PR #1, opened 2017-09-05
PIP dependency support

mattmoor:
This is a proposal for some PIP rules that are heavily based on a prototype I wrote a month or two back.  Very open to suggestions here, but it seems to work across a number of examples I've tried so far.


README.md
```diff
@@ -39,6 +47,51 @@ py_binary(
...
+
+## Consuming `pip` dependencies
+
+Once a set of dependencies has been imported via `pip_import` and `pip_install`
+we can start consuming them in our `py_{binary,library,test}` rules.  In support
+of this, the generated `requirements.bzl` also contains a `packages` method,
+which can be used directly in `deps=[]` to reference an imported `py_library`.
+
+```python
+load("@my_deps//:requirements.txt", "packages")
```
damienmg: You probably meant `requirements.bzl` here

python/pip.sh
```diff
@@ -0,0 +1,84 @@
+#!/bin/bash -e
```
damienmg: General comment for this file: the file already complex, why not in python instead?

python/whl.sh
```diff
@@ -0,0 +1,82 @@
+#!/bin/bash -e
+
```
damienmg: Same here, bash can become quickly hard to read...

.gitignore
```diff
@@ -0,0 +1,39 @@
+# Compiled Object files
```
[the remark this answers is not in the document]
mattmoor: This is just my standard `.gitignore` file I copy from repo to repo.  I never touch it again.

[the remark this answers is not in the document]
damienmg: Repo .gitignore are supposed to be repo specific, so not includes things like *~...

README.md
mattmoor: Thanks

.travis.yml
```diff
@@ -0,0 +1,27 @@
+sudo: required
```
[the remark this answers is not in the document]
mattmoor: Done

python/pip.sh
mattmoor: See line 17.  I was halfway through rewriting one of these scripts in Python before I realized: I can't invoke a `py_binary` during `WORKSPACE`, I'd have to publish a PAR file.  I immediately `git stash`ed and rewrote my TODO instead. :(

python/whl.sh
mattmoor: Ack

python/pip.sh
damienmg: :( I actually do things without py_binary for simple python_script (I know that is a shame) using simply python <script.py>. Anyway up to you.

mattmoor: I suppose that is true.  I want to be able to factor libraries and have a clean separation of unit tests etc.  I'll experiment with my stashed changes with this in mind.

mattmoor:
@duggelz Do you have any suggestions on the best way to download and invoke `pip` without using `get-pip.py` to install it as a system-wide dependency?  This seems to be why the OSX build is failing.

mattmoor:
@sebgoa You'd asked for this at one point, if you are still interested I'd love your feedback.

.gitignore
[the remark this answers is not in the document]
duggelz: I advice I got, didn't like, but eventually learned to appreciate the wisdom of, is to create a ~/.gitignore_global with all the editor autosave filenames, etc (things that are particular to me, not to the repo I'm working on), and do `git config --global core.excludesfile ~/.gitignore_global`

README.md
```diff
@@ -48,3 +101,47 @@ See Bazel core [documentation](https://docs.bazel.build/versions/master/be/pytho
...
 
 See Bazel core [documentation](https://docs.bazel.build/versions/master/be/python.html#py_binary).
+
+<a name="py_test"></a>
+## py_test
+
+See Bazel core [documentation](https://docs.bazel.build/versions/master/be/python.html#py_test).
+
+<a name="pip_import"></a>
+## pip_import
```
duggelz: There's a thing called [skydoc](https://github.com/bazelbuild/skydoc) that will turn structured docstrings into Bazel documentation.  It's kind of basic and a bit, well, not beautiful.  https://google.github.io/subpar/subpar.html

mattmoor: I'll TAL at skydoc for this repo, but in `rules_docker` and `rules_k8s` I'm blocked on [this](https://github.com/bazelbuild/rules_docker/issues/32#issuecomment-324164587)

mattmoor: Ok, I've switched over to skydoc for the docs here (based on how google/subpar does it).  Feedback welcome :)

mattmoor:
I moved `whl.sh` into a little Python script that I can invoke directly, but also reasonably unit test.

I have not moved `pip.sh` yet, and thinking about it left me with an idea.  Considering:
* Shelling out to `pip` from Python is a bit silly,
* @damienmg's trick won't work if I have a `pip` library dependency,
* I still haven't solved the `pip` fetching problem.

**What if we build / publish a `PAR` (similar to the process `rules_docker` uses), which links the `pip` library, and removes the need to `get-pip.py` somehow?  WDYT?**

I still haven't found a clean process for doing this from a single repo (`google/containerregistry` and `bazelbuild/rules_docker` are separate repos), but that seems like a smaller problem to solve.

mattmoor:
I found [this](https://github.com/pypa/get-pip/blob/master/template.py#L164-L168), which may be relevant to why certs aren't working properly from `pip` in a PAR file.  It is similar to [the trick](https://github.com/google/containerregistry/blob/master/tools/patched_.py#L28) we do for `httplib2` in `google/containerregistry` tools.

AlecBenzer:
I tried this out and it worked for me! Yay! :D 

Few questions:

1. It feels a little verbose? I don't understand all the details, but naively it seems like it should be possible to specify the requirements either only in a `requirements.txt` or only in the `BUILD` files, not both? Is there a way to use this omitting the explicit `requirements.txt` and to infer the dependencies from the BUILD files? 
    
    Extra work that needs to be done here seems more relevant just because it needs to be done every time a new dependency is added.
1. The overall structure feels a little cumbersome? I.e., in `WORKSPACE`, importing one rule to generate a package that we then import which contains another rule that we invoke, and then importing the generated package within `BUILD` and using the `packages` call. Is there any way to simplify the number of pieces here?

    Though admittedly I think this is probably less important since it's mostly just a one-time setup thing.

README.md
damienmg: Not updated? Maybe you just want to delete that part all in profit of the skydoc generated.

mattmoor:
@AlecBenzer The two-phase import pattern is hard (maybe impossible?) to avoid and still allow users to use `requirements.txt` as their source of truth.  I completely agree it is a mouthful.

My bias towards this approach is driven by:
 * Once this boilerplate is in place, you (should) never need to touch `WORKSPACE` again (it is 100% driven by `requirements.txt`)
 * The boilerplate code basically doesn't change repo to repo (unless you start juggling multiple `requirements.txt` files).  **My hope is that this makes the "it's a mouthful" less of an issue** because they can copy/paste directly from our `README.md` and that should work 99% of the time.
 * Enabling the use of `requirements.txt` directly reduces the number of concepts new Python users need to grok in order to start using Bazel.


*(thinking out loud)*
We could certainly consider an approach like Gazelle, where we would (when run manually) generate files that you would then check into your tree.  In our case `WORKSPACE,` in their case `BUILD`.  My issue with this is that it forks the source of truth.  `WORKSPACE` is a copy of `requirements.txt` and you have to build your own processes to ensure they stay in sync.

I think my preference would be to have these as users' starter rules (less error prone), and perhaps support something like Gazelle for when they have reached the point of wanting to switch off of `requirements.txt` as their source of truth and use pure Bazel.

That said, I don't think `.whl` (or pypi's hosting of them) is ready for Bazel to consume it directly since some of the time `pip wheel` will actually build the `.whl` from the legacy packaging, which complicates things.  When `.whl` coverage is more comprehensive (and the tooling can just support telling us the list of URLs for `.whl` files), then perhaps we can consider this option.


I'm ofc happy to discuss, I'd love to do even better!  At first I didn't even think using the language-idiomatic source of truth was possible at all until this pattern dawned on me (after it ate away at my brain for a number of hours of background thought).

README.md
mattmoor: Ha, I think this bug was in two places.  thanks for catching it again :)

python/whl.sh
```diff
@@ -0,0 +1,82 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+# TODO(mattmoor): Ideally we'd rewrite this in Python for portability, but
+# since the executes during WORKSPACE instantiation that would require
+# redistributing PAR files, since we cannot invoke built tools at that time.
+WHL="$1"
```
duggelz: `# Label of wheel file`

python/whl.sh
```diff
@@ -0,0 +1,82 @@
...
+REQUIREMENTS="$2"
+PKG=$(basename "${WHL}" | cut -d'-' -f 1)
+DIST_INFO="$(basename "${WHL}" | cut -d'-' -f -2).dist-info"
+
+unzip "${WHL}" > /dev/null
```
duggelz: I guess I should know this, but what directory do genrules run in these days?  Some arbitrary empty directory?

python/whl.sh
```diff
@@ -0,0 +1,82 @@
...
+# limitations under the License.
+
+
+# TODO(mattmoor): Ideally we'd rewrite this in Python for portability, but
+# since the executes during WORKSPACE instantiation that would require
+# redistributing PAR files, since we cannot invoke built tools at that time.
+WHL="$1"
+REQUIREMENTS="$2"
+PKG=$(basename "${WHL}" | cut -d'-' -f 1)
+DIST_INFO="$(basename "${WHL}" | cut -d'-' -f -2).dist-info"
```
duggelz: `# E.g. google_cloud-0.27.0-py2.py3-none-any.whl` -> `google_cloud-0.27.0.dist-info`

python/whl.sh
```diff
@@ -0,0 +1,82 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+# TODO(mattmoor): Ideally we'd rewrite this in Python for portability, but
+# since the executes during WORKSPACE instantiation that would require
+# redistributing PAR files, since we cannot invoke built tools at that time.
+WHL="$1"
+REQUIREMENTS="$2"
```
duggelz: `# Label of requirements.bzl file`

python/whl.sh
```diff
@@ -0,0 +1,82 @@
...
+unzip "${WHL}" > /dev/null
+
+# Enumerate the names of dependencies from the WHL metadata.
+function dependencies() {
+    cat | python <<EOF
+import json
+
+# TODO(mattmoor): find a schema / specification to follow.
+# Why are there so many layers?
+with open("${DIST_INFO}/metadata.json", "r") as f:
```
duggelz: There are some specs that contradict each other a little bit, and wheel and pip don't exactly follow any of them.  So, I'd go by what actually works empirically.

python/whl.sh
```diff
@@ -0,0 +1,82 @@
...
+  reqs = json.loads(f.read()).get("run_requires", [])
```
duggelz: In the interests of Python 2/3 compatibility, I'd suggest some minor changes:

```
import io
import json

with io.open("${DIST_INFO}/metadata.json", "r", encoding="utf8") as f:
    for req in reqs:
    requires = req.get("requires", [])
        for entry in requires:
            (name, constraint) = entry.split(" ", 1)
            print(name)
```

Also, add a TODO that we should do something about version constraints.

python/whl.sh
```diff
@@ -0,0 +1,82 @@
...
+      (name, constraint) = entry.split(" ", 1)
+      print name
+EOF
+}
+
+DEPS=($(dependencies))
+
+function name() {
+    cat | python <<EOF
+import json
```
duggelz:
```
import io
import json

with io.open("${DIST_INFO}/metadata.json", "r", encoding="utf8") as f:
    print(json.loads(f.read())["name"])
```

duggelz:
Sheesh, all that, and I was review the wrong commit.

README.md
```diff
@@ -39,6 +47,51 @@ py_binary(
...
+```python
+load("@my_deps//:requirements.txt", "packages")
+
+py_library(
+    name = "mylib",
+    srcs = ["mylib.py"],
+    deps = [
+        ":myotherlib",
+	# This takes the name as specified in requirements.txt
+	packages("importeddep"),
```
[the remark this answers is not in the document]
mattmoor: Renamed.  This is a vestige of a time when I was exporting `packages` as a dict of packages, but I had to make it a function to allow passing in `_` instead of `-`.

examples/helloworld/helloworld_test.py
```diff
@@ -0,0 +1,41 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import unittest
+
+from examples.helloworld import helloworld
+
+
+class HelloWorldTest(unittest.TestCase):
```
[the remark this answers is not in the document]
mattmoor: Hmm, I started to add a test based on [this](https://stackoverflow.com/questions/20180543/how-to-check-version-of-python-modules), which works for other modules, but doesn't seem to work for `futures`...

```shell
$ python
Python 2.7.6 (default, Oct 26 2016, 20:30:19) 
[GCC 4.8.4] on linux2
>>> from concurrent import futures
>>> print futures.__version__
  File "<stdin>", line 1, in <module>
AttributeError: 'module' object has no attribute '__version__'
>>> 
```

I've added a new example that uses `pip` at an exact version (not what I have on my system), which seems to work.

python/pip.bzl
```diff
@@ -0,0 +1,54 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""A rule for importing Python dependencies via requirements.txt."""
+
+def _import_impl(repository_ctx):
+  """Core implementation of pip_import."""
+
+  # Add an empty top-level BUILD file.
+  repository_ctx.file("BUILD", "")
```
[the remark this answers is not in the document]
mattmoor: I'm not sure I understand your concern.  This file is being created within the context of the repository, not in the user's main build tree.  I only create this because otherwise skylark rules within the repository cannot be loaded.

I added a better comment to clarify.

python/pip.bzl
```diff
@@ -0,0 +1,54 @@
...
+
+  result = repository_ctx.execute([
+    repository_ctx.path(repository_ctx.attr._script),
+    repository_ctx.attr.name,
+    repository_ctx.path(repository_ctx.attr.requirements),
+    repository_ctx.path("requirements.bzl"),
+    repository_ctx.path(""),
+  ])
+  if result.return_code:
+    fail("pip_import failed: %s (%s)" % (result.stdout, result.stderr))
```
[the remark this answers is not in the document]
mattmoor: AIUI this is still challenging to do *through Bazel*.  I'd love to add this kind of negative checking to several rules repos, and have already resorted to bash-script-on-Travis for testing things like `bazel run` does the right thing.  @damienmg Has this improved (or even moved) at all in the last 6 months?

python/pip.bzl
```diff
@@ -0,0 +1,54 @@
...
+            mandatory = True,
+            single_file = True,
+        ),
+        "_script": attr.label(
+            executable = True,
+            default = Label("//python:pip.sh"),
+            cfg = "host",
+        ),
+    },
+    implementation = _import_impl,
```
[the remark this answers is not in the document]
mattmoor: Done

python/pip.sh
mattmoor: The only remaining shell script now does this (I copied it from you :D)

python/pip.sh
```diff
@@ -0,0 +1,84 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# TODO(mattmoor): Ideally we'd rewrite this in Python for portability, but
+# since the executes during WORKSPACE instantiation that would require
+# redistributing PAR files, since we cannot invoke built tools at that time.
+NAME="$1"
```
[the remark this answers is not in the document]
mattmoor: Ack, this is now in Python.  Sorry for the cross signals here, but hopefully the Python is clearer to follow (except perhaps my import-fallback logic in `piptool.py`, but this is temporary if/until I can switch to a PAR we distribute).

python/pip.sh
```diff
@@ -0,0 +1,84 @@
...
+REQUIREMENTS_TXT="$2"
+REQUIREMENTS_BZL="$3"
+REPOSITORY_DIR="$4"
+
+# TODO(mattmoor): Switch to a version of PIP we download.
+pip wheel -w "${REPOSITORY_DIR}" -r "${REQUIREMENTS_TXT}"
```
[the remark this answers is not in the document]
mattmoor: Do you have a recommendation for this in the fallback in `piptool.py`?

python/whl.sh
mattmoor: Added to `whl.py`

python/whl.sh
mattmoor: Repository rules (a bit different) run in the context that will become the new repository, so this is unpacking the `.whl` into a parallel source tree.

python/whl.sh
mattmoor: That's what I've been doing :)

I expect this will be a little bit of whackamole as folks try things out.

python/whl.sh
mattmoor: This changed a little bit in `piptool.py`, it's now using `zipfile.open`.

I look forward to you beating me up on the best ways to write Python 2/3 (and ideally enforce cross-compatibility, if it's possible in Bazel).  Apologies again for the crossed signals.

mattmoor:
@damienmg Can we add a buildifier check to the Jenkins job, or do I have to set up Travis for that?

damienmg:
Not right now unfortunately. I want to add it (bazelbuild/continuous-integration#13) but we kind of need bazelbuild/continuous-integration#100.

duggelz:
Design question for you:

Assumption: Bazel isn't doing some super-clever multi-level hierarchical WORKSPACE namespacing that I'm not aware of.

Scenario setup:

Workspace A depends on pip package P1 at version V1.  Workspace B depends on pip package P1 at version V2.  Workspace C depends on A and B via `git_repository()` rules in c/WORKSPACE file.  A and B have a_install() and b_install() functions that are also called in c/WORKSPACE.  A and B have py_xxx targets that C will use in some manner.

Scenario 1: V1 = V2

Result: Bazel rejects the configuration because a_install and b_install create duplicate targets.

Question: Can B detect that it doesn't need to create P1 target?

Scenario 2: V1 != V2

Result: Same as Scenario 1

Scenario 3: V1 != V2, and we only create P1 once

Result: Build succeeds, but Python programs in C might get V1 or V2 depending on the order of things in c/WORKSPACE.

Scenario 4: V1 != V2, and we incorporate versions into target labels like `mock_2_0_0` instead of `mock`, so we create two P1 sets of targets.

Now assume a py_binary in C depends on a py_library from A and another py_library from B.

Result: Build succeeds, but Python programs in C can depend on P1.V1 and P1.V2 without Bazel realizing there is a clash.  Python import semantics mean that the program might get files from V1, V2, or both, at runtime, in fairly hard to predict ways.

Note that current Google repos that manually depend on pip packages in WORKSPACE via `http_repository` and similar, use both naming conventions (`@mock` and `@mock_2_0_0`), so we actually get all of the problems described above.

Scenario 5: We require that V1 = V2, and make it work correctly.

Result: A, B and C are now version-deadlocked.  C can never update unless A and B happen to both update to the same V3.  Now multiply by the combinatorial explosion of dependencies and versions that we find in the real world.

Suggestion: I don't think this can be completely solved without making new py_library() and py_binary() rules that delay resolving py_library() deps until the final py_binary() step.

In the meantime:  I prefer baking version numbers into package labels (e.g. `@mock_2_0_0//` instead of `@mock`) so we don't have deadlock.  The missing piece is something in Bazel that looks at deps, or runtimes, or I'm not sure what, and complains loudly if it finds `@foo_anything` and `@foo_anything_else` in the same target.

mattmoor:
@duggelz I think that's a **very** good question, and I think it breaks down where you have:
> A and B have a_install() and b_install() functions that are also called in c/WORKSPACE

As you are clearly driving at, we have 3 sets of requirements that are being merged **after the constraints have been solved**, which I think is a losing proposition.

I think that my expectation is that instead of `a_install` and `b_install` we would `pip_import` a blended `requirements.txt` from `A` and `B`, and just `c_install`.

This story is only half-baked because:
1. We don't have a rule for combining `@foo//:requirements.txt` and `@bar//:requirements.txt` (but I suspect this is pretty easy).
1. `A`, `B`, and `C` may each references this collection of dependencies with different repository names (but I believe this is doable by aliasing repositories with [`bind`](https://docs.bazel.build/versions/master/be/workspace.html#bind)).  At one point, with this exact problem in mind, I debated removing users capacity to name the repository itself, which would effectively guarantee a single resolution step, but I thought this was a bit extreme (e.g. what if I want to write an invoked tool that has its own `requirements.txt`, that's ok).

---
I think the crux of this issue is that `WORKSPACE` currently serves two roles (I hint at this above):
1. Resolving dependencies based on fuzzy constraints (incl. transitive resolution).
1. Fetching dependencies based on concrete specification.

The analogs to this outside of Bazel are things like `[frozen] requirements.txt` or `package[-lock].json`.

I think the guidance should be for users not to mix independently resolved external dependencies, but I'm not 100% sure if/how we can enforce it.  Perhaps we can do this through Bazel's constraint system, but I have not looked any further at how we might express this.


Is what I'm describing clear?  Does my logic make sense?

duggelz:
@mattmoor I agree with what you wrote.  Some additional points:

1. Bazel has a native.existing_rules that make fixing my points 1 and 2 easy.
2. Looking at various repository rules, there is a nice [omit_foo pattern](https://github.com/bazelbuild/rules_closure/blob/master/closure/repositories.bzl) that allows C to override A and B as desired.
3. The conflict between "fuzzy constraints" and "concrete specification" is real.  Python aims to address this with `Pipfile` vs `Pipfile.lock`, but meanwhile we muddle through.

I don't think we need to make a tool that tries to merge A and B automatically or anything.  I think it's reasonable to say that C must provide a `requirements.txt` that satisfies every need that A and B have.

After reading some discussions about Java/Maven, I do think that we should put the names in a canonical format that we document and agree on.  Specifically, I think we should have the literal prefix `pypi_` on names that would otherwise be very short and ambigous.  E.g. `@pypi__package_name_here//:package_name` or `@pypi__package_name_here__version_here//:package_name`

mattmoor:
@duggelz Given that things are hidden behind `package(...)` in the expected usage today, I think this is completely fine.  The main change is that we wouldn't namespace packages based on the enclosing `pip_import`, which I think is fine.

I've gone with `pypi__{distribution}_{version}`, used `existing_rules` to avoid multiple imports, but each `pip_import` will still exposes the library through `package()` even if it didn't import it.  I included a section in the `README.md` covering canonical naming.

mattmoor:
@duggelz Hooray, green CI :)

PR #9, opened 2017-09-20
always lowercase requirements dictionary keys

justinwp:
not sure if this has any other ramifications

### DIFF
--- a/rules_python/piptool.py
+++ b/rules_python/piptool.py
@@ -141,12 +141,12 @@ _requirements = {{
 all_requirements = _requirements.values()
 
 def requirement(name):
-  name = name.replace("-", "_")
+  name = name.replace("-", "_").lower()
   return _requirements[name]
 """.format(input=args.input,
            whl_libraries='\n'.join(map(whl_library, whls)),
            mappings=','.join([
-             '"%s": "@%s//:pkg"' % (wheel.distribution(), wheel.repository_name())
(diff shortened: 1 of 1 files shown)


bazel-io:

justinwp:
I signed it!

duggelz:
From [PEP 508](https://www.python.org/dev/peps/pep-0508/) it appears that you can register packages with any capitalization, and install them with any capitalization, even if it's different.  I.e. you can register package 'MyPackage' and install it with `mypackage`, `MYPACKAGE`, or even `mYpAcKaGe`.

That is not an ideal spec, but is it the spec, and the tools implement it, so we should to.

duggelz:
Patch looks ok to me, but we'll want to add a test (package `PyYAML` seems a good choice), and then an admin (@mattmoor) will need to rebuild the piptool.par file.

duggelz:
Sorry, that was actually a "yes this seems like a good patch", not a "no".  We should accept `requirements("PyYAML")` and normalize it to "pyyaml" like pip does.

duggelz:
I guess the real test would be if some library had `requirements("PyYAML")` and some other library had `requirements("pyyaml")`.  We would want to realize that those are actually the same thing.

mattmoor:
This LGTM, not sure why @justinwp closed/deleted, but I'll recreate it with an updated PAR.  thanks Justin.

PR #10, opened 2017-09-20
Add a simple test that the built and checked in PAR files match.

mattmoor:


mattmoor:
@duggelz Let's see if Bazel CI thinks this is gonna work :)

mattmoor:
@damienmg what is going on here?

damienmg:
Tests failed: http://ci.bazel.io/blue/organizations/jenkins/PR%2Frules_python/detail/rules_python/37/tests (fwiw Jenkins team is finally working on making it easier to mark the good step at failing). When you see the job Is yellow, that means that tests have failed and should have a test report somewhere).

mattmoor:
@duggelz Hmm, @duggelz any sense for what might be making this non-deterministic?

duggelz:
Doh.  I was going to add a non-determinism test a while back but I couldn't get Bazel-in-bazel to work.  Apparently I never filed a bug, and then it slipped off my radar.

The non-determinism stems from the timestamps of the files inside the compressed archive.  This is fixed in the internal tool by setting all timestamps to 12:00:01 AM, Jan 1, 1980 (there's a reason for this specific date).  I will do the same.

mattmoor:
The failure here is the same we were seeing with `Flask`.  I'm rebasing this on [this PR](https://github.com/bazelbuild/rules_python/pull/40) to see if that unblocks having the test here enabled.

mattmoor:
@duggelz At long last, this should be RFAL :)

thanks for fixing this :D

PR #11, opened 2017-09-20
Allow any capitalization in requirement names.

mattmoor:
Quoting @duggelz

> From PEP 508 it appears that you can register packages with any capitalization, and install them with any capitalization, even if it's different. I.e. you can register package 'MyPackage' and install it with mypackage, MYPACKAGE, or even mYpAcKaGe.
>
> That is not an ideal spec, but is it the spec, and the tools implement it, so we should to.

Thanks to @justinwb for the original PR.


bazel-io:

mattmoor:
@damienmg Any idea why the bot doesn't recognize my super powers? :D

mattmoor:
test this please

damienmg:
(seems like it did :p)

PR #15, opened 2017-09-21
Use the lowercased version of package names

aronasorman:
Some packages use the lowercased version of their dependencies, and pip would install them. So even if I have `pillow==3.8.1` in my `requirements.txt`, it would still work.

However, it looks like that `pip_import` always uses the case sensitive name of a package. So even if I have `django==1.10` in my requirements.txt, `pip_import` would always name the package as `Django` because that's what pip uses.

This fixes that issue by always using the lowercase version of packages in the `_requirement` dictionary.


bazel-io:

aronasorman:
I signed it!

mattmoor:
Is this a duplicate of https://github.com/bazelbuild/rules_python/pull/11 ?

mattmoor:
The diff matches, so I'm going to close this and bug @duggelz :D

Issue #20, opened 2017-10-04
Pip fails dictionary lookup.


afking:
Requirements file of `boto3==1.4.7` fails with
```
ERROR: /private/var/tmp/_bazel_edwardmcfarlane/ad32b996cb424fa4cc44a5c86220cb2f/external/pypi__boto3_1_4_7/BUILD:6:1: Traceback (most recent call last):
	File "/private/var/tmp/_bazel_edwardmcfarlane/ad32b996cb424fa4cc44a5c86220cb2f/external/pypi__boto3_1_4_7/BUILD", line 6
		py_library(name = "pkg", srcs = glob(["**/*.p..."]), <3 more arguments>)
	File "/private/var/tmp/_bazel_edwardmcfarlane/ad32b996cb424fa4cc44a5c86220cb2f/external/pypi__boto3_1_4_7/BUILD", line 13, in py_library
		requirement("botocore>=1.7.0,<1.8.0")
	File "/private/var/tmp/_bazel_edwardmcfarlane/ad32b996cb424fa4cc44a5c86220cb2f/external/pip/requirements.bzl", line 150, in requirement
		_requirements[name]
key "botocore>=1.7.0,<1.8.0" not found in dictionary.
```

afking:
Would be nice to be to fix all sub dependencies. For example `botocore==1.7.22` that would satisfy version range.

mattmoor:
Hmm, this is an artifact of how we are parsing the `metadata.json` from the `.whl` file, typically this version information is expressed differently.  I will try to TAL and see how hard a fix this is.

mattmoor:
Ok, I think that this is hopefully just switching things over to use `re.split` and expanding the character set.  I'll hopefully have a PR you can test shortly.  (famous last words)

afking:
@mattmoor awesome! Works, thanks for the quick fix.

afking:
@mattmoor think it might be an error on my side but failing to build on Ubuntu, mac is working fine. Fails with:
```
bazel build //:my_python_lib
ERROR: error loading package '': Encountered error while reading extension file 'requirements.bzl': no such package '@pip//': pip_import failed:  (ERROR: 'pip wheel' requires the 'wheel' package. To fix this, run: pip install wheel
)
ERROR: error loading package '': Encountered error while reading extension file 'requirements.bzl': no such package '@pip//': pip_import failed:  (ERROR: 'pip wheel' requires the 'wheel' package. To fix this, run: pip install wheel
[...]

PR #21, opened 2017-10-04
Split on more characters when stripping the version metadata.

mattmoor:
`boto3`s `metadata.json` file doesn't delimit the version specifier the way other packages I've come across seem to, so support splitting on the kinds of characters this variation might foist upon us.

Fixes: https://github.com/bazelbuild/rules_python/issues/20


mattmoor:
@damienmg Can you review?  @duggelz is out of the office.

damienmg:
No tests?

mattmoor:
@damienmg `boto_test.py` is the test.  I want to make sure we can build and import it, by the time that happens I've covered what I want :)

PR #22, opened 2017-10-05
Correctly decode metadata.json into a utf8 str

kgreenek:
Without this, the metadata.json file is read as bytes instead of a str, and the call
to json.loads() fails with the following error:
```bash
TypeError: the JSON object must be str, not 'bytes'
```

### DIFF
--- a/rules_python/whl.py
+++ b/rules_python/whl.py
@@ -58,7 +58,7 @@ class Wheel(object):
     # directory.
     with zipfile.ZipFile(self.path(), 'r') as whl:
       with whl.open(os.path.join(self._dist_info(), 'metadata.json')) as f:
-        return json.loads(f.read())
+        return json.loads(f.read().decode("utf-8"))
 
   def name(self):
     return self.metadata().get('name')


bazel-io:

kgreenek:
I signed it!

rules_python/whl.py
```diff
@@ -58,7 +58,7 @@ def metadata(self):
     # directory.
     with zipfile.ZipFile(self.path(), 'r') as whl:
       with whl.open(os.path.join(self._dist_info(), 'metadata.json')) as f:
-        return json.loads(f.read())
+        return json.loads(f.read().decode("utf-8"))
```
duggelz: Looking at the `json.loads` function, it seems like this should (originally) have just been `return json.load(f)` (note change from `loads` to `load`), which handles all the encoding/decoding/python2-vs-3.  Can you try that and see if it does indeed work for you?

kgreenek: Tried it, and it actually didn't work. When I called load instead of loads, I got the following traceback:
```bash
  File "/home/kevin/.cache/bazel/_bazel_kevin/cb16c580e8d350a5b65c591962f69d45/external/io_bazel_rules_python/rules_python/whl.py", line 131, in <module>
    main()
  File "/home/kevin/.cache/bazel/_bazel_kevin/cb16c580e8d350a5b65c591962f69d45/external/io_bazel_rules_python/rules_python/whl.py", line 127, in main
    for d in whl.dependencies()
  File "/home/kevin/.cache/bazel/_bazel_kevin/cb16c580e8d350a5b65c591962f69d45/external/io_bazel_rules_python/rules_python/whl.py", line 126, in <listcomp>
    'requirement("%s")' % d
  File "/home/kevin/.cache/bazel/_bazel_kevin/cb16c580e8d350a5b65c591962f69d45/external/io_bazel_rules_python/rules_python/whl.py", line 68, in dependencies
    run_requires = self.metadata().get('run_requires', [])
  File "/home/kevin/.cache/bazel/_bazel_kevin/cb16c580e8d350a5b65c591962f69d45/external/io_bazel_rules_python/rules_python/whl.py", line 61, in metadata
    return json.load(f.read())
  File "/usr/lib/python3.5/json/__init__.py", line 265, in load
    return loads(fp.read(),
AttributeError: 'bytes' object has no attribute 'read'
```

kgreenek: Oh oops, I was passing "f.read()" to the load function instead of passing "f". I tried again, this time correctly, and I get the same str/bytes related error, but in a deeper function call. Here's the traceback:
```bash
  File "/home/kevin/.cache/bazel/_bazel_kevin/cb16c580e8d350a5b65c591962f69d45/external/io_bazel_rules_python/rules_python/whl.py", line 131, in <module>
    main()
  File "/home/kevin/.cache/bazel/_bazel_kevin/cb16c580e8d350a5b65c591962f69d45/external/io_bazel_rules_python/rules_python/whl.py", line 127, in main
    for d in whl.dependencies()
  File "/home/kevin/.cache/bazel/_bazel_kevin/cb16c580e8d350a5b65c591962f69d45/external/io_bazel_rules_python/rules_python/whl.py", line 126, in <listcomp>
    'requirement("%s")' % d
  File "/home/kevin/.cache/bazel/_bazel_kevin/cb16c580e8d350a5b65c591962f69d45/external/io_bazel_rules_python/rules_python/whl.py", line 68, in dependencies
    run_requires = self.metadata().get('run_requires', [])
  File "/home/kevin/.cache/bazel/_bazel_kevin/cb16c580e8d350a5b65c591962f69d45/external/io_bazel_rules_python/rules_python/whl.py", line 61, in metadata
    return json.load(f)
  File "/usr/lib/python3.5/json/__init__.py", line 268, in load
    parse_constant=parse_constant, object_pairs_hook=object_pairs_hook, **kw)
  File "/usr/lib/python3.5/json/__init__.py", line 312, in loads
    s.__class__.__name__))
TypeError: the JSON object must be str, not 'bytes'
```

duggelz: Ok, then what you originally have is fine.

Can you tell me what operating system and Python version you're using?  There's a variety of important differences between 2.6, 2.7, and 3.x that are relevant to this bit of code.

We don't have good test coverage for those differences either, which is a separate issue.

kgreenek: I am on Ubuntu 16.04.

I tested this with python 3.5.2 and 2.7.12.

To test, I used a virtualenv with python3 installed. Built my sample repo using bazel, and then ran my binary. Then I ran "bazel clean --expunge", opened a new terminal without the virtualenv activated (i.e. python 2.7), rebuilt with bazel, and re-ran. Both worked for me.

I also confirmed that without this patch, the bazel build step fails with python3, but works with python2.7.

mattmoor:
Test this please

kgreenek:
@mattmoor I tested with both python 3 and python 2.7 (see my comment in the thread in whl.py). Is that good enough for this PR?

mattmoor:
@kgreenek Sorry, that's supposed to be the magic phrase to kick Bazel's Jenkins instance, but clearly it's not working.  @damienmg what's going on with Bazel CI?

damienmg:
Someone broke it apparently, I guess I'll have a not fun workday tomorrow.

damienmg:
This should be fixed.

Jenkins, test this please.

mattmoor:
Thanks!

PR #39, opened 2017-11-13
Add a better error message when deps not found

hwright:
Use the skylark `fail` function instead of letting Python throw a
key-not-found exception.

(There may be a prettier way of doing this: I'm happy to be shown
differently, or to bikeshed the prose.)


bazel-io:

mattmoor:
Please test this

rules_python/piptool.py
```diff
@@ -138,8 +138,10 @@ def pip_install():
 all_requirements = _requirements.values()
 
 def requirement(name):
-  name = name.replace("-", "_").lower()
-  return _requirements[name]
+  name_key = name.replace("-", "_").lower()
+  if name_key not in _requirements:
+    fail("Could not find pip-provided dependency: '%s'" % name)
```
mattmoor: Maybe it would be interesting to include what's available?

hwright: We could, but since that list is unbounded, I'm concerned that it would be too much for this error string.

mattmoor:
Test this please

rules_python/piptool.py
mattmoor: This is Python, not Node.js ;-)

SGTM

PR #40, opened 2017-11-16
This updates WORKSPACE to the latest subpar, and ./update_tools.sh.

mattmoor:
This also fixes a couple bugs I hit after regenerating `piptool.par`:
 - `os.environ['PYTHONPATH']` may result in a `KeyError`
 - The cert expansion logic was removed, but things fail without it (after I `./update_tools.sh`).  I thought this was intentional (should the PAR changes have fixes this?), but I'm restoring for now to keep `HEAD` in a good state.


mattmoor:
@duggelz This makes `bazel test ...` clean after `./update_tools.sh` (it is broken at `HEAD`).  The subpar update is an artifact of me thinking that might fix the certs issue, but it didn't seem to.

I'd like to reach a point where we have [this](https://github.com/bazelbuild/rules_python/pull/10) done, so that CI makes sure the above passes (since `./update_tools.sh` should do nothing!).

LMK if you want to dig into this at all, I'm just trying to get HEAD into a good state.

WORKSPACE
```diff
@@ -52,7 +52,7 @@ _piptool_install()
 git_repository(
     name = "subpar",
     remote = "https://github.com/google/subpar",
-    tag = "1.0.0",
+    commit = "b1b6bd13ff80ede09f57539031cfb20459c28237",
```
duggelz: I tagged 1.1.0 if you want to use that instead (You'll miss the CI changes which are irrelevant here).

rules_python/piptool.py
```diff
@@ -48,8 +48,10 @@ def extract_packages(package_names):
     # Add extracted directories to import path ahead of their zip file
     # counterparts.
     sys.path[0:0] = dirs_to_add
-    existing_pythonpath = os.environ['PYTHONPATH'].split(':')
-    os.environ['PYTHONPATH'] = ':'.join(dirs_to_add + existing_pythonpath)
+    existing_pythonpath = os.environ.get('PYTHONPATH')
+    if existing_pythonpath:
+        dirs_to_add.append(existing_pythonpath.split(':'))
```
duggelz: `append` will do the wrong thing here (adds the whole list as a single element of the first list), you want `extend`

rules_python/piptool.py
```diff
@@ -68,7 +70,12 @@ def extract_packages(package_names):
 
 
 def pip_main(argv):
-    argv = ["--disable-pip-version-check"] + argv
+    # Extract the certificates from the PAR following the example of get-pip.py
+    # https://github.com/pypa/get-pip/blob/430ba37776ae2ad89/template.py#L164-L168
+    cert_path = os.path.join(tempfile.mkdtemp(), "cacert.pem")
+    with open(cert_path, "wb") as cert:
+      cert.write(pkgutil.get_data("pip._vendor.requests", "cacert.pem"))
+    argv = ["--disable-pip-version-check", "--cert", cert_path] + argv
```
duggelz: Yes, this shouldn't be necessary, but ok to add them back in while I investigate more.

duggelz:
Oops, can I unapprove?  See my comment about `append`.

rules_python/piptool.py
mattmoor: thanks, good catch.

mattmoor:
Updated, thanks.

mattmoor:
@duggelz To be clear, I'll hold off until you re-approve :)

duggelz:
LGTM to me, submit.  (I still can't find any button to unapprove or re-approve).

PR #46, opened 2017-11-27
Split the extras param by comma

jac-stripe:
When the bazel implementation calls into the wheel python script, it joins the extras arg by commas (https://github.com/bazelbuild/rules_python/blob/3e167dcfb17356c68588715ed324c5e9b76f391d/python/whl.bzl#L27).

When that arg is parsed, it should be split by commas. In the case of multiple args being passed with the same argument name, we should continue to use an Append parser, but with modifications to split by comma.

### DIFF
--- a/python/whl.bzl
+++ b/python/whl.bzl
@@ -24,7 +24,10 @@ def _whl_impl(repository_ctx):
   ]
 
   if repository_ctx.attr.extras:
-    args += ["--extras", ",".join(repository_ctx.attr.extras)]
+    args += [
+      "--extras=%s" % extra
+      for extra in repository_ctx.attr.extras
+    ]
 
   result = repository_ctx.execute(args)
   if result.return_code:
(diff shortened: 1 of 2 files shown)


bazel-io:

mattmoor:
This is a pretty easy fix in `whl.bzl`, which I prefer to relying on a private class in `argparse`.

I would change the line you linked to:
```python
   args += [
       "--extras=%s" % extra
       for extra in repository_ctx.attr.extras
   ]
```

This should do the right thing for `argparse`.  Switching between this and `gflags` all the time trips me up... :-/

mattmoor:
I should also have said: thanks for the PR!

jac-stripe:
Great suggestion, updated

mattmoor:
Test this please

(for the Bazel CI bot)

mattmoor:
thanks again for the PR.

Issue #49, opened 2017-12-05
Direct dependency `funcsigs` of pip_imported `mock` package not available


nikhaldi:
A minimal repro needs multiple files (also in the attachment [funcsigs-repro.tar.gz](https://github.com/bazelbuild/rules_python/files/1533133/funcsigs-repro.tar.gz) for convenience):

WORKSPACE:
```
git_repository(
    name = "io_bazel_rules_python",
    remote = "https://github.com/bazelbuild/rules_python.git",
    commit = "346b898e15e75f832b89e5da6a78ee79593237f0"
)

load("@io_bazel_rules_python//python:pip.bzl", "pip_import")
pip_import(
   name = "requirements",
   requirements = "//:requirements.txt",
)

load("@requirements//:requirements.bzl", "pip_install")
pip_install()
```

BUILD:
```
exports_files(["requirements.txt"])

load("@requirements//:requirements.bzl", "requirement")

py_binary(
    name="run",
    srcs=["run.py"],
    deps=[
        requirement('mock'),
    ],
)
```

requirements.txt:
```
mock==2.0.0
```

run.py:
```
import mock
```

Then to reproduce run:

```
$ bazel run //:run
INFO: Found 1 target...
Target //:run up-to-date:
  bazel-bin/run
INFO: Elapsed time: 1.138s, Critical Path: 0.03s

INFO: Running command line: bazel-bin/run
Traceback (most recent call last):
  File "/private/var/tmp/_bazel_nik/5a08842075e4d3de340bcb5074db6110/execroot/__main__/bazel-out/darwin_x86_64-fastbuild/bin/run.runfiles/__main__/run.py", line 1, in <module>
    import mock
  File "/private/var/tmp/_bazel_nik/5a08842075e4d3de340bcb5074db6110/execroot/__main__/bazel-out/darwin_x86_64-fastbuild/bin/run.runfiles/pypi__mock_2_0_0/mock/__init__.py", line 2, in <module>
    import mock.mock as _mock
  File "/private/var/tmp/_bazel_nik/5a08842075e4d3de340bcb5074db6110/execroot/__main__/bazel-out/darwin_x86_64-fastbuild/bin/run.runfiles/pypi__mock_2_0_0/mock/mock.py", line 80, in <module>
    import funcsigs
ImportError: No module named funcsigs
ERROR: Non-zero return code '1' from command: Process exited with status 1.
```

[...]

PR #50, opened 2017-12-05
Evaluate PEP 508 environment markers for package dependencies

nikhaldi:
Previously any wheel dependencies that had an environment marker
(such as 'python_version>3.3') were simply ignored, leading to
missing packages in the Python environment constructed by bazel.

Fixes https://github.com/bazelbuild/rules_python/issues/49


bazel-io:

nikhaldi:
Two concerns that I still have with this change:

1. This definitely has the potential to break existing builds. For example through version conflicts between explicit packages in requirements.txt and the transitive dependencies now newly pulled in through bazel. Not sure what the rules are around breaking changes.
2. This relies on `pkg_resources.evaluate_marker()` which comes from setuptools. I assume it's only available from some setuptools version onwards (I can try to find out). I wonder if it's a good idea to put a lower boundary setuptools version in `python/requirements.txt`.

rules_python/whl_test.py
```diff
@@ -60,11 +63,19 @@ def test_mock_whl(self):
     self.assertEqual(wheel.name(), 'mock')
     self.assertEqual(wheel.distribution(), 'mock')
     self.assertEqual(wheel.version(), '2.0.0')
+    if VERSION_TUPLE < (3, 3):
+      expected_deps = ['funcsigs', 'pbr', 'six']
+    else:
+      expected_deps = ['pbr', 'six']
```
mattmoor: @duggelz Do you have any tricks to get coverage of both paths?

nikhaldi:
@mattmoor rebased

mattmoor:
Test this please 

(For Bazel CI)

rules_python/whl_test.py
nikhaldi: Good point. I think it can be done by mocking `platform.python_version()`. I could add this to this PR or send a separate change after.

nikhaldi:
About the CI failure (https://ci.bazel.io/blue/organizations/jenkins/PR%2Frules_python/detail/rules_python/67/pipeline): It seems the linux-x86_64 environment on Jenkins has a somewhat old setuptools version which does not yet support some common syntax in version markers. It's failing on '<='. This appears to be supported from setuptools 17.1 onwards.

I added another commit which pins the version of setuptools in piptool to the latest (38.2.4). I think this is a good idea in any case as it should make builds more predictable in general. It does mean that the piptool.par executable is now 1.8M instead of 1.3M as setuptools is embedded in it.

mattmoor:
Test this please

mattmoor:
@nikhaldi This is because we are actually using two different scripts to do this, `piptool.par` is one, which we bundle as a PAR to avoid needing the host to have `pip` installed, and `whl.py`, which we [invoke directly](https://github.com/bazelbuild/rules_python/blob/master/python/whl.bzl#L47).

### So why are these (checked in `.par` or `.py`) our only options?

During `BUILD` steps, you can use `py_binary` targets as "tools", but you cannot do this during `WORKSPACE`.  So whatever you invoke must be a standalone executable.

### So how do we get past this?

Given a dependency on a particular `setuptools`, I think the option that makes the most sense would be to do with `whl.py` what we do for `piptool.par` and have `update_tools.sh` check a PAR file into the repo (yuck, but oh well).

LMK if that [doesn't] make sense, and I'm happy to elaborate.

nikhaldi:
@mattmoor thanks for the explanation. It makes somewhat sense, though it feels like there is a lot of duct tape holding this thing together.

I can see how to do a separate `whl.par`. An alternative might be to vendor (copy) the `packaging` module. That's where the environment marker parsing is actually implemented. I only realized recently that `setuptools` is doing the same thing, it's just [vendoring](https://github.com/pypa/setuptools/blob/master/pkg_resources/_vendor/vendored.txt) `packaging`. The existing uses in piptool of `pkg_resources.Requirement` could probably be replaced with `packaging` as well which would remove `setuptools` from `piptool.par`. Thoughts?

I won't have time to look at this further until late December/early January but I can spend some more time on it then. I'm not unhappy either if somebody else takes over in the meantime.

mattmoor:
@nikhaldi You cannot vendor the `packaging` module because `whl.py` has to be a self-contained script to work properly.  This is what I meant when I said it couldn't be a `py_binary`.

FWIW, I agree with your characterization (re: duct tape), but I think it's largely working around some limitations of Bazel that aren't easily fixed (though I've been trying!).  For things that aren't, I'm certainly open to discussing the trade-offs and making improvements.

> The existing uses in piptool of pkg_resources.Requirement could probably be replaced with packaging as well which would remove setuptools from piptool.par. Thoughts?

Defer to @duggelz wisdom, but superficially this SGTM.

> I won't have time to look at this further until late December/early January but I can spend some more time on it then. I'm not unhappy either if somebody else takes over in the meantime.

No worries, everything slows down this time of year :)

I really appreciate you taking the time to send us this PR.

nikhaldi:
Test this please

mattmoor:
Test this please

nikhaldi:
This appears to be working now with a separate whltool.par. Also refactored the tests to not use ugly branching based on Python versions.

I opted to leave the `setuptools` package in over `packaging` because piptool uses some `pkg_resources` features which I'm not sure can easily be rewritten with `packaging`. This could be a separate refactoring if the size of the par files is a real concern.

mattmoor:
Test this please

PR #61, opened 2018-01-25
Multiplatform

dududko:
In this patch is proposed the way to resolve the issue of building platform independent python binaries and images with BAZEL. This will allow to build a docker image for any platform using python_rules and still be able to run python binary using bazel native rule.

> I would like to mention that it was difficult to patch your code, so I created piptool_v2.py and whl_v2.py modules as well as additional rules for pip.bzl.

There are two main steps I did for solution implementation:

1. Download and extract wheel files for specified platforms. I implemented this using “pip download --platform <platform>” instead of “pip wheel”.
2. Extract wheel files for the same package in a single directory so that all platform dependens .so files could be found in a package directory.

To solve the problem with packages that can’t be downloaded with “pip download” I create them manually and connect as external repository. Then I pass the path to this external repository to “pip download” using [--find links](https://pip.pypa.io/en/stable/reference/pip_download/#cmdoption-f) command. Pip will automatically resolve all existing wheel files and copy them to requirements repo. This solution also allowed me to cache the files which are already downloaded with pip.

Here you can find the example where I use my patch.
https://github.com/dududko/example_rules_python


bazel-io:

dududko:
I signed it!

evanj:
(I am not a Bazel contributor, just a user) Interesting approach, thanks! I'm going to have to take a closer look at this, to see if I can borrow parts of your approach. I've done something similar, with storing built wheels an external location. BTW, there is some discussion about trying to get some collaboration going on these Python rules. You may wish to see the following bazel-discuss thread: https://groups.google.com/d/msg/bazel-discuss/NMO6KPyPKh4/VnxtEVP6EQAJ

dududko:
@evanj , thank you for the link!

excavador:
@evanj by the way, please take a look here: https://groups.google.com/forum/#!topic/bazel-discuss/oXz0o6B9tAI

excavador:
@evanj ah, it was you! :)))

Issue #67, opened 2018-02-21
Tests failing on Bazel CI


katre:
I am seeing failures in Bazel CI for both Bazel HEAD and Bazel 0.10.1: https://ci.bazel.build/blue/organizations/jenkins/Global%2Frules_python/detail/rules_python/276/tests

I can reproduce this locally and get the same error messages.

katre:
This seems to be due to 3175797bd07aac4ff35fa711f0a82285f2005e42, @mattmoor can you comment?

mattmoor:
@duggelz made PAR construction deterministic, so without probing deeper I would guess that you changed something (transitively) about how the PAR is built (the same `__init__.py` change that broke `rules_docker` seems plausible?).

buchgr:
How can we resolve this issue? Who would be the person to talk to?

duggelz:
Bazel has apparently just switched to a new CI system:

https://groups.google.com/d/msg/bazel-discuss/8spcVZMv9yE/nz-ya0tpAQAJ

buchgr:
Heya,

we are still seeing this error on our CI. `//tools:par_test` is failing with Bazel from `HEAD` but it works with the latest released Bazel.

```
exec ${PAGER:-/usr/bin/less} "$0" || exit 1
-----------------------------------------------------------------------------
FF
======================================================================
FAIL: test_piptool_matches (__main__.WheelTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/3dd5dfaab0d9fd8c7dd47297b1d500e7/sandbox/5266514015113038112/execroot/io_bazel_rules_python/bazel-out/k8-fastbuild/bin/tools/par_test.runfiles/io_bazel_rules_python/tools/par_test.py", line 33, in test_piptool_matches
    'The checked in tools/piptool.par does not match the latest build.')
AssertionError: The checked in tools/piptool.par does not match the latest build.

======================================================================
FAIL: test_whltool_matches (__main__.WheelTest)
----------------------------------------------------------------------
Traceback (most recent call last):
[...]

PR #81, opened 2018-03-23
Switch from using .par files to pure python for piptool and whltool

duggelz:
Addresses #67 

This vendors several third_party packages, which increases the repository size by 12MB and 116000 lines.  Deleting the .par files will get about 6MB of that back.

The .par files will be removed in a separate change.

This change was tested against the rules_k8s repository, and various manual testing.

Alternatives to this change are:
1) Do nothing, continue with the current process
2) Create an actual release process for the .par files, which builds and stores them in a separate repository along with the .bzl files.


duggelz:
@mattmoor for your consideration

duggelz:
I suggest using GitHub's "Show changes from commit range..." feature, and skip the ginormous third_party commits.

c4urself:
I'd love to see this get in, debugging issues with the PAR files is not fun!!

sitaktif:
Can we do this instead? https://github.com/soniaai/rules_poetry/blob/master/rules_poetry/defs.bzl#L3-L61

thundergolfer:
Closing as release 0.1.0 included a packaging rules implementation that no longer relied on `.par` files.

PR #86, opened 2018-04-06
Update dist_info location generation to account for build tags

rydrman:
The [pep-427](https://www.python.org/dev/peps/pep-0427/#file-name-convention) spec that is being followed when discovering dist_info data in python wheels allows for an optional build tag to exist. 

Previously this was not taken into account and could make the dist_info appear missing when a build tag was used.
 
These changes update the way the dist_info path is generated to allow for the case where a build tag is included.


thundergolfer:
Will close due lack of activity.

Issue #87, opened 2018-04-06
Rolling TLS brownouts breaking pip integration


jkinkead:
I'm seeing mysterious failures that I'm pretty sure are [this issue](https://github.com/pypa/packaging-problems/issues/130): pypi deprecating old TLS support.

The fix is to update to pip 9.0.3. I have a patch for this coming in a PR in a moment.

jkinkead:
@mattmoor / @duggelz - Per one of the comments in the linked issue, this is going to break the pip rules completely starting April 8th (Monday). This should be fixed ASAP.

duggelz:
Pull request merged, thank you.

jkinkead:
Thanks for fixing this quickly!

PR #88, opened 2018-04-06
Update pip version to 9.0.3.

jkinkead:
Partial fix for #87 . This still needs `./update_tools.sh` to be run; but I figured the binary files should be added by a trusted source.


duggelz:
I still haven't figured out the right way to add my own commits to someone else's PR, so I'm going to just force merge this and then make a new PR to regenerate the .par files, then force merge that.  I don't see any possible way that could go wrong :)

jkinkead:
lol.

FWIW fetching the wheels to the shared global pip cache with a newer version of pip works pretty well. ;)

Issue #70, opened 2018-02-27
rules_python doesn't read dependencies from METADATA files


drigz:
This means that rules that depend (directly or indirectly) on wheels that don't contain `metadata.json` (eg [celery==3.1.11](https://pypi.python.org/pypi/celery/3.1.11)), will fail as described in [this message to bazel-discuss](https://groups.google.com/d/msg/bazel-discuss/Nt4F-_4vmJc/hOd0_EdEAQAJ).

The cause appears to be [this TODO](https://github.com/bazelbuild/rules_python/blob/115e3a0dab4291184fdcb0d4e564a0328364571a/rules_python/whl.py#L110):
```
# TODO: handle fields other than just name
```

This can be worked around by manually declaring the dependencies in requirements.txt and your own rules.

drigz:
This is likely to become more pressing since wheel v0.31.0 stopped generating metadata.json. google-auth versions after 1.4.1 don't have metadata.json, so you either have to pin to 1.4.1 or manually declare the dependencies.

https://github.com/pypa/wheel/blob/3dc261abc98a5e43bc7fcf5783d080aaf8f9f0cf/CHANGES.txt#L13

evanj:
My pull request #90 is an attempt to resolve this. I haven't tested it extremely carefully, but so far this patch has worked for the packages we use.

brandjon:
FTR this issue is the cause of #98.

brandjon:
Prioritizing because this can easily cause annoying breakages as shown by #98. Will look at this after other CI failures / blockers.

thundergolfer:
No longer a problem with `pip_install` repo rule. Closing issue.

PR #90, opened 2018-04-09
whl: Parse requirements from METADATA files

evanj:
Some wheels only include METADATA files (examples: wheel 0.31.0,
tenacity 4.10.0). Previously whl did not parse the requirement
sections. Add code to parse these sections.

The most complicated part is parsing out the extra == 'something'
clauses from the boolean expression. Use the same expression parser
used by pkg_resources.evaluate_marker which is already used by this
code to find those clauses and produce a version with them removed.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -85,6 +85,15 @@ http_file(
 )
 
 http_file(
+    name = "wheel_0_31_1_whl",
+    sha256 = "80044e51ec5bbf6c894ba0bc48d26a8c20a9ba629f4ca19ea26ecfcf87685f5f",
+    # From https://pypi.org/project/wheel/
+    url = ("https://files.pythonhosted.org/packages/81/30/" +
+           "e935244ca6165187ae8be876b6316ae201b71485538ffac1d718843025a9/" +
+           "wheel-0.31.1-py2.py3-none-any.whl"),
+)
+
+http_file(
     name = "mock_whl",
     sha256 = "5ce3c71c5545b472da17b72268978914d0252980348636840bd34a00b5cc96c1",
     # From https://pypi.python.org/pypi/mock
--- a/rules_python/BUILD
+++ b/rules_python/BUILD
@@ -35,6 +35,7 @@ py_test(
         "@google_cloud_language_whl//file",
         "@grpc_whl//file",
         "@mock_whl//file",
+        "@wheel_0_31_1_whl//file",
     ],
     deps = [
         ":whl",
--- a/rules_python/whl.py
+++ b/rules_python/whl.py
@@ -16,9 +16,81 @@
 import argparse
 import json
 import os
-import pkg_resources
 import re
+import rfc822
+import sys
 import zipfile
+
+from pkg_resources._vendor.packaging import markers
+import pkg_resources
+
+def recurse_split_extra(parsed_parts):
+  extra = ''
+  remaining = []
+
+  i = 0
+  for part in parsed_parts:
+    if isinstance(part, list):
+      # parenthesized expressions are lists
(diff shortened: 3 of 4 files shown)


evanj:
I ran into this with our `rules_pyz` rules, but since we basically use a version of this tool ourselves, I decided I might as well submit this improvement upstream: https://github.com/TriggerMail/rules_pyz

jdelfino:
Hi @evanj - thanks for working on this PR, I was happy to find it when I hit this issue. Unfortunately it doesn't seem to catch all cases, and I haven't dug into why yet. See https://github.com/bazelbuild/rules_python/issues/95.

ensonic:
Got the same problem. Could you please rebase the change, so that I can verify it?

evanj:
I just rebased this change. I also updated the version of wheel that is used by the test to the latest version.

ensonic:
I found one issue that does not work. Look at the last line of the generated BUILD file:
```python
# ... /external/pypi__gapic_google_cloud_logging_v2_0_91_3/BUILD

load("@ros_log_deps//:requirements.bzl", "requirement")

py_library(
  name = "pkg",
  srcs = glob(["**/*.py"]),
  data = glob(["**/*"], exclude=["**/*.py", "**/* *", "BUILD", "WORKSPACE"]),
  imports = ["."],
  deps = [requirement("google-gax"),requirement("googleapis-common-protos[grpc]"),requirement("oauth2client"),requirement("proto-google-cloud-logging-v2[grpc]"
)],
     )
```
this results into:
```
key "googleapis_common_protos[grpc]" not found in dictionary
```

rogerhub:
Also, do `tools/piptool.par` and `tools/whltool.par` need to be updated to include these changes?

bbarnes52:
any ETA when this will be resolved?

lberki:
Soon! Jon will start working on this around middle August. Sorry for the continuing delays.

c4urself:
whl.py should not have any Python2-only syntax (such as `basestring`) -- the problem is that when you parse requirements with `pkg_resources.evaluate_marker` it will use the current script's Python version.  If whltool.par is Python2 then we will always only get Python2 dependencies. This probably works in a Python2 world but does not work at all Python3.

c4urself:
I think the right way to do is is to merge https://github.com/bazelbuild/rules_python/pull/82 first and then use `six` to make `whl.py` Python2/3 compatible.

gfl-chris:
I merged this PR into my fork and I do see BUILD files in the `./bazel-<proj>/external/*` directories, but the transitive dependencies don't seem to be automatically picked up. Any idea what I need to add to my BUILD files?

gfl-chris:
I take that back. This does generate dependencies for some packages, but not for others. For example, the BUILD file for Flask-1.0.2 looks like this:

asford:
@brandjon Any chance you can review this in connection to your triage of #70?

greggdonovan:
@evanj @brandjon What's the latest on this PR? We're interested in this, as we have had to pause any Python dependency updates for effected dependencies. E.g. [`kubernetes-client`|https://github.com/kubernetes-client/python].

evanj:
I couldn't convince my organization to use Bazel, so I no longer have the time to figure out how to get Python to work with it, sorry.

pcn:
https://github.com/bazelbuild/rules_python/pull/121 works as-is with python2, but doesn't work on its latest commits with python3.

greggdonovan:
@brandjon Do you have any advice for how to move forward? If this PR is rebased is it mergeable?

hrfuller:
Wondering the same. What is needed to get this merged?

groodt:
This PR seems closable now? It's very old and the new third-party packaging seems to handle these situations correctly now.

PR #101, opened 2018-05-23
Allow environment variables and extra args to be passed to pip.

### LINKED ISSUE #74: Support for pip --index-url, --extra-index-url, --find-links
Atm it doesn't seem possible to fetch wheels from anywhere except the default PyPI index. We have some custom built wheels that we store in a directory and commit with the repository, and want to use those with `--find-links`, not something from PyPI.

Globegitter:
This should fix #74 but was really motivated by the fact that I wanted to install https://github.com/uploadcare/pillow-simd with the `cc -mavx2` option reliably. So I needed the option to pass in environment variables that get propagated to pip and I further ran into this issue: https://github.com/uploadcare/pillow-simd/pull/26 so also needed to be able to pass in custom arguments.

So an example would be:

```
pip_import(
   name = "py_deps",
   requirements = "//:requirements.txt",
   environment = {
       "CC": "cc -maxv2",
   },
   extra_args = ["--no-binary", "pillow-simd"],
)
```

### DIFF
--- a/rules_python/piptool.py
+++ b/rules_python/piptool.py
@@ -151,10 +151,10 @@ def determine_possible_extras(whls):
   }
 
 def main():
-  args = parser.parse_args()
+  args, unkwown = parser.parse_known_args()
 
   # https://github.com/pypa/pip/blob/9.0.1/pip/__init__.py#L209
-  if pip_main(["wheel", "-w", args.directory, "-r", args.input]):
+  if pip_main(["wheel", "-w", args.directory,] + unkwown + ["-r", args.input]):
     sys.exit(1)
 
   # Enumerate the .whl files we downloaded.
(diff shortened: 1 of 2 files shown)


rules_python/piptool.py
```diff
@@ -151,10 +151,10 @@ def is_possible(distro, extra):
   }
 
 def main():
-  args = parser.parse_args()
+  args, unkwown = parser.parse_known_args()
```
stefanhdao: Nit: `unknown`

dsully:
Any progress here? Would love to see this committed to the repo. Thanks.

georgegiovanni:
bump

ajgokhale:
bump

xhuoroku:
Any update?

thundergolfer:
@xhuoroku look `pip_install` which has this functionality. `pip_import` is deprecated.

PR #108, opened 2018-07-16
Parse requirements from METADATA

siddharthab:
This is needed for some packages like python_dateutil which have not yet started using metadata.json.


rogerhub:
There's already a PR for this feature (#90).

At a glance, that one seems to handle more complex ways that METADATA can specify the dependencies. I didn't look too closely though.

siddharthab:
Aah! I did not see that. I am going to close this now.

PR #109, opened 2018-07-16
Add prebuilt option

marczych:
This adds an option to the `pip_import()` rule that allows users to
specify a prebuilt tarball which skips the slow `pip wheel ...` command
that happens frequently during bazel commands. Details:

 - `prebuilt`: Tar archive that contains `*.whl` and `requirements.txt`
   generated from running `pip wheel -r requirements.txt`
 - `generate_prebuilt_and_exit`: Development option that allows users to
   generate the prebuilt archive in order to get it into their
   workspace.

The update process looks like this:

 - Developer updates their `requirements.txt`, runs a `bazel command`
   which will fail because the `requirements.txt` in the prebuilt
   doesn't match the one in the repository
 - Add `generate_prebuilt_and_exit = True` to the `pip_import()` command
   and run the build again
 - The prebuilt will be generated and the command will fail
 - The developer takes the prebuilt, uploads it for deployment, updates
   the reference in the `WORKSPACE`, removes
   `generate_prebuilt_and_exit`, and is off to the races

### DIFF
--- a/python/pip.bzl
+++ b/python/pip.bzl
@@ -23,13 +23,24 @@ def _pip_import_impl(repository_ctx):
   repository_ctx.file("BUILD", "")
 
   # To see the output, pass: quiet=False
-  result = repository_ctx.execute([
+  args = [
     "python", repository_ctx.path(repository_ctx.attr._script),
     "--name", repository_ctx.attr.name,
     "--input", repository_ctx.path(repository_ctx.attr.requirements),
     "--output", repository_ctx.path("requirements.bzl"),
     "--directory", repository_ctx.path(""),
-  ])
+  ]
+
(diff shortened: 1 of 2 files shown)


marczych:
I signed it!

marczych:
I admit that the API here is a little clunky (specifically, the steps to generate a new tarball) so I'm open to suggestions on how to improve that. It has drastically reduced the amount of time we spend downloading dependencies so I think that it's very useful even in its current state. Also, the documentation needs to be updated before merging but I'm holding off on that until the rest of the feature is reviewed.

marczych:
For what it's worth, I'm using a slightly different approach for this these days. Is it fair to say that `pip_parse` serves a similar purpose?

PR #121, opened 2018-09-19
add support for METADATA 2.1 in wheels

ali5h:
`wheel` has dropped support for `metadata.json`. This change adds the
functionality to parse METADATA 2.1 file and return the same
dictionary as deprecated `metadata.json`.


rules_python/whl.py
```diff
@@ -107,9 +112,43 @@ def expand(self, directory):
 
   # _parse_metadata parses METADATA files according to https://www.python.org/dev/peps/pep-0314/
   def _parse_metadata(self, content):
-    # TODO: handle fields other than just name
-    name_pattern = re.compile('Name: (.*)')
-    return { 'name': name_pattern.search(content).group(1) }
+    metadata = {}
+    pkg_info = email.parser.Parser().parse(content)
```
clintharrison: This is not compatible with both Python 2 and 3. Running this with 2 will fail when there are unicode characters in the METADATA file (Grab python_dateutil-2.7.5-py2.py3-none-any.whl from PyPI for an example), as the email parser package uses cStringIO, which is only compatible with ASCII strings.

shariat: StringIO is compatible with unicode, but not mixed unicode and ASCII. I tested the change with the wheel you mentioned and it works

shariat: the output is 
```

load("test", "requirement")

py_library(
    name = "pkg",
    srcs = glob(["**/*.py"]),
    data = glob(["**/*"], exclude=["**/*.py", "**/* *", "BUILD", "WORKSPACE"]),
    imports = ["."],
    deps = [requirement("six")],
)
```

clintharrison: Err, you're right. Sorry, my bad. I got mixed up in my Python versions -- the issue is actually with Python 3, as the `email.feedparser.BufferedSubFile` used under the hood by `email.parser.Parser` uses `StringIO`. `zipfile.ZipFile.open()` always reads as bytes, which causes an error like the following:

```
  File "/Users/clint/code/rules_python/rules_python/whl.py", line 74, in metadata
    return self._parse_metadata(f)
  File "/Users/clint/code/rules_python/rules_python/whl.py", line 116, in _parse_metadata
    pkg_info = email.parser.Parser().parse(content)
  File "/usr/local/Cellar/python/3.7.1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/email/parser.py", line 57, in parse
    feedparser.feed(data)
  File "/usr/local/Cellar/python/3.7.1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/email/feedparser.py", line 175, in feed
    self._input.push(data)
  File "/usr/local/Cellar/python/3.7.1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/email/feedparser.py", line 103, in push
    self._partial.write(data)
TypeError: string argument expected, got 'bytes'
```

ali5h:
@brandjon this is basically #90. I did it without knowing that #90 exists. Although, this one has much less code changes

pcn:
`rfc822` doesn't exist in python3, I'm not sure the move to using it is going to work long-term.

ali5h:
@pcn I have one with `email` package, will push the changes

pcn:
That'd be cool, thanks.

cristifalcas:
Hi. Can we have some updates on this one please?

mdvorsky:
Here's an alternative implementation, with some tests as well: https://github.com/bazelbuild/rules_python/pull/198

ali5h:
I wrote these fetch rules for pip using pip-compile https://github.com/ali5h/rules_pip_lock

PR #128, opened 2018-11-02
Add experimental support for building wheels.

pstradomski:
Currently only pure python wheels are supported.

This extension builds wheels directly, invoking a simple python script that creates the zip archive in the desired format instead of using distutils/setuptools.

This will make building platform-dependent wheels easier in the future, as bazel
will have full control on how extension code is built.


experimental/python/wheel.bzl
```diff
@@ -0,0 +1,142 @@
...
+py_wheel = rule(
+    implementation=_py_wheel_impl,
+    attrs={
+        "deps": attr.label_list(
+            doc="Dependencies of this package, e.g. py_library rules"),
+        "packages": attr.string_list(
+            mandatory=True,
+            allow_empty=False,
+            doc="""\
+List of Python packages to include in the distribution.
```
pstradomski: This is controversial and perhaps worth discussing.

A typical package will have dependencies on other packages; some might be first party but packaged separately, some might be third party. From bazel point of view they will probably form a tree of py_library dependencies.

When creating a wheel package one does not want to include the whole tree, so we need to tell bazel where to cut the chain. I see three ways to do that:
1) Don't follow dependencies. Only package up the libraries listed directly in deps.
2) Follow dependencies and only package specific python packages
  a) Including sub-packages, so e.g. if "foo" is listed then "foo.bar" gets packaged too
  b) Not including sub-packages.

The implementation here does "2a", but I could see how other options could be preferred. I'm happy to change if you think other options are better.

pstradomski:
@brandjon: any chances somebody with commit rights takes a look? Thanks!

dsully:
Would love to see this make it into the repo. Any progress here?

Thanks

pstradomski:
Still waiting for the maintainers to respond.

Did you try out this patch? Did it work for you? Any issues? I'm currently
aware of encoding issues if one of the metadata fields (eg. author)
contains non-ascii letters that I still need to address, though I think
there's value in the patch even as it is now.

pon., 14 sty 2019, 16:40: Dan Sully <notifications@github.com> napisał(a):

pstradomski:
Btw, I've fixed the encoding issues. Building wheels with non-ascii characters in metadata should now work in both Python 2 and Python 3.

pstradomski:
> Would love to see this make it into the repo. Any progress here?
> 
> Thanks

pstradomski:
Going to try with a new pull request to avoid all the merge mess here.

Issue #98, opened 2018-05-15
boto_test broken on mac


lfpino:
Hi there,

I'm the Bazel release manager and I've been investigating the rules_python failures (https://buildkite.com/bazel/bazel-with-downstream-projects-bazel/builds/233#14f47bce-4863-42cf-9179-fe4b1be650a8) in the new release candidate for Bazel 0.14.0. After trying to reproduce this in my macOS I realized that the failures are also present with Bazel 0.12.0 as follows:

```
$ git clone https://github.com/bazelbuild/rules_python.git
$ cd rules_python
$ /tmp/download/bazel-0.12.0-darwin-x86_64 test //examples/boto:boto_test
INFO: Analysed target //examples/boto:boto_test (23 packages loaded).
INFO: Found 1 test target...
FAIL: //examples/boto:boto_test (see /private/var/tmp/_bazel_lpino/4cea827263cf09da0a5ddbf48df3f327/execroot/io_bazel_rules_python/bazel-out/darwin-fastbuild/testlogs/examples/boto/boto_test/test.log)
Target //examples/boto:boto_test up-to-date:
  bazel-bin/examples/boto/boto_test
INFO: Elapsed time: 3.315s, Critical Path: 1.16s
INFO: Build completed, 1 test FAILED, 5 total actions
//examples/boto:boto_test                                                FAILED in 1.0s
  /private/var/tmp/_bazel_lpino/4cea827263cf09da0a5ddbf48df3f327/execroot/io_bazel_rules_python/bazel-out/darwin-fastbuild/testlogs/examples/boto/boto_test/test.log

$ cat /private/var/tmp/_bazel_lpino/4cea827263cf09da0a5ddbf48df3f327/execroot/io_bazel_rules_python/bazel-out/darwin-fastbuild/testlogs/examples/boto/boto_test/test.log
exec ${PAGER:-/usr/bin/less} "$0" || exit 1
-----------------------------------------------------------------------------
Traceback (most recent call last):
  File "/private/var/tmp/_bazel_lpino/4cea827263cf09da0a5ddbf48df3f327/bazel-sandbox/5352791971010170471/execroot/io_bazel_rules_python/bazel-out/darwin-fastbuild/bin/examples/boto/boto_test.runfiles/io_bazel_rules_python/examples/boto/boto_test.py", line 15, in <module>
    import boto3
[...]

Issue #117, opened 2018-09-05
//tools:par_test red on bazel ci again


mhlopko:
Hello :)

I see the par_test is red again: https://buildkite.com/bazel/bazel-with-downstream-projects-bazel/builds/425#47df845e-6759-4c77-8867-7e81e02ddac0

We had problems with it in https://github.com/bazelbuild/rules_python/issues/67 and in https://github.com/bazelbuild/rules_python/issues/80. Can you help me understand why the test is failing with bazel@HEAD? What can we do to prevent it in the future? Should we run the test on our CI with bazel@HEAD?

Thanks!

brandjon:
To recap my understanding of [Douglas's comment](https://github.com/bazelbuild/rules_python/issues/67#issuecomment-377085251): This test is intended to provide a security guarantee that the checked-in pre-built .par files have not been tampered with. The problem is that the check for exact equality is too brittle, and a looser check would compromise confidence in the test.

The pre-built .par files are needed for bootstrapping, since piptool and whltool are themselves defined in terms of pip_import. I would argue that since they're used for bootstrapping, this defeats the point of the test anyway, since a malicious piptool.par would taint its own rebuild.

I agree with Douglas's comment that we should just delete the test. That avoids giving people the false impression that we have an automated way to verify the integrity of the .par files. We definitely should not accept PRs modifying the .par files from untrusted sources.

Long term, to avoid a checked-in binary, we could break the circular dependency of piptool/whltool on themselves by using a non-pip_import way of obtaining the dependencies in [rules_python/python/requirements.txt](https://github.com/bazelbuild/rules_python/blob/7e5adb05b7aa4cba64a6dd1138f817670888b220/python/requirements.txt).

PR #131, opened 2018-11-12
Fix boto_test

brandjon:
WIP toward fixing #98. AFAIK creating a PR is necessary so I can trigger our CI to run.


brandjon:
Rebased to the real fix (I hope).

brandjon:
Confirmed, boto_test is no longer failing on mac. The only remaining failure is pre-existing and unrelated (#117).

PR #135, opened 2018-11-15
Switch to new http repo rules

### LINKED ISSUE #105: Please move from native to Skylark rules for external git/http repositories
Bazel is planning to move away from native rules for external repositories to Skylark rules, mainly because bundling every version control system in bazel does not scale. As part of this effort, we plan to submit https://bazel-review.googlesource.com/c/bazel/+/55932 soon, which will cause bazel by default to insist that the skylark versions the git and http rules for external repositories be used. This can be disabled by a flag for at least the next 6 months, but we still would appreciate if you could adopt your repository to use the Skylark versions of those rules soon.

A test run of downstream projects with that flag flip in bazel enabled can be found at https://buildkite.com/bazel/bazel-with-downstream-projects-bazel/builds/286

For more background, see
- the original announcement https://groups.google.com/d/topic/bazel-discuss/dO2MHQLwJF0/discussion
- and the documentation of the incompatible changes https://docs.bazel.build/versions/master/skylark/backward-compatibility.html#remove-native-git-repository

Thanks.

brandjon:
This upgrades our dependencies to newer versions that are compatible with the new repo rules, and handles the fallout. Also add some comments to WORKSPACE.

rules_python now builds under Bazel 0.19.1 with `--incompatible_remove_native_git_repository` and `incompatible_remove_native_http_archive`.

Fixes #105.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -11,22 +11,47 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
 workspace(name = "io_bazel_rules_python")
 
-# Skydoc stuff
+load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_file")
+
+################################
+# Skydoc and its dependencies. #
+################################
+
+# Skydoc's sass dependency is not covered by skydoc_repositories(), so we have
+# to redeclare it here. Watch that the version matches when updating Skydoc's
+# version.
+
 git_repository(
     name = "io_bazel_rules_sass",
+    commit = "8b61ad6953fde55031658e1731c335220f881369",  # Taken from skydoc's WORKSPACE
     remote = "https://github.com/bazelbuild/rules_sass.git",
-    tag = "0.0.3",
 )
 
-load("@io_bazel_rules_sass//sass:sass.bzl", "sass_repositories")
+load("@io_bazel_rules_sass//:package.bzl", "rules_sass_dependencies")
+
+rules_sass_dependencies()
+
+# Node is used by sass. This weird (anti-?)pattern of initializing a repo we
+# didn't directly import is taken from Skydoc's WORKSPACE.
+load("@build_bazel_rules_nodejs//:defs.bzl", "node_repositories")
+
+node_repositories()
+
+load("@io_bazel_rules_sass//:defs.bzl", "sass_repositories")
 
 sass_repositories()
 
+# We implicitly depend on Skydoc importing Skylib under `@bazel_skylib`.
+# We don't redeclare it here in order to avoid repeating a definition that
+# could get out of sync with Skydoc.
+
 git_repository(
     name = "io_bazel_skydoc",
-    commit = "e9be81cf5be41e4200749f5d8aa2db7955f8aacc",
+    commit = "77e5399258f6d91417d23634fce97d73b40cf337",  # 2018-10-29
     remote = "https://github.com/bazelbuild/skydoc.git",
 )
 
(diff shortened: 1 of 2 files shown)


brandjon:
Rupert, sending this to you "by default" since you did a couple trivial ones earlier (thanks!), but feel free to send it back for me to find someone else.

WORKSPACE
```diff
@@ -51,58 +79,69 @@ _piptool_install()
...
+
 http_file(
     name = "grpc_whl",
+    downloaded_file_path = "grpcio-1.6.0-cp27-cp27m-manylinux1_i686.whl",
     sha256 = "c232d6d168cb582e5eba8e1c0da8d64b54b041dd5ea194895a2fe76050916561",
     # From https://pypi.python.org/pypi/grpcio/1.6.0
-    url = ("https://pypi.python.org/packages/c6/28/" +
-           "67651b4eabe616b27472c5518f9b2aa3f63beab8f62100b26f05ac428639/" +
-           "grpcio-1.6.0-cp27-cp27m-manylinux1_i686.whl"),
+    urls = [("https://pypi.python.org/packages/c6/28/" +
```
katre: Why is this constructed using string concatenation? I'd prefer just one long string, because that would make it clearer that there's only a single URL in the urls attribute.

WORKSPACE
```diff
@@ -51,58 +79,69 @@ _piptool_install()
...
     name = "google_cloud_language_whl",
+    downloaded_file_path = "google_cloud_language-0.29.0-py2.py3-none-any.whl",
     sha256 = "a2dd34f0a0ebf5705dcbe34bd41199b1d0a55c4597d38ed045bd183361a561e9",
     # From https://pypi.python.org/pypi/google-cloud-language
-    url = ("https://pypi.python.org/packages/6e/86/" +
-           "cae57e4802e72d9e626ee5828ed5a646cf4016b473a4a022f1038dba3460/" +
-           "google_cloud_language-0.29.0-py2.py3-none-any.whl"),
+    urls = [("https://pypi.python.org/packages/6e/86/" +
+             "cae57e4802e72d9e626ee5828ed5a646cf4016b473a4a022f1038dba3460/" +
+             "google_cloud_language-0.29.0-py2.py3-none-any.whl")],
```
katre: You need a hygenic macro to create these.

WORKSPACE
```diff
@@ -11,30 +11,58 @@
...
+# Skydoc and its dependencies. #
+################################
+
+# Skydoc's sass dependency is not covered by skydoc_repositories(), so we have
+# to redeclare it here. Watch that the version matches when updating Skydoc's
+# version.
+
 git_repository(
     name = "io_bazel_rules_sass",
+    commit = "8b61ad6953fde55031658e1731c335220f881369",  # Taken from skydoc's WORKSPACE
```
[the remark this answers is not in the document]
brandjon: Well skydoc didn't explain why they use that version. Updated the comment to add a date.

WORKSPACE
brandjon: The multiline form has the advantage of being able to easily read the distinct pieces of the string, for instance that the wheel part of the URL matches the file path attr above. I'm not sure the value of having an unbroken URL is worth it in this case.

WORKSPACE
brandjon: You mean to factor out the whl name from the url and the `downloaded_file_path`? I'm not sure what value the macro adds besides code golfing -- What would the macro's arguments be, and what would the URL-minus-the-wheel-name be? Or would you suggest parsing the URL to determine `downloaded_file_path`?

I'm not even sure I like the downloaded file path being the wheel name as opposed to the default string, `<repo name>/file/downloaded`, but I kept it this way for compatibility with the existing tests. If I create a macro, that risks users loading it and adopting that convention without us having really thought it out.

katre: Okay, that makes sense. Leave it as-is.

WORKSPACE
katre: SGTM.

WORKSPACE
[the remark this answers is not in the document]
katre: Thanks!

brandjon:
Thanks for the review!

PR #136, opened 2018-11-23
Update requirements.

uri-canva:
Updates pip, setuptools and wheel. Based on #82 because there's no way to merge the par files, but logically independent of it.

Diff between the two PRs: https://github.com/uri-canva/rules_python/compare/f96a5dbf2ebc035f7fbda51a3da068d7933f14ea..be39b7906da9c924b09ca78ce3fa152a10f782ae.


PoncinMatthieu:
It would be really great to get this in, the pip tools are really outdated and missing a few features. Also the new versions allows to get deterministic builds for the remote cache to be working.

c4urself:
This is a great -- but please consider: https://github.com/bazelbuild/rules_python/pull/81 -- no more having distribute a par that could potentially contain something malicious and ends up being impossible to debug properly.

PR #152, opened 2019-01-05
Add loading of git_repository rule to README

danieljanes:
Newer versions of Bazel do not contain a native `git_repository` rule so users of `rules_python` are required to load the `git_repository` rule before they can use it to fetch `rules_python`. This PR updates the documentation accordingly.


danieljanes:
I signed it!

PR #85, opened 2018-04-06
Add a pip3_import target for Python 3 compatibility.

jkinkead:
Fix / workaround for #33 .

Much thanks to @joshclimacell for doing essentially this.

PR #158, opened 2019-01-14
Add optional attr python_interpreter to run pip with python 3

acumon:
While it's possible to select which interpreter to run python rules with, pip will always be run with python from the system path.

This is an alternative solution to PR #82 and PR #85. it doesn't add pip3_import(). instead it adds python_interpreter attribute to pip_import(). Rationale is that if people want to use python3 for pip without changing the system's default ```python```, they are likely also want to use ```python3.6``` or ```python3.7``` without changing the system's default ```python3``` (for me it points to python 3.5) Using this approach users can any version of python, including customized built python with the interpreter path specified.


acumon:
I signed it!

acumon:
I signed it!

acumon:
I signed it!

acumon:
I signed it!

beneisner:
Has there been any progress on this? Very excited to see this change, will help our workflow out a lot.

akashgangil:
Can someone please look into this?

pickledgator:
+1 for this!

khurshidatcadre:
+1

README.md
```diff
@@ -65,8 +65,16 @@ load("@io_bazel_rules_python//python:pip.bzl", "pip_import")
 # This rule translates the specified requirements.txt into
 # @my_deps//:requirements.bzl, which itself exposes a pip_install method.
 pip_import(
-   name = "my_deps",
-   requirements = "//path/to:requirements.txt",
+  name = "my_deps",
+  requirements = "//path/to:requirements.txt",
+)
+
+# Python interpreter can be specify optionally to support customized version,
```
ValentinMouret:
```suggestion
# Python interpreter can be specified optionally to support customized version,
```

rules_python/piptool.py
```diff
@@ -84,6 +84,9 @@ def pip_main(argv):
 parser = argparse.ArgumentParser(
     description='Import Python dependencies into Bazel.')
 
+parser.add_argument('--python_interpreter', action='store',
+                    help=('The python python interpreter to use '))
```
ValentinMouret:
```suggestion
                    help=('The python interpreter to use '))
```

llchan:
Should this relate to `--python_top` in some way?  It seems that if I'm already specifying a `py_runtime(...)`, the pip steps should use the same python, no?

llchan:
Nevermind, seems like that is on its way out.  The python toolchain stuff is very actively changing (see e.g. https://github.com/bazelbuild/bazel/issues/7899), so it might be good to let the dust settle there before we make a move here.

bshashank:
Bump.
> The python toolchain stuff is very actively changing (see e.g. bazelbuild/bazel#7899), so it might be good to let the dust settle there before we make a move here.

I'm not sure it'll help. The rules here are used at the analysis phase - in the WORKSPACE file, and these rules only have access to the [repository_ctx](https://docs.bazel.build/versions/0.28.0/skylark/lib/repository_ctx.html), and this object does not have access to toolchains.

So we the workaround of specifying the interpreter in every `pip_import` invocation seems like a good workaround.

cristifalcas:
Hi. We also really need this. Any love for this PR?

Ragora:
I had to create a custom version of this repo just for this capability for work.

brandjon:
Thanks for the PR Chen. I've prioritized this feature as #249 and rebased your commit in #252.

As @bshashank pointed out, this is pretty independent of Python toolchains because this happens at `WORKSPACE` evaluation time. I imagine most users will only ever need to set it to `"python"`, `"python2"`, or `"python3"`, so it should be sufficient to add `pip2_import` / `pip3_import` wrappers around this feature. But in a pinch, if you had some weird system with a non-standard Python command, you could hack up your `WORKSPACE` to override the pip_import repo rules with equivalent versions that have a modified `python_interpreter `line.

PR #160, opened 2019-01-30
Create CODEOWNERS

dslomov:
Adding all GitHub repo admins to Global CODEOWNERS


dslomov:
lmk if we need to update this list

dslomov:
@mattmoor and @buchgr do you need/want to be CODEOWNERS?

PR #164, opened 2019-02-15
Update toolchain proposal to clarify wording around platform runtimes

brandjon:
I was going to add a boolean field "is_hermetic" to the provider, but the word "hermetic" is problematic and the field really isn't necessary. So instead I clarified some wording and added a short description of the provider constructor.


brandjon:
(x-ref to [bazelbuild/bazel#7375](https://github.com/bazelbuild/bazel/issues/7375))

PR #173, opened 2019-04-12
Add py_import rule

beasleyr-vmw:
This change adds adds a `py_import` rule to import Python eggs like
`java_import` imports Java jars.

Partially addresses bazelbuild/bazel#7312.


beasleyr-vmw:
Whoops. Insufficient testing on my part.

PR #174, opened 2019-04-12
Add py_import rule

beasleyr-vmw:
This change adds adds a `py_import` rule to import Python eggs like
`java_import` imports Java jars.

py_import.egg generated using `zipper`[1]:

```console
$ third_party/ijar/zipper Cc examples/py_import/py_import.egg examples/py_import/helloworld.py=examples/helloworld/helloworld.py examples/__init__.py= examples/py_import/__init__.py=
```

Partially addresses bazelbuild/bazel#7312.
Addresses #222 .

[1]: https://github.com/bazelbuild/bazel/tree/master/third_party/ijar

Testing Done:
```console
$ bazelisk test --override_repository=rules_python=$PWD/../.. ...
//:py_import_test                                                        PASSED in 0.6s
```


beasleyr-vmw:
How might I attract a reviewer's attention? :) cc @brandjon

aiuto:
Can you sync and merge to bring this up to date. It looks pretty reasonable to me.

beasleyr-vmw:
> Can you sync and merge to bring this up to date. It looks pretty reasonable to me.

Sure. Thanks for looking.

beasleyr-vmw:
Apologies for all of the push spam. The project had changed a bit since I last amended this PR.

groodt:
🚗 🔫 

This looks like it could come in handy. Can this be updated at some stage to support unpacking wheels in future as well? It doesn't even need to resolve additional dependencies, that can be left for the developer, only check that appropriately named wheels exist in `deps` from the wheel METADATA file.

Should this PR at least be verifying that you've got a valid `egg`? https://setuptools.readthedocs.io/en/latest/formats.html#the-internal-structure-of-python-eggs

There's not much to validate, but at least check the mandatory fields of PKG-INFO?
https://www.python.org/dev/peps/pep-0314/

README.md
```diff
@@ -20,9 +20,9 @@ accordingly.
 ## Overview
 
 This repository is the home of the core Python rules -- `py_library`,
-`py_binary`, `py_test`, and related symbols that provide the basis for Python
-support in Bazel. It also contains packaging rules for integrating with PyPI
-(`pip`). Documentation lives in the
+`py_binary`, `py_test`, `py_import`, and related symbols that provide the
```
alexeagle, line 23: I don't think this list is meant to be exhaustive, plus it's clearly very insufficient for documenting the rules. I'm looking at generating a better docsite similar to the https://github.com/bazelbuild/rules_nodejs one.
I'll revert this file to keep the delta small

python/defs.bzl
```diff
@@ -71,6 +71,47 @@ def py_test(**attrs):
...
+            ]),
+            imports = depset(direct = import_paths, transitive = [
+                dep[PyInfo].imports
+                for dep in ctx.attr.deps
+            ]),
+        ),
+    ]
+
+py_import = rule(
+    doc = "This rule allows the use of Python eggs as libraries for " +
```
alexeagle, line 98: I'll update the docstring here, in particular I think we need to mention whl_library as the alternative if you don't have .egg files

alexeagle (approved):
Thanks @beasleyr-vmw !!

alexeagle:
you mention `examples/py_import/helloworld.py` in the description, but it's not included in the PR, is that intentional?

alexeagle:
I just found that as well, trying to get the new integration test to pass, thanks!

alexeagle:
@googlebot I consent.
you silly

Issue #178, opened 2019-04-16
Allow users to override system Python


beasleyr-vmw:
rules_python's `pip_import` and `whl_library` repository rules call out to `python`. On CentOS 6, the system Python is ancient---Python 2.6.6---and it does not satisfy rules_python's need for dict comprehensions. I have newer Python distributions available, but no way to get `pip_import` to use them---prepending the alternate Python's location to my `PATH` doesn't work.

It'd be great if rules_python had some way to let me override its use of the system python.

brandjon:
Dedupping this to #249.

PR #179, opened 2019-04-16
pip_import: Let users override system Python via PYTHON env var

beasleyr-vmw:
CentOS 6 provides Python 2.6.6, which is too old to satisfy `pip_import`'s
dict comprehension requirement.  I need some way to point `pip_import` to
an alternate Python distribution.

This change adds system Python override to `pip_import` and `whl_library`
via the `PYTHON` environment variable.

Resolves https://github.com/bazelbuild/rules_python/issues/178.

### DIFF
--- a/python/pip.bzl
+++ b/python/pip.bzl
@@ -24,7 +24,7 @@ def _pip_import_impl(repository_ctx):
 
     # To see the output, pass: quiet=False
     result = repository_ctx.execute([
-        "python",
+        repository_ctx.os.environ.get("PYTHON", "python"),
         repository_ctx.path(repository_ctx.attr._script),
         "--name",
         repository_ctx.attr.name,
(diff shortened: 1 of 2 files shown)


brandjon:
Hi Ryan, thanks for the PR. The recently merged #252 allows you to include a `python_interpreter` attr, and we're about to add a `pip3_import` (#256) as a wrapper.

Compared with those approaches, this one allows the user to easily inject an alternate Python interpreter by setting an env var. But I don't think it's necessary to directly support that use case if we have an attribute users can set in their WORKSPACE. Note that the env var behavior can now be obtained by writing a macro wrapper around `pip_import`.

In light of this, I'll let #256 mark this PR closed.

beasleyr-vmw:
> In light of this, I'll let #256 mark this PR closed.
Sounds good.

Issue #199, opened 2019-07-03
py_wheel generated package structure


sha1n:
I'm trying to build wheel packages using the experimental `py_wheel` rule and after looking at the code, I realize that the paths inside the package are starting from the workspace root and I cannot change that using rule parameters. My package code is located deep in a sub-tree of my repo and my imports are *absolute*, but relative to that sub-tree. This setup results in a broken wheel package, since my imports are not aligned with the filesystem structure. 

In addition, I would love to be able to have a `__main__.py` script at the root of the distributed package, so that my users can run the tools that I provide using `python -m <package-name>` after they install the package. If I understand the current implementation correctly, the only way I can make that work is to put my main file in the root of the repo, which is not an option for me.

I believe this should be relatively easy to solve, if we add an attribute to the `py_wheel` rule, that will allows the user to specify the package root path. This is somewhat similar to the `py_test` and `py_binary` `imports` attribute that allows you to add custom import paths for the interpreter based on your tree structure.

Is there anything I can do to implement my requirements considering my constraints and using existing functionality? If not, would you consider a PR that adds the extra attributes I described above, or any other approach you can think of?

Thanks!

pstradomski:
This is not exactly the same as the "imports" rule, but I think this is an adequate way to achieve compatibility.

There's a separate request to allow adding a prefix to paths in the resulting package. If I understand your request is about removing a prefix from paths in the resulting package.

sha1n:
Yes that what I need. I updated #200 today to support multiple prefixes, following your review.

Thanks!

sha1n:
Fixed in #200

PR #200, opened 2019-07-03
added the ability to specify a custom wheel package root

sha1n:
This PR suggests a possible way to address the issues described in https://github.com/bazelbuild/rules_python/issues/199. I added an optional attribute `package_root_path` to the `py_wheel` rule, that specifies the root of the package, relative to the workspace path. 

This patch only fixes the problem for cases in which the entire package is under the same path. If the package bundles dependencies from other paths, those will not be fixed. It would be nice if we could define the prefix on the sources and strip the prefix when we create the final package. Not sure how to do that and will be very happy to hear suggestions, or alternative approaches.

Thanks.

```
py_wheel(
    package_root_path = "experimental"
    name = ...
    distribution = ...
    python_tag = ...
    version = ...
    deps = ...
)
```

### DIFF
--- a/experimental/python/wheel.bzl
+++ b/experimental/python/wheel.bzl
@@ -102,7 +102,7 @@ def _py_wheel_impl(ctx):
     args.add("--abi", ctx.attr.abi)
     args.add("--platform", ctx.attr.platform)
     args.add("--out", outfile.path)
-    args.add("--strip_path_prefix", ctx.attr.strip_path_prefix)
+    args.add_all(ctx.attr.strip_path_prefixes, format_each = "--strip_path_prefix=%s")
 
     args.add_all(inputs_to_package, format_each = "--input_file=%s", map_each = _input_file_to_arg)
 
@@ -243,9 +243,9 @@ to refer to the package in other packages' dependencies.
         "license": attr.string(default = ""),
         "classifiers": attr.string_list(),
         "description_file": attr.label(allow_single_file = True),
-        "strip_path_prefix": attr.string(
-            default = "none",
-            doc = "path to the root of the generated package (default is workspace path)",
+        "strip_path_prefixes": attr.string_list(
+            default = [],
+            doc = "path prefixes to strip from files added to the generated package",
         ),
         # Requirements
         "requires": attr.string_list(
--- a/experimental/python/wheel.bzl
+++ b/experimental/python/wheel.bzl
@@ -102,7 +102,7 @@ def _py_wheel_impl(ctx):
     args.add("--abi", ctx.attr.abi)
     args.add("--platform", ctx.attr.platform)
     args.add("--out", outfile.path)
-    args.add("--package_root_path", ctx.attr.package_root_path)
+    args.add("--strip_path_prefix", ctx.attr.strip_path_prefix)
 
     args.add_all(inputs_to_package, format_each = "--input_file=%s", map_each = _input_file_to_arg)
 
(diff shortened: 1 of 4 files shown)


experimental/python/wheel.bzl
```diff
@@ -242,6 +243,10 @@ to refer to the package in other packages' dependencies.
         "license": attr.string(default = ""),
         "classifiers": attr.string_list(),
         "description_file": attr.label(allow_single_file = True),
+        "package_root_path": attr.string(
```
pstradomski: I'd suggest renaming this to something like "strip_path_prefix" to make it more explicit what happens here. In particular, I'd read package_root_path as path to prepend (there's a request to do that as well).

In addition, perhaps it would make sense to make this a list so that multiple prefixes to strip could be provided?

experimental/rules_python/wheelmaker.py
```diff
@@ -93,8 +94,18 @@ def add_string(self, filename, contents):
 
     def add_file(self, package_filename, real_filename):
         """Add given file to the distribution."""
+        def strip_prefix(path, prefix):
+            if path.startswith(prefix):
+                return path[len(prefix):]
+            else:
+                return path
+
+        resolved_package_filename = package_filename \
```
pstradomski: Using backslashes for line continuaiton is frowned upon.

Can you change strip_prefix to be no-op when prefix is None and call it unconditionally?

experimental/rules_python/wheelmaker.py
```diff
@@ -254,7 +268,9 @@ def main():
                     python_tag=arguments.python_tag,
                     abi=arguments.abi,
                     platform=arguments.platform,
-                    outfile=arguments.out) as maker:
+                    outfile=arguments.out,
+                    package_root_path=arguments.package_root_path,
```
pstradomski: If I recall correctly having trailing comma on argument list is not supported on all python versions. Please remove it.

experimental/rules_python/wheelmaker.py
```diff
@@ -93,8 +94,18 @@ def add_string(self, filename, contents):
...
+            else:
+                return path
+
+        resolved_package_filename = package_filename \
+            if self._package_root_path is None \
+            else strip_prefix(package_filename, self._package_root_path)
+
         # Always use unix path separators.
-        arcname = package_filename.replace(os.path.sep, '/')
+        arcname = resolved_package_filename.replace(os.path.sep, '/')
```
pstradomski: Shouldn't this happen before strip_prefix? I'd suppose BUILD file would specify the prefix in unix style regardless of build platform, so stripping would need to use '/' separator to work reliably on windows.

experimental/python/wheel.bzl
sha1n: Done

experimental/rules_python/wheelmaker.py
sha1n: Done

experimental/rules_python/wheelmaker.py
sha1n: Done

experimental/rules_python/wheelmaker.py
sha1n: Correct! Fixed.
I incorporated all the logic into one function that returns the final `arcname` in order to avoid another variable that can be accidentally used instead of the correct one.

experimental/python/wheel.bzl
sha1n: Pay attention to the fact that the order of prefixes is important. I created separate tests to show that and documented it in the `wheelmaker` argument and hinted that in the rule attribute description.

pstradomski:
Looks good to me.

pstradomski:
You should be able to "Squash and merge" the change now.

sha1n:
> You should be able to "Squash and merge" the change now.

@pstradomski I don't have any of the merge options, only close and comment.

pstradomski:
@sha1n Weird. Merged it for you.

brandjon:
Merging PRs requires write permission, which Pawel has in this repository (for ownership of the rules under the experimental dir).

PR #202, opened 2019-07-19
Add exports for core Python logic that's bundled with Bazel

brandjon:
This series of commits adds exports to `rules_python` for Bazel-packaged Python rule logic. Most importantly, this includes the native rules and providers like `py_binary` and `PyInfo`, but it also includes Starlark-defined constructs that are distributed in `@bazel_tools`.

Work toward bazelbuild/bazel#8893. These exports constitute the new API we wish to migrate users to for that change.

Blocked on:
- [x] Still needs redirects for the runfiles library
- [x] Need to rename the canonical workspace name from `@io_bazel_rules_python` to `@rules_python` (#203)

### DIFF
--- a/python/python.bzl
+++ b/python/python.bzl
@@ -12,26 +12,44 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-def py_library(*args, **kwargs):
-    """See the Bazel core py_library documentation.
+"""Re-exports for some of the core Bazel Python rules.
 
-    [available here](
-    https://docs.bazel.build/versions/master/be/python.html#py_library).
+This file is deprecated; please use the exports in defs.bzl instead. This is to
+follow the new naming convention of putting core rules for a language
+underneath @rules_<LANG>//<LANG>:defs.bzl. The exports in this file will be
+disallowed in a future Bazel release by
+`--incompatible_load_python_rules_from_bzl`.
+"""
+
+def py_library(*args, **kwargs):
+    """See the Bazel core [py_library](
+    https://docs.bazel.build/versions/master/be/python.html#py_library)
+    documentation.
+
+    Deprecated: This symbol will become unusuable when
+    `--incompatible_load_python_rules_from_bzl` is enabled. Please use the
+    symbols in `@rules_python//python:defs.bzl` instead.
     """
     native.py_library(*args, **kwargs)
 
 def py_binary(*args, **kwargs):
-    """See the Bazel core py_binary documentation.
+    """See the Bazel core [py_binary](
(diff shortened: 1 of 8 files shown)


python/python.bzl
```diff
@@ -12,26 +12,44 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+"""Re-exports for some of the core Bazel Python rules.
```
clintharrison: Can/should we follow a similar pattern as [skylib's old `lib.bzl`](https://github.com/thomasvl/bazel-skylib/blob/79c9da076f7b869b1c09b2f7dd16b2458a34f2f2/lib.bzl#L17-L21) and `print()` a warning when this file is used?

I know it can get a bit noisy, but otherwise I suspect no one will see these deprecation warnings.

brandjon:
@clintharrison: I got an email of your post but can't see a comment here, did you delete it or is github being weird?

In response to your point about warning when a deprecated API (`python.bzl`) is used: While (in my opinion) there's not necessarily a clear consensus about whether to use or proscribe `print()` for deprecation warnings, I personally try to avoid it. The philosophy is to instead make warnings into failures. This may block a user from upgrading to a new version of the repo until they migrate, but that's more acceptable for repo rules than it is for Bazel itself (since users have more flexibility about not upgrading).

In addition, we can sometimes make it a soft failure that can be bypassed during a transition period (a la Bazel's `--incompatible_*` flags). In this case it might be done by having the macro inject an extra dependency on a target that fails unless a certain flag is passed on the command line (implemented using Starlark-based configurations / build settings).

clintharrison:
@brandjon so

> @clintharrison: I got an email of your post but can't see a comment here, did you delete it or is github being weird?
> 
> In response to your point about warning when a deprecated API (`python.bzl`) is used: While (in my opinion) there's not necessarily a clear consensus about whether to use or proscribe `print()` for deprecation warnings, I personally try to avoid it. The philosophy is to instead make warnings into failures. This may block a user from upgrading to a new version of the repo until they migrate, but that's more acceptable for repo rules than it is for Bazel itself (since users have more flexibility about not upgrading).
> 
> In addition, we can sometimes make it a soft failure that can be bypassed during a transition period (a la Bazel's `--incompatible_*` flags). In this case it might be done by having the macro inject an extra dependency on a target that fails unless a certain flag is passed on the command line (implemented using Starlark-based configurations / build settings).

Sorry, I intended on reposting it as a full comment, rather than an inline one, but you beat me to the reply :)

Thanks for explaining the reasoning there -- I think that approach makes a lot of sense!

Issue #208, opened 2019-07-24
Workflow for regenerating par files is broken


brandjon:
Running the `update_tools.sh` script within the docker contain leads to an error. First there's an incompatible change issue

> sequence element must be a string (got 'path'). See https://github.com/bazelbuild/bazel/issues/7802 for information about --incompatible_string_join_requires_strings.

but this is a red herring, since it occurs within the control flow of reporting another error. Bypassing it by adding `--incompatible_string_join_requires_strings=false` on the bazel command line in the Dockerfile, we get

> java.io.IOException: Cannot run program "git" (in directory "/root/.cache/bazel/_bazel_root/eb88a7e7a2050c89c11b77b84df21520/external/io_bazel_rules_sass"): error=2, No such file or directory

So it appears that we need to ensure git is available within the docker container.

This is blocking #203, which itself blocks #202.

brandjon:
#210 proposes to pin the bazel dependency within the docker image when regenerating the tools. This sounds like the right idea, in addition to the fix to be compatible with a newer bazel.

The docker image is also run with `--no-cache`. As best I can tell, there's no need to disregard the cache by default. We don't pin other dependencies in the image, so it's not like avoiding the cache ensures determinism; it just ensures agreement with other machines that attempt to regenerate the files at roughly the same time. I'll see about making that an opt-in.

brandjon:
Reopening for some continued cleanups...

brandjon:
I notified Github support of the odd issue where this bug appears as "Open" despite having also been closed by d852e8a. They responded so closing manually now.

PR #209, opened 2019-07-24
Refactor update_tools.sh workflow and regenerate par files

brandjon:
I created a (not yet pushed) fix for #208, and reran `update_tools.sh` in the docker container. This PR is to verify these files in CI.

Point of order: Do we even have integration tests for the precompiled par files, or only of the source code? Must look into that. We can also test other projects manually against this commit.

The regenerated `piptool.par` and `whltool.par` files grew by 9731 bytes and 9733 bytes respectively, relative to a total filesize of 5.7 MB and 1.4 MB respectively. The similar increases suggests that the growth is due to changes in common tooling (i.e. subpar), and is any case not in and of itself problematic.

Assuming these files are good, I'll merge the fix for #208 and then update this PR.


brandjon:
> grew by 9731 bytes and 9733 bytes respectively

FTR that's associated with bumping the bazel version to 0.28.

PR #210, opened 2019-07-24
pin upstream bazel to 0.24

ptarjan:
Fixes: #208
It looks like you just reference the latest bazel, which means your builds will just break on bazel releases. Instead I suggest pinning to a certain version and then you can walk the version up at your leisure without master sitting there broken.

### DIFF
--- a/tools/update_tools/Dockerfile
+++ b/tools/update_tools/Dockerfile
@@ -1,10 +1,9 @@
 FROM gcr.io/gcp-runtimes/ubuntu_16_0_4:latest
 
 # Install Bazel (https://docs.bazel.build/versions/master/install-ubuntu.html)
-RUN apt-get update -y && apt-get install openjdk-8-jdk -y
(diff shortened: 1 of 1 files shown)


brandjon:
Thanks! Pinning the bazel version is probably a good idea even if we want to keep compatibility with the bleeding edge. Will merge later today.

brandjon:
Thinking about this more, I don't think we need to change from the apt repo to `wget`ting the install script. It looks like old versions are not pruned from the bazel-apt package source, so that should still be viable, plus it avoids needing to manually install dependencies (beyond openjdk, as [described](https://docs.bazel.build/versions/0.28.0/install-ubuntu.html#step-1-install-the-jdk-optional) on the install page).

I'll see if pinning it in the install line works, then maybe refactor this file a bit.

brandjon:
Turns out we don't keep old versions around in our apt repo after all, despite the fact that the files themselves are still present and reported in the XML page at the apt's URL. Oh well.

The other options someone suggested were

- `wget`ting the binary and running it directly without installation, or

- using [bazelisk](https://github.com/bazelbuild/bazelisk)

I think your approach is simpler.

PR #211, opened 2019-07-24
Fix extra dependencies

ptarjan:
This will remove duplicate dependencies. Without this patch we are seeing errors like:

```
Label '@pypi__enum34_1_1_6//:pkg' is duplicated in the 'deps' attribute of rule 'pkg'
```


ptarjan:
@brandjon gentle ping

brandjon:
Regenerated the par files, rebased, and pushed to your branch. Merging momentarily.

PR #212, opened 2019-07-26
Rename canonical workspace name to "@rules_python"

### LINKED ISSUE #203: Rename canonical workspace to `@rules_python`
The canonical workspace name -- i.e. the name appearing in the `workspace()` declaration, documentation, examples, and code -- is currently `@io_bazel_rules_python`. We should change this to `@rules_python` for consistency with `@rules_cc`, `@rules_java`, etc.

This reflects how this repository is (or ought to be) the canonical home for all things Python in Bazel, and is especially important in light of bazelbuild/bazel#8893. As part of that work, Bazel itself may begin to assume that the core Python rules (`py_binary`, etc.) are available under a `@rules_python` repo.

brandjon:
Fixes #203.


README.md
```diff
@@ -4,6 +4,10 @@ Status: This is **ALPHA** software.
 
 [![Build status](https://badge.buildkite.com/0bcfe58b6f5741aacb09b12485969ba7a1205955a45b53e854.svg)](https://buildkite.com/bazel/python-rules-python-postsubmit)
 
+## Recent updates
```
c-parsons: I'm often a little unclear on whether README.md  refers to the most recent release of the repository (and thus should only be updated alongside cutting a new release) or whether it affects state at HEAD. I leave it up to you -- just wanted to make sure you've thought about its role in the release process of rules_python :)

brandjon: Good point. I don't feel too bad about it being HEAD given that this repo doesn't even have any tags, but I don't know what a general solution would be. Aside from not iterating so quickly on the API. :)

Issue #118, opened 2018-09-11
Is the --default-timeout flag working?


badc0re:
Is there a way to provide the timeout flag when installing with pip, or does the PIP_DEFAULT_TIMEOUT env variable works instead?

https://pip.pypa.io/en/stable/user_guide/#environment-variables

brandjon:
Should be addressed by #217.

PR #217, opened 2019-08-01
Add a timeout attribute to pip_import rules.

ChrisCummins:
The combination of a large `requirements.txt` file and a slow network can
cause the default timeout for `repository_ctx.execute()` to be reached.
This patch exposes that timeout to the `pip_import()` caller so they can
manually specify a longer timeout. E.g.

    pip_import(
        name = "requirements",
        requirements = "//long_list_of_requirements.txt",
        timeout = 3600,
    )

### DIFF
--- a/docs/pip.md
+++ b/docs/pip.md
@@ -5,7 +5,7 @@
 ## pip_import
 
 <pre>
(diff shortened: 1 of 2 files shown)


brandjon:
Sounds reasonable, but this only affects Bazel's timeout when evaluating a repo rule. What about pip's own timeout? See also #118.

michael4338:
@brandjon , the pip itself uses "pip wheel" instead of "pip install", the former does not have a "timeout" arg.

brandjon:
Thanks for the PR, merging.

PR #218, opened 2019-08-02
Update README and add py_repositories()

brandjon:
Update the README file to explain the difference between the core Python rules
and the packaging rules. Add the py_repositories() hook and recommend that
users add it to their WORKSPACE files.

Background on the py_repositories() hook: I don't know what this might be
useful for at the moment, but I found pip_repositories() to be helpful when I
added a deprecation message for renaming the workspace to @rules_python. So
it's probably a good idea to have this just in case.

Work toward bazelbuild/bazel#9006.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,12 +30,11 @@ expected for any language supported in Bazel. The packaging rules provide
 support for integration with dependencies that, in a non-Bazel environment,
 would typically be managed by `pip`.
 
-Historically, the core rules have been bundled with Bazel itself, and some of
-them are even implemented natively (not in Starlark). The Bazel team is in the
-process of transitioning these rules to live in bazelbuild/rules_python
-instead. In the meantime, all users of Python rules in Bazel should migrate
-their builds to load these rules and their related symbols (`PyInfo`, etc.)
-from `@rules_python` instead of using built-ins or
+Historically, the core rules have been bundled with Bazel itself. The Bazel
+team is in the process of transitioning these rules to live in
+bazelbuild/rules_python instead. In the meantime, all users of Python rules in
+Bazel should migrate their builds to load these rules and their related symbols
+(`PyInfo`, etc.) from `@rules_python` instead of using built-ins or
 `@bazel_tools//tools/python`.
 
 ## Setup
@@ -53,10 +52,11 @@ git_repository(
     commit = "{HEAD}",
 )
 
+# This call should always be present.
 load("@rules_python//python:repositories.bzl", "py_repositories")
 py_repositories()
 
-# Only needed if you're using the packaging rules.
+# This one is only needed if you're using the packaging rules.
 load("@rules_python//python:pip.bzl", "pip_repositories")
 pip_repositories()
 ```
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -10,21 +10,38 @@ Status: This is **ALPHA** software.
...
-rules that are available from Bazel core are simple aliases to that bundled
-functionality.  On top of that, this repository provides support for installing
-dependencies typically managed via `pip`.
+This repository provides two sets of Python rules for Bazel. The core rules
+provide the essential library, binary, test, and toolchain rules that are
+expected for any language supported in Bazel. The packaging rules provide
+support for integration with dependencies that, in a non-Bazel environment,
+would typically be managed by `pip`.
+
+Historically, the core rules have been bundled with Bazel itself, and some of
```
katre: I'm not sure you need to mention that the existing version of the rules are native, but if you do, mention that the future rules_python versions will be implemented in Starlark.

README.md
```diff
@@ -36,19 +53,18 @@ git_repository(
     commit = "{HEAD}",
```
katre: I know this already exists but it seems like worrying advice, since people will just copy and paste this.

README.md
```diff
@@ -36,19 +53,18 @@ git_repository(
...
 )
 
-# Only needed for PIP support:
-load("@rules_python//python:pip.bzl", "pip_repositories")
+load("@rules_python//python:repositories.bzl", "py_repositories")
```
katre: Add a similar comment that this is needed to use py_library, etc.

python/repositories.bzl
```diff
@@ -0,0 +1,14 @@
+"""This file contains macros to be called during WORKSPACE evaluation.
+
+For historic reasons, pip_repositories() is defined in //python:pip.bzl.
+"""
+
+def py_repositories():
+    """Pull in dependencies needed to use the core Python rules.
+
+    At the moment this is a placeholder hook, in that it does not actually pull
```
katre: This should be a normal comment, not part of the documentation.

README.md
brandjon: Removing reference to native rules, it's probably not of interest to the reader.

README.md
brandjon: I think that's ok at the moment. I'd rather they copy/paste, get immediately broken, and look for a recent commit, than that they copy/paste, don't get broken right away, but get an old commit (assuming we're not updating this readme every commit).

In the near future I'll cut a tag `0.1.0`, and change this to a true copy/paste recipe. (Tony is also working on a way to auto-generate recipes as part of the tag description, which sounds pretty cool.)

README.md
brandjon: Clarified. I don't think it's important to say it's for the core rules specifically since all uses of this repo should also include this call.

python/repositories.bzl
brandjon: Done, and updated the analogous docstring in pip.bzl's `pip_repositories()`.

README.md
katre: SGTM

PR #223, opened 2019-08-09
Comply with the standards of the Bazel federation

fweikert:
This commit allows rules_python to be a member of the Bazel federation, since it adds the required bzl files related to setup and dependencies.
Moreover, it also changes the WORKSPACE to fetch all non-pip_import dependencies through the federation.

rules_python_internal_setup() in internal_setup.bzl still needs some work since some of the dependencies don't export their proper setup functions yet, which means that the present commit has to work around that particular problem.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -27,7 +27,7 @@ load("@bazel_federation//:repositories.bzl", "rules_python_deps")
 rules_python_deps()
 
 load("@bazel_federation//setup:rules_python.bzl",  "rules_python_setup")
-rules_python_setup()
+rules_python_setup(use_pip=True)
 
 load("//:internal_deps.bzl", "rules_python_internal_deps")
 rules_python_internal_deps()
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -14,160 +14,23 @@
 
 workspace(name = "rules_python")
 
-load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
-load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_file")
-
-#################################
-# Stardoc and its dependencies. #
-#################################
-
-# Initialization taken from:
-# https://skydoc.bazel.build/docs/getting_started_stardoc.html
-
-git_repository(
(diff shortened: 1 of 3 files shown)


brandjon:
Exciting! I'm virtually OOO today but will review first thing Monday.

WORKSPACE
```diff
@@ -14,160 +14,23 @@
...
-_version_install()
+load("@bazel_federation//:repositories.bzl", "rules_python_deps")
+rules_python_deps()
 
-pip_import(
-    name = "examples_boto",
-    requirements = "//examples/boto:requirements.txt",
-)
+load("@bazel_federation//setup:rules_python.bzl",  "rules_python_setup")
+rules_python_setup()
```
brandjon: In light of bazelbuild/bazel-federation#34, please set `use_pip=True`.

WORKSPACE
```diff
@@ -14,160 +14,23 @@
...
 
-_version_install()
+load("@bazel_federation//:repositories.bzl", "rules_python_deps")
+rules_python_deps()
 
-pip_import(
-    name = "examples_boto",
-    requirements = "//examples/boto:requirements.txt",
-)
+load("@bazel_federation//setup:rules_python.bzl",  "rules_python_setup")
```
brandjon: If rules_python's deps change at head, does that mean I update this function's definition in the federation at head? Or do I override the definition in this WORKSPACE file, and let the federation figure out how to merge my updated requirements into the deps function later on?

internal_deps.bzl
```diff
@@ -0,0 +1,46 @@
...
+    futures_3_whl()
+    google_cloud_language_whl()
+    grpc_whl()
+    mock_whl()
+
+    piptool()
+    examples()
+
+
+def  piptool():
```
brandjon: Nit: extra whitespace

WORKSPACE
fweikert: Yes, you would update deps and setup functions in the federation. Users shouldn't use the federation at head.

PR #228, opened 2019-08-14
Rename //rules_python package to //packaging

### LINKED ISSUE #227: `whl_test` fails when invoked from outside main repo
Reported by @fweikert.

Similar to #225, the test target `//rules_python:whl_test` fails when invoked in the Bazel federation's CI. This one happens because the `rules_python` top-level package name conflicts with the `rules_python` repo name (when we're not the main repo), and the Python import mechanism in Bazel does not handle this ambiguity well. I.e., it fails to find whl.py because it's looking inside rules_python/ instead of inside rules_python/rules_python.

Solution is probably to rename the `//rules_python` package to avoid this naming conflict. The targets in that package currently have public visibility, but I think they can actually be made private -- the rules themselves don't force public visibility because they only depend on the precompiled versions, not the sources.

brandjon:
This avoids a name clash between the package name and the repo name, which
prevented whl_test from running in the Bazel federation's CI.

Regenerated the par files.

Tested that this still works in environments where the system Python has a
library named "packaging" installed, by creating a virtualenv, pip installing
"packaging", and running `bazel test //examples/...`.

Fixes #227.


fweikert:
Nice, thank you very much!

PR #233, opened 2019-10-03
remove unneeded references to sass and skydoc.

aiuto:
stardoc does not need  them


brandjon:
Verified by checking out the change and rerunning `update_docs.sh` for no diff.

PR #236, opened 2019-10-08
Add configurable timeout for pip_import repository fetch.

scarito1:
The default 600 second timeout can be insufficient for projects with lots of requirements or if packages need to be built from source.


ali5h:
we had this issue and implemented parallel fetch for pip packages in https://github.com/ali5h/rules_pip. The caveat is you must use pip-compile to generate a full `requirements.txt` so the rules can fetch the packages in parallel.

psigen:
Duplicate of https://github.com/bazelbuild/rules_python/pull/217?

That one seems to include a docstring though.

brandjon:
Dedupping to #217.

PR #237, opened 2019-10-09
Update README

### LINKED ISSUE #229: Status of rules_python
Is this repository still Experimental / Alpha software?
Or do we recommend Bazel users to use it (instead of calling directly the native rules)?

It would be nice to clarify the documentation:
- https://github.com/bazelbuild/rules_python#experimental-bazel-python-rules
- https://docs.bazel.build/versions/master/be/python.html

brandjon:
Expand the Overview. Add more context on core rules vs packaging rules and
level of support. Add instructions for using Bazel Federation. Move pip stuff
to sub-headings.

Fixes #229.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,12 +1,12 @@
 # Python Rules for Bazel
 
+[![Build status](https://badge.buildkite.com/0bcfe58b6f5741aacb09b12485969ba7a1205955a45b53e854.svg)](https://buildkite.com/bazel/python-rules-python-postsubmit)
+
 ## Recent updates
 
 * 2019-07-26: The canonical name of this repo has been changed from `@io_bazel_rules_python` to just `@rules_python`, in accordance with [convention](https://docs.bazel.build/versions/master/skylark/deploying.html#workspace). Please update your WORKSPACE file and labels that reference this repo accordingly.
 
 ## Overview
-
-[![Build status](https://badge.buildkite.com/0bcfe58b6f5741aacb09b12485969ba7a1205955a45b53e854.svg)](https://buildkite.com/bazel/python-rules-python-postsubmit)
 
 This repository is the home of the core Python rules -- `py_library`,
 `py_binary`, `py_test`, and related symbols that provide the basis for Python
@@ -157,6 +157,7 @@ fix is available to automatically migrate BUILD and .bzl files to add the
 appropriate `load()` statements and rewrite uses of `native.py_*`.
 
 ```sh
+# Also consider using the -r flag to modify an entire workspace.
 buildifier --lint=fix --warnings=native-py <files>
 ```
 
@@ -189,7 +190,7 @@ The .par files need to be regenerated whenever their sources are updated. This
 can be done by running
 
 ```shell
-# Optionally pass --nodocker if Docker is not available on your system.
+# You can pass --nodocker if Docker is not available on your system.
 ./update_tools.sh
 ```
 
--- a/README.md
+++ b/README.md
@@ -1,67 +1,77 @@
-# Experimental Bazel Python Rules
-
-Status: This is **ALPHA** software.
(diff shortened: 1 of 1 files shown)


brandjon:
Setting Tony as reviewer, but I'm going to force submit in the interest of getting a release out tonight since I'm OOO tomorrow. Feel free to leave me comments to forward-fix afterwards.

brandjon:
It looks kind of strange that the WORKSPACE boilerplate produced by the //distro:relnotes target doesn't mention the federation at all. Is that because it's expected that `py_repositories` would load the federation? It doesn't, but that's because our distributable has no other repos as dependencies.

README.md
```diff
@@ -1,67 +1,77 @@
...
+setup" heading. This will look something like the following:
 
-## Setup
+```python
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+http_archive(
+    name = "rules_python",
+    # NOT VALID: Replace with actual version and SHA.
+    url = "https://github.com/bazelbuild/rules_python/releases/download/<RELEASE>/rules_python-<RELEASE>.tar.gz",
+    sha256 = "<SHA>",
```
laurentlb: I'd encourage you to put an actual release version and sha. I think it's a better user-experience when the code can be copy-pasted.

README.md
```diff
@@ -1,67 +1,77 @@
...
+)
+load("@rules_python//python:repositories.bzl", "py_repositories", "rules_python_toolchains")
+py_repositories()
+rules_python_toolchains()
+```
 
-To use this repository, first modify your `WORKSPACE` file to load it and call
-the initialization functions as needed:
+Otherwise, you may import rules_python in a standalone way by copying the
+following:
```
laurentlb: I don't see the difference between the two blocks of code. You just replaced `http_archive` with `git_repository`? (or did I miss something?)

README.md
```diff
@@ -72,12 +82,14 @@ py_binary(
 )
```
laurentlb: Replace the dots with: `srcs = ["main.py"]`

README.md
```diff
@@ -134,34 +146,57 @@ format in the future.
...
+```
+
+Currently the WORKSPACE file needs to be updated manually as per [Getting
+started](#Getting-started) above.
+
+Note that Starlark-defined bundled symbols underneath
+`@bazel_tools//tools/python` are also deprecated. These are not yet rewritten
+by buildifier.
+
 ## Development
```
laurentlb: In the future, consider moving this to a different page.

README.md
brandjon: I didn't do this because I thought it'd make the hash self-referential. Then I realized the README isn't included in the distributable being hashed. Will change.

README.md
brandjon: It started out more different, but then converged when I fixed errors in the federation boilerplate. Now I'm at a loss to understand how exactly the federation comes into play here. @aiuto, can you comment on what in particular a user does to use the federation? It's possible I conflated using the federation with using the distributable tar.gz.

README.md
brandjon: You mean like a contributing.md file?

laurentlb: Yes. I'd keep the README.md file focused on the user.

README.md
brandjon: Discussed offline with Tony. What I had is a comparison of using the distributable vs source checkout, but using the federation is different. Adding a third way to write WORKSPACE files, arranged in order of preference.

README.md
brandjon: Done.

README.md
brandjon: Done.

brandjon:
See #239 for post-merge fixes.

PR #243, opened 2019-10-21
Add pytest_test support to rules_python.

### LINKED ISSUE #240: pytest support
`py_test` is a Bazel rule and `pytest` is a Python library for writing and running tests 

https://docs.pytest.org 
https://github.com/pytest-dev/pytest

I've found two third-party Bazel rules for running tests written with pytest [bazel_rules_pex](https://github.com/benley/bazel_rules_pex/blob/0c5773db01ab8aeb3ae749b2fc570749b93af41f/pex/pex_rules.bzl#L458) and [rules_pyz](https://github.com/TriggerMail/rules_pyz/blob/c2cc31b2fa749cb81f7e9c57ec6385b25a32cf80/rules_python_zip/rules_python_zip.bzl#L350), but both projects are unmaintaned and don't work anymore. **Edit 2020-01-15**: see https://github.com/ali5h/rules_pip/blob/b4e94d0ddc95367217b7059bcdb926c159d28602/defs.bzl#L139 as well.

pytest has been mentioned a few times on the bazel mailing list https://groups.google.com/forum/?nomobile=true#!searchin/bazel-discuss/pytest%7Csort:date

I'm not sure what the correct way to add support for it would be. Probably another rule but I don't know what it should be called, because the names are so similar. `pytest` `pytest_py_test`  `py_test_pytest` all seem pretty funny. Or maybe it could be an option for `py_test` or a global one in WORKSPACE. 

`pytest` [can run tests written with `unittest`](https://docs.pytest.org/en/latest/unittest.html) (and has [nicer output when a test fails](https://docs.pytest.org/en/latest/assert.html#assert)), so theoretically you could just change the `py_test` rule to run tests with pytest but that's obviously not happening.

pstradomski:
Ported from https://bitbucket.org/pstradomski/rules_python_extra.

Fixes #240.

### DIFF
--- a/experimental/examples/pytest/BUILD
+++ b/experimental/examples/pytest/BUILD
@@ -15,9 +15,9 @@ package(default_visibility = ["//visibility:public"])
 
 licenses(["notice"])  # Apache 2.0
 
-load("//python:defs.bzl", "py_library")
-load("//experimental/python:pytest.bzl", "pytest_test")
 load("@examples_pytest//:requirements.bzl", "requirement")
+load("//experimental/python:pytest.bzl", "pytest_test")
+load("//python:defs.bzl", "py_library")
 
 py_library(
     name = "conftest",
--- a/experimental/examples/pytest/BUILD
+++ b/experimental/examples/pytest/BUILD
@@ -37,6 +37,7 @@ pytest_test(
         requirement("importlib_metadata"),
         requirement("more_itertools"),
         requirement("packaging"),
+        requirement("pathlib2"),
         requirement("pluggy"),
         requirement("py"),
         requirement("wcwidth"),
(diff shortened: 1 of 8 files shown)


edbaunton:
Is this something that could be reviewed and merged some time soon? We would love to use pytest with Bazel!

ali5h:
i personally don't think this should be part of `rules_python`. With a simple macro you can achieve this,  check this as an example https://github.com/ali5h/rules_pip/blob/master/defs.bzl#L139

thundergolfer:
We already use PyTest with Bazel. It works with the standard `py_test` rule if you add the following snippet to the file run as `main`. 

```
if __name__ == "__main__":
    dir_path = os.path.dirname(os.path.realpath(__file__))
    sys.exit(pytest.main([dir_path, "-W", "ignore::DeprecationWarning"]))
```

Compared with our issues we have in our Python-Bazel codebase, this is a minor annoyance.

pstradomski:
Withdrawing as it seems there's no consensus this is the right approach.

peloton-mqiu:
> https://github.com/ali5h/rules_pip/blob/master/defs.bzl#L139

Hi @ali5h , 
is the link you shared still points to the line you originally intended?
master seems to have advanced.

PR #245, opened 2019-10-23
Enable buildifier presubmit check

hlopko:
This PR enables all buildifier checks on presubmit, plus it migrates the repo to fix violations.

### DIFF
--- a/experimental/python/wheel.bzl
+++ b/experimental/python/wheel.bzl
@@ -152,7 +152,7 @@ def _py_wheel_impl(ctx):
         data_runfiles = ctx.runfiles(files = [outfile]),
     )]
 
-def _concat_dicts(**dicts):
+def _concat_dicts(*dicts):
     result = {}
     for d in dicts:
         result.update(d)
--- a/experimental/python/wheel.bzl
+++ b/experimental/python/wheel.bzl
@@ -18,14 +18,16 @@ def _path_inside_wheel(input_file):
     # input_file.short_path is sometimes relative ("../${repository_root}/foobar")
     # which is not a valid path within a zip file. Fix that.
     short_path = input_file.short_path
-    if short_path.startswith('..') and len(short_path) >= 3:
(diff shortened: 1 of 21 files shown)


python/defs.bzl
```diff
@@ -47,6 +47,8 @@ def py_library(**attrs):
     Args:
       **attrs: Rule attributes
     """
+
+    # buildifier: disable=native-python
```
brandjon: Can we disable this just once for the whole file? (It'd be appropriate given the purpose of this file.)

python/private/reexports.bzl
```diff
@@ -36,5 +36,5 @@ the original name.
 # Don't use underscore prefix, since that would make the symbol local to this
 # file only.
 
-internal_PyInfo = PyInfo
-internal_PyRuntimeInfo = PyRuntimeInfo
+InternalPyInfo = PyInfo
```
brandjon: -1. The nonconventional name makes it clearer that this is a private name and not a public provider.

python/defs.bzl
brandjon: Looks like we can't.

experimental/python/wheel.bzl
```diff
@@ -197,18 +199,21 @@ py_wheel(
...
-
-Note it's usually better to package `py_library` targets and use
-`console_scripts` attribute to specify entry points than to package
-`py_binary` rules. `py_binary` targets would wrap a executable script that
-tries to locate `.runfiles` directory which is not packaged in the wheel.
+console_script entry points, e.g. 'experimental.examples.wheel.main:main'.
 """,
         ),
+        "description_file": attr.label(allow_single_file = True),
         # Attributes defining the distribution
```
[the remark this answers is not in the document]
brandjon: Fixed in my commit.

python/private/reexports.bzl
brandjon: Done.

brandjon:
Thanks for the PR!

PR #247, opened 2019-10-31
Add badge for incompatible flag pipeline

hlopko:
https://buildkite.com/bazel/rules-python-plus-bazelisk-migrate


brandjon:
Ha, looks like we need to migrate ourselves for `--incompatible_load_python_rules_from_bzl`. :)

PR #251, opened 2019-11-07
Fix packaging rules documentation generation

### LINKED ISSUE #250: Fix docgen for packaging rules
Needs regeneration (dunno why it wasn't generated right the first time), and the docstrings need updating to format the args correctly. Will do imminently.

brandjon:
Fix formatting of docstrings and regenerate the pages. Fixes #250.

### DIFF
--- a/docs/whl.md
+++ b/docs/whl.md
@@ -8,6 +8,31 @@
 whl_library(<a href="#whl_library-name">name</a>, <a href="#whl_library-extras">extras</a>, <a href="#whl_library-requirements">requirements</a>, <a href="#whl_library-whl">whl</a>)
 </pre>
 
+A rule for importing <code>.whl</code> dependencies into Bazel.
+
+<b>This rule is currently used to implement <code>pip_import</code>,
+it is not intended to work standalone, and the interface may change.</b>
+See <code>pip_import</code> for proper usage.
+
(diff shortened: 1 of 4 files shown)


docs/whl.md
```diff
@@ -8,6 +8,22 @@
 whl_library(<a href="#whl_library-name">name</a>, <a href="#whl_library-extras">extras</a>, <a href="#whl_library-requirements">requirements</a>, <a href="#whl_library-whl">whl</a>)
 </pre>
 
+A rule for importing `.whl` dependencies into Bazel.
+
+<b>This rule is currently used to implement `pip_import`. It is not intended to
```
aiuto: Is this the right visual? What happens with    `stuff` when you are already in ```?

brandjon: It's not in a markdown triple-backtick, it's in a Python triple quote string (in source code, not the md file).

brandjon: You can preview the md file in github -- top right ellipses -> View file.

brandjon:
I originally tried replacing all the `<code>` tags with backticks, but it didn't render properly in args. That's a known issue in stardoc and I think it can already be fixed by upgrading to a newer stardoc and/or changing which template we use. Will file that detail away in the tracking bug.

Issue #249, opened 2019-11-06
Enable straightforward use of Python 3 for packaging rules


brandjon:
Tracking bug for the various dups of this frequently proposed functionality. Will merge in a rebased version of #158, and use that to provide a `pip3_import` rule.

zlalanne:
Thanks for this! Any chance at a 0.0.2 release soon?

brandjon:
Plan is to have one corresponding to the Bazel Federation release for Bazel 2.0. Opened #261 to track.

PR #252, opened 2019-11-07
Add python_interpreter attr to pip rules

brandjon:
This is a rebase of @acumon's #158. It adds a `python_interpreter` attribute to `pip_import` and `whl_library` to allow customizing the command used to launch the Python interpreter. We'll follow this up with creating `pip2_import` and `pip3_import` as wrappers around this feature.

Diffbased against #251, which fixes documentation for these rules.

Work toward #249. Closes #158.

### DIFF
--- a/docs/pip.md
+++ b/docs/pip.md
@@ -5,7 +5,7 @@
 ## pip_import
 
 <pre>
-pip_import(<a href="#pip_import-name">name</a>, <a href="#pip_import-requirements">requirements</a>)
+pip_import(<a href="#pip_import-name">name</a>, <a href="#pip_import-python_interpreter">python_interpreter</a>, <a href="#pip_import-requirements">requirements</a>)
 </pre>
 
 A rule for importing `requirements.txt` dependencies into Bazel.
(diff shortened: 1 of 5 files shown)


keith:
I can't tell from this but is there a way to use this with your custom defined python toolchains that already encode the specific paths to interpreters? Or would this work by duplicating those paths?

thundergolfer:
It doesn't look like it. `python_interpreter` is a string not a label. We run within Nix, so I think for us it'd suffice to put the string `"python3"` and that would resolve to our Nix environment Python3.6, but even that isn't ideal.

brandjon:
This won't interact with toolchains at all because the pip rules run at `WORKSPACE` evaluation time. It's just directly running the given string as a command name using `repository_ctx.execute`. The underlying mechanism hasn't changed, it's just allowing that string to now be customized.

This has the downside that a given repo definition has no way to run different Python commands on different systems.

brandjon:
Note that 2571dcfa695186aebfd1f887ea77f1e9489978a4 is a rebased and slightly modified (by me) version of @acumon's 3e1f35ce229bf9b96e13163972f5f92600104846. The googlebot doesn't detect this because the email address on the commit metadata doesn't match, but it's the same email as on the original PR #158. Overriding the googlebot check since consent to contribute was already established in #158.

brandjon:
Force-pushed a rebase now that #251 is merged. Overriding the CLA for the same reasons, hopefully for the last time on this PR.

PR #256, opened 2019-11-12
Add pip3_import

### LINKED ISSUE #33: PyYAML pip import fails with python3
Using [this repository](https://github.com/bazelbuild/rules_python/files/1462177/foo.tar.gz):

```
$ bazel run :hello
INFO: Found 1 target...
Target //:hello up-to-date:
  bazel-bin/hello
INFO: Elapsed time: 1.910s, Critical Path: 0.03s

INFO: Running command line: bazel-bin/hello
Hello
$
```

But forcing python 3 gives an error:
```
$ bazel run --python_path=/usr/local/bin/python3 :hello 
INFO: Found 1 target...
Target //:hello up-to-date:
  bazel-bin/hello
INFO: Elapsed time: 0.174s, Critical Path: 0.01s

INFO: Running command line: bazel-bin/hello
  File "/private/var/tmp/_bazel_hwright/0c1dcd494edf6933a7c645d083d11d41/execroot/foo/bazel-out/darwin_x86_64-fastbuild/bin/hello.runfiles/foo/hello.py", line 1, in <module>
    import yaml
  File "/private/var/tmp/_bazel_hwright/0c1dcd494edf6933a7c645d083d11d41/execroot/foo/bazel-out/darwin_x86_64-fastbuild/bin/hello.runfiles/pypi__PyYAML_3_12/yaml/__init__.py", line 2, in <module>
    from error import *
ModuleNotFoundError: No module named 'error'
ERROR: Non-zero return code '1' from command: Process exited with status 1.
$
```

This doesn't happen with all packages pulled in through pip, just PyYAML.  I believe it might be related to the way that PyYAML handles py2/3 compat: instead of enabling the same code to run under both environments, PyYAML has a separate tree for each version internal to the package, and presumably selectively installs one or the other depending on the request.  It doesn't look like Bazel is picking the right tree when running under python3.

### LINKED ISSUE #165: Document how to flush out bad pip deps with bazel clean --expunge
Hello. Thank you very much for your work.

I was experimenting with Bazel + PIP rules in my Python project that used Python 3.
My default interpreter was Python 2 when I first created my Bazel project with all PIP dependencies in it.
I tried running some unit-tests with Bazel and got errors about running in Python 2.

Changed test's `python_version` to PY3 and got errors about fetched PIP dependencies working only with Python 2, not 3. It was an error somewhere in a recursive dependency (Python 2's `future` library, in particular).

I tried to make `rules_python` to work with a different interpreter, and it didn't work. It was easier to change my default Python interpreter to Python 3. 

After doing that, the dependencies were not downloaded again to work with Python 3. I tried calling `bazel clean` multiple times, that didn't work. It was hard for me to debug the problem. I spent several hours doing that as I didn't have a lot of experience with Bazel.

In order to make it work, I had to call `bazel clean --expunge`.

I wish it was documented somewhere. It would have saved me several hours of debugging. If you consider to merge one of the PRs: https://github.com/bazelbuild/rules_python/pull/85 or https://github.com/bazelbuild/rules_python/pull/158, then it will also be a problem using the functionality.

Anyway, thank you for the project. This is not really a huge problem. The issue is here to make it easier for someone else online in the future.

### LINKED ISSUE #249: Enable straightforward use of Python 3 for packaging rules

brandjon:
Add `pip3_import` as a macro built on the new `python_interpreter` attribute to `pip_import`.

Also change canonical wheel repo naming to include the `pip_import` name, so the same wheel can be named by multiple `pip_import`s.

Update README, regenerate docs and tools.

Thanks go out to the authors of similar PRs #179, #85, and #216.

Fixes #33 (yaml import problem).
Fixes #165 (documentation request).
Closes #85 (similar PR).
Closes #179 (similar PR).
Fixes #249 (feature tracking bug).


python/pip.bzl
```diff
@@ -102,6 +102,20 @@ py_binary(
 """,
 )
 
+def pip2_import(**kwargs):
+    """A wrapper around pip_import that uses the `python2` system command."""
+    # We don't advertise "Use this for PY2 programs" because it's not clear
+    # whether this alias is always available, and on most (but not all) systems
+    # `python` is Python 2 anyway.
+    pip_import(python_interpreter = "python2", **kwargs)
```
keith: This won't work on macOS I guess

brandjon: IIRC macOS has `python` and `python2.7` but not `python2`, right?

I was ambivalent about whether to add `pip2_import` at all, or just leave that use case to `pip_import` (with or without a custom `python_interpreter` attr). I guess it's not ideal to encourage people to switch Python 2 deps to use `pip2_import` and thereby break mac users by default.

I suppose we could make this macro smarter to detect what platform it's running on and choose a different default. Thoughts?

keith: Correct. Yea I think related to https://github.com/bazelbuild/bazel/pull/9402 it would be safe to default this to `python2.7` on macOS, but I agree that this isn't a huge issue since this still provides complete customizability for users

brandjon: I'm going to revert the addition of `pip2_import` for the purposes of this PR. Filed #258 to track the feature of `pip_import` reliably detecting a suitable Python interpreter.

brandjon:
@aiuto This adds a wrapper around the existing `pip_import` and adds a bunch of README text.

It also changes repo naming behavior, which breaks users who depended directly on wheel repos rather than going through the helper function as an intermediary.

I think we'd like to release this functionality in a 0.0.2 release for inclusion in the federation for Bazel 2.0.

PR #262, opened 2019-12-03
Fix immutable frozen set bug in defs.bzl

jamiesnape:
When adding tags to a native `py_library` rule that is created through a macro we
were not properly considering the case where the tags came from a different
file and therefore were frozen. This caused an error.

Analog of https://github.com/bazelbuild/rules_cc/commit/cfe68f6bc79dea602f2f6a767797f94a5904997f.


brandjon:
Thanks!

PR #277, opened 2020-02-20
Move load statements to the top to remove the Buildifier warnings

laurentlb:


laurentlb:
The CI is failing at head (https://buildkite.com/bazel/rules-python-python), and this PR should fix it.

PR #284, opened 2020-03-02
Remove .travis.yml

dhalperi:
This looks like it's left over from before the switch to buildkite.
The Travis tests are not run in precommit now, and I can't even tell
if they were run in PR #1.

This causes pre-commit errors in forks if folks have Travis
enabled in other projects.

---
This change is [Reviewable](https://reviewable.io/reviews/bazelbuild/rules_python/284)


brandjon:
Thanks!

PR #290, opened 2020-04-07
Add information regarding support

laurentlb:


keith:
Can you clarify if Google isn't providing support here who has access to review / merge changes besides folks from Google or if the Google folks will continue to provide that role?

thundergolfer:
>  who has access to review / merge changes besides folks from Google

Seconded, when I read this:

> This repository is maintained by the Bazel community.

A "maintainer" is to my mind someone who have reviewer and push permissions to the repo, and also participates/owns the release process. As far as I know currently only Google employees are "maintainers" under this definition, which if true means this repository isn't really maintained by the community.

If this PR is indication that `rules_python` is moving to a model similar to `rules_scala` where invested companies like Stripe and Wix have employees that are maintainers, that's good 🙂.

laurentlb:
Thanks for the feedback. My goal is indeed to move to a model like rules_scala.

PR #295, opened 2020-04-22
WIP: Move rules_python implementation to Starlark

abergmeier:
This is a work in progress. Mostly to identify which issues are still to be fixed before this can happen.
Contributions very welcome.


abergmeier:
Depends on https://github.com/bazelbuild/rules_python/pull/300

PR #300, opened 2020-04-29
Add initial test for py_binary working.

abergmeier:
This for now only tests whether py_binary produces an executable script.


abergmeier:
Is the Mac worker flaky?

PR #303, opened 2020-05-06
Update version in version.bzl

thundergolfer:
Just realised I [release `0.0.2` at latest `master`](https://github.com/bazelbuild/rules_python/releases) but neglected to bump this file beforehand 😖 .

### DIFF
--- a/version.bzl
+++ b/version.bzl
@@ -13,4 +13,4 @@
 # limitations under the License.
 """The version of rules_python."""
 
-version = "0.0.1"
+version = "0.0.2"


brandjon:
Apparently both 0.0.1 and 0.0.2 are marked in github as `Pre-release`. Does that justify forcibly overriding the tag, or is that still heresy even for pre-releases?

thundergolfer:
> Does that justify forcibly overriding the tag, or is that still heresy even for pre-releases?

I considered it, but think it's a user-unfriendly move and would probably only do it on project's that aren't actively used. It's a stuff up, but I'd guess that almost nobody will open up the release and get confused seeing `0.0.1` instead of `0.0.2`.

Issue #241, opened 2019-10-14
Add rule transition for setting the Python version


brandjon:
FR: Add a transition to set the Python version (PY2 vs PY3). This would enable arbitrary Starlark rules to behave just like py_binary and py_test.

I thought it wasn't possible to use Starlark-Based Configurations (SBC) to set native options, but it turns out this is doable via the `//command_line_option:foo` syntax. So it'd be a nice thing to add to rules_python. No need to put it in `@bazel_tools` given that it doesn't require any native code, and we want to migrate everything in `@bazel_tools//tools/python` to here anyway.

brandjon:
It turns out that a Starlark-based version transition has the side-effect of changing the output directory name by appending a hash of the configuration. For this reason, I don't necessarily recommend using a Starlark-based transition to replace the native transition. But I'm uploading it to `//experimental/examples` anyway for reference.

PR #309, opened 2020-05-12
Add proof-of-concept for a Starlark-based Python version transition

brandjon:
This is a demo, not production ready.

Notes for #241.


andyscott:
This is pretty neat. Do you have any sense for when the upstream functionality in Bazel will be production ready?

brandjon:
Do you mean Starlark-Based Configuration transitions? +@juliexxia for that.

There actually exists a native Python version transition but it's experimental (intended for Google's own Python 2-to-3 transition) and can be removed at any time. Ideally we don't want to add more expressiveness to the native rules when a Starlark-based solution exists. As far as I'm aware you could use this Starlark proof of concept today if you don't mind the whitelist verbiage / experimentality of Starlark transitions and the output directory name mangling.

In the long term, it's possible the path to output files may become based on their inputs rather than a hash of the config, at which point the name mangling won't matter. But that's outside my wheelhouse.

juliexxia:
Everything SBC-related in this PR is fully functional and ready to use. I'm actually in the process of removing all the experimental flags right now thought the APIs have been stable for a while.  

> As far as I'm aware you could use this Starlark proof of concept today if you don't mind the whitelist verbiage / experimentality of Starlark transitions and the output directory name mangling.
Yep!

> In the long term, it's possible the path to output files may become based on their inputs rather than a hash of the config, at which point the name mangling won't matter. But that's outside my wheelhouse
I believe bazel has purposefully never made promises about the structure of the output path and so I wouldn't depend on this changing anytime soon (and even if it does, i wouldn't depend on it staying any kind of way (: ). the output directory is changed to ensure that artifacts from transitioned targets don't class with artifacts from non-transitioned.

thundergolfer:
@juliexxia has the been any movement internally at Google on a transition to Starlark-based rules? This is an important change that can't really happen without a lot of coordination between Google and community maintainers 🙂.

Issue #257, opened 2019-11-12
Provide a way to better connect pip_import to the Python toolchain


brandjon:
(Issue forked from #37.)

This issue tracks the fact that the Python interpreter used by `pip_import` (at WORKSPACE evaluation time) is not necessarily related in any way to the Python interpreter used by the toolchain (at analysis time). This is a consequence of the fact that toolchain information is not generally available at WORKSPACE evaluation time.

This can lead to confusion when the wrong major version of Python is used (before #249, it's hard to avoid this problem). It can be even more confusing when the discrepancy is only in minor version.

For major version, we could consider a modification to `whl_library` that sets `srcs_version = "PY[2|3]ONLY"`. This would prevent accidental mixing of pip dependencies between PY2 vs PY3, at the expense of prohibiting the same pip package from being used for both when its sources support that use case. (We could also make this an optional feature, controlled by an attribute of `pip_import`.)

To address the stronger problem, where you want to guarantee that exactly the same interpreter gets used at both WORKSPACE eval time and analysis, it's not clear what we can do. Perhaps we could have a repo macro/rule that generates the toolchain in such a way that it's the same system interpreter as is passed to `pip_import`. Then at least users who use this construct wouldn't have a problem.

Note that it doesn't make sense to use the same interpreter for WORKSPACE eval and analysis if the interpreter is not a prebuilt binary, since Bazel can't build its own tools at WORKSPACE eval time.

thundergolfer:
Will leave this open as it is still an issue with `pip_import`, but will not that `pip_install` (the newer repo rule) does provide a `python_interpreter_target` attr which can be pass a label to a pre-built Python binary. 

We've had success using `rules_nixpkgs` to pull a specific Python binary like so: 

```
nixpkgs_package(
    name = "python3",
    build_file_content = """
package(default_visibility = ["//visibility:public"])
filegroup(name = "interpreter", srcs = ["bin/python"])
""",
    nix_file = "//build/nix:python3.nix",
    repositories = {"nixpkgs": "@nixpkgs//:default.nix"},
)
```

and then pass it to `pip_install` for it to use:

```python
pip_install(
    name = "pip",
    python_interpreter_target = "@python3//:bin/python3",
    requirements = "//:requirements.txt",
)
```

PR #312, opened 2020-05-16
Support python interpreter target in pip_import.

kku1993:
This allows users to use a custom python interpreter that is built by
another repository rule instead of using a pre-built interpreter binary
that is checked-in.

This tangentially addresses #257 since a common setup is to use the
custom built interpreter in the python toolchain.

For example, see: https://github.com/kku1993/bazel-hermetic-python


gergelyfabian:
This is a great feature unblocking the use of in-build Python interpreter with `pip_import`. I tested it for my use case and it works great.

PR #315, opened 2020-05-20
Fix errors with incompatible_disallow_empty_glob

gergelyfabian:
Allow py_library sources to be empty.
If --incompatible_disallow_empty_glob is set then generated py_library
targets will fail if there are no .py files.

Examples are pymssql==2.1.4 and cx-Oracle==7.2.3.

Set `allow_empty = True` for glob().

Bazel issue for incompatible_disallow_empty_glob:
https://github.com/bazelbuild/bazel/issues/8195

### DIFF
--- a/packaging/whl.py
+++ b/packaging/whl.py
@@ -133,6 +133,13 @@ parser.add_argument('--extras', action='append',
                     help='The set of extras for which to generate library targets.')
 
 def main():
+  """
+  Generate a BUILD file for an unzipped Wheel
+
+  We allow for empty Python sources as for Wheels containing only compiled C code
+  there may be no Python sources whatsoever (e.g. packages written in Cython: like `pymssql`).
+  """
+
   args = parser.parse_args()
   whl = Wheel(args.whl)
 
@@ -148,7 +155,7 @@ load("{requirements}", "requirement")
 
 py_library(
     name = "pkg",
-    srcs = glob(["**/*.py"]),
+    srcs = glob(["**/*.py"], allow_empty = True),
     data = glob(["**/*"], exclude=["**/*.py", "**/* *", "BUILD", "WORKSPACE"]),
     # This makes this directory a top-level in the python import
     # search path for anything that depends on this.


thundergolfer:
Hey @gergelyfabian, 

The `.par` files are binary files and therefore unreviewable, so for security reasons we will regenerate the .par files for you when merging your pull request. see [CONTRIBUTING.md](https://github.com/bazelbuild/rules_python/blob/master/CONTRIBUTING.md). 

you can remove them from your PR 🙂

gergelyfabian:
Thanks! Will do.

Jonathon Belotti <notifications@github.com> ezt írta (időpont: 2020. máj.
21., Cs 4:41):

gergelyfabian:
> Thanks! Will do.

I have removed the .par files.

gergelyfabian:
Just a comment: rules_python_external is not experiencing this problem, because it adds a `__init__.py` file to the generated Bazel structure:

rules_python_external (for pymssql):

```
.
├── BUILD
├── _mssql.cpython-36m-x86_64-linux-gnu.so
├── pymssql-2.1.4.dist-info
│   ├── __init__.py
│   ├── METADATA
│   ├── RECORD
│   ├── top_level.txt
│   └── WHEEL
└── pymssql.cpython-36m-x86_64-linux-gnu.so
```

While rules_python does (for pymssql):

```
.
├── BUILD
├── _mssql.cpython-36m-x86_64-linux-gnu.so
├── pymssql-2.1.4.dist-info
│   ├── METADATA
│   ├── RECORD
│   ├── top_level.txt
│   └── WHEEL
├── pymssql.cpython-36m-x86_64-linux-gnu.so
└── WORKSPACE
```

So maybe the fix should be to add the `__init__.py` file creation, rather then to add `allow_empty = True`?

gergelyfabian:
Answering my own question on `__init__.py`. In rules_python_external the decision was made, that it should not depend on some generated code (done by the plugin), but rather allow for no Python sources if this is e.g. a pip package implemented in Cython (that means no Python sources in Bazel cache for that).

So I'd argument that the fix should be to add allow_empty = True.

thundergolfer:
@gergelyfabian I think this change would be good to merge with a comment about which scenarios produce packages without py srcs. Like was done here -> https://github.com/dillon-giacoppo/rules_python_external/pull/40/files 👍

gergelyfabian:
@thundergolfer I've updated the PR.

PR #316, opened 2020-05-20
Remove files from py_library data glob() if there is space in the mid…

gergelyfabian:
…dle of path

Directory names may also contain spaces in some cases, that would fail
Bazel.
Files with spaces in their name were already excluded.

### DIFF
--- a/packaging/whl.py
+++ b/packaging/whl.py
@@ -149,6 +149,7 @@ load("{requirements}", "requirement")
 py_library(
     name = "pkg",
     srcs = glob(["**/*.py"]),
+    # Remove files and directories with spaces in their names as those are invalid in Bazel.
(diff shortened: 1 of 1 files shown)


andyscott:
Is there a good test or example exhibiting the incorrect behavior? Additionally, can we add some comments so that later folks don't accidentally revert this change if we merge it?

gergelyfabian:
I think before adding tests for this one, maybe there would be a better idea to achieve the same purpose.

I've created a patch for rules_python_external that would achive the same purpose, but it's more configurable:

https://github.com/gergelyfabian/rules_python_external/compare/extra_pip_args...gergelyfabian:pip_data_exclude

This adds an additional argument that can be used from the user's WORKSPACE to define what additional files to exclude. Then the pattern for excluding dirs with spaces can be added by the user (I'm doing that from my workspace file).

If you think, it's a better approach, then maybe I could implement it here too.

PR #320, opened 2020-05-29
Fix for when there are so many file arguments it creates the Command To Long error

opbro:
When there are to many file arguments to py_wheel from py_package on Windows 10 I was getting 
this error related to wheelmaker.exe 
```
command is longer than CreateProcessW's limit (32768 characters)
```
Added additional rules to get around this issue.

### DIFF
--- a/experimental/python/wheel.bzl
+++ b/experimental/python/wheel.bzl
@@ -81,37 +81,6 @@ Sub-packages are automatically included.
     },
 )
 
-def _py_wrapped_input_files_impl(ctx):
-    inputs_to_package  = depset(
-        direct = ctx.files.deps
-    )
-    packageinputfile = ctx.actions.declare_file(ctx.attr.name + '_target_wrapped_inputs.txt')
-    content = ''
-    for input_file in inputs_to_package.to_list():
-        content += _input_file_to_arg(input_file) + '\n'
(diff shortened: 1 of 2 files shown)


opbro:
@googlebot I signed it!

experimental/python/wheel.bzl
```diff
@@ -81,6 +81,37 @@ Sub-packages are automatically included.
...
+    
+    return [
+        DefaultInfo(
+            files = depset(direct = [packageinputfile])
+        )
+    ]
+
+
+
+py_wrapped_input_files = rule(
```
pstradomski, line 102: Would it make sense to always create a file with list of files to packaging instead of asking users to make this decision?

opbro: Yeh I was playing with that before. Let me take another stab at how that feels.

pstradomski:
andyscott, thundergolfer: For some reason github would let me merge this PR; I'm not sure why though, as I'm not listed in CODEOWNERS. As this looks like unintentional misconfiguration, I'm defering to your decision whether to merge or not. The change looks good to me.

opbro:
@thundergolfer  or @andyscott  have any thoughts on this change?

Issue #301, opened 2020-05-02
Can there be a simpler project example? 


oconnore:
I'm trying to copy+paste from the setup instructions to get a working hello world app, and this isn't working for me. 

How does `internal_deps.bzl` relate to the examples in `/examples/hello_world/...`? 

https://github.com/bazelbuild/rules_python/blob/a0fbf98d4e3a232144df4d0d80b577c7a693b570/internal_deps.bzl#L33-L50

Also, where does requirements.bzl come from when there's no `requirements.bzl` file in the whole project? 

https://github.com/bazelbuild/rules_python/blob/a0fbf98d4e3a232144df4d0d80b577c7a693b570/examples/helloworld/BUILD#L15

https://github.com/bazelbuild/rules_python/search?q=requirements.bzl&unscoped_q=requirements.bzl

oconnore:
I was able to get this working, but I'll leave this open since it was rather still rather painful. Here's my hello world:

Project: [hello_world.zip](https://github.com/bazelbuild/rules_python/files/4566968/hello_world.zip)

Workspace:
```
# SNIP *********

load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
http_archive(
    name = "rules_python",
    url = "https://github.com/bazelbuild/rules_python/releases/download/0.0.1/rules_python-0.0.1.tar.gz",
    sha256 = "aa96a691d3a8177f3215b14b0edc9641787abaaa30363a080165d06ab65e1161",
)
load("@rules_python//python:repositories.bzl", "py_repositories")
py_repositories()
load("@rules_python//python:pip.bzl", "pip_repositories")
pip_repositories()

# Define Deps
load("@rules_python//python:pip.bzl", "pip_import")
pip_import(
   name = "analytics_dependencies",
   requirements = "//sql_monitor:requirements.txt",
)
# Load Analytics Deps
load("@analytics_dependencies//:requirements.bzl", "pip_install")
pip_install()
```

thundergolfer:
Thanks for the feedback. I haven't tried the example myself, but you'd hope running it was a simple matter of copying the folder out of the repo and running it. I'll check it out.

oconnore:
If Bazel just searches for the next closest workspace file up the file tree, then the examples could perhaps stay in this repository as independent projects from the rule implementation?

thundergolfer:
[...]

PR #326, opened 2020-06-14
Updated 'helloworld' example

thundergolfer:
Update is motivated by https://github.com/bazelbuild/rules_python/issues/301

This change makes the 'helloworld' example much more copy-pasteable by beginners, by decoupling the code from project's `WORKSPACE`. I think it can be harder to understand this example given it's coupling to the main project, like with load statements like this: `load("//python:defs.bzl", "py_library", "py_test")`

I think I would prefer it if all the examples in the `examples/` folder were managed by one `WORKSPACE`, instead of a `WORKSPACE` per example, but I'll leave it as is for now and wait for comment. _Edit:_ @alexeagle has [good rebuttal to this below.](https://github.com/bazelbuild/rules_python/pull/326#issuecomment-655527483)

----

cc @oconnore

### DIFF
--- a/.bazelignore
+++ b/.bazelignore
@@ -1,0 +1,1 @@
+examples/helloworld
--- a/examples/helloworld/WORKSPACE
+++ b/examples/helloworld/WORKSPACE
@@ -1,7 +1,5 @@
 workspace(name = "examples_helloworld")
 
-load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
-
 # If copying this WORKSPACE file to start a new Bazel workspace, replace this with a
 # `http_archive` repository rule as specified here: https://github.com/bazelbuild/rules_python/releases
 local_repository(
@@ -14,11 +12,9 @@ load("@rules_python//python:repositories.bzl", "py_repositories")
 py_repositories()
 
 # Only needed if using the packaging rules.
-load("@rules_python//python:pip.bzl", "pip_repositories")
+load("@rules_python//python:pip.bzl", "pip_repositories", "pip_import")
 
 pip_repositories()
-
-load("@rules_python//python:pip.bzl", "pip_import")
 
 pip_import(
     name = "helloworld_pypi_deps",
--- a/examples/helloworld/WORKSPACE
+++ b/examples/helloworld/WORKSPACE
@@ -1,0 +1,31 @@
+workspace(name = "examples_helloworld")
+
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+# If copying this WORKSPACE file to start a new Bazel workspace, replace this with a
+# `http_archive` repository rule as specified here: https://github.com/bazelbuild/rules_python/releases
+local_repository(
+    name = "rules_python",
+    path = "../../",
+)
+
+load("@rules_python//python:repositories.bzl", "py_repositories")
+
+py_repositories()
+
+# Only needed if using the packaging rules.
+load("@rules_python//python:pip.bzl", "pip_repositories")
+
+pip_repositories()
+
+load("@rules_python//python:pip.bzl", "pip_import")
+
+pip_import(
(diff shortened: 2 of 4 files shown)


andyscott:
This seems reasonable to me as long as we update CI to check that the nested workspace builds successfully. I'll see if I can convince buildkit to make that happen. If not, we might need to use some shell scripts.

alexeagle:
Even better than customized CI, let's just make a bazel integration test that loads and tests the nested workspace using bazel-in-bazel. I'll contribute the bits we use from rules_nodejs.

I think one WORKSPACE per example is best - in some cases an example will require additional WORKSPACE content (like protobuf for example) and it's a lot easier to understand if that content is self-contained in the example directory. Also it means the labels in the example will look like the users project, not needing an extra `example_name/` segment at the beginning.

alexeagle:
After looking through these "examples" though, I think there's just one example there which is how to run pip. There's no example of a py_binary, nor a genrule/custom rule that executes a python action. I think if users come to this examples directory they are looking for something that matches the shape of their repo.

My rough proposal is to
- maybe drop examples/version because I don't see what that adds - in examples/boto we also assert the version of a pip-installed package
- put the remaining three directories together into examples/pip
- create issues for other scenarios we need examples for, maybe from combing issues or looking at our own company repos for shapes that matter

thundergolfer:
> I'll contribute the bits we use from rules_nodejs.

Can you point me at this. I haven't done 'bazel-in-bazel' before. 

> maybe drop examples/version because I don't see what that adds 

Looking at [`version_test.py`] is looks vaguely like an integration test and I faded memories of seeing code comments saying `rules_python` needs to validate the `pip` version it uses, but this code is super legacy and probably no longer useful.

> create issues for other scenarios we need examples for,

A useful one is showcasing a basic Python+Protobuf setup, using `rules_proto`'s `py_proto_library` until this project maintains its own like [`rules_go`](https://github.com/bazelbuild/rules_go/blob/master/proto/core.rst#go_proto_library).

alexeagle:
Our integration test setup has a lot of moving pieces, and we might want to upstream it first. I think this is something Greg or I should contribute

alexeagle:
okay #337 has the nested workspace. #338 has that same commit, plus a `bazel_integration_test` rule that tests the nested WORKSPACE.
It's roughly copied from rules_nodejs, we can explain the design considerations more.

.bazelignore
```diff
@@ -0,0 +1 @@
+examples/helloworld
```
alexeagle, line 1: of course if you bazelignore the nested folder, you can't use it as sources for a test.
In my #337 I did what we do in rules_nodejs which is tell bazel that `examples/helloworld/BUILD` is just a regular file and does not indicate a Bazel package.

examples/helloworld/WORKSPACE
```diff
@@ -0,0 +1,27 @@
...
+# If copying this WORKSPACE file to start a new Bazel workspace, replace this with a
+# `http_archive` repository rule as specified here: https://github.com/bazelbuild/rules_python/releases
+local_repository(
+    name = "rules_python",
+    path = "../../",
+)
+
+load("@rules_python//python:repositories.bzl", "py_repositories")
+
+py_repositories()
```
alexeagle, line 12: this doesn't do anything, and I think if rules_python were to grow dependencies it would be horribly breaking because WORKSPACE deps semantics are so bad, users will have to re-order their WORKSPACE if they can get it to work at all. WDYT about removing this notion that rules_python always makes you install additional runtime bazel deps? we should get what we need from python packages on pypi.

thundergolfer: Yep, I've seen your bazelcon talk and I'm onboard 😅

.bazelignore
thundergolfer: yeh that's better. will close in favour of #337

thundergolfer:
Closing in favour of #337

PR #333, opened 2020-06-27
Use internal pip unzip_file definition to ensure executable bit is retained when unpacking

gmweaver:
`zipfile.ZipFile` does not retain file permissions, which leads to the loss of the executable bit on files. This impacts the installation of any Python packages that rely on executable scripts (i.e. `pyspark`). The `unzip_file` function used by pip corrects this issue by adding the executable bit back:

https://github.com/pypa/pip/blob/ebd331fb45f4994bc075a799e38252a0b77bd566/src/pip/_internal/utils/unpacking.py#L107

This request replaces the call to `zipfile.ZipFile` with a call to a slightly modified `unzip_file` function. The minor modifications to the original source code from unpacking.py remove any functions/classes that are not necessary for core functionality and/or error checking. Alternatively, pip could be added as a dependency and the function could be imported directly.


gmweaver:
@googlebot I signed it!

gmweaver:
@thundergolfer @andyscott any thoughts on the proposed changes? As an alternative, I also considered a modified version of `_extract_member()` in `zipfile` that adds back the executable after copying, but thought mirroring `pip` might be better.

thundergolfer:
@gmweaver Check out https://github.com/dillon-giacoppo/rules_python_external/pull/46/files for a leaner implementation. 

Also we can't accept the compiled `.par` files in a pull request for security reasons. The process is that we compile the tools after accepting the PR. 🙂

gmweaver:
@thundergolfer thanks for the reference! Definitely a much cleaner solution. Updated to lean implementation and removed changes to compiled `.par` files.

rotemtam:
Perfect, looking forward to this!

thundergolfer:
@googlebot I consent.

PR #337, opened 2020-07-13
feat(examples): move examples to a nested WORKSPACE

alexeagle:
This lets users understand the example in isolation. They can copy/paste the example directory
and it works correctly.

This refactors the existing examples which are quite weak, only really demonstrating pip usage.
This makes room for examples demonstrating other features (like protocol buffers) or package
managers (like poetry).

In a later commit I'll add bazel-integration-testing so we get a test target that confirms
the examples build (including their WORKSPACE being self-contained)


alexeagle:
@thundergolfer ping! I've got some more examples queued up after this lands

.bazelrc
```diff
@@ -0,0 +1,8 @@
+# For bazel-in-bazel testing
```
thundergolfer, line 1: Ahh, so that's how this works. Clever.

examples/pip/extras/BUILD
```diff
@@ -12,8 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-load("@examples_extras//:requirements.bzl", "requirement")
-load("//python:defs.bzl", "py_test")
+load("@extras_deps//:requirements.bzl", "requirement")
+load("@rules_python//python:defs.bzl", "py_test")
 
 package(default_visibility = ["//visibility:public"])
```
thundergolfer, line 18: Not directly related to your change, but this looks to be unnecessary, and I'm not sure we want to leave it here in case it leads to newbies copy-pasting it around their BUILD files (I did this in my first Bazel migration).

alexeagle: good callout, agreed we should train users to default to `["//:__subpackages__"]` or more specific where possible.

PR #338, opened 2020-07-13
bazel_integration_test to test examples/ nested workspaces

alexeagle:


examples/pip/helloworld/helloworld_test.py
```diff
@@ -12,10 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import helloworld
```
alexeagle, line 15: this is surprising - it's just one directory closer to WORKSPACE now so I'd expect to remove one segment here but this effectively removes two - anyone understand why? `from helloworld import helloworld` does not work.

alexeagle:
Wow that's interesting, it passes on Ubuntu 18 but on Ubuntu 16 we get
```

File "/var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/ec321eb2cc2d0f8f91b676b6d4c66c29/sandbox/linux-sandbox/17/execroot/rules_python/_tmp/f84da081940674fccb1ee3be5e69fd27/_bazel_buildkite-agent/28a56a7da3adfa85df487e8fe49aab88/sandbox/linux-sandbox/3/execroot/pip_example/bazel-out/k8-fastbuild/bin/boto/boto_test.runfiles/boto_deps_pypi__futures_3_3_0/concurrent/futures/_base.py", line 414
--
  | raise exception_type, self._exception, self._traceback
  | ^
  | SyntaxError: invalid syntax


```
@thundergolfer any experience with a difference between python interpreter or stdlibs on these images?

alexeagle:
Greened this up by excluding it from Ubuntu16 for now

examples/pip/helloworld/helloworld_test.py
thundergolfer: This arrangement of modules and packages doesn't work in Bazel, and I think it wouldn't even work sometimes outside of Bazel, depending on how the test file is run.

Python is adding the directory of the test file to `sys.path[0]`. So this is the first entry in `sys.path`: `'/Users/jbelotti/work/rules_python/examples/pip/helloworld'`.

Now because you've correctly listed `deps = [":helloworld"],`, the workspace root is correctly listed in the `sys.path` but for me it's entry 6 or 7. 

So when you do `from helloworld import helloworld`, Python scans the `sys.path` and immediately finds a `'helloworld'` module available under the first entry, `...example/pip/helloworld`. So it imports `helloworld.py` and breaks because the module does not contain any `helloworld` symbol. 

When you do `import helloworld` Python does the same thing but this time the code is compatible with importing the module not the parent package.

AFAIK even though Bazel [is removing the `sys.path[0]` entry in the stub file](https://github.com/bazelbuild/bazel/blob/1a0852199d05bef95fcceaa1ef244c36d393bfb0/src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt#L19), it can't remove the one that Python puts there when the  'second-stage' interpreter runs the target module (our `helloworld_test.py`).

This problem doesn't occur when the package has a different name to the modules within. If I try `from helloworld import helloworld` inside the `boto/` folder it works. 

I think given this we shouldn't use this arrangement. It's a gotcha.

thundergolfer:
@alexeagle I can repro that `SyntaxError` on my laptop. It's because Python2 is used to install `futures` but the test gets executed by Python3. 

Without looking at the source code, I suspect that Python3 is favoured if it's present where Bazel expects it. By adding `python_version = "PY2",` to the boto test that error will be fixed and you can remove the tag. 👍

examples/pip/helloworld/helloworld_test.py
alexeagle: yes that makes sense thanks. I was looking at the sys.path but not thinking enough about the ordering.

I don't quite follow your suggestion though. What do you think should change? There's no __init__.py in the example so there isn't a package from which I could `from . import helloworld`. Are you saying the directory name should change? helloworld.py should change?

alexeagle: Okay so I think this is just me learning python and nothing to do with Bazel.

https://docs.python.org/3/library/sys.html#sys.path is clear that the directory containing the entry point main script is the zeroth element on the sys.path. This is true outside of Bazel and correctly behaves the same inside Bazel. So this example shows a valid test setup and `import helloworld` is the right way to hold it.

IMO our examples should prove that any python code that works outside Bazel also works under Bazel, so whether we recommend a pattern or not, it's a valid example.

alexeagle:
Green without any exclusions now by pinning the python version.

PR #339, opened 2020-07-14
Remove mention and usage of Bazel Federation

alexeagle:
It's currently a stalled project so it's not useful for us to direct new users there in our README.
Separately it is harder to develop on rules_python since it is currently not self-contained.
For example it's hard to find or adjust the version of rules_pkg without looking/editing in the federation repo.

Tony says this is an okay change: https://github.com/bazelbuild/bazel-federation/commit/63f9746afd4e8564d1033c529ce5065f69d6c7f9#commitcomment-40577834

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,32 +50,7 @@ contribute](CONTRIBUTING.md) page for information on our development workflow.
 ## Getting started
 
 To import rules_python in your project, you first need to add it to your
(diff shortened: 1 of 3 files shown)


thundergolfer:
Thanks for following up on this. I reached out to them on Slack about this but never sent the follow-up email as they requested.

PR #344, opened 2020-07-19
Add missing windows CI

### LINKED ISSUE #2: Windows support
We should get `rules_python` to a state where Bazel can enable:
```json
                {"node": "windows-x86_64"}
```

... in `.ci/rules_python.json`.

Some early scouting of this on top of my `pip_import` PR reveals that there are [at least some blocking issues](http://ci.bazel.io/blue/organizations/jenkins/PR%2Frules_python/detail/rules_python/26/pipeline/25) (I believe) unrelated to my changes.

alexeagle:
Fixes #2 (if we can get it green)


thundergolfer:
Failing due to https://github.com/google/subpar/issues/39#issuecomment-391621235

thundergolfer:
There's a whole lot of Windows specific logic in https://github.com/bazelbuild/bazel/blob/1a0852199d05bef95fcceaa1ef244c36d393bfb0/src/main/java/com/google/devtools/build/lib/bazel/rules/python/BazelPythonSemantics.java that ends up creating the `<target_name>.exe` that is in this case `compiler.exe`. 

Somewhere in there a non-UTF-8 char is being introduced, or the whole file is actually just binary and not a text-based bash file like in Unix. 

I'd say I'd be better to address this issue be replacing the `.par` files, like we did in `rules_python_external`.

alexeagle:
Agreed that the best path forward here is to upstream rules_python_external approach.

UebelAndre:
@alexeagle this could probably be closed now that https://github.com/bazelbuild/rules_python/pull/585 has been merged

PR #349, opened 2020-08-20
wheel: support for 'plugin' type entry_points

dhalperi:
## PR Checklist

- [X] Does not include precompiled binaries, eg. `.par` files. See [CONTRIBUTING.md](/CONTRIBUTING.md) for info

## PR Type

Python wheels have a rich concept called `entry_points`. The most common use of these is the existing `console_scripts` support: command-line wrappers for Python code.

However, there are many more uses. Another common one is `plugins` -- the Python version of, say, Java `@AutoService`. This is how `pytest` finds modules on the `PYTHONPATH`.

This PR is something we've been using for a while to add our own plugins to the `entry_points` file; but the change is worth discussing.

In one sense, this is stupid. There's some interference between `console_scripts` and `plugins` -- in reality, the underlying structure is:

```
[group name, like console_scripts]
key1 = value1
key2 = value2
...

[g2]
key3 = v3
key4 = v4
...

...
```

So maybe this would be best as a single argument merging `console_scripts` in: `{group: {k1: v1, k2: v2, ...}, ...}`... if we wanted to make a breaking change.

But since `console_scripts` already existed, I added `plugins` separately. (I did not call it `entry_points` since it doesn't handle `console_scripts`).

(There are of course, other backwards-compat options like adding an `entry_points` arg that throws if `console_scripts` is present.)

Issue Number: not yet, but happy to file one if needed

## What is the new behavior?

Wheels built by `rules_python` support `entry_points` other than `console_scripts`, hopefully without breaking existing `console_scripts` support.

- [X] No - At least, I tried not to.

## Other information

See https://packaging.python.org/specifications/entry-points/ and specifically https://packaging.python.org/guides/creating-and-discovering-plugins/#using-package-metadata

Sorry for not ticking all the boxes from jump - this has been sitting around a while and when I realized PRs are being reviewed again I thought it might be worth starting the discussion.

---
This change is [Reviewable](https://reviewable.io/reviews/bazelbuild/rules_python/349)


experimental/python/wheel.bzl
```diff
@@ -209,6 +212,11 @@ _entrypoint_attrs = {
     "console_scripts": attr.string_dict(
         doc = """\
 console_script entry points, e.g. 'experimental.examples.wheel.main:main'.
+""",
+    ),
+    "plugins": attr.string_dict(
```
pstradomski, line 217: That should be string_list_dict to allow more than one entry point in single group, right?

experimental/rules_python/wheelmaker.py
```diff
@@ -157,12 +158,26 @@ def add_metadata(self, extra_headers, description, classifiers, requires,
         metadata += "\n"
         self.add_string(self.distinfo_path('METADATA'), metadata)
 
-    def add_entry_points(self, console_scripts):
+    def add_entry_points(self, console_scripts, plugins):
```
pstradomski, line 161: I'd change it so that the .bzl file just merges console_scripts and plugins, and wheelmaker just always sees only entry_points.

experimental/python/wheel.bzl
pstradomski, line 217: I think entry_points would be a better name.

experimental/python/wheel.bzl
```diff
@@ -209,6 +212,11 @@ _entrypoint_attrs = {
     "console_scripts": attr.string_dict(
```
pstradomski, line 212: The .bzl rule can just merge console_scripts into entry_points/plugins (i.e. treat entries in "console_scripts" as if they were entries in "entry_points" in the "console_script" group). I guess the question is whether console_scripts should be marked as deprecated or do we want to live with this merging forever. I'm fine with marking as deprecated.

experimental/rules_python/wheelmaker.py
```diff
@@ -157,12 +158,26 @@ def add_metadata(self, extra_headers, description, classifiers, requires,
...
+        pattern = re.compile(r'([^=:]+):([^:=]+)=([^=]+)')
+        lines = []
+        if console_scripts:
+            lines += ["[console_scripts]"] + console_scripts
+        if plugins:
+            for line in plugins:
+                match = pattern.fullmatch(line)
+                if not match:
+                    raise ValueError('{line} is not a valid entry point'.format(line=line))
+                plugin_type = match.group(1).strip()
```
pstradomski, line 174: This can produce multiple groups with the same name. Even if that's allowed by the spec (I'm not sure?) it would be better to order the entries so that each group appears only once.

pstradomski (changes requested):

experimental/python/wheel.bzl
dhalperi: Implemented.

I documented `console_scripts` as deprecated based on how other things in this project are. I didn't see a specific Starlark flag to make an attr as deprecated.

experimental/python/wheel.bzl
dhalperi: Yep.

dhalperi: Done.

experimental/rules_python/wheelmaker.py
dhalperi: Done.

experimental/rules_python/wheelmaker.py
dhalperi: This did not happen before (at least from `wheel.bzl`) because there was only one entry allowed per group. As part of switching that to allow multiple entry points per group, rewrite this code from scratch. I believe all such concerns are now handled.

experimental/rules_python/wheelmaker.py
```diff
@@ -157,13 +158,38 @@ def add_metadata(self, extra_headers, description, classifiers, requires,
         metadata += "\n"
         self.add_string(self.distinfo_path('METADATA'), metadata)
 
-    def add_entry_points(self, console_scripts):
+    def add_entry_points(self, entry_points):
         """Write entry_points.txt file to the distribution."""
         # https://packaging.python.org/specifications/entry-points/
-        if not console_scripts:
+        import re
+        pattern = re.compile(r'([^;]+);([^=]+)=(.*)') # group;name=ref
```
dhalperi, line 165: FYI, an earlier draft of this code required `ref` to be in the format `a:b`. Turns out that's only one of the allowed formats for an object reference, so I dropped it.

experimental/rules_python/wheelmaker.py
```diff
@@ -157,13 +158,38 @@ def add_metadata(self, extra_headers, description, classifiers, requires,
...
+        for line in entry_points:
+            match = pattern.fullmatch(line)
+            if not match:
+                raise ValueError('{line} is not a valid entry point'.format(line=line))
+            group = match.group(1).strip()
+            group_dict = groups.setdefault(group, {})
+            name = match.group(2).strip()
+            object_reference = match.group(3).strip()
+            if name in group_dict:
+                raise ValueError("Duplicate entry for name {name} in group {group}".format(name=name, group=group))
```
dhalperi, line 180: Is there a stock formatter for this repo? I tried `yapf` but it changed way more than just the modified code.

experimental/python/wheel.bzl
pstradomski: I think the best we can do is print("console_scripts is deprecated, please use entry_points instead") in the skylark rule when it sees non-empty console_scripts.

experimental/python/wheel.bzl
```diff
@@ -140,8 +140,14 @@ def _py_wheel_impl(ctx):
...
 
+    for group, entries in ctx.attr.entry_points.items():
+        for e in entries:
+            # Assemble a string that can be passed as a CLI argument and parsed unambiguously.
+            args.add("--entry_point", group + ";" + e)
+
+    # Merge console_scripts into entry_points
     for name, ref in ctx.attr.console_scripts.items():
-        args.add("--console_script", name + " = " + ref)
+        args.add("--entry_point", "console_scripts;{name} = {ref}".format(name = name, ref = ref))
```
pstradomski, line 150: Perhaps it would be better to generate entry_points.txt in .bzl (like _target_wrapped_inputs.txt does above) and just package it in wheelmaker.py.

I leave it up to you to decide which approach you prefer.

pstradomski (changes requested):
@andyscott @thundergolfer : This change looks good to me (except for the comments I've left). Should I merge it?

I authored the initial implementation of py_wheel for this repo and kept maintianing it, but with you as new maintainers I'd like a clarification if I should continue merging changes here or would you prefer to review/merge all new PRs?

experimental/python/wheel.bzl
dhalperi: Done.

experimental/python/wheel.bzl
dhalperi: Rewrote entirely. The ugliness moved, but is not gone. The wheelmaker.py file is much cleaner though.

experimental/python/wheel.bzl
dhalperi: Hmm, this `print` statement fails `buildifier`. Is there an alternative?

pstradomski (changes requested):
Presubmit complains about the print() function.

Let's remove it.

pstradomski (approved):
Looks good to me

PR #359, opened 2020-09-20
Promote rules_python_external to public API; deprecate pip_import

alexeagle:
Follows up on next steps from https://github.com/bazelbuild/rules_python/pull/354
See individual commits for more detail (this should be rebased and merged, not squashed)

Reminder of the plan:

- ~~bring in the repo to a subdirectory, preserving git history and contributors~~
- ~~clean up the code, remove mypy, get it functional in that subdirectory~~
- ~~add an example/integration test proving that it works~~
- ~~cut a patch release and get some users to test it out~~
- communicate our deprecation plan to rules_python users
- **THIS PR** -> BIG SWITCH: move the rules_python_external implementation to the root, make a legacy/ directory for current impl
- fill in any missing docstrings and include new rules in stardoc
- cut a breaking release (a minor, since semver starting with 0. states that minors are breaking)
- help users switch over
- in a later breaking release, actually remove the old implementation


thundergolfer (commented):
Looks good. I'll give it a run through in our codebase.

examples/legacy_pip_import/WORKSPACE
```diff
@@ -8,7 +8,7 @@ http_archive(
     sha256 = "e46612e9bb0dae8745de6a0643be69e8665a03f63163ac6610c210e80d14c3e4",
```
thundergolfer, line 8: The version used here would have to be updated at the same time as your change below, otherwise the example is broken.

python/pip.bzl
```diff
@@ -13,161 +13,51 @@
...
+    ```
+
+    Args:
+      requirements: A 'requirements.txt' pip requirements file.
+      name: A unique name for the created external repository (default 'pip').
+      **kwargs: Keyword arguments passed directly to the `pip_repository` repository rule.
     """
-    pip_import(python_interpreter = "python3", **kwargs)
-
-def pip_repositories():
```
thundergolfer, line 141: Doing this is unfortunately not backwards compatible with `rules_docker`: https://github.com/bazelbuild/rules_docker/blob/master/repositories/py_repositories.bzl#L34

We can leave it in and/or submit a change to `rules_docker` to remove what's currently just a "placeholder"

thundergolfer: Example of the failure: 

```
ERROR: /private/var/tmp/_bazel_jbelotti/a8a1f81a1422a72c79c6c81641a8c816/external/io_bazel_rules_docker/repositories/py_repositories.bzl:22:1: file '@rules_python//python:pip.bzl' does not contain symbol 'pip_repositories'
ERROR: error loading package '': in /private/var/tmp/_bazel_jbelotti/a8a1f81a1422a72c79c6c81641a8c816/external/io_bazel_rules_docker/repositories/deps.bzl: Extension file 'repositories/py_repositories.bzl' has errors
```

alexeagle: ah yeah, that one always gets us, I'll add back the function as

```
def pip_repositories():
    # buildifier: disable=print
    print("DEPRECATED: the pip_repositories rule has been replaced with pip_install, please see release notes")
```

examples/legacy_pip_import/WORKSPACE
alexeagle: I think the example at head will be broken between this landing and the next release going out, yes. It's too hard to make it compatible such that you can run this example either against a release or against our sources at HEAD, during a breaking change like this. If we want to be really careful about it we should land this on a 0.1.x feature branch instead of master, then after cutting the release we change the branches back. Up to you if it's worth it.

thundergolfer: Probably not worth it.

python/pip.bzl
thundergolfer: I should've gone further down with my comment, as right after you hit a `pip_import` usage which has the same problem: https://github.com/bazelbuild/rules_docker/blob/master/repositories/py_repositories.bzl#L36

The implication of this is that upgrading to `0.1.*` will make `rules_python` incompatible with any existing repositories that a `WORKSPACE` pulls and that use `pip_import`. Those repos would need to get a PR and then ship a release that makes itself compatible with `0.1.*`. Some of those repos would include: 

* https://github.com/etsy/rules_grafana/blob/ff452327df3dcec541e861bb0fe31114c4294bec/grafana/workspace.bzl#L2
* https://github.com/graknlabs/bazel-distribution/blob/b6ae4297ab66484107f9469d88a4e5d047b36c53/pip/deps.bzl
* https://github.com/stackb/rules_proto/blob/1d6b84118399828511faeecc145d399c1e7bdee2/WORKSPACE
* `rules_docker` (already noted here)

alexeagle: those `pip_import` rules will also print the warning, and yeah they'll need an update to get past the breaking change. We can send those PRs out - is that your proposal? Obviously end-users need to wait until all their deps are compatible before they can update (we've had the same thing with rules_docker in some breaking changes for rules_nodejs)

OR are you suggesting we do something more to adapt the existing API for these users to make it less breaking?

thundergolfer: > We can send those PRs out - is that your proposal? 

Yeah I think we'd be better off being proactive about it.

thundergolfer: >  are you suggesting we do something more to adapt the existing API for these users to make it less breaking?

I thought about it. These Bazel repo rules are awkward to version as you can't declare your repo rule deps, things just break at sync/build time.

thundergolfer (approved):

PR #362, opened 2020-09-27
Make workspace dependency hostname configurable

hrfuller:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes
Changes to experimental rules_python_external upstreamed from rules_python_external https://github.com/dillon-giacoppo/rules_python_external/pull/58

Issue Number: N/A

## What is the new behavior?
Users can pass in the hostname of a mirror of files.pythonhosted.org which is used to fetch rule dependencies.

- [ ] Yes
- [x] No


hrfuller:
Hey @alexeagle could you ptal? This is upstreamed from the rules_python_external pr you discussed interest in.

experimental/rules_python_external/repositories.bzl
```diff
@@ -6,22 +6,21 @@ load("@bazel_tools//tools/build_defs/repo:utils.bzl", "maybe")
...
     (
         "pypi__setuptools",
-        "https://files.pythonhosted.org/packages/54/28/c45d8b54c1339f9644b87663945e54a8503cfef59cf0f65b3ff5dd17cf64/setuptools-42.0.2-py2.py3-none-any.whl",
+        "https://{hostname}/packages/54/28/c45d8b54c1339f9644b87663945e54a8503cfef59cf0f65b3ff5dd17cf64/setuptools-42.0.2-py2.py3-none-any.whl",
         "c8abd0f3574bc23afd2f6fd2c415ba7d9e097c8a99b845473b0d957ba1e2dac6",
     ),
     (
         "pypi__wheel",
-        "https://files.pythonhosted.org/packages/00/83/b4a77d044e78ad1a45610eb88f745be2fd2c6d658f9798a15e384b7d57c9/wheel-0.33.6-py2.py3-none-any.whl",
+        "https://{hostname}/packages/00/83/b4a77d044e78ad1a45610eb88f745be2fd2c6d658f9798a15e384b7d57c9/wheel-0.33.6-py2.py3-none-any.whl",
```
bshashank, line 23: If we are going to support other places to host the files, both the `https` and the `packages` need to get parametrized. For example in our internal artifactory this file is at: 

https://artifactory.example.com/pypi-remote-cache/00/83/b4a77d044e78ad1a45610eb88f745be2fd2c6d658f9798a15e384b7d57c9/wheel-0.33.6-py2.py3-none-any.whl

And I can see other deployments where the mirror is just plain HTTP instead of HTTPS. Can the whole thing be called `server_prefix` for example? where it defaults to: `server_prefix=https://files.pythonhosted.org/packages`

hrfuller: Sounds reasonable to me. Two thought are:
1.  Is having the packages hosted under a `packages` directory a part of the structure of a PyPI hosting mirror? If it is standard we probably shouldn't allow it to change (In artifactory you could put a `packages` dir under the pypi-remote-cache repo)
2. Using an insecure protocol is generally not a good idea, but as long as we are using rules that verify the hashes of downloaded files it could be okay.

thundergolfer: For **1**, all I've got from Google so far is this in [PEP-0381](https://www.python.org/dev/peps/pep-0381/#special-pages-a-mirror-needs-to-provide)

> A mirror is a subset copy of PyPI, so it provides the same structure by copying it.
> * simple: rest version of the package index
> * packages: packages, stored by Python version, and letters
> * serversig: signatures for the simple pages

That indicates to me that `packages/` is supposed to be a fixed part of the URL. If we're going to support parameterising a prefix we need to know that the suffix is going to be the same. In particular, how is the `00/83/` part determined?

For **2**, yes I'd agree `http` is not a good idea as it's really only the use of the `sha256` verification that's keeping things safe. Even if you use a host which is only accessible on internal network, I think DNS or BGP spoofing still leave you vulnerable.

hrfuller: @thundergolfer I’m in agreement on both those front. Packages shouldn’t be overridden. Assuming we don’t make changes to support that does this PR look good now?

thundergolfer:
@hrfuller Have you seen this? https://github.com/bazelbuild/bazel/pull/12170

It appears relevant to this use-case, and looks like it could land in the next Bazel release.

hrfuller:
I hadn't but that does look useful thanks. Currently not on the leading edge release so that wouldn't solve my issue directly but I'd understand if you think rules_python should defer to that.

thundergolfer:
> Currently not on the leading edge release so that wouldn't solve my issue directly but I'd understand if you think rules_python should defer to that.

Yes I would like to hold off a bit on this and watch that issue for progress. The approach in bazelbuild/bazel#12170 will allow for the required functionality without changes to multiple rulesets including this one.

PR #365, opened 2020-09-30
make pip_repository rule a local repo

hrfuller:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

I ran into an issue where the pip_install rule does not re-run if the underlying http_archive that it is imported from changes. If the API in `@pip//:requirements.bzl` changes, say to create a new helper function, that function won't be importable unless you run `bazel clean --expunge` to force pip_install to re run with the new rules_python source in the http_archive.

Steps to repro:
1. run some some bazel command like `bazel query //...`
2. switch to a commit with a different rules_python version, and a new function in @pip//:requirements.bzl generated by `python.pip_install.extract_wheels.lib:generate_requirements_file_contents`
3. try to load that new symbol in a BUILD file and run `bazel query //...` see some error like:
`file '@pip//:requirements.bzl' does not contain symbol '<new symbol you added>' `
4. run `bazel clean --expunge`
5. See that error go away as the pip_install repo rule is re-run.

While it is surprising that the generated `@pip//:requirements.bzl` is not invalidated when the http_archive changes, it could be a corner case of invalidation that is not covered because that file is not actually a part of the archive. It is generated by the repo_rule that comes from the archive. If that .bzl file is the only thing requested from the repo_rule then the repo_rule may not be considered invalid.

## What is the new behavior?
Run steps 1-3 above and see no error. The pip_install repo rule is re-run if bazel sees the source of its import (the http_archive) change.

- [ ] Yes
- [x] No

## Other information
There may be other side effects of having the pip_install repository rule declared as local=True but this fixes the issue, and doesn't cause any extra analysis time work in the workspace. If there are other concerns with local=True I would appreciate input on how to invalidate the pip_install repo in other ways when the underlying bazel source changes.


hrfuller:
Hey @thundergolfer I can't request you for a review, can you add yourself?

thundergolfer:
The fact that this behaviour is observed is annoying but I _think_ it makes sense given the documentation. 

In https://docs.bazel.build/versions/master/skylark/repository_rules.html#when-is-the-implementation-function-executed

> Finally, for non-local repositories, only a change in the following dependencies might cause a restart:
>  * `.bzl` files needed to define the repository rule.

The `http_archive` pulls  a new set of files, and the relevant file in this case would be `python/pip_install/pip_repository.bzl`. That `.bzl` may not have changes, _but_ the `python.pip_install.extract_wheels` function that it calls out to has changed. This change must be invisible to Bazel because it's not in a `.bzl` file, so the repo rule doesn't re-run. 

Looking over things, I suspect we're getting this problem because `python.pip_install.extract_wheels` is not a hidden attribute of the `repository_rule`.

thundergolfer:
Finally done some more investigation, with WIP code at: https://github.com/bazelbuild/rules_python/tree/jonathon--fix-repo_rule-cache-key-calc-jan2020

The relevant line of the Bazel doc is actually: 

> Content of any file used and referred to by a label (e.g., `//mypkg:label.txt` not `mypkg/label.txt`).

If I add a reference to `bazel.py` as a hidden attr on the repo_rule, like I've done in the above branch, I can make changes to that file which will change `requirements.bzl` and successfully have Bazel recompute the repository. 

```
+        "_lib_bazel": attr.label(
+            executable = False,
+            default = Label("//python/pip_install/extract_wheels/lib:bazel.py"),
+        )
```

So in this situation, the current modularisation of the python code actually hurts us, as changes in nearly all of the modules are invisible to Bazel's repo_rule caching, and making their changes visible would involve adding **`N`** hidden attributes on the rule. 

I tried referencing the filegroup `@rules_python//pip_install/extract_wheels:distribution` but that doesn't work (and the docs don't say that it would).

thundergolfer:
Working on a refactored `extract_wheel` package that shouldn't have this issue: https://github.com/bazelbuild/rules_python/tree/ jonathon--refactor-extract_wheels-folder-into-single-module

hrfuller:
Thanks for the research / fix on that subtlety @thundergolfer. When you have a PR open from your branch ping me for review please!

hrfuller:
Closing as this isn't the solution to the issue. But we do need to get all the source files that the rule depends on into the cache keys of the repo rule.

PR #367, opened 2020-10-03
Add support for relative 'file:' requirements in pip_install

### LINKED ISSUE #366: Support local wheel files in pip_install requirements.txt
<!--🔅🔅🔅🔅🔅🔅🔅🔅🔅🔅🔅🔅🔅🔅🔅🔅🔅🔅🔅🔅🔅🔅🔅🔅🔅🔅🔅🔅🔅🔅🔅

Oh hi there! 😄 

To expedite issue processing please search open and closed issues before submitting a new one.
Existing issues often contain information about workarounds, resolution, or progress updates.

🔅🔅🔅🔅🔅🔅🔅🔅🔅🔅🔅🔅🔅🔅🔅🔅🔅🔅🔅🔅🔅🔅🔅🔅🔅🔅🔅🔅🔅🔅🔅🔅🔅-->


# 🚀 feature request

### Relevant Rules

`pip_install`

### Description

If you have a local .whl file in the workspace, you can add it to the requirements.txt file like so:

```
./path/to/wheel.whl
file:path/to/wheel.whl
```

Presently, this results in an error when attempting fetch the wheels, since pip is not running in the workspace root:

```
  File "<snip>/external/pypi__pip/pip/_internal/cli/base_command.py", line 153, in _main                                                                                                         
    status = self.run(options, args)                                                                                                                                                                                                                                           
  File "<snip>/external/pypi__pip/pip/_internal/commands/wheel.py", line 158, in run                                                                                                             
    resolver.resolve(requirement_set)                                                                                                                                                                                                                                          
  File "<snip>/external/pypi__pip/pip/_internal/legacy_resolve.py", line 201, in resolve                                                                                                         
    self._resolve_one(requirement_set, req)                                                                                                                                                                                                                                    
  File "<snip>/external/pypi__pip/pip/_internal/legacy_resolve.py", line 365, in _resolve_one                                                                                                    
    abstract_dist = self._get_abstract_dist_for(req_to_install)                                                                                                                                                                                                                
  File "<snip>/external/pypi__pip/pip/_internal/legacy_resolve.py", line 313, in _get_abstract_dist_for                                                                                          
    req, self.session, self.finder, self.require_hashes                                                                                                                                                                                                                        
  File "<snip>/external/pypi__pip/pip/_internal/operations/prepare.py", line 194, in prepare_linked_requirement                                                                                  
    progress_bar=self.progress_bar                                                                                                                                                                                                                                             
  File "<snip>/external/pypi__pip/pip/_internal/download.py", line 452, in unpack_url                                                                                                            
    unpack_file_url(link, location, download_dir, hashes=hashes)                                                                                                                                                                                                               
  File "<snip>/external/pypi__pip/pip/_internal/download.py", line 398, in unpack_file_url                                                                                                       
    hashes.check_against_path(link_path)                                                                                                                                                                                                                                       
  File "<snip>/external/pypi__pip/pip/_internal/utils/hashes.py", line 104, in check_against_path                                                                                                
    with open(path, 'rb') as file:                                                                                                                                                                                                                                             
FileNotFoundError: [Errno 2] No such file or directory: '<snip>/<snip>-py3-none-any.whl'                                                                                                                                                                       
  File "/usr/local/lib/python3.7/runpy.py", line 193, in _run_module_as_main                                                                                                                                                                                                   
    "__main__", mod_spec)                                                                                                                                                                                                                                                      
  File "/usr/local/lib/python3.7/runpy.py", line 85, in _run_code                                                                                                                                                                                                              
    exec(code, run_globals)                                                                                                                                                                                                                                                    
  File "<snip>/external/rules_python/python/pip_install/extract_wheels/__main__.py", line 5, in <module>                                                                                         
    main()                                                                                                                                                                                                                                                                     
  File "<snip>/external/rules_python/python/pip_install/extract_wheels/__init__.py", line 87, in main                                                                                            
    subprocess.run(pip_args, check=True)                                                                                                                                                                                                                                       
  File "/usr/local/lib/python3.7/subprocess.py", line 512, in run                                                                                                                                                                                                              
    output=stdout, stderr=stderr)                                                                                                                                                                                                                                              
subprocess.CalledProcessError: Command '['/usr/bin/python3.7', '-m', 'pip', 'wheel', '-r', '<snip>/requirements.txt']' returned non-zero exit status 2. 
```


The `extract_wheels` script could convert relative wheel paths to absolute wheel paths prior to running `pip wheel`. It already has the full path to the requirements.txt file, so would just snip off the end and replace with relative path provided in requirements.txt.


- Using `whl_library`, but this is not recommended and does not load dependencies of the local wheel.
- Using absolute path in requirements.txt. This appears to work, but the path is not portable across different environments.

aaliddell:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: closes #366 

Relative `file:` requirements do not work due to `pip wheel` being run from the generated repository root.

## What is the new behavior?

Relative `file:` URIs in requirements.txt are converted to absolute `file:///` URIs based on the directory containing the requirements.txt file. This allows wheel files within the workspace to take part in dependency resolution, since the alternative `whl_library` rule is both non-recommended and does not attempt to fetch dependencies based on the wheel metadata.

Supports both `file:some/path/to/wheel.whl` and `requirement @ file:some/path/to/wheel.whl`. If a requirement is already absolute it will not be altered. For this to work, the requirements.txt file has to be copied into the generated repository, since we do not want to amend the source file in place.

This does not attempt to support local wheels with paths specified without `file:`, since these are equivalent but the bare requirement is harder to patch. Supporting that version consistently would require either writing a full parser for the requirements.txt format or depending on the internal implementation in `pip._internal`.

- [ ] Yes
- [x] No

## Other information

You may be able to deprecate `whl_library` rule as a result of this, if desired.


aaliddell:
I'm not sure how you'd like to test this, since we'd require having a .whl file (or other supported requirement type) checked into the repo, perhaps in the `examples/pip_install` folder. We could either choose to find a very small existing .whl file or create a tiny one ourselves. However, I can understand hesitation towards adding another binary file, since the .par files have been a pain to review in the past also.

Let me know what you'd like?

pstradomski:
Dnia sobota, 3 października 2020 15:21:29 CEST Adam Liddell pisze:
> I'm not sure how you'd like to test this, since we'd require having a .whl
> file (or other supported requirement type) checked into the repo, perhaps
> in the `examples/pip_install` folder. We could either choose to find a very
> small existing .whl file or create a tiny one ourselves. However, I can
> understand hesitation towards adding another binary file, since the .par
> files have been a pain to review in the past also.
> 
> Let me know what you'd like?

Can you use one of the wheels built in experimental/examples/wheel ?

-- 
Paweł Stradomski

aaliddell:
Unfortunately I don't think that can work easily, since the wheel has to exist at the time the repo rules are run. The wheel files created in that folder won't be present until much later when the execution phase has completed.

Since they are technically separate workspaces, you could get it to work in CI by forcing the test order and copying the wheel at the right point, but it would not be a portable solution that would work for any contributors machine.

LouisStAmour:
As an alternative to adding a binary .whl file, perhaps we could re-use a source folder with a very minimal setup.py file? For example https://github.com/maet3608/minimal-setup-py

I came here with an alternative idea however. Since the problem with relative path resolution is that pip can resolve paths to other requirements.txt files relatively just fine, but that it can't resolve package paths to anything but the current working directly, what if we changed the current working directory?

https://github.com/bazelbuild/rules_python/blob/c639955c18bbc16e6a9e5228540b6151fd3b10b3/python/pip_install/pip_repository.bzl#L53-L61

Just add `working_directory` to the above, and give it the path to the requirements.txt file? Seems simpler than editing and rewriting the requirements.txt file and supports any new syntax that comes out. For bonus points, make the working_directory an optional parameter that can be overridden as necessary. Heck, if it's too complicated to figure out what working directory to use, I'd be very happy with an optional parameter that I could add to override current behaviour and specify a working directory. In my case, I'd like to be able to say it should execute within a subfolder.

Not sure if my suggestion for `working_directory` actually solves the above issue for .whl files as I don't have one of those. But I do want to use code from folder /A with a setup.py file as a module for folder /B yet my requirements.txt is expecting to run within a venv in folder /B from folder /B as the cwd. Works great that way, doesn't work when running inside Bazel due to the different working directory. It resolves `./parent-requirements.txt` from `B/child-requirements.txt` but it can't resolve `../A` from `B/parent-requirements.txt` correctly.

aaliddell:
> I came here with an alternative idea however. Since the problem with relative path resolution is that pip can resolve paths to other requirements.txt files relatively just fine, but that it can't resolve package paths to anything but the current working directly, what if we changed the current working directory?

That may be suitable, I'll have to test. Provided we don't end up messing up where the outputs are written, but I think that's already managed IIRC.

> As an alternative to adding a binary .whl file, perhaps we could re-use a source folder with a very minimal setup.py file? For example https://github.com/maet3608/minimal-setup-py or see A/setup.py in https://github.com/LouisStAmour/rules_python-pip-cwd-bug

Do you mean for the test or in general? For a test, any of the supported requirement types would be suitable, I just picked .whl as that's what I had in mind when working on this. I guess a non-binary file would be more palatable :+1:

aaliddell:
Superceded by #433

PR #370, opened 2020-10-16
Remove py_repositories

alexeagle:
This adds useless setup code to a users WORKSPACE file.

The idea that we could add some transitive WORKSPACE dependencies is flawed without a working support for this in Bazel. Anything we add in that function later is a breaking change for users, in that they might call our py_repositories() before fetching rules_xx, and we already installed an incompatible rules_xx.

Since adding anything here is a breaking change, we can always put it back later as a breaking change to the rules.

However I would argue strongly that rules_python is too core in the dependency chain for it to *ever* grow transitive dependencies. Like rules_nodejs we should either vendor code we need privately or strip development-only dependencies from our distribution, and never suggest that users call a rules_python function that installs other starlark code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,9 +86,6 @@ http_archive(
     strip_prefix = "rules_python-{}".format(rules_python_version),
     url = "https://github.com/bazelbuild/rules_python/archive/{}.zip".format(rules_python_version),
 )
-
-# Then load and call py_repositories() as
-# above.
 ```
 
 Once you've imported the rule set into your `WORKSPACE` using any of these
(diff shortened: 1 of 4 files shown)


thundergolfer (approved):
Agree with your argument. Without the ability to express version compatibility across external repositories, any additional dependency added under `py_repositories()` adds likelihood that end-users experience undocumented breakages. 

Looking at https://github.com/bazelbuild/rules_python/pull/218, the PR that intro'd it over a year ago, there didn't seem to be a great need for it originally, and over time transitive external repo deps have seem to me more fraught with risks. cc @brandjon.

PR #405, opened 2021-01-24
pip: 20.3.3 -> 20.3.4

groodt:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

## What is the new behavior?

Updates the vendored version of pip from 20.3.3 to 20.3.4.

Notably, this is slated to be the last version of pip that will support Python 2.
https://github.com/pypa/pip/issues/8936#issuecomment-762080038

It might be prudent to prepare and tag a release of rules_python if this is merged.

I would also suggest that rules_python, at least from the `pip_install` perspective, should target Python 3 only and mention this in the README files. I understand that the Bazel rules_python, in general, will need to support Python 2 until this requirement is removed from Bazel itself (and perhaps internally at Google), but the `pip_install` elements of the rules need not support Python 2 in my opinion. It is EOL and further developments in Python packaging are not going to support Python 2 (or Python <3.5) so it might be time to start enforcing this too.

Note, there is an additional PR [here](https://github.com/groodt/rules_python/pull/1) with pip 21.0 that can be merged after this one. Which I would then consider a clean breaking change for the Python 2 vs Python 3 support for rules_python `pip_install`. In the same way that pip supports Python 2 with `pip<21` it would be possible to point users at a commit or version of rules_python `pip_install` that does work with Python 2 if they need it. 

I suspect my recommendation is overly cautious, I suspect the number of users that are using rules_python (for external pip dependencies) and Python 2 to be 0 or vanishingly close to 0.

- [ ] Yes
- [x] No

This PR would not be introducing a breaking change, the breaking change is in the follow-up PR that introduces pip 21.0

### DIFF
--- a/python/pip_install/repositories.bzl
+++ b/python/pip_install/repositories.bzl
@@ -6,8 +6,8 @@ load("@bazel_tools//tools/build_defs/repo:utils.bzl", "maybe")
 _RULE_DEPS = [
     (
         "pypi__pip",
-        "https://files.pythonhosted.org/packages/54/eb/4a3642e971f404d69d4f6fa3885559d67562801b99d7592487f1ecc4e017/pip-20.3.3-py2.py3-none-any.whl",
-        "fab098c8a1758295dd9f57413c199f23571e8fde6cc39c22c78c961b4ac6286d",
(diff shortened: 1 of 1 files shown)


hrfuller (approved):
I agree with your suggestion of tagging a new version indicating this is the last version that supports python2 out of the box, but users will always be able to override these deps to use older versions of pip with pip_install if they would like. Not sure how we can force python3 on all our users given they can override the python_interpreter attr in pip_install anyway.

Issue #421, opened 2021-03-01
Update setuptools to current (54.0)


phlax:
# 🚀 feature request (+ ~bug)

### Relevant Rules

Setup tools install

### Description

Currently the setuptools version is 42.x this appears to have problems with building wheels in windows

for ref - see https://github.com/envoyproxy/envoy/pull/15236#issuecomment-788050782 for PR which hit this windows problem

### Describe the solution you'd like

Update to current (54.0) or at least > 51.0 as this seems to fix the problem we have had

I would be happy to raise a PR to fix if there is interest to update

### Describe alternatives you've considered

we are patching rules_python to update setuptools

groodt:
Can you try again from master instead of the release tag? The release tag is ancient.

Long shot, but it might help since this PR did bump things a little:
https://github.com/bazelbuild/rules_python/pull/390
(setuptools 44.1.0)

There is a bit of queue of PRs that tasks that I think we're waiting for:
* Maintainers to cut new release that will be the last release with any semblance of Python 2.7 and 3.5 support for `pip_install` (it was never really supported, but kinda worked sometimes).
* Remove Python 3.5 support from CI https://github.com/bazelbuild/rules_python/pull/417
* Upgrade to pip 21.0.1 (pip dropping Python 2.7 and this adds support for Big Sur and Apple Silicon wheels. The setuptools upgrade could possibly be added here tbh, since setuptools > 45 requires Python > 3.5.

phlax:
> Can you try again from master instead of the release tag? The release tag is ancient.

I thnk master still points to the old setuptools version:

https://github.com/bazelbuild/rules_python/blob/c639955c18bbc16e6a9e5228540b6151fd3b10b3/python/requirements.txt#L2

and here

https://github.com/bazelbuild/rules_python/blob/c7e068d38e2fec1d899e1c150e372f205c220e27/python/pip_install/repositories.bzl#L19

phlax:
ah wait - i see that is 44.0 not 42 - ill try again to see if that fixes the problem i had

phlax:
@groodt seem like the problem remains in setuptools 44.1.0

for ref the failing ci can be seen here https://dev.azure.com/cncf/envoy/_build/results?buildId=67847&view=logs&j=959800d2-504c-5424-ae5b-beaf4b89831b&t=5012f2a4-f1a0-5822-b3e6-7b2c1f819b7c&l=3382

ill open a PR to bump the version...

[...]

PR #417, opened 2021-02-13
Remove Ubuntu 16.04 from CI

groodt:
## PR Checklist

Please check if your PR fulfills the following requirements:

- [ ] Does not include precompiled binaries, eg. `.par` files. See [CONTRIBUTING.md](/CONTRIBUTING.md) for info
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type

What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature (please, look at the "Scope of the project" section in the README.md file)
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] Other... Please describe:

## What is the current behavior?

Issue Number: N/A

## Does this PR introduce a breaking change?

- [ ] Yes
- [ ] No

PR #412, opened 2021-02-02
pip: 20.3.4 -> 21.0.1

groodt:
**Note**: No urgency to rush this. It introduces a breaking change where pip_install will only work for Python3+. If any release of rules_python needs to be done prior to prepare a final Python2 compatible release, it should be done before merging this.

## PR Checklist

Please check if your PR fulfills the following requirements:

- [x] Does not include precompiled binaries, eg. `.par` files. See [CONTRIBUTING.md](/CONTRIBUTING.md) for info
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type

What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature (please, look at the "Scope of the project" section in the README.md file)
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [x] Other... Please describe:

## What is the current behavior?

Issue Number: N/A

## What is the new behavior?

Follow up to https://github.com/bazelbuild/rules_python/pull/405

This PR introduces pip 21.0.1 as the vendored version of pip. This version of pip will only support Python 3+ and is pip breaking away from Python 2 support since pip 21.0: https://pip.pypa.io/en/stable/news/#id1

My suggestion would be to first merge https://github.com/bazelbuild/rules_python/pull/405 and prepare a release or tag along with documentation around the `pip_install` Python 2 support. Then this PR can be rebased against upstream bazelbuild/rules_python and merged to support Python 3 `pip_jnstall` support only for future development.

## Does this PR introduce a breaking change?

- [x] Yes
- [ ] No

## PR Checklist

Please check if your PR fulfills the following requirements:

- [x] Does not include precompiled binaries, eg. `.par` files. See [CONTRIBUTING.md](/CONTRIBUTING.md) for info
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type

What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature (please, look at the "Scope of the project" section in the README.md file)
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [x] Other... Please describe:

## What is the current behavior?

Issue Number: N/A

[...]

Issue #409, opened 2021-01-28
pip_install Python2 support?


gfl-chris:
# 🐞 bug report

### Affected Rule

The issue is caused by the rule: `pip_import`

### Is this a regression?

No. Yes. Maybe.

### Description

I'm trying to invoke the `pip_install` workspace rule with Python2, setting `python_interpreter` to `"python2"`. But `pip_install` fails with the error `/usr/bin/python2: No module named python.pip_install`. I tracked this down to missing `__init__.py` files in the `rules_python/python` and `rules_python/python/pip_install` directories. Python2 must require __init__.py files in the whole module path whereas Python3 doesn't... Anyhow, once the files are there, `python2` fails with the error:
```
  File "python/pip_install/extract_wheels/__init__.py", line 18
    def configure_reproducible_wheels() -> None:
                                        ^
SyntaxError: invalid syntax
```
Which tells me that that's Python3 syntax. :( Is Python2 not supported for `pip_install`?

## 🌍  Your Environment

**Output of `bazel version`:**

<pre>
  <code>
Bazelisk version: development
Build label: 3.7.1
Build target: bazel-out/k8-opt/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Tue Nov 24 17:38:30 2020 (1606239510)
Build timestamp: 1606239510
Build timestamp as int: 1606239510
  </code>
</pre>

**Rules_python version:**

<pre>
  <code>
8537cea11d867333c07c465d536778c1b7560cde
  </code>
</pre>

**python2 --version:**
Python 2.7.18

thundergolfer:
Thanks for the issue report. [rules_python_external](https://github.com/dillon-giacoppo/rules_python_external#prerequisites), which the new packing rules are from, was always only supporting 3.5+. Though this is not clearly communicated in the repo. 

Going forward, Python 2 won't be supported by the packaging rules. Python 2 is officially [sunsetted](https://www.python.org/doc/sunset-python-2/) and no longer supported. Pip has [dropped support for Python 2](https://pip.pypa.io/en/stable/news/#id1) and we want to move forward along with Pip. 

The legacy rules still should work for Python 2, so we should make it much clearer that that's the only option for Python 2.

PR #422, opened 2021-03-02
setuptools: Bump version to 54.0.0

phlax:
## PR Checklist

- [X] Does not include precompiled binaries, eg. `.par` files. See [CONTRIBUTING.md](/CONTRIBUTING.md) for info
- [?] Tests for the changes have been added (for bug fixes / features)
- [?] Docs have been added / updated (for bug fixes / features)

## PR Type

- [X] Bugfix
- [ ] Build related changes
- [ ] CI related changes
- [X] Other... Please describe:

dependency update

rules_python uses old version of setuptools which can break some packages on windows

Issue Number: #421 

## What is the new behavior?

rules_python uses current setuptools stable (54.0) which fixes above problem

- [ ] Yes
- [ ] No

?


python/requirements.txt
```diff
@@ -1,5 +1,5 @@
 pip==9.0.3
-setuptools==54.0.0
+setuptools==44.0.0
```
phlax: not sure why this wont let me set newer setuptools - versions available only seem to go to 44.1

aaliddell: Version 44 was the last to support both Python 2 & 3. So until rules_python fully drops Python 2 support, it may be stuck with this version.

python/pip_install/repositories.bzl
```diff
@@ -16,8 +16,8 @@ _RULE_DEPS = [
     ),
     (
         "pypi__setuptools",
-        "https://files.pythonhosted.org/packages/ab/b5/3679d7c98be5b65fa5522671ef437b792d909cf3908ba54fe9eca5d2a766/setuptools-44.1.0-py2.py3-none-any.whl",
-        "992728077ca19db6598072414fb83e0a284aca1253aaf2e24bb1e55ee6db1a30",
+        "https://files.pythonhosted.org/packages/70/06/849cc805ac6332210083f2114a95b22ee252ce81ed4e1be4f1d2b87c9108/setuptools-54.0.0-py3-none-any.whl",
```
phlax, line 19: @aaliddell 

> Version 44 was the last to support both Python 2 & 3. So until rules_python fully drops Python 2 support, it may be stuck with this version.

does that also preclude this change ?

aaliddell: Perhaps, but I think the new `pip_install` is only intended to support Python 3 (#409). The issue seems to be that there's also the older packaging rules still present, which support Python 2.

phlax: it would be really helpful if this were included from our pov, but my bazel knowledge is a bit limited so im not totally sure of the effect

groodt:
I personally think this will work itself out and be easier to manage once the other train of PRs is merged as mentioned here https://github.com/bazelbuild/rules_python/issues/421#issuecomment-788350959

In summary:
* Maintainers to cut new release that will be the last release with any semblance of Python 2.7 and 3.5 support for pip_install (it was never really supported, but kinda worked sometimes).
* Remove Python 3.5 support from CI #417
* Upgrade to pip 21.0.1 (pip dropping Python 2.7 and this adds support for Big Sur and Apple Silicon wheels. The setuptools upgrade could possibly be added here tbh, since setuptools > 45 requires Python > 3.5. https://github.com/bazelbuild/rules_python/pull/412

This change can't really be done until the above is done. It might also be possible to include this as part of #412, but perhaps also as a fast-follow too, since it is an unrelated changed to #412.

hrfuller (approved):

PR #432, opened 2021-03-08
Incrementally download wheels at workspace time.

hrfuller:
Commits in this PR can be reviewed individually, they build on each other.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

The current pip_install repository rule requires that all pip packages be resolved and downloaded before bazel can start to run build actions. This can lead to painful wait times when invoking pip_install with large requirement.txt files on clean repos, or whenever changes to the external repo configurations are made.

Issue Number: #395 

## What is the new behavior?
Add an experimental repository rule called pip_install_incremental. Which sets up an external repository for each package in the requirements lock file passed by users.

The largest difference from pip_install is that pip_install_incremental doesn't do a full resolve over the requirements_lock file passed in. As such users must supply a requirements_lock.txt file containing fully resolved and pinned transitive closure of the pypi dependencies needed by their workspace. We don't provide a way to resolve a requirements lock file in this PR, nor do we verify the integrity of the lock file, but we could provide something in the future to do this.

- [ ] Yes
- [x] No


python/pip.bzl
```diff
@@ -56,6 +56,17 @@ def pip_install(requirements, name = "pip", **kwargs):
         **kwargs
     )
 
+def pip_install_incremental(requirements_lock, name = "pip_incremental", **kwargs):
```
UebelAndre, line 59: Why is this a separate macro and not just a conditional on the existing `pip_install` macro?

examples/pip_install_incremental/WORKSPACE
```diff
@@ -0,0 +1,40 @@
...
+    # 1. Python interpreter that you compile in the build file (as above in @python_interpreter).
+    # 2. Pre-compiled python interpreter included with http_archive
+    # 3. Wrapper script, like in the autodetecting python toolchain.
+    #python_interpreter_target = "@python_interpreter//:python_bin",
+
+    # (Optional) You can set quiet to False if you want to see pip output.
+    #quiet = False,
+
+    # Uses the default repository name "pip_incremental"
+    requirements_lock = "//:requirements_lock.txt",
```
UebelAndre, line 34: Is this analogous to [pip-tool's](https://pypi.org/project/pip-tools/) `requirements.txt` where a "non-locked" or "floating" requirements file would be `requirements.in`?

python/pip.bzl
hrfuller: That macro itself gives a good example of how to do it.

examples/pip_install_incremental/WORKSPACE
hrfuller: Yeah. I called it something else to not conflict with the existing pip_install rule. And requirements_lock.txt should be somewhat reminiscent of package-lock.json

README.md
```diff
@@ -136,6 +136,41 @@ re-executed in order to pick up a non-hermetic change to your environment (e.g.,
...
+repo which contains a macro called `install_deps()` which is used
+to create child repos for each package in your requirements_lock.txt.
+
+
+```python
+load("@rules_python//python:pip.bzl", "pip_install_incremental")
+
+# Create a central repo that knows about the dependencies needed for
+# requirements.txt.
+pip_install(
```
UebelAndre, line 165: Should this not be `pip_install_incremental`?

hrfuller:
Thanks, probably a copy paste error on my part.

README.md
```diff
@@ -136,6 +136,41 @@ re-executed in order to pick up a non-hermetic change to your environment (e.g.,
...
+repo which contains a macro called `install_deps()` which is used
+to create child repos for each package in your requirements_lock.txt.
+
+
+```python
+load("@rules_python//python:pip.bzl", "pip_install_incremental")
+
+# Create a central repo that knows about the dependencies needed for
+# requirements.txt.
+pip_install_incremental(
```
alexeagle, line 165: maybe we should name this `pip_fetch` or `pip_fetch_deps`

README.md
```diff
@@ -136,6 +136,41 @@ re-executed in order to pick up a non-hermetic change to your environment (e.g.,
 updating your system `python` interpreter), you can completely flush out your
 repo cache with `bazel clean --expunge`.
 
+### Importing `pip` dependencies incrementally (experimental)
```
alexeagle, line 139: "incrementally" doesn't feel like the right word to me, I would have expected sth like "partial download"

README.md
```diff
@@ -136,6 +136,41 @@ re-executed in order to pick up a non-hermetic change to your environment (e.g.,
...
+
+One pain point with `pip_install` is the need to download all dependencies resolved by
+your requirements.txt before the bazel analysis phase can start. For large python monorepos
+this can take a long time, especially on slow connections.
+
+To download only the pip packages needed to build targets in the
+subgraph of top level targets in your bazel invocation, you can experiment with using `pip_install_incremental`.
+The interface of `pip_install_incremental` mirrors `pip_install` as closely as possible.
+
+The only user facing difference between `pip_install` and `pip_install_incremental` is that for the latter you need
```
alexeagle, line 149: not really the only difference, since you also need an extra load() and execute the install_deps function

README.md
```diff
@@ -136,6 +136,41 @@ re-executed in order to pick up a non-hermetic change to your environment (e.g.,
...
+to supply a fully resolved and pinned requirements_lock.txt file (named to distinguish it from requirments.txt
+used in `pip_install`). The `requirements` attribute is replaced with a `requirements_lock` attribute to make it
+clear that a fully pinned transitive resolve is needed.
+
+To add incremental pip dependencies to your `WORKSPACE` load
+the `pip_install_incremental` function, and call it to create a main
```
alexeagle, line 155: this feels a bit dense, I'd suggest making breaking into more sentences/paragraphs. Say what the first rule does, then explain "that rule actually just translates your requirements lockfile into Bazel's starlark language. Then you must load that resulting starlark function and call it"

README.md
```diff
@@ -136,6 +136,41 @@ re-executed in order to pick up a non-hermetic change to your environment (e.g.,
...
+load("@rules_python//python:pip.bzl", "pip_install_incremental")
+
+# Create a central repo that knows about the dependencies needed for
+# requirements.txt.
+pip_install_incremental(
+   name = "my_deps",
+   requirements_lock = "//path/to:requirements_lock.txt",
+)
+
+load("@my_deps//:requirements.bzl", "install_deps")
```
alexeagle, line 170: add another inline doc comment here that continues the story

examples/pip_install_incremental/WORKSPACE
```diff
@@ -0,0 +1,40 @@
+workspace(name = "example_repo")
+
+load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
+
+# TODO(hrfuller) change this to a released http_archive when a release exists.
```
alexeagle, line 5: nit: in rules_nodejs we usually just let the example go ahead of the repo a bit, with the assumption that releases are frequent enough. You don't want all these examples features to require a second PR to patch up what the first one did

The integration test runner always swaps in the current code anyway. It expects the releases/download URL
https://github.com/bazelbuild/rules_python/blob/master/tools/bazel_integration_test/test_runner.py#L26-L29

examples/pip_install_incremental/requirements_lock.txt
```diff
@@ -0,0 +1,11 @@
+# Transitive dependency set from :requirements.txt
```
alexeagle, line 1: nit: maybe give what command you used to generate this

README.md
```diff
@@ -136,6 +136,41 @@ re-executed in order to pick up a non-hermetic change to your environment (e.g.,
...
+the `pip_install_incremental` function, and call it to create a main
+repo which contains a macro called `install_deps()` which is used
+to create child repos for each package in your requirements_lock.txt.
+
+
+```python
+load("@rules_python//python:pip.bzl", "pip_install_incremental")
+
+# Create a central repo that knows about the dependencies needed for
+# requirements.txt.
```
alexeagle, line 164: is there ever a use-case like in rules_go where I'd want to check in that resulting requirements.bzl file so I could hand-edit? like to patch maybe?

python/pip_install/create_incremental_repo/BUILD
```diff
@@ -0,0 +1,21 @@
...
+load("//python/pip_install:repositories.bzl", "requirement")
+
+py_library(
+    name = "lib",
+    srcs = ["__init__.py"],
+    deps = [requirement("pip")]
+)
+
+py_test(
+    name = "test_create_incremental_repo",
```
alexeagle, line 11: if you change the word "incremental" then change here too, like maybe "python/pip_install/translate_lockfile_to_bzl" or sth

python/pip_install/create_incremental_repo/test_create_incremental_repo.py
```diff
@@ -0,0 +1,34 @@
...
+    def test_incremental_requirements_bzl(self) -> None:
+        with NamedTemporaryFile() as requirements_lock:
+            requirement_string = "foo==0.0.0"
+            requirements_lock.write(bytes(requirement_string, encoding="utf-8"))
+            requirements_lock.flush()
+            args = argparse.Namespace()
+            args.requirements_lock = requirements_lock.name
+            args.repo = "pip_incremental"
+            contents = generate_incremental_requirements_contents(args)
+            library_target = sanitised_repo_library_label("foo", create_incremental_repo_prefix(args.repo))
```
alexeagle, line 24: in a test I'd prefer you hardcode expected values rather than calculate them, both to document what it looks like, and as an additional assertion that the sanitised_* helper does what we expect

python/pip_install/extract_wheels/lib/utilities.py
```diff
@@ -0,0 +1,24 @@
+from argparse import ArgumentParser
```
alexeagle, line 1: nit: instead of utilities.py can this be arguments.py since we hope to not grow other random utilities?

README.md
groodt: I've always felt the same and said the same in the original issue: https://github.com/bazelbuild/rules_python/issues/395

`Fetching pip dependencies lazily (experimental)`?

Lazy vs Eager?
Ahead of time vs Just in time?

README.md
groodt: It's also a bit of a shame that `pip` itself has `install` and `fetch` subcommands. :)

In the pip rules, it is actually fetching (or building) wheels through the `pip wheel` command. I don't know how important or confusing that makes things.

README.md
groodt: That's an interesting idea. I think they'd probably be platform dependent potentially.

README.md
hrfuller: I was going off the convention in that issue, but I agree incrementally doesn't feel very explanatory. @alexeagle suggested something more like "parse_pip_{deps|requirements}" (into starlark functions) which makes more sense with the follow up install_deps() call.

examples/pip_install_incremental/WORKSPACE
hrfuller: Ah good to know.

examples/pip_install_incremental/requirements_lock.txt
hrfuller: I used pip-compile which is not a tool we are supporting afaict, but I will add that.

python/pip_install/extract_wheels/lib/utilities.py
hrfuller: 😂  sure!

README.md
hrfuller: Maybe, but to me that would suggest that the resolver you used didn't do a good job applying your requirements / constraints. Let's revisit that if it comes up?

examples/pip_install_incremental/requirements_lock.txt
UebelAndre: What does it mean to "support" a tool? Does it matter how a pinned requirements file is generated? I personally have only ever used pip-compile

hrfuller:
Hey @alexeagle would you take another look when you get a change. I believe I addressed all concerns.

examples/pip_install_incremental/requirements_lock.txt
hrfuller: I meant we don't have any rules/helpers/tooling around it in the repo. I think its fine to suggest it, though it won't work for everyone, which is why I mentioned using a virtualenv and pip freeze in the README.

python/pip_install/extract_wheels/lib/arguments_test.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class ArguementsTestCase(unittest.TestCase):
+    def test_arguments(self) -> None:
+        parser = argparse.ArgumentParser()
+        parser = arguments.parse_common_args(parser)
+        repo_name = "foo"
+        index_url = "--index_url=pypi.org/simple"
+        args_dict = vars(parser.parse_args(
+            args=["--repo", repo_name, "--extra_pip_args={index_url}".format(index_url=index_url)]))
```
skliarpawlo, line 14: Hey, thanks for the feature, we are already trying to use it from the PR directly and it's a game-changer!

**Problem**: we use our own PyPi index in requirements.txt (--index-url=... at the top of the requirements file) and it seems like it is being ignored (I can be wrong or misusing the tool), but if that is the case, it would be great if 
that is passed to extra_pip_args automatically.

Also when we try to pass the pypi url via `pip_extra_args` explicitly, like this:
```
pip_parse(
    name = "my_repo",
    quiet = False,
    timeout = 1500,
    requirements_lock = "@//.../requirements-lock.txt",
    python_interpreter = "python3.7",
    extra_pip_args=["--index-url=http://<pypi-host>/repository/pypi-all/simple", "--trusted-host=<pypi-host>"]
)
```

We get an error, like this:
```
ERROR: .../WORKSPACE:50:13: //external:repo_pypi__setuptools: expected value of type 'list(string)' for attribute 'extra_pip_args' in 'whl_library' rule, but got "{\"args\":[\"--index-url=http://.../repository/pypi-all/simple\"]}" (string)
```

hrfuller: Hey thanks for the feedback, I probably missed the --index-url argument. I'll try to repro the issue with extra_pip_args as well.

hrfuller: I fixed the issue with --extra_pip_args not being passed to the child repos correctly. I will leave the issue of pip flags in a requirements.txt file as a TODO, as this is an experimental rule for the moment.

alexeagle (approved):
Nice!!

examples/pip_parse/WORKSPACE
```diff
@@ -0,0 +1,39 @@
+workspace(name = "example_repo")
+
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+http_archive(
+    name = "rules_python",
+    url = "https://github.com/bazelbuild/rules_python/releases/download/0.1.0/rules_python-0.1.0.tar.gz",
```
thundergolfer, line 7: I don't think this example works because `0.1.0` does not contain `pip_parse`

examples/pip_parse/WORKSPACE
```diff
@@ -0,0 +1,39 @@
...
+    # acts as an executable). The latter can be anything that could be used as Python interpreter. E.g.:
+    # 1. Python interpreter that you compile in the build file (as above in @python_interpreter).
+    # 2. Pre-compiled python interpreter included with http_archive
+    # 3. Wrapper script, like in the autodetecting python toolchain.
+    #python_interpreter_target = "@python_interpreter//:python_bin",
+
+    # (Optional) You can set quiet to False if you want to see pip output.
+    #quiet = False,
+
+    # Uses the default repository name "pip_incremental"
```
thundergolfer, line 32: Default actually `pip_parsed_deps` I think (shown below)

examples/pip_parse/WORKSPACE
hrfuller: Thats what I thought as well, but as @alexeagle pointed out the integration test runner replaces this string with a pointer to a local copy of the up to date repo.

PR #446, opened 2021-04-02
Augment extra args with flags from requirements

srstrickland:
Requirements files may contain lines like:

```
--index-url=http://some-pypi-instance/whatever
--trusted-host=some-pypi-instance
```

Currently we're forced to manually redefine them as `extra_pip_args`.
Here we introduce a minor change to pull appropriate lines from
the requirements lock file and register them as extra pip args.

## PR Checklist

I'm not sure if this warrants documentation updates (other than a changelog entry surely), because it seems like this is how things should've worked in the first place.  Which leaves me torn on whether this is a bugfix or a feature 😛  ... Anyways looking for guidance here, happy to make any documentation updates necessary.

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: 438

Currently, when we use:

```
pip_parse(
    name = "my_pip_requirements",
    requirements_lock = "@//3rdparty/python:full-requirements.txt",
)
```

We have to also manually specify:

```
extra_pip_args=[
        "--index-url=http://our-pypi-server/repository/pypi-all/simple",
        "--trusted-host=our-pypi-server",
]
```

... despite the fact that those flags are already in `full-requirements.txt`.  This is cumbersome because it means we have to define this in two places: one where we generate the `full-requirements.txt`, and they _have_ to line up.

## What is the new behavior?

`extra_pip_args` as passed are still used, but are augmented with lines from the `requirements_lock` file that correspond to pip arguments.

- [ ] Yes
- [x] No

## Other information

A somewhat suboptimal implementation, as it reads the requirements file for a second time, and manually looks for flag-like lines (those that start with `-` when stripped).  While the main parsing routine (`parse_requirements()`) is already extracting the flags (and verifying them), I found that internally it sets attributes on objects like `session` and `finder` ... but there wasn't an obvious way to turn those values (which become somewhat "esoteric" in terms of the classes that need them) back into the args that constructed them.  So I thought I would propose this first before embarking on a more "intrusive" approach.

### DIFF
--- a/python/pip_install/parse_requirements_to_bzl/__init__.py
+++ b/python/pip_install/parse_requirements_to_bzl/__init__.py
@@ -1,6 +1,8 @@
 import argparse
 import json
+import re
 import textwrap
+import shlex
 import sys
 from typing import List, Tuple
 
@@ -38,6 +40,27 @@ def deserialize_structured_args(args):
     return args
 
 
+def read_embedded_pip_args(args):
+    """Augment extra_pip_args from directives in requirements_lock file
+        Args:
+            args: deserialized args to read and modify
+    """
+    extra_pip_args = args.get("extra_pip_args")
+    requirements_lock = args["requirements_lock"]
+    with open(requirements_lock) as f:
+        embedded = [
+            line for line in f.readlines()
+            if line.strip().startswith('-')
+        ]
+
+    embedded = [
+        arg for args in embedded
+        for arg in shlex.split(args, comments=True)
+    ]
+    if embedded:
+        args["extra_pip_args"] = embedded + (extra_pip_args or [])
+
+
 def generate_parsed_requirements_contents(all_args: argparse.Namespace) -> str:
     """
     Parse each requirement from the requirements_lock file, and prepare arguments for each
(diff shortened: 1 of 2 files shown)


hrfuller:
@srstrickland Thanks for working on a fix here. What if you read the lines of the requirment file into a List and zip that together with the output of parse_requirements(). Whenever parse_requirements() returns None you can append the line of the same index into an auxillary list of flags returned from https://github.com/bazelbuild/rules_python/pull/446/files#diff-5c50a51b32223d31cf82a95e84a40688d85582316939818e6c1f8dbb4da830c5L13

srstrickland:
@hrfuller I could be misunderstanding your suggestion, but from what I can tell, doing so would only work if `parse_requirement()` yields something for every row, which it doesn't; it only `yield`s [when there is a "valid" _requirement_ line](https://github.com/pypa/pip/blob/main/src/pip/_internal/req/req_file.py#L151).  So the "flag" lines (and comment and blank lines) will just be skipped (well, flag lines are handled internally but nothing is `yield`ed).

I suppose in the same spirit of this approach, I could take all the lines (read manually) and remove from that set the lines that came out of the parsed set that seem to match... but that would leave me worried that too much is hinging on some potentially-fragile "matching" code.

hrfuller:
Closing in favor of #456

PR #452, opened 2021-04-12
add map_path_prefixes arg to py_wheel

andrewcunnin:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

py_wheel targets have a strip_path_prefixes argument, but not a map_path_prefixes argument.

Issue Number: N/A

## What is the new behavior?

py_wheel targets now have a map_path_prefixes argument, which allows developers to alter the prefixes within their wheels by passing in an array of '='-separated key-value pairs. this makes it easier for people shipping python wheels out of larger repositories to pull in dependencies from other directories.

- [ ] Yes
- [x] No


andrewcunnin:
I added examples to demo the new feature and I also typed up some documentation for both this feature and for the already-existing strip_path_prefixes attribute of py_wheel targets, but I wasn't sure where the best place to put it is. I'd be happy to add that in somewhere if that's wanted.

andrewcunnin:
@andyscott @brandjon @lberki @thundergolfer 
howdy! sorry for the ping, just wanted to make sure this pr doesn't get buried. let me know if you want me to change anything on this!

andrewcunnin:
> > this makes it easier for people shipping python wheels out of larger repositories to pull in dependencies from other directories.
> 
> If you've got examples out in the world of people doing this, that would be appreciated. Bazel or otherwise. It doesn't seem like a Bazel-specific need.

sorry, I'm just seeing this message! The goal is to emulate the setup.py [package_dir argument](https://docs.python.org/3/distutils/setupscript.html#listing-whole-packages). Right now I'm working with an example where we have an internally published python library that is a part of a larger mono-repo because it's being imported directly within adjacent projects and it also shares some logic with other parts of the project, but it is also being used outside of the repo. 

We currently have to repackage this using a bash script that we launch with bazel, which basically just unpacks the bazel-generated wheel and then repackages it with the updated prefixes. It'd be nice to be able to do away with that script and have an easier mechanism to package our library!

mbeards:
> > this makes it easier for people shipping python wheels out of larger repositories to pull in dependencies from other directories.
> 
> If you've got examples out in the world of people doing this, that would be appreciated. Bazel or otherwise. It doesn't seem like a Bazel-specific need.

We're doing this as well - it's tricky to get the import paths to line up between bazel and venv-oriented workflows, so an escape hatch like this would help a lot.

PR #454, opened 2021-04-13
Add configuration for stale bot

alexeagle:
Ideally we would have the resources to address and resolve all issues, and review and merge all pull requests. But we don't.
Rather than leave these open indefinitely, it's better to set user expectations that these will just age out.


hrfuller (approved):
LGTM! Thanks.

PR #456, opened 2021-04-14
Parse requirements files for global pip flags.

### LINKED ISSUE #438: Parse pip flags from requirements_lock.txt
### Description

As a follow up to #432 and the issue brought up by @skliarpawlo we aren't capturing pip flags that may be passed in the requirements_lock.txt file.

E.g
requirements_lock.txt:
```
--index-url=some.index.com/simple
foo > 0.0.0
bar==1.0.0
...
```
--index-url is ignored, but in `pip_install` it works because the requirements file is passed verbatim to pip there.

Let's use pip to parse those arguments out of the requirements lock and amend the extra_pip_args array with them.

### LINKED ISSUE #447: pip_parse ignores hashes and is incompatible with custom sources
# 🐞 bug report

### Affected Rule

The issue is caused by the rule: `pip_parse`

### Is this a regression?

No.

### Description

`pip_parse` does not work with custom (URL-based) sources of packages.

## 🔬 Minimal Reproduction

If I have this listed in my `requirements_lock`, the rule will actually install `spacy-transformers` from PyPI with no version specified instead of downloading the archive from github. 

```
https://github.com/explosion/spacy-transformers/archive/015daa34efbf2202b56080954832d029b99f821a.zip#egg=spacy-transformers \
    --hash=sha256:f81470a17f999975b525d265a41d28a417944984774a710387cafeba47e0407
```

Additionally, it looks like the `--hash` is completely ignored for all packages.

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
Linux
  </code>
</pre>

<pre>
  <code>
Build label: 4.0.0
Build time: Thu Jan 21 07:33:24 2021 (1611214404)
Build timestamp: 1611214404
Build timestamp as int: 1611214404
  </code>
</pre>

**Rules_python version:** 

16965937f5e1f1ddc888c523323d0db8155241b1

hrfuller:
Follow up to #432 
Parse requirements files for global pip flags, then add them to the extra_pip_args array.

Pass a copy of each cleaned up line (comments and line-breaks removed) to
each whl_library repo verbatim. Then link based requirements or requirement
lines with requirement specific flags work.

Due to the pip restriction on using requirement specific flags in
requirements.txt files only, write each requirement line to a temp file
before invoking pip in each whl_library repo.

Fixes #438 
Fixes #447

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

All flags and url requirements in pip_parse are ignored.

Issue Number: N/A

## What is the new behavior?
Global pip flags requirement specific pip flags, and url style requirement specs are respected in pip_parse

- [ ] Yes
- [x] No


examples/pip_parse/requirements_lock.txt
```diff
@@ -2,15 +2,26 @@
 # This file is autogenerated by pip-compile
 # To update, run:
 #
-#    pip-compile --output-file=requirements_lock.txt requirements.txt
+#    pip-compile --generate-hashes --output-file=requirements_lock.txt requirements.txt
```
alexeagle, line 5: I am surprised there isn't a second place in this example where the generate-hashes argument was supplied.
If the point of the PR is to allow users to give pip flags, shouldn't there be a source file where they do so? (as opposed to this generated file?)

python/pip_install/parse_requirements_to_bzl/__init__.py
```diff
@@ -7,23 +7,40 @@
...
 
 
-def parse_install_requirements(requirements_lock: str) -> List[InstallRequirement]:
-    return [
-        constructors.install_req_from_parsed_requirement(pr)
-        for pr in parse_requirements(requirements_lock, session=PipSession())
-    ]
+def parse_install_requirements(requirements_lock: str, extra_pip_args: List[str]) -> List[Tuple[InstallRequirement, str]]:
+    ps = PipSession()
+    # This is roughly taken from pip._internal.req.req_file.parse_requirements in order to keep
```
alexeagle, line 16: thanks for the reference!
nit: even better to link to the sources you're talking about at the specific version, I wouldn't know where to find this

alexeagle (approved):

examples/pip_parse/requirements_lock.txt
hrfuller: I would think of this file, which happens to be generated, as the only input from the user. The fact that there is also a requirements.txt in this directory is because we are using pip-compile to create the lock file. I can modify this file to add some pip arguments for the sake of demonstration.

python/pip_install/parse_requirements_to_bzl/__init__.py
hrfuller: Will do.

examples/pip_parse/requirements_lock.txt
hrfuller: Adding this turned up a small bug, so yay.

Issue #440, opened 2021-03-26
Repositories generated by pip_parse contain a BUILD and a BUILD.bazel


person142:
### Affected Rule

`pip_parse`

### Is this a regression?

No

### Description

The repositories generated by `pip_parse` contain a `BUILD` file and a `BUILD.bazel` file; the former shouldn't be there. For example `ls`ing the contents of an `attrs` repository gives:

```
BUILD
BUILD.bazel
WORKSPACE
attr
attrs-20.3.0.dist-info
```

In addition, the `BUILD.bazel` file doesn't `exclude` the `BUILD` file in the glob (only the `BUILD.bazel`), so it gets injected into the built artifacts.

## 🔬 Minimal Reproduction

Inspect the contents of a repository generated by `pip_parse`.

## 🔥 Exception or Error

None

## 🌍  Your Environment

**Operating System:**

Seen on both Linux/macOS

**Output of `bazel version`:**

```
$ bazelisk version
Bazelisk version: v1.5.0
Build label: 3.3.1
Build target: bazel-out/darwin-opt/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Tue Jun 30 15:14:05 2020 (1593530045)
Build timestamp: 1593530045
Build timestamp as int: 1593530045
```

**Rules_python version:**

`7aaf7627485c84c4ecd976f7b3fa940f60bb007c`

**Anything else relevant?**

No

PR #427, opened 2021-03-02
Use BUILD.bazel file when creating py_library rules of extracted whls

kageiit:
## PR Checklist

Please check if your PR fulfills the following requirements:

- [x] Does not include precompiled binaries, eg. `.par` files. See [CONTRIBUTING.md](/CONTRIBUTING.md) for info
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type

What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature (please, look at the "Scope of the project" section in the README.md file)
- [ ] Code style update (formatting, local variables)
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] Other... Please describe:

## What is the current behavior?

Using wheels like keyring (https://pypi.org/project/keyring) , `pip_install` fails with an error like
```
  File "/private/var/tmp/_bazel_gkorlam/a1f9cb65ef7d02cf04534e97e406c6b9/external/rules_python/python/pip_install/extract_wheels/lib/bazel.py", line 187, in extract_wheel
    with open(os.path.join(directory, "BUILD"), "w") as build_file:
IsADirectoryError: [Errno 21] Is a directory: 'pypi__keyring/BUILD'
```

This is because the extracted whl file has a `build` directory in it.
Screen Shot 2021-03-02 at 2 00 39 PM

Issue Number: N/A

## What is the new behavior?
By changing the build file to `BUILD.bazel` it is much less likely to encounter this issue. `pip_install` succeeds in importing the keyring whl after this change.

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

PR #457, opened 2021-04-17
Also name the BUILD file in external repository roots BUILD.bazel

person142:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Closes https://github.com/bazelbuild/rules_python/issues/440

The build file in the repository roots generated by `pip_repository` is still `BUILD` and not `BUILD.bazel`. Since `pip_parse` has only one layer of directories, the external repository then ends up with a `BUILD` and a `BUILD.bazel`.

## What is the new behavior?

It's now called `BUILD.bazel`, which means `pip_parse` later overwrites it with the final `BUILD.bazel`.

- [ ] Yes
- [x] No

## Other information

This prevents a regression of https://github.com/bazelbuild/rules_python/pull/427 with `pip_parse`.

### DIFF
--- a/python/pip_install/pip_repository.bzl
+++ b/python/pip_install/pip_repository.bzl
@@ -12,7 +12,7 @@ def _construct_pypath(rctx):
         rctx: Handle to the repository_context.
     Returns: String of the PYTHONPATH.
     """
-    rctx.file("BUILD", "")
+    rctx.file("BUILD.bazel", "")
 
     # Get the root directory of these rules
     rules_root = rctx.path(Label("//:BUILD")).dirname
--- a/python/pip_install/pip_repository.bzl
+++ b/python/pip_install/pip_repository.bzl
@@ -12,7 +12,6 @@ def _construct_pypath(rctx):
         rctx: Handle to the repository_context.
     Returns: String of the PYTHONPATH.
     """
-    rctx.file("BUILD.bazel", "")
 
     # Get the root directory of these rules
     rules_root = rctx.path(Label("//:BUILD")).dirname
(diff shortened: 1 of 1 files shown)


person142:
This comment: https://github.com/bazelbuild/rules_python/pull/441#issuecomment-821870343 gives an example that fails on master but passes on this branch.

python/pip_install/pip_repository.bzl
```diff
@@ -12,7 +12,7 @@ def _construct_pypath(rctx):
         rctx: Handle to the repository_context.
     Returns: String of the PYTHONPATH.
     """
-    rctx.file("BUILD", "")
+    rctx.file("BUILD.bazel", "")
```
hrfuller, line 15: Can you move this line inside the else block [here](https://github.com/bazelbuild/rules_python/pull/457/files#diff-bced444149cf9f25b36837efb2792070f362650b01f553d9d01bef080ebfefabR82)?

It shouldn't have been part of _construct_pypath in the first place, which was my mistake while refactoring. The root level BUILD.bazel file is only needed by pip_install. pip_parse already creates a root level BUILD.bazel via the extract_wheel function.

person142: Done in https://github.com/bazelbuild/rules_python/pull/457/commits/6e922451e38acef7e80105df20b2f9ffb4dd47eb.

The `here` link above links back to the comment, but I'm assuming you were referring to this `else`:

https://github.com/bazelbuild/rules_python/blob/master/python/pip_install/pip_repository.bzl#L82

I put it outside that conditional, since we do want the empty build file inside the repository where the `requirements.bzl` is generated-otherwise user code can't load from it. But it will be left out of the repositories for the individual packages still because it's not being injected here:

https://github.com/bazelbuild/rules_python/blob/master/python/pip_install/pip_repository.bzl#L215

hrfuller: Thank you! Yes you read my mind. I was mixing up with whl_library invocation for that else block.

hrfuller (approved):

person142:
Thanks for reviewing @hrfuller.

PR #467, opened 2021-05-08
compile_pip_requirements: Fix defaulted macro args behaviour

thundergolfer:

* The `requirements_in` arg cannot be popped from `**kwargs` because it is a named argument. So the arg value is always: `name + ".in"`
* The `requirements_txt` arg is actually set by passing `requirements_locked` to the macro. If you pass the latter, you can set the former. If you just pass the former, the arg value is always `name + ".txt"`, similar to the above.

The result is that the `.in` file must always be `${name}.in` and you have to set the lock file name via an undocumented kwarg.

## What is the new behavior?

**Both args are given `None` defaults which the macro body will immediately detect and replace with the `name + ".in"` and `name + ".txt"` value if needed.**

With this change the following macro instantiation now works as expected: 

```python
compile_pip_requirements(
    name = "requirements_annotated",
    requirements_in = "requirements.in",
    requirements_locked = "requirements.annotated.txt",
)
```

I could have done `requirements_in = requirements_in or name + ".in`, but I wanted to explictly check fo `None` and not catch anything falsey.

Another alternative is keeping the `= "requirements.in"` defaulting and abandoning the defaulting to the `name + <ext>` setup.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

- [ ] Yes
- [x] No

### DIFF
--- a/python/pip_install/requirements.bzl
+++ b/python/pip_install/requirements.bzl
@@ -7,8 +7,8 @@ def compile_pip_requirements(
         name,
         extra_args = [],
         visibility = ["//visibility:private"],
-        requirements_in = "requirements.in",
-        requirements_txt = "requirements.txt",
+        requirements_in = None,
+        requirements_txt = None,
         **kwargs):
     """
     Macro creating targets for running pip-compile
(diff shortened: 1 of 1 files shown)


alexeagle (approved):
Thanks!!

Issue #478, opened 2021-06-01
requirements_test always pass if pip-compile fails to resolve dependency


reiyw:
# 🐞 bug report

### Affected Rule

The issue is caused by the rule: `compile_pip_requirements`

### Is this a regression?

No.

### Description

`compile_pip_requirements` produces a test target for checking pip-compile. This test generates a lock file from given `requirements_in`, and fails if the content is different from the original `requirements_txt`. However, if pip-compile fails to resolve the dependency, the test passes regardless of the contents of `requirements_txt`.

The cause of this issue is in: https://github.com/bazelbuild/rules_python/blob/c6970fc44877dbbbce84d17845d9bc797aefe299/python/pip_install/pip_compile.py#L70-L89
[pip-compile exits with code 2](https://github.com/jazzband/pip-tools/blob/5.5.0/piptools/scripts/compile.py#L459) when it fails to resolve the dependency. However, the error code is not checked, and here the contents of `requirements_txt` and `requirements_out` are identical (copied immediately before and not updated by pip-compile), so the test passes.

I'm ready to send a PR to fix this issue, if relevant.

### Example Project

WORKSPACE
```starlark
load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")

# HEAD commit as of 2021-06-01.
RULES_PYTHON_COMMIT = "c6970fc44877dbbbce84d17845d9bc797aefe299"
RULES_PYTHON_DIGEST = "c43edeb59c6cb009f23f8e5603ce8fa1ca25a9084aae54513c82acfff06c3aaa"

http_archive(
    name = "rules_python",
    urls = [
        "https://github.com/bazelbuild/rules_python/archive/" + RULES_PYTHON_COMMIT + ".tar.gz",
    ],
    sha256 = RULES_PYTHON_DIGEST,
    strip_prefix = "rules_python-" + RULES_PYTHON_COMMIT,
)

load("@rules_python//python:pip.bzl", "pip_parse")
pip_parse(
    name = "my_deps",
    requirements_lock = "//:requirements.txt",
)

load("@my_deps//:requirements.bzl", "install_deps")
install_deps()
```

BUILD
```starlark
load("@rules_python//python/pip_install:requirements.bzl", "compile_pip_requirements")

compile_pip_requirements(
    name = "requirements",
    extra_args = [
        "--quiet",
    ],
)
```

requirements.in
```python
grpcio==1.38.0
[...]

PR #482, opened 2021-06-08
Check pip-compile failed to resolve requirements

reiyw:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #478 

When pip-compile fails to resolve the dependency, `requirements_test` will pass regardless of the content of the lock file.

## What is the new behavior?

When pip-compile fails to resolve the dependency, exit with an error message, and the test will fail.

- [ ] Yes
- [x] No


reiyw:
@googlebot I signed it!

python/pip_install/pip_compile.py
```diff
@@ -73,7 +73,15 @@
         cli()
         print("cli() should exit", file=sys.stderr)
         sys.exit(1)
-    except SystemExit:
+    except SystemExit as e:
+        if e.code == 2:
```
thundergolfer, line 77: This improves things for sure, but still leaves us open to cases where `e.code not in {0, 2}`. Looking at the piptools source, it only ever exits 0 or 2, but my preference would be to check for the unexpected cases and `exit(1)` instead of incorrectly succeeding .

```python
if e.code == 2:
   # handle 2
   # ...
   sys.exit(1)
elif e.code == 0:
   # handle 0
   # ...
   sys.exit(0)
else:
   # handle unexpected
   # ...
   sys.exit(1)
```

thundergolfer (approved):
Suggested adding extra error checking, but happy with the contributed fix 🙏

python/pip_install/pip_compile.py
reiyw: I totally agree. I have fixed it that way.

thundergolfer:
Thanks for the contribution 👍

PR #505, opened 2021-07-27
pip_parse: Fix when using a python wrapper script

plule-ansys:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

When using `pip_parse` with a python wrapper script in `python_interpreter` or `python_interpreter_target`, the declared `whl_library` do not use the wrapper script but directly the underlying python interpreter given by `sys.executable`.

We use a portable python distribution that needs some environment adaptation in the `PATH` and `LD_LIBRARY_PRELOAD` variables to run properly. Calling directly the python interpreter binary fails, so `pip_parse` itself is successful but the wheels cannot be retrieved.

## What is the new behavior?

 - `pip_parse` now transmits `python_interpreter` and `python_interpreter_target` attributes to the `whl_library` targets
 - `whl_library` now takes `python_interpreter_target` in account

- [ ] Yes
- [x] No

### DIFF
--- a/python/pip_install/parse_requirements_to_bzl/__init__.py
+++ b/python/pip_install/parse_requirements_to_bzl/__init__.py
@@ -141,7 +141,9 @@ dependencies from a fully resolved requirements lock file."
     )
     parser.add_argument(
         "--python_interpreter_target",
-        help="Bazel target of the python interpreter. It will take precedence over python_interpreter.",
+        help="Bazel target of a python interpreter.\
+It will be used in repository rules so it must be an already built interpreter.\
+If set, it will take precedence over python_interpreter.",
     )
     parser.add_argument(
         "--quiet",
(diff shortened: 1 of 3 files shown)


python/pip_install/parse_requirements_to_bzl/__init__.py
```diff
@@ -119,6 +119,14 @@ def main() -> None:
         required=True,
         help="Path to fully resolved requirements.txt to use as the source of repos.",
     )
+    parser.add_argument(
+        "--python_interpreter",
+        help="The python interpreter that will be used to download and unpack the wheels.",
+    )
+    parser.add_argument(
+        "--python_interpreter_target",
+        help="Bazel target of the python interpreter. It will take precedence over python_interpreter.",
```
hrfuller, line 128: Some documentation explaining that this has to be a reference to another external repository would be helpful. To me the name implies that this could be a reference to a target in the current workspace, which wouldn't work.

plule-ansys: That's right thanks, outside of the `.bzl` this becomes much less obvious, I've updated the documentation. I've not changed the name of the argument to keep it consistent with `whl_library`/`pip_parse` though. (sorry for the slow reaction, I was away from computers!).

thundergolfer:
@hrfuller :bump: this changes looks mergeable 👍

alexeagle (approved):
Thanks!!

Issue #260, opened 2019-11-15
"Cross compilation" of py_binary/py_image/py_library targets


nimish:
Hi,

I have a py_binary that depends on a python pip library (grpcio) that has a native extension bundled in. This means that to create a linux container i'd need to have the pip_import rule download the manylinux wheel, not the host one (macos in my case).

Is there a way to force this? Otherwise py_image will happily just bundle up wheels with darwin native libs. py_binary will also only make host-runnable things.

ali5h:
use https://github.com/ali5h/rules_pip/ and use `pip_install(["--platform=linux_x86_64"])`

nimish:
@ali5h this works (thanks!) but is there integration into the bazel platform selection functions? I don't want separate targets for Linux and Mac.

E.g. Something that works with https://docs.bazel.build/versions/master/platforms.html

ali5h:
you can define multiple piplib repos for different platforms and use select to pickup correct one

nimish:
Is that documented and supported anywhere? That would be the ideal case, for bazel to automatically pick up the right pip repo for the right target platform.

RemiTorracinta:
Would really, really like to see this as well.

brduarte:
Help

UebelAndre:
Friendly ping to @alexeagle and @meastham, are either of you able to take another look at your pull requests? I'd love to have that functionality 😃

UebelAndre:
I wonder if [pipenv](https://github.com/pypa/pipenv) could be used to facilitate generating cross platform dependency graphs. Does anyone have experience with the tool?

meastham:
> Friendly ping to @alexeagle and @meastham, are either of you able to take another look at your pull requests? I'd love to have that functionality 😃

Sorry for the long delay on this! I did some investigation on some of the open questions, I'll see if we can find a path to getting this merged.

jvolkman:
I'm all for the wheel-only approach taken in #510. I don't want native compilation happening at all in the loading phase, as it can lead to hard-to-diagnose cache misses between machines (in my experience).

[...]

PR #510, opened 2021-08-14
Add support for multi-platform `pip download` to pip_parse

meastham:
## PR Checklist

- [X] Does not include precompiled binaries, eg. `.par` files. See [CONTRIBUTING.md](https://github.com/bazelbuild/rules_python/blob/master/CONTRIBUTING.md) for info

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

There is no multi-platform support: https://github.com/bazelbuild/rules_python/issues/260

## What is the new behavior?

Adds a new argument, `pip_platform_definitions`, to `pip_parse`. The value of the argument looks like this:
```
    pip_platform_definitions = {
        "//platforms:linux_x86_64_build": "manylinux2014_x86_64-37-cp-cp37m",
        "//platforms:macos_x86_64_build": "macosx_10_14_x86_64-37-cp-cp37m",
    },
```
Each key is a `config_setting`, and each value is a string of the fom `<platform>-<python_version>-<implementation>-<abi>`.

When `pip_platform_definitions` is specified, instead of running `pip wheel`, the rules run `pip download` with constraints extracted from the values in the map values. This makes it possible to do cross-platform Python builds, e.g. on MacOS run bazel with `--platforms=//platforms:macos_x86_64` and have linux dependencies fetched properly. In the above example, we would end with 3 targets for a pip package called `foo`:
* `@pip_pypi__macosx_10_14_x86_64_37_cp_cp37m__foo//:pkg` is the `py_library` for the MacOS version of the package.
* `@pip_pypi__manylinux2014_x86_64_37_cp_cp37m__foo//:pkg` is the `py_library` for the Linux version of the package.
* `@pip_pypi__foo` is an `alias` with its `actual` determined by a `select` over the provided `config_settings` pointing at the above two targets. This is the target that the `requirement` macro points at.

Of course this only works if all of the required packages have wheels available which match the configured platform constraints, either in PyPI or another configured repository.

- [ ] Yes
- [X] No

## Other information

In its current state, this PR is more of a proof of concept that is intended to start a conversation rather than something that is ready to merge. If I get feedback from the maintainers that this general approach is something that could be merged, then I would be happy to do the work to get it into a better state.

We've been successfully been using this at my company to build Docker images with native Python dependencies (e.g. pandas, gRPC) on MacOS. We have a reasonably large collection of Python deps (124), and in practice it has not been a large problem to obtain wheels for all of them. All except 3 are already in PyPI, and for the 3 that are missing it's possible to build universal wheels, which we just do manually outside of Bazel.

Things that I think likely need improvement before this is ready to merge are:
* The interface to this new functionality is clunky. It's basically the fastest thing I could get working to see if the approach would work at all. I think the right way may be to create a `toolchain_type` for this purpose, and put the pip download constraints in a toolchain provider, but have not investigated this thoroughly.
* Documentation
* Examples/Tests
* Some general code cleanup. The new interface for `repo_names_and_requirements` is pretty confusing.
Feedback on any of these items would be great!

### DIFF
--- a/python/pip_install/parse_requirements_to_bzl/extract_single_wheel/__init__.py
+++ b/python/pip_install/parse_requirements_to_bzl/extract_single_wheel/__init__.py
@@ -21,6 +21,10 @@ def main() -> None:
         required=True,
         help="A single PEP508 requirement specifier string.",
     )
+    parser.add_argument(
+        "--pip_platform_definition",
+        help="A pip platform definition in the form <platform>-<python_version>-<implementation>-<abi>",
+    )
     arguments.parse_common_args(parser)
     args = parser.parse_args()
     deserialized_args = dict(vars(args))
@@ -28,10 +32,20 @@ def main() -> None:
 
     configure_reproducible_wheels()
 
-    pip_args = (
-        [sys.executable, "-m", "pip", "--isolated", "wheel", "--no-deps"] +
-        deserialized_args["extra_pip_args"]
-    )
+    pip_args = [sys.executable, "-m", "pip", "--isolated"]
+    if args.pip_platform_definition:
+        platform, python_version, implementation, abi = args.pip_platform_definition.split("-")
+        pip_args.extend([
+            "download",
+            "--only-binary", ":all:",
+            "--platform", platform,
+            "--python-version", python_version,
+            "--implementation", implementation,
+            "--abi", abi
+        ])
+    else:
+        pip_args.append("wheel")
+    pip_args.extend(["--no-deps"] + deserialized_args["extra_pip_args"])
 
     requirement_file = NamedTemporaryFile(mode='wb', delete=False)
     try:
--- a/python/pip_install/pip_repository.bzl
+++ b/python/pip_install/pip_repository.bzl
@@ -183,6 +183,11 @@ pip_repository_attrs = {
         default = False,
         doc = "Create the repository in incremental mode.",
     ),
+    "pip_platform_definitions": attr.label_keyed_string_dict(
+        doc = """
+A map of select keys to platform definitions in the form <platform>-<python_version>-<implementation>-<abi>"
+        """,
+    ),
     "requirements": attr.label(
         allow_single_file = True,
         doc = "A 'requirements.txt' pip requirements file.",
@@ -195,11 +200,6 @@ of 'requirements' no resolve will take place and pip_repository will create indi
 wheels are fetched/built only for the targets specified by 'build/run/test'.
 """,
     ),
-    "pip_platform_definitions": attr.label_keyed_string_dict(
-        doc = """
-A map of select keys to platform definitions in the form <platform>-<python_version>-<implementation>-<abi>"
-        """
-    )
 }
 
 pip_repository_attrs.update(**common_attrs)
@@ -282,6 +282,9 @@ def _impl_whl_library(rctx):
     return
 
 whl_library_attrs = {
+    "pip_platform_definition": attr.string(
+        doc = "A pip platform definition in the form <platform>-<python_version>-<implementation>-<abi>",
+    ),
     "repo": attr.string(
         mandatory = True,
         doc = "Pointer to parent repo name. Used to make these rules rerun if the parent repo changes.",
@@ -290,9 +293,6 @@ whl_library_attrs = {
         mandatory = True,
         doc = "Python requirement string describing the package to make available",
     ),
-    "pip_platform_definition": attr.string(
-        doc = "A pip platform definition in the form <platform>-<python_version>-<implementation>-<abi>",
-    )
 }
 
 whl_library_attrs.update(**common_attrs)
@@ -317,16 +317,16 @@ def _impl_platform_alias(rctx):
     rctx.file(
         "BUILD",
         content = _PLATFORM_ALIAS_TMPL.format(
-            select_items = rctx.attr.select_items
+            select_items = rctx.attr.select_items,
         ),
         executable = False,
     )
 
 platform_alias = repository_rule(
     attrs = {
-        "select_items": attr.string_dict()
+        "select_items": attr.string_dict(),
     },
     implementation = _impl_platform_alias,
     doc = """
-An internal rule used to create an alias for a pip package for the appropriate platform."""
+An internal rule used to create an alias for a pip package for the appropriate platform.""",
 )
--- a/python/pip_install/pip_repository.bzl
+++ b/python/pip_install/pip_repository.bzl
@@ -100,6 +100,11 @@ def _pip_repository_impl(rctx):
             "--timeout",
             str(rctx.attr.timeout),
         ]
+        if rctx.attr.pip_platform_definitions:
+            args.extend([
+                "--pip_platform_definitions",
+                struct(arg = {str(k): v for k, v in rctx.attr.pip_platform_definitions.items()}).to_json(),
+            ])
     else:
         args = [
             python_interpreter,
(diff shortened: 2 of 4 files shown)


mattem:
As an overall feature I think this is okay, but I'd be careful in the cross / multi platform wording.

The general issue is that pip tools (the rule that powers `pip_compile`) doesn't produce a locked requirements file that is cross platform, so while you're getting the Linux wheels for what's listed, any dependencies that have a platform constraint next to them in the input requirements file (or transitive requirements) that don't match the platform the lock file was generated on will be missing from the lock file, at which point you could be missing requirements for a given platform except the one it was locked on. 

I guess what I'm trying to say is, this is constrained by the locked requirements file as the input.

meastham:
> As an overall feature I think this is okay, but I'd be careful in the cross / multi platform wording.
> 
> The general issue is that pip tools (the rule that powers `pip_compile`) doesn't produce a locked requirements file that is cross platform, so while you're getting the Linux wheels for what's listed, any dependencies that have a platform constraint next to them in the input requirements file (or transitive requirements) that don't match the platform the lock file was generated on will be missing from the lock file, at which point you could be missing requirements for a given platform except the one it was locked on.
> 
> I guess what I'm trying to say is, this is constrained by the locked requirements file as the input.

Yes, you're 100% correct that that is a gotcha that you have to be on the lookout for. I think this can already affect users of `pip_compile` if they're using the same lock file on multiple host platforms though right?

Also I see that you're a collaborator, do you know how I can find out if I would be able to get this merged if I fix up the list of issues in the PR description?

python/pip_install/pip_repository.bzl
```diff
@@ -178,6 +183,11 @@ pip_repository_attrs = {
         default = False,
         doc = "Create the repository in incremental mode.",
     ),
+    "pip_platform_definitions": attr.label_keyed_string_dict(
+        doc = """
+A map of select keys to platform definitions in the form <platform>-<python_version>-<implementation>-<abi>"
```
thundergolfer, line 188:
```suggestion
A map of select keys to Python platform definitions in the form <platform>-<python_version>-<implementation>-<abi>"
```

Just a nit, but helpful to distinguish from a Bazel `platform`.

python/pip_install/parse_requirements_to_bzl/extract_single_wheel/__init__.py
```diff
@@ -21,17 +21,31 @@ def main() -> None:
...
 
     configure_reproducible_wheels()
 
-    pip_args = (
-        [sys.executable, "-m", "pip", "--isolated", "wheel", "--no-deps"] +
-        deserialized_args["extra_pip_args"]
-    )
+    pip_args = [sys.executable, "-m", "pip", "--isolated"]
+    if args.pip_platform_definition:
+        platform, python_version, implementation, abi = args.pip_platform_definition.split("-")
```
thundergolfer, line 37: Q: why this format instead of the format listed here in [PEP 425](https://www.python.org/dev/peps/pep-0425/#id8)? 

`{python tag}-{abi tag}-{platform tag}`

The extra bit in this PR's string is `{implementation}`, `cp = CPython`, but this is covered by the `{python tag}` from PEP 425. 

> The Python tag indicates the implementation and version required by a distribution.

thundergolfer:
Python packaging is hell, isn't it? Appreciate the PR and the context you've provided in "other information". I think the biggest roadblocks to supporting cross-platform this way are the sharp-edges/pitfalls users have to navigate:

* Package supports becomes 'wheel only'. How is this communicated? Ref: "Of course this only works if all of the required packages have wheels available which match the configured platform constraints"
* A single locked `requirements.txt` file is only really valid on the platform that produced it. So do we support multiple files? @alexeagle has a draft out for this now https://github.com/bazelbuild/rules_python/pull/531/files
* Feels strange to allow providing strings which may include requirements (eg. python 3.10) which relate to no toolchain in the workspace. You touch on this here: "I think the right way may be to create a toolchain_type for this purpose". If I'm going to download third-party build artefacts that become inputs to some `py_library/binary/test`, they ought to only be build artefacts that match up to a Python runtime defined in the workspace.

But generally I think moving to a 'download only' install phase is solid.

thundergolfer:
Python packaging is hell, isn't it? Appreciate the PR and the context you've provided in "other information". I think the biggest roadblocks to supporting cross-platform this way are the sharp-edges/pitfalls users have to navigate:

* Package supports becomes 'wheel only'. How is this communicated? Ref: "Of course this only works if all of the required packages have wheels available which match the configured platform constraints"
* A single locked `requirements.txt` file is only really valid on the platform that produced it. So do we support multiple files? @alexeagle has a draft out for this now https://github.com/bazelbuild/rules_python/pull/531/files


But generally I think moving to a 'download only' install phase is solid.

NathanHowell:
> Currently I am setting up a bazel repository and the setup is that most of the developers are using Macs where as our CI is running on Ubuntu20.04, so wheel only installation would be something that would be really good to be able to use.

poetry and pipenv-resolve, if not others, support a single lock file across platform for (most? all?) wheels... worth considering if you have wheel only dependencies. it'd be nice if pip-compile added an option for the same. https://github.com/soniaai/rules_poetry has Poetry rules for Bazel (note: I'm the original author).

meastham:
Thanks for the replies and sorry for the long delay on this!

> * A single locked `requirements.txt` file is only really valid on the platform that produced it. So do we support multiple files? @alexeagle has a draft out for this now https://github.com/bazelbuild/rules_python/pull/531/files

This is a good point, but in practice I've found it not cause very many problems in practice with our repo. For the few cases where we have platform-specific transitive dependencies I've just added them as unconditional top level dependencies in our `requirements.in`. We haven't run into the case where there's a platform-specific dep that isn't available on other platforms; in that case you would have to do some hacky like manually create a dummy empty wheel for that dep and include it in an extra index.

Do you think coming up with a better solution to this problem would block landing this or could we just note it as a caveat when documenting the feature?

> * Feels strange to allow providing strings which may include requirements (eg. python 3.10) which relate to no toolchain in the workspace. You touch on this here: "I think the right way may be to create a toolchain_type for this purpose". If I'm going to download third-party build artefacts that become inputs to some `py_library/binary/test`, they ought to only be build artefacts that match up to a Python runtime defined in the workspace.

I dug into how this would look, and I'm not really sure how much of a practical benefit there would be. `py_runtime`/`py_runtime_pair` don't contain the necessary information, so I think we would need to use a separate parallel "pip toolchain." Additionally, toolchains aren't available to `repository_rule`s, which is where we need to run pip download. That would mean we would need a list of all of the "pip toolchains" passed in to `pip_parse` (which is not how toolchains usually work), and then like the current implementation we would download and extract all of the candidate wheels at repository_rule time and later use the resolved toolchain to select the correct version at build time. This seems pretty awkward and introduces new error modes (i.e. not passing all of the registered toolchains to `pip_parse`), so it doesn't really seem better to me.

Does that analysis seem correct? Is there a better way I'm missing?

python/pip_install/parse_requirements_to_bzl/extract_single_wheel/__init__.py
meastham: It's because `pip download` takes four flags: `--platform`, `--python-version`, `--implementation`, `--abi`. I'm not sure why pip doesn't line up with the PEP 425 format.

groodt:
Decided to go ahead with this simpler approach, that has a similar outcome. https://github.com/bazelbuild/rules_python/pull/773

Issue #335, opened 2020-07-09
Values in /etc/pip.conf and /etc/xdg/pip/pip.conf are not ignored by pip despite `--isolated` flag


georgegiovanni:
Use of the `--isolated ` flag in the bundled `piptool.par` ignores environment variable `PIP_INDEX_URL` as expected, but not index urls defined in `/etc/pip.conf` and `/etc/xdg/pip/pip.conf`. This may be the intended behaviour for pip: https://github.com/pypa/pip/issues/3828, but presents challenges for consistent builds across platforms. Seems to ignore `PIP_CONFIG_FILE` as well.

A passing build on one platform that has a private pypi index server specified in a system configuration file such as `/etc/pip.conf`, will fail on another that doesn't have this file of course, but there's not even a way to provide a value for this custom index url with an environment variable and `--action_env` flag. Therefore, in an automated continuous integration environment one cannot simply add an environment variable for this non-hermetic operation, and instead needs to create the system configuration file and place it exactly at `/etc/pip.conf`.

Support for extra pip args was recently added in, https://github.com/bazelbuild/rules_python/commit/dd7f9c5f01bafbfea08c44092b6b0c8fc8fcb77f, which can solve this issue by hardcoding the index urls in the `pip_import` rule declaration, but then one loses the flexibility of having an environment variable that can have different values depending on the platform.

The location of the index server should not matter for hermetic build purposes as long as the underlying data is the same and produces the same hash. Same argument goes for tools found under `PATH` and HTTP mirror urls.

In short, what's the best way to provide a custom value for the index url so one can install private repositories?

PR #512, opened 2021-08-16
Allow the `--isolated` pip flag to be optionally unset

UebelAndre:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

In adopting a wider use of `pip_parse` I've run into an issue where the pip config file which defines the cert location for my index_url is not being parsed, leading to failures when attempting to download packages.

Issue Number: N/A

## What is the new behavior?

This PR introduces flags for optionally disabling the use of the [--isolated](https://pip.pypa.io/en/stable/cli/pip/#cmdoption-isolated) flag that's currently always passed when invoking pip commands. Users can set `isolated = False` on uses of [pip_parse](https://github.com/bazelbuild/rules_python/blob/304b85cdb7be888a8aa8dce80c1e49374abdac18/docs/pip.md#pip_parse) and [pip_install](https://github.com/bazelbuild/rules_python/blob/304b85cdb7be888a8aa8dce80c1e49374abdac18/docs/pip.md#pip_install). Additionally, the environment variable `RULES_PYTHON_PIP_ISOLATED` can be set to a falsey value (`"0"`, `"False"`) to accomplish the same. The environment variable is intended to allow for this kind of configuration in `.bazelrc` files where users could define `common --repo_env=RULES_PYTHON_PIP_ISOLATED` to allow for this behavior in specific cases (like on CI machines).

I'm not sure if making `--isolated` optional is the best approach, but I do want to be able to rely on the machine's pip config to determine how to download things. The first alternative I reached for was to use `extra_pip_args` to pass [--cert](https://pip.pypa.io/en/stable/cli/pip/#cmdoption-cert) but next ran into an issue where `--cert` was different on different machines. I'm happy to hear some other alternatives though.

- [ ] Yes
- [x] No


UebelAndre:
Another option for fixing this without a change to the repo would be https://github.com/bazelbuild/rules_python/issues/335 but for me this is undesirable since since it requires an infra change for something that seems like inconsistent behavior in pip(?)

python/pip_install/pip_repository.bzl
```diff
@@ -36,6 +36,20 @@ def _parse_optional_attrs(rctx, args):
     Returns: Augmented args list.
     """
 
+    # Determine whether or not to pass the pip `--isloated` flag to the pip invocation
+    use_isolated = rctx.attr.isolated
+
+    # The environment variable will take precedence over the attribute
+    isolated_env = rctx.os.environ.get("RULES_PYTHON_PIP_ISOLATED", None)
+    if isolated_env != None:
+        if isolated_env.lower() in ["", "0", "false"]:
```
hrfuller, line 45: The empty string doesn't make sense as a falsey value to me. Can we keep this to set values?
```suggestion
        if isolated_env.lower() in ("0", "false"):
```

UebelAndre: Yeah, sounds good to me 😃

UebelAndre:
@hrfuller Hey, anything more I need to do for this PR?

UebelAndre:
@hrfuller Hey, sorry to ping again but it'd be really great if this were to be merged 🙏 😅

hrfuller (approved):

hrfuller:
LGTM. Merging. Thanks @UebelAndre

PR #531, opened 2021-09-08
Allow for requirements files to differ per platform

alexeagle:
As a common example, we need a compiled requirements file for linux that differs from mac os

## PR Checklist

- [ ] Does not include precompiled binaries, eg. `.par` files. See [CONTRIBUTING.md](https://github.com/bazelbuild/rules_python/blob/master/CONTRIBUTING.md) for info

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+docs/*.md linguist-generated=true
--- a/docs/BUILD
+++ b/docs/BUILD
@@ -24,6 +24,7 @@ licenses(["notice"])  # Apache 2.0
 _DOCS = {
     "packaging": "//docs:packaging-docs",
     "pip": "//docs:pip-docs",
+    "pip_repository": "//docs:pip-repository",
     "python": "//docs:core-docs",
 }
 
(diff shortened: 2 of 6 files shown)


mattem (approved):

UebelAndre:
I think this is slightly incorrect. Instead of selecting the requirements based on the host, a dependency graph should be created for each target platform, so if I build on a macos host and target a linux platform (maybe for a docker container), I get the right set of requirements. Right now it'd be wrong unless my dependency graph happens to be the same.

alexeagle:
@UebelAndre the actual install has to happen on the target platform so in a docker container you'd copy in the Linux requirements and use pip_install. There's no Mac outputs of pip_parse that can be useful there.

alexeagle:
Note to self: this PR only updated the pip_parse/incremental=True codepath to use the platform-specific lockfile. `pip_install` usage still ends up with the unlocked requirements, and isn't platform-aware.

Issue #526, opened 2021-09-06
Increase min tested Bazel version to 4.0.0


UebelAndre:
# 🚀 feature request

### Description

 A clear and concise description of the problem or missing capability...

Writing tests and new features for these rules is difficult since there are a few new features (for me, the `env` attribute and `json` module) that are incredibly useful pretty un-fun to work around, making changes take more time than the need to. The min tested version should be bumped to 4.0.0 to enable the use of these features.

### Describe the solution you'd like

 If you have a solution in mind, please describe it.
Update the bazel versions in [.bazelrc](https://github.com/bazelbuild/rules_python/blob/bef2244d295ea5bf979261513b31bf38b941ac28/.bazelversion) and [version.bzl](https://github.com/bazelbuild/rules_python/blob/bef2244d295ea5bf979261513b31bf38b941ac28/version.bzl#L23) to `4.0.0`

### Describe alternatives you've considered

 Have you considered any alternative solutions or workarounds?
Alternatively, it'd be nice to have a well documented way to conditionally use things in newer versions of Bazel. Perhaps `native.bazel_version` could be made available to rules and macros (not just repository rules) (https://github.com/bazelbuild/bazel/issues/8305) but that would likely also require some way to compare versions which I'm only aware of [@bazel_skylib//lib:versions.bzl](https://github.com/bazelbuild/bazel-skylib/blob/1.0.3/docs/versions_doc.md).

alexeagle:
Agreed - it's a breaking change for users, but we made the same in rules_nodejs to reduce support burden. LTS and rolling releases are the only ones I think we need to support.

PR #533, opened 2021-09-13
Updated min tested Bazel version to 4.0.0

UebelAndre:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: https://github.com/bazelbuild/rules_python/issues/526

## What is the new behavior?

Updates the min tested Bazel version to `4.0.0`

- [x] Yes
- [ ] No

## Other information
closes https://github.com/bazelbuild/rules_python/issues/526


hrfuller (approved):
I'm +1 on this, will wait for @alexeagle stamp as well. How risky is this upgrade, it may make future releases incompatible for people who can't go to newer bazel version right?

alexeagle:
We'll need to document this as a breaking change, that users must upgrade to 4.0. We'll start using new features like the `json` module which will fail. It has been the LTS version for a while, and I think we are not required to support older ones, so policy-wise I believe this is fine. (We ought to have some testing against rolling releases as well. I hope the rules authors SIG will make some guidance about what rulesets are expected to do)

One thing to figure out (maybe before merging this?) is how to give the right error message so developers know they must upgrade, instead of just reporting [name 'json' is not defined](https://github.com/bazelbuild/rules_nodejs/issues/2938)
@UebelAndre could you find the right spot to hook up a `versions.check` https://docs.aspect.dev/bazelbuild/bazel-skylib/1.0.3/docs/versions.html#versions-check
?

UebelAndre:
I unfortunately don't have a whole lot of bandwidth so don't think I'll be able to get to that in the near future 😞 Is this something I can leave to the `rules_python` maintainers?

alexeagle:
I think I could find time for it. I'm looking around for where the check could go - there's no single call from users WORKSPACE files into a repository rule that we could put it. So I guess we just check at the beginning of several of them like `pip_install` and `pip_import`

alexeagle:
@googlebot I consent.

python/pip_install/repositories.bzl
```diff
@@ -3,6 +3,9 @@
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 load("@bazel_tools//tools/build_defs/repo:utils.bzl", "maybe")
 
+# Avoid a load from @bazel_skylib repository as users don't necessarily have it installed
```
hrfuller, line 6: Looks like versions.bzl tries to load skylib anyway. https://github.com/bazelbuild/rules_python/pull/533/files#diff-86c5bb17276af4fdb85612635eabfdaf9b91676ae6bac37ee78a09947dfd6045R1

hrfuller:
@googlebot I consent.

python/pip_install/repositories.bzl
alexeagle: thanks for fixing it!

PR #534, opened 2021-09-14
Exposed docs for `compile_pip_requirements`

UebelAndre:
## PR Checklist

Please check if your PR fulfills the following requirements:

- [x] Does not include precompiled binaries, eg. `.par` files. See [CONTRIBUTING.md](https://github.com/bazelbuild/rules_python/blob/master/CONTRIBUTING.md) for info
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type

What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature (please, look at the "Scope of the project" section in the README.md file)
- [ ] Code style update (formatting, local variables)
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] Other... Please describe:

## What is the current behavior?

`compile_pip_requirements` is not currently in the docs and seems to otherwise be hidden as a result. This PR fixes that.

Issue Number: N/A

## What is the new behavior?
`compile_pip_requirements` can now be loaded via `@rules_python//python:pip.bzl` which is also used to populate `docs/pip.md` with the appropriate docs for the rule.

## Does this PR introduce a breaking change?

- [ ] Yes
- [ ] No

PR #535, opened 2021-09-16
Update docstrings for packaging rules/macros

thundergolfer:
Prompted by https://github.com/bazelbuild/rules_python/pull/534 and reading through the docstrings, I'm putting up these edits to clean up things.

* Avoiding use of "import". Going way back to `pip_import`, I thought "import" was a misleading term. The rules need to install packages, and this involves running sometimes arbitrary on the system if a wheel isn't available. "import" also clashes with Python's `import` keyword. Rulesets such as `rules_jvm_external` use "resolve" and "fetch", which is appropriate for JVM dependencies because they are just file bundles you download, but this Python stuff you're installing.
* Adding deprecation notices in docstrings, because #534 showed me that the generated documentation doesn't show anything for our deprecated rules and it's a bit confusing to see no detail.
* Generally trying to bring docstring wording closer to rules_jvm_external. eg. use of "generates", "this macro runs a repository rule". 

---

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

- [ ] Yes
- [x] No

### DIFF
--- a/python/pip.bzl
+++ b/python/pip.bzl
@@ -17,9 +17,11 @@ load("//python/pip_install:pip_repository.bzl", "pip_repository")
 load("//python/pip_install:repositories.bzl", "pip_install_dependencies")
 
 def pip_install(requirements, name = "pip", **kwargs):
-    """Imports a `requirements.txt` file and generates a new `requirements.bzl` file.
+    """Accepts a `requirements.txt` file and installs the dependencies listed within.
 
-    This is used via the `WORKSPACE` pattern:
+    Those dependencies become available in a generated `requirements.bzl` file.
+
+    This macro runs a repository rule that invokes `pip`. In your WORKSPACE file:
 
     ```python
     pip_install(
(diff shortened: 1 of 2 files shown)


python/pip.bzl
```diff
@@ -42,9 +44,9 @@ def pip_install(requirements, name = "pip", **kwargs):
     )
     ```
```
alexeagle, line 46:
```suggestion
Note that this convenience comes with a cost. Every target in the BUILD file which loads the requirements helper in this way will cause an eager-fetch of all the pip dependencies, even if no python targets are requested to be built. In a multi-language repo, this may cause developers to fetch dependencies they don't need, so consider using the long form for imports if this happens.
```

alexeagle (commented):
I think it's not desirable that the markdown in the docs/ folder no longer matches the starlark generation if this PR lands. Someone would have to remember to run the update_docs.sh and we probably won't.

Let's do https://github.com/bazelbuild/rules_python/pull/540 so it's ergonomic and quick to keep them in sync always.

PR #540, opened 2021-09-28
docs: enforce that api docs are up-to-date

alexeagle:
Previously we just suggested you run a shell script to update, but there was no clue when that was needed in a PR


alexeagle:
I don't understand if the `main` branch is green, how the futures problem is just happening now. I'll try a rebase just to roll the dice

PR #545, opened 2021-10-04
Add support for stamping the version in py_wheel

tmspicer:
## PR Checklist

- [ X] Does not include precompiled binaries, eg. `.par` files. See [CONTRIBUTING.md](https://github.com/bazelbuild/rules_python/blob/master/CONTRIBUTING.md) for info

## PR Type

- [ ] Bugfix
- [ X] Feature (please, look at the "Scope of the project" section in the README.md file)
- [ ] Build related changes
- [ ] CI related changes

Currently the version attribute for py_wheel is a string which requires hardcoding it in the BUILD file.

Issue Number: 544

## What is the new behavior?

The version attribute can either be a hardcoded string or it can be a stamped version which will be resolved during the process.
Example:
```
version = "{STABLE_GIT_COMMIT}"
```

In the event that `py_wheel` has `stamp = True` then the output will be a directory with the stamp not resolved containing a whl file with the stamp resolved.   This is to work around bazel not being able to resolve stamps until the execution stage but we must declare an output.  Not completely ideal but it was the only way I was able to workaround the file being dynamic depending on the version stamp.

Example of new usage:
```py_wheel(
              version = "{STABLE_GIT_COMMIT}",
              stamp = True,
              ...
      )
```

- [ ] Yes
- [ X] No

### DIFF
--- a/python/packaging.bzl
+++ b/python/packaging.bzl
@@ -83,7 +83,7 @@ Sub-packages are automatically included.
     },
 )
 
-def _escape_filename_segment(segment):
+def _escape_filename_segment(segment, stamp = False):
     """Escape a segment of the wheel filename.
 
     See https://www.python.org/dev/peps/pep-0427/#escaping-and-unicode
@@ -95,20 +95,26 @@ def _escape_filename_segment(segment):
     escaped = ""
     for character in segment.elems():
         # isalnum doesn't handle unicode characters properly.
-        if character.isalnum() or character == ".":
+        if character.isalnum() or character == "." or (stamp and (character in ["{", "}"])):
             escaped += character
         elif not escaped.endswith("_"):
             escaped += "_"
     return escaped
 
 def _py_wheel_impl(ctx):
-    outfile = ctx.actions.declare_file("-".join([
+    out_filename = "-".join([
         _escape_filename_segment(ctx.attr.distribution),
-        _escape_filename_segment(ctx.attr.version),
+        _escape_filename_segment(ctx.attr.version, ctx.attr.stamp),
         _escape_filename_segment(ctx.attr.python_tag),
         _escape_filename_segment(ctx.attr.abi),
         _escape_filename_segment(ctx.attr.platform),
-    ]) + ".whl")
+    ]) + ".whl"
+
+    if ctx.attr.stamp:
+        # this is to trick bazels output as we don't know the full file name
+        outfile = ctx.actions.declare_directory(out_filename)
+    else:
+        outfile = ctx.actions.declare_file(out_filename)
 
     inputs_to_package = depset(
         direct = ctx.files.deps,
@@ -134,6 +140,8 @@ def _py_wheel_impl(ctx):
     args.add("--abi", ctx.attr.abi)
     args.add("--platform", ctx.attr.platform)
     args.add("--out", outfile.path)
+    if ctx.attr.stamp:
+        args.add("--stamp_info", ctx.info_file.path)
     args.add_all(ctx.attr.strip_path_prefixes, format_each = "--strip_path_prefix=%s")
 
     args.add("--input_file_list", packageinputfile)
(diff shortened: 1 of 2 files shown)


python/packaging.bzl
```diff
@@ -310,6 +319,9 @@ _other_attrs = {
         ),
         default = "",
     ),
+    "stamp": attr.bool(
+        doc = "Enable stamping of the version attribute",
```
UebelAndre, line 323: Can you add an example of what this would resolve?
```suggestion
        doc = "Enable stamping of the version attribute. Stamp variables should be wrapped in brackets. Eg: `1.2.3-{BUILD_HOST}`",
```
Something like that

python/packaging.bzl
```diff
@@ -134,6 +140,8 @@ def _py_wheel_impl(ctx):
     args.add("--abi", ctx.attr.abi)
     args.add("--platform", ctx.attr.platform)
     args.add("--out", outfile.path)
+    if ctx.attr.stamp:
+        args.add("--stamp_info", ctx.info_file.path)
```
UebelAndre, line 144: Could this be `version_info` instead? I don't think the use of stamping should always trigger rebuilds and I feel like if folks want stable info, then a `version_file` attribute should probably be added allowing them to resolve it themselves.

UebelAndre (changes requested):
Just a bystander excited about the change 😅

mattem:
Thanks for the PR!

Generally stamping is controlled by passing the `--stamp` flag to bazel on the command line, rather than an attr on the rule itself. 
Perhaps we can introduce the same pattern used in rules_go and rules_nodejs and use a provider to handle the stamping context?

https://github.com/bazelbuild/rules_nodejs/blob/aa09b5796faaf2470b84246a8f00e6e27181c63b/internal/node/context.bzl#L14
https://github.com/bazelbuild/rules_nodejs/blob/aa09b5796faaf2470b84246a8f00e6e27181c63b/internal/BUILD.bazel#L46

This then gets attached as an attr on rules, and the impl function reads from the provider:
https://github.com/bazelbuild/rules_nodejs/blob/f6ffed591aad320778bd9d6f35cb860dbde8e7af/packages/esbuild/esbuild.bzl#L166

It also means there can be tests that override the stamping behaviour:
https://github.com/bazelbuild/rules_nodejs/blob/f6ffed591aad320778bd9d6f35cb860dbde8e7af/packages/esbuild/test/define/BUILD.bazel#L35

UebelAndre:
> Generally stamping is controlled by passing the `--stamp` flag to bazel on the command line, rather than an attr on the rule itself.

There's prior-art for target specific stamping without the `--stamp` flag.
https://docs.bazel.build/versions/4.0.0/user-manual.html#workspace_status
> Use these options to "stamp" Bazel-built binaries: to embed additional information into the binaries, such as the source control revision or other workspace-related information. You can use this mechanism with rules that support the `stamp` attribute, such as `genrule`, `cc_binary`, and more. 

I don't quite care what the implementation is as long as I can stamp wheels with volatile status info in such a way that changes to stable status info does not trigger a rebuild. Making `ctx.info_file` an input to the action would do this. Though, I do feel target level stamping feels more ergonomic.

python/packaging.bzl
tmspicer: I could switch it to use version_file instead.  My current scenario required it to use the stable versions but if the volatile versions would be better for the general public and not cause it to rebuild then I'm fine with changing it.

tmspicer:
> Perhaps we can introduce the same pattern used in rules_go and rules_nodejs and use a provider to handle the stamping context?

I'll take a look at switching to using a provider and the global stamp option.

python/packaging.bzl
UebelAndre: The only thing I care about is being able to stamp targets without triggering rebuilds. If stable status is required, then I'd rather have a utility to generate it separately from the wheel and somehow wire that up (eg, some `version_file` attribute which would contain the version to use).

tmspicer:
Updated code to implement check for the --stamp command line variable instead of a target specific attribute.
It was modeled and structured off of the `rules_nodejs` implementation

Issue #541, opened 2021-09-29
pip install example not working


XiaoConstantine:
# 🐞 bug report

### Affected Rule

 The issue is caused by the rule: 

### Is this a regression?

 Yes, the previous version in which this bug was not present was: ....

### Description

 A clear and concise description of the problem...

OS: MacOS Big Sur(11.6) Intel Chip
Bazel Version: 3.7.0

I just cloned the latest master and try to run through `examples` directory.

```bash
bazel run :main
```

Output:

```bash
/Users/xiao/development/rules_python/examples/pip_install/build-support/bazel/bin/bazel-3.7.0 run :main
ERROR: Traceback (most recent call last):
	File "/Users/xiao/development/rules_python/examples/pip_install/BUILD", line 9, column 40, in <toplevel>
		load("@rules_python//python:pip.bzl", "compile_pip_requirements")
Error: file '@rules_python//python:pip.bzl' does not contain symbol 'compile_pip_requirements'
```

## 🔬 Minimal Reproduction
1. Get the repo
2. Go to `rules_python/examples/pip_install`
3. `bazel run :main`

## 🔥 Exception or Error

<pre><code>

/Users/xiao/development/rules_python/examples/pip_install/build-support/bazel/bin/bazel-3.7.0 run :main
ERROR: Traceback (most recent call last):
	File "/Users/xiao/development/rules_python/examples/pip_install/BUILD", line 9, column 40, in <toplevel>
		load("@rules_python//python:pip.bzl", "compile_pip_requirements")
Error: file '@rules_python//python:pip.bzl' does not contain symbol 'compile_pip_requirements'

</code></pre>

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
OS: MacOS Big Sur(11.6) Intel Chip
  </code>
</pre>

**Output of `bazel version`:**

<pre>
  <code>
3.7.0
  </code>
</pre>

**Rules_python version:**

<pre>
  <code>
latest master
  </code>
</pre>

**Anything else relevant?**

thundergolfer:
Sorry about this. Reasonably often people land changes in the examples/ which aren't compatible with the example's rules_python version. We should be better about following up on these breakages. 

Latest `main` commit works, so I've set the example up with that in #547.

PR #547, opened 2021-10-12
fix: rules_python 0.4.0 is not recent enough for pip_install example

thundergolfer:
Addresses https://github.com/bazelbuild/rules_python/issues/541

with fix I can get `:main` to run. Output is: 

```
Target //:main up-to-date:
  bazel-bin/main
INFO: Elapsed time: 22.670s, Critical Path: 0.58s
INFO: 5 processes: 5 internal.
INFO: Build completed successfully, 5 total actions
INFO: Build completed successfully, 5 total actions
['DEFAULT_SESSION', 'NullHandler', 'Session', '__author__', ....
...
```

### Does this PR introduce a breaking change?

- [ ] Yes
- [X] No


alexeagle:
What's the principled answer? We've always had this situation in rules_nodejs and our answer has been to cut releases often enough that the skew is rarely noticed, but sometimes get a similar issue filed.

Obviously we could just use `local_repository(path="../..", name = "rules_python")` so that the example always works in the repo, but then it's not a standalone example that ppl can copy into their code.
(In any case we'd still want the integration test runner to replace that with our distro package, so we get test coverage for the distro rather than working against the sources.)

PR #573, opened 2021-11-18
Remove return code type check.

phst:
Only floating-point objects have an is_integer
method (cf. https://docs.python.org/3.10/library/stdtypes.html#typesnumeric),
so this will actually raise an AttributeError whenever the main function
returns an integer or None.  sys.exit works just fine with any argument type,
so there’s no need to check anything.  The example at
https://packaging.python.org/specifications/entry-points/#use-for-scripts also
doesn’t contain a type check.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/python/pip_install/extract_wheels/lib/bazel.py
+++ b/python/pip_install/extract_wheels/lib/bazel.py
@@ -35,8 +35,7 @@ def generate_entry_point_contents(entry_point: str, shebang: str = "#!/usr/bin/e
         from {module} import {method}
         if __name__ == "__main__":
             rc = {method}()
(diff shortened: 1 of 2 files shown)


thundergolfer (approved):
Happy with this change

https://github.com/python/cpython/blob/main/Python/sysmodule.c#L837 lays out the argument handling of `sys.exit` nicely. Non-integer is handled, and treated as exit(1).

Issue #574, opened 2021-11-25
Gazelle support is not possible with a different package index than the default


aignas:
# 🐞 bug report

Sorry about a bug report on the `HEAD` version of the `rules_python`, but wanted to share a few findings whilst trying to use it, since I've spent some time tracking this down.

### Affected Rule

Gazelle python manifest is affected.
```
gazelle_python_manifest(
    name = "gazelle_python_manifest",
    modules_mapping = "@modules_map//:modules_mapping.json",
    pip_deps_repository_name = "pip_deps",
    requirements = "//third_party/pip:requirements.txt",
)
```

And it is not a regresseion

### Description

Currently [the rule](https://github.com/bazelbuild/rules_python/blob/main/gazelle/manifest/defs.bzl#L7) does not allow to use a custom package index.

There are multiple things related to this:

#### 1. It is not possible to define an index other than the default one for the manifest generation

Since the previously linked bazel rule does not have sufficient arguments, the only way to get it sort of working is to use extra definitions in the requirements file as below:
```txt
#
# This file is autogenerated by pip-compile
# To update, run:
#
#    bazel run //third_party/pip:requirements.update
#
--extra-index-url https://my.internal.mirror.com

affine==2.3.0 \
    --hash=sha256:2e045def1aa29e613c42e801a7e10e0b9bacfed1a7c6af0cadf8843530a15102 \
    --hash=sha256:34b05b070d954c382e56f02c207a372d8a32621a87653cc30cdd31cd7f65799f
    # via rasterio
```

#### 2. modules_mapping cannot handle such requirements file

The parser breaks [here](https://github.com/bazelbuild/rules_python/blob/main/gazelle/modules_mapping/def.bzl#L168). It does not expect an extra index line to be present at all and breaks because the `version` attribute within the wheel `struct` is not present and this is a bit difficult to debug, because the starlark code assumes that the parsing is successful, but in reality it could break at any line.

#### 3. JSON information is fetched from PyPI only

If a package only exists in a third_party PyPI index, this will fail. I am not sure if this is a problem with the implementation, but for a user who may be using a custom package index, it would be good to know that this is a requirement.

## 🔬 Minimal Reproduction

Since I am not aware of any publicly available Python PI alternative, I am not sure I can create an example, but I do hope that the analysis helps.

## 🔥 Exception or Error

```
INFO: Repository modules_map instantiated at:
  <repo>/src/WORKSPACE:65:16: in <toplevel>
[...]

PR #578, opened 2021-12-09
refactor: make modules_mapping a regular rule

f0rmiga:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

## What has changes

The current `modules_mapping` is a repository rule. This changeset moves it to be a regular rule, consuming the wheels from pip_install instead of re-downloading and building (when necessary) them twice.

- [ ] Yes
- [x] No

Since the Gazelle plugin is not contained in any release, I don't consider it to be a breaking change.


gazelle/README.md
```diff
@@ -61,7 +57,7 @@ gazelle_python_manifest(
     name = "gazelle_python_manifest",
     # The @modules_map refers to the name we gave in the modules_mapping
     # rule in the WORKSPACE
```
kormide, line 59: Reference to the old repository name.

kormide:
Was there a reason why `modules_mapping` was originally a repository rule rather than a _rule_ rule?

f0rmiga:
@kormide In the early days, the modules mapping was considerably faster compared to depending on all the wheels from `pip_install`. It was important because we were generating it on the fly (not checked-in as part of the gazelle manifest like we do now). This is not true anymore.

gazelle/README.md
```diff
@@ -23,22 +20,12 @@ Add this to your `WORKSPACE`:
...
-)
 ```
 
-Next, we'll make a pair of targets for consuming that `modules_mapping` we
-fetched, and writing it as a manifest file for Gazelle to read.
+Next, we'll fetch metadata about your Python dependencies, so that gazelle can
+determine which package a given import statement comes from. This is provided
+by the `modules_mapping` rule. We'll make a target for consuming this
+`modules_mapping`, and writing it as a manifest file for Gazelle to read.
 This is checked into the repo for speed, as it takes some time to calculate
```
alexeagle, line 29: Is there still a file checked into the repo for speed?

alexeagle (approved):

gazelle/README.md
f0rmiga: Yes

gazelle/README.md
f0rmiga: Fixed in https://github.com/bazelbuild/rules_python/pull/578/commits/fa6a7cf2f3b11eed1a0e382a0c1823f2d7637cfe.

corypaik:
I was working on this a bit earlier since the previous version didn't work with `--find-links` and thought it could make sense to generate a `modules_mapping.json` for each wheel individually. I've refactored what I had to work with this rule and drafted a change that runs the generator once per wheel then merges the result. It seems to address the issue of having to recompute the module mapping for all dependencies each time one is updated. 

Not really sure of a good way to benchmark it beyond rough timings and checking for cached actions, but it seems like it should be more efficient. Could also combine the json merger with the yaml generator if having them separate adds too much overhead, but I wasn't sure if there was another use for the merged json file beyond making the yaml one. I pushed the changes [here](https://github.com/corypaik/rules_python/tree/simplify-modules-mapping).

f0rmiga:
@corypaik - interesting approach. We have been using it on a large project and haven't felt the need to do this performance improvement since the testing to check whether the checked-in manifest is up to date or not doesn't require recomputing the whole file.

The most expensive part of updating the modules mapping is downloading and, in some cases, compiling native modules. The tree traversal to generate the mappings is usually super fast and I don't think the performance improvement would be noticeable?

aignas:
I think making this a simple rule also fixed #574 because the same fetch packages can be now used here. So thanks for this change!

PR #581, opened 2021-12-23
Make spread_purelib_into_root behave like the wheel was installed by pip

hrfuller:
Merge purelib dir into toplevel even if purelib and toplevel have
subdirs with the same name. See tensorflow-io for an example of
a package which was not installed correctly by rules_python before this
change.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

In pip_install and pip_parse rule if you try to install a wheel which has Root-Is-Purelib: false (tensorflow-io is an example), python sources in the .data/purelib dir will not be merged into the root because we skip children of the purelib which have the same basename as a directory in the top level of the wheel.

What rules_python does currently: https://gist.github.com/hrfuller/ccc0420cd5b72b38f6151fa1b5c811e0

After running the example open up the created repo in the bazel output base, and notice that purelib directory is not merged.

This leads to issues where tensorflow_io python sources cannot be imported.

Issue Number: N/A

## What is the new behavior?

Descendants of the purelib directory will be merged with existing directories in the top level, but will not overwrite any files in the toplevel. This is the same thing that pip does when you install a wheel.

To verify this run

`pip install --no-deps --only-binary :all: --target tfio tensorflow-io==0.19.0`

Compare the shape of the tfio directory with the external repo bazel generated using the gist above.

- [ ] Yes
- [x] No

### DIFF
--- a/python/pip_install/extract_wheels/lib/purelib.py
+++ b/python/pip_install/extract_wheels/lib/purelib.py
@@ -1,4 +1,5 @@
 """Functions to make purelibs Bazel compatible"""
+import os
 import pathlib
 import shutil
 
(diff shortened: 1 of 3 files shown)


alexeagle (approved):
Thanks for the great writeup explaining the problem and approach :)

python/pip_install/extract_wheels/lib/purelib.py
```diff
@@ -34,20 +35,33 @@ def spread_purelib_into_root(wheel_dir: str) -> None:
...
-        # Some purelib Wheels, like Tensorflow 2.0.0, have directories
-        # split between the root and the purelib directory. In this case
-        # we should leave the purelib 'sibling' alone.
-        # See: https://github.com/dillon-giacoppo/rules_python_external/issues/8
-        if not pathlib.Path(root_dir, grandchild.name).exists():
-            shutil.move(
+    for child in purelib_dir.iterdir():
+        if child.is_dir():
+            backport_copytree(src=child, dst=pathlib.Path(root_dir, child.name))
+        elif not pathlib.Path(root_dir, grandchild.name).exists():
```
thundergolfer, line 63: I think `grandchild` -> `child` is a needed patch, to avoid `NameError`?

hrfuller: Yes, thanks for catching that. Will provide a patch.

Issue #436, opened 2021-03-19
Replace `subpar` with a cross-platform solution


UebelAndre:
# 🚀 feature request

### Description

 A clear and concise description of the problem or missing capability...

I opened this as a feature request because I otherwise didn't know how to raise open the issue

Right now the rules use [subpar](https://github.com/google/subpar) to generate some executable which are then committed back and reused for other rules:
https://github.com/bazelbuild/rules_python/blob/c37ba2215eccab53ae1da5f827a335281f81b9e1/packaging/BUILD#L45-L62

Is there a reason these couldn't be converted to using Bazel's `python_zip_file` functionality to accomplish the same thing? (introduced in https://github.com/bazelbuild/bazel/pull/9453)

UebelAndre:
If this is possible, would it solve for the outstanding issue on https://github.com/bazelbuild/rules_python/pull/344

UebelAndre:
is the generated zip (zipapp? [PEP-0441](https://www.python.org/dev/peps/pep-0441/)) cross platform? Does anyone know it's limitations there? I'm pretty surprised this is not a more advertised feature.

hrfuller:
The zipapp won't be cross-platform. Any rules_python pip dependencies are fetched only for the local platform. In some cases with universal wheels they *could* be cross-platform, but it isn't a guarantee.

UebelAndre:
Is there currently a cross platform implementation for self contained Python apps?

bshashank:
> The .zip won't be cross-platform. Any rules_python pip dependencies are fetched only for the local platform. In some cases with universal wheels they could be cross-platform, but it isn't a guarantee. Having said that and without in depth knowledge I don't think subpar's are truly cross-platform either.

I agree, subpar and `--build_python_zip` behave in a similar manner wrt cross-platform. The one advantage of subpar is that you can ask for the par to be unzipped before execution - which helps Python discover compiled code (`.so` or `.dlls`) more easily.

Since pip and wheel are both pure-python, I think both subpar and `--build_python_zip` would work in the same manner.

UebelAndre:
Except in that the zips are producible on windows and the pars are not, right?

hrfuller:
FWIW if cross-platform support is a necessity, which it may not be here, https://github.com/pantsbuild/pex has cross-platform support.

UebelAndre:
Nice! Are there Bazel rules for that?

[...]

PR #582, opened 2021-12-24
Deleted legacy pip_import rule

UebelAndre:
## PR Checklist

- [ ] Does not include precompiled binaries, eg. `.par` files. See [CONTRIBUTING.md](https://github.com/bazelbuild/rules_python/blob/master/CONTRIBUTING.md) for info

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

As described in https://github.com/bazelbuild/rules_python/issues/436#issuecomment-985699038, maybe it's time to clean out the old pip repository rules. In investigating cache invalidation issues. I've often been confused by what code path was being used since the two are quite similar.

Issue Number: https://github.com/bazelbuild/rules_python/issues/436#issuecomment-985699038

## What is the new behavior?
This change deletes some old rules that have long since been considered legacy and have modern replacements that are actively being developed.

- [x] Yes
- [ ] No


UebelAndre:
@alexeagle @hrfuller Based on the conversation at https://github.com/bazelbuild/rules_python/issues/436#issuecomment-985699038, I've opened this PR. I think the rules have been communicated as EOL for some time so just deleting them feels acceptable to me but if you're hoping for a more transitional closure this code then I'm happy to defer to a PR from either one of you 😄

.gitignore
```diff
@@ -37,6 +37,7 @@
 /bazel-genfiles
 /bazel-out
 /bazel-testlogs
+user.bazelrc
```
hrfuller, line 40: Is this a drive-by?

examples/pip_install/.bazelrc
```diff
@@ -1 +1,2 @@
-test --test_output=errors
+# https://docs.bazel.build/versions/main/best-practices.html#using-the-bazelrc-file
+try-import %workspace%/user.bazelrc
```
hrfuller, line 2: This change seems good but also unrelated to this PR. Should it be its own?

hrfuller (approved):
LGTM, modulo comment about PR hygiene.

examples/pip_install/.bazelrc
UebelAndre: Does it matter hugely if this is included? I was a way for me to test the changes since the examples are already a little difficult to run. I found this to be necessary in validating the changes were safe and would opt to leave it in.

.gitignore
UebelAndre: This is a recommended pattern and was used in testing this Pr. I'd recommend it stay if possible

UebelAndre:
@hrfuller i think the changes here should be merged as is. Do you feel super strongly otherwise?

tools/BUILD
```diff
@@ -23,12 +23,6 @@ py_binary(
     srcs = ["wheelmaker.py"],
 )
 
-# These files are generated and updated by ./update_tools.sh
-exports_files([
-    "piptool.par",
-    "whltool.par",
```
alexeagle, line 29: I bet there's also some docs or checks for these files we can remove

alexeagle (approved):
Thanks!

tools/BUILD
UebelAndre: You were right. I found the CONTRIBUTING docs and removed them.

alexeagle (approved):
ooh this also has the amazing side-effect of reducing the size of this repo enough that I think we can just ship the GH archive file rather than uploading our own release artifact

examples/pip_install/.bazelrc
hrfuller: It doesn't matter hugely. In cases where we have to revert functionality it can matter, because this is a feature we want regardless of this PR.

PR #587, opened 2021-12-29
Simplify release

### LINKED ISSUE #586: remove release artifact
Now that #582 landed, the github-served artifact is small

```
-rw-rw-r-- 1 alexeagle alexeagle 2.8M Dec 28 14:52 beed9a257cc493ebf5aa7b8e6bbe577fa637e8f6.zip
-rw-rw-r-- 1 alexeagle alexeagle 338K Dec 28 14:51 main.zip
```

so there's no need for rules_python to distribute a release artifact, we can just let GitHub serve a source archive.
This lets us use a much simpler release pattern from https://github.com/bazel-contrib/rules-template where maintainers just push a tag and the rest is automated.

alexeagle:
Instead of building our own .tgz file with release artifacts, just let GitHub do it.
Makes the release instructions much quicker, and doesn't require a review pass for releases, so we'll do them more frequently.

Note that GitHub does serve a stable SHA, and automated release notes include the workspace snippet with the SHA.
This produces an archive of 338K that users will fetch. Our previous release was 45K, so this is an order of magnitude larger since our test code, examples, and documentation are now included. However IMO it's not that important, compared with simpler maintenance.

This follows the new example from the rules authors SIG: https://github.com/bazel-contrib/rules-template

fixes #586


.github/workflows/release.yml
```diff
@@ -0,0 +1,27 @@
...
+          XDG_CACHE_HOME: ~/.cache/bazel-repo
+        run: bazel --bazelrc=.github/workflows/ci.bazelrc --bazelrc=.bazelrc test //...
+      - name: Prepare workspace snippet
+        run: .github/workflows/workspace_snippet.sh ${{ env.GITHUB_REF_NAME }} > release_notes.txt
+      - name: Release
+        uses: softprops/action-gh-release@v1
+        with:
+          # Use GH feature to populate the changelog automatically
+          generate_release_notes: true
+          body_path: release_notes.txt
```
hrfuller, line 27: Does writing to the releases_notes.txt here append to or overwrite the workspace snippet?

tools/bazel_integration_test/test_runner.py
```diff
@@ -58,17 +33,11 @@ def main(conf_file):
...
-        for command in config["bazelCommands"]:
-            bazel_args = command.split(" ")
-            try:
-                doubleHyphenPos = bazel_args.index("--")
-                print("patch that in ", doubleHyphenPos)
-            except ValueError:
-                pass
+        for command in config['bazelCommands']:
+            bazel_args = command.split(' ')
+            bazel_args.append("--override_repository=rules_python=%s/rules_python" % os.environ['TEST_SRCDIR'])
```
hrfuller, line 38: Very cool.

tools/bazel_integration_test/test_runner.py
```diff
@@ -58,17 +33,11 @@ def main(conf_file):
...
 
-            # Bazel's wrapper script needs this or you get
+            # Bazel's wrapper script needs this or you get
```
hrfuller, line 40: nit trailing ws.

hrfuller (approved):
This is great. Should we cut a release after this merges to make sure it works? :)

.github/workflows/release.yml
alexeagle: This line is reading from the release notes. Result is the thing you want (both snippet and description of changes)

hrfuller: I see. The comment about generate_release_notes: true is confusing because it seems like we are writing generated release notes into that file path.

alexeagle:
@hrfuller I'll push a tag to the repo now to verify.

Issue #571, opened 2021-11-16
Non-determinism in pip_parse or pip_import


alexeagle:
https://github.com/bazelbuild/rules_python/pull/551#issuecomment-970499748 reports that different machines running the pip operations result in different outputs, making cache busts on subsequent actions that depend on python requirements in their inputs.

aignas:
Funny enough I ran into this today and was wondering as to why the bazel-remote cache was not working as expected in our CI, but was working correctly in local setups.

taylorjohnson:
Hi. Is this addressed by https://github.com/bazelbuild/rules_python/pull/570 or is there additional work needed? Thanks!

UebelAndre:
There is still more work. At the very least, packages which compile C/C++ code are subject to non-deterministic targets. There's still one open question on that PR which may mitigate this just a bit more (https://github.com/bazelbuild/rules_python/pull/570#issuecomment-982143297) but ultimately, the `pip` rules need a way to allow users to express generally things that should be excluded from the generated targets and allow them to provide the same information from Bazel. Or, that's my own opinion. It'd be great to blanket ignore all `*.so` files and somehow annotate `pip_parse` to take a `cc_binary` at build time as a replacement. I'm not sure how that would be wired up though.

UebelAndre:
Though that said, if the packages you depend on are pure python packages (don't build any extra files when setting up the wheel) then I don't think you'll run into determinism issues. Though, if you're impacted by `*.pyc` files and `__pycache__` directories, odds are your packages are doing just that.

aignas:
How is this related to use cases when users are using binary wheels and even though we have `*.so` files present in the wheel, they are not built by a local `bazel` setup, but rather downloaded from PyPI from a know wheel.

groodt:
> How is this related to use cases when users are using binary wheels and even though we have `*.so` files present in the wheel, they are not built by a local `bazel` setup, but rather downloaded from PyPI from a know wheel.

Are you sure the entire transitive closure comprises of Wheels? 

It's often the case that there is an sdist somewhere in the transitive closure and this builds on every host machine. This can easily introduce non-determinism, since the building of dependencies in Python is essentially `eval setup.py`.

aignas:
> > How is this related to use cases when users are using binary wheels and even though we have `*.so` files present in the wheel, they are not built by a local `bazel` setup, but rather downloaded from PyPI from a know wheel.
> 
[...]

PR #589, opened 2021-12-29
Added support for annotating rendered pip dependencies

UebelAndre:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

This change adds an `annotations` field to `pip_parse` and `pip_install` which allow users to make changes to generated python targets. The goal of this functionality is to allow users to filter out any non-determinsitic behavior they may find in their dependencies to resolve remote cache issues.

Issue Number: https://github.com/bazelbuild/rules_python/issues/571

## What is the new behavior?

Below is an example of `package_annotation`. The rule allows us to add additional targets or set additional flags to packages like `wheel` in the example.

```python
load("@rules_python//python:pip.bzl", "package_annotation", "pip_parse")

pip_parse(
    name = "pip_parsed",
    annotations = {
        "wheel": package_annotation(
            build_content = """\
load("@bazel_skylib//rules:write_file.bzl", "write_file")
write_file(
    name = "generated_file",
    out = "generated_file.txt",
    content = ["Hello world from build content file"],
)
""",
            copy_executables = {"@pip_repository_annotations_example//:data/copy_executable.py": "copied_content/executable.py"},
            copy_files = {"@pip_repository_annotations_example//:data/copy_file.txt": "copied_content/file.txt"},
            data = [":generated_file"],
            data_exclude_glob = ["*.dist-info/RECORD"],
        ),
    },
)

```

The annotations here introduce the following changes to the `BUILD` file
```diff
diff --git a/BUILD.pip_parsed_wheel.bazel b/BUILD.pip_parsed_wheel.bazel
index 74dbeb0..8e3f483 100644
--- a/BUILD.pip_parsed_wheel.bazel
+++ b/BUILD.pip_parsed_wheel.bazel
@@ -22,7 +22,7 @@ filegroup(
 py_library(
     name = "pkg",
     srcs = glob(["**/*.py"], exclude=[], allow_empty = True),
-    data = [] + glob(["**/*"], exclude=["**/* *", "**/*.py", "**/*.pyc", "**/__pycache__/**", "*.whl", "BUILD.bazel", "WORKSPACE", "rules_python_wheel_entry_point*.py"]),
+    data = [":generated_file", "copied_content/executable.py", "copied_content/file.txt"] + glob(["**/*"], exclude=["**/* *", "**/*.py", "**/*.pyc", "**/__pycache__/**", "*.dist-info/RECORD", "*.whl", "BUILD.bazel", "WORKSPACE", "rules_python_wheel_entry_point*.py"]),
     # This makes this directory a top-level in the python import
     # search path for anything that depends on this.
     imports = ["."],
@@ -38,3 +38,24 @@ py_binary(
     imports = ["."],
     deps = ["pkg"],
 )
+
+copy_file(
+    name = "copied_content/file.txt.copy",
+    src = "@pip_repository_annotations_example//:data/copy_file.txt",
+    out = "copied_content/file.txt",
+    is_executable = False,
+)
+
+copy_file(
+    name = "copied_content/executable.py.copy",
+    src = "@pip_repository_annotations_example//:data/copy_executable.py",
+    out = "copied_content/executable.py",
+    is_executable = True,
+)
+
+load("@bazel_skylib//rules:write_file.bzl", "write_file")
+write_file(
+    name = "generated_file",
+    out = "generated_file.txt",
+    content = ["Hello world from build content file"],
+)
```

- [ ] Yes
- [x] No


python/pip_install/pip_repository.bzl
```diff
@@ -367,6 +392,40 @@ whl_library = repository_rule(
...
+        data (list, optional): A list of labels to add as `data` dependencies to the generated `py_library` target.
+        data_exclude_glob (list, optional): A list of exclude glob patterns to add as `data` to the generated
+            `py_library` target.
+        srcs_exclude_glob (list, optional): A list of labels to add as `srcs` to the generated `py_library` target.
+
+    Returns:
+        str: A json encoded string of the provided content.
+    """
+    return json.encode(struct(
+        build_content = build_content,
```
mattem, line 425: Could we also supply a label of a BUILD file in the repo here? Similar to `http_archive` (https://docs.bazel.build/versions/main/repo/http.html#http_archive-build_file) etc. It might be a nice way to completely override the BUILD file for a requirement.

python/pip_install/pip_repository.bzl
```diff
@@ -367,6 +392,40 @@ whl_library = repository_rule(
...
+        copy_executables = {},
+        data = [],
+        data_exclude_glob = [],
+        srcs_exclude_glob = []):
+    """Annotations to apply to the BUILD file content from package generated from a `pip_repository` rule.
+
+    [cf]: https://github.com/bazelbuild/bazel-skylib/blob/main/docs/copy_file_doc.md
+
+    Args:
+        build_content (str, optional): Raw text to add to the generated `BUILD` file of a package.
```
mattem, line 411: nit: might be nice to mirror the same attribute names as some builtin repo rules for `build_content` -> `build_file_content` https://docs.bazel.build/versions/main/repo/http.html#http_archive-build_file_content

docs/pip.md
```diff
@@ -35,6 +35,33 @@ It also generates two targets for running pip-compile:
 | kwargs |  other bazel attributes passed to the "_test" rule   |  none |
 
 
+<a name="#package_annotation"></a>
```
mattem, line 38: I think the docs here brush over some of the advanced use of this feature, or don't go into much detail about how a user might hold this. The example in the examples and mock dir is great, at a minimum this should link to it. Perhaps add an example of supplying a BUILD file for some cc_library, as that might be a common task here?

python/pip_install/pip_repository.bzl
UebelAndre: Ah, this is actually a misleading parameter. `build_content` is really `additive_build_content`. It's a way for us to add whatever we want to the generated BUILD file. Do you think there's a big use case for this? I have yet to come across it so I would opt out of adding it to this PR.

python/pip_install/pip_repository.bzl
UebelAndre: This isn't intended to be the whole build file. Just a small snippet to add to the bottom of the generated one. It's been renamed to `additive_build_content`.

docs/pip.md
UebelAndre: I can do that. Can you suggest a package to use for the example? Hopefully something that doesn't require compiling something massive 😅

python/pip_install/pip_repository.bzl
mattem: If you haven't hit it then don't add it, we can always add it on later. Some folks may like the content in a file rather than inline however, but it's not a blocker imo.

docs/pip.md
mattem: Heh, I don't have a good one off the top of my head, there's normally some hidden complexity.

mattem (approved):

docs/pip.md
UebelAndre: Agreed. If someone can suggest one then I can try to add it to docs later. But I think the demo shows the general functionality of each annotation one could apply.

UebelAndre:
For better visibility, one issue came out of this PR: https://github.com/bazelbuild/rules_python/commit/7740b22d0bae942af0797967f2617daa19834cb3#r64135831

Issue #414, opened 2021-02-07
Allow accessing requirements directly by labels


person142:
# 🚀 feature request

### Relevant Rules

`py_binary`, `py_library`, and `py_test`

### Description

Per the README accessing requirements by labels is not supported:

> To avoid breakage, please use requirement() instead of depending directly on wheel repo labels.

The requirements also have somewhat unintuitive labels since they prepend `pypi__` onto the (sanitized) package name. Not working with labels directly has some consequences:

- It doesn't work with buildifier (https://github.com/bazelbuild/buildtools/issues/955)
- It doesn't work with buildozer (https://github.com/bazelbuild/buildtools/issues/565#issuecomment-645556864)
- In the next release it will be hard to work with `all_whl_requirements` without digging into the requirements workspace, so the abstraction is leaky. (Though uncommon, I find I occasionally need to dig through the requirements workspace in some other settings too.)
- It's fairly obfuscated; i.e. it adds an extra layer on top of the normal mechanisms for referring to targets from external workspaces
- It doesn't scale well-there used to be `requirement`, now there's `whl_requirement`, if you started wrapping console scripts then there would need to be a new function for that, etc. etc.

### Describe the solution you'd like

Is it possible to:

- Shorten the labels so that it's `@<workspace>//<sanitized-package-name>` (For the short form of the label.)
- Officially support working with the labels directly

### Describe alternatives you've considered

We don't enforce that Python deps be sorted. When I need to buildozer a requirement I do some hacky find-replace to stringify it and then unstringify it.

person142:
Ok, so the reason for the status quo is mentioned in the wheel extracting code:

> Further, rules-python automatically adds the repository root to the PYTHONPATH, meaning a package that has the same name as a module is picked up. We workaround this by prefixing with `pypi__`. Alternatively we could require `--noexperimental_python_import_all_repositories` be set, however this breaks rules_docker. See: https://github.com/bazelbuild/bazel/issues/2636

This happens inside the shim for launching a `py_binary`:

```python
# Returns repository roots to add to the import path.
def GetRepositoriesImports(module_space, import_all):
  if import_all:
    repo_dirs = [os.path.join(module_space, d) for d in os.listdir(module_space)]
    repo_dirs.sort()
    return [d for d in repo_dirs if os.path.isdir(d)]
  return [os.path.join(module_space, '__main__')]
```

[...]

PR #594, opened 2022-01-03
Update README to say that using `requirement()` is optional

person142:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: https://github.com/bazelbuild/rules_python/issues/414

Currently the documentation recommends only using `requirement()` to access the `py_library` targets generated by `pip_install` and `pip_parse`.

## What is the new behavior?
Update the documentation to specify that using `requirement()` is optional, and that accessing the labels directly is ok.

- [ ] Yes
- [x] No

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,7 +124,7 @@ Bazel will only fetch/build wheels for the requirements in the subgraph of your 
 
 There are API differences between `pip_parse` and `pip_install`:
 1. `pip_parse` requires a fully resolved lock file of your python dependencies. You can generate this by using the `compile_pip_requirements` rule,
-   running `pip-compile` directly, or using virtualenv and `pip freeze`. `pip_parse` uses a label argument called `requirements_lock` instead of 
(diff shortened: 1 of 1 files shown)


alexeagle (approved):
Awesome thank you for picking up that TODO :)

README.md
```diff
@@ -169,12 +168,37 @@ py_library(
...
+```
+
+Here `name` is the `name` attribute that was passed to `pip_parse` and
+`package` is the pip package name with characters that are illegal in
+Bazel label names (e.g. `-`, `.`) replaced with `_`. If you need to
+update `name` from "old" to "new", then you can run the following
+buildozer command:
+
+```
+buildozer 'substitute deps @old_([^/]+)//:pkg @new_${1}//:pkg' //...
```
aignas, line 192: This does not work as the target list needs to be specified as `//...:*` for all of the targets.

hrfuller: good point @aignas

person142: Derp, I ran this locally to test but copy pasted wrong apparently...

PR #601, opened 2022-01-09
changes to pip_repository source files now re-trigger the repo rule

UebelAndre:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

`pip_repository` rules currently only track the root of `rules_python` in order to determine the location of the necessary python files for rendering BUILD files for python packages. Because only the BUILD file is tracked, the repository rule is never made aware of the python sources. This results in situations where the `@rules_python//python/pip_install` source code may be changed but existing `pip_repository` rules will not be updated. 

Issue Number: N/A

## What is the new behavior?
This change generates a list of source files and passes them to the `pip_repository` rules so it can react to changes made. Now if a change is made to the python code, the repository rule will correctly be re-triggerd.

- [ ] Yes
- [x] No

### DIFF
--- a/python/pip_install/pip_repository.bzl
+++ b/python/pip_install/pip_repository.bzl
@@ -1,6 +1,7 @@
 ""
 
 load("//python/pip_install:repositories.bzl", "all_requirements")
+load("//python/pip_install/private:srcs.bzl", "PIP_INSTALL_PY_SRCS")
 
 def _construct_pypath(rctx):
     """Helper function to construct a PYTHONPATH.
@@ -111,6 +112,18 @@ def _parse_optional_attrs(rctx, args):
 
     return args
 
+def _detect_changes(rctx):
+    """Ensure sources used by the `pip_repository` rule are tracked as inputs.
+
+    This function ensures that if any source file used by the rule changes,
+    that the repository rule is re-triggered.
+
+    Args:
+        rctx (repository_ctx): The repository rule's context object
+    """
+    for src in rctx.attr._py_srcs:
+        rctx.path(src)
+
 _BUILD_FILE_CONTENTS = """\
 package(default_visibility = ["//visibility:public"])
 
@@ -126,6 +139,8 @@ def _pip_repository_impl(rctx):
 
     # We need a BUILD file to load the generated requirements.bzl
     rctx.file("BUILD.bazel", _BUILD_FILE_CONTENTS)
+
+    _detect_changes(rctx)
 
     pypath = _construct_pypath(rctx)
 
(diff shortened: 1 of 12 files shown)


python/pip_install/private/pip_install_utils.bzl
```diff
@@ -0,0 +1,114 @@
...
+    )
+
+    return DefaultInfo(
+        files = depset([output]),
+        runfiles = ctx.runfiles(files = [target_file]),
+        executable = output,
+    )
+
+_srcs_installer = rule(
+    doc = "A rule for writing a file back to the repository",
```
hrfuller, line 76: Can you also clarify that this rule generates the srcs.bzl file here?

python/pip_install/private/srcs.bzl
```diff
@@ -0,0 +1,19 @@
+"""A generate file containing all source files used for `@rules_python//python/pip_install:pip_repository.bzl` rules"""
```
hrfuller, line 1: It would be good to mention how to regenerate this file, and when that is needed.

python/pip_install/private/test/BUILD
```diff
@@ -0,0 +1,14 @@
+load("@bazel_skylib//rules:diff_test.bzl", "diff_test")
+
+diff_test(
+    name = "srcs_diff_test",
+    failure_message = "Please run `bazel run @rules_rust//python/pip_install/private:srcs_module.install`",
```
hrfuller, line 5: s/rules_rust/rules_python/

hrfuller, line 5: You should provide a description of the side effects of running that command as well.

python/pip_install/pip_repository.bzl
```diff
@@ -111,6 +112,18 @@ def _parse_optional_attrs(rctx, args):
 
     return args
 
+def _detect_changes(rctx):
```
hrfuller, line 115: I don't think this is strictly necessary. Just having a label list of the source files in the attrs of the repo rule will add the labels to the repository marker file.

hrfuller (commented):
Thanks much! This has been an outstanding broken window for a while. It generally only effects rules developers, which is why it hasn't been addressed yet.

python/pip_install/private/test/BUILD
hrfuller: In addition I think we should assume anyone re-running this rule will be a rule developer working inside rules_python so it should be `bazel run //python/pip_install/private:srcs_module.<verb>`

python/pip_install/private/pip_install_utils.bzl
UebelAndre: "install" communicates that something is being put into place where "generate" just means something is being created. I ultimately don't care a ton which it is but I'd opt for "install" unless you feel strongly about it. Just let me know.

UebelAndre: I've changed this to `update` to match the existing pattern set by `compile_pip_requirements`.

hrfuller: I like that more than install. Thanks.

hrfuller (approved):

PR #605, opened 2022-01-19
restrict visibility on filegroup target only consumed in 1 place

thundergolfer:
Reviewd #601 after it merged. Thanks for fixing this long outstanding annoyance 🙂  @UebelAndre 

If I understand this code correctly the filegroup is only consumed by `//python/pip_install/private:srcs_module` and so you don't need public visibility on the target.

### DIFF
--- a/python/pip_install/BUILD
+++ b/python/pip_install/BUILD
@@ -26,7 +26,7 @@ filegroup(
         "//python/pip_install/extract_wheels:py_srcs",
         "//python/pip_install/parse_requirements_to_bzl:py_srcs",
     ],
-    visibility = ["//visibility:public"],
+    visibility = ["//:__subpackages__"],
 )
 
 exports_files(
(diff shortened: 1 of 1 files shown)


UebelAndre:
Thanks for the extra review and sorry about the miss. I'll be sure to pay more attention to visibility going forward 😀

thundergolfer:
@UebelAndre no worries at all, and don't worry about it. It's such a minor thing 🙂

python/pip_install/BUILD
```diff
@@ -26,7 +26,7 @@ filegroup(
         "//python/pip_install/extract_wheels:py_srcs",
         "//python/pip_install/parse_requirements_to_bzl:py_srcs",
     ],
-    visibility = ["//visibility:public"],
+    visibility = ["//:__subpackages__"],
```
f0rmiga, line 29:
```suggestion
    visibility = ["//python/pip_install/private:__pkg__"],
```

f0rmiga (approved):
Since it's only used by `//python/pip_install/private:srcs_module`, as stated in the PR description, we could be even more diligent.

I'll leave it up to you to take the suggestion or not.

PR #606, opened 2022-01-19
Use alias for `copy_file` in `pip_repository` rules

UebelAndre:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

## What is the new behavior?
This is a minor cleanup to help users easily identify what rules or macros are actually in use.

- [ ] Yes
- [x] No


python/pip_install/extract_wheels/lib/bazel.py
```diff
@@ -97,8 +97,12 @@ def generate_copy_commands(src, dest, is_executable=False) -> str:
         str: A `copy_file` instantiation.
     """
     return textwrap.dedent(
+        # Here we use the aliased rule `rules_python_copy_file` to avoid
```
mattem, line 100: If this is in a generated file, when does this conflict happen? I think this makes it more confusing tbh about what this rule is, the load site (`rules_python`) makes it pretty clear that this is vendored, and not via `bazel_skylib`?

UebelAndre: It wouldn't "conflict". If users write their own "copy_file" rule and load it in this file, I'd like to make the vendored rule more obvious since. This is should have no change in behavior but was a minor thing that came up in a project of mine where a new Bazel user got confused.

thundergolfer (approved):
Fine with this, but I think `_copy_file = "copy_file"` is the more typical way of accomplishing the same thing?

UebelAndre:
> Fine with this, but I think `_copy_file = "copy_file"` is the more typical way of accomplishing the same thing?

My goal was to make it obvious that the already loaded `copy_file` rule was unique to `rules_python`. If you feel strongly I can rename it but do feel `ules_python_` stands out a bit more than `_`. Just LMK.

PR #612, opened 2022-01-28
Fix recursive use of `maybe` in `pip_repository` rules.

UebelAndre:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Users trying to wrap `pip_parse` or `pip_install` with [maybe](https://docs.bazel.build/versions/4.0.0/repo/utils.html#maybe) will run into a failure for attempting to use the rule recursively. This can be fixed by avoiding `maybe` in these macro repository rules.

Issue Number: N/A

## What is the new behavior?
[maybe](https://docs.bazel.build/versions/4.0.0/repo/utils.html#maybe) can now be used with `pip_parse` and `pip_install`.

- [ ] Yes
- [ ] No


python/pip_install/repositories.bzl
```diff
@@ -74,11 +73,11 @@ def pip_install_dependencies():
...
     for (name, url, sha256) in _RULE_DEPS:
-        maybe(
-            http_archive,
-            name,
-            url = url,
-            sha256 = sha256,
-            type = "zip",
-            build_file_content = _GENERIC_WHEEL,
-        )
+        if not native.existing_rule(name):
```
hrfuller, line 76: FWIW bazel recommends not using native.existing_rule.
>If possible, avoid using this function. It makes BUILD files brittle and order-dependent.

thundergolfer:
So you're attempting to do:

```python
maybe(
   pip_install,
   "<name>",
   **kwargs,
 )
```

is that right? What's the use case?

python/pip_install/repositories.bzl
thundergolfer: Also

> [Note that the existing_rule(s) functions are poorly defined, have design issues, can lead to performance and maintenance issues. So we don't really want to encourage their use.](https://github.com/bazelbuild/bazel/issues/7496)


It seems `rules_go` has for many years used `native.existing_rules` to have a 'fake `maybe`' https://github.com/bazelbuild/rules_go/pull/782/files.

UebelAndre:
> So you're attempting to do:
> 
> ```python
> maybe(
>    pip_install,
>    "<name>",
>    **kwargs,
>  )
> ```
> 
> is that right? What's the use case?

I have a repository I'm sharing between different workspaces which uses python. In each workspace loading it, I want the `pip_parse` repository which produces the dependencies to be used in the shared repository. There's a macro which loads all dependencies including the `pip_parse` repository and `pip_parse` is the only one that can't be wrapped via `maybe`. I think that it should be able to be.

python/pip_install/repositories.bzl
UebelAndre: I think that note is describing common issues in WORKSPACE management. I don't understand why `maybe` would be ok but not the native functions. One argument might be because `maybe` can't be used recursively but I think this situation is notably different in that users should be able to treat the `pip_repository` rules as normal repository rules, but there's some macro magic involved to inject the necessary dependencies. I think the use of these native functions here is fine but I'm also happy to learn better practices 😄

UebelAndre:
@thundergolfer @hrfuller friendly ping 😄

UebelAndre:
@thundergolfer @hrfuller another ping, would love to get this in 🙏

thundergolfer:
Still feel uneasy about this. 

I went on another code search to learn about the use of `native.existing_rule*` in the wild and here's you ripping out `native.existing

thundergolfer:
> Can you elaborate on...

It's mostly just https://github.com/bazelbuild/rules_python/pull/612#discussion_r796505691

Unless we're solving a critical case, opting into functionality the Bazel devs have explicitly advised against seems bad practice. So while you do have a valid case, solving for it might be 'one step forward two steps back'

If you opted out of the `pip_parse` macro and used `pip_repository` directly, wouldn't you have things working?

UebelAndre:
Thinking more about `pip_repository`. I think the thing I ultimately want is for `rules_python` to provide a `rules_python_dependencies` macro in accordance with [bazel-contrib](https://github.com/bazel-contrib/rules-template/blob/f26177bcbdd9eeab373f23ca568c98704efddd3f/mylang/repositories.bzl) and https://docs.bazel.build/versions/main/skylark/deploying.html instead of trying to hide them in macros. That way `pip_parse` and `pip_install` could be updated to be regular repository rules. But if that's not desirable, then I think this change is an acceptable middle ground.

alexeagle:
As one mitigation here, in #531 I'm adding stardoc for the `pip_repository` rule so it's better documented that you don't have to use the two wrapper macros.

UebelAndre:
> As one mitigation here, in #531 I'm adding stardoc for the `pip_repository` rule so it's better documented that you don't have to use the two wrapper macros.

@alexeagle I think this change is still useful. I commented on that PR: https://github.com/bazelbuild/rules_python/pull/531#issuecomment-1106734325

PR #616, opened 2022-02-01
chore: bump local version of bazel used to 4.2.0

mattem:
This bumps the version of bazel used locally when building to 4.2.0. This version is the first bazel version in 4.x that supports M1 correctly.
This bump doesn't change what is expected from users. However, should it?


alexeagle:
Let's go to 4.2.2 and also update version.bzl ?

thundergolfer:
@mattem 

> However, should it? 

Yeh the `version.bzl` says as much, originally written by Alex 🙂 

```
# This version should be updated together with the version of the Bazel
# in .bazelversion.
```

mattem:
Yeah, I'm familiar with the semantics (and the typo) from rnj, my question was more do we want to now bump the minimum supported version for rules_python to 4.2.2, or do we want to test against an array of versions (which will be slower on CI here). Sounds like the consensus is to bump it.

f0rmiga (approved):

Issue #293, opened 2019-02-19
Ship a repo rule to fetch a hermetic Python interpreter and generate a py_runtime / toolchain for it


brandjon:
Tracking issue for the feature request proposed [here](https://github.com/bazelbuild/bazel/issues/4815#issuecomment-464419291).

brandjon:
See also bazelbuild/bazel#8265, which can be reopened in this repo.

ddlees:
@brandjon I've been posting to GH Issues like this w/ a tiny project I put together that might fill the gap here. It should help some people at least while there isn't a native solution in bazel or in these rules. Anyway, I hope you guys have time to take a gander and provide some feedback if you think it's warranted. Cheers!

https://github.com/digital-plumbers-union/rules_pyenv

brandjon:
Thanks for the contribution! I'll leave a decision on the approach for this issue to the new maintainers, @thundergolfer and @andyscott.

thundergolfer:
@f0rmiga let's organize in this issue. 

👀  also say this recent issue comment: https://github.com/bazelbuild/rules_python/issues/42#issuecomment-1001222292

f0rmiga:
@thundergolfer Take a look at https://github.com/bazelbuild/rules_python/pull/618.

PR #618, opened 2022-02-03
feat: cpython toolchains for linux and macos

f0rmiga:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

- [ ] Yes
- [x] No

## Other information

Fixes https://github.com/bazelbuild/rules_python/issues/293.


UebelAndre:
Would it be possible to add python 3.7 if possible since it still appears to be within it's lifespan https://www.python.org/dev/peps/pep-0537/ ?

UebelAndre:
It also seems there's Windows binaries available? Could those be added as well?

UebelAndre:
Also, it'd be awesome to have the binaries added to the google mirror

f0rmiga:
@UebelAndre

- Python 3.7 is only published for 3.7.9 in https://github.com/indygreg/python-build-standalone/releases/tag/20200822. I can see if I can get a more up-to-date patch built.
- I don't have a Windows machine to test. I guess I can rely on CI for this. Still, I'd prefer to do it in a follow up PR.
- Agree. I wonder if we should even go one step further and start building the binaries in our infrastructure.

sikko-mode-25:
thanks for doing this!  one thing i wanted to note - I tried using these toolchains on my machine, and found they don't work with `--build_python_zip`, it would be really nice if they did.

thundergolfer:
> and found they don't work with --build_python_zip, it would be really nice if they did.

@sikko-mode-25 would be good to get details of "don't work". What's the error?

python/private/versions.bzl
```diff
@@ -0,0 +1,46 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""The Python versions we use for the toolchains.
+"""
+
+RELEASE_URL = "https://github.com/indygreg/python-build-standalone/releases/download/20211017"
```
thundergolfer, line 18: Have we established a line of comms with Gregory? I've been a very minor Github sponsor of him for a while, and have Twitter DM'd him, but taking this dependency does feel reminiscent of this XKCD comic: https://xkcd.com/2347/

I think at the least I'd like to get more hands on with this repo, and will message you on Slack in case I run into any speedbumps trying to compile these interpreters myself.

python/repositories.bzl
```diff
@@ -1,8 +1,164 @@
...
+    url = "{release_url}/{release_filename}".format(
+        release_url = RELEASE_URL,
+        release_filename = release_filename,
+    )
+    rctx.download(
+        url = url,
+        integrity = TOOL_VERSIONS[version][rctx.attr.platform],
+        output = release_filename,
+    )
+    unzstd = rctx.which("unzstd")
```
thundergolfer, line 54: This is the bit that doesn't feel production-ready. We either inspect the host system for the needed binary, or download the source and compile it on the host. (Do Windows machines have `make` available by default?)

So if we don't want to adopt this non-hermeticity, what options do we have?

* Try and land non-zstd tarball distributions in the python-standalone repo. Bazel natively supports regular tarballs.
    *  Do we know if Gregory is against using it? 
* Get zstd support landed in Bazel. https://github.com/bazelbuild/bazel/pull/11968 is an old PR, but there's been activity from Googlers on it in the last 18 days (mid-Jan 2022) and it seems like it might be mergeable?

thundergolfer (commented):
I think we should be sure to support Windows if we're gonna land this. I'm in progress on downloading a Windows VM (when will Github Codespaces support the Windows platform, eugh).

schultetwin:
Very cool. I use `rules_python` regularly and would love to see this.

As an FYI, I've (tried to) set-up a similar system and wanted to list a few cases here that we ran in to.

1. These standalone versions of python still have system dependencies. See here for a full list. https://python-build-standalone.readthedocs.io/en/latest/running.html#runtime-requirements. In particular, on Linux you need `libdl.so.2` in-order to use `ctypes`. I ended up with an auto-detection system used in the `repository_rule` that looks a bit like this:

```
       res = repository_ctx.execute([path_to_new_python_interpreter, "-c", "import ctypes; ctypes.CDLL('libdl.so.2')"])
        if res.return_code:
            fail("Failed to find libdl.so.2 installed on system.")
```

I'm sure other options exist too (including just documenting this case), but wanted to highlight it.

2. On macOS, there is some dependency on XCode version that I haven't tracked down yet. When trying to compile c extensions in pip packages, if the wrong version of XCode is installed, we get an error like:
```
 Compiling with an SDK that doesn't seem to exist: /Applications/Xcode_12.4.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.1.sdk
```
This one I haven't put in a solution for yet. But my next step is to look at the auto-configured x-code rules and see if I can detect if the proper version of Xcode is installed in the repository rule. Again, other options likely exist too, and maybe this is only worth documenting.

3. Finally, it's worth linking to the behavior quirks page: https://python-build-standalone.readthedocs.io/en/latest/quirks.html. This lists the various gotchas that may occur when using this version of python.

Not asking for any changes in particular in this PR, but hope the above context is useful for gotchas with a similar system!

python/private/versions.bzl
f0rmiga: I have not interacted with Gregory other than fixing a couple Make targets his repository uses to build OpenSSL.
Feel free to reach me on Slack.

python/repositories.bzl
f0rmiga: This is no different than using `gzip` with `tar`. I don't see anyone worried about which version of `gzip` the system is using to decompress a `.tar.gz` file (at least not at this level).

I.e. if `unzstd` exits code 0, it successfully extracted the software we want - and that is deterministic. For Windows we can follow a similar approach to compiling it without Make if it's not supported on that system.

f0rmiga:
@schultetwin thanks for the points!

1. I think we can just document this behaviour, which is standard when loading dynamic objects at runtime.
2. I'm not versed in XCode, but this sounds like something fixable from the user side? Probably downloading an extra SDK?
3. I think it's worth extracting the behaviour quirks that apply to the binaries we picked. Not all of those quirks apply.

python/repositories.bzl
```diff
@@ -1,8 +1,164 @@
...
+        fail(exec_result.stderr)
+
+    build_content = """\
+# Generated by python/repositories.bzl
+
+load("@bazel_tools//tools/python:toolchain.bzl", "py_runtime_pair")
+
+filegroup(
+    name = "files",
+    srcs = glob(["bin/**", "lib/**", "include/**", "share/**"]),
```
aignas, line 89: Be sure to exclude `__pycache__` and `*.pyc` files as that may introduce cache busts in production systems.

python/private/versions.bzl
```diff
@@ -0,0 +1,46 @@
...
+        "x86_64-apple-darwin": "sha384-Ix7lxb+niRWcDCOsK1q53dAkOp0CIw40kM8wHv//WQs2O/il8SRJ60cP3R67fEnm",
+        "aarch64-apple-darwin": "sha384-JSi3XZdHVZ+6DMybPYKgEYSQuBPpjXgBlj1uGqB9f/r3Wi6P0+CnYRG12TEzgcs6",
+        "x86_64-unknown-linux-gnu": "sha384-+PNcmKXJ+ZiyKyZ2WR1XedDXJ05ujC2w9TdXl80vloYMqfIOpcVPOWUgre+btI+3",
+    },
+    "3.10.0": {
+        "x86_64-apple-darwin": "sha384-eVTq704heZyKW6SB/DzarWB9S5bRH1LTQ4rFHzlKTE9jDjHDCBKyQhSgYy8a62lt",
+        "aarch64-apple-darwin": "sha384-NbhxnZL0pBTKpzEfoCYWl6s2GYdfiI9HOSSHn1iCMZnIY6htt/KhzjwIgCP+Nj2u",
+        "x86_64-unknown-linux-gnu": "sha384-iYJF9Y9o2Ain3YkuuF7ZGrGuJ+MyiD/xnrjJSap0TF2DR+I9XDx4stunNgx17gSn",
+    },
+}
```
aignas, line 39: Do we have to adopt this path of hardcoding shas in here? Could we accept a parameters to the repository rule as a parameter?
* python version
* a dict where key is platform and the value is sha256.

python/repositories.bzl
```diff
@@ -1,8 +1,164 @@
...
+    python_version = "PY3",
+)
+
+py_runtime_pair(
+    name = "python_runtimes",
+    py2_runtime = ":py2_runtime",
+    py3_runtime = ":py3_runtime",
+)
+"""
+    rctx.file("BUILD.bazel", build_content)
```
aignas, line 112: Consider returning here the values of parameters to the bazel rule, which would make the rule reproducible. See https://docs.bazel.build/versions/main/skylark/repository_rules.html#implementation-function

For example:
```starlark
    shas = dict(**repository_ctx.attr.sha256)
    shas[arch] = sha256 # put in the sha256 that we get when we download the interpreter
    return {
        "name": repository_ctx.attr.name,
        "sha256": shas,
        "timestamp": repository_ctx.attr.timestamp,
        "version": repository_ctx.attr.version,
    }
```

This assumes that the repository rule accepts:
* `name` for the repository.
* `sha256` dict where we have a mapping from `platform` to `sha256` of the downloaded artifact.
* `timestamp` is the timestamp value in the download url
* `version` the python version to download.

For example, the rule can be executed with the following params:
```starlark
        name = "python_interpreter",
        sha256 = {
            "x86_64-apple": "12c5bce1b48d2b896049ec2524648d5429eba982539bc571d43c1a2ec3997630",
            "x86_64-linux": "33cb6a4895418b9de2b770f8ab72d1fd2dbebb95747a81f6d1824c1900062df8",
        },
        timestamp = "20211017T1616",
        version = "3.9.7",
```

If the sha256 is missing from the dict (e.g. someone wants to add a download for ARM apple), then `bazel` would output a warning on how to change the input parameters.

python/repositories.bzl
```diff
@@ -1,8 +1,164 @@
...
+        version = version,
+        platform = rctx.attr.platform,
+    )
+    url = "{release_url}/{release_filename}".format(
+        release_url = RELEASE_URL,
+        release_filename = release_filename,
+    )
+    rctx.download(
+        url = url,
+        integrity = TOOL_VERSIONS[version][rctx.attr.platform],
```
aignas, line 51: If sha is an empty string, bazel will emit a warning, which the user can chose to act upon. This way we can avoid hardcoding shas of the releases and we can avoid coupling the releases of `indygreg` Python interpreter and `rules_python`.

python/repositories.bzl
```diff
@@ -1,8 +1,164 @@
...
+                name = name,
+                platform = platform,
+            ),
+            platform = platform,
+            **kwargs
+        )
+        native.register_toolchains("@{name}_toolchains//:{platform}_toolchain".format(
+            name = name,
+            platform = platform,
+        ))
```
aignas, line 159: How does this work with `pip_parse`. As far as I remember, it is impossible to pass something with a `select`. Would having a single python_repository be preferable?

python/repositories.bzl
sluongng: just FYI internally we compile a zstd cc_binary that we use to extract the python toolchain http download.

So making zstd binary configurable is very much desirable.

Let me know if you are interested in the BUILD file for zstd

python/private/versions.bzl
UebelAndre: I think accepting a parameter would be nice but I would like there to be some hard-coded fallbacks.

josh-newman:
Coincidentally, we recently started using indygreg/python-build-standalone's releases to define our Python toolchain (in @grailbio's internal repository). We hadn't yet had the time to make a PR here, so I'm excited to see this. Thanks @f0rmiga!

I've also experimented with using the artifacts from that standalone release to build self-contained, single-file Python executables (like .par, but also including the interpreter + stdlib). I've extracted a minimal example of this in [grailbio/py_singlefile_binary](https://github.com/grailbio/py_singlefile_binary); it defines a very slightly modified Python main, links with libpython, and uses itself as the import path because it is also a PEP 441 .zip.

Would you be open to expanding file visibility in the standalone Python repo's BUILD, to `//visibility:public`? Or even defining a libpython `cc_library` ([example from our prototype](https://github.com/grailbio/py_singlefile_binary/blob/462f05fc4b8a60cb3b0c4cf01cd44d9879fae8d9/third_party/python/cpython/BUILD.cpython#L24)) and stdlib.zip package there, for use cases like ours?

I'm also curious if there's interest in rules_python supporting self-contained executables in the future. Our interpreter+.zip approach works so far for some internal tools, though there may be better ways to accomplish this. @sluongng mentioned github.com/indygreg/PyOxidizer to me, for example. (Of course that'd be separate PRs/issues, I just thought I'd mention it here where there's context.)

python/repositories.bzl
```diff
@@ -1,8 +1,164 @@
...
+    version = rctx.attr.python_version
+    if version in MINOR_MAPPING:
+        version = MINOR_MAPPING[version]
+    release_filename = "cpython-{version}-{platform}-pgo+lto-{release_date}.tar.zst".format(
+        release_date = RELEASE_DATE,
+        version = version,
+        platform = rctx.attr.platform,
+    )
+    url = "{release_url}/{release_filename}".format(
+        release_url = RELEASE_URL,
```
lummax, line 46: Can this be made configurable? To allow to pass in a patched version of the standalone pythons.

I guess the same applies to the `_zstd` attributes.

python/repositories.bzl
f0rmiga: Not sure I follow why the pre-compiled modules would introduce cache busts? Could you expand, please?

python/repositories.bzl
f0rmiga: Addressed in https://github.com/bazelbuild/rules_python/pull/618/commits/0d84ae6114368208704eaa26bf5b888247118221.

python/repositories.bzl
f0rmiga: No, the native toolchain system cannot be used in repository rules. We have to pass the interpreter explicitly to pip_parse/pip_install.

python/repositories.bzl
sluongng: https://github.com/grailbio/py_singlefile_binary/tree/main/third_party%2Fzstd
And here is how we put it together https://github.com/grailbio/py_singlefile_binary/blob/main/bazel/zstd/rules.bzl

jheaff1: `zstd` decompression isn’t necessary since [this](https://github.com/indygreg/python-build-standalone/issues/79) issue was actioned, which added `.tar.gz` release artefacts.

python/repositories.bzl
mattem: In older versions of Python the generated bytecode was non-deterministic, but given we are fetching prebuilt versions this may be a non-issue as they wouldn't change, unless Python will continually write new `.pyc` files here?
We should add a test against the SHA of these files perhaps.

python/repositories.bzl
ewhauser: +1 for using tarballs

python/repositories.bzl
aignas: Yes, I understand that, I probably did not find a better place to comment on that, but it would be great to have an explicit example on how we would be using such platform specific python interpreter repository in a `pip_parse` rule, but maybe I am missing something.

From my experience a different system python version may lead to incompatible Python byte-code at runtime. For example, a user is using an old Mac OS X version and the system python is `3.8`, but the repo is using `3.9`, then the test targets would fail if there is a minor version mismatch. To make things worse (or better), the default Python used by `pip_parse` cannot be upgraded by doing `brew install` (at least by default).

As long as we cannot have a consistent way to use the same Python to do `pip_parse` and as a toolchain, the usefulness may be slightly limited. The only way I found it possible to use together with `pip_parse` is by having a single repository rule that does `uname -r` and downloads an appropriate binary and exposing it as `@interpreter//:bin/python`, which can then be passed to `pip_parse` via the `python_interpreter_target` attribute.

Or maybe we should handle the this (i.e. selecting the correct python interpreter) within `pip_parse`?

python/repositories.bzl
mattem: Do you know what changed within those files? My guess is that the interpreter does indeed update these files again and they contain timestamps or absolute paths.

We should exclude `RECORD` files for the same reason then.

python/repositories.bzl
thundergolfer: Thanks for the heads up @jheaff1, that's awesome.

thundergolfer:
> I'm also curious if there's interest in rules_python supporting self-contained executables in the future.

@josh-newman it's something that would be great to provide a good default for, but building a good solution can be done in a separate repo first I think. `rules_python` also already has the baggage of the zip output.  It would be preferable to have the official rules only support one way of doing things. 

Thanks for open sourcing [grailbio/py_singlefile_binary](https://github.com/grailbio/py_singlefile_binary).

python/repositories.bzl
aignas: I cannot find the logs right now, but my guess was that:
1. We download the python interpreter files and extract them.
2. We run pip_parse, which uses `bin/python` from the repository rule via the `python_interpreter_target` attribute.
3. This causes extra `.pyc` files being created during the `pip_parse`
4. Since the `.pyc` files are non-deterministic, we re-execute the Python test targets every time.

Having a minimal example should not be too difficult, but unfortunately I don't have time right now to help.

python/repositories.bzl
thundergolfer: So they latest release publishes the 'install only' tarballs that we want, for Python 3.9 and 3.10, and each of those for Windows, Apple, and Linux. 

e.g [github.com/indygreg/python-build-standalone/releases/download/20211017/cpython-3.10.0-x86_64-apple-darwin-install_only-20211017T1616.tar.gz](https://github.com/indygreg/python-build-standalone/releases/download/20211017/cpython-3.10.0-x86_64-apple-darwin-install_only-20211017T1616.tar.gz)

@f0rmiga the 'install only' tarballs will remove a decent amount of complexity and non-hermeticity. Let's give them a try.

thundergolfer: Just a heads up that I'm getting hands on with the 'install only' tarballs over in https://github.com/thundergolfer/example-bazel-monorepo/pull/54 and documenting my troubles.

python/private/toolchains_repo.bzl
```diff
@@ -0,0 +1,83 @@
...
+This follows guidance here:
+https://docs.bazel.build/versions/main/skylark/deploying.html#registering-toolchains
+
+The "complex computation" in our case is simply downloading large artifacts.
+This guidance tells us how to avoid that: we put the toolchain targets in the
+alias repository with only the toolchain attribute pointing into the
+platform-specific repositories.
+"""
+
+PLATFORMS = {
```
UebelAndre, line 26: The way this dict is used appears to lead to failures where if you specify python 3.8 as the target, which does not have an artifact for `aarch64-apple-darwin`, Bazel will error out. Specifying the desired version should only define what's available.

python/repositories.bzl
```diff
@@ -1,8 +1,199 @@
...
 def py_repositories():
     # buildifier: disable=print
     print("py_repositories is a no-op and is deprecated. You can remove this from your WORKSPACE file")
+
+########
+# Remaining content of the file is only used to support toolchains.
+########
+_DOC = "Fetch external tools needed for python toolchain."
+_ATTRS = {
+    "integrity": attr.string(mandatory = True),
```
UebelAndre, line 32: Why is this using `integrity` and not `sha256`, which I feel is far more common?

UebelAndre (changes requested):
For greater visibility, I've got a branch that adds windows support and uses `tar.gz` files at https://github.com/bazelbuild/rules_python/pull/628 hopefully that helps 😄

python/private/toolchains_repo.bzl
```diff
@@ -0,0 +1,83 @@
...
+# these targets.
+
+"""
+
+    for [platform, meta] in PLATFORMS.items():
+        build_content += """\
+toolchain(
+    name = "{platform}_toolchain",
+    exec_compatible_with = {compatible_with},
+    target_compatible_with = {compatible_with},
```
UebelAndre, line 63: Is it correct to restrict the toolchain to the same target and exec environment? I would kinda have expected `target` to be left empty since I don't think there is a "target" restriction in the toolchain. Or does the toolchain somehow become embedded in runfiles?

UebelAndre (changes requested):

python/private/toolchains_repo.bzl
f0rmiga: Fixed in https://github.com/bazelbuild/rules_python/pull/618/commits/0e91655421bdc5fd16a50ba9e47733fbeb45502f.

python/repositories.bzl
f0rmiga: Took the opportunity to switch it in https://github.com/bazelbuild/rules_python/pull/618/commits/d7ac042f2ad92670f60e25d4f97fcdef567c7859.

UebelAndre (changes requested):
Would it be possible to integrate these toolchains into `pip_repository` as the default interpreters?
https://github.com/bazelbuild/rules_python/blob/27d0c7bb8e663dd2e2e9b295ecbfed680e641dfd/python/pip_install/pip_repository.bzl#L231-L238

f0rmiga:
> Would it be possible to integrate these toolchains into `pip_repository` as the default interpreters?
> 
> https://github.com/bazelbuild/rules_python/blob/27d0c7bb8e663dd2e2e9b295ecbfed680e641dfd/python/pip_install/pip_repository.bzl#L231-L238

Possibly. @alexeagle will push a commit adding some aliases to the host hermetic interpreter.

UebelAndre:
> Possibly. @alexeagle will push a commit adding some aliases to the host hermetic interpreter.

[alias](https://docs.bazel.build/versions/main/be/general.html#alias) may not work since this is a repository rule. There'd have to be some way of selecting the right repository and extracting only the needed toolchain

python/private/toolchains_repo.bzl
alexeagle: I agree this is probably wrong. I think the bug originates in the template
https://github.com/bazel-contrib/rules-template/blob/main/mylang/private/toolchains_repo.bzl#L91-L92

exec_compatible_with toolchain would be used for build actions. If the user wants a python interpreter in the target (like for a docker container) that might be a use case for target_compatible_with, but probably not needed in this PR

alexeagle: to be precise, when the target platform differs from the execution platform, I think this code will result in no toolchains being selected

python/repositories.bzl
f0rmiga: @aignas could you try the latest state of this PR? I made the `lib` directory read-only, so it should be immutable. I wonder if you will still get the same error.

python/repositories.bzl
```diff
@@ -1,8 +1,252 @@
...
+
+py_runtime_pair(
+    name = "python_runtimes",
+    py2_runtime = ":py2_runtime",
+    py3_runtime = ":py3_runtime",
+)
+""".format(
+        python_path = python_bin,
+    )
+    rctx.file("BUILD.bazel", build_content)
```
UebelAndre, line 138: Curious why a `WORKSPACE.bazel` file was not needed. Is that provided somewhere else?

UebelAndre (changes requested):

python/repositories.bzl
alexeagle: One is always generated for a repository rule, and you cannot influence the content

python/private/toolchains_repo.bzl
alexeagle: Users can totally make their own if they want to constrain a toolchain on some other platforms. I don't think we have to cover long tail use cases, at least until they are requested

python/repositories.bzl
```diff
@@ -1,8 +1,252 @@
...
+            values = PLATFORMS.keys(),
+        ),
+        "python_version": attr.string(
+            doc = "The Python version.",
+            mandatory = True,
+            values = TOOL_VERSIONS.keys() + MINOR_MAPPING.keys(),
+        ),
+        "sha256": attr.string(
+            doc = "The SHA256 integrity hash for the Python interpreter tarball.",
+            mandatory = True,
```
UebelAndre, line 163: I don't think this should be mandatory. Allowing users to not set it will cause bazel to post the usual logging informing users they can improve their experience by filling this in. My general workflow when adding new dependencies is to add the url, do a sync, then fill in the blanks based on the warnings.

UebelAndre (changes requested):

python/repositories.bzl
f0rmiga: See https://github.com/bazelbuild/rules_python/pull/618/commits/d12143673d10660387a9bfdd87bfca1a55c9e0bf.

python/repositories.bzl
f0rmiga: Done for zstd here 77383e3.

python/repositories.bzl
f0rmiga: I created a tool to get the hashes here: https://github.com/bazelbuild/rules_python/pull/618/commits/d7ac042f2ad92670f60e25d4f97fcdef567c7859.

README.md
```diff
@@ -54,6 +54,28 @@ http_archive(
...
+```python
+load("@rules_python//python:repositories.bzl", "python_register_toolchains")
+
+# Multiple versions are available in the @rules_python//python:versions.bzl file.
+python_register_toolchains(
+    name = "python310",
+    python_version = "3.10",
+)
+
+load("@python39_host//:defs.bzl", "interpreter")
```
UebelAndre, line 68: should this be `310`?
```suggestion
load("@python310_host//:defs.bzl", "interpreter")
```

python/repositories.bzl
```diff
@@ -1,8 +1,224 @@
...
+    name = "python_runtimes",
+    py2_runtime = ":py2_runtime",
+    py3_runtime = ":py3_runtime",
+)
+""".format(
+        python_path = python_bin,
+    )
+    rctx.file("BUILD.bazel", build_content)
+
+    return {
```
UebelAndre, line 140: this should be deleted of no attribute is optional so no one forgets to update this in the future

python/versions.bzl
```diff
@@ -0,0 +1,108 @@
...
+    url = "/".join([_RELEASE_BASE_URL, release_filename])
+    return (release_filename, url)
+
+# When updating the versions and releases, run the following command to get
+# the hashes:
+#   bazel run //python/private:print_toolchains_checksums
+#
+# buildifier: disable=unsorted-dict-items
+TOOL_VERSIONS = {
+    "3.8.10": {
```
UebelAndre, line 35: Why are there two versions of python3.8?

UebelAndre (changes requested):

README.md
f0rmiga: Yes, it should! Thanks.

README.md
```diff
@@ -54,6 +54,28 @@ http_archive(
...
+```python
+load("@rules_python//python:repositories.bzl", "python_register_toolchains")
+
+# Multiple versions are available in the @rules_python//python:versions.bzl file.
+python_register_toolchains(
+    name = "python310",
+    python_version = "3.10",
+)
+
+load("@python310_host//:defs.bzl", "interpreter")
```
mattem, line 68: I kinda find this confusing that it's called `host` toolchain. It's resolved for the host platform, but it's not the hosts interpreter.

python/private/toolchains_repo.bzl
```diff
@@ -0,0 +1,150 @@
...
+    python3_binary_path = "python.exe" if (os_name == WINDOWS_NAME) else "bin/python3"
+
+    # Base BUILD file for this repository.
+    build_contents = """\
+# Generated by python/repositories.bzl
+package(default_visibility = ["//visibility:public"])
+alias(name = "files",           actual = "@{py_repository}_{host_platform}//:files")
+alias(name = "py3_runtime",     actual = "@{py_repository}_{host_platform}//:py3_runtime")
+alias(name = "python_runtimes", actual = "@{py_repository}_{host_platform}//:python_runtimes")
+alias(name = "python3",         actual = "@{py_repository}_{host_platform}//:{python3_binary_path}")
```
mattem, line 91: Can we include an alias for `pip`?

python/versions.bzl
alexeagle: because some users need that. We want to ensure that this new API doesn't force users to change their current interpreter version, it ought to allow any version that we know how to fetch

README.md
alexeagle: what would you want to name it? this one matches rules_nodejs

python/private/toolchains_repo.bzl
f0rmiga: Can only provide it for Linux and Mac.

f0rmiga: Addressed with fd9f983.

README.md
f0rmiga: Any suggestions for better naming?

mattem: I liked the `resolved` naming scheme, matches the `resolved_toolchain`?
https://github.com/bazelbuild/rules_nodejs/blob/9dddd19bc9bc91cb5a0fd932e3cc2d601133a9ea/nodejs/private/toolchains_repo.bzl

I know this isn't the toolchain, but `resolved_interpreter` or similar?

python/repositories.bzl
```diff
@@ -1,8 +1,225 @@
...
+    ),
+)
+
+exports_files(["{python_path}"])
+
+py_runtime(
+    name = "py2_runtime",
+    interpreter_path = "/bin/false",
+    python_version = "PY2",
+)
```
mattem, line 122: What's the reason for defining this? 
It means checks like `if toolchain.py2_runtime:` will always be true, which seems odd given that it points to `/bin/false`.
Could the `py2_runtime` be unset?

f0rmiga: Addressed in f509b46.

README.md
f0rmiga: Addressed in [7eacdf1](https://github.com/bazelbuild/rules_python/pull/618/commits/7eacdf1eb22c1f24df10d332c0908413c21f30f4).

python/repositories.bzl
f0rmiga: Will leave for now.

alexeagle (approved):
I put some suggestions in https://github.com/bazelbuild/rules_python/pull/642
in particular, fixing the "target_compatible_with" that doesn't belong, matching https://github.com/bazel-contrib/rules-template/pull/9

otherwise looks great to me!

sluongng:
🎉  amazing work!

aaliddell:
Awesome 👍

aignas:
Thank you for the work. I tried to use it on my Mac and everything worked except for chmoding the extracted Python folder. Since I am the only one in here facing this issue, it could be something to do with my system, but wanted to give a heads up in case anyone else encounters this: https://github.com/bazelbuild/rules_python/commit/bed8c1b410d435bc795a1d03acdc737b8aa1b37f#r68477007

README.md
UebelAndre: I've opened https://github.com/bazelbuild/rules_python/pull/656 which relates to this discussion.

PR #623, opened 2022-02-17
add constraints option to pip_repository

dymart:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Currently you cannot pass a constraint file to pip_repository 

## What is the new behavior?

You can pass a constraints file to pip_repository

- [ ] Yes
- [x] No

### DIFF
--- a/python/pip_install/extract_wheels/__init__.py
+++ b/python/pip_install/extract_wheels/__init__.py
@@ -65,6 +65,12 @@ def main() -> None:
         help="Path to requirements.txt from where to install dependencies",
     )
     parser.add_argument(
+        "--constraints",
+        action="store",
+        required=False,
+        help="Path to constraints.txt, to find the versions that should be used",
+    )
+    parser.add_argument(
         "--annotations",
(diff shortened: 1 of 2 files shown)


mattem:
At a minimum we should document this arg in the `pip_install` marco here https://github.com/bazelbuild/rules_python/blob/main/python/pip.bzl#L23

thundergolfer (commented):
I think I'd prefer if we avoid adding this complexity into the `pip_repository` rule and instead have have this part of the ruleset take the `requirements.txt` file 'as is'. 

`compile_pip_requirements` is a utility macro for turning a `requirements.in` file into a transitively locked `requirements.txt` file. If we require adding support for a _constraints file_, I think I'd prefer having it in that place. 

Ref: https://pip.pypa.io/en/stable/user_guide/#constraints-files

dymart:
Currently don't have the bandwidth to upstream as there would be more prefactors required for this to be useful. 

We may come back and upstream py_requirements though.

PR #630, opened 2022-03-02
feat: concurrent parsing for Gazelle

f0rmiga:
No breaking changes, just some performance improvements for the Gazelle extension.


gazelle/parse.py
```diff
@@ -38,25 +42,44 @@ def parse_comments(content):
...
-def parse(stdout, filepath):
-    with open(filepath, "r") as file:
+def parse(repo_root, rel_package_path, filename):
+    rel_filepath = os.path.join(rel_package_path, filename)
+    abs_filepath = os.path.join(repo_root, rel_filepath)
+    with open(abs_filepath, "r") as file:
         content = file.read()
-        modules = parse_import_statements(content)
-        comments = parse_comments(content)
+        with concurrent.futures.ThreadPoolExecutor(max_workers=2) as executor:
```
alexeagle, line 50: where did you come up with the number 2 here? maybe worth a comment

alexeagle (approved):

gazelle/parse.py
f0rmiga, line 50:
```suggestion
       # From simple benchmarks, 2 workers gave the best performance here.
        with concurrent.futures.ThreadPoolExecutor(max_workers=2) as executor:
```

f0rmiga: Addressed with [9732ff7](https://github.com/bazelbuild/rules_python/pull/630/commits/9732ff794232ec4c13819be5e16b00fc32d77e74).

PR #636, opened 2022-03-03
Allow previous indygreg releases

alexeagle:
This gives us more python patch versions

### DIFF
--- a/python/private/tests/toolchains/versions_test.bzl
+++ b/python/private/tests/toolchains/versions_test.bzl
@@ -21,14 +21,13 @@ load("//python/private:versions.bzl", "MINOR_MAPPING", "TOOL_VERSIONS")
 
 required_platforms = [
     "x86_64-apple-darwin",
-    "x86_64-pc-windows-msvc",
     "x86_64-unknown-linux-gnu",
 ]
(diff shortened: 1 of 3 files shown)


python/private/tests/toolchains/versions_test.bzl
```diff
@@ -21,14 +21,13 @@ load("//python/private:versions.bzl", "MINOR_MAPPING", "TOOL_VERSIONS")
 
 required_platforms = [
     "x86_64-apple-darwin",
-    "x86_64-pc-windows-msvc",
```
alexeagle, line 24: note, there *are* windows distros of 3.8.10:

 cpython-3.8.10-x86_64-pc-windows-msvc-shared-pgo-20210506T0943.tar.zst
 cpython-3.8.10-x86_64-pc-windows-msvc-static-noopt-20210506T0943.tar.zst

but they don't follow the naming convention that darwin and linux use, so we'd have to make the templating even more complex. I think we can just say that windows is not always included in every patch version.

f0rmiga: I'm fine with this.

PR #637, opened 2022-03-03
Allow METADATA files in pip_repository generated targets

UebelAndre:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Currently `*.dist-info/METADATA` files are excluded in `pip_repository` generated targets. This is incorrect as it's consumed by https://docs.python.org/3/library/importlib.metadata.html

Issue Number: N/A

## What is the new behavior?
This change allows `METADATA` files to be included in `pip_repository` generated targets, enabling the use of https://docs.python.org/3/library/importlib.metadata.html.

- [ ] Yes
- [x] No

### DIFF
--- a/examples/pip_repository_annotations/pip_repository_annotations_test.py
+++ b/examples/pip_repository_annotations/pip_repository_annotations_test.py
@@ -64,6 +64,34 @@ class PipRepositoryAnnotationsTest(unittest.TestCase):
         stdout = proc.stdout.decode("utf-8").strip()
         self.assertEqual(stdout, "Hello world from copied executable")
 
+    def test_data_exclude_glob(self):
+        current_wheel_version = "0.37.1"
+
+        r = runfiles.Create()
+        dist_info_dir = (
+            "pip_repository_annotations_example/external/{}/wheel-{}.dist-info".format(
+                self.wheel_pkg_dir(),
+                current_wheel_version,
+            )
+        )
+
+        # Note: `METADATA` is important as it's consumed by https://docs.python.org/3/library/importlib.metadata.html
+        # `METADATA` is expected to be there to show dist-info files are included in the runfiles.
+        metadata_path = r.Rlocation("{}/METADATA".format(dist_info_dir))
+
+        # However, `WHEEL` was explicitly excluded, so it should be missing
+        wheel_path = r.Rlocation("{}/WHEEL".format(dist_info_dir))
+
+        # Because windows does not have `--enable_runfiles` on by default, the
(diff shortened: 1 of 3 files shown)


examples/pip_repository_annotations/pip_repository_annotations_test.py
```diff
@@ -64,6 +64,34 @@ def test_copy_executables(self):
...
+        dist_info_dir = (
+            "pip_repository_annotations_example/external/{}/wheel-{}.dist-info".format(
+                self.wheel_pkg_dir(),
+                current_wheel_version,
+            )
+        )
+
+        # Note: `METADATA` is important as it's consumed by https://docs.python.org/3/library/importlib.metadata.html
+        # `METADATA` is expected to be there to show dist-info files are included in the runfiles.
+        metadata_path = r.Rlocation("{}/METADATA".format(dist_info_dir))
```
UebelAndre, line 80: This is a restored regression test from https://github.com/bazelbuild/rules_python/pull/626 but has new meaning as it now uses `METADATA` files to ensure this isn't removed from the `data` attributes of generated targets.

mattem (approved):

python/pip_install/extract_wheels/lib/bazel.py
```diff
@@ -139,12 +139,10 @@ def generate_build_file_contents(
     there may be no Python sources whatsoever (e.g. packages written in Cython: like `pymssql`).
     """
 
-    # `dist-info` contains non-determinisitc files which can change any time
-    # the repository rules run. Below is a list of known patterns to these
-    # files. However, not all files should be ignored as certain packages
-    # require things like `top_level.txt`.
     dist_info_ignores = [
-        "**/*.dist-info/METADATA",
+        # RECORD is known to contain sha256 checksums of files which might include the checksums
```
ianoc, line 143: Are there known examples this happens in? Some import lib functions fail still because they use data from RECORD, the files() method on a distribution for instance.

groodt: I too have been wondering why we're doing so much non-standard unpacking and exclusion of files from wheel packages. 

I'm trying to land a change that uses `installer` to bring some PEP standards into this rule set. See: https://github.com/bazelbuild/rules_python/pull/715

I think we need a better way to handle non-determinism. I think we need to either make it opt-in somehow, or direct people in how to use a "wheel only" (using `--only-binary=:all:`). If you do that, then the only non-determinism in generated `.pyc` files, which can be excluded/included via some parameters.

PR #644, opened 2022-03-08
feat: add requires_file attr for py_wheel

weixiao-huang:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No


pstradomski:
Any background for why this is needed? Maybe the documentation should be updated to reflect that and what's the interaction with "requires" attribute?

weixiao-huang:
Because sometimes we may write codes like this in `setup.py`

```python
#!/usr/bin/env python

from setuptools import setup, find_packages

with open("requirements.txt") as fp:
    install_requires = fp.read()


setup(
    # get requires from file
    install_requires=install_requires,
)
```

However, in `py_wheel`, it only provides requires from bazel string. It seems we cannot get requires from files. So I add `requires_file` in `py_wheel`.

If `requires` specified, `requires_file` will not work

weixiao-huang:
Is there any update of this PR?

UebelAndre:
@weixiao-huang would you be able to rebase this and address conflicts?

python/packaging.bzl
```diff
@@ -188,6 +188,12 @@ def _py_wheel_impl(ctx):
 
     if ctx.attr.python_requires:
         metadata_contents.append("Requires-Python: %s" % ctx.attr.python_requires)
+
+    if len(ctx.attr.requires) == 0 and ctx.attr.requires_file:
```
UebelAndre, line 192: Would it be possible to make this additive?

tools/wheelmaker.py
```diff
@@ -258,6 +258,12 @@ def parse_args() -> argparse.Namespace:
     metadata_group.add_argument(
         "--platform", type=str, default="any", help="Target platform. "
     )
+    metadata_group.add_argument(
+        "--requires_file",
+        type=Path,
+        default="",
```
UebelAndre, line 264: I don't think there's a need to set a default here.

tools/wheelmaker.py
```diff
@@ -398,6 +404,9 @@ def main() -> None:
                       encoding="utf-8") as metadata_file:
                 metadata = metadata_file.read()
 
+        if arguments.requires_file:
+            with open(arguments.requires_file) as fp:
```
UebelAndre, line 408: Shouldn't this instead be written to `Requires-Dist:` in the metadata? My use case for this is to be able to define 3rdparty dependencies in `requirements.in` files and be able to continue to use the existing `requires` field for local dependencies.

UebelAndre (changes requested):
I'm no maintainer but did have some requests 😅

tools/wheelmaker.py
```diff
@@ -398,6 +403,10 @@ def main() -> None:
...
+                additive_requires_list = ["Requires-Dist: {}".format(line) for line in fp.read().strip().split("\n")]
+                metadata += "\n" + "\n".join(additive_requires_list)
```
pstradomski, line 409: Won't that leave you with two \n in a row, so everything below will be interpreted as description?

Can we have a test?

tools/wheelmaker.py
```diff
@@ -398,6 +403,10 @@ def main() -> None:
                       encoding="utf-8") as metadata_file:
                 metadata = metadata_file.read()
 
+        if arguments.requires_file:
+            with open(arguments.requires_file) as fp:
+                additive_requires_list = ["Requires-Dist: {}".format(line) for line in fp.read().strip().split("\n")]
```
pstradomski, line 408: why read().strip().split() and not readlines()?

chrislovecnm:
@weixiao-huang knudge

weixiao-huang:
Sorry, I'll check it recently

PR #653, opened 2022-03-14
docs: avoid users on 3.10 by accident

alexeagle:


thundergolfer (approved):

thundergolfer:
`3.10` seems a bit too bleeding edge as of early March 2022. Wheel coverage is certainly not as high as 3.9, and wheel misses will cause repo rule execution errors. Scikit-learn is a major package that released 3.10 wheels only 2-2.5 months ago (https://github.com/scikit-learn/scikit-learn/issues/21401). My anecdotal experience is that even a workspace with a small number of 'more complex' third-party packages (eg. scikit, matplotlib) is going to get a wheel miss on 3.10.

PR #654, opened 2022-03-14
fix: take custom tools versions into account when selecting Python interpreter

mattem:
When providing a custom dict for the `tool_versions`, take this into account when selecting the filename and url.

This also changes the `get_release_url` function to allow for different URLs for each platform (where a platform is added after the initial release of the interpreter version).


python/versions.bzl
```diff
@@ -45,7 +36,11 @@ TOOL_VERSIONS = {
         },
     },
     "3.8.12": {
-        "url": "20220227/cpython-{python_version}+20220227-{platform}-{build}.tar.gz",
+        "url": {
+            "x86_64-apple-darwin": "20220227/cpython-{python_version}+20220227-{platform}-{build}.tar.gz",
```
alexeagle, line 40: these still look the same, is there a reason to make url a dict now instead of a string?

python/repositories.bzl
```diff
@@ -187,10 +185,18 @@ python_repository = repository_rule(
             mandatory = True,
             values = TOOL_VERSIONS.keys() + MINOR_MAPPING.keys(),
         ),
+        "release_filename": attr.string(
```
alexeagle, line 188: hmm it's smelly there's no docs/ change in this PR, something must not be wired up

alexeagle (approved):

python/versions.bzl
mattem: From what @indygreg mentioned in https://github.com/indygreg/python-build-standalone/issues/118#issuecomment-1066180403, I was expecting the `aarch64-apple-darwin` platform to have a different release to the others, so this entry will eventually need to be a dict.

PR #655, opened 2022-03-15
Document how to vendor a pip_parse requirements.bzl file

### LINKED ISSUE #608: pip_parse is burdensome for downstream projects
Typically, to install an upstream project, it looks something like

```bzl
http_repository(
  name = "example",
  # ...
)

load("@example//:workspace.bzl", "example_repositories")
example_repositories()
```

However, suppose `example` uses pip dependencies.

The install must be at minimum:

```bzl
http_repository(
  name = "example",
  # ...
)

# create pip_parse repo
load("@example://workspace_part_1.bzl", "example_repositories_part_1")
example_repositories_part_1()

# load requirements.bzl from pip_parse repo and call install_deps()
load("@example//:workspace_part_2.bzl", "example_repositories_part_2")
example_repositories_part_2()
```

This looks weird, and further burdens already lengthy WORKSPACE files.

---

A solution is to expose generating requirements.bzl. Then the upstream project can commit that to the source tree (probably the same step as generating the requirements lock file today). And then downstream projects can initialize repos with only one step.

This allows the repository creation to happen in one step.

alexeagle:
fixes #608


python/pip_install/parse_requirements_to_bzl/__init__.py
```diff
@@ -166,7 +166,7 @@ def coerce_to_bool(option):
     return str(option).lower() == "true"
 
 
-def main() -> None:
+def main(output) -> None:
```
UebelAndre, line 169: nit: Can you type annotate `output`? Perhaps even add a docstring?

UebelAndre (changes requested):

hrfuller (approved):
Thanks! LGTM modulo comments.

python/pip_install/pip_repository.bzl
```diff
@@ -178,6 +180,9 @@ def _pip_repository_impl(rctx):
     if result.return_code:
         fail("rules_python failed: %s (%s)" % (result.stdout, result.stderr))
 
+    # We need a BUILD file to load the generated requirements.bzl
+    rctx.file("BUILD.bazel", _BUILD_FILE_CONTENTS.format(" ".join([str(a) for a in args])))
```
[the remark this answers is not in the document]
alexeagle: sure, my thinking here was just that there's no harm documenting the command that was used to generate a requirements.bzl for non-incremental as well, it will just be the extract_wheels command line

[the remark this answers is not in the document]
hrfuller: Yeah feel free to ignore. Just thought it might be confusing / unneccesary.

python/pip.bzl
```diff
@@ -165,6 +166,31 @@ def pip_parse(requirements_lock, name = "pip_parsed_deps", **kwargs):
         name = "flake8",
         actual = entry_point("flake8"),
     )
+
+    ## Vendoring the requirements.bzl file
```
thundergolfer, line 170: This commentary is place in the middle of a Starlark codeblock. The gen'd `.md` has syntax highlighting on it. 

I think the above codeblock just needs to be closed before starting this commentary. It's currently closed at https://github.com/bazelbuild/rules_python/pull/655/files#diff-b92f94a67df862cde9714ff224afa676e39fe02da4fd4c6993daf684fbbc7cbdR194

alexeagle: oops! thanks

PR #664, opened 2022-03-19
Added support for custom `strip_prefix` args in toolchains

UebelAndre:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Not all toolchains are packaged the same. Even older releases of https://github.com/indygreg/python-build-standalone have different archive structures which are incompatible with how toolchains are currently setup.

Issue Number: N/A

## What is the new behavior?
This change allows for a `strip_prefix` key to be added to `python_register_toolchains.tool_versions` so that users can use reasonably different archive layouts.

- [ ] Yes
- [x] No

### DIFF
--- a/python/repositories.bzl
+++ b/python/repositories.bzl
@@ -195,6 +195,7 @@ python_repository = repository_rule(
         ),
         "strip_prefix": attr.string(
             doc = "A directory prefix to strip from the extracted files.",
+            mandatory = True,
         ),
         "url": attr.string(
             doc = "The URL of the interpreter to download",
--- a/python/repositories.bzl
+++ b/python/repositories.bzl
@@ -81,7 +81,7 @@ def _python_repository_impl(rctx):
         rctx.download_and_extract(
             url = url,
             sha256 = rctx.attr.sha256,
-            stripPrefix = "python",
+            stripPrefix = rctx.attr.strip_prefix,
         )
 
     # Write distutils.cfg to the Python installation.
@@ -157,6 +157,7 @@ py_runtime_pair(
         "python_version": python_version,
         "release_filename": release_filename,
         "sha256": rctx.attr.sha256,
+        "strip_prefix": rctx.attr.strip_prefix,
         "url": url,
     }
 
@@ -191,6 +192,9 @@ python_repository = repository_rule(
         "sha256": attr.string(
             doc = "The SHA256 integrity hash for the Python interpreter tarball.",
             mandatory = True,
+        ),
+        "strip_prefix": attr.string(
+            doc = "A directory prefix to strip from the extracted files.",
         ),
         "url": attr.string(
             doc = "The URL of the interpreter to download",
@@ -244,7 +248,7 @@ def python_register_toolchains(
         if not sha256:
             continue
 
-        (release_filename, url) = get_release_url(platform, python_version, base_url, tool_versions)
+        (release_filename, url, strip_prefix) = get_release_url(platform, python_version, base_url, tool_versions)
 
         python_repository(
             name = "{name}_{platform}".format(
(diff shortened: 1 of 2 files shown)


python/versions.bzl
```diff
@@ -118,21 +122,25 @@ def get_release_url(platform, python_version, base_url = DEFAULT_RELEASE_BASE_UR
...
+        A tuple of (filename, url, and archive strip prefix)
     """
 
     url = tool_versions[python_version]["url"]
 
     if type(url) == type({}):
         url = url[platform]
 
+    strip_prefix = tool_versions[python_version].get("strip_prefix", None)
+    if type(strip_prefix) == type({}):
```
alexeagle, line 134: nit: https://docs.aspect.build/bazelbuild/bazel-skylib/1.1.1/docs/types_doc_gen.html#types.is_dict is slightly better

python/repositories.bzl
```diff
@@ -192,6 +193,9 @@ python_repository = repository_rule(
             doc = "The SHA256 integrity hash for the Python interpreter tarball.",
             mandatory = True,
         ),
+        "strip_prefix": attr.string(
+            doc = "A directory prefix to strip from the extracted files.",
```
alexeagle, line 197: either add `mandatory=True` or add the `default="python"` ?

I don't think a default of empty string is ever useful when actually extracting?

python/versions.bzl
```diff
@@ -34,6 +34,7 @@ TOOL_VERSIONS = {
             "x86_64-apple-darwin": "8d06bec08db8cdd0f64f4f05ee892cf2fcbc58cfb1dd69da2caab78fac420238",
             "x86_64-unknown-linux-gnu": "aec8c4c53373b90be7e2131093caa26063be6d9d826f599c935c0e1042af3355",
         },
+        "strip_prefix": "python",
```
alexeagle, line 37: curious, are these required here? or are you just being explicit by setting this to the default value?

alexeagle (approved):

python/versions.bzl
UebelAndre: That would require introducing a dependency on `bazel-skylib` or vendoring this file which I think would be better done in an explicit cleanup PR. Does that sound reasonable?

python/repositories.bzl
UebelAndre: It could be if someone packaged a toolchain that had everything in the root directory. I think setting `mandatory=True` is incorrect and `default="python"` forces a precedent which doesn't need to be true. An empty string would be equivalent to no strip prefix.

All that said though, since all the toolchains currently come from https://github.com/indygreg/python-build-standalone, I could set `default="python"` if you'd like but still advise against it. Your thoughts?

python/versions.bzl
UebelAndre: It's kinda both. There's no default so without this the toolchain would not be structured correctly. So to continue to have these particular ones function correctly, I added the strip_prefixes. I kinda like having these here since I noticed the archive structure changed some time ago. Looking at https://github.com/indygreg/python-build-standalone/releases/tag/20200822, the toolchains are in a `python/install` directory. My hope in adding these here is that it'd draw attention to the fact that this might be needed to get toolchains working.

python/versions.bzl
alexeagle: sure - but I'm pretty sure rules_python already requires users to have bazel-skylib installed? agree we shouldn't add a new dependency

python/repositories.bzl
alexeagle: I'd like one of mandatory (you can pass empty string if that's the right value) or default="python"

python/versions.bzl
UebelAndre: I'm not aware of that being a requirement. If that's the case, it'd be nice if there was a `python_rules_dependencies` macro that loaded all dependencies (adhering to [bazel-contrib](https://github.com/bazel-contrib/rules-template/blob/f26177bcbdd9eeab373f23ca568c98704efddd3f/mylang/repositories.bzl)).

python/repositories.bzl
UebelAndre: The attribute is now mandatory

alexeagle (approved):

Issue #624, opened 2022-02-17
Update `compile_pip_requirements` to support `requirements.in` files from multiple external repositories


UebelAndre:
# 🚀 feature request

### Relevant Rules

[compile_pip_requirements](https://github.com/bazelbuild/rules_python/blob/0.6.0/python/pip_install/requirements.bzl#L6-L33)

### Description

I have some external repositories that have some rules backed by python targets and have some requirements. Instead of needing to create multiple `pip_parse` or `pip_install` repositories, I would like to generate a `requirements.txt` file that has all the necessary packages so I don't end up downloading things multiple times.

### Describe the solution you'd like

I would like to be able to pass the `requirements.in` files to `compile_pip_requirements` and have that inject them as `-r {requirements}` before compiling.
```python
compile_pip_requirements(
    name = "requirements",
    extra_args = ["--allow-unsafe"],
    requirements_in = "requirements.in",
    requirements_txt = "requirements.txt",
    external_requirements = [
        "@tool_a//:requirements.in",
        "@tool_b//:requirements.in",
        "@tool_c//:requirements.in",
    ],
)
```

### Describe alternatives you've considered

 Have you considered any alternative solutions or workarounds?
Open to ideas but I'd like to easily be able to merge other sets of requirements into one collection (which is one of the benefits of being able to compile a locked `requirements.txt` file).

UebelAndre:
This is still desired. 

I don't think `rules_python` gets so many issues that this bot needs to continually delete them. This functionality should be disabled.

PR #670, opened 2022-03-22
Wrap reqs defined by `pip_parse.install_deps` in `maybe`

UebelAndre:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Currently with the new hermetic toolchains some packages don't work like they used to when using system python. A potential solution is to build custom wheels that *do* work with the new toolchain and explicitly define them as an `http_archive`. Unfortunately, the repository you get (to me) is ambiguous since there's no `maybe` wrapper around the `whl_library` repositories.

Issue Number: N/A

## What is the new behavior?
This change wraps `whl_library` repositories defined by `pip_parse` in maybe to guarantee repositories with the same name defined first in the workspace get used.

- [ ] Yes
- [x] No


f0rmiga:
@mattem What do you think about this?

mattem:
I feel this kinda a stopgap, as in, if the http_archive should take precedence, then define it _after_ the pip rule runs. Ideally we could provide a why to remove it from the lock file as I'm assuming it's only in there so pip can resolve dependencies for it?

UebelAndre:
Ideally there would only be 1 pip repository. But I think without something like https://github.com/bazelbuild/rules_python/issues/624 or guidance on how two different repos can coordinate a shared dependency pool based on what the root workspace is, this solves for confusion and allows users to maintain that "the repository defined first is what's used".

PR #679, opened 2022-04-01
change approach for vendoring parsed requirements

### LINKED ISSUE #677: Regression caused by #655
# 🐞 bug report

### Affected Rule

`pip_repository` breaks if `BUILD_WORKING_DIRECTORY` is set where the `requirements.bzl` file is not written.

### Is this a regression?

This is a regression caused by https://github.com/bazelbuild/rules_python/pull/655#discussion_r839916855

### Description

`pip_repository` should always write a `requirements.bzl` file. Printing the contents should be a sideeffect.

## 🔬 Minimal Reproduction

From the root of `rules_python`:

```shell
cd examples/pip_parse && export BUILD_WORKING_DIRECTORY=$(pwd) && bazel test //...
```

## 🔥 Exception or Error

```
pip_parse % bazel test //...
INFO: Invocation ID: 9b05ae6c-b5e9-42f6-9820-376b018e7233
ERROR: error loading package '': cannot load '@pypi//:requirements.bzl': no such file
INFO: Elapsed time: 6.276s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (0 packages loaded)
FAILED: Build did NOT complete successfully (0 packages loaded)
```

alexeagle:
Instead of printing something different in the pip_parse tooling, just suggest that users continue
to run the repository rule, but check the output into their repo, then load from that instead
of from the generated repository.

This is simpler than trying to work out what correct arguments to pass to the tool when running it
outside of the pip_parse starlark context.

Fixes #677.


examples/pip_parse_vendored/requirements.bzl
```diff
@@ -0,0 +1,49 @@
...
+_config = {'python_interpreter': 'python3', 'python_interpreter_target': '@python39_aarch64-apple-darwin//:bin/python3', 'quiet': True, 'timeout': 600, 'repo': 'pip_parsed', 'isolated': True, 'extra_pip_args': [], 'pip_data_exclude': [], 'enable_implicit_namespace_pkgs': False, 'environment': {}, 'repo_prefix': 'pip_parsed_'}
```
alexeagle, line 13: ugh, this line gets a resolved interpreter which is platform-dependent and can't be checked in.
`@python39//:python3` would work fine here, as it's an alias to the platform interpreter, however we use `rctx.rpath` on the label provided and that doesn't work with alias labels.
Hmm.

alexeagle: okay we can just "fix it in post" - when we copy this file, replace it back with the symbolic reference

alexeagle:
This is an annoying change because I can't find a way to exclude the generated .bzl file from buildifier, so to get a green CI I had to make formatting tweaks to fix what we generate, and then more tweaks on the result in the genrule, sigh

docs/pip.md
```diff
@@ -230,19 +230,11 @@ See https://github.com/bazelbuild/rules_python/issues/608
...
-  --requirements_lock ./requirements_lock.txt \
-  --quiet False --timeout 120 --repo pypi --repo-prefix pypi_ > requirements.bzl
-```
-
-Then load the requirements.bzl file directly, without using `pip_parse` in the WORKSPACE.
+To do this, use the "write to source file" pattern documented in
+https://blog.aspect.dev/bazel-can-write-to-the-source-folder
+to put a copy of the generated requirements.bzl into your project.
+Then load the requirements.bzl file directly rather than from the generated repository.
+See the example in rules_python/examples/pip_parse_vendored.
```
hrfuller, line 237: Can we provide guidance for those who don't want to use the aspect lib? Something like `cp $(bazel info output_base)/external/<pip_parse_repo_name>/requirements.bzl ./requirements.bzl`

hrfuller (approved):
Thanks for fixing this!

examples/pip_parse_vendored/WORKSPACE
```diff
@@ -0,0 +1,42 @@
...
+    python_version = "3.9",
+)
+
+load("@python39//:defs.bzl", "interpreter")
+load("@rules_python//python:pip.bzl", "pip_parse")
+
+# This repository isn't referenced, except by our test that asserts the requirements.bzl is updated.
+# It also wouldn't be needed by users of this ruleset.
+pip_parse(
+    name = "pip_parsed",
```
hrfuller, line 30: is `parsed` intentional vs `parse`?

docs/pip.md
alexeagle: Yeah I'll make a comment about that at the callsite of `write_source_files` in the example.

alexeagle: Actually, I just replaced it with bazel_skylib so there's no usage of aspect_bazel_lib in the example.

examples/pip_parse_vendored/WORKSPACE
alexeagle: just copied from examples/pip_repository_annotations. I'll change it to just "pip".

rahul-theorem:
@alexeagle would you be willing to cut a hotfix containing this fix?

alexeagle:
I plan to cut a regular release tomorrow.

PR #688, opened 2022-04-15
Replace host cfg with exec

alexeagle:
Something changed on Bazel CI (probably the version of buildifier) such that this suddenly broke on the latest nightly master despite us not changing anything.

See https://buildkite.com/bazel/rules-python-python/builds/2694#d430bd0c-f1cb-496c-81ae-6a9eae413f21


alexeagle:
Note, the alternative would be for us to pin our buildifier version in our .bazelci config

f0rmiga (approved):

PR #690, opened 2022-04-16
pip_compile: remove external/workspace_name prefix from generated requirements.txt

### LINKED ISSUE #689: the test created by compile_pip_requirements will fail when run as an external dependency
# 🐞 bug report

### Affected Rule

compile_pip_requirements 

### Is this a regression?

Not sure.

### Description

The test created by compile_pip_requirements will fail when run as an external dependency.

```diff
     # via
-    #   -r x/requirements.in
+    #   -r external/y/x/requirements.in
     #   z
```

## 🔥 Exception or Error

The test should result in the same outcome however it is run.

## 🌍  Your Environment

**Operating System:**

Fedora 35

5.1.0

**Rules_python version:**

0.8.0

BoleynSu:
Fixes #689

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #689

## What is the new behavior?

In the generated file `via -r filename` will be replaced with `via -r -`.

- [x] Yes
- [ ] No

The existing generated tests will fail and a regeneration is required.


BoleynSu:
The test is not added yet and I am expecting to break some existing ones.
I will fix them and add a new one once the PR is SGTMed.

BoleynSu:
Friendly ping.

alexeagle:
I don't think this is really an improvement - these `from` lines are useful documentation to say where a requirement came from, and the user doesn't understand what stdin is in this case, it's hidden by Bazel.

Can we just fixup the `external` pathing before writing the file instead?

BoleynSu:
Sure. That makes sense. Just that's more work to do and I want to make sure
the upstream maintainer would think the issue is acceptable to be fixed
(which is not always the case) in the first place before I spend more time
on it.

BoleynSu:
As a byproduct, this also allows |compile_pip_requirements| to comsume a generated requirements.in.
@alexeagle PTAL. I can add some tests after you think the PR is OK.

f0rmiga:
@alexeagle Seems valid to me with proper tests. Anything else you want @BoleynSu to change?

alexeagle:
Yup, seems like a targeted fix. Even a simple test by checking in a requirements file that currently has an external/ segment would help.

BoleynSu:
Cool. I will add some tests later. I have implemented something simliar to `@unpinned_maven//:pin` which depend on this fix. Maybe I can use that as a test here.

BoleynSu:
Hi @alexeagle  and @f0rmiga, I have added the tests PTAL. Also I implemented something simliar to `@unpinned_maven//:pin` in the test. Would you mind also check if that can be part of the contributed repository rules?

tests/pip_deps/WORKSPACE
```diff
@@ -0,0 +1,26 @@
...
+    name = "python310",
+    python_version = "3.10",
+)
+
+load("@python310//:defs.bzl", "interpreter")
+load("//:pip_deps.bzl", "pip_deps")
+
+PIP_PACKAGES = {"sphinx": "4.5.0"}
+
+pip_deps(
```
alexeagle, line 22: this was the intent of the `pip_parse_vendored` example introduced in https://github.com/bazelbuild/rules_python/commit/12662b68d89976967a4b4653e79a9a88a6b29271

could you just update that example instead of introducing another e2e test? They are kinda slow so it would be better not to accumulate more.

Also this doc https://github.com/bazelbuild/rules_python/commit/12662b68d89976967a4b4653e79a9a88a6b29271#diff-b92f94a67df862cde9714ff224afa676e39fe02da4fd4c6993daf684fbbc7cbdR182-R186 should maybe be updated?

.bazelci/presubmit.yml
```diff
@@ -36,5 +36,6 @@ platforms:
       - "-//gazelle/..."
       # The dependencies needed for this test are not cross-platform: https://github.com/bazelbuild/rules_python/issues/260
       - "-//tests:pip_repository_entry_points_example"
+      - "-//tests:pip_deps_example"
```
alexeagle, line 39: why doesn't it work on Windows?

python/pip_install/pip_compile.py
```diff
@@ -14,7 +14,12 @@
     sys.exit(1)
 
 requirements_in = os.path.relpath(sys.argv.pop(1))
-requirements_txt = sys.argv.pop(1)
+requirements_txt = os.path.relpath(sys.argv.pop(1))
+parts = requirements_in.split(os.path.sep, 2)
+if parts[0] == "external":
+    requirements_in = parts[2]
```
alexeagle, line 20: shouldn't this be all the segments joined other than the first two? what if requirements.in isn't in the root of the external repository, but is in `external/other_repo/some_folder/requirements.in` ?

tests/pip_deps/pip_deps.bzl
```diff
@@ -0,0 +1,55 @@
...
+def pip_deps(
+        *,
+        name = "pip",
+        packages = {},
+        requirements_lock_target = Label("//:requirements_txt"),
+        requirements_lock_file = Label("//:requirements.txt"),
+        python_interpreter_target = None,
+        **kwargs):
+    _requirements_in(
+        name = "unpinned_" + name,
```
alexeagle, line 46: I think the term "unpinned" may be inaccurate here. I'm not sure how it compares with rules_jvm_external, but here, the existing locked requirements file is still an input to pip-compile, so any already-pinned transitive dependencies will remain pinned to the same version. Only versions which no longer satisfy the constraints in requirements.in will actually change.

If the behavior is different from rules_jvm_external, then maybe another name would be better?

tests/pip_deps/pip_deps.bzl
```diff
@@ -0,0 +1,55 @@
...
+        content = "".join(["{package} == {version}\n".format(
+            package = package,
+            version = version,
+        ) for (package, version) in repository_ctx.attr.packages.items()]),
+    )
+    repository_ctx.file("WORKSPACE", content = "")
+    repository_ctx.file("BUILD", content = """
+load("@rules_python//python:pip.bzl", "compile_pip_requirements")
+
+compile_pip_requirements(
```
alexeagle, line 17: this is really clever to drop the updater binary into an external repo like this :)

tests/pip_deps/WORKSPACE
BoleynSu: > this was the intent of the `pip_parse_vendored` example introduced in [12662b6](https://github.com/bazelbuild/rules_python/commit/12662b68d89976967a4b4653e79a9a88a6b29271)
> 
The major issue with pip_parse_vendored and the approach used for golang is that the meta data is unknown to Bazel. With pip_deps, one can save all the metadata in a .bzl file so that the downstream can consume it, e.g., use it to resolve diamond dependency.

I did something similar with golang dependencies at https://github.com/BoleynSu-Org/monorepo/blob/main/configs/deps/go_deps.bzl.

> could you just update that example instead of introducing another e2e test? They are kinda slow so it would be better not to accumulate more.
>

Sure, I think I can merge it into another e2e test.
 
> Also this doc [12662b6#diff-b92f94a67df862cde9714ff224afa676e39fe02da4fd4c6993daf684fbbc7cbdR182-R186](https://github.com/bazelbuild/rules_python/commit/12662b68d89976967a4b4653e79a9a88a6b29271#diff-b92f94a67df862cde9714ff224afa676e39fe02da4fd4c6993daf684fbbc7cbdR182-R186) should maybe be updated?

python/pip_install/pip_compile.py
BoleynSu: In this case, `parts[2]` would be `some_folder/requirements.in`. Then it should work as expected.

tests/pip_deps/pip_deps.bzl
BoleynSu: > I think the term "unpinned" may be inaccurate here. I'm not sure how it compares with rules_jvm_external, but here, the existing locked requirements file is still an input to pip-compile, so any already-pinned transitive dependencies will remain pinned to the same version. Only versions which no longer satisfy the constraints in requirements.in will actually change.
> 

The input to _requirements_in is `packages` and output is a `requirements.in` file. None of them have anything to do with the locked requirements file.

> If the behavior is different from rules_jvm_external, then maybe another name would be better?

tests/pip_deps/pip_deps.bzl
BoleynSu: Haha, thanks!

.bazelci/presubmit.yml
BoleynSu: I am not sure. I guessed it was because `bazel_integration_test` did not work well on Windows, which is likely wrong.

The error is caused by `os.chdir("external\\unpinned_pip")` reporting no such directory. I do not have a Windows machine to debug it at the moment. Would you mind take a look if possible?

python/pip_install/pip_compile.py
alexeagle: ah, I see the `split` is limited to two

tests/pip_deps/pip_deps.bzl
alexeagle: There's a requirements_lock attribute here that gets passed to pip_compile. Am I missing something?

BoleynSu:
Hi @alexeagle, I will remove `pip_deps` for now and add the test to existing an e2e test. I will propose another PR regarding `pip_deps` if you think it is worth the effort. If so, maybe we can first decide a better name for it.

BoleynSu:
Hi @alexeagle, I implemented another pip_deps at https://github.com/BoleynSu-Org/monorepo/blob/main/configs/deps/pip_deps.bzl which addressed this issue you pointed out. Thanks!

Use site:
https://github.com/BoleynSu-Org/monorepo/blob/main/configs/build/workspace_2.bzl#L21
https://github.com/BoleynSu-Org/monorepo/blob/main/configs/build/workspace_3.bzl#L7

Please let me know if you think this can be upstreamed.

alexeagle (approved):
Thanks! sorry this got stalled for so long.

alexeagle:
I'll get this in so that #689 can be fixed.

PR #692, opened 2022-04-25
pip/deps: Bump `pip-tools` -> 6.6.0

phlax:
Signed-off-by: Ryan Northey <ryan@synca.io>

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes
- [C] Other... Please describe:

Updates `pip-tools` to latest version 

This may be a breaking change for some as more recent version of `pip-compile` enforce `-` separation of namespace package names (rather than `.`)

OTOH it makes it usable with dependabot which uses (at least by default) a recent version  

Cant be used with dependabot without a lot of unnecessary changes 

## What is the new behavior?

Hopefully at least - compiled files will be ~same as produced by dependabot so will not keep trying to revert each others changes

- [X] Yes - of sorts
- [ ] No

`requirements.txt` files compiled with `compile_pip_requirements` may need to be recompiled and/or fail existing test expectations


mattem:
This is also a breaking change as pip-tools drops support for Python 3.6 in 6.5.0 (I know it's EoL, just a note for the diff) 
Also 6.4.0 requires a minimum of pip 21.2.0 should be noted too.

phlax:
@mattem lmk if i need to add anything, happy to update

testing out the PR i saw a couple of other issues that may cause issues with dependabot 

one is that dependabout runs in the directory of the requirements file so all of the paths are relative to the reqs file not the WORKSPACE

re that issue im wondering if we can add some flag or similar to the rule that allows it run in the directory etc

the other issue im wondering about is the `pip-compile` command in the requirements.txt - its helpful that it shows the correct bazel command, but i think to prevent unnecessary changes with dependabot it needs to just be `pip-compile` 

for ref one of the reqs files i hope to manage with in this way is here https://github.com/envoyproxy/envoy/blob/334974a0081d70b6a258f27cd40a827caf49aabf/tools/base/requirements.txt#L5

im wondering for this case we can make the command customizable in some way

f0rmiga:
@mattem Want to guide @phlax on any specific changes it may need, or could we merge it since Python 3.6 is EOL as you mentioned?

phlax:
@groodt re https://github.com/phlax/rules_python/pull/1#issuecomment-1144195163 i would be happy to add to this PR but as this is a ~breaking change its maybe best to pass this one through first/alone

@f0rmiga any chance of landing this?

its not a total solution to using `rules_python` with dependabot but its a step in the right direction (and i guess to `pip-compiling` more generally)

looks like this needs a main merge, which ill do now...

lmk if there is anything else that needs to be updated

groodt:
@phlax SGTM. Land the merge and this PR and I'll create a fresh new one.

python/pip_install/repositories.bzl
```diff
@@ -29,14 +29,19 @@ _RULE_DEPS = [
     ),
     (
         "pypi__pip_tools",
-        "https://files.pythonhosted.org/packages/6d/16/75d65bdccd48bb59a08e2bf167b01d8532f65604270d0a292f0f16b7b022/pip_tools-5.5.0-py2.py3-none-any.whl",
-        "10841c1e56c234d610d0466447685b9ea4ee4a2c274f858c0ef3c33d9bd0d985",
+        "https://files.pythonhosted.org/packages/ba/42/5a9dfbaac47794c087d115fa275d3248a41614686b84d85cbb0bd146ebe5/pip_tools-6.6.0-py3-none-any.whl",
```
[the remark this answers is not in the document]
phlax: yep, it was late when i merged, added back now

python/pip_install/requirements.bzl
```diff
@@ -60,9 +60,11 @@ def compile_pip_requirements(
     deps = [
         requirement("click"),
         requirement("colorama"),
+        requirement("pep517"),
         requirement("pip"),
         requirement("pip_tools"),
         requirement("setuptools"),
+        requirement("tomli"),
```
thundergolfer, line 67: How it this one used by pip-tools. I checked the install-requires of pip-tools and it isn't there.

thundergolfer (approved):

phlax:
im guessing we want a changelog entry somewhere - as this might break some expectations

python/pip_install/requirements.bzl
groodt: `# via pep517`

thundergolfer: Ah cheers. Didn't expect `pep517` to take any deps.

thundergolfer:
Yes I think this would at least go in the _What's Changed_ section of release notes.

groodt:
Mergeable?

PR #694, opened 2022-04-30
Use "shared" msvc hermetic toolchain

jheaff1:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Using python packages with C modules, e.g. numpy, would fail when using
the "static" toolchain. See 
Issue Number: indygreg/python-build-standalone#124

## What is the new behavior?
When using the "shared" toolchain, importing numpy is successful

- [ ] Yes
- [x] No

### DIFF
--- a/python/repositories.bzl
+++ b/python/repositories.bzl
@@ -117,6 +117,7 @@ filegroup(
     srcs = glob(
         include = [
             "*.exe",
+            "*.dll",
             "bin/**",
             "DLLs/**",
             "extensions/**",
--- a/python/versions.bzl
+++ b/python/versions.bzl
@@ -42,7 +42,7 @@ TOOL_VERSIONS = {
             "aarch64-apple-darwin": "f9a3cbb81e0463d6615125964762d133387d561b226a30199f5b039b20f1d944",
             # no aarch64-unknown-linux-gnu build available for 3.8.12
             "x86_64-apple-darwin": "f323fbc558035c13a85ce2267d0fad9e89282268ecb810e364fff1d0a079d525",
-            "x86_64-pc-windows-msvc": "924f9fd51ff6ccc533ed8e96c5461768da5781eb3dfc11d846f9e300fab44eda",
+            "x86_64-pc-windows-msvc": "4658e08a00d60b1e01559b74d58ff4dd04da6df935d55f6268a15d6d0a679d74",
             "x86_64-unknown-linux-gnu": "5be9c6d61e238b90dfd94755051c0d3a2d8023ebffdb4b0fa4e8fedd09a6cab6",
         },
         "strip_prefix": "python",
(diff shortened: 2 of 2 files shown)


jheaff1:
The build failure is due to the ` //:examples/wheel:filename_escaping` target. If I run ` bazel build -s //:examples/wheel:filename_escaping` to print out the command that is actually run, and then run said command, it seems to run successfully on this new branch but not on main, which is odd

f0rmiga:
> The build failure is due to the ` //:examples/wheel:filename_escaping` target. If I run ` bazel build -s //:examples/wheel:filename_escaping` to print out the command that is actually run, and then run said command, it seems to run successfully on this new branch but not on main, which is odd

Are you referring to the build failure on CI?

jheaff1:
> > The build failure is due to the ` //:examples/wheel:filename_escaping` target. If I run ` bazel build -s //:examples/wheel:filename_escaping` to print out the command that is actually run, and then run said command, it seems to run successfully on this new branch but not on main, which is odd
> 
> Are you referring to the build failure on CI?

Yes

UebelAndre:
@jheaff1 @f0rmiga Howdy folks! Any updates here?

jheaff1:
> @jheaff1 @f0rmiga Howdy folks! Any updates here?

Hey there! I’m on holiday at the moment so don’t have the time to look into this sadly. It might be a few more weeks before I can take a look

phlax:
we are hitting similar problems to the ones resolved here trying to make use of the hermetic toolchain on windows.

the error we are receiving when trying to make use of an egg that requires compilation is:

```console
  error: Microsoft Visual C++ 14.0 or greater is required. Get it with "Microsoft C++ Build Tools": https://visualstudio.microsoft.com/visual-cpp-build-tools/
```

but afaict we do have c++ build tools in the env (ref https://github.com/envoyproxy/envoy/pull/20432#issuecomment-1129747148) so  im guessing this is an issue related to using the static standalone binary

i tested using this PR - sadly it didnt work at all (afaict some kind of file not found issue - perhaps the expected shared libs - struggling to get useful info out of our win CI)

jheaff1:
> @jheaff1 @f0rmiga Howdy folks! Any updates here?

CI now passes

jheaff1:
> we are hitting similar problems to the ones resolved here trying to make use of the hermetic toolchain on windows.
> 
> the error we are receiving when trying to make use of an egg that requires compilation is:
> 
> ```
>   error: Microsoft Visual C++ 14.0 or greater is required. Get it with "Microsoft C++ Build Tools": https://visualstudio.microsoft.com/visual-cpp-build-tools/
> ```
> 
> but afaict we do have c++ build tools in the env (ref [envoyproxy/envoy#20432 (comment)](https://github.com/envoyproxy/envoy/pull/20432#issuecomment-1129747148)) so im guessing this is an issue related to using the static standalone binary
> 
> i tested using this PR - sadly it didnt work at all (afaict some kind of file not found issue - perhaps the expected shared libs - struggling to get useful info out of our win CI)

To fix the CI build I had to modify the BUILD file for the toolchain repo to include the DLLs provided in the standalone python tarball. Does this updated PR resolve your issue?

phlax:
> Does this updated PR resolve your issue?

not exactly, but it works now

we kinda ~fixed the problem already by upgrading our deps - i guess the "fix" was mostly just not needing to compile

i have tested this branch with/out the "fix" - without it i get the same original error about missing `Visual C++` - im wondering if there is something amiss with our win env.

f0rmiga (approved):

f0rmiga:
Thanks for fixing this!

Issue #646, opened 2022-03-10
Pip install uses wrong Macos SDK version from standalone interpreters.


twheys:
Hi, my team discovered an issue trying to use `pip_install` to install python dependencies that use cpython on an M1 Mac with a full xcode installation.

Here's the error from the build logs:
```
  building '_ruamel_yaml' extension
  creating build/temp.macosx-11.0-arm64-3.9
  Compiling with an SDK that doesn't seem to exist: /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.3.sdk
  Please check your Xcode installation
  clang -Wno-unused-result -Wsign-compare -Wunreachable-code -DNDEBUG -g -fwrapv -O3 -Wall -arch arm64 -mmacosx-version-min=11.0 -Wno-nullability-completeness -Wno-expansion-to-defined -Wno-undef-prefix -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.3.sdk -fPIC -I/var/folders/c8/z52g4gc5679gygt8z0dx0jqm0000gn/T/tmpjbwqfg64/tools/deps/include -I/var/folders/c8/z52g4gc5679gygt8z0dx0jqm0000gn/T/tmpjbwqfg64/tools/deps/include/ncursesw -I/var/folders/c8/z52g4gc5679gygt8z0dx0jqm0000gn/T/tmpjbwqfg64/tools/deps/include/uuid -Werror=unguarded-availability-new -g0 -I/private/var/tmp/_bazel_alex/bfcdc3c2b5b01e8ec0b8b8dd179fe4b8/external/r9python_darwin_arm64/include/python3.9 -c _ruamel_yaml.c -o build/temp.macosx-11.0-arm64-3.9/_ruamel_yaml.o
  clang: warning: no such sysroot directory: '/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.3.sdk' [-Wmissing-sysroot]
  In file included from _ruamel_yaml.c:6:
  /private/var/tmp/_bazel_alex/bfcdc3c2b5b01e8ec0b8b8dd179fe4b8/external/r9python_darwin_arm64/include/python3.9/Python.h:25:10: fatal error: 'stdio.h' file not found
  #include <stdio.h>
           ^~~~~~~~~
```

[...]

PR #697, opened 2022-05-09
Patch pip environment with xcode sdk location.

hrfuller:
Fixes issues in #646 by setting CPPFLAGS to the correct xcode sdks.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

If you use one of the hermetic standalone interpreters provided by rules_python pip fails to build sdists with native code extensions because the sysconfig clang flags provided by the standalone interpreters match the interpreter build environment which is almost always different than the execution environment. We fix this issue by providing the correct libc system headers location via CFLAGS env var.

## What is the new behavior?
Pip installing a wheel which needs native code build will succeed.

- [ ] Yes
- [x] No

## Other information
The Windows image in buildkite doesn't have a c++ toolchain available for pip to use to in order to test changes I had to introduce a platform specific requirement into examples/pip_install/requirements.in. In order to appease the platform specific requirements.txt files resulting from compile_pip_requirements I added support for platform specific lock files to compile_pip_requirements. pip-compile can't generate cross platform requirement files from an arbitrary execution platform, so that is left up to the user to generate in their own environments.


mattem:
There were a few other flags I had to override from sysconfig, I'll dig them out.

mattem:
I also found that we needed to override the include paths set for `gcc`:

```
  res = rctx.execute([
      "gcc",
      "-Wp,-v",
      "-xc++",
      "/dev/null",
      "-fsyntax-only",
  ])

  include_paths = [
      paths.normalize(line.replace("(framework directory)", "").strip())
      for line in res.stderr.split("\n")
      if line.startswith(" /usr") or line.startswith(" /Library")
  ]
  include_flags = " ".join(["-I%s" % p for p in include_paths])
```

And set them on `CPPFLAGS`. 

`--sysroot` may also need setting so that it applies to libraries too.

python/pip_install/pip_repository.bzl
```diff
@@ -61,6 +64,27 @@ def _resolve_python_interpreter(rctx):
...
+        rctx.path(Label("@{}//:{}".format(rctx.attr.python_interpreter_target.workspace_name, STANDALONE_INTERPRETER_FILENAME)))
+    ):
+        xcode_sdk_location = rctx.execute(["xcode-select", "-p"])
+        if xcode_sdk_location.return_code == 0:
+            xcode_root = xcode_sdk_location.stdout.strip()
+            if "commandlinetools" not in xcode_root.lower():
+                # This is a full xcode installation somewhere like /Applications/Xcode13.0.app/Contents/Developer
+                # so we need to change the path to to the macos specific tools.
+                xcode_root = "{}/Platforms/MacOSX.platform/Developer".format(xcode_root)
+            environment[CFLAGS] = "-isysroot {}/SDKs/MacOSX.sdk".format(xcode_root)
```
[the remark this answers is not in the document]
hrfuller: Yeah it will, good catch. I believe I was checking for an existing CFLAGS before but must have lost it in some refactoring. My idea was to not overwrite it if CFLAGS was already set in the environment with the assumption that the user knows what they want if they are specifically setting CFLAGS.

hrfuller:
My understanding is that all the standard library includes can be found under the sysroot. 

From [clang docs](https://clang.llvm.org/docs/UsersManual.html#relocatable-pch-files):
>If the headers are installed in some other system root, the -isysroot option can be used provide a different system root from which the headers will be based. For example, -isysroot /Developer/SDKs/MacOSX10.4u.sdk will look for mylib.h in Developer/SDKs/MacOSX10.4u.sdk/usr/include/mylib.h.

Is the difference that when compiling cpp files you need to set CPPFLAGS in addition to CFLAGS? If so does the -isysroot not apply to cpp compiles?

Though the documentation is a bit confusing clang supports `--sysroot` in some fashion though its undocumented. I believe that the [gcc docs](https://gcc.gnu.org/onlinedocs/gcc/Directory-Options.html) provide some answer to why this is:
>-isysroot dir
This option is like the --sysroot option, but applies only to header files (except for Darwin targets, where it applies to both header files and libraries). See the --sysroot option for more information.

`-isysroot` also applies to the library include path on darwin apparently. Since this change only applies to the darwin platform I *think* that -isysroot should be enough.

hrfuller:
@mattem I think I resolved the issues you brought up except for the extra include flags, which I don't think are required based on docs I've read. If there aren't any blockers can we land this and iterate on any edge cases after? Alternatively if you have a repro of an extension that doesn't build with the current PR I could work off of that.

python/pip_install/pip_repository.bzl
```diff
@@ -61,6 +64,28 @@ def _resolve_python_interpreter(rctx):
...
+    Pip won't be able to compile c extensions from sdists with the pre built python distributions from indygreg
+    otherwise. See https://github.com/indygreg/python-build-standalone/issues/103
+    """
+    if (
+        rctx.os.name.lower().startswith("mac os") and
+        rctx.attr.python_interpreter_target != None and
+        rctx.path(Label("@{}//:{}".format(rctx.attr.python_interpreter_target.workspace_name, STANDALONE_INTERPRETER_FILENAME))) and
+        not environment.get(CPPFLAGS)
+    ):
+        xcode_sdk_location = rctx.execute(["xcode-select", "-p"])
```
f0rmiga, line 80:
```suggestion
        xcode_sdk_location = rctx.execute(["xcode-select", "--print-path"])
```

python/pip_install/pip_repository.bzl
```diff
@@ -61,6 +64,28 @@ def _resolve_python_interpreter(rctx):
...
+        if xcode_sdk_location.return_code == 0:
+            xcode_root = xcode_sdk_location.stdout.strip()
+            if "commandlinetools" not in xcode_root.lower():
```
f0rmiga, line 83: `commandlinetools` is a magic string here that could use a const to improve the readability here what this `if` block means. Otherwise, a comment would be good to explain the logic here.

hrfuller:
@f0rmiga I resolved your comments. I tried using `--sysroot` instead of `-isysroot` but it appears to not include the header files in clang when doing so, so we need to use `-isysroot` anyway.

f0rmiga:
> @f0rmiga I resolved your comments. I tried using `--sysroot` instead of `-isysroot` but it appears to not include the header files in clang when doing so, so we need to use `-isysroot` anyway.

Oh, interesting!

Thanks for addressing the comments.

mattem (approved):

Issue #691, opened 2022-04-23
py_binary with hermetic toolchain requires a system interpreter


jpgxs:
# 🐞 bug report

### Affected Rule

`py_binary`

### Description

Running `py_binary` without a system interpreter (using a toolchain configured with `python_register_toolchain`)
fails with the following error:

```
/usr/bin/env: 'python3': No such file or directory
```

After installing a system `python3`, the rule runs fine and uses the correct Python interpreter (not the system one).

### Files

```
# WORKSPACE
load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
http_archive(
    name = "rules_python",
    sha256 = "cdf6b84084aad8f10bf20b46b77cb48d83c319ebe6458a18e9d2cebf57807cdd",
    strip_prefix = "rules_python-0.8.1",
    url = "https://github.com/bazelbuild/rules_python/archive/refs/tags/0.8.1.tar.gz",
)
load("@rules_python//python:repositories.bzl", "python_register_toolchains")
python_register_toolchains(
    name = "python3_10",
    python_version = "3.10",
)
```

```
# BUILD
exports_files(["interp_version.py"])
py_binary(
    name = "interp_version",
    srcs = ["interp_version.py"],
    main = "interp_version.py",
)
```

```
# interp_version.py
import sys
print(sys.version_info)
```

### Repro

Using the `ubuntu:focal` Docker image:

```
root # bazel run //:interp_version
INFO: Analyzed target //:interp_version (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
Target //:interp_version up-to-date:
  bazel-bin/interp_version
INFO: Elapsed time: 0.773s, Critical Path: 0.01s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
INFO: Build completed successfully, 1 total action
/usr/bin/env: 'python3': No such file or directory  <-- Error

root # apt-get install python3-minimal
[...]

root # python3 --version
Python 3.8.10


root # bazel run //:interp_version
INFO: Analyzed target //:interp_version (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
Target //:interp_version up-to-date:
  bazel-bin/interp_version
INFO: Elapsed time: 0.707s, Critical Path: 0.01s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
INFO: Build completed successfully, 1 total action
[...]

PR #699, opened 2022-05-11
Add ability for Bazel >=5.0.0 users to use downloaded py interpreter in py_binary stub script

thundergolfer:
## PR Checklist

- [ ] Docs have been added / updated (for bug fixes / features) **TODO**

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: https://github.com/bazelbuild/rules_python/issues/691

## What is the new behavior?

Bazel users on a version greater than or equal to version 5 will avoid using a system Python interpreter when running `py_binary` or `py_test` targets.

- [ ] Yes
- [X] No


python/repositories.bzl
```diff
@@ -154,6 +168,7 @@ py_runtime(
     files = [":files"],
     interpreter = "{python_path}",
     python_version = "PY3",
+    {stub_shebang_assignment}
```
thundergolfer, line 171: I think this functionality could be behind a default `True` configuration knob. I'm not confident it will be the right move in all scenarios. 

When doing `bazel run` (or `test`) it will work, but if the `py_binary` is moved in any way, for example to layer it into a container, then the absolute path will not be valid. 

With 'in-build' Python toolchains, I'm pretty sure the toolchain interpreter is moved with the rest of the application files. So it can be found at a relative path to the stub_script, for example `foo.runfiles/python39_x86_64-apple-darwin/bin/python3`.

python/repositories.bzl
```diff
@@ -35,6 +35,13 @@ def py_repositories():
 # Remaining content of the file is only used to support toolchains.
 ########
 
+# Parse the bazel version string from `native.bazel_version`.
+# e.g.
+# "0.10.0rc1 abc123d" => (0, 10, 0)
+# "0.3.0" => (0, 3, 0)
+def _parse_native_bazel_version(bazel_version):
+    return tuple([int(n) for n in bazel_version.split(".")])
```
thundergolfer, line 43: I took this from https://github.com/bazelbuild/bazel-skylib/blob/main/lib/versions.bzl#L22 but I don't really want this file to take a dependency on skylib for such simple functionality. 🤷

python/repositories.bzl
```diff
@@ -104,6 +111,12 @@ def _python_repository_impl(rctx):
             fail(exec_result.stderr)
 
     python_bin = "python.exe" if ("windows" in platform) else "bin/python3"
+    print(native.bazel_version)
+    if _parse_native_bazel_version(native.bazel_version) >= (5,0,0):
+        abs_python_bin_path = rctx.path(python_bin)
+        stub_shebang_assignment = "stub_shebang = \"#!{}\"".format(abs_python_bin_path)
```
f0rmiga, line 117:
```suggestion
    if _parse_native_bazel_version(native.bazel_version) >= (5,0,0):
        execroot = "/".join(str(rctx.path(".")).replace("\\", "/").split("/")[:-2])
        python_bin_path = str(rctx.path(python_bin)).removeprefix(execroot + "/")
        shebang_line = "#!./" + python_bin_path
        stub_shebang_assignment = "stub_shebang = '{}',".format(shebang_line)
```

f0rmiga: This will find the binary relative to the execroot.

lummax: I tried something similar, unfortunately this does not work for me (bazel 5.1.0, rules_python 0.9.0) as the python repository is not part of the linux sandbox dependencies when running the `py_binary()`.

Did you run this successfully?

PR #702, opened 2022-05-17
Handle entry_points that end with .py

groodt:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

There are some entry_points in the wild that end with a `.py` suffix. This is an extreme edge-case, but this clashes with the native Python rules in Bazel that forbids `py_binary` to end with `.py`.

Issue Number: N/A

- [ ] Yes
- [x] No

## Other information
An example entry_point in the wild that ends with `.py` https://github.com/jupyterhub/jupyterhub-idle-culler/blob/9dba16033485f3c70352edd3255bb0c5905f16ce/setup.py#L15

### DIFF
--- a/python/pip_install/extract_wheels/lib/bazel.py
+++ b/python/pip_install/extract_wheels/lib/bazel.py
@@ -48,7 +48,7 @@ def generate_entry_point_contents(
     )
 
 
-def generate_entry_point_rule(script: str, pkg: str) -> str:
+def generate_entry_point_rule(name: str, script: str, pkg: str) -> str:
     """Generate a Bazel `py_binary` rule for an entry point script.
 
     Note that the script is used to determine the name of the target. The name of
@@ -56,6 +56,7 @@ def generate_entry_point_rule(script: str, pkg: str) -> str:
     directories within a wheel.
 
     Args:
+        name (str): The name of the generated py_binary.
         script (str): The path to the entry point's python file.
         pkg (str): The package owning the entry point. This is expected to
             match up with the `py_library` defined for each repository.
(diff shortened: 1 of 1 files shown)


python/pip_install/extract_wheels/lib/bazel.py
```diff
@@ -409,13 +409,18 @@ def extract_wheel(
     directory_path = Path(directory)
     entry_points = []
     for name, entry_point in sorted(whl.entry_points().items()):
-        entry_point_script = f"{WHEEL_ENTRY_POINT_PREFIX}_{name}.py"
-        (directory_path / entry_point_script).write_text(
+        # There is an extreme edge-case with entry_points that end with `.py`
+        # See: https://github.com/bazelbuild/bazel/blob/09c621e4cf5b968f4c6cdf905ab142d5961f9ddc/src/test/java/com/google/devtools/build/lib/rules/python/PyBinaryConfiguredTargetTest.java#L174
+        entry_point_without_py = name[:-3] if name.endswith(".py") else name
```
thundergolfer, line 414:
```suggestion
        entry_point_without_py = name.removesuffix(".py")
```

https://docs.python.org/3/library/stdtypes.html#str.removesuffix

thundergolfer (approved):

python/pip_install/extract_wheels/lib/bazel.py
groodt: Incredible as it may seem `New in version 3.9.` and I would prefer this to be compatible with older versions of PY3.

thundergolfer: Bah, so it is.

PR #713, opened 2022-05-25
Don't include pyc files as part of the hermetic toolchain

tomgr:
## PR Checklist

I'm not sure how to test this.

## PR Type

- [X] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Using rules_python with the hermetic toolchain via `python_register_toolchains` results in cache misses when using bazel with a remote cache. This appears to be because pyc files that are generated on the build host are included as part of one of the globs, but since pyc files don't appear to be generated deterministically. For example, from one run we saw one action that used the hermetic toolchain had the following file as an input:
```
        {
          "path": "external/python3_9_x86_64-pc-windows-msvc/lib/__pycache__/_markupbase.cpython-39.pyc",
          "digest": "acc87949c12c0a73eace38bed1a4d9e311ff283fe29a7e28bdfb339ff267ed27"
        },
```
but from another run we saw:
```
        {
          "path": "external/python3_9_x86_64-pc-windows-msvc/lib/__pycache__/_markupbase.cpython-39.pyc",
          "digest": "afb43608c750e3f7d4dcb0c968aec6d84f678cebc506666325b43c24ce964e5a"
        },
```
The hashes above are taken from the bazel execution log as per https://bazel.build/docs/remote-execution-caching-debug.
Note that although the above are with the windows toolchain, we saw the same behaviour on all MacOS, Linux, and Windows.

## What is the new behavior?

pyc files are excluded from the glob. We're no longer seeing cache misses when using a cache.

- [ ] Yes
- [X] No

### DIFF
--- a/python/repositories.bzl
+++ b/python/repositories.bzl
@@ -128,7 +128,8 @@ filegroup(
             "share/**",
         ],
         exclude = [
-            "**/* *", # Bazel does not support spaces in file names.
+            "**/* *",   # Bazel does not support spaces in file names.
+            "**/*.pyc", # These are not deterministically generated, breaking remote caching
         ],
     ),
 )


thundergolfer (approved):

f0rmiga (changes requested):
I first would like to understand why this happens before simply excluding `.pyc` files. To clarify, we download the python toolchain deterministically, then make the directory tree read-only. So, excluding the `.pyc` sounds like burying the real problem.

f0rmiga:
@tomgr You said you see this on all platforms, right? The hash you posted is from the Windows toolchain, which we don't make the tree read-only.

f0rmiga:
@tomgr Could you check if https://github.com/bazelbuild/rules_python/commit/248599047210f1a8f3e9a0efac3bb7e6bbe3c4fb solves the problem under Windows?

f0rmiga:
Sorry, https://github.com/bazelbuild/rules_python/commit/43bdb71b1e387a1aa4d3fc5423deb7e450dda400 should be the one...

f0rmiga:
@tomgr When you say "first run" and then "second run", is there any `bazel clean` involved between them, or different machines? Can you ever get into this state from the same machine without cleaning the Bazel cache?

f0rmiga:
It throws me off that somehow read-only is not taking any effects.
Nevertheless, we could be hitting https://docs.python.org/3/reference/import.html#pyc-invalidation. I have to check if the metadata of those files changes between extraction.

tomgr:
> @tomgr When you say "first run" and then "second run", is there any `bazel clean` involved between them, or different machines? Can you ever get into this state from the same machine without cleaning the Bazel cache?

There's no `bazel clean`, but I was deliberately running them on separate machines to reproduce the issue. I've never noticed it happening from the same machine: it seems that once the pyc files have been modified, they are then stable. Although we run our Linux builds inside Docker containers, we do share the repository cache between the containers. I suspect that removing the shared repository cache would reproduce the same issue.

Is there any extra data that I can gather to help debug why the read-only setting is not working?

mattoberle:
@tomgr any chance your builds are running as `root` inside the Docker containers?
When running as a user with `CAP_DAC_OVERRIDE` the read-only mode on the hermetic toolchain directory is ignored.

alexgartner-bc:
> @tomgr any chance your builds are running as `root` inside the Docker containers? When running as a user with `CAP_DAC_OVERRIDE` the read-only mode on the hermetic toolchain directory is ignored.

Interesting. I'm seeing similar issues but only on our stateless build servers (which are running builds as root) vs our stateful build servers which run builds as uid 1000.

This only started happening after we configured `python_register_toolchains`. I noticed it because our rules_docker python images started permadiffing:

<details>
<summary>container-diff diff</summary>

```
$ container-diff diff --type=history --type=file us-west2-docker.pkg.dev/REDACTED/oci/mock_isv@sha256:45de6b34dddce7c24935a5d2f3706973750d06a9f3d27ad2f8f23a9894036ba2 us-west2-docker.pkg.dev/REDACTED/oci/mock_isv@sha256:ee7655603252efbf1d852f7d065622c6d3017c2bb01d8e2ae516183c5de68b09

-----File-----

These entries have been added to us-west2-docker.pkg.dev/REDACTED/oci/mock_isv@sha256:45de6b34dddce7c24935a5d2f3706973750d06a9f3d27ad2f8f23a9894036ba2: None

These entries have been deleted from us-west2-docker.pkg.dev/REDACTED/oci/mock_isv@sha256:45de6b34dddce7c24935a5d2f3706973750d06a9f3d27ad2f8f23a9894036ba2: None

These entries have been changed between us-west2-docker.pkg.dev/REDACTED/oci/mock_isv@sha256:45de6b34dddce7c24935a5d2f3706973750d06a9f3d27ad2f8f23a9894036ba2 and us-west2-docker.pkg.dev/REDACTED/oci/mock_isv@sha256:ee7655603252efbf1d852f7d065622c6d3017c2bb01d8e2ae516183c5de68b09:
FILE                                                                                                                                                                 SIZE1        SIZE2
/app/src/python/sense/services/mock_isv/image.binary.runfiles/python3_8_x86_64-unknown-linux-gnu/lib/python3.8/__pycache__/zipfile.cpython-38.pyc                    57.3K        57.3K
/app/src/python/sense/services/mock_isv/image.binary.runfiles/python3_8_x86_64-unknown-linux-gnu/lib/python3.8/asyncio/__pycache__/base_events.cpython-38.pyc        49.9K        49.9K
/app/src/python/sense/services/mock_isv/image.binary.runfiles/python3_8_x86_64-unknown-linux-gnu/lib/python3.8/__pycache__/pathlib.cpython-38.pyc                    43.3K        43.3K
/app/src/python/sense/services/mock_isv/image.binary.runfiles/python3_8_x86_64-unknown-linux-gnu/lib/python3.8/json/__pycache__/decoder.cpython-38.pyc               9.7K         9.7K
/app/src/python/sense/services/mock_isv/image.binary.runfiles/python3_8_x86_64-unknown-linux-gnu/lib/python3.8/__pycache__/_markupbase.cpython-38.pyc                7.7K         7.7K
/app/src/python/sense/services/mock_isv/image.binary.runfiles/python3_8_x86_64-unknown-linux-gnu/lib/python3.8/__pycache__/hashlib.cpython-38.pyc                    6.7K         6.7K


-----History-----

Docker history lines found only in us-west2-docker.pkg.dev/REDACTED/oci/mock_isv@sha256:45de6b34dddce7c24935a5d2f3706973750d06a9f3d27ad2f8f23a9894036ba2: None

Docker history lines found only in us-west2-docker.pkg.dev/REDACTED/oci/mock_isv@sha256:ee7655603252efbf1d852f7d065622c6d3017c2bb01d8e2ae516183c5de68b09: None
```

</details>

Sounds like an interim solution might be `capsh --drop=CAP_DAC_OVERRIDE` before the build?

f0rmiga:
A better ida is not to run as root!

phlax:
i use a temporary container locally for working on a project using `rules_python` which currently uses `root` inside

i get this is not "best-practice" but im well aware of the risks and in this case there arent really any

it seems a bit arbitrary to enforce that you cant use `rules_python`  at all as root

Cartman75:
BTw, its seems impossible to run this in gitlab because it forces root. Did someone figure out how to get around that? 

I even tried: 
```    - |
      sudo -H -i -u citests bash -e -x << EOS
      whoami
      cd \$CI_BUILDS_DIR
      ./bazelisk --output_base bazel_cache test //...
      EOS
```
and claimed it was it running as root and it was a no-go.

keith:
FYI there is a setting to opt out of this root warning, https://github.com/bazelbuild/rules_python//commit/e67e7dd719d34d5a13c15b24d0234c1ac753b52d

ph03:
@Cartman75 did you figure out a way to run gitlab CI as non-root without using the `ignore_root_user_error` parameter, which annoyingly results in

```
DEBUG: Rule 'python3_x86_64-unknown-linux-gnu' indicated that a canonical reproducible form can be obtained by dropping arguments ["ignore_root_user_error"]
```

as well as cache misses?

Cartman75:
I did, but we decided not to go this route since you still need the operating system python3 to bring it all together to build a .zip. (or at least that's where we ended up. 
The EOS thing above will work. I think I had to adjust the sudo command or something. I don't remember now. Sorry.

qzmfranklin:
A few CI installations (circle, buildbuddy) by default run with `root`. This would yield errors such as:

https://app.buildbuddy.io/invocation/55238174-54c3-459c-8d80-72722f2d00f9

Setting up `non-root` in those systems aren't very straightforward. Also, it does not look like a responsibility of a user of `rules_python` to have to do this.

SlausB:
Modern docker utilizes user mapping through uidmap to implement rootless setup where root user inside docker nicely maps to host user - finally solving annoying ownership/permission issues.
If you create another user inside docker (one would suggest it as a solution to this issue), then such user won't be mapped to host user anymore.
I am currently trying to buiild/run [biomes-game](https://github.com/ill-inc/biomes-game) inside docker which uses `npm` to run `python` to run `bazel` to run `yarn` which fails with "error: you are root". Reinstalling everything as non-root will be a headache. It's huge mess.

learnsomethingtoday:
@ph03 were you able to find any solution to run this as non-root user via gitlab?

Cartman75:
I can say gitlab is getting closer to allowing it: 
https://gitlab.com/gitlab-org/gitlab/-/merge_requests/137907

learnsomethingtoday:
I see, so do we need to wait until this is rolled out and then proceed?
Or did u find any other bypass to do this..?

Cartman75:
I did do this recently and it seems to be working. This is when I moved to MODULE.bazel
```
python.toolchain(
    configure_coverage_tool = False,
    python_version = "3.11",
)
```
Pointing to `ignore_root_user_error` but posted the rest for context.

learnsomethingtoday:
I'm sorry I don't know where to do this change, can you please let me know regarding that.

Cartman75:
That I won't be able to walk you through. That is a whole thing if you do not understand what I am doing.
https://bazel.build/external/migration
But we are using bzlmod to load our external modules.

ph03:
> @ph03 were you able to find any solution to run this as non-root user via gitlab?

I don't recall the details anymore, but we're happily running gitlab CI without root user by using an image that setup a non-root user like this in the related Dockerfile

```
# Run as developer
RUN \
    useradd --create-home --groups sudo,docker developer && \
    echo "developer:developer" | sudo chpasswd
USER developer
```

DebkanchanSamadder:
This is causing GCP cloud build to fail

layus:
Did anyone ever consider and/or test `chattr +i` to make files and folder immutable, even for root ?

loeffel-io:
Hey, for clarification

I do need to add this

```starlark
python.toolchain(
    configure_coverage_tool = False,
    python_version = "3.11",
)
```

while one of my deps is using `rules_python` for my builds since i do not "use" `rules_python`?

it works btw but this does not feel good

Thanks

PR #727, opened 2022-06-10
feat: cc_import for libpython

f0rmiga:
Exposes a `cc_import` target for libpython. So far, we've only been exposing the headers but on hermetic environments without libpython being found in the system search paths, we need to expose those libraries too.


python/repositories.bzl
```diff
@@ -148,6 +148,16 @@ cc_library(
     ],
 )
 
+cc_import(
+    name = "libpython",
+    hdrs = [":includes"],
+    shared_library = glob([
```
UebelAndre, line 154: Can you make this a select so it can be more explicit? Shouldn't all toolchains be consistent in exposing this?

f0rmiga: Fixed in 043f2e388dcf6854260253ef19e7079d32babbbe.

f0rmiga: Windows is always the weird one.

hrfuller (approved):
LGTM.

PR #729, opened 2022-06-13
fix: versioned shared object for libpython on Linux

f0rmiga:
We need the versioned shared object on Linux rather than the symlink that points to the versioned one.


thundergolfer (commented):
What's the error you encounter?

f0rmiga:
We need the real thing rather than the symlink as I pointed out in the PR description. The symlink just doesn't work.

f0rmiga:
It links against `libpython.so` and then throws `error while loading shared libraries: libpython3.10.so.1.0: cannot open shared object file: No such file or directory` when trying to run.

thundergolfer (approved):

PR #732, opened 2022-06-18
Validation to ensure requirements_lock is pinned.

groodt:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Currently nothing is checking if the `requirements_lock` is fully pinned. It's not technically possible to ensure that a lockfile is fully pinned (without actually attempting to resolve it), but a lightweight approximation is to ensure all individual requirements are pinned (i.e. have `==` or `===`).

## What is the new behavior?
Checks if the requirement is pinned, or throws an error.

- [ ] Yes
- [x] No


python/pip_install/parse_requirements_to_bzl/__init__.py
```diff
@@ -35,8 +35,14 @@ def parse_install_requirements(
         parser.parse(requirements_lock, constraint=False), preprocess(content)
     ):
         if parsed_line.is_requirement:
+            install_req = constructors.install_req_from_line(parsed_line.requirement)
+            if not install_req.is_pinned:
```
thundergolfer, line 39: What about accumulating into an `unpinned_reqs` and then raising the `RuntimeError` at the end, with the following error directive line:

`The following requirements were not pinned: {unpinned_reqs}"""))`

The current way is fail-fast on the first, which is sufficient to communicate "not fully pinned", but gives the user less info about their problem.

thundergolfer (approved):

python/pip_install/parse_requirements_to_bzl/__init__.py
groodt: 👍

anuragkanungo:
Hi, looks like this breaks the ability to specify a hosted whl file in the requirements.in

Having something like this in `requirements.in`
```
https://files.pythonhosted.org/packages/24/93/f5b001dc0f5de84ce049a34ff382032cd9478e1080aa6ac48470fa810577/onnx-1.11.0-cp39-cp39-manylinux_2_12_x86_64.manylinux2010_x86_64.whl
```

generates this in `requirements.lock`
```
onnx @ https://files.pythonhosted.org/packages/24/93/f5b001dc0f5de84ce049a34ff382032cd9478e1080aa6ac48470fa810577/onnx-1.11.0-cp39-cp39-manylinux_2_12_x86_64.manylinux2010_x86_64.whl \
    --hash=sha256:67c6d2654c1c203e5c839a47900b51f588fd0de71bbd497fb193d30a0b3ec1e9
```

which basically then fails with

```
RuntimeError: The `requirements_lock` file must be fully pinned. See `compile_pip_requirements`.
Alternatively, use `pip-tools` or a similar mechanism to produce a pinned lockfile.

The following requirements were not pinned:
onnx@ https://files.pythonhosted.org/packages/24/93/f5b001dc0f5de84ce049a34ff382032cd9478e1080aa6ac48470fa810577/onnx-1.11.0-cp39-cp39-manylinux_2_12_x86_64.manylinux2010_x86_64.whl from https://files.pythonhosted.org/packages/24/93/f5b001dc0f5de84ce049a34ff382032cd9478e1080aa6ac48470fa810577/onnx-1.11.0-cp39-cp39-manylinux_2_12_x86_64.manylinux2010_x86_64.whl
)
```


Seems like the check provided by pip, looks for `==` and `===` but not `@`.
https://github.com/pypa/pip/blob/main/src/pip/_internal/req/req_install.py#L252


Unless, I am missing something, it's breaking backward compatibility.

PR #739, opened 2022-06-30
refactor: move all re-exports to private/reexports.bzl

mattem:
Moves all re-exported symbols in to `reexports.bzl`. 

This is a needed prefactor step for https://github.com/bazelbuild/rules_python/pull/723 as we need to consume `py_library` internall in the macro, and doing so from `defs.bzl` would introduce a cycle.


python/defs.bzl
```diff
@@ -21,26 +21,22 @@ migrated to Starlark, their implementations will be moved here.
```
alexeagle, line 21: the comment above feels like it should go to the new file, and the comment here should just be "public API" or sth

alexeagle (approved):

PR #743, opened 2022-07-04
Flatten python packages

groodt:
This PR attempts to simplify the `pip_install / pip_parse` python source packages. The code had grown organically and I found it difficult to follow and unnecessarily nested into packages. 

In general, the code should be quite simple:
1. A tool to install all wheels eagerly in a requirements.txt (pip_install)
2. A tool to parse a requirements.txt to a bzl file (pip_parse & parse_requirements_to_bzl)
3. A tool to install a single wheel (pip_parse & whl_library)

There is some additional utility around annotations to bazel rules, extracting wheels and updating namespace_pkgs, but I still believe the code doesn't need to be so fragmented across multiple packages. There is a lot of very similar / duplicated logic when extracting wheels that is probably unnecessary. 

This PR simply moves existing code and imports around and should be fully backwards compatible.

Given that `pip_parse` seems to be the more popular choice (based on discussions and most of the recent GH issues) it is probably justified to remove the eager installation logic (pip_install) and standardise on `pip_parse`. It should be possible to run with both aliases. I'll do this in a follow-up refactor.


hrfuller (approved):
LGTM. I agree it was getting ungainly and all the dunder files were ugly. Can you please follow up after squashing and merging, by adding this commit to `//:.git-blame-ignore-revs`

PR #744, opened 2022-07-05
chore: support bzlmod

alexeagle:
Part of https://github.com/bazelbuild/bazel-central-registry/issues/124


examples/bzlmod/__init__.py
```diff
@@ -0,0 +1,8 @@
+# TODO: bzlmod should grant access to pip_install dependencies as well
```
f0rmiga, line 1: Do we need to leave this TODO (not possible yet?) here or should we resolve before merging?

MODULE.bazel
```diff
@@ -0,0 +1,17 @@
+module(
+    name = "rules_python",
+    version = "0.4.0",
```
f0rmiga: This is not the current release version.

f0rmiga: It perhaps makes sense to set this to 0.0.0 and automatically commit from CI when doing a release?

f0rmiga:
Your force-push ate my comments.

examples/bzlmod/__init__.py
alexeagle: I'd like to get this much in since it's feature parity for what bazel-central-registry currently provides (for an old version of rules_python)

I think it will take a bit of research and tinkering to see how pip deps can be exposed too.

MODULE.bazel
alexeagle: yes! actually 0.0.0 can stay here, the file in the registry is modified

examples/bzlmod/MODULE.bazel
```diff
@@ -0,0 +1,12 @@
+module(
+    name = "example_bzlmod",
+    compatibility_level = 1,
+    version = "0.0.0",
+)
+
+bazel_dep(name = "rules_python", version = "0.0.0")
+
+local_path_override(
```
f0rmiga, line 9: Interesting

alexeagle: you like and approve? :)

f0rmiga: I don't have enough experience on the modules to tell. My first reaction is that I like.

f0rmiga (approved):
Looks reasonable. Feel free to merge and I foresee it getting some tweaks in the near future.

PR #748, opened 2022-07-12
Fix remote cache miss

jheaff1:
In some cases, the hermetic python toolchain would generate .pyc files
when being used, resulting in a remote cache miss. This commit excludes
generated files from the "files" filegroup to resolve this issue

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/python/repositories.bzl
+++ b/python/repositories.bzl
@@ -142,6 +142,7 @@ filegroup(
         ],
         exclude = [
             "**/* *", # Bazel does not support spaces in file names.
+            "**/__pycache__/**",
         ],
     ),
 )


hrfuller (approved):
LGTM. Could you please clarify when the pyc's are generated. Is it during some build step of the hermetic interpreter external repo?

f0rmiga:
Please, see: https://github.com/bazelbuild/rules_python/pull/713.

jheaff1:
After reading #713, it seems that the issue shouldn’t occur if not building as root. I was indeed building as the root user “ContainerAdministrator” within a Windows Container. I’ll try building as non-root. Thanks!

f0rmiga:
Cool! I'm working on a solution to warn the user about this.

f0rmiga:
@jheaff1 I created https://github.com/bazelbuild/rules_python/pull/749. I'll add the same functionality under Windows next.

jheaff1:
@f0rmiga This issue does indeed go away when building as non-root, however when building as non-root in a Windows Container, symlinks cannot be created, which prevents the usage of features like `--enable_runfiles` (See [here](https://docs.bazel.build/versions/main/windows.html#enable-symlink-support)), which is used in rules_python in fact:

https://github.com/bazelbuild/rules_python/blob/aef17ad72919d184e5edb7abf61509eb78e57eda/.bazelrc#L19-L22

Given this scenario, and perhaps others where it is troublesome to build as non-root, should we instead exclude the pyc files from the filegroup?

PR #750, opened 2022-07-12
safely check for standalone interpreter sentinel files

hrfuller:
## PR Checklist
Fixes #746 
Please check if your PR fulfills the following requirements:

- [ ] Tests for the changes have been added (for bug fixes / features)
     Tests omitted because we don't have a builtin interpreter which is not of the standalone interpreter variety. 
     Unfortunately we can't unit test these bzl files that use the rctx.
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type

What kind of change does this PR introduce?

- [x] Bugfix
- [ ] Feature (please, look at the "Scope of the project" section in the README.md file)
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] Other... Please describe:

## What is the current behavior?

Issue Number: N/A

## What is the new behavior?
Fixes regression to using custom python_interpreter_target 

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

PR #753, opened 2022-07-13
Fix logical error in pip_repository.bzl

dflems:
`rctx.attr.python_interpreter_target.workspace_name` was being accessed even if `rctx.attr.python_interpreter_target` was unset. This change returns from the `_maybe_set_xcode_location_cflags` function early if the preconditions aren't met.

Looks like this was very-recently merged in #750


dflems:
cc @hrfuller @groodt

hrfuller:
Fix is here #755

PR #757, opened 2022-07-14
Rename pip_parse to pip_install

groodt:
This is a fairly large refactor and unfortunately some breaking changes. The intent here is to reduce complexity and duplication.

Historically, we had 2 rules to install packages from PyPI (`pip_install` and `pip_parse`), with `pip_install` being the original. It worked well, but had a significant flaw in that it would eagerly fetch and install all packages. This was slow and frustrating when there were a large number of dependencies, so `pip_parse` was born!

`pip_parse` (even though it is still a repository rule) lazily fetched and installed packages. This was adopted by most of the community based on evidence from issues raised, the bazel Slack channels and the rule maintainers.

Maintaining both versions should no longer be necessary. There was a lot of duplication in both versions and it required extra effort and consideration for adding any new features.

I've taken the approach (probably 🌶️ ) that `pip_install` is a better name, so I've introduced this change where both `pip_install` and `pip_parse` share the same code, but with a small shim to handle the `requirements_lock` vs `requirements` difference in argument. I think a deprecation log message should be added and then pip_parse should be removed.

2 Breaking Changes:

1. The small breaking change in this PR, will be for the users (probably a small number) who still use a pip_install version prior to this PR *and* who *DO NOT* use the `requirement()` macro to reference dependencies. The naming style of the external repositories has changed, so this will require a `sed` across any `.bazel` files if these users upgrade. This isn't a big deal in my opinion.


groodt:
Closing. Proceeding with smaller and less extreme approach here. https://github.com/bazelbuild/rules_python/pull/807

Issue #709, opened 2022-05-19
bazel run //:gazelle doesn't work with imports with format ("from x import y")


sramirezmartin:
# 🐞 bug report

### Affected Rule

 The issue is caused by the rule: 
<pre><code>
gazelle(
    name = "gazelle",
    data = GAZELLE_PYTHON_RUNTIME_DEPS,
    gazelle = "@rules_python//gazelle:gazelle_python_binary",
)
</code></pre>

### Is this a regression?

Unsure about this.

### Description

When importing a google.cloud modules in the formar `from google.cloud import aiplatform`, we get the following error when running `bazel run //:gazelle`.

<pre><code>
gazelle: ERROR: failed to validate dependencies for target "@toy_build_file_generation_example//subdir1/subdir2": "google.cloud" at line 4 from "subdir1/subdir2/test2.py" is an invalid dependency: possible solutions:
        1. Add it as a dependency in the requirements.txt file.
        2. Instruct Gazelle to resolve to a known dependency using the gazelle:resolve directive.
        3. Ignore it with a comment '# gazelle:ignore google.cloud' in the Python file.

</code></pre>

But when we use the format `import google.cloud.aiplatform as aiplatform`, the command `bazel run //:gazelle` finishes successfully and the BUILD files are updated accordingly. 

## 🔬 Minimal Reproduction

Here is a link to a toy example where the issue can be reproduced: https://github.com/sramirezmartin/gazelle-toy-example

The commands to run are:
<pre><code>
git clone https://github.com/sramirezmartin/gazelle-toy-example.git
cd gazelle-toy-example
pip-compile --generate-hashes --output-file=subdir1/requirements_lock.txt subdir1/requirements.txt
bazel run //subdir1:gazelle_python_manifest.update
bazel run //:gazelle
</code></pre>

## 🔥 Exception or Error

<pre><code>
gazelle: ERROR: failed to validate dependencies for target "@toy_build_file_generation_example//subdir1/subdir2": "google.cloud" at line 4 from "subdir1/subdir2/test2.py" is an invalid dependency: possible solutions:
        1. Add it as a dependency in the requirements.txt file.
        2. Instruct Gazelle to resolve to a known dependency using the gazelle:resolve directive.
        3. Ignore it with a comment '# gazelle:ignore google.cloud' in the Python file.

</code></pre>

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
NAME="Ubuntu"
[...]

PR #760, opened 2022-07-17
709/from imports

aptenodytes-forsteri:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

- `from foo import bar` adds `//foo` as a dep instead of `//foo/bar`.
- `from foo.bar import baz` adds `//foo/bar` as a dep instead of `//foo/bar:baz` (when `//foo/bar:baz` is a separate py_library that contains foo/bar/baz.py as a source).
- `from google.cloud import aiplatform` does not add `"@gazelle_python_test_google_cloud_aiplatform//:pkg",` as a dep (fails to resolve `from` imports for pip packages).

Issue Number: #709 

## What is the new behavior?
When gazelle encounters `from foo.bar import baz`, try to resolve, in order from most specific to least specific:

```
foo.bar.baz
foo.bar
foo
```

This way, if `foo.bar.baz` is in the manifest file or in the index of existing rules (e.g. from a rule that contains the source file foo/bar/baz, it can be found). If `foo.bar.baz` can't be resolved, try `foo.bar` (maybe `baz` is just a function or variable in `foo/bar.py`).

- [ ] Yes
- [x] No - all existing tests continue to pass.

## Other information
May want to consider antagonistic cases - are there cases where imports are truly ambiguous?


gazelle/testdata/invalid_imported_module/test.yaml
```diff
@@ -2,7 +2,8 @@
 expect:
   exit_code: 1
   stderr: |
-    gazelle: ERROR: failed to validate dependencies for target "//:invalid_imported_module": "grpc" at line 2 from "__init__.py" is an invalid dependency: possible solutions:
+    gazelle: Failed to find any valid modules from [grpc]. Errors: ["grpc" at line 2 from "__init__.py" is an invalid dependency: possible solutions:
```
aptenodytes-forsteri, line 5: I can clean up this error message

f0rmiga: Yeah, I wouldn't expect it to change.

aptenodytes-forsteri: Fixed.

f0rmiga (approved):
Thanks!!

PR #758, opened 2022-07-16
Exclude static libraries and tests to reduce zipped Python executables

tetsuok:
## PR Checklist

Please check if your PR fulfills the following requirements:

- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type

What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature (please, look at the "Scope of the project" section in the README.md file)
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [x] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] Other... Please describe:

## What is the current behavior?

Issue Number: N/A

Build artifacts by `bazel build` with the `--build_python_zip` option are unnecessary big. The size of the artifacts is at least 90MB (e.g., a zipped python binary which just prints "hello, world"). This is not great when build artifacts are included to docker images. The cause of this size issue is that the hermetic Python toolchain include the static libraries of the Python toolchain and tests for the standard libraries even though majority of users of rules_python don't need them. These files occupy 47MB (52% of an artifact of size 90MB).

## Repro steps

`WORKSPACE`:
```starlark
load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")

http_archive(
    name = "rules_python",
    sha256 = "a3a6e99f497be089f81ec082882e40246bfd435f52f4e82f37e89449b04573f6",
    strip_prefix = "rules_python-0.10.2",
    url = "https://github.com/bazelbuild/rules_python/archive/refs/tags/0.10.2.tar.gz",
)

load("@rules_python//python:repositories.bzl", "python_register_toolchains")

python_register_toolchains(
    name = "python3_10",
    python_version = "3.10.4",
)
```

`BUILD`:
```starlark
load("@rules_python//python:defs.bzl", "py_binary")

py_binary(
    name = "hello",
    srcs = ["hello.py"],
)
```

`hello.py`:
```python
print("hello, world!")
```

```shell
$ bazel build --build_python_zip //:hello
$ ls -lh bazel-bin/hello.zip
-r-xr-xr-x 1 t docker 91M Jul 17 11:55 bazel-bin/hello.zip
```

You can see some static libraries are included in the zip file (we have duplicated `libpython3.10.a`. The file size (uncompressed) of the static library is 38MB):

```shell
$ unzip -l bazel-bin/hello.zip | sort -n | grep ".a$"
[...]

PR #707, opened 2022-05-18
python: Update minor toolchain versions

phlax:
## PR Checklist

Please check if your PR fulfills the following requirements:

- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type

What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature (please, look at the "Scope of the project" section in the README.md file)
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [X] Other... Please describe:

Updates the python toolchain version for 3.10 -> 3.10.4

## What is the current behavior?

3.10.2 is installed

Issue Number: N/A

## What is the new behavior?

3.10.4 is installed

## Does this PR introduce a breaking change?

- [X] Yes
- [ ] No

if you have 3.10.2 hardcoded as your version it would break

not sure if it is preferable to add 3.10.4 in addition rather than replacing existing version

PR #779, opened 2022-08-04
Update Python minor toolchain versions to allow smaller Python binaries

tetsuok:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Updates the Python toolchain versions for 3.8.13, 3.9, and 3.10. Adds 3.9.13 and 3.10.6. The `install_only` archives of 3.8.13, 3.9.13, and 3.10.6 are smaller than before, because the static libpython libraries and tests for the stdlib itself are removed from the archives. This is essentially equivalent to what we do in #758, but now in the python-build-standalone side. We can remove our own filtering introduced in #758 once the support of earlier versions than 3.8.13, 3.9.13, and 3.10.6 is dropped.
For details of the toolchain side changes, see https://github.com/indygreg/python-build-standalone/releases/tag/20220802.

Issue Number: N/A

For 3.9, 3.9.12 is installed.
For 3.10, 3.10.4 is installed.

## What is the new behavior?

For 3.9, 3.9.13 is installed.
For 3.10, 3.10.6 is installed.

- [x] Yes
- [ ] No

When user sets `python_version = "3.9"` in `python_register_toolchains`, 3.9.13 is installed.
When user sets `python_version = "3.10"` in `python_register_toolchains`, 3.10.6 is installed.

### DIFF
--- a/python/versions.bzl
+++ b/python/versions.bzl
@@ -48,13 +48,13 @@ TOOL_VERSIONS = {
         "strip_prefix": "python",
     },
     "3.8.13": {
-        "url": "20220502/cpython-{python_version}+20220502-{platform}-{build}.tar.gz",
+        "url": "20220802/cpython-{python_version}+20220802-{platform}-{build}.tar.gz",
         "sha256": {
-            "aarch64-apple-darwin": "ba604867d8c6e0a1a85b1be789cad4c69af8a1699043f51e8a85998b55979127",
+            "aarch64-apple-darwin": "ae4131253d890b013171cb5f7b03cadc585ae263719506f7b7e063a7cf6fde76",
             # no aarch64-unknown-linux-gnu build available for 3.8.13
-            "x86_64-apple-darwin": "52e3541f41d165002e19a60e11dcc145e90ffe1fe8a6f94b17d5b72a674674ea",
-            "x86_64-pc-windows-msvc": "1435b77b4d89f2a99719918fcf917a67cd711cad46f67516bed2462d18fbefb3",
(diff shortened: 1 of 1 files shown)


python/versions.bzl
```diff
@@ -102,13 +113,24 @@ TOOL_VERSIONS = {
...
+    },
 }
 
 # buildifier: disable=unsorted-dict-items
 MINOR_MAPPING = {
     "3.8": "3.8.13",
-    "3.9": "3.9.12",
-    "3.10": "3.10.4",
+    "3.9": "3.9.13",
+    "3.10": "3.10.6",
```
mattem, line 133: I guess this PR is a superset of https://github.com/bazelbuild/rules_python/pull/785. I'm not sure I agree with changing the minor mappings and silently bumping users versions. Thinking about it again though they explicitly set the version with no patch, so maybe it is okay...?

tetsuok: @mattem I changed this by following the previous similar PR #707.

tetsuok: I'm totally fine to remove this minor mapping change if separate discussion is needed on when to / how to update minor mappings.

mattem: Ah okay, we've set a precedence that these get updated previously. I think we are okay here then 👍

mattem (approved):

python/versions.bzl
tetsuok: Sure. Thanks for the review!

PR #781, opened 2022-08-04
Added importlib + reqs to pip-compile for python < 3.8

Faqa:
## PR Type

- [X] Bugfix
- [ ] Build related changes
- [ ] CI related changes

You cannot run the `compile_pip_requirements.update` target with a Python interpreter at a version < 3.8 - it will fail on importing `importlib_metadata` in the `build` package

## What is the new behavior?
You can run the `compile_pip_requirements.update` target with a Python interpreter at a version < 3.8 

- [ ] Yes
- [X] No

## Other information

I could see a scenario where part of the setup instructions involve telling the user to produce a lockfile for `pip-compile` from their local env, but it's probably just easier to package up all the reqs for any version.

### DIFF
--- a/python/pip_install/requirements.bzl
+++ b/python/pip_install/requirements.bzl
@@ -76,6 +76,9 @@ def compile_pip_requirements(
         requirement("pip_tools"),
         requirement("setuptools"),
         requirement("tomli"),
+        requirement("importlib_metadata"),
+        requirement("zipp"),
+        requirement("more_itertools"),
     ]
 
     attrs = {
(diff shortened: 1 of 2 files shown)


python/pip_install/requirements.bzl
```diff
@@ -76,6 +76,9 @@ def compile_pip_requirements(
         requirement("pip_tools"),
         requirement("setuptools"),
         requirement("tomli"),
+        requirement("importlib_metadata"),
+        requirement("zipp"),
+        requirement("more_itertools"),
```
mattem, line 81: Are these all noops in higher language version? I know `import_metadata` is. I'm surprised this still works with versions like 3.6 given the version of pip-tools used here is now incompatible with that version.

Faqa: I can't speak for 3.6, but this rule definitely doesn't work for 3.7, which is the version we're trying to work with. Far as I can see, the other two aren't nops, but they should be compatible with the relevant interfaces.

mattem: Yeah I guess we've dropped 3.6 support here given it's EoL, so that's fine. Thanks for checking!

mattem (approved):

PR #784, opened 2022-08-08
chore: remove distribution filegroups

alexeagle:
Since https://github.com/bazelbuild/rules_python/pull/587 we just distribute the whole source archive from GitHub

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -22,27 +22,6 @@ exports_files([
     "version.bzl",
 ])
 
-filegroup(
-    name = "distribution",
-    srcs = [
-        "BUILD",
-        "MODULE.bazel",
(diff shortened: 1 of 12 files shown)


aiuto:
I still prefer human built artifacts to the auto-generated github ones, so that we can track usage, but the rest of this is sound.
LGTM.
Good question who the right reviewers should be. brandjon@ has not really worked on Python in 3 years. lberki@ is not really the right one either.

alexeagle:
ah, I forgot, these are still required for integration testing https://github.com/bazel-contrib/rules_bazel_integration_test#1-declare-a-filegroup-to-represent-the-parent-workspace-files

PR #807, opened 2022-08-27
Standardise on pip_parse

groodt:
Less extreme version of the proposal outlined here:
https://github.com/bazelbuild/rules_python/pull/757

<hr />

# What is changing?
This PR standardises on `pip_parse` behaviour as the only mechanism for installation of 3rd-party python packages. The internals of `pip_install` have been changed to use `pip_parse` behind the scenes, with a small compatibility shim over rule attribute names to ensure minimal disruption to any existing users of `pip_install`.

# Why is it changing?
Historically, we had 2 rules to install packages from PyPI (pip_install and pip_parse), with pip_install being the original. It worked well, but had a significant flaw in that it would eagerly fetch and install all packages. This was slow and frustrating when there were a large number of dependencies, so pip_parse was born!

pip_parse (even though it is still a repository rule) lazily fetched and installed packages. This was adopted by most of the community based on evidence from issues raised, the bazel Slack channels and the rule maintainers.

Maintaining both versions should no longer be necessary. There is a lot of duplication in both versions and it requires extra effort and time. This increases maintenance burden and significantly slows progress to improve the rules.

# Breaking Changes:

* A small breaking change in this PR, will be for users (probably a small number) who still use a pip_install version prior to this PR and who DO NOT use the requirement() macro to reference dependencies. The naming style of the external repositories has changed, so this will require a sed across any .bazel files if these users upgrade. This isn't a big deal in my opinion.
* Users of `pip_install` will need to add the calls as you do with `pip_parse` e.g.
```
load("@pip//:requirements.bzl", "install_deps")

install_deps()
```


groodt:
@f0rmiga for early review before opening up to other maintainers.

f0rmiga:
The breaking change is acceptable. Gazelle can produce the right label for users who rely on it. I have not reviewed the code yet, thought. Will do next.

docs/pip.md
```diff
@@ -73,88 +73,24 @@ Annotations to apply to the BUILD file content from package generated from a `pi
...
 
 **PARAMETERS**
 
 
 | Name  | Description | Default Value |
 | :-------------: | :-------------: | :-------------: |
-| requirements |  A 'requirements.txt' pip requirements file.   |  <code>None</code> |
-| name |  A unique name for the created external repository (default 'pip').   |  <code>"pip"</code> |
-| kwargs |  Additional arguments to the [<code>pip_repository</code>](./pip_repository.md) repository rule.   |  none |
+| requirements |  <p align="center"> - </p>   |  <code>None</code> |
```
f0rmiga, line 83: We may want to preserve the descriptions here.

python/pip.bzl
```diff
@@ -21,88 +21,11 @@ compile_pip_requirements = _compile_pip_requirements
 package_annotation = _package_annotation
 
 def pip_install(requirements = None, name = "pip", **kwargs):
-    """Accepts a `requirements.txt` file and installs the dependencies listed within.
+    # pip_install is now considered deprecated.
```
f0rmiga, line 24: I think we should throw a deprecation warning message here and leave it for only 1 release. Then we delete pip_install.

python/pip.bzl
```diff
@@ -203,10 +127,14 @@ def pip_parse(requirements_lock, name = "pip_parsed_deps", **kwargs):
     # Just in case our dependencies weren't already fetched
     pip_install_dependencies()
 
+    # Temporary compatibility shim
+    # pip_install was previously document to use requirements while pip_parse was using requirements_lock
+    # We would prefer everyone move to using requirements_lock, but we maintain a temporary shim
```
f0rmiga, line 132:
```suggestion
    # Temporary compatibility shim.
    # pip_install was previously document to use requirements while pip_parse was using requirements_lock.
    # We would prefer everyone move to using requirements_lock, but we maintain a temporary shim.
```

python/pip_install/extract_wheels/extract_single_wheel.py
```diff
@@ -8,11 +8,32 @@
...
+def configure_reproducible_wheels() -> None:
+    """Modifies the environment to make wheel building reproducible.
+    Wheels created from sdists are not reproducible by default. We can however workaround this by
+    patching in some configuration with environment variables.
+    """
+
+    # wheel, by default, enables debug symbols in GCC. This incidentally captures the build path in the .so file
+    # We can override this behavior by disabling debug symbols entirely.
+    # https://github.com/pypa/pip/issues/6505
+    if "CFLAGS" in os.environ:
```
f0rmiga, line 22: We should be careful with environment variables for the C and C++ compilers. Ideally, we should instruct users to set these instead of doing them ourselves. Setting `-g0` is only part of making them deterministic.

If we determine that we actually want to mess with them, then setting `CPPFLAGS` is where we want to start.

groodt: I agree. I'm not really confident in these either. My personal preference would be to not have any sdist being built in repository rules at all and encourage "binary-only" wheel installations instead.

All I've done here is to bring over the existing function that was already being called in both `pip_install` and `pip_parse` that was previously in extract_wheels.py (which will be deleted in this PR).

groodt:
Any further comments @f0rmiga ?

python/pip_install/extract_wheels/extract_single_wheel.py
f0rmiga: It's good to keep it this way for now. I thought I had commented on this already. We can make improvements to this part on a follow-up PR.

f0rmiga (approved):

alexeagle:
Will users run into a problem with passing uncompiled requirements to pip_parse? Do you get an error if your requirements.txt just contains "pep8" or something?

groodt:
> Will users run into a problem with passing uncompiled requirements to pip_parse? Do you get an error if your requirements.txt just contains "pep8" or something?

Yes, folks will get a clear error message instructing them what to do:
```
The `requirements_lock` file must be fully pinned. See `compile_pip_requirements`.
Alternatively, use `pip-tools` or a similar mechanism to produce a pinned lockfile.

The following requirements were not pinned:
```

python/pip.bzl
```diff
@@ -92,17 +42,11 @@ def pip_install(requirements = None, name = "pip", **kwargs):
...
-
-    pip_repository(
-        name = name,
-        requirements = requirements,
-        repo_prefix = "pypi__",
-        **kwargs
-    )
+    # buildifier: disable=print
+    print("pip_install is deprecated. Please switch to pip_parse. pip_install will be removed in a future release.")
+    pip_parse(requirements = requirements, name = name, **kwargs)
```
ArmaanT, line 47: Can you do something like `pip_parse(requirements_lock = requirements, ...` here so that you don't need to add a requirements parameter to `pip_parse`?

I'm not at all involved with `rules_python`, just excited about the feature!

groodt:
> So its almost certain a requirements.txt that was used with pip_install won't work with pip_parse, if pip_install is an alias of pip_parse.

Can you explain what you mean? This is literally what we did at $dayjob and it worked just fine. The recommendation has always been to use a fully pinned requirements.txt and I was the person who added the checks for it in pip_parse. 

It's unlikely somebody will have a partially fly pinned requirements.txt and we can easily add that to the release notes or error message. Whats your recommendation?

hrfuller:
My point has to do with fully-transitively resolved and pinned lock file (what we expect in pip_parse) vs a fully-pinned requirements.txt that only has top level distributions. pip_install would run pip to resolve the transitive set, where are pip parse expects the requirements file to already by fully transitively resolved hence the compile-pip-requirements thing. See https://github.com/bazelbuild/rules_python/blob/main/python/pip_install/extract_wheels/extract_single_wheel.py#L41 --no-deps is the difference.

groodt:
> a fully-pinned requirements.txt that only has top level distributions

Got it. Do you expect those to be common? It's quite unlikely those individuals will have a good bazel experience because their results would be non-deterministic as resolution results would change when different package versions are published (of the non-direct dependencies).

Seems like a possible, but unlikely (and generally u supportable?) situation to me.

groodt:
Is the suggestion you are making @hrfuller to rather NOT try to reuse the existing requirements.txt (even though many will be transitively locked)? I can get behind that. I don't mind. I just thought it could help users with existing requirements.txt that are locked (which has also been the recommendation and most should have or they would see lots of cache misses and non-determinism).

hrfuller:
>Is the suggestion you are making @hrfuller to rather NOT try to reuse the existing requirements.txt

Not really, just pointing out that the current solution won't work for some users. Maybe we should just document that fact and move on. I'm sure it's not common at large companies to use an unlocked requirements file, but it is much more tenable for small repos.

groodt:
> Maybe we should just document that fact

@hrfuller Ok, that works for me. Happy if I just add a note to the depreciation paragraph in README.md?

groodt:
@hrfuller I've added a small note to the README. LGTY?

```
The maintainers have taken all reasonable efforts to faciliate a smooth transition, but some users of `pip_install` will
need to replace their existing `requirements.txt` with a fully resolved set of dependencies using a tool such as
`pip-tools` or the `compile_pip_requirements` repository rule.
```

alexeagle:
Cool, @groodt we should make sure the next release includes a breaking change section in the release notes as well to make this discoverable

Issue #812, opened 2022-08-28
Output generated by gazelle_python_manifest is not deterministic


abhishekshivanna:
# 🐞 bug report

### Affected Rule

 The issue is caused by the rule:  
`load("@rules_python//gazelle/manifest:defs.bzl", "gazelle_python_manifest")`

### Is this a regression?

 Don't think so, probably always existed

### Description

When invoking the `update` rule on `gazelle_python_manifest` the output generated to the `gazelle_python.yaml` mapping file is not deterministic. This non-determinism causes the `test` rule (`gazelle_python_manifest.test`) to always fail.

## 🔬 Minimal Reproduction
I've created a sample repository with a dependency that triggers this bug: https://github.com/abhishekshivanna/rules_python_bug

You can run `bazel run //:gazelle_python_manifest.update` to see that the `gazelle_python.yaml` constantly changes on subsequent runs.

## 🔥 Exception or Error

<pre>

bazel run //:gazelle_python_manifest.test
INFO: Analyzed target //:gazelle_python_manifest.test (8 packages loaded, 213 targets configured).
INFO: Found 1 target...
Target //:gazelle_python_manifest.test up-to-date:
  bazel-bin/gazelle_python_manifest.test
INFO: Elapsed time: 0.707s, Critical Path: 0.44s
INFO: 6 processes: 4 internal, 2 darwin-sandbox.
INFO: Build completed successfully, 6 total actions
INFO: Build completed successfully, 6 total actions
exec ${PAGER:-/usr/bin/less} "$0" || exit 1
Executing tests from //:gazelle_python_manifest.test
-----------------------------------------------------------------------------
2022/08/28 19:36:37 ERROR: "/Users/<redacted>/code/rules_python_bug/gazelle_python.yaml" is out-of-date, follow the intructions on this file for updating.
</pre>

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
Darwin Kernel Version 21.3.0: Wed Jan  5 21:37:58 PST 2022; root:xnu-8019.80.24~20/RELEASE_ARM64_T6000 arm64
  </code>
</pre>

**Output of `bazel version`:**

<pre>
  <code>
Bazelisk version: development
Build label: 5.2.0
Build target: bazel-out/darwin_arm64-opt/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Tue Jun 7 16:11:09 2022 (1654618269)
Build timestamp: 1654618269
[...]

PR #813, opened 2022-08-28
Make gazelle_python_manifest output deterministic

abhishekshivanna:
When invoking the update rule on `gazelle_python_manifest`
the output generated to `gazelle_python.yaml` mapping file 
is not deterministic.
This non-determinism causes the test rule for `gazelle_python_manifest`
to always fail. 
This patch fixes the bug by sorting the keys before writing it to the
mapping file, making the output deterministic

Test Plan: Existing unit tests to verify the output

Fixes https://github.com/bazelbuild/rules_python/issues/812

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: https://github.com/bazelbuild/rules_python/issues/812

## What is the new behavior?
See commit summary

- [ ] Yes
- [x] No


gazelle/manifest/manifest.go
```diff
@@ -94,11 +95,31 @@ func (f *File) Decode(manifestPath string) error {
 	return nil
 }
 
+// ModulesMapping is the type used to map from importable Python modules to
+// the wheel names that provide these modules
```
f0rmiga, line 99:
```suggestion
// ModulesMapping is the type used to map from importable Python modules to
// the wheel names that provide these modules.
```

gazelle/manifest/manifest.go
```diff
@@ -94,11 +95,31 @@ func (f *File) Decode(manifestPath string) error {
...
+type ModulesMapping map[string]string
+
+// MarshalYAML makes sure that we sort the module names before marshalling
+// the contents of `ModulesMapping` to a yaml file. This ensures that the
+// file is deterministically generated from the map
```
f0rmiga, line 104:
```suggestion
// MarshalYAML makes sure that we sort the module names before marshaling
// the contents of `ModulesMapping` to a YAML file. This ensures that the
// file is deterministically generated from the map.
```

gazelle/manifest/manifest.go
```diff
@@ -94,11 +95,31 @@ func (f *File) Decode(manifestPath string) error {
...
+func (m ModulesMapping) MarshalYAML() (interface{}, error) {
+	var mapslice yaml.MapSlice
+	var keys []string
+	for key := range m {
+		keys = append(keys, key)
+	}
+	sort.Strings(keys)
```
f0rmiga, line 111: I'd use https://github.com/emirpasic/gods/tree/dbdbadc158ae6b453820b3cfb8c6cb48be4d7ddf#treeset as it may give some performance improvement while adding the elements to the set. We already have this library as a dependency, so it's straightforward to use it.

In your approach, you make a whole copy of the slice and then sort it, which is not very optimal.

f0rmiga (changes requested):

f0rmiga:
@abhishekshivanna thanks for the contribution!

gazelle/manifest/manifest.go
abhishekshivanna: ah sweet, didn't know the dep was already pulled in! (was trying to avoid new deps) :)
thanks for the review!

f0rmiga (approved):

f0rmiga:
@abhishekshivanna thanks for addressing my comments.

PR #814, opened 2022-08-29
feat: add alias generation for pip parse

corypaik:
Generate alias packages and targets for `pip_parse` under the parent repository. This serves two purposes. It allows for users to reference packages using a simpler `@pypi//package_name` without the import issues associated with putting the contents of the package there (as may be the case with `pip_install` following this naming pattern). Additionally, this allows users to override individual libraries either by default or based on select statements.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

Currently, in order to refer to a dependency by name with `pip_parse` one must use the syntax `@pypi_package_name//:pkg`.  It it also not currently possible to override the dependencies of pip repositories using local packages or those built with Bazel. The only workaround available now is to define a repository (e.g., `@pypi_package_name` prior to calling `pip_parse`. This is not ideal, and does not allow for selecting a version to use (e.g., local vs. from pip) via flags.

## What is the new behavior?

The alias packages add the option to refer to dependencies under one central repository using the syntax `@pypi//package_name`.  Additionally, it allows users to override the library targets by default or using any valid select statement. 

Note that the original names (e.g., `@pypi_package_name//:pkg`) remain unmodified, as do all functions in `requirements.bzl`. The only exception is to `all_requirements`, which now uses the aliases instead. Happy to adjust this behavior as desired, my primary usage is `@pypi//package_name` and a few interactive targets that use `all_requirements`.

- [ ] Yes
- [x] No

## Other information

I originally was working on a version which made the creation of alias packages optional (and disabled by default). After trying to profile both with and without the alias repositories I could not find any measurable difference so I got rid of the option entirely.

I've refactored the existing tests to work with these changes and added some notes to the `pip_parse` example, but I'd also be happy to add additional tests if need be.


UebelAndre:
@corypaik would you be willing to rebase this PR? I think it's a very useful change!

rickeylev:
Sorry, this fell off my radar.

Can you explain more about the motivating use case a bit more? Essentially, you want to allow an arbitrary select condition to control what target is used? I'm not necessarily against doing so, but it strikes me as a bit strange -- ignoring Bazel for a moment, it'd be like pip installing something, then mucking with the install or sys.path to override particular libraries, which makes me raise my eye brows.

Does pip have a builtin mechanism to ignore a particular dependency? Or maybe we should modify the requirements.txt processing to remove the desired library and replace it with the desired alias?


[There's a Bazel flag to override specific repositories](https://bazel.build/docs/external#overriding-repositories) -- does that satisfy the same need?

Each pypi package being its own repo is most likely an intentional design decision (an unreferenced repo doesn't get loaded or fetched at all; I think it better isolates the repo-rule impact, too?). Putting everything into a single repo likely defeats that to some extent.

I think the aliases need to be moved into the respective package's repo target -- this ensures that only a single actual target for a library is used. Otherwise, what you can do is e.g. `deps = ["@pypy//foo", "@pypi_foo//foo"]`, and end up with two different versions of the same library and behavior is undefined as to which you get. This behavior happens _transitively_ -- i.e. if the transitive closure of a target contains both those targets, you'll end up with undefined behavior.

> as may be the case with pip_install

I think we deprecated pip_install in favor of pip_parse? I don't think we should add features to something we've deprecated.

Issue #796, opened 2022-08-17
Linking `rules_python` libpython with PyO3


omerbenamram:
# 🐞 bug report

### Is this a regression?

Not AFAIK. 

### Description

When an interpreter is downloaded, `rules_python` names the affiliated `libpython` as 
`libpython<version>.so.1.0` (https://github.com/bazelbuild/rules_python/blob/main/python/repositories.bzl#L206)

When used as a dependency:

```
    rust_shared_library(
        name = wheel_name,
        srcs = rust_srcs,
        crate_name = crate_name,
        edition = "2021",
        deps = rust_deps + [
            "@python//:libpython",
        ],
        **rust_kwargs
    )
```

the following linker flags will be emitted:
`-lpython3.9`
`-Lnative=bazel-out/k8-fastbuild/bin/_solib_k8/_U@python_Ux86_U64-unknown-linux-gnu_S_S_Clibpython___Ulib`

However, the directory `-Lnative=bazel-out/k8-fastbuild/bin/_solib_k8/_U@python_Ux86_U64-unknown-linux-gnu_S_S_Clibpython___Ulib` only contains said `libpython3.9.so.1.0`, which isn't picked up by the linker as a valid match for `-lpython3.9`

```
-as-needed" "-Wl,-z,relro,-z,now" "-B/usr/lib/llvm-14/bin" "-ldl" "-lpthread" "-ldl" "-lpthread" "-ldl" "-lpthread" "-ldl" "-lpthread" "-ldl" "-lpthread" "-ldl" "-lpthread" "-ldl" "-lpthread" "-ldl" "-lpthread" "-ldl" "-lpthread" "-ldl" "-lpthread" "-lstdc++" "-lm"
  = note: ld.lld: error: unable to find library -lpython3.9
          clang: error: linker command failed with exit code 1 (use -v to see invocation)
          
```

Running against a modified `rules_python` which uses `lib/libpython{python_version}.so` instead of `lib/libpython{python_version}.so.1.0` solves this.
```
cc_import(
    name = "libpython",
    hdrs = [":includes"],
    shared_library = select({{
        "@platforms//os:windows": "python3.dll",
        "@platforms//os:macos": "lib/libpython{python_version}.dylib",
        "@platforms//os:linux": "lib/libpython{python_version}.so",
    }}),
)
```

## 🔬 Minimal Reproduction

https://github.com/omerbenamram/rules_python_pyo3_repro/

## 🌍  Your Environment

[...]

PR #820, opened 2022-09-02
fix: replace cc_import with cc_library for libpython

scasagrande:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #796 

## What is the new behavior?

This allows rust library targets that utilize `pyo3` to correctly be able to include `libpythonX.Y.so` via depending on `@python//:libpython`.

- [ ] Yes
- [x] No


rickeylev:
@f0rmiga would you be willing/able to have a look? From the associated issue, it sounds like you would be better qualified to judge.

At a glance, I would have expected cc_import to Just Work, since it's doc indicates it's for prebuilt shared libraries, which would imply either (a) something downstream of it is incorrect, or (b) something is missing in our cc_import() call. But I don't know enough about the difference between cc_import and cc_library (and the lower-level linker stuff mentioned in the issue) to say with any certainty.

scasagrande:
I *just* came up with another solution for myself that completely avoids having `libpython` as a dependency for my `pyo3` target. I should have thought of this ahead of time, but that's important for manylinux compliance.

I enabled the `"extension-module"` and `"abi3-py38"` features for `pyo3` and now I don't need `libpython` as a dependency

rickeylev:
> I just came up with another solution for myself that completely avoids having libpython as a dependency for my pyo3 target. I should have thought of this ahead of time, as this is important for manylinux compliance.

Great, does that mean changing it to cc_library is unnecessary now? If so, if you want to pivot this PR to adding a comment, e.g. `# NOTE: if you're depending this from rust, then ...`, we can finish review. It sounds like a good hint to give future people trying to do rust+python interop

f0rmiga:
I still think the approach in the PR is correct. In practical terms, the difference between cc_library and cc_import is that cc_import only takes 1 file, so in the case of Linux that has 2 files (the real shared object and the symlink to it), there's no way to use cc_import. For one case I worked on, the cc_import was fine because we were using RBE, and since RBE converts the symlinks into the realpath contents, it worked fine.

f0rmiga (approved):

PR #832, opened 2022-09-20
Implement py_proto_library

comius:
`py_proto_library` was tested manually on Bazel repository using Bazel 6.0.0 with and without using bzlmod.

Extend README.md to mention also `py_proto_library`.

With bzlmod `py_proto_library` just works. Users without bzlmod, need to import `rules_proto` and `protobuf` repositories into their `WORKSPACE` file. 

`py_proto_library` is compatible with Bazel `>=5.4.0` and `>=6.0.0`.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -7,8 +7,8 @@ module(
 bazel_dep(name = "platforms", version = "0.0.4")
 
 # Those are loaded only when using py_proto_library
-bazel_dep(name = "rules_proto", version = "4.0.0")
-bazel_dep(name = "protobuf", repo_name = "com_google_protobuf", version = "3.19.2")
+bazel_dep(name = "rules_proto", version = "5.3.0-21.7")
+bazel_dep(name = "protobuf", repo_name = "com_google_protobuf", version = "21.7")
 
 internal_deps = use_extension("@rules_python//python:extensions.bzl", "internal_deps")
 
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -6,6 +6,10 @@ module(
 
 bazel_dep(name = "platforms", version = "0.0.4")
 
+# Those are loaded only when using py_proto_library
+bazel_dep(name = "rules_proto", version = "4.0.0")
+bazel_dep(name = "protobuf", repo_name = "com_google_protobuf", version = "3.19.2")
+
 internal_deps = use_extension("@rules_python//python:extensions.bzl", "internal_deps")
 
 internal_deps.install()
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 ## Overview
 
 This repository is the home of the core Python rules -- `py_library`,
-`py_binary`, `py_test`, and related symbols that provide the basis for Python
+`py_binary`, `py_test`, `py_proto_library`, and related symbols that provide the basis for Python
 support in Bazel. It also contains package installation rules for integrating with PyPI and other package indices. Documentation lives in the
 [`docs/`](https://github.com/bazelbuild/rules_python/tree/main/docs)
 directory and in the
--- a/internal_deps.bzl
+++ b/internal_deps.bzl
@@ -124,3 +124,24 @@ def rules_python_internal_deps():
         strip_prefix = "bazel-integration-testing-165440b2dbda885f8d1ccb8d0f417e6cf8c54f17",
         sha256 = "2401b1369ef44cc42f91dc94443ef491208dbd06da1e1e10b702d8c189f098e3",
     )
+
+    maybe(
+        http_archive,
+        name = "rules_proto",
+        sha256 = "66bfdf8782796239d3875d37e7de19b1d94301e8972b3cbd2446b332429b4df1",
+        strip_prefix = "rules_proto-4.0.0",
+        urls = [
+            "https://github.com/bazelbuild/rules_proto/archive/refs/tags/4.0.0.tar.gz"
+        ],
+    )
+
+    maybe(
+        http_archive,
+        name = "com_google_protobuf",
+        sha256 = "4dd35e788944b7686aac898f77df4e9a54da0ca694b8801bd6b2a9ffc1b3085e",
+        strip_prefix = "protobuf-3.19.2",
+        urls = [
+            "https://github.com/protocolbuffers/protobuf/archive/v3.19.2.tar.gz",
+            "https://mirror.bazel.build/github.com/protocolbuffers/protobuf/archive/v3.19.2.tar.gz",
+        ],
+    )
--- a/internal_deps.bzl
+++ b/internal_deps.bzl
@@ -128,20 +128,20 @@ def rules_python_internal_deps():
     maybe(
         http_archive,
         name = "rules_proto",
-        sha256 = "66bfdf8782796239d3875d37e7de19b1d94301e8972b3cbd2446b332429b4df1",
-        strip_prefix = "rules_proto-4.0.0",
+        sha256 = "dc3fb206a2cb3441b485eb1e423165b231235a1ea9b031b4433cf7bc1fa460dd",
+        strip_prefix = "rules_proto-5.3.0-21.7",
         urls = [
-            "https://github.com/bazelbuild/rules_proto/archive/refs/tags/4.0.0.tar.gz"
+            "https://github.com/bazelbuild/rules_proto/archive/refs/tags/5.3.0-21.7.tar.gz",
         ],
     )
 
     maybe(
         http_archive,
         name = "com_google_protobuf",
-        sha256 = "4dd35e788944b7686aac898f77df4e9a54da0ca694b8801bd6b2a9ffc1b3085e",
-        strip_prefix = "protobuf-3.19.2",
+        sha256 = "75be42bd736f4df6d702a0e4e4d30de9ee40eac024c4b845d17ae4cc831fe4ae",
+        strip_prefix = "protobuf-21.7",
         urls = [
-            "https://github.com/protocolbuffers/protobuf/archive/v3.19.2.tar.gz",
-            "https://mirror.bazel.build/github.com/protocolbuffers/protobuf/archive/v3.19.2.tar.gz",
+            "https://mirror.bazel.build/github.com/protocolbuffers/protobuf/archive/v21.7.tar.gz",
+            "https://github.com/protocolbuffers/protobuf/archive/v21.7.tar.gz",
         ],
     )
--- a/internal_deps.bzl
+++ b/internal_deps.bzl
@@ -20,10 +20,10 @@ def rules_python_internal_deps():
         http_archive,
         name = "rules_pkg",
         urls = [
-            "https://mirror.bazel.build/github.com/bazelbuild/rules_pkg/releases/download/0.2.4/rules_pkg-0.2.4.tar.gz",
-            "https://github.com/bazelbuild/rules_pkg/releases/download/0.2.4/rules_pkg-0.2.4.tar.gz",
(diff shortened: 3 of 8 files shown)


comius:
cc @rickeylev @haberman

proto/MODULE.bazel
```diff
@@ -0,0 +1,8 @@
+module(
+    name = "rules_python_proto",
+    compatibility_level = 1,
+    version = "0.0.0",
+)
+
+bazel_dep(name = "rules_proto", version = "5.3.0-21.5")
```
meteorcloudy, line 7: There is no version 5.3.0-21.5 for rules_proto in the BCR.
https://github.com/bazelbuild/bazel-central-registry/tree/main/modules/rules_proto

proto/MODULE.bazel
```diff
@@ -0,0 +1,8 @@
...
+bazel_dep(name = "rules_python", version = "0.4.0")
```
meteorcloudy, line 8: I thought you want to have a separate module to avoid having extra dependencies of rules_python, but here you depend on rules_python already, then having a sub module doesn't seem necessary anymore?

proto/MODULE.bazel
comius: I know, we need to fix this.

proto/MODULE.bazel
comius: The other way around. I don't want `rules_python` to depend on `rules_proto` or on `com_google_protobuf`.

meteorcloudy: I see, that makes sense.

proto/MODULE.bazel
```diff
@@ -0,0 +1,9 @@
+module(
+    name = "rules_python_proto",
+    compatibility_level = 1,
+    version = "0.0.0",
```
meteorcloudy, line 4: nit, "0.0.1"?

meteorcloudy (approved):
LGTM, from dependencies management perspective.

proto/python/py_proto_library.bzl
```diff
@@ -0,0 +1,157 @@
...
+        ),
+        PyInfo(
+            transitive_sources = default_outputs,
+            # Proto always produces 2- and 3- compatible source files
+            has_py2_only_sources = False,
+            has_py3_only_sources = False,
+        ),
+    ]
+
+py_proto_library = rule(
```
UebelAndre, line 151: Would you be willing to add docs to the rule and attribute?

proto/python/py_proto_library.bzl
```diff
@@ -0,0 +1,157 @@
...
+    )
+
+    return [
+        _PyProtoInfo(
+            runfiles_from_proto_deps = runfiles_from_proto_deps,
+            transitive_sources = transitive_sources,
+        ),
+    ]
+
+py_proto_aspect = aspect(
```
UebelAndre, line 103: Same here, would you be willing to add docs to the rule and attribute? Why does the aspect need an attribute for the toolchain?

UebelAndre (changes requested):

proto/python/py_proto_library.bzl
```diff
@@ -0,0 +1,157 @@
...
+        ),
+    ]
+
+py_proto_library = rule(
+    implementation = py_proto_library_rule,
+    attrs = {
+        "deps": attr.label_list(providers = [ProtoInfo], aspects = [py_proto_aspect]),
+    },
+    provides = [PyInfo],
+)
```
rickeylev, line 157:
```suggestion
py_proto_library = rule(
    implementation = py_proto_library_rule,
    doc = ("Creates a Python importable library for Protocol Buffer messages. " +
    "By convention, the name is the base proto-library name with `_pb2`, e.g. " +
    "`foo_pb2` from wrapping `foo_proto`"
    ),
    attrs = {
        "deps": attr.label_list(
        doc = ("ProtoInfo-providing targets to generate Python libraries for. "+
        "Usually this is just the one target: the proto library of interest.",
        providers = [ProtoInfo], aspects = [py_proto_aspect]),
    },
    provides = [PyInfo],
)
```

proto/python/py_proto_library.bzl
```diff
@@ -0,0 +1,157 @@
...
+        "_aspect_proto_toolchain": attr.label(
+            default = ":python_toolchain",
+        ),
+    },
+    attr_aspects = ["deps"],
+    required_providers = [ProtoInfo],
+    provides = [_PyProtoInfo],
+)
+
+def py_proto_library_rule(ctx):
```
rickeylev, line 115: We should probably make everything in here private (except for the public py_proto_library symbol), or move everything into a directory named "private" or "internal" (I think these are established conventions to help prevent people loading things they shouldn't).

Right now, it's a bit confusing which of py_proto_library.bzl or defs.bzl is the one you should load -- I remember having this confusion when first looking at repos. Repos that use the "private" subdir make it much more clear what I should be loading

UebelAndre: Yeah, I thought it was more idiomatic to do something like the following:
```suggestion
def _py_proto_library_impl(ctx):
```

I've never seen a public rule implementation. But I have seen public APIs for creating actions (e.g. [cc_common](https://bazel.build/rules/lib/cc_common))

proto/python/py_proto_library.bzl
comius: Done.

proto/python/py_proto_library.bzl
comius: Well to run it. I didn't really see other rules where implicit dependecies/toolchain would be extensively documented.

aaliddell:
FWIW transitive proto compilation with an aspect is probably not the right method to use an longer. Both rules_proto and rules_proto_grpc have dropped aspect based compilation in favour of more explicit direct compilation of the named protos only within a target. This is because transitive compilation is both artificially limiting in terms of passable options and more importantly leads to multiple definitions of the same proto message if depended on via differing trees (a serious problem in C++, a nuisance in Python). See here for further details: https://rules-proto-grpc.com/en/latest/transitivity.html

proto/python/py_proto_library.bzl
comius: Done.

proto/MODULE.bazel
comius: Done.

comius:
> FWIW transitive proto compilation with an aspect is probably not the right method to use any longer. Both rules_proto and rules_proto_grpc have dropped aspect based compilation in favour of more explicit direct compilation of the named protos only within a target. This is because transitive compilation is both artificially limiting in terms of passable options and more importantly leads to multiple definitions of the same proto message if depended on via differing trees (a serious problem in C++, a nuisance in Python). See here for further details: https://rules-proto-grpc.com/en/latest/transitivity.html

Compiling only directly is a good choice for gRPC generated libraries. 

For non-service proto libraries, it's better to use `lang_proto_library` with an aspect. It removes redundancies - having to add `py_proto_library` next to every `py_proto_library` in the transitive closure. Aspects take care that you 1:1 mapping from proto dependency graph to python graph - even if `lang_proto_library` is duplicated. If there are problems with proto dependency graph, they are not going to be solved by either direct or aspect based solution. Aspect based solution will probably fail, whereas no-aspect/direct solution will potentially hide problems.

proto/MODULE.bazel
comius: Done.

proto/python/private/py_proto_library.bzl
```diff
@@ -0,0 +1,191 @@
...
+      Use `py_proto_library` to generate Python libraries from `.proto` files.
+
+      The convention is to name the `py_proto_library` rule `foo_py_pb2`,
+      when it is wrapping `proto_library` rule `foo_proto`.
+
+      `deps` must point to a `proto_library` rule.
+
+      Example:
+
+```
```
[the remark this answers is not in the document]
comius: I don't think indentation matters in md, which is used by Stardoc (except for the codeblock).

rickeylev:
https://rules-proto-grpc.com/en/latest/transitivity.html says

> In hindsight, [recursive aspect] was not the correct behaviour and led to many bugs, since you may end up creating a library that contains compiled proto files from a third party, where you should instead be depending on a proper library for that third party’s protos.

Is there a place to further discuss this (this PR is probably not the best venue)? This situation sounds similar to some of the problems we have with SWIG (and C bindings in general) within Google, and pushing the codegen burden elsewhere has its own set of problems. @aaliddell

aaliddell:
> Is there a place to further discuss this (this PR is probably not the best venue)? This situation sounds similar to some of the problems we have with SWIG (and C bindings in general) within Google, and pushing the codegen burden elsewhere has its own set of problems. @aaliddell 

You can email me or put it in a new discussion [here](https://github.com/rules-proto-grpc/rules_proto_grpc/discussions); not sure I’ll be much help though with SWIG. Or put it on the bazel-discuss mailing list if you want a wider audience. Or slack. Too many options 😵‍💫

> Compiling only directly is a good choice for gRPC generated libraries.

> For non-service proto libraries, it's better to use lang_proto_library with an aspect. It removes redundancies - having to add py_proto_library next to every proto_library in the transitive closure. Aspects take care that your 1:1 mapping from proto dependency graph to python graph - even if lang_proto_library is duplicated. If there are problems with proto dependency graph, they are not going to be solved by either direct or aspect based solution. Aspect based solution will probably fail, whereas no-aspect/direct solution will potentially hide problems.

Ok, I thought I’d check your weren’t just blindly copying some of the other language rulesets’ behaviour without knowing the implications; seems like you’re on top of it 👍

tpudlik:
I just ran into an interesting bug in the unofficial `py_proto_library` from https://github.com/protocolbuffers/protobuf/: it doesn't respect `tags = ["manual"]`. No tags are forwarded to [the code generation rule](https://github.com/protocolbuffers/protobuf/blob/2bed3439bb817fabf165dc24d756c907a887b737/protobuf.bzl#L536-L547), so the code generation runs unconditionally. So I thought I'd confirm: this "official" implementation doesn't suffer from this bug, does it?

jiawen:
I just discovered [this comment](https://github.com/protocolbuffers/protobuf/blob/main/protobuf.bzl#L570) in com_google_protobuf//:protobuf.bzl.py_proto_library(). It suggests that Bazel 5.3 will include py_proto_library() internally.

Is this the implementation? But only available when using bzlmod and/or Bazel 6?

hrfuller:
Still wrapping my head around bzlmod stuff. Is the idea that users would have to depend on the py_proto_module seperately from rules_python in their MODULE.bazel file in order to use the py_proto_library rule; or would any dependency on rules_python include the ability to use py_proto_library?

comius:
> I just discovered [this comment](https://github.com/protocolbuffers/protobuf/blob/main/protobuf.bzl#L570) in com_google_protobuf//:protobuf.bzl.py_proto_library(). It suggests that Bazel 5.3 will include py_proto_library() internally.
> 
> Is this the implementation? But only available when using bzlmod and/or Bazel 6?

Yes, this is the implementation. It will work with Bazel 5.3 and Bazel 6. And also without bzlmod.

> Still wrapping my head around bzlmod stuff. Is the idea that users would have to depend on the py_proto_module seperately from rules_python in their MODULE.bazel file in order to use the py_proto_library rule; or would any dependency on rules_python include the ability to use py_proto_library?

I fixed the PR, so that the users only need to depend on rules_python. And proto repositories are fetched only if you use py_proto_library,

comius:
> per chat with Ivo: PR is not yet ready to merge; he's going to refactor things into a single bzlmod module file because deps for modules are fetched lazily (and thus no proto dep will be brought in unless necessary)

The PR is now ready.

rickeylev (approved):

alexeagle:
This forgot to update the docs/ folder so that the new stardoc comments are published for users to find.

PR #846, opened 2022-10-04
feat: multi-toolchain support

f0rmiga:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Most mono repos using Bazel have the problem of needing support for multiple Python versions. Currently, there's no simple way to accomplish this.

## What is the new behavior?

Add support for registering multiple Python toolchains with different versions as an opt-in feature so that users can gradually migrate portions of their mono repos to this new feature.

Using load statements to override `py_binary` and `py_test`, this feature implementation also plays well with the Gazelle extension by using Gazelle's `# gazelle:map_kind from_kind to_kind to_kind_load` feature. Also, pure Python libraries can take advantage of this by testing against multiple Python versions in the same BUILD file (see examples).

This feature is achieved by using platform transitions behind the scenes.

- [ ] Yes
- [x] No

## Other information


python/platforms/transition.bzl
```diff
@@ -0,0 +1,178 @@
...
+    test = True,
+)
+
+def _py_rule(rule, transition_rule, name, target_platform, **kwargs):
+    args = kwargs.pop("args", None)
+    data = kwargs.pop("data", None)
+    env = kwargs.pop("env", None)
+
+    # Attributes common to all build rules.
+    # https://bazel.build/reference/be/common-definitions#common-attributes
```
rickeylev, line 111: Lets move the list of these attribute names into a separate file. It will most assuredly get reused by other code.

python/platforms/transition.bzl
```diff
@@ -0,0 +1,178 @@
...
+    features = kwargs.pop("features", None)
+    restricted_to = kwargs.pop("restricted_to", None)
+    tags = kwargs.pop("tags", None)
+    target_compatible_with = kwargs.pop("target_compatible_with", None)
+    testonly = kwargs.pop("testonly", None)
+    toolchains = kwargs.pop("toolchains", None)
+    visibility = kwargs.pop("visibility", None)
+
+    rule(
+        name = "_" + name,
```
rickeylev, line 126: The "name" arg is being removed in an upcoming release; it ended up causing some sort of cache poisoning type of issue or something.

You'll need to change this to bind the return value to an global variable.

python/platforms/transition.bzl
```diff
@@ -0,0 +1,178 @@
...
+        tools = data,
+        args = args,
+        env = env,
+        is_windows = select({
+            "@platforms//os:windows": True,
+            "//conditions:default": False,
+        }),
+        target = ":_" + name,
+
+        # Attributes common to all build rules.
```
rickeylev, line 158: It seems like there's a lot of copy/paste for these common attributes. How about instead of manually carrying around ~11 variables, we stick them in a dict? 11 is a lot to carry around and makes it easy to forget one.

python/repositories.bzl
```diff
@@ -349,6 +395,7 @@ def python_register_toolchains(
         distutils: see the distutils attribute in the python_repository repository rule.
         distutils_content: see the distutils_content attribute in the python_repository repository rule.
         register_toolchains: Whether or not to register the downloaded toolchains.
+        set_python_version_constraint: When set to true, target_compatible_with for the toolchains will include a version constraint.
```
rickeylev, line 398: Why does this need to be an opt-in setting instead of on by default?

python/repositories.bzl
```diff
@@ -381,17 +430,67 @@ def python_register_toolchains(
...
+        fail("python_versions must not be empty")
+
+    if not default_version:
+        default_version = python_versions.pop(0)
+    for python_version in python_versions:
+        if python_version == default_version:
+            # We register the default version lastly so that it's not picked first when --platforms
+            # is set with a constraint during toolchain resolution. This is due to the fact that
+            # Bazel will match the unconstrained toolchain if we register it before the constrained
+            # ones.
```
rickeylev, line 475: Can you explain this a bit more? I don't fully understand the problem.

Each of the registered toolchains should have e.g. compatible_with=[X] such that they're mutually exclusive. When a rule does toolchain lookup, the rule should already have a python version set so that when Bazel matches its config to the available toolchains, it can only match one.

It sounds like the either the "mutually exclusive" part or the "rule's setting is set" part isn't true, and multiple toolchains are matching?

python/platforms/transition.bzl
```diff
@@ -0,0 +1,178 @@
...
+    target_compatible_with = kwargs.pop("target_compatible_with", None)
+    testonly = kwargs.pop("testonly", None)
+    toolchains = kwargs.pop("toolchains", None)
+    visibility = kwargs.pop("visibility", None)
+
+    rule(
+        name = "_" + name,
+        args = args,
+        data = data,
+        env = env,
```
rickeylev, line 129: Also pass along env_inherit.

python/platforms/transition.bzl
```diff
@@ -0,0 +1,178 @@
...
+    exec_properties = kwargs.pop("exec_properties", None)
+    features = kwargs.pop("features", None)
+    restricted_to = kwargs.pop("restricted_to", None)
+    tags = kwargs.pop("tags", None)
+    target_compatible_with = kwargs.pop("target_compatible_with", None)
+    testonly = kwargs.pop("testonly", None)
+    toolchains = kwargs.pop("toolchains", None)
+    visibility = kwargs.pop("visibility", None)
+
+    rule(
```
rickeylev, line 125: I think you also want to pass along "srcs" and "deps" here. They can also participate in expansion for args and env.

rickeylev (commented):
I'm not sure about the usage of platforms here. It seems like too coarse an object. All that's is being varied is the python version, but a platform encompasses a lot more than that.

I guess, stepping back, what's the usage of this look like and use case?

When you mentioned it in chat, I thought you meant something like:

```
py_binary(name="a", python_version="3.8", data=[":b"], deps=[":c"])
py_binary(name="b", python_version="3.9", deps=[":c"])
whatever(name="c", srcs = select({
  "py3.8": ...,
  "py3.9": ...
}))
```

And the end result is two runtimes end up in runfiles, and each respective binary uses its associated interpreter, and dependencies are able to just naturally select on the python version.

But, I see platforms here, and there's many edits to the pip stuff, so I'm not sure if what I expected is entirely accurate.

f0rmiga:
> I'm not sure about the usage of platforms here. It seems like too coarse an object. All that's is being varied is the python version, but a platform encompasses a lot more than that.

We definitely want to use platforms here. While we can't yet do cross-platform builds with native wheels, we can support this soon (a common case being macos -> linux containers).

> I guess, stepping back, what's the usage of this look like and use case?

Commonly, a mono repo has multiple teams and projects, leading to different Python version requirements. This allows projects to move to newer Python versions independently.

> When you mentioned it in chat, I thought you meant something like:
> 
> ```
> py_binary(name="a", python_version="3.8", data=[":b"], deps=[":c"])
> py_binary(name="b", python_version="3.9", deps=[":c"])
> whatever(name="c", srcs = select({
>   "py3.8": ...,
>   "py3.9": ...
> }))
> ```

Why would someone want to deal with selects manually? Transitions are here to solve exactly this problem.

> And the end result is two runtimes end up in runfiles, and each respective binary uses its associated interpreter, and dependencies are able to just naturally select on the python version.

That's exactly the case with my implementation. See the tests.

> But, I see platforms here, and there's many edits to the pip stuff, so I'm not sure if what I expected is entirely accurate.

Your expectation and the community's expectations may not be aligned on this.

examples/multi_python_versions/.bazelrc
```diff
@@ -0,0 +1,5 @@
+test --test_output=errors
+
+# Windows requires these for multi-python support:
```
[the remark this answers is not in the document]
f0rmiga: This is a good question. I don't know many people developing with Windows. @groodt What do you think?

python/platforms/transition.bzl
f0rmiga: Could you point me to some docs or PRs talking about this, please?

python/repositories.bzl
f0rmiga: This allows a default toolchain to be picked when multi-python support is not enabled.

examples/multi_python_versions/.bazelrc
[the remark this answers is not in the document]
alexeagle: Not that it's necessarily correct, but rules_js has a hard requirement of enabling runfiles on Windows and I don't know of any complaints we've gotten from Windows developers who aren't able to use it.

rickeylev:
So, keep in mind I'm coming into this without much context. A lot of this code is new to me, and I only know the broad strokes of how the repository rules like requirements, pip parse, and the standalone interpreter rules work. I'm happy to VC to chat about some the points in this PR.

> Commonly, a mono repo has multiple teams and projects, leading to different Python version requirements. This allows projects to move to newer Python versions independently.

This basically sounds like the python 2/3 situation (i.e. the `py_binary.python_version` attribute and --python_version flag), except applied to 3.X value. Is that correct?

That, too uses a transition, however, it has two key differences:
 * That transition only affects the python version, not the entire platform.
 * The python version is a parameter of the rule, rather than a rule itself.

The first means any surrounding platform configuration is inherited. This essentially allows an easy way to have a single target definition that can be built for multiple platforms.

The second means only a single rule is necessary, rather than N rules. It also means a default can be obtained from elsewhere rather than it _always_ being forced.

> 
> > When you mentioned it in chat, I thought you meant something like:
> > ```
> > py_binary(name="a", python_version="3.8", data=[":b"], deps=[":c"])
> > py_binary(name="b", python_version="3.9", deps=[":c"])
> > whatever(name="c", srcs = select({
> >   "py3.8": ...,
> >   "py3.9": ...
> > }))
> > ```
> 
> Why would someone want to deal with selects manually? Transitions are here to solve exactly this problem.

Selects and transitions solve different but related problems. Transitions are for forcing a transitive closure to have a particular configuration. Selects are for selecting different dependencies/inputs based upon the currently active configuration.

For the problem of: "I need to use a different target depending on the python version, OS, cpu, etc", then the answer is to use select. Whether it's a wheel, py_library, filegroup, or what not, doesn't particularly matter. Usually the consumed target hides this detail from the consumed target by using alias() and/or varying the inputs to its canonical target.

As an example, lets say we wanted to build an binary using importlib resources for multiple Python versions. I use this example because importlib_resources (underscore) is a backport of functionality that is in the stdlib later. At the same time, the contents of the importlib_resources wheel could contain Python version specific artifacts (e.g. pyc files). To make this work, you would end up with something like this:

```
# myapp/BUILD
py_binary(
  name="myapp",
  deps = select({
    "py3.12": [] # 3.12 finally has all the features we need; no dep needed
    "default": [requirements("importlib_resources")]
  })
)

# importlib_resources/BUILD
# I forget the exact generated BUILD file contents, but I hope this
# gets the gist across.
# Method 1: a single entry point whose inputs are varied. Makes sense when
# the variation is just on some of the attributes.
py_wheel(
  name = "importlib_resources",
  srcs = select({
    "py3.7": ["importlib_resources-py37.wheel"],
    "py3.8": ["importlib_resources-py38.wheel"],
    "py3.9": ...
    "py3.10": ...
    "py3.11": ...
    "py3.12": ...
  })
)
# Method 2: an alias that dispatches to the appropriate target. Makes sense
# when the variation is more nuanced.
alias(
  name = "importlib_resources",
  actual = select({
    "py3.8": ":_importlib_resources_38",
    "py3.9": ":_importlib_resources_39",
    ...
}))
py_library(name="_importlib_resources_{XX}", ...) # etc
```

The above is mostly orthogonal to a transition. The transition just changes where the active configuration is being set. Note I'm just a simple first-order dependency relationship for clarity. The path from "myapp" to "importlib_resources" could have an arbitrary number of intervening dependencies, especially for a monorepo that code shares freely.

Another chunk I don't understand `compile_pip_requirements_X_Y` -- why must there be a separate rule for each version? I think it can be replaced with e.g.
```
compile_pip_requirements(
  name = "requirements",
  requirements_txt = select({"py38": "requirements_38.txt", "py39": "requirements_39.txt", etc})
)
```

And if the version does needs to be forced, an attribute can specify that.

Similarly, `multi_pip_parse` takes version strings as inputs to map the versions to interpreters -- this looks suspiciously like the sort of thing that toolchain resolution is supposed to handle? e.g. a toolchain is registered to be compatible for python 3.8, and then pip_parse obtains the interpreter from the toolchain, and finding the matching interpreter for the current configuration is handled by toolchain resolution.

For the problem of: "I need this target to build with a specific configuration", then the answer is to use a transition.

To be clear, yes, a transition definitely needs to be involved here as, at a minimum, the python version must be changed to affect the transitive closure. My issue isn't with using a transition, but that the platform needs to be transitioned instead of simply the python version.

As an example, if I'm understanding this code correctly, this looks to be possible, but probably not intended:

```
py_binary(
  name="outer", target_platform="@rules_python//:darwin_py38",
  data = [":inner"],
)
py_binary(
  name = "inner", target_platform="@rules_python//:linux_py38"
)
```

The net effect is "outer" is built for darwin and includes the "inner" binary built for linux. This is probably not intended. Making this work requires inner to add a separate target with a platform-specific name and associated alias mapping for each platform a consumer needs, and also results in the consumer needing to handle the runfiles artifact name varying depending on the platform.

python/platforms/transition.bzl
rickeylev: https://bazel.build/rules/lib/globals#rule (last arg in the args table).
https://github.com/bazelbuild/bazel/issues/16301

python/repositories.bzl
rickeylev: Hm. Ok. Why is there a toolchain without a version constraint? Do we not know the python version when the toolchain is registered?

examples/multi_python_versions/.bazelrc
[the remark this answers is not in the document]
rickeylev: That sounds very promising. I also found these:
https://github.com/bazelbuild/bazel/issues/5807
https://github.com/bazelbuild/bazel/issues/8704

In the reference history of the first one, it looks like several other rules also enabled symlinks for windows. The second appears to be a canonical "enable by default" bug. 

I also found this doc: https://bazel.build/configure/windows#symlink
Which seems to both recommend using and not using it :shrug:

rickeylev:
Per chat:

re: specifying the python version using the rule name vs attribute: I'm fine with having e.g. `py_binary_3_8` etc; I don't have a strong opinion there. There's pros/cons to encoding this in the rule name vs attribute. If people see value in version-specific macros, then yeah, lets define those (Thulio mentioned something about how Gazelle can be told to rewrite the rule used, which definitely sounds like an easy and appealing way to bulk-update build files)

re: transitioning platforms: I asked some other Starlark experts, and they agreed transitioning platform is a pretty heavy weight way to do it. A better way is to create a flag that a transition affects and use config_setting to map the flag to constraints for the toolchains.

Putting it all together, I think it looks something like this:

```
# https://bazel.build/extending/config#predefined-settings
load("@bazel_skylib//rules:common_settings.bzl", "string_flag")
string_flag(name="python_version", default="XXX", values=["3.6", "3.7", ...)

config_setting(
  name = "is_py3.6",
  flag_values = {":python_version": "3.6"}
)
config_setting(
  name = "is_py3.7"
  flag_values = {":python_version": "3.7"}
)

py_runtime(
  name = "py3.6_runtime",
  target_compatible_with = [":is_py3.6"],
)
py_runtime(
  name = "py3.7_runtime",
  target_compatible_with = [":is_py3.7"],
)
toolchain(
  name = "py3.6_toolchain",
  target_compatible_with = [":is_py3.6", ...],
  toolchain = ":py3.6_runtime",
  toolchain_type = ...
)
toolchain(
  name = "py3.7_toolchain",
  target_compatible_with = [":is_py3.7", ...],
  toolchain = ":py3.7_runtime",
  toolchain_type = ...
)

def py_version_transition_impl(...):
  return {":python_version": ctx.attr.python_version}

def py_versioned_binary_impl(ctx):
  ...same as before...
py_versioned_binary = rule(cfg=py_version_transition, ...)
```

f0rmiga:
@rickeylev

re: specifying the python version using the rule name vs attribute: you don't necessarily need to rename, or follow what I did; see below:

https://github.com/bazelbuild/rules_python/blob/1053ab06b16c7e346e8b6a253fd3c45bcc4e686d/examples/multi_python_versions/tests/BUILD.bazel#L1-L4

I can think of a few cases on a monorepo for doing the renaming. The most obvious would be a monorepo where a library is shared between multiple projects, and those projects end up having different python version requirements. The maintainers of the shared library will want to assert the tests pass on all python versions, so a rename like this would be required. Otherwise, it makes more sense to just change the import statement on a BUILD file and it's like magic to move to another version. E.g. when adopting this new feature:

```diff
-load("@rules_python//python:defs.bzl", "py_binary", "py_test")
+load("@python//3.8:defs.bzl", "py_binary", "py_test")
```

re: transitioning platforms: I pushed https://github.com/bazelbuild/rules_python/pull/846/commits/1053ab06b16c7e346e8b6a253fd3c45bcc4e686d with this. Indeed much simpler. Thanks!

python/platforms/transition.bzl
f0rmiga: Addressed in d813740.

python/platforms/transition.bzl
f0rmiga: Addressed in 161f068.

python/platforms/transition.bzl
f0rmiga: I actually missed a few test and binary-specific attributes. Addressed in 66b33f3.

python/platforms/transition.bzl
f0rmiga: Addressed in cecb3c8.

python/config_settings/transition.bzl
```diff
@@ -0,0 +1,200 @@
...
+            files = depset([zipfile_symlink] if zipfile_symlink else [], transitive = [target[DefaultInfo].files]),
+            runfiles = ctx.runfiles([zipfile_symlink] if zipfile_symlink else []).merge(target[DefaultInfo].default_runfiles),
+        ),
+        target[PyInfo],
+        target[PyRuntimeInfo],
+        target[InstrumentedFilesInfo],
+        target[OutputGroupInfo],
+        # TODO(f0rmiga): testing.TestEnvironment is deprecated in favour of RunEnvironmentInfo but
+        # RunEnvironmentInfo is not exposed in Bazel < 5.3.
+        # https://github.com/bazelbuild/bazel/commit/dbdfa07e92f99497be9c14265611ad2920161483
```
rickeylev, line 53: Ugh, this is annoying. Does checking hasattr(native, "RunEnvironmentInfo") work?

Otherwise lets mention this FR in the comment for now: https://github.com/bazelbuild/bazel/issues/8305

I also filed #901 for this.

iirc, TestEnvironment is an alias to RunEnvironmentInfo, so it's NBD really.

python/config_settings/transition.bzl
```diff
@@ -0,0 +1,200 @@
...
+    "python_version": attr.string(
+        mandatory = True,
+    ),
+    "srcs": attr.label_list(
+        allow_files = True,
+        mandatory = False,
+    ),
+    "target": attr.label(
+        executable = True,
+        cfg = _transition_python_version,
```
rickeylev, line 77: You'll want to apply the transition to the whole rule (`rule(cfg=...)`, otherwise things like deps, srcs, and tools won't select based upon the same version of python as what's in the `target` attribute.

python/config_settings/transition.bzl
```diff
@@ -0,0 +1,200 @@
...
+        env[k] = ctx.expand_location(v)
+    providers = [
+        DefaultInfo(
+            executable = executable,
+            files = depset([zipfile_symlink] if zipfile_symlink else [], transitive = [target[DefaultInfo].files]),
+            runfiles = ctx.runfiles([zipfile_symlink] if zipfile_symlink else []).merge(target[DefaultInfo].default_runfiles),
+        ),
+        target[PyInfo],
+        target[PyRuntimeInfo],
+        target[InstrumentedFilesInfo],
```
rickeylev, line 49: To make sure the coverage instrumentation filters match the wrapped binary, specify the target attribute as one of the dependency attributes. The other attributes with the common names (srcs, deps, etc) should be automatically handled.
```suggestion
        # Ensure that the binary we're wrapping is included in code coverage.
        coverage_common.instrumented_files_info(
            ctx,
            dependency_attributes = ["target"],
        ),      
```

python/config_settings/transition.bzl
```diff
@@ -0,0 +1,200 @@
...
+        expected_target_path = target[DefaultInfo].files_to_run.executable.short_path[:-4] + ".zip"
+        for file in target[DefaultInfo].default_runfiles.files.to_list():
+            if file.short_path == expected_target_path:
+                zipfile = file
+        zipfile_symlink = ctx.actions.declare_file(ctx.attr.name + ".zip")
+        ctx.actions.symlink(
+            is_executable = True,
+            output = zipfile_symlink,
+            target_file = zipfile,
+        )
```
rickeylev, line 37: So I understand what this logic is doing, but why is it doing it? The underlying zip file will be there. Is it just to make sure that "{name}.zip" shows up alongside the regular "{name}" instead of the underlying "_{name}.zip"?

That's fine, but why does it matter? Please add a brief comment to explain.

It's unfortunate we have to flatten here; I don't see a way to avoid that. A minor improvement would be to flatten the default outputs instead of runfiles (it'd be a smaller depset); though I'm not sure how much it'll actually help because, under the hood, building the zip requires flattening anyways.

python/pip.bzl
```diff
@@ -163,3 +164,200 @@ def pip_parse(requirements = None, requirements_lock = None, name = "pip_parsed_
         bzlmod = bzlmod,
         **kwargs
     )
+
+def _multi_pip_parse_impl(rctx):
```
rickeylev, line 168: I don't understand the "why" of these pip/wheel changes, so lgtm!

python/config_settings/BUILD.bazel
```diff
@@ -0,0 +1,12 @@
+load("//python:versions.bzl", "TOOL_VERSIONS")
+load(":config_settings.bzl", "construct_config_settings")
+
+filegroup(
+    name = "distribution",
```
rickeylev, line 5: What are these distribution targets?

rickeylev (changes requested):
Mostly looks good! Ok to merge after addressing comments.

python/config_settings/BUILD.bazel
f0rmiga: Used to collect sources for bazel-in-bazel integration tests.

python/config_settings/transition.bzl
f0rmiga: Addressed in 906bb6a.

python/config_settings/transition.bzl
f0rmiga: Addressed in 9c833f1.

python/config_settings/transition.bzl
f0rmiga: Checking if `native` has it works, thanks for the suggestion. Addressed in aa0fc9f.

python/config_settings/transition.bzl
```diff
@@ -74,7 +81,7 @@ _COMMON_ATTRS = {
     ),
     "target": attr.label(
         executable = True,
-        cfg = _transition_python_version,
+        cfg = "exec",
```
rickeylev, line 84: This needs to be cfg=target, otherwise it could build for a different platform that what it'll run on. e.g., with remote execution, it might build for linux, while locally you build and run a test for mac.

```suggestion
        cfg = "target",
```

rickeylev (changes requested):

python/config_settings/transition.bzl
f0rmiga: Of course... addressed in 6896f15.

rickeylev (approved):

PR #847, opened 2022-10-05
fix: assert expected bazel version

f0rmiga:
Asserts that the expected Bazel version is in sync between `.bazelversion` and `version.bzl`.


groodt (approved):
LGTM

PR #855, opened 2022-10-12
refactor: ci with multiple tasks

f0rmiga:
Refactor CI to run examples as integration tests as separate tasks. This makes CI faster and improves the iteration experience when developing features multi-platform.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -3,60 +3,148 @@ buildifier:
   version: latest
   # keep this argument in sync with .pre-commit-config.yaml
   warnings: "all"
-all_targets: &all_targets
+.reusable_config: &reusable_config
   build_targets:
+    - "--"
     - "..."
     # As a regression test for #225, check that wheel targets still build when
     # their package path is qualified with the repo name.
     - "@rules_python//examples/wheel/..."
+     # Gazelle is not fully Windows compatible: https://github.com/bazelbuild/bazel-gazelle/issues/1122
+    - "-//gazelle/..."
   build_flags:
-      - "--keep_going"
-  # We control Bazel version in integration tests, so we don't need USE_BAZEL_VERSION for tests.
-  skip_use_bazel_version_for_test: true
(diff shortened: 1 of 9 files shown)


examples/BUILD
```diff
@@ -20,24 +20,31 @@ licenses(["notice"])  # Apache 2.0
 bazel_integration_test(
     name = "pip_install_example",
     timeout = "long",
+    tags = ["integration-test"],
```
alexeagle, line 23: nit: why doesn't `bazel_integration_test` add the tags? it's already a macro

.bazelci/presubmit.yml
```diff
@@ -3,60 +3,148 @@ buildifier:
...
+    name: build_file_generation integration tests on macOS
+    working_directory: examples/build_file_generation
+    platform: macos
+  integration_test_build_file_generation_windows:
+    <<: *reusable_build_test_all
+    name: build_file_generation integration tests on Windows
+    working_directory: examples/build_file_generation
+    platform: windows
+
+  integration_test_bzlmod_linux:
```
alexeagle, line 70: could you use https://github.com/bazelbuild/continuous-integration/blob/5a1a944d644bac905da8cab4111b14b6497b7316/buildkite/README.md#matrix-testing here to reduce the repetition?

examples/BUILD
f0rmiga: Right, can do that.

.bazelci/presubmit.yml
f0rmiga: I didn't use it because I knew some wouldn't work on Windows. I fixed most of them, but there's still one test that needs fixing, then we can move to a matrix.

examples/BUILD
f0rmiga: Fixed in https://github.com/bazelbuild/rules_python/pull/855/commits/913daf82a4503e21038d6b235e2e29af76cdd7b2.

alexeagle (approved):

Issue #382, opened 2020-11-19
sys.path[0] breaks out of runfile tree.


dillon-giacoppo:
# 🐞 bug report

### Affected Rule

`py_binary`, `py_test`.

### Is this a regression?

No

### Description

When Python initializes it adds the directory of the script to the [`sys.path`](https://docs.python.org/3/library/sys.html#sys.path):

> As initialized upon program startup, the first item of this list, path[0], is the directory containing the script that was used to invoke the Python interpreter. If the script directory is not available (e.g. if the interpreter is invoked interactively or if the script is read from standard input), path[0] is the empty string, which directs Python to search modules in the current directory first. Notice that the script directory is inserted before the entries inserted as a result of PYTHONPATH.

This would imply that a `py` target such as `{REPO_DIR}/bazel-bin/{target_path}/{target}.runfiles/{script_path}/{python_script}.py` would cause `{REPO_DIR}/bazel-bin/{target_path}/{target}.runfiles/{script_path}/` to be inserted as `sys.path[0]`.

However, because the script is a symlink to the _real_ python script in the repository, Python follows the symlink and appends the actual underlying directory to `sys.path[0]`, in this case: `{REPO_DIR}/{script_path}/`. This issue was raised in Python's bug tracker and noted as expected behaviour that won't be fixed [issue17639](https://bugs.python.org/issue17639).

This allows the script to "break out" of the runfiles tree and import code from the same directory which it does not necessarily have access to via `src` or `deps`.

This causes issues when creating multiple targets in the same directory which should not automatically depend on each other.

For example, given:

```
repo/
  src/
    main.py
    foo.py
```

`main` can depend on `foo` without the file being explicitly listed in the BUILD file.

## 🔬 Minimal Reproduction

`src/BUILD`
```py3
package(default_visibility = ["//visibility:public"])

py_binary(
    name = "main",
    srcs = ["main.py"],
    python_version = "PY3",
    srcs_version = "PY3",
)
```

`src/main.py`
```py3
import sys

import foo

print(sys.path[0])
```

`src/foo.py`
```py3
print("foo imported")
```

The output of `bazel run //src:main` will be:

```
{REPO}/src/
foo imported
```

Which should not be possible where `foo.py` is not an input to `//src:main`.

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
macOS 11.0.1 (20B29)
  </code>
</pre>

**Output of `bazel version`:**

<pre>
  <code>
[...]

PR #863, opened 2022-10-20
An attempt to run Python exe in a more hermetic way

mvukov:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Current custom toolchains allow the repo path and user's local Python packages to sneak in the Python path. For instance, `sys.path` can be something like:

```
[
  "/home/mvukov/dev/tmp/hermetic_python",  # The repo root.
  "/home/mvukov/.cache/bazel/_bazel_mvukov/bc9f8eb0563a76ddb0353ce7869ddf8d/execroot/hermetic_python/bazel-out/k8-fastbuild/bin/demo.runfiles",
  "/home/mvukov/.cache/bazel/_bazel_mvukov/bc9f8eb0563a76ddb0353ce7869ddf8d/execroot/hermetic_python/bazel-out/k8-fastbuild/bin/demo.runfiles/hermetic_python",
  "/home/mvukov/.cache/bazel/_bazel_mvukov/bc9f8eb0563a76ddb0353ce7869ddf8d/execroot/hermetic_python/bazel-out/k8-fastbuild/bin/demo.runfiles/python3_8_x86_64-unknown-linux-gnu",
  "/home/mvukov/.cache/bazel/_bazel_mvukov/bc9f8eb0563a76ddb0353ce7869ddf8d/external/python3_8_x86_64-unknown-linux-gnu/lib/python38.zip",
  "/home/mvukov/.cache/bazel/_bazel_mvukov/bc9f8eb0563a76ddb0353ce7869ddf8d/external/python3_8_x86_64-unknown-linux-gnu/lib/python3.8",
  "/home/mvukov/.cache/bazel/_bazel_mvukov/bc9f8eb0563a76ddb0353ce7869ddf8d/external/python3_8_x86_64-unknown-linux-gnu/lib/python3.8/lib-dynload",
  "/home/mvukov/.local/lib/python3.8/site-packages",  # The user's local folder.
  "/home/mvukov/.cache/bazel/_bazel_mvukov/bc9f8eb0563a76ddb0353ce7869ddf8d/external/python3_8_x86_64-unknown-linux-gnu/lib/python3.8/site-packages"
]
```

with a toolchain registered as

```
python_register_toolchains(
    name = "python3_8",
    # Available versions are listed in @rules_python//python:versions.bzl.
    # We recommend using the same version your team is already standardized on.
    python_version = "3.8",
)
```
## What is the new behavior?

With the proposed fix, based on [rules_py](https://github.com/aspect-build/rules_py/blob/6c90da27ee6495f93223aa1a5602264c59cbb919/py/private/utils.bzl#L4) I created a small wrapper script for the interpreter that removes the repo root and the local folder from the path.

- [ ] Yes
- [x] No

## Other information

I didn't look into a wrapper for Windows yet. Please let me know if this could end up in the main branch and based on that I can proceed further. An alternative could be to have an extra argument like `interpreter_flags` in `python_register_toolchains`.


mvukov:
No sure why CI fails. When I run `bazel test //...` locally, all green. Can anyone provide more info, please?

python/repositories.bzl
```diff
@@ -179,6 +179,16 @@ def _python_repository_impl(rctx):
             "share/**",
         ]
 
+    python_wrapper = "python_wrapper.sh"
+    python_wrapper_content = """\
+#!/bin/bash
+
+set -o errexit -o nounset -o pipefail
+
+external/{repo_name}/{python_bin} -B -s -I "$@"
```
rickeylev, line 188: Do we want -S (upper case) as well?

rickeylev, line 188: I think the "external/" part here might depend on if a flag is set?
https://bazel.build/reference/command-line-reference#flag--legacy_external_runfiles

I think you can drop the "external/" part, but not 100% sure.

Actually, what's the CWD when this is invoked? The relative path here means it would have to be the runfiles root. But what is doing that chdir()? (Sorry if I missed it)

groodt:
I've got a few questions about this:
* Whats the plan for Windows?
* How does this inter-play with the [python_stub_template.txt](https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt)

Is this the right place to be solving the problem? To me, this feels like a similar problem to:
https://github.com/bazelbuild/rules_python/issues/691, but now we've introduced a hardcoded dependency on the host of `/bin/bash`.

python/repositories.bzl
mvukov: I don't think this would harm in the case a custom toolchain from this repo is used. In the case a system python is used, this may even be desirable.

mvukov: Thanks for pointing this out. If `--nolegacy_external_runfiles`, then this path doesn't work. The CWD in my example is

```bash
/home/mvukov/.cache/bazel/_bazel_mvukov/bc9f8eb0563a76ddb0353ce7869ddf8d/execroot/hermetic_python/bazel-out/k8-fastbuild/bin/demo.runfiles/hermetic_python
```
where hermetic_python is the repo name. When `--legacy_external_runfiles` is used, then `external/` has to be used...

> But what is doing that chdir()? (Sorry if I missed it)
I'm not doing any chdir in this PR :) Can you be a bit more specific?

mvukov:
In the meantime, I found that using -I won't work as per doc "-I     : isolate Python from the user's environment (implies -E and -s)". This effectively nulls PYTHONPATH set by the stub. I removed that in [10ffa53](https://github.com/bazelbuild/rules_python/pull/863/commits/10ffa539de5b8abba67d917a354f11f5113fa011). After this, `sys.path` outputs for this example:

```
[
  "/home/mvukov/dev/tmp/hermetic_python",
  "/home/mvukov/.cache/bazel/_bazel_mvukov/bc9f8eb0563a76ddb0353ce7869ddf8d/execroot/hermetic_python/bazel-out/k8-fastbuild/bin/demo.runfiles",
  "/home/mvukov/.cache/bazel/_bazel_mvukov/bc9f8eb0563a76ddb0353ce7869ddf8d/execroot/hermetic_python/bazel-out/k8-fastbuild/bin/demo.runfiles/hermetic_python",
  "/home/mvukov/.cache/bazel/_bazel_mvukov/bc9f8eb0563a76ddb0353ce7869ddf8d/execroot/hermetic_python/bazel-out/k8-fastbuild/bin/demo.runfiles/python3_8_x86_64-unknown-linux-gnu",
  "/home/mvukov/.cache/bazel/_bazel_mvukov/bc9f8eb0563a76ddb0353ce7869ddf8d/external/python3_8_x86_64-unknown-linux-gnu/lib/python38.zip",
  "/home/mvukov/.cache/bazel/_bazel_mvukov/bc9f8eb0563a76ddb0353ce7869ddf8d/external/python3_8_x86_64-unknown-linux-gnu/lib/python3.8",
  "/home/mvukov/.cache/bazel/_bazel_mvukov/bc9f8eb0563a76ddb0353ce7869ddf8d/external/python3_8_x86_64-unknown-linux-gnu/lib/python3.8/lib-dynload",
  "/home/mvukov/.cache/bazel/_bazel_mvukov/bc9f8eb0563a76ddb0353ce7869ddf8d/external/python3_8_x86_64-unknown-linux-gnu/lib/python3.8/site-packages"
]
```
Still, it's a pity that the first entry still sneaks in the path.

> Whats the plan for Windows?

If we decide to move forward with this PR, I'd work on support for Windows.

> How does this inter-play with the [python_stub_template.txt](https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/bazel/rules/python/python_stub_template.txt)

The generated stub driven by a platform interpreter calls the generated wrapper, the wrapper then calls a custom interpreter.

> Is this the right place to be solving the problem? To me, this feels like a similar problem to:
https://github.com/bazelbuild/rules_python/issues/691, but now we've introduced a hardcoded dependency on the host of /bin/bash.

Not sure, to be honest. I am looking into an intermediate solution until https://github.com/bazelbuild/bazel/issues/15897 gets implemented or until https://github.com/aspect-build/rules_py gets out of experimental.

TBH, I would be just happy at the moment if py_runtime had an extra argument `interpreter_flags` that could be hard-coded in the stub. Would this alternative be better?

mattem:
👋 As you found, `-I` won't work with rules_python, as it makes Python ignore all env vars (including `PYTHON_PATH` that bazel crafts), so it effectively breaks us. 

rules_py on the other hand can set this, (I'm the author of rules_py 😄) as it doesn't use the runfiles layout for running `py_binary` etc, so we don't need to directly meddle with `PYTHON_PATH`, so therefore can set the `-I` flag, and deals with the side affects.

It's probably worth setting the new `-P` flag in Python 3.11 that was designed for this, but we'd have to detect the interpreter version etc. https://docs.python.org/3.11/using/cmdline.html#cmdoption-P

On the topic of rules_py, I can help with issues using it to push it out of experimental if needed.

jvolkman:
This is related to https://github.com/bazelbuild/rules_python/issues/382 and https://github.com/bazelbuild/bazel/issues/7091. Also worth noting that @groodt already added `PYTHONSAFEPATH` (the env var equivalent of `-P`) to the stub template in https://github.com/bazelbuild/bazel/pull/15701, but of course that only applies to 3.11.

mvukov:
@mattem Many thanks. I'll try rules_py with a large monorepo to see how that works.

mvukov:
I tested https://github.com/aspect-build/rules_py on a large monorepo and works as expected (after a couple of minor issues were resolved -- the current main branch works fine). Haven't tried out remote execution, but, the ruleset plays well now with remote caching. It would be nice to have rules_python and rules_py in the same repo eventually. :)

PR #865, opened 2022-10-24
Support annotations on pip packages with extras.

william-smith-skydio:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

When using `pip_parse` with packages with extras, e.g. `requests[security]>=2.8.1`, it is not possible to add annotations. The `requests` package will otherwise work correctly, but trying to add an annotation to `requests` will not result in the annotation actually being applied. This is because annotations have separate logic for parsing requirements in the generated .bzl file. It would previously turn the requirement into `requests[security]` rather than just `requests` and fail to match, as just `requests` is expected.

## What is the new behavior?

The custom parsing logic has been changed to support extras, e.g. `requests[security]>=2.8.1`.

- [ ] Yes
- [x] No


f0rmiga (approved):
LGTM. @hrfuller could you take a second pass, please?

hrfuller (approved):
LGTM as well. IIRC the requirement strings come from the requirements lock file. Ideal the lock file should have "simple" requirement strings and then extras should just be on their own lines in the lock file. Given that pip_compile doesn't work like that this seems like the best solution.

PR #867, opened 2022-10-25
Update bzlmod example to use latest Bazel release 

kormide:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Build related changes
- [ ] CI related changes

The existing bzlmod example does not support the latest version of Bazel.

## What is the new behavior?

Bzlmod example uses a prerelease version of Bazel 6 and properly overrides the module path during integration testing.

- [ ] Yes
- [x] No

### DIFF
--- a/examples/BUILD
+++ b/examples/BUILD
@@ -40,4 +40,6 @@ bazel_integration_test(
 
 bazel_integration_test(
     name = "bzlmod_example",
+    bzlmod = True,
+    override_bazel_version = "6.0.0rc1",
 )
--- a/tools/bazel_integration_test/test_runner.py
+++ b/tools/bazel_integration_test/test_runner.py
@@ -53,6 +53,15 @@ def main(conf_file):
                     % os.environ["TEST_SRCDIR"]
                 )
 
+                # TODO: --override_module isn't supported in the current BAZEL_VERSION (5.2.0)
+                # This condition and attribute can be removed when bazel is updated for
+                # the rest of rules_python.
+                if (config["bzlmod"]):
+                    bazel_args.append(
+                        "--override_module=rules_python=%s/rules_python"
+                        % os.environ["TEST_SRCDIR"]
+                    )
+
                 # Bazel's wrapper script needs this or you get
                 # 2020/07/13 21:58:11 could not get the user's cache directory: $HOME is not defined
                 os.environ["HOME"] = str(tmp_homedir)
--- a/internal_setup.bzl
+++ b/internal_setup.bzl
@@ -31,6 +31,13 @@ def rules_python_internal_setup():
     # Depend on the Bazel binaries for running bazel-in-bazel tests
     bazel_binaries(versions = SUPPORTED_BAZEL_VERSIONS)
 
+    # Bazel 5.3.0 has bzlmod bugs so we use 6.0 prerelease for the bzlmod example.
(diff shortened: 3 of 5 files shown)


kormide:
FYI @alexeagle @Wyverald

tools/bazel_integration_test/test_runner.py
```diff
@@ -52,6 +52,19 @@ def main(conf_file):
                     "--override_repository=rules_python=%s/rules_python"
                     % os.environ["TEST_SRCDIR"]
                 )
+                bazel_args.append(
+                    "--override_module=rules_python=%s/rules_python"
+                    % os.environ["TEST_SRCDIR"]
+                )
```
kormide, line 58: This is the equivalent of `--override_repository` but for MODULE deps.

tools/bazel_integration_test/test_runner.py
```diff
@@ -52,6 +52,19 @@ def main(conf_file):
...
+
+                # TODO: Despite not declaring dependencies on these in MODUILE.bzl
+                # or in the bzlmod example's MODULE.bzl, bazel is complaining about
+                # the following versions being yanked. For now, allow all yanked
+                # versions in integration tests.
+                #
+                #   protobuf@3.19.2
+                #   zlib@1.2.12
+                bazel_args.append("--allow_yanked_versions=all")
```
kormide, line 67: @Wyverald Any idea why this was required? Those versions are indeed yanked in the BCR, but I don't have any direct or transitive dependencies on them via bzlmod (that I know of).

examples/BUILD
```diff
@@ -40,4 +40,5 @@ bazel_integration_test(
 
 bazel_integration_test(
     name = "bzlmod_example",
+    override_bazel_version = "6.0.0-pre.20221012.2",
```
alexeagle, line 43: there's a 6.0.0-rc0 this morning, maybe use that?

alexeagle (approved):
Won't merge until we get an answer on the yanked problem

examples/BUILD
kormide: That release tag doesn't appear to be available on GitHub. I think the one I'm referencing here is the latest.

tools/bazel_integration_test/test_runner.py
fmeum: It works with `last_green` and was fixed by https://github.com/bazelbuild/bazel/commit/d6d6183dc17484a8d5addcd68a7480ce7a5a97f6. While your module didn't depend on the yanked `zlib`, `@bazel_tools` did.

examples/BUILD
Wyverald: it should be 6.0.0-rc1

Wyverald: sorry, "6.0.0rc1" without the dash

tools/bazel_integration_test/test_runner.py
Wyverald: Yep -- `bazel_tools` is a built-in module that depends on stuff like `rules_cc`, `rules_java`, etc. Unfortunately it also depends on these two yanked versions, but using 6.0.0rc1 should fix this.

kormide: That works, thanks!

examples/BUILD
alexeagle: oops sorry I was too fast and inaccurate :)

alexeagle (approved):

PR #868, opened 2022-10-25
Updating docs

chrislovecnm:
I ran into the problem where I did not have python installed, and I did not want to install it.  Hey I use bazel! When you do not have python installed you cannot use pre-submit to configure running buildifier.  It took me forever to figure out how to run it, so I included the command in the CONTRIBUTING.md.


groodt (approved):
LGTM

PR #870, opened 2022-10-28
Support bzlmod

kormide:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

The requirements file is parsed in a python script that generates `requirements.bzl` and passes information from the file into `whl_library` called by `install_deps` within the generated file. In order to declare the wheel repositories in a bzlmod extension, we need to be able to parse that same information in Starlark.

## What is the new behavior?

Can parse a requirements file in Starlark which will be uses in a pip_parse bzlmod extension. This in in a similar vein to how `rules_js` parses a pnpm lockfile in starklark for the `npm_translate_lock` extension.

- [ ] Yes
- [x] No

## Other information

This isn't currently used, but it's a nice reviewable chunk for the bzlmod work.


kormide:
fyi: @alexeagle, @Wyverald

alexeagle:
My understanding is that something was added after you wrote the yaml parser, which permits one module extension to see a workspace that was created by another, and so perhaps that means it's possible to fetch some tool (like yq) and then execute it from a module extension.

Perhaps that means we could use an existing requirements.txt parser like from pip or poetry that's likely to handle all the edge cases correctly. I doubt it, but might be worth a bit of poking.

python/pip_install/private/test/requirements_parser_tests.bzl
```diff
@@ -0,0 +1,185 @@
...
+    # via requests
+"""))
+
+    return unittest.end(env)
+
+def _parse_requirements_lockfile_test_impl(ctx):
+    env = unittest.begin(ctx)
+
+    asserts.equals(env, [
+        ("certifi", "certifi==2021.10.8     --hash=sha256:78884e7c1d4b00ce3cea67b44566851c4343c120abd683433ce934a68ea58872     --hash=sha256:d62a0163eb4c2344ac042ab2bdf75399a71a2d8c7d47eac2e2ee91b9d6339569"),
```
alexeagle, line 67: do you only need the name for the module extension implementation, and not the hashes? I thought you'd need all the data required to produce a requirements.bzl file, like the hashes for downloading the packages.

OH I remember - the pip_parse rule doesn't actually know how to download anything, it will still end up calling `pip install` on individual packages which are required by the build...

alexeagle (commented):
Maybe @hrfuller can help with reviews as well, he has a good memory and wrote the pip_parse stuff.

kormide:
> My understanding is that something was added after you wrote the yaml parser, which permits one module extension to see a workspace that was created by another, and so perhaps that means it's possible to fetch some tool (like yq) and then execute it from a module extension.
> 
> Perhaps that means we could use an existing requirements.txt parser like from pip or poetry that's likely to handle all the edge cases correctly. I doubt it, but might be worth a bit of poking.

Yeah originally I wanted to call `pip_parse` in the nodule extension but have it output an extra JSON file containing the result of parsing requirements that I could easily load in starlark. But where I got stuck was finding some what to load that file from the pip_parse repo within the extension. I'm not sure that's possible. But maybe I could call `module_ctx.execute` to execute a python script that uses the same parsing library? Can a module extension create and own a file like that?

python/pip_install/private/test/requirements_parser_tests.bzl
kormide: Yeah this is the same requirement info that gets passed to `whl_library`.

Wyverald:
> Yeah originally I wanted to call pip_parse in the module extension but have it output an extra JSON file containing the result of parsing requirements that I could easily load in starlark.

That's not directly possible (you can't declare a repo in an extension and immediately read from that repo), but you can use the trick mentioned by Alex: have an extension declare a `pip_parse` repo, `use_repo` that repo in your module, and then load from it in another extension.

> But maybe I could call module_ctx.execute to execute a python script that uses the same parsing library? Can a module extension create and own a file like that?

Alternatively, this works too. A module extension can create files, but these files are not addressable after the extension impl function finishes.

kormide:
> > Yeah originally I wanted to call pip_parse in the module extension but have it output an extra JSON file containing the result of parsing requirements that I could easily load in starlark.
> 
> That's not directly possible (you can't declare a repo in an extension and immediately read from that repo), but you can use the trick mentioned by Alex: have an extension declare a `pip_parse` repo, `use_repo` that repo in your module, and then load from it in another extension.
> 
> > But maybe I could call module_ctx.execute to execute a python script that uses the same parsing library? Can a module extension create and own a file like that?
> 
> Alternatively, this works too. A module extension can create files, but these files are not addressable after the extension impl function finishes.

Thanks for the clarification. That's interesting that a module context can create temporary files. I think one extension is a better developer experience than two, so I'll look into that approach and see where it leads me.

python/pip_install/private/requirements_parser.bzl
```diff
@@ -0,0 +1,89 @@
...
+    if input == "\n":
+        if len(requirements) > 0 and len(requirements[-1]) == 1:
+            return (_STATE.ParseRequirement, buffer)
+        return (_STATE.ConsumeSpace, "")
+    return (_STATE.ConsumeComment, buffer)
+
+def _handleParseDependency(input, buffer, requirements):
+    if input == EOF:
+        fail("Enountered unexpected end of file while parsing requirement")
+    if input.isspace() or input in [">", "<", "~", "=", ";", "["]:
```
hrfuller, line 74: I'm not sure it is valid for a constraint portion of a requirement string to start with a semi-colon, but you should add a test case for it anyway. e.g `parse("SomePackage;") == ("SomePackage", "SomePackage;")`

hrfuller (approved):
Cool use of a state machine, thanks for working on this.

groodt:
Looks good. I've got a different proposal to consider and wondered what you might think.

1. Update rules_python to consume a pip "installation report" instead of a transitively locked requirements.txt https://pip.pypa.io/en/stable/reference/installation-report/
2. Parse the locked json report with https://bazel.build/rules/lib/json#decode

The benefits of the json installation report has a few benefits:
* It has the direct urls
* It captures the environment (sys info) of the locking platform

The direct url info in particular will be helpful in future, because it would enable us to parse the format into `.bzl` http_file or http_archive for retrieval in repository rules using the bazel machinery without having to invoke pip.

Wyverald:
> The direct url info in particular will be helpful in future, because it would enable us to parse the format into .bzl http_file or http_archive for retrieval in repository rules using the bazel machinery without having to invoke pip.

This might be built into Bzlmod in the future, without you guys having to do anything. With the planned lockfiles feature, Bzlmod will store the results of module extension resolution in the lockfile; said result is actually just the names and definitions of the repos that the extension generated.

groodt:
> > The direct url info in particular will be helpful in future, because it would enable us to parse the format into .bzl http_file or http_archive for retrieval in repository rules using the bazel machinery without having to invoke pip.
> 
> This might be built into Bzlmod in the future, without you guys having to do anything. With the planned lockfiles feature, Bzlmod will store the results of module extension resolution in the lockfile; said result is actually just the names and definitions of the repos that the extension generated.

That would be something different to what I mention above. That would be for locking the bazel rules dependency graph, not the dependency graph(s) of resolved python packages from external sources. There is no mechanism for bzlmod to do that presently because we aren’t capturing it presently. Hence my line of questioning.

alexeagle:
I don't think it's ever a good idea for Bazel to manage the lockfile when a language-idiomatic package manager manages the dependencies and constraints. Then you get a different result outside of Bazel and the language tooling gets confused.

Anyhow this PR is just to make a bzlmod module extension for the existing repository rules, not to redesign them.

python/pip_install/private/requirements_parser.bzl
kormide: I thought I saw an example of this somewhere, but I'll double check/add a test case.

kormide:
I'm going to hold off on this for now and see if I can just execute a python script in the module extension to do the parsing (using the same library that pip_parse uses).

python/pip_install/private/requirements_parser.bzl
kormide: There are some examples here: https://pip.pypa.io/en/stable/reference/requirement-specifiers/#requirement-specifiers
E.g.
```
SomeProject ; sys_platform == 'win32'
```
Looks like I already do have a test case.

fmeum:
> Thanks for the clarification. That's interesting that a module context can create temporary files. I think one extension is a better developer experience than two, so I'll look into that approach and see where it leads me.

@kormide Note that even though `rules_python` would have to define two module extensions, end users would not need to interact with the one that provides the tools (such as `yq`) for the other.

kormide:
> > Thanks for the clarification. That's interesting that a module context can create temporary files. I think one extension is a better developer experience than two, so I'll look into that approach and see where it leads me.
> 
> @kormide Note that even though `rules_python` would have to define two module extensions, end users would not need to interact with the one that provides the tools (such as `yq`) for the other.

That's good to know. I had ruled that out because I thought that users would have to declare both extensions.

I'm going to put up a solution with a single module extension first because I'm close to getting that working, and if it feels like there's too much duplication I can go back to the two extension approach.

kormide:
I repurposed this PR to include the full bzlmod solution.

Commit 1: requirements starlark parser (added support for parsing pip options)
Commit 2: bzlmod extension
Commit 3: bzlmod release support (workspace snippet, BCR app config)

I ended up going with the starlark parsing of requirements as an initial solution here. The alternative is to add an additional module extension that parses the requirements file and outputs something like JSON that starlark can read. In the end, (I think) we need to have the `whl_library` declarations inside of the pip_parse extension, hence the need for either a parser or another extension to load the wheel arguments.

.bcr/template.metadata.json
```diff
@@ -0,0 +1,20 @@
...
+      "name": "Richard Levasseur",
+      "email": "rlevasseur@google.com",
+      "github": "rickeylev"
+    },
+    {
+      "name": "Greg Roodt",
+      "email": "groodt@gmail.com",
+      "github": "groodt"
+    }
+  ],
```
kormide, line 14: This is a template for an entry that will appear the Bazel Central Registry ([example](https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/aspect_rules_js/metadata.json)).

I wasn't sure who the main maintainers of rules_python are. @rickeylev @groodt are you okay with being listed here?

python/pip_install/extract_wheels/parse_requirements_to_bzl.py
```diff
@@ -131,20 +153,29 @@ def generate_parsed_requirements_contents(
...
         _config = {args}
         _annotations = {annotations}
+        _bzlmod = {bzlmod}
 
         def _clean_name(name):
             return name.replace("-", "_").replace(".", "_").lower()
 
         def requirement(name):
+            if _bzlmod:
+                return "@@{repo}//:" + _clean_name(name) + "_{py_library_label}"
```
kormide, line 163: This points calls to `xyz_requirement(...)` to the aliases inside the pip_parse repo, so that we don't need to declare a bunch of `use_repo`s in MODULE.bazel.

kormide:
Just realized that I still need to support registering the python toolchain via bzlmod. The tests are failing on CI because the bzlmod example doesn't use a consistent python version on all platforms (I think it's using whatever the executor has installed), and it's outputting the python version to the vendored requirements file, causing a diff_test to fail.

kormide:
Added a module extension for registering a python toolchain.

examples/bzlmod/MODULE.bazel
```diff
@@ -10,3 +10,25 @@ local_path_override(
     module_name = "rules_python",
     path = "../..",
 )
+
+python = use_extension("@rules_python//python:extensions.bzl", "python")
+
+python.toolchain(
```
fmeum, line 16: Does every root module have to manually register a toolchain or is the host Python used if nothing else is specified?

kormide: In the case of this example, I had to use it to fix the python version because the generated `requirements_lock.txt` file stamps the python version in a comment, and on CI the different platforms had different versions installed, so the diff_test was failing.

.bcr/template.metadata.json
rickeylev: SGTM

f0rmiga: @kormide you can also add me.

.bcr/template.metadata.json
```diff
@@ -0,0 +1,20 @@
...
+    {
+      "name": "Richard Levasseur",
+      "email": "rlevasseur@google.com",
+      "github": "rickeylev"
+    },
+    {
+      "name": "Greg Roodt",
+      "email": "groodt@gmail.com",
+      "github": "groodt"
+    }
```
f0rmiga, line 13:
```suggestion
    },
    {
      "name": "Thulio Ferraz Assis",
      "email": "thulio@aspect.dev",
      "github": "f0rmiga"
    }
```

kormide: :+1: added

kormide:
This PR needs another review since adding the full bzlmod solution, but it's not clear to me who I should ping for that.

f0rmiga:
@alexeagle is probably the best person to do the final review.

alexeagle (approved):
nice! sorry it took me a while to get to this

.bcr/config.yml
```diff
@@ -0,0 +1,3 @@
+fixedReleaser:
```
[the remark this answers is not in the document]
kormide: The publish-to-bcr app needs to know who the releaser is in order to:
* send error emails
* tag their username in the bcr PR
* make them the commit author

When a repository uses release automation and the author of a release is the `github-actions` bot (rules_python does this), then the user who caused the release is obscured. `fixedReleaser` also helps in the case where a user is the releaser author, but their email is hidden from their profile, so the app can't access it to send emails or properly attribute the commit to them.

.github/workflows/workspace_snippet.sh
```diff
@@ -9,7 +9,40 @@ PREFIX="rules_python-${TAG}"
...
+python = use_extension("@rules_python//python:extensions.bzl", "python")
+
+python.toolchain(
+    name = "python",
+    python_version = "3.9",
+)
+
+use_repo(python, "python_toolchains")
+
+register_toolchains(
```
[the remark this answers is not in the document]
kormide: I found that registering the toolchain outside of bzlmod didn't work. Note that in bazel-lib, we only handle the toolchain internally and it is registered via bzlmod in that repo's `MODULE.bazel`. The user doesn't have to do it but they also can't declare their own version via bzlmod.

.bcr/config.yml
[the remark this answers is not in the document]
alexeagle: So if you "push a tag to release", we can never fix this? What if you had to push an annotated tag to release, so that there's a commit-like object in a git ref?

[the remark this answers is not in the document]
kormide: The problem lies with the GitHub api. It uses the identity of whatever authorized token is used to create the release to set as the author. It will not look at annotated tag information and there are no fields to explicitly set the author. So if the actions GITHUB_TOKEN is used to create the release, there's nothing we can really do unless they expand the api.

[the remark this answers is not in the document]
alexeagle: okay well this isn't the PR to dig into that anyhow :)

alexeagle (approved):

PR #881, opened 2022-11-10
fix(determinism): copy two missing entries from data_excludes

alexeagle:
We install some pip packages to bootstrap the pip_install/pip_parse rules, and these were allowing .pyc files as data deps. In some clients I observe that pip install is creating these with non-determinism and busting the python action caches. We already had a correct solution for user-installed packages, so we just need to include those entries for the built-ins.


hrfuller (approved):
LGTM. Thanks for catching this.

Issue #895, opened 2022-11-23
stripped .py extension from the generated dependency name breaks builds.


rbuckland:
# 🐞 bug report

### Affected Rule

when `install_deps(..)` is run with `rpyc` as a `requirement()`, the Bazel build errors due to duplicate `py_binary` entries in the generated `pip_dependencies` set.

### Is this a regression?

Have not run a previous version of baxel, with the rpyc dependency.

### Description

Issue #702 introduced an edge case bug, where wheels with duplicate console_script entries (by name), differing by the `.py` extension fail due to duplicate dependencies being generated.

https://github.com/bazelbuild/rules_python/blob/main/python/pip_install/extract_wheels/bazel.py#L399
alters the name by stripping the extension `.py`

Which in the case of the pip dependency `rpyc`, creates duplicate `py_binary(...)` rules.

```
[console_scripts]
rpyc_classic = rpyc.cli.rpyc_classic:main
rpyc_classic.py = rpyc.cli.rpyc_classic:main
rpyc_registry = rpyc.cli.rpyc_registry:main
rpyc_registry.py = rpyc.cli.rpyc_registry:main
```
has 4x entries, pairs differeing with a `.py` extension.

this in turn results in "duplicate" py_binary() rules generated.

**(🔥 Exception)**
```
❯ bazel test lib/internal/mylib7/src:mylib7_py 
Loading: 
Loading: 0 packages loaded
Analyzing: target //lib/internal/mylib7/src:mylib7_py (0 packages loaded, 0 targets configured)
ERROR: Traceback (most recent call last):
        File "/home/user8/.cache/bazel/_bazel_user8/09d901cfcd7f772f56461320444edcec/external/pip_dependencies_310_rpyc/BUILD.bazel", line 42, column 10, in <toplevel>
                py_binary(
        File "/home/user8/.cache/bazel/_bazel_user8/09d901cfcd7f772f56461320444edcec/external/rules_python/python/private/reexports.bzl", line 79, column 21, in py_binary
                native.py_binary(**_add_tags(attrs))
Error in py_binary: py_binary rule 'rules_python_wheel_entry_point_rpyc_classic' in package '' conflicts with existing py_binary rule, defined at /home/user8/.cache/bazel/_bazel_user8/09d901cfcd7f772f56461320444edcec/external/pip_dependencies_310_rpyc/BUILD.bazel:33:10
[...]

PR #897, opened 2022-11-23
fix: #895 append _py if ending with .py

rbuckland:
## PR Checklist

## PR Type

- [X] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #895 

## What is the new behavior?

`rpyc` dependency generates separate `py_binary` rules.

- [ ] Yes
- [ ] No
- [X] Unsure


alexeagle:
My understanding is that this is breaking, yes. Users may have to add back the `_py` suffix if they're already using one of these. But it was documented as an "extreme edge case" and you're right that removing `.py` suffix is a recipe for collisions, so I think this is okay. Thanks!

alexeagle (approved):

PR #907, opened 2022-11-30
python_repository: Exclude pycache files

illicitonion:
After the first time the python_repository is used, files are created in
`__pycache__` directories. These invalidate the configured target for
the python_repository target, and cause spurious rebuilds.

Here's an example of the set of pycache files I've seen in one
repository.

```
lib/python3.8/distutils/__pycache__/__init__.cpython-38.pyc
lib/python3.8/distutils/__pycache__/archive_util.cpython-38.pyc
lib/python3.8/distutils/__pycache__/cmd.cpython-38.pyc
lib/python3.8/distutils/__pycache__/config.cpython-38.pyc
lib/python3.8/distutils/__pycache__/core.cpython-38.pyc
lib/python3.8/distutils/__pycache__/debug.cpython-38.pyc
lib/python3.8/distutils/__pycache__/dep_util.cpython-38.pyc
lib/python3.8/distutils/__pycache__/dir_util.cpython-38.pyc
lib/python3.8/distutils/__pycache__/dist.cpython-38.pyc
lib/python3.8/distutils/__pycache__/errors.cpython-38.pyc
lib/python3.8/distutils/__pycache__/extension.cpython-38.pyc
lib/python3.8/distutils/__pycache__/fancy_getopt.cpython-38.pyc
lib/python3.8/distutils/__pycache__/file_util.cpython-38.pyc
lib/python3.8/distutils/__pycache__/log.cpython-38.pyc
lib/python3.8/distutils/__pycache__/spawn.cpython-38.pyc
lib/python3.8/distutils/__pycache__/sysconfig.cpython-38.pyc
lib/python3.8/distutils/__pycache__/util.cpython-38.pyc
lib/python3.8/distutils/command/__pycache__/__init__.cpython-38.pyc
lib/python3.8/distutils/command/__pycache__/build.cpython-38.pyc
lib/python3.8/distutils/command/__pycache__/install.cpython-38.pyc
lib/python3.8/email/__pycache__/_header_value_parser.cpython-38.pyc
lib/python3.8/email/__pycache__/contentmanager.cpython-38.pyc
lib/python3.8/email/__pycache__/headerregistry.cpython-38.pyc
lib/python3.8/email/__pycache__/policy.cpython-38.pyc
```

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

If you cquery twice in a row, you get different output, because extra files were created in the repository by the first cquery invocation.

Issue Number: N/A

## What is the new behavior?
The consecutive cqueries now produce identical results.

- [ ] Yes
- [x] No

### DIFF
--- a/python/repositories.bzl
+++ b/python/repositories.bzl
@@ -196,8 +196,20 @@ def _python_repository_impl(rctx):
 
     python_bin = "python.exe" if ("windows" in platform) else "bin/python3"
 
+    glob_include = []
+
+    if rctx.attr.ignore_root_user_error:
+        glob_include += [
+            "# These pycache files are created on first use of the associated python files.",
+            "# Exclude them from the glob because otherwise between the first time and second time a python toolchain is used,",
+            "# the definition of this filegroup will change, and depending rules will get invalidated.",
+            "# See https://github.com/bazelbuild/rules_python/issues/1008 for unconditionally adding these to toolchains so we can stop ignoring them.",
+            "**/__pycache__/*.pyc",
+            "**/__pycache__/*.pyo",
+        ]
+
     if "windows" in platform:
-        glob_include = [
+        glob_include += [
             "*.exe",
             "*.dll",
             "bin/**",
@@ -210,7 +222,7 @@ def _python_repository_impl(rctx):
             "share/**",
         ]
     else:
-        glob_include = [
+        glob_include += [
             "bin/**",
             "extensions/**",
             "include/**",


alexeagle:
what version of rules_python? I made a recent fix https://github.com/bazelbuild/rules_python/commit/3f0d62d18d5dfead0701b061e2f27774db6f64d6 that's in 0.14.0

f0rmiga:
Are you running as `root` and set `ignore_root_user_error`?

illicitonion:
> what version of rules_python? I made a recent fix [3f0d62d](https://github.com/bazelbuild/rules_python/commit/3f0d62d18d5dfead0701b061e2f27774db6f64d6) that's in 0.14.0

I have repro'd in 0.15.0 - this is around the python interpreter's distro itself, rather than than when using pip.

> Are you running as `root` and set `ignore_root_user_error`?

I am. I could make the globbing conditional on `ignore_root_user_error` being set, if you think that'd be better?

f0rmiga:
When you set that attribute, it's assumed that you stumbled across https://github.com/bazelbuild/rules_python/pull/713. rules_python is not supported running as root. I'm hesitant to start making exceptions as it involves many side effects (and I'm afraid we don't even know all of them).

illicitonion:
Apologies - `ignore_root_user_error` was being set in a patch I didn't see when I started investigating this, so I actually _didn't_ see #713 before coming here.

I'm an infrastructure provider for dozens of teams who are using Bazel, many of whom are for sure running bazel as root inside docker containers - realistically to use hermetic python interpreters we need a patch like this, and we would really like to not have to keep an internal fork of `rules_python` if we can avoid it...

What's the reasoning behind making the directory readonly? Presumably part of the point is to avoid generating these pyc files, in which case, excluding them from the glob feels like it's working towards the same aim? Are these other reasons to do so too?

I've added a new commit to https://github.com/illicitonion/rules_python/tree/exclude-pycache-in-python-repository which only excludes these files if you're running as root, to minimise the surface area of the change to "people who have already opted in to ignoring reasonable advice", but I do think it may be worth merging this _without_ the only-when-root patch...

f0rmiga:
I noticed that people are running as root by accident, not intentionally. Making the pyc read-only prevents the interpreter from updating them while taking advantage of the improved performance. Why isn't it realistic to stop building as root? As I said, we (rules_python maintainers) don't understand all of the side effects that running as root brings. This was only one of the side effects we caught. Speaking more broadly, and outside Bazel, running as root defeats the whole concept of file permissions on UNIX-like systems, so if we want to keep the "correct" part of `{ Fast, Correct } — Choose two`, the least we can do is stop building as root on UNIX systems.

gibfahn:
>Why isn't it realistic to stop building as root?

(context: I work on the same team as @illicitonion )

If running in Docker using [User Namespaces](https://docs.docker.com/engine/security/userns-remap/), then you really want to be running as root inside the container so you end up mapping to a specific user outside the container. This makes e.g. cleanup tasks outside the container more performant. When running inside a single-use docker container as root, where docker is configured with user namespaces, I don't think it's the case that that "running as root defeats the whole concept of file permissions on UNIX-like systems".

We've been running builds (using a variety of build systems) as root for several years now without issue.

f0rmiga:
@gibfahn thank you for the extra information on how you use Docker. Let's schedule a call to discuss this better.

pjjw:
hey, similar situation here, curious if consensus was reached.

illicitonion:
We had a really useful call, and think the underlying issue here is that https://github.com/indygreg/python-build-standalone includes a somewhat arbitrary subset of `.pyc` files in its toolchains. @f0rmiga is going to try to arrange upstream that a complete set of `.pyc` files is included in the toolchains, so that new ones don't get re-generated, and we don't need to ignore them.

f0rmiga (approved):
@illicitonion could you update this PR? I'll merge as is and create a separate issue to address this upstream.

illicitonion:
@f0rmiga done - thanks!

Issue #908, opened 2022-11-30
compile_pip_requirements no longer uses predictable paths


vonschultz:
# 🐞 bug report

### Affected Rule

The issue is caused by the rule: `compile_pip_requirements`

### Is this a regression?

Yes, the previous version in which this bug was not present was: 0.13.0. It is present in 0.15.0. I didn't explicitly check 0.14.0, but I'll actually go one step further: a `git bisect` says the first bad commit is b15e15fb95d584df80339a5f3bdfd253fcdff2dc. (Ping @philsc  and @f0rmiga.)

### Description

As it says in python/pip_install/pip_compile.py: 
 
>    We don't want fully resolved, absolute paths in the generated requirements file.
>    The paths could differ for every invocation. Replace them with a predictable path.

I haven't double checked that that's the right file for this bug report, but the comment describes a desirable outcome that I'm no longer seeing.

## 🔬 Minimal Reproduction

See https://gist.github.com/vonschultz/9e242d1612ec70b1848057cce681e037. I'm using it with `bazelisk run :pip_dependencies.update`. 

## 🔥 Exception or Error

You see the problem in [requirements_common_lock-txt](https://gist.github.com/vonschultz/9e242d1612ec70b1848057cce681e037#file-requirements_common_lock-txt):
<pre><code>    # via -r /tmp/rules_python-tests/9e242d1612ec70b1848057cce681e037/requirements_extra.in</code></pre>

Before b15e15fb95d584df80339a5f3bdfd253fcdff2dc that line reads
<pre><code>    # via -r ./requirements_extra.in</code></pre>
which is the desired behavior.

## 🌍  Your Environment

**Operating System:**

Ubuntu 20.04

**Output of `bazel version`:**

<pre>
  <code>
Bazelisk version: v1.9.0
Build label: 5.3.2
Build target: bazel-out/k8-opt/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Wed Oct 19 18:22:12 2022 (1666203732)
Build timestamp: 1666203732
Build timestamp as int: 1666203732
  </code>
</pre>

**Rules_python version:**

<pre><code>0.15.0</code></pre>

f0rmiga:
I was able to reproduce this and it only seems to happen with the added `-r requirements_extra.in`.

philsc:
Whoops. Sorry about that. Thanks for fixing it @f0rmiga!

vonschultz:
[...]

PR #909, opened 2022-11-30
fix: pip_compile to handle multiple generated requirements.in

f0rmiga:
Fixes https://github.com/bazelbuild/rules_python/issues/908.


examples/pip_parse/BUILD
```diff
@@ -55,16 +55,6 @@ alias(
     actual = entry_point("yamllint"),
 )
 
-# The requirements.in file can be checked in to the source tree or it can be
-# generated. Pretend that we do some generating of the file. For this example,
-# the "template" is already the file we want.
-genrule(
```
alexeagle, line 61: this was just dead code, so this is a drive-by cleanup right?

alexeagle (approved):

examples/pip_parse/BUILD
f0rmiga: No, it was not dead code. But I don't think we should be testing the generated requirements feature in the main `pip_parse` example. This is why I set up a more comprehensive test elsewhere.

PR #920, opened 2022-12-06
fix: don't test on release

f0rmiga:
The tags that are released already have a more comprehensive test run attached to them, so we don't need to spend a lot of time re-testing on GitHub Actions during release.


rickeylev:
> The tags that are released already have a more comprehensive test run attached to them

What test run is this referring to? The release steps in the DEVELOPING.md doc doesn't mention anything about pre-testing steps?

f0rmiga:
We are only pushing tags to a commit on origin/main. The main CI already runs on all commits landed on main.

rickeylev (approved):

PR #933, opened 2022-12-19
Remove unused code

groodt:
Removes unused code. The code that is generating `requirements.bzl` is `parse_requirements_to_bzl.py`

Aside: I do think these generated helper functions in the repository rule could probably be replaced with simpler macros similar to the macro included in rules_jvm_external. See: https://github.com/bazelbuild/rules_jvm_external#artifact-helper-macro


hrfuller (approved):
LGTM.

Issue #845, opened 2022-10-01
py_wheel built with stamping doesn't embed version in METADATA file


pstradomski:
### Affected Rule

py_wheel

### Is this a regression?

No

### Description

//examples/wheel:minimal_with_py_library_with_stamp generates a wheel with the following metadata

```
Metadata-Version: 2.1
Name: example_minimal_library
Version: 0.1.{BUILD_TIMESTAMP}

UNKNOWN
```

There's a verabatim `{BUILD_TIMESTAMP}` there: the replacement was not applied correctly.

ns-ggeorgiev:
I am hitting the same bug as well. Any way to help with the fix?

PR #935, opened 2022-12-19
fix: embed stamped version in py_wheel METADATA

mattoberle:
## PR Checklist

## PR Type

- [x] Bugfix

Issue Number: #845 

Prior to `rules_python==0.14.0` the `dist-info/METADATA` file built by `py_wheel` resolved stamped data.
This can be confirmed by doing the following:

```sh
git checkout 0.13.0
git cherry-pick 921ad79
bazel test //examples/wheel/...
```

Where `921ad79` is the commit from this PR that adds the test for the resolved version.

Beginning with `0.14.0`, version is incorrectly set.

## What is the new behavior?

This PR restores the previous behavior by writing the `Version:` in `wheelmaker.py`.
One side effect is that the order of the `METADATA` file is modified.
`Version:` was previously after `Name:`. Now `Version:` is the final key before the "description".

> Note: I also had to run `bazel run //:vendor_requirements` from within `examples/pip_parse_vendored` to fix CI. Does not seem related to this PR, I can remove that change if it should not be included.

- [ ] Yes
- [x] No


tools/wheelmaker.py
```diff
@@ -167,10 +167,12 @@ def add_wheelfile(self):
...
         self.add_string(self.distinfo_path("WHEEL"), wheel_contents)
 
-    def add_metadata(self, metadata, description):
+    def add_metadata(self, metadata, description, version):
         """Write METADATA file to the distribution."""
         # https://www.python.org/dev/peps/pep-0566/
         # https://packaging.python.org/specifications/core-metadata/
+        metadata += "Version: " + version
+        metadata += "\n"
         metadata += "\n"
```
mattoberle, line 176:
```suggestion
        metadata += "\n\n"
```

Oops, if preferred...

f0rmiga (approved):

PR #953, opened 2022-12-29
Refactor and separate concerns of external python package handling code

groodt:
Refactors the code that deals with external python packages retrieved from repositories (i.e. PyPI, artifactory etc) into 3 main concerns:

1. wheel_installer: The intention is for this to be *only* Wheel unpacking, which would be a hermetic operation and could run as a build action. However, at the moment, there is also a pip interaction that happens here to build wheels. This can be refactored out into a wheel_builder in a future PR
2. lock_file_generator: This generates a `requirements.bzl` lock file for the external python packages from a fully resolved `requirements.txt` file (optionally generated from dependency_resolver. See 3)
3. dependency_resolver: Fully resolves a `requirements.in` file of direct dependencies into a `requirements.txt` file.

I think the refactor makes the code easier to work with and understand. This should make further improvements easier over time.


python/pip_install/tools/lock_file_generator/BUILD.bazel
```diff
@@ -0,0 +1,51 @@
+load("@rules_python//python:defs.bzl", "py_binary", "py_library", "py_test")
```
rickeylev, line 1: Should this be "@rules_python//" or just "//"?

python/pip_install/tools/wheel_installer/BUILD.bazel
```diff
@@ -0,0 +1,68 @@
...
+    deps = [":lib"],
+)
+
+py_test(
+    name = "namespace_pkgs_test",
+    size = "small",
+    srcs = [
+        "namespace_pkgs_test.py",
+    ],
+    tags = ["unit"],
```
rickeylev, line 33: Is this tags=unit used by something elsewhere (didn't see it)? I'd think that size=small was sufficient info to filter (small usually means unit test)

rickeylev (approved):
Just nits, really; otherwise LGTM

python/pip_install/tools/wheel_installer/BUILD.bazel
groodt: I think it's probably unused. I saw a few `test_tag_filters` to exclude integration test tags. I'll try remove it and see.

PR #959, opened 2023-01-02
chore(docs): fix bad html formatting

alexeagle:
Our docs are markdown with embedded html. Bare left-angle-brackets will be interpreted as HTML tags, so they should always be in code blocks


groodt (approved):
LGTM

PR #961, opened 2023-01-02
Support gazelle plugin when using bzlmod

aignas:
## PR Checklist

- [X] Tests for the changes have been added (for bug fixes / features)

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

`gazelle` does not work with rules_python when using bzlmod.

## What is the new behavior?

`gazelle` works with rules_python when using bzlmod.

- [ ] Yes
- [X] No

## Change summary

- Add gazelle target definition and fix bzlmod dependencies

  Since with bzlmod we have isolation of module workspace, we only need to add `gazelle` to the example and there is no need to setup the go toolchain, which is an improvement over non-bzlmod usage.

- Add modules_map generation based on the whl list

  Even though I am changing the contents of the `all_requirements` and `all_whl_requirements` in `bzlmod` usecases, I believe these variables where not previously usable under `bzlmod`, so it should be fine.

- Add gazelle_python_manifest generation to the example

  The manifest generation part did not need any changes.

- Allow configuring the pip dependency naming convention

  This also allows us to support the multi-platform Python builds as described in https://www.youtube.com/watch?v=Bjaiu8tZZhs. The alias targets may be defined under `//third_party/pip:$distribution_name$_pkg`, so using the newly added `gazelle` directives allows the alias targets to be used by gazelle.


aignas:
Will split into multiple PRs with separate tests

Issue #965, opened 2023-01-03
Support gazelle plugin when using bzlmod


aignas:
# 🚀 feature request

Creating this issue to encompass all of the PRs related to this feature, as the single research PR (#961) got a little bit too big for easy review.

### Relevant Rules

`gazelle` Python plugin.

### Description

Currently the `gazelle` plugin does not work if using `bzlmod` with `bazel >= 6.0.0`. This is due to:
* Gazelle and go toolchain is not present in `MODULE.bazel` of `rules_python`, which means that the plugin reference fails.
* The `all_whl_requirements` have labels, that do not work in `bzlmod` due to repository remapping. Hence, the `modules_mapping` rule fails unless we do something about it.
* Gazelle will generate dependency labels in the targets that do not work in `bzlmod` due to repository remapping.

### Describe the solution you'd like

As researched in #961, we can fix these three issues separately:
1. Adding gazelle and go as bzlmod deps.
1. Changing the contents of the `all_whl_requirements` and `all_requirements` based on wether `bzlmod` is used.
1. Have extra declaratives for changing the generated labels to work with bzlmod.
1. Have extra aliases in the `@pip` repo that work with both, `bzlmod` and classic gazelle. This might be similar to #814, however this is not a requirement to make gazelle with `bzlmod` work.

### Describe alternatives you've considered

Rejected alternatives:
* Having a flag in `gazelle_python.yaml` manifest, which would say that we are using `bzlmod`. The manifest should be the same irrespective of whether we are using `bzlmod` or not due to the same reason why we may want to have aliases in `@pip` repo -- because we want to enable a smooth transition to `bzlmod` as described in the [migration guide](https://docs.google.com/document/d/1JtXIVnXyFZ4bmbiBCr5gsTH4-opZAFf5DMMb-54kES0/edit)

aignas:
Some ideas that came from the comment in #970.

Regarding the metadata for gazelle to consume from the `requirements.bzl` file, I see that there are 2 pieces for gazelle to work:
- `modules_mapping` creates a mapping between import names and Python package names. It downloads the `whl` files to get this information, so it depends on the said `whl` targets.
- `gazelle_python.yaml` manifest generation depends on the mapping and a few extra variables, which help it map the Python package name back to the bazel label for the `pkg` target.

Thinking from outside-in, I'd like to use something like below in my BUILD.bazel files:
```starlark
# BUILD.bazel contents
[...]

PR #966, opened 2023-01-03
feat(pip_parse)!: use user friendly aliases to pip repos in requirements arrays

aignas:
## PR Checklist

- [X] Tests for the changes have been added (for bug fixes / features)

## PR Type

- [X] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Currently the labels in the `all_requirements` and `all_whl_requirements` are not useable, which is needed for gazelle bzlmod support.

Issue number: #965

## What is the new behavior?

We change the labels to point to the aliases that are present when `bzlmod` is used.

- [ ] Yes
- [X] No


examples/bzlmod/BUILD.bazel
```diff
@@ -14,7 +14,8 @@ py_library(
     name = "lib",
     srcs = ["__init__.py"],
     deps = [
-        requirement("pylint"),
+        # You can reference libraries using plain text form or the 'requirement' macro
+        "@pip//:pylint_pkg",
```
alexeagle, line 18: can it just be `@pip//pylint` ? That's the nicest syntax users would enjoy most.
 (we'd generate a `external/pip/pylint/BUILD.bazel` file with appropriate aliases)

alexeagle (commented):
What if gazelle was changed to use the `requirement` function?

That's how it works in Java:
https://github.com/bazelbuild/rules_jvm_external/blob/master/examples/bzlmod/java/src/com/github/rules_jvm_external/examples/bzlmod/BUILD#L7

groodt:
> What if gazelle was changed to use the requirement function?
> 
> That's how it works in Java:
https://github.com/bazelbuild/rules_jvm_external/blob/master/examples/bzlmod/java/src/com/github/rules_jvm_external/examples/bzlmod/BUILD#L7

I am working on a PR that extracts the external package artifact macros out of the repository. Which would make things a lot simpler for both bzlmod and general maintenance of the rules. The artifact macros in rules_jvm_external work quite differently to the present repository rules in rules_python.

examples/bzlmod/BUILD.bazel
aignas: I would love to use this and would be willing to implement this if this approach was agreed universally. @rickeylev, any opinions here?

aignas:
@groodt, should we wait for the #989 to be done before we think about this again?

I personally found that something like `@pip//pylint` would be much nicer than `artifact("pylint")` because I can easily query the external targets as there are fewer layers of abstraction here. Any opinions from others?

examples/bzlmod/BUILD.bazel
aignas: @alexeagle, what do you think about the inclusion of the user friendly aliases?

aignas:
This is getting stale, so I'll close this in favour of something else in the somewhat near future.

PR #971, opened 2023-01-03
Gazelle bzlmod poc v2

aignas:
Following #970, this is an alternative implementation of bzlmod and gazelle integration.
The code is not clean, but somewhat functional.

The main idea is that:
* `gazelle_python_manifest` macro is generated in `@pip//gazelle.bzl`.
* The user needs to worry only about that macro when generating gazelle python manifest.

Benefits:
* No need for extra directives in the build files. We just regenerate the manifest when switching to bzlmod and rerun gazelle.
* No need to specify `modules_mapping` rule as we call it for the user.
* Only the `name` argument to the `gazelle_python_manifest` is truly necessary, we can infer the rest.

Drawbacks:
* Whenever `gazelle.bzl` code is updated (e.g. codefixes), this will cause re-fetching of the pip wheels if not using `bzlmod`, because the code detects changes in the `@pip` repository.
* The fact that the user is using bzlmod is leaking into the manifest.


chrislovecnm:
I know this is WIP, but I am wondering if we should have this as a separate example and not modify the existing example.  I would love to have this as an example, but I know that lots of people are not using bzmod yet.

aignas:
@chrislovecnm, the existing example is working with and without bzlmod minus the differences in the gazelle_python.yaml manifest. Would you still like a separate example?

chrislovecnm:
@aignas I think having a separate example using bzlmod has value.  Many companies need to start using it, and we must show off using it.  Once we are at a point where bzlmod is more the standard, we can remove the old example.  I would even tell people to use the new example.  Thoughts?

Issue #43, opened 2017-11-20
Support coverage for `py_test`


hwright:
`bazel coverage` should generate a coverage report for the python tests it runs.  Ideally, it gathers the list of dependencies, and then computes coverage over that for a given test.  When multiple tests are run in the same Bazel invocation, their reports should be merged into a single one.

I'm dropping this here mostly as a python-specific version of bazelbuild/bazel#1118.  I know that the `py_test` rule is currently implemented natively in Bazel, and the implementation in `rules_python` just forwards there, but this is fundamentally a python issue, so this repo feels appropriate.

The general plan is to modify the python test runner within Bazel to use `coverage.py` to generate reports when running under coverage.  I suspect it is a journey fraught with peril.

Alternatively, if it is possible to invoke `coverage.py` in ways _other_ than modifying core Bazel, I'm open to suggestions.

mattmoor:
@damienmg assigning to you to redirect appropriately.

hwright:
fwiw, I've also been poking at this as documented here: https://groups.google.com/d/msg/bazel-discuss/LC97zBrMG84/cfVNdus-AwAJ

It's been slow going, as I'm pretty unfamiliar with Bazel internals, and haven't been able to get the ear of somebody who is.

brandjon:
Setting P4 because this isn't a blocker for 1.0 support.

JoshSullivan11:
Has this feature been implemented yet? If so, in what release version? It seems like coverage has been able to cover every other bit of our code repo except for all python files which is makes up most of our code base. Please let me know!

siddharthab:
Is there a way to make `bazel coverage //some_py_test` not fail even if it does not collect coverage information. Currently, it fails because `:lcov_merger` attribute is not defined in py_test rules.

This is kind of important because we want to be able to run `bazel coverage ...` on a monorepo.

sgammon:
how is it possible that coverage, python, and bazel don't work together, in the year of our lord 2020

thundergolfer:
IIUC this problem lies in `bazelbuild/bazel`, and we need to work to get https://github.com/bazelbuild/bazel/issues/10660 triaged there.

EnergySRE:
We at Camus Energy still want this. Python doesn't scale well without good test coverage, and coverage reports are valuable leverage.

[...]

PR #977, opened 2023-01-09
feat(coverage): Register coverage.py to hermetic toolchains

aignas:
## PR Checklist

- [X] Some examples have been modified to work with `bazel coverage`.
- [ ] Docs have been added / updated (for bug fixes / features).

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

The `coverage_tool` cannot be included via the toolchian.

Issue Number: #43

## What is the new behavior?

Summary:
- A script to update URLs and sha256 values used to download platform specific coverage.py wheels.
- Allow to specify `coverage_tool` attribute in `python_repository`.
- Register `coverage.py` when running `pip_install_dependencies` function.
- Ensure that `use_repo` includes coverage tool, so that `bzlmod` users can use it too.
- Have a boolean flag, which disables setting `coverage_tool` in the toolchain code.

Does not work/bad/ugly:
- Use coverage.py v6.5.0 because the latest has a file [`types.py` in the package directory, which imports from Python's stdlib `types`][1]. Somehow the Python interpreter is thinking that the `from types import FrameType` is referring to the currently interpreted file and everything breaks. I would have expected the package to use absolute imports and only attempt to import from `coverage.types` if we use `coverage.types` and not just a plain `types` import.
- The `multi_python_versions` example cannot show coverage for the more complex tests that are using `subprocess`. I am wondering if this is related to the fact that we are including `coverage.py` via the toolchain and not through [other mechanisms][2].
- The `__init__.py` files in the root of the WORKSPACE in `bzlmod` is breaking, when running under `bazel coverage //:test`. However, it started working when I renamed `__init__.py` to `lib.py`. I am suspecting that this has to do with the fact that the layer of indirection that `coverage` introduces could be something to do with that.
- The coverage collection script is written in bash so I removed support for the automatic registration of coverage toolchain for Windows for now. The issue tracking upstream support for coverage in Windows is [here][3].

Work towards #43.

[1]: https://github.com/nedbat/coveragepy/blob/master/coverage/types.py
[2]: https://bazel.build/configure/coverage
[3]: https://github.com/bazelbuild/bazel/issues/15835

- [ ] Yes
- [X] No

We may potentially break setups where coverage tooling is already included via
`COVERAGE_PYTHON` variable setting. In that case the users should use the flag to
disable coverage_tool inclusion in the toolchain. Setting
`register_coverage_tool = False` when registering toolchains would be sufficient.


aignas:
This is mostly building on the support added in https://github.com/bazelbuild/bazel/commit/9d0163002ca63f4cdbaff9420380d72e2a6e38b3

python/repositories.bzl
```diff
@@ -241,6 +242,7 @@ exports_files(["python", "{python_path}"])
 py_runtime(
     name = "py3_runtime",
     files = [":files"],
+    coverage_tool = {coverage_tool},
```
adam-azarchs, line 245: Our team's internal rules do something like
```starlark
config_setting(
    name = "coverage_enabled",
    values = {"collect_code_coverage": "true"},
)

py_runtime(
    name = "py3_runtime",
    coverage_tool = select({
        ":coverage_enabled": "{coverage_tool}",
        "//conditions:default": None,
    }),
    ...
)
```
so that bazel doesn't need to fetch the coverage repo in order to configure the toolchain except when running for code coverage, which is nice.

aignas: Fixed this, thanks for the comment!

python/private/coverage_deps.bzl
```diff
@@ -0,0 +1,105 @@
...
+
+def install_coverage_deps():
+    """Register the dependency for the coverage dep.
+    """
+    for name, url, sha256, platform in _coverage_deps:
+        maybe(
+            http_archive,
+            name = name,
+            build_file_content = """
+py_library(
```
adam-azarchs, line 94: I'm not sure this can be a `py_library`.  Doesn't that lead to a dependency when you add it as a dependency for the `python_runtime`?  Or is the python toolchain not actually a dependency of `py_library` targets (as opposed to `py_binary`)?

aignas: `coverage_tool` needs to be a single file from the toolchain point of view and that is why there is only a single `src` in there. I was following official docs from https://bazel.build/configure/coverage#python and since the manual tests with `bazel coverage` worked in some cases, I thought that was fine?

aignas: TIL, thanks for explaining. Should we have an extra line of docs somewhere?

f0rmiga:
This is awesome! I was going to work on it this week. I did a pass on your code so far and will do a more thorough review today.

python/private/coverage_deps.bzl
rickeylev: I would suggest not using py_library for now -- it implies a set of functionality that simply isn't respected (e.g. imports attribute won't be respected, anything in its PyInfo is ignored, etc).

It's currently possible to put a py_library there, but is a bit risky. This is because py_library doesn't, today, depend on the toolchain. However, I don't see how it can avoid that forever (in particular, I'm thinking of for when pyc precompilation is added).

I _do_ think a py_library-like-equivalent dependency is the correct conceptual type of dependency that should be there, though. Running with code coverage is, essentially, adding an extra dependency to the binary and wrapping the regular entry point in another entry point. Unfortunately, it's not immediately clear to me how to do make that work cleanly via (or with support from) the toolchain (I have a few ideas, but haven't thought them through).

To use the coverage settings of the toolchain, you need to:
1. Point coverage_tool to a single file. Only this file is consumed. Any additional files, runfiles, dependencies, etc aren't used.
2. Put any files coverage_tool needs in coverage_files, which will be added to the binary's runfiles.

adam-azarchs: I mean `filegroup` is [clearly documented](https://bazel.build/reference/be/general#filegroup.srcs) that `srcs` is in the files output for the rule (which we need) and `data` is in the `runfiles`.  I don't think we need to make it more complicated than that (it is also true that bazel will treat such a `filegroup` as an executable target, however we don't think we actually depend on that behavior).

python/private/coverage_deps.bzl
```diff
@@ -0,0 +1,123 @@
...
+    name = "coverage",
+    srcs = ["coverage/__main__.py"],
+    data = glob(["coverage/*", "coverage/**/*.py", "coverage/*.so"]),
+    exec_compatible_with = {compatible_with},
+    visibility = ["//visibility:public"],
```
rickeylev, line 117: Should this be public visibility? It's really only the toolchain that should be referencing it. It'd be rather hard to use it correctly because its just a group of platform-specific files.

python/repositories.bzl
```diff
@@ -238,9 +239,20 @@ cc_library(
 
 exports_files(["python", "{python_path}"])
 
+# Used to only download coverage toolchain when the coverage is collected by
+# bazel.
+config_setting(
+    name = "coverage_enabled",
+    values = {{"collect_code_coverage": "true"}},
+)
+
```
rickeylev, line 248: oh awesome, thank you for adding this!

Can you restrict visibility of this? It's not the py rule's responsibility to tell if coverage is enabled or not, so lets not expose a way for people to use it and think it is.

python/repositories.bzl
```diff
@@ -255,13 +267,15 @@ py_runtime_pair(
         python_path = python_bin,
         python_version = python_short_version,
         python_version_nodot = python_short_version.replace(".", ""),
+        coverage_tool = rctx.attr.coverage_tool if rctx.attr.coverage_tool == None and "windows" not in rctx.os.name else "\"{}\"".format(rctx.attr.coverage_tool),
```
rickeylev, line 270: I think this logic has a bug?

If tool=None and os=windows, then it'll do format(None), resulting in `"None"`, which isn't a valid value.

Is this supposed to be "or" instead of "and" ?

python/repositories.bzl
```diff
@@ -238,9 +239,20 @@ cc_library(
...
+    values = {{"collect_code_coverage": "true"}},
+)
+
 py_runtime(
     name = "py3_runtime",
     files = [":files"],
+    coverage_tool = select({{
+        ":coverage_enabled": {coverage_tool},
+        "//conditions:default": None,
+    }}),
```
rickeylev, line 255: Per my other comment, you'll also want to set coverage_files

python/private/coverage_deps.bzl
rickeylev: Er, sorry, I mixed up the py_runtime and PyRuntimeInfo apis. Disregard the "you have to specify two attributes" part. The splitting of the target into (entry point, files) is handled by py_runtime.

python/repositories.bzl
```diff
@@ -278,6 +292,25 @@ python_repository = repository_rule(
     _python_repository_impl,
     doc = "Fetches the external tools needed for the Python toolchain.",
     attrs = {
+        "coverage_tool": attr.label(
+            # Mirrors the definition at
+            # https://github.com/bazelbuild/bazel/blob/master/src/main/starlark/builtins_bzl/common/python/py_runtime_rule.bzl
+            allow_files = False,
+            doc = """
```
rickeylev, line 299: Lets also mention the py_runtime.coverage_tool docs and refer the reader there for more information. It'll be easy for these two attribute docs to get out of sync.

rickeylev (changes requested):

python/private/coverage_deps.bzl
adam-azarchs: > > I don't think we need to make it more complicated than that
> 
> I think you're referring to my "it should be a py library-like-equivalent dependency" comment?
> 
> There's a big distinction between "bundle of pre-built, platform-specific files" and a "dependency" (for lack of a better term). The former carries no information other than "here's a bag of files". A dependency can carry additional information, such as modifications to make to the import path, additional linker flags, propagating a _proper_ runfiles object (we currently omit symlinks, for example), or other necessary metadata.

Yes, and in this particular case, so long as the target has its visibility restricted to the toolchain, a bag of pre-built platform-specific files is exactly what this is - the `py_runtime` rule won't pay any attention to additional providers from the target, and if it did honor e.g. `imports` from this input, I can't see a situation where that wouldn't do more harm than good.  But, this is a very good reason to make sure the visibility is appropriately restricted.

> "depend" is a strong word, but we do allow the "executable" of the target to be an indicator as to what the entry point is. The basic logic for converting the Target to the (entry point, files) is:
> 
> * If there is one output, use that
> * otherwise, use whatever the executable file is
> * Otherwise, error

Yes, I do remember how I implemented that.  But in this case I was referring to the fact that a `filegroup` with a single file in `src` will have a `FilesToRun` provider supplying that file as an executable, which would make the second option you list above there work, but we don't depend on that behavior (which is, admittedly, a little weird) because it'll stop at the first in this case.

At any rate, if a `filegroup` is what we want to use here, then the upstream docs in the bazel repo should maybe be updated?  Except that there _might_ be use cases for having it as a `py_library` in the general case; just in this particular case we don't need it to be and by using restricted visibility and a `filegroup` we can afford to not think about all of the ways that could go wrong.

aignas: I changed it to use filegroup, so it should be hopefully an improvement over `py_library` and we can figure out something better as a separate PR.

python/private/coverage_deps.bzl
```diff
@@ -0,0 +1,123 @@
...
+    """
+    for name, url, sha256, platform in _coverage_deps:
+        maybe(
+            http_archive,
+            name = name,
+            build_file_content = """
+py_library(
+    name = "coverage",
+    srcs = ["coverage/__main__.py"],
+    data = glob(["coverage/*", "coverage/**/*.py", "coverage/*.so"]),
```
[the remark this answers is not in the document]
aignas: Done, please take a look again.

python/repositories.bzl
aignas: Thanks, there was a bug, fix it now.

python/repositories.bzl
aignas: Done.

python/private/coverage_deps.bzl
```diff
@@ -0,0 +1,165 @@
...
+load("@bazel_tools//tools/build_defs/repo:utils.bzl", "maybe")
+load(
+    "//python:versions.bzl",
+    "MINOR_MAPPING",
+    "PLATFORMS",
+)
+
+# Update with './tools/update_coverage_deps.py <version>'
+#START: managed by update_coverage_deps.py script
+_coverage_deps = {
```
rickeylev, line 14: Ah hm. Something I just realized is we run the risk of conflicting with a user installed coverage library. e.g.
the toolchain uses coverage X
The user, in their requirements, uses coverage Y

At runtime, we'll have 2 different coverage libraries available, and it's not well defined which will be used.

I'm not sure what we can do about this, though. I guess add an arg to toolchain creation and dependency setup that allows the user to specify the coverage target to use, and skip having one automatically installed? Or...IDK. The key attribute we're looking for is that the toolchain and user binary reference the same target (or equiv; that we only end up with 1 coverage library in runfiles).

python/tests/toolchains/run_acceptance_test.py.tmpl
```diff
@@ -53,6 +53,15 @@ class TestPythonVersion(unittest.TestCase):
...
         subprocess.run("bazel test //...", shell=True, check=True)
 
+        stream = os.popen("bazel info execution_root")
+        exec_root = pathlib.Path(stream.read().strip())
+        external = exec_root / "external"
+        self.assertEqual([], list(external.glob("pypi__coverage_*")), "coverage toolchain should not be downloaded upon bazel test")
+
+    @unittest.skipIf(%is_windows%, "not supported on Windows")
+    def test_coverage_toolchain(self):
+        subprocess.run("bazel coverage //...", shell=True, check=True)
```
rickeylev, line 63: Is there another way to test this? I was actually looking to remove this run_acceptance_test thing because it prevents upstream bazel from running our tests with upcoming bazel versions.

examples/bzlmod/.bazelrc
```diff
@@ -1 +1,3 @@
 common --experimental_enable_bzlmod
+
+coverage --java_runtime_version=remotejdk_11
```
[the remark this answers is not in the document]
adam-azarchs: My guess would be because they don't have java installed on their dev machine, or at least not in a way that bazel is able to find it, and including it in the PR was accidental.

python/private/coverage_deps.bzl
adam-azarchs: That would be an argument in favor of using `py_library` as the target type in the remote for `coverage_dep`, so things other than the toolchain could use it.

I don't think there's a way around having the coverage tool used by the toolchain being potentially different from what's declared in requirements, because you can't actually do a `pip install` until the toolchain is configured.  I don't think it's a dealbreaker, though; it might be a little surprising to find out that coverage is running with a version of `coveragepy` distinct from what was in your `requirements.txt`, but at the same time I don't think it'll actually _interefere_ with anything.

examples/bzlmod/.bazelrc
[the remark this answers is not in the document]
aignas: Should I remove it? I installed `bazel` via `bazelisk` and it does not pull in any `java` dependencies by default. What is the official way to do this?

python/tests/toolchains/run_acceptance_test.py.tmpl
aignas: I am not sure about that. I was using what was available, but if you have a suggestion on how to configure `bazelci` in order to run coverage, I would be happy to remove the changes in this file.

python/private/coverage_deps.bzl
rickeylev: For now, can you default `register_coverage_deps=False`? Otherwise we'll potentially cause issues with people who have coverage in their deps already, and we don't really have any solution or workaround available to them.

Yeah, we were kicking around ideas on how to best address this. I'm not sure there is much to be done for this in this PR. The only thing I can think of would be to change the  `register_coverage_tool` arg from a bool to something that allows the caller to pass in their own target.

> you can't actually do a pip install until the toolchain is configured. 

Hmmm, well, what if you _could_? If we had a separate "pip tools" or "requirements resolution" toolchain, then maybe we'd be able to do that. Just thinking out loud.

The two more appealing ideas we came up with were:
 * Making sure the toolchains' coverage was last in the import order (probably by modifying the bootstrap, when the `bootstrap_template` arg becomes available). That would at least give well-defined order for which version gets used and lets the user have more control over the version used.
 * Have the toolchain only have the coverarge tool entry point, not the coverage library itself; each test is responsible for bringing the necessary deps for the coverage tool (and try to figure out how to make that as easy/transparent to test writers as possible)

examples/bzlmod/.bazelrc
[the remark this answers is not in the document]
rickeylev: So according to this: https://bazel.build/docs/bazel-and-java#hermetic-testing

What this basically means is instead of using a locally installed JDK, it uses one it downloads (it sounds similar to how we're doing things).

So this is probably a net positive for getting started at the cost of having to download a JDK. I think that's worth it, so go ahead and leave it in.

python/tests/toolchains/run_acceptance_test.py.tmpl
```diff
@@ -53,6 +53,15 @@ class TestPythonVersion(unittest.TestCase):
 
         subprocess.run("bazel test //...", shell=True, check=True)
 
+        stream = os.popen("bazel info execution_root")
```
rickeylev, line 56: Please use the subprocess module instead. Right now, this is leaking an opened file description, which spams the logs with a warning.

python/tests/toolchains/run_acceptance_test.py.tmpl
rickeylev: I think we just have to add `coverage_targets: <targets>` to the presubmit.yml?

https://github.com/bazelbuild/continuous-integration/blob/master/buildkite/README.md#running-bazel-coverage

But eh, I'm ok with doing that in a separate PR.

rickeylev (changes requested):

python/tests/toolchains/run_acceptance_test.py.tmpl
aignas: Reverted modifications to this file and relying on the coverage_targets in the examples for now.

rickeylev (approved):

rickeylev:
btw, re: coverage of subprocesses not working: in the bootstrap, there's some code that unsets some coverage variables. I _think_ keeping that set is necessary to propagate coverage things to subprocesses

Issue #979, opened 2023-01-11
Include extra dependencies for `entry_point`


xinbinhuang:
# 🚀 feature request
related: https://github.com/bazelbuild/rules_python/issues/341

### Relevant Rules

`entry_point`

### Description

Currently, the `entry_point` is a pre-generated `py_binary` to expose the `console_scripts`. It works for tools that takes file as input and with limited reliance on module resolution e.g. static analysis tools, `flake8`, `black`. 

However, it doesn't quite work with things that expect dependencies on the import path, which is common for running dev server for web development frameworks, e.g `flask`, `uvicorn`, `strawberry`. In order for these use cases to work, we need a mechanism to properly propagate the import paths when executing the console scripts target.

### Describe the solution you'd like

One solution would be to convert `entry_point` into a macro that dynamically generate `py_binary` on the fly, so that it can take additional `deps` attribute to construct the console scripts `py_binary`.

Roughly can be implemented as below?

In the generated `@pip//:requirements.bzl`
```
load("@rules_python//python:defs.bzl", "py_binary")
load("@pip//:requirements.bzl", "requirement")

def new_entry_point(name, pkg, script = None, **kwargs):
    if not script:
        script = pkg
    entry_point_py = "@{repo_prefix}" + _clean_name(pkg) + "//:{entry_point_prefix}_" + script + ".py"

    py_binary(
        name = name,
        srcs = kwargs.pop("srcs", []) + [entry_point_py],
        main = entry_point_py,
        deps = kwargs.pop("deps", []) + [requirement(pkg)],
        **kwargs
    )
```

In `BUILD.bazel`
```

new_entry_porint(
    name = "strawberry",
    pkg = "strawberry-graphql",
    script = "strawberry",
    deps = [
        ":schema",
    ],
)
```

### Describe alternatives you've considered

use environment variables `PYTHONPATH` when calling `PYTHONPATH=... bazel run ... `. but it's hard and not trivial to figure out all import paths and add to it.

rafmagns-skepa-dreag:
linked PR author asked questions that were never answered on publicly github. would like to see them answered before this issue is closed

chrislovecnm:
@aignas can you take a look?

aignas:
[...]

PR #980, opened 2023-01-11
Allow `extry_point` to include extra targets

xinbinhuang:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: https://github.com/bazelbuild/rules_python/issues/979

```bazel
alias(
    name = "flake8",
    actual = entry_point(pkg="flake8", script="flake8")
)
```

## What is the new behavior?

```bazel
entry_point(
    name = "flake8",
    pkg = "flake8",
    script="flake8",
    deps = [...]        # any other kwargs applicable to py_binary
)
```

- [ ] Yes
- [x] No


examples/pip_parse/BUILD.bazel
```diff
@@ -55,6 +56,11 @@ alias(
     actual = entry_point("yamllint"),
 )
 
+entry_point_binary(
+    name = "yamllint_binary",
+    pkg = "yamllint",
+)
```
rickeylev, line 62: Since the basic feature is to allow additional dependencies, please provide an example where it does so. Otherwise it isn't clear what the difference is between these two, and its harder for readers to see code reflective of their case.

python/pip.bzl
```diff
@@ -235,6 +235,9 @@ def dist_info_requirement(name):
 def entry_point(pkg, script = None):
     fail("Not implemented yet")
 
+def entry_point_binary(name, pkg, script = None, **kwargs):
+    fail("Not implemented yet")
```
rickeylev, line 239: Instead of simply "not implemented yet", lets change the message to better indicate that codegen generates the implementation for this, so if you hit this, it means you're loading the wrong thing.

rickeylev (changes requested):

alexeagle:
It feels to me like the `entry_point` is syntax sugar for the simple case, and when you have additional dependencies you can easily write your own `py_binary` to run the tool with those dependencies (as you did for *every* package before `entry_point` was introduced). Also rules_js does a similar generation of convenience binaries for 3p packages, and doesn't have a feature like this.

So IMO this isn't needed and adds maintenance burden for rules_python.

xinbinhuang:
> It feels to me like the `entry_point` is syntax sugar for the simple case, and when you have additional dependencies you can easily write your own `py_binary` to run the tool with those dependencies (as you did for _every_ package before `entry_point` was introduced). 

You're totally right that it's a syntax sugar and user can build their own to involve additional dependencies. However, from a user perspective, it's not obvious at first glance on how to do that, especially for users not familiar with Bazel and `rules_python`. IMO, it seems a bit awkward to provide a half-functioned solution that indicate to the users that it'll work, but in fact with significant caveats. 

If we keep the current solution, at least we should clearly call out the limitation and list out the alternatives in the documentation so users don't get confused on this. However, I would argue that i'll be better time spent to just provide a better solution instead of writing those documentation.

> Also rules_js does a similar generation of convenience binaries for 3p packages, and doesn't have a feature like this.

I think `rules_js` already has this feature ([examples](https://github.com/aspect-build/rules_js/blob/2a603e7737fd57168f8e63ef3b845f8b78bb41b4/examples/webpack_cli/BUILD.bazel#L25-L39)). Or are you referring to something else?
 
> So IMO this isn't needed and adds maintenance burden for rules_python.

I think the end goal here is to only keep the new `entry_point_binary` and remove the old alias later. My original plan was to augmenet/modify the existing `entry_point`, but it seems a bit akward given how the current one is used (I'm happy to expand on what this means if you feel otherwise). So I introduced the `entry_point_binary` to avoid breaking changes and ease for user migration - similar to what is done for `pip_parse` and `pip_install`.

groodt:
The maintainers have discussed various plans to redesign how entry_points and the various generated functions are handled. We’re going to try move away from such a heavy reliance on generated repository rules. The functionality described in this PR would work with simple macros too (and would probably be more appropriate since it’s syntax sugar).

xinbinhuang:
Hey @groodt, thx for the context. One quick question

> PR would work with simple macros too

Do you mean by a macro provided by `rules_python` to extract entry point from the package?

xinbinhuang:
@groodt @alexeagle @rickeylev hey, just wanna check to see if we still want to have this PR merged in some way. Or if we want to take a totally different approach, that's ok too. Let me know how you think.

rafmagns-skepa-dreag:
PR author asked questions that were never answered on publicly github. would like to see them answered before this PR is auto-closed

rickeylev:
Over in #979 we're talking about and spec'ing out a new rule/macro to handle this case. There's a pending PR with most of the functionality implemented.

At this point, we're just finishing up some details and waiting for the PR author to come back from vacation. As such, I'll close this PR.

PR #985, opened 2023-01-12
runfiles: Fix usage instructions

fmeum:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Docs reference `bazel_tools` and import instructions fail with Bzlmod.

## What is the new behavior?

The canonical location for the runfiles library going forward will be the rules_python repo, so users should load from it. Since repository names are essentially dynamic with Bzlmod, they should not be used in import statements.

- [ ] Yes
- [x] No

### DIFF
--- a/python/runfiles/runfiles.py
+++ b/python/runfiles/runfiles.py
@@ -27,12 +27,12 @@ USAGE:
       py_binary(
           name = "my_binary",
           ...
-          deps = ["@bazel_tools//tools/python/runfiles"],
+          deps = ["@rules_python//python/runfiles"],
       )
 
 2.  Import the runfiles library.
 
-      from bazel_tools.tools.python.runfiles import runfiles
+      from python.runfiles import runfiles
 
 3.  Create a Runfiles object and use rlocation to look up runfile paths:
 


fmeum:
CC @alexeagle

python/runfiles/runfiles.py
```diff
@@ -27,12 +27,12 @@
...
           name = "my_binary",
           ...
-          deps = ["@bazel_tools//tools/python/runfiles"],
+          deps = ["@rules_python//python/runfiles"],
       )
 
 2.  Import the runfiles library.
 
-      from bazel_tools.tools.python.runfiles import runfiles
+      from python.runfiles import runfiles
```
rickeylev, line 35: Any idea which name is more correct/supported/defactor-correct? "import rules_python.python.runfiles" or "import python.runfiles" ?

I see a mix in our own tests.

Ah -- will "import python.runfiles" work better with bzlmod? Since the repo-name gets renamed to the versioned name.

otherwise LGTM.

alexeagle: It seems to me that we should suggest an import statement that will work when the package is installed from pypi. I would think something like `bazel_runfiles` would be the obvious form.

There's no reason this should match a bazel repository name, if I understand the way sys.path is constructed in the runtime shim for py_binary.

rickeylev: +1 to "bazel_runfiles", great name.

That said, it's a new name, so we can't exactly update these docs right now to say that name. We'll have to wait until either the pypi package is available and/or the source tree has been reorganized before recommending that name.

In the mean time, we should update these docs to at least drop the bazel_tools reference.

> There's no reason this should match a bazel repository name

100% agree

alexeagle: yeah let's land this one, I volunteered to take the pip package publishing task, and will update this again with a README.

alexeagle (approved):

PR #995, opened 2023-01-19
chore: publish a runfiles library as a wheel

alexeagle:
Wire it up to GH actions so it is published for each release.

Tested locally with:
```
$ bazel build python/runfiles:wheel --embed_label=1.0.2 --stamp
$ PYTHONPATH=bazel-bin/python/runfiles/bazel_runfiles-_BUILD_EMBED_LABEL_-py3-none-any.whl python
>>> import runfiles
>>> runfiles.Create()
```

Note, I would have liked to call the package bazel-runfiles, but this isn't possible without either refactoring the paths in this repo, or doing some fancy starlark to copy files around to create a folder that we turn into the wheel. There is no project https://pypi.org/project/runfiles though there is a https://pypi.org/project/runfile

We could try harder to get the name we prefer.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -13,7 +13,17 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
       - name: Prepare workspace snippet
-        run: .github/workflows/workspace_snippet.sh ${{ env.GITHUB_REF_NAME }} > release_notes.txt
+        run: .github/workflows/workspace_snippet.sh > release_notes.txt
+      - name: Build wheel dist
+        run: bazel build --stamp --embed_label=${{ env.GITHUB_REF_NAME }} //python/runfiles:wheel
+      - name: Publish runfiles package to PyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          # Note, the PYPI_API_TOKEN was added on
+          # https://github.com/bazelbuild/rules_python/settings/secrets/actions 
+          # and currently uses a token which authenticates as https://pypi.org/user/alexeagle/
+          password: ${{ secrets.PYPI_API_TOKEN }}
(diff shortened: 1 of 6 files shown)


alexeagle:
Looks like our buildifier presubmit check is broken, it complains about `MODULE.bazel` which I didn't edit in this PR.

```
##### :bazel: buildifier: found 1 format issue in your WORKSPACE, BUILD and *.bzl files
--
  | Please download <a href="https://github.com/bazelbuild/buildtools/releases/tag/6.0.0">buildifier 6.0.0</a> and run the following command in your workspace:<br/><pre><code>buildifier MODULE.bazel</code></pre>
```

fmeum:
It's broken because the new buildifier release formats `MODULE.bazel` files differently. You either need to pin to the old release or apply the fix.

alexeagle:
Yeah, I rebased over a commit I added on `main` that updated that formatting, and yeah I knew about your buildifier format fix. It feels like a bug to me that our presubmit.yml uses "latest" of buildifier but not a fight I want to pick today :)

f0rmiga (approved):

fmeum:
The logic in `CurrentRepository` makes some assumptions about the runfiles path of the file that contains it. We probably have to add tests for it when loaded from the wheel.

alexeagle:
Yeah I'll try adding the integration test before merging this.

rickeylev (approved):
overall LGTM.
+1 to having some sort of test; I'm ok with that being a separate change though, too

alexeagle:
@rickeylev looks like your personal email isn't properly connected with your GitHub identity in a way that Google's CLA bot understands - this is the result of applying your suggestions. You might need to add that email address in http://go/github ?

rickeylev:
CLA issue fixed

alexeagle:
Cool, thanks!

Issue #667, opened 2022-03-21
there is a vendored in version of skylib under third_party


aiuto:
That's just an anti-pattern. We should use the real one, or recreate the few needed features and not call it skylib.

alexeagle:
We did this because bzlmod isn't a thing yet, and didn't want to make a breaking change where we require users to install some rules_python dependencies because rules_python is so foundational and users easily get a skewed version of bazel_skylib if we depended on one.
I don't really agree it's an antipattern. I think core rulesets should just ship with zero dependencies until most users are on bzlmod.

johnynek:
The old "the old thing is deprecated and the new thing is not ready yet".

alexeagle:
Obviously the other problem is that skylib wasn't accepting contributions for many months, so like other users we had to fork.

PR #1003, opened 2023-01-21
cleanup: remove vendored copy of skylib

alexeagle:
Following #1001 we require that users install bazel-skylib, so we are now free to load from it.

Follow-up from #932


oliverlee:
resolves #667

f0rmiga (approved):

Issue #99, opened 2018-05-15
Provide a way to publish Python packages


calder:
Add a rule for building Python packages for distribution via PyPi. See https://github.com/horia141/bazel-pypi-package for an initial defunct attempt this.

rkhir:
Hello, is there an update on this issue? Or perhaps an alternative solution to publish Python packages?

alexeagle:
We now need this for rules_python itself, as we want to publish the runfiles library as a wheel to pypi.
I tried the simple thing of using `twine` via https://github.com/pypa/gh-action-pypi-publish but it failed our release https://github.com/bazelbuild/rules_python/actions/runs/4001340558/jobs/6867466794

In any event, I think this isn't a good solution because it's non-hermetic - you need python and `twine` and other stuff on the environment where you publish the wheel. I think we need a proper target that can be used with `bazel run` that does the publish, assuming nothing about the machine where it's run.

alexeagle:
0.18.2 should be released today which finally creates a `[name].publish` target on py_wheel.

PR #1014, opened 2023-01-25
refactor: py_wheel now a macro

alexeagle:
Pure refactoring that should have no user-visible effect.

This gives us the place to insert other targets produced by it, such as a publishing target.

Note: due to stardoc limitation, we have to expose two different symbols to the user in the docs.
I've linked them together and not tried to "hide" the fact that there's both a macro and a rule.
This way we don't have to duplicate all the docs for the attributes into the single docstring of the macro,
and declare all the matching parameters and remember to pass them through to the rule.

See docs preview: https://github.com/bazelbuild/rules_python/blob/bc38c7e28a5684fecd0d7dd38fffc00b763f6ad2/docs/packaging.md#py_wheel

Part of #99

### DIFF
--- a/docs/packaging.md
+++ b/docs/packaging.md
@@ -26,21 +26,84 @@ This rule is intended to be used as data dependency to py_wheel rule.
 | <a id="py_package-packages"></a>packages |  List of Python packages to include in the distribution. Sub-packages are automatically included.   | List of strings | optional | <code>[]</code> |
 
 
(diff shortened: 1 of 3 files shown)


rickeylev (approved):

f0rmiga (approved):
Looks like only macOS is stuck on CI but the rest is already green.

PR #1015, opened 2023-01-26
feat: wheel publishing

alexeagle:
Tested manually with:
```
$ cd examples/wheel
$ bazel run --stamp --embed_label=1.2.3 -- :customized.publish --repository testpypi
```
The result was published here: https://test.pypi.org/project/example-customized/1.2.3/

### DIFF
--- a/docs/packaging.md
+++ b/docs/packaging.md
@@ -121,7 +121,7 @@ Information about a wheel produced by `py_wheel`
 ## py_wheel
 
 <pre>
-py_wheel(<a href="#py_wheel-name">name</a>, <a href="#py_wheel-kwargs">kwargs</a>)
+py_wheel(<a href="#py_wheel-name">name</a>, <a href="#py_wheel-twine">twine</a>, <a href="#py_wheel-kwargs">kwargs</a>)
 </pre>
 
 Builds a Python Wheel.
@@ -168,6 +168,31 @@ py_wheel(
 )
 ```
 
+To publish the wheel to Pypi, the twine package is required.
+rules_python doesn't provide twine itself, see https://github.com/bazelbuild/rules_python/issues/1016
+However you can install it with pip_parse, just like we do in the WORKSPACE file in rules_python.
+
+Once you've installed twine, you can pass its label to the `twine` attribute of this macro,
+to get a "[name].publish" target.
+
+Example:
+
+```python
+py_wheel(
+    name = "my_wheel",
+    twine = "@publish_deps_twine//:pkg",
+    ...
+)
+```
+
+Now you can run a command like the following, which publishes to https://test.pypi.org/
+
+```sh
+% TWINE_USERNAME=__token__ TWINE_PASSWORD=pypi-*** \
+    bazel run --stamp --embed_label=1.2.4 -- \
+    //path/to:my_wheel.publish --repository testpypi
+```
+
 
 **PARAMETERS**
 
@@ -175,6 +200,7 @@ py_wheel(
 | Name  | Description | Default Value |
 | :------------- | :------------- | :------------- |
 | <a id="py_wheel-name"></a>name |  A unique name for this target.   |  none |
+| <a id="py_wheel-twine"></a>twine |  A label of the external location of the py_library target for twine   |  <code>None</code> |
 | <a id="py_wheel-kwargs"></a>kwargs |  other named parameters passed to the underlying [py_wheel rule](#py_wheel_rule)   |  none |
 
 
--- a/python/packaging.bzl
+++ b/python/packaging.bzl
@@ -68,7 +68,7 @@ This also has the advantage that stamping information is included in the wheel's
     },
 )
 
-def py_wheel(name, **kwargs):
+def py_wheel(name, twine = None, **kwargs):
     """Builds a Python Wheel.
 
     Wheels are Python distribution format defined in https://www.python.org/dev/peps/pep-0427/.
@@ -113,16 +113,63 @@ def py_wheel(name, **kwargs):
     )
     ```
 
+    To publish the wheel to Pypi, the twine package is required.
+    rules_python doesn't provide twine itself, see https://github.com/bazelbuild/rules_python/issues/1016
+    However you can install it with pip_parse, just like we do in the WORKSPACE file in rules_python.
+
+    Once you've installed twine, you can pass its label to the `twine` attribute of this macro,
+    to get a "[name].publish" target.
+
+    Example:
+
+    ```python
+    py_wheel(
+        name = "my_wheel",
+        twine = "@publish_deps_twine//:pkg",
+        ...
+    )
+    ```
+
+    Now you can run a command like the following, which publishes to https://test.pypi.org/
+
+    ```sh
+    % TWINE_USERNAME=__token__ TWINE_PASSWORD=pypi-*** \\
+        bazel run --stamp --embed_label=1.2.4 -- \\
+        //path/to:my_wheel.publish --repository testpypi
+    ```
+
     Args:
         name:  A unique name for this target.
+        twine: A label of the external location of the py_library target for twine
         **kwargs: other named parameters passed to the underlying [py_wheel rule](#py_wheel_rule)
     """
+    _dist_target = "{}.dist".format(name)
     py_wheel_dist(
-        name = "{}.dist".format(name),
+        name = _dist_target,
         wheel = name,
         out = kwargs.pop("dist_folder", "{}_dist".format(name)),
     )
 
     _py_wheel(name = name, **kwargs)
+
+    if twine:
+        if not twine.endswith(":pkg"):
+            fail("twine label should look like @my_twine_repo//:pkg")
+        twine_main = twine.replace(":pkg", ":rules_python_wheel_entry_point_twine.py")
+
+        # TODO: use py_binary from //python:defs.bzl after our stardoc setup is less brittle
+        # buildifier: disable=native-py
+        native.py_binary(
+            name = "{}.publish".format(name),
+            srcs = [twine_main],
+            args = [
+                "upload",
+                "$(rootpath :{})/*".format(_dist_target),
+            ],
+            data = [_dist_target],
+            imports = ["."],
+            main = twine_main,
+            deps = [twine],
+        )
 
 py_wheel_rule = _py_wheel
--- a/python/runfiles/BUILD.bazel
+++ b/python/runfiles/BUILD.bazel
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-load("//python:defs.bzl", "py_binary", "py_library")
+load("//python:defs.bzl", "py_library")
 load("//python:packaging.bzl", "py_wheel")
 
 filegroup(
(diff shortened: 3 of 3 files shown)


python/private/py_twine.bzl
```diff
@@ -0,0 +1,87 @@
+"""Implementation of py_twine rule.
```
f0rmiga, line 1: Missing license header.

python/private/py_twine.bzl
```diff
@@ -0,0 +1,106 @@
...
+# to avoid a dependency from bazelbuild -> aspect-build
+def _to_manifest_path(ctx, file):
+    """The runfiles manifest entry path for a file.
+
+    This is the full runfiles path of a file including its workspace name as
+    the first segment. We refert to it as the manifest path as it is the path
+    flavor that is used for in the runfiles MANIFEST file.
+    We must avoid using non-normalized paths (workspace/../other_workspace/path)
+    in order to locate entries by their key.
+    Args:
```
f0rmiga, line 63: I'd avoid Bash for a few reasons.

1. It's not super portable, especially on Windows.
2. You end up using more subprocesses that are not Bash anymore.
3. The rlocation is a mess of code that could eventually break, and we keep copy-pasting it.

I'd suggest the following:

1. Rewrite it in Python (you can inherit the hermetic interpreter).
2. Ditch the rlocation for the expand_location in Starlark.

f0rmiga (changes requested):

python/private/py_twine.bzl
alexeagle: For 1 this is true, I'd need a cmd.exe equivalent, we have that in lots of places.
For 2, I don't think a bash parent and python subprocess is any worse than two python processes. It might make a difference if we wanted to customize the twine CLI, but I don't want that.
For 3, I think it's pretty reliable, we could use the aspect_bazel_lib utils.bzl which exports that constant, but sadly I think we'll get in trouble using that from bazelbuild

Really what I want is not to have a parent process at all. It's a tiny annoying shim. How important do you think it is? I'm kinda out of time to work on this.

`expand_location` can't do the right thing if there's no runfiles tree, that's why the runfiles library needs to be used within the action. I think if it were a python parent process it would have to use the very runfiles library we are publishing here :)

f0rmiga: It would be funny to use the runfiles library to publish itself. :)

If you don't want to have parent processes at all, Python is the way to go because you can load twine and execute the code directly. Also, you don't need two implementations for multiple platforms, which adds maintenance complexity. I wonder how other maintainers feel about this.

Also, something that I forgot to suggest earlier is that you don't need to copy the wheel, you can symlink it to avoid duplication.

alexeagle:
The CI failure also indicates a design problem, how to reference the twine tool we've fetched for ourselves without users trying to reference a non-existent label.

python/private/py_twine.bzl
```diff
@@ -0,0 +1,106 @@
...
+load(":py_wheel.bzl", "PyWheelInfo")
+
+_attrs = {
+    "twine_bin": attr.label(
+        doc = """\
+A py_binary that runs the twine tool.
+The default value assumes you have `twine` listed in your own requirements.txt, and have run
+`pip_parse` with the default name of `pypi`.
+
+If these don't apply, you might use the `entry_point` helper to supply your own twine binary:
```
rickeylev, line 16: This all seems rather brittle or convoluted...

Is this something people could instead specify during their WORKSPACE setup? e.g.
`rules_python_setup(twine="//my:twine")`? And if unspecified, we install or otherwise assume some default?

Is this something bzlmod could handle better than workspace? If so, maybe we should make this bzlmod-only?

This can also be where toolchains might help -- have a twine toolchain, then use the usual toolchain registration/setup to figure out how a user provides the twine executable.

python/private/py_twine.bzl
```diff
@@ -0,0 +1,106 @@
...
+    if file.short_path.startswith("../"):
+        return file.short_path[3:]
+    else:
+        return ctx.workspace_name + "/" + file.short_path
+
+_exec_tmpl = """\
+#!/usr/bin/env bash
+{rlocation}
+tmp=$(mktemp -d)
+# The namefile is just a file with one line, containing the real filename for the wheel.
```
rickeylev, line 79: Why cant the produced file have this name?

python/private/py_twine.bzl
```diff
@@ -0,0 +1,106 @@
...
+wheel_filename=$tmp/$(cat "$(rlocation {wheel_namefile})")
+cp $(rlocation {wheel}) $wheel_filename
+$(rlocation {twine_bin}) upload $wheel_filename "$@"
```
rickeylev, line 82: I think the plain `File.path` might allow you to avoid using rlocation? i.e.

bash_script = """
{twine} upload {wheel}
""".format(twine.path, wheel.path)

I'm pretty sure the CWD is well-defined to be the repo root.

python/private/py_twine.bzl
rickeylev: I think what you want is to use `args`, so something like:

```
def py_package(name, ..):
  py_wheel(name=name, ...)
  run_twine(
    name=name + ".publish",
    args=['upload', "$(location :{wheel})".format(name)]
    data = [name]
  )

run_twine = rule(
  attrs={"_twine": attr.label(executable=True)}
)
def run_twine_impl(...): return DefaultInfo(executable=...)
```

Unfortunately, there isn't a good way to "embed" the args into the output executable in a particularly easy way.

python/private/py_twine.bzl
rickeylev: Another comment/question -- the requirements.txt this is referring to is the same one used for e.g. my py_binary's, right? Must my py_binary and twine share the same requirements.txt settings? I'm hoping that the particular build/version of twine is independent of _what_ it is publishing? Coupling the two seems like it would cause friction if twine's transitive deps aren't compatible with your py_binary's transitive deps.

python/private/py_twine.bzl
alexeagle: Because the analysis phase wants to predict the output files and the status files can only be read during action execution.

python/private/py_twine.bzl
alexeagle: Yeah that's the problem, Bazel doesn't capture args or env for programs which aren't run as actions.

rickeylev: The run command (and test command) respects `args` and `env` of the target it's given. Running as an action won't respect those. Passing one target to another won't automatically respect the inner's args/env.

python/private/py_twine.bzl
rickeylev: What part of the file name isn't knowable ahead of time? We're producing a final name such as: `runfiles-0.1.2-py39-none-any.whl` right? The first part is defined, the last 3 are part of configuration info available to the rule implementation. The 2nd part is variable, but is something that the CI knows (because it gets it from the git tag, right?) and can pass along on the command line.

python/private/py_twine.bzl
alexeagle: That assumes the user has runfile symlinks enabled though, which is why the rlocation function exists.

python/private/py_twine.bzl
```diff
@@ -0,0 +1,106 @@
...
+Simply wraps the tool with a bash script and a runfiles manifest.
+See https://twine.readthedocs.io/
+"""
+
+load(":py_wheel.bzl", "PyWheelInfo")
+
+_attrs = {
+    "twine_bin": attr.label(
+        doc = """\
+A py_binary that runs the twine tool.
```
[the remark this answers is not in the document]
alexeagle: It's a py_binary in every case I can think of - that's what `entry_point` helper returns. It's true that we don't care what type of rule provides an executable here, but it's nice to help users understand that `py_binary` is likely what they'd use to provide one.

python/private/py_twine.bzl
alexeagle: https://github.com/bazelbuild/rules_python/issues/741

alexeagle: "pass along on the command line" - I'm not sure what you mean by this, we have an action under `bazel build` that outputs the file

python/private/py_twine.bzl
alexeagle: Yes that's right, I had it backwards. I'll play with this and see if we can avoid wrapping the twine binary

python/private/py_twine.bzl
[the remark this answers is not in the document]
groodt: I would avoid using the generated entry_point if we can please. Thoughts of building our own tool that wraps twine.

WORKSPACE
```diff
@@ -67,3 +67,21 @@ load("@rules_python_gazelle_plugin//:deps.bzl", _py_gazelle_deps = "gazelle_deps
...
+
+load("@python//3.11.1:defs.bzl", "interpreter")
+
+#####################
+# Install twine for our own runfiles wheel publishing
+# Eventually we might want to install twine automatically for users too, see:
+# See https://github.com/bazelbuild/rules_python/issues/1016
+load("@rules_python//python:pip.bzl", "pip_parse")
+
+pip_parse(
```
groodt, line 79: Why are we unpacking twine differently to the rest of our tools in the repo that use external libraries? Feels inconsistent. I think all of the rules internal tools should load dependencies in the same private dependency closure.

python/private/py_twine.bzl
groodt: Should the rule even mention twine? Or is it a py_pypi_publish rule? I’m not even sure why it needs to be exposed as a rule. This could be part of py_wheel and implemented in python as mentioned elsewhere I think.

python/private/py_twine.bzl
alexeagle: The user already has stamping setup for their Bazel workspace, when they say `--stamp` and maybe `--workspace_status_command=./some/script.sh` that produces values in the `bazel-out/stable-status.txt` and `bazel-out/volatile-status.txt` files. That's the API contract for stamping. I think it's not ideal for us to say "by the way, if some of your artifacts are wheel files, then stamping isn't enough and you need to change how you run Bazel in your CI/CD pipeline"

WORKSPACE
alexeagle: We don't use any tools with pip_parse yet - we exploded their dependency graphs into our `python/pip_install/dependencies.bzl` file.

groodt: Precisely, why don't we do that here too?

python/private/py_twine.bzl
rickeylev: Well, I don't agree, because stamping is metadata about the whole build, not specific artifacts. But lets take the convo to the issue instead of the PR, where stamping and wheels are the dedicated topic, as it doesn't need to block this PR.

python/private/py_twine.bzl
rickeylev: Ah. Is that just on windows?

WORKSPACE
alexeagle: It has a really long list of transitives, see the `requirements.txt` introduced in this PR. I don't think we want to splat that into our repo for our own usage any more than a user would want to do that.

python/private/py_twine.bzl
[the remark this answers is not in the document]
alexeagle: Sorry, I don't understand your proposal. Do you just want to recommend that users spell out the longer label which provides the twine py_binary? Or do you want to write a bazel-specific publishing tool? If the latter, I don't understand the motivation there. Bazel should be able to run existing tools.

WORKSPACE
groodt: The length of the transitives isn't really an issue tbh. It's easy to automate or generate that. Being aware of what they are also yields strong benefits so we ensure they are compatible on the same PYTHONPATH and also are pure python so we can use them more easily.

I feel that we should handle the "rule-private" dependencies of our tools consistently.

alexeagle:
FYI I'm out for a week and a half, not sure when I'll have time again to pick this up

alexeagle:
Hi @groodt @rickeylev @f0rmiga I've updated this PR with a much simpler, less ambitious proposal. It could still be a stepping-stone to a nicer way if we resolved #1016 

For now I just want to call wheel publishing "possible" for users.

WORKSPACE
alexeagle: I could imagine a helper like `py_twine_repositories` in repositories.bzl that does the 'pip_parse`, making this easier for users. I split that out to #1016 for now.

f0rmiga (approved):

python/private/py_twine.bzl
alexeagle: Yes that's right, users can install twine with a separate pip_parse rule.

PR #1020, opened 2023-01-26
fix(release): use correct GH workflows syntax

alexeagle:
See https://docs.github.com/en/actions/learn-github-actions/contexts#github-context


alexeagle:
Still working to unblock the 0.17 release - failure was https://github.com/bazelbuild/rules_python/actions/runs/4020009800/jobs/6907480131

rickeylev (approved):

f0rmiga (approved):

PR #1021, opened 2023-01-26
fix(release): minimum needed to run twine to publish

alexeagle:
Earlier attempts to publish 0.17 have [failed mysteriously](https://github.com/bazelbuild/rules_python/actions/runs/4020082465/jobs/6907623158) at the `pypa/gh-action-pypi-publish` step. This PR replaces that with a command that I can test locally:

```
TWINE_USERNAME=__token__ TWINE_PASSWORD=pypi-*** bazel run --stamp --embed_label=1.2.4 //python/runfiles:wheel.publish -- --repository testpypi

Uploading bazel_runfiles-1.2.4-py3-none-any.whl
100% ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 11.0/11.0 kB • 00:00 • 27.3 MB/s

View at:
https://test.pypi.org/project/bazel-runfiles/1.2.4/
```

This is a (hopefully) less contentious subset of #1015 where we just run a binary ourselves but don't try to make a public API out of it yet.


python/runfiles/BUILD.bazel
```diff
@@ -40,12 +48,27 @@ py_wheel(
         "Development Status :: 5 - Production/Stable",
         "License :: OSI Approved :: Apache Software License",
     ],
-    description_file = "README.md",
+    description_file = "README.rst",
```
alexeagle, line 51: note, this change is actually required, otherwise pypi rejects the package. There's some way to configure the package metadata to allow markdown instead, but it seems easier to just use their defaults.

.github/workflows/release.yml
```diff
@@ -28,6 +28,15 @@ jobs:
         uses: actions/checkout@v2
       - name: Create release archive and notes
         run: .github/workflows/create_archive_and_notes.sh
+      - name: Publish wheel dist
+        env:
+          # This special value tells pypi that the user identity is supplied within the token
+          TWINE_USERNAME: __token__
+          # Note, the PYPI_API_TOKEN was added on
+          # https://github.com/bazelbuild/rules_python/settings/secrets/actions 
+          # and currently uses a token which authenticates as https://pypi.org/user/alexeagle/
```
f0rmiga, line 37: What's the long-term plan here to avoid using personal tokens?

WORKSPACE
```diff
@@ -67,3 +67,21 @@ load("@rules_python_gazelle_plugin//:deps.bzl", _py_gazelle_deps = "gazelle_deps
 # This rule loads and compiles various go dependencies that running gazelle
 # for python requirements.
 _py_gazelle_deps()
+
+#####################
+# Install twine for our own runfiles wheel publishing
+# Eventually we might want to install twine automatically for users too, see:
+# See https://github.com/bazelbuild/rules_python/issues/1016
```
f0rmiga, line 74:
```suggestion
# Install twine for our own runfiles wheel publishing.
# Eventually we might want to install twine automatically for users too, see:
# https://github.com/bazelbuild/rules_python/issues/1016.
```

python/runfiles/BUILD.bazel
```diff
@@ -40,12 +40,27 @@ py_wheel(
...
     deps = [":runfiles"],
 )
+
+# TODO(alexeagle): carry forward #1015 to make this part of the py_wheel macro
+py_binary(
+    name = "wheel.publish",
+    srcs = ["@publish_deps_twine//:rules_python_wheel_entry_point_twine.py"],
+    args = [
+        "upload",
+        "$(location :wheel.dist)/*",
```
f0rmiga, line 60:
```suggestion
        "$(execpath :wheel.dist)/*",
```

tools/publish/BUILD.bazel
```diff
@@ -0,0 +1,8 @@
+load("//python:pip.bzl", "compile_pip_requirements")
+
+compile_pip_requirements(
+    name = "requirements",
+    requirements_windows = "requirements_windows.txt",
+    # This fails on RBE and MacOS right now, and we don't need coverage there
+    tags = ["manual"],
```
f0rmiga, line 7: Could you remove this? I landed a fix here: https://github.com/bazelbuild/rules_python/commit/e7b51d7e557a2a5a244c58a81f77b5ab090f4973.

f0rmiga (changes requested):

.github/workflows/release.yml
alexeagle: oh, Richard already fixed this to use a Google-owned token. I'll update the comment

tools/publish/BUILD.bazel
alexeagle: it will still fail, I'll comment out this line so you see the failure

f0rmiga (approved):

rickeylev (approved):

f0rmiga (approved):

PR #1022, opened 2023-01-27
chore: revert publish wheel on releases

alexeagle:
our 0.17 release is stuck on this step.
Partially reverts #995

We can try again after #1021 lands


groodt (approved):
LGTM

PR #1023, opened 2023-01-27
Using label as id for py targets

linzhp:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Gazelle plugin generates a UUID for each `py_library` target for `py_binary` and `py_test` in the same package to refer to. However, creating another identifier such as UUID is not necessary because Bazel already has id for each target, which is its label.

Furthermore, in some large monorepos where it's too slow for Gazelle to index the whole repo, we turn off the indexing with `bazel run //:gazelle -- -index=false` and rely on [CrossResolver](https://github.com/bazelbuild/bazel-gazelle/blob/77e0f34c43a9f5e95f6af93485a4bd83942ed3c6/resolve/index.go#L66) to resolve dependencies based on paths conventions. The UUID makes it impossible to resolve dependencies without indexing the whole repository.

## What is the new behavior?
Reuse Bazel label as the id for the library and avoid depending on the UUID library. Now CrossResolvers can easily resolve same-package dependencies

- [ ] Yes
- [x] No

## Other information
If this approach is acceptable, I can create a follow up PR to rename variables to reflect the fact that we are no longer using UUID

### DIFF
--- a/gazelle/python/generate.go
+++ b/gazelle/python/generate.go
@@ -210,7 +210,7 @@ func (py *Python) GenerateRules(args language.GenerateArgs) language.GenerateRes
 		}
 
 		pyLibrary = newTargetBuilder(pyLibraryKind, pyLibraryTargetName, pythonProjectRoot, args.Rel).
-			setUUID(fmt.Sprintf("//%s:%s", args.Rel, pyLibraryTargetName)).
+			setUUID(label.New("", args.Rel, pyLibraryTargetName).String()).
 			addVisibility(visibility).
 			addSrcs(pyLibraryFilenames).
 			addModuleDependencies(deps).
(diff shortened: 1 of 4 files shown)


gazelle/python/generate.go
```diff
@@ -212,7 +210,7 @@ func (py *Python) GenerateRules(args language.GenerateArgs) language.GenerateRes
 		}
 
 		pyLibrary = newTargetBuilder(pyLibraryKind, pyLibraryTargetName, pythonProjectRoot, args.Rel).
-			setUUID(uuid.Must(uuid.NewUUID()).String()).
+			setUUID(fmt.Sprintf("//%s:%s", args.Rel, pyLibraryTargetName)).
```
f0rmiga, line 213: Could you use the Label type for this, please?

f0rmiga:
Does go mod tidy prune the uuid dependency?

f0rmiga (approved):
Thanks for addressing my comments.

Issue #1031, opened 2023-01-30
0.15.0 Sha256 Sum Changed


loganasherjones:
# 🐞 bug report

### Affected Rule

N/A

### Is this a regression?

N/A

### Description

Recently, the `0.15.0` version of `rules_python` seems to have a different `sha256sum`. According to the [release page for 0.15.0](https://github.com/bazelbuild/rules_python/releases/tag/0.15.0) the `sha256` should be `fda23c37fbacf7579f94d5e8f342d3a831140e9471b770782e83846117dd6596` but I'm currently getting `99e0d77f6ca445134adb5ef3d321a065d3b55b47a834cca110c30bf519d919bb`.

## 🔬 Minimal Reproduction

Create a `WORKSPACE` using the instructions from the 0.15.0 release page:

```
load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
http_archive(
    name = "bazel_skylib",
    sha256 = "74d544d96f4a5bb630d465ca8bbcfe231e3594e5aae57e1edbf17a6eb3ca2506",
    urls = [
        "https://mirror.bazel.build/github.com/bazelbuild/bazel-skylib/releases/download/1.3.0/bazel-skylib-1.3.0.tar.gz",
        "https://github.com/bazelbuild/bazel-skylib/releases/download/1.3.0/bazel-skylib-1.3.0.tar.gz",
    ],
)
http_archive(
    name = "rules_python",
    sha256 = "fda23c37fbacf7579f94d5e8f342d3a831140e9471b770782e83846117dd6596",
    strip_prefix = "rules_python-0.15.0",
    url = "https://github.com/bazelbuild/rules_python/archive/refs/tags/0.15.0.tar.gz",
)
```

`main.py`

```python
print("Hello, World")
```

`BUILD`:

```
load("@rules_python//python:defs.bzl", "py_binary")

py_binary(
  name = "main",
  srcs = ["main.py"],
)
```

```
bazel run //:main
```

Will fail with a checksum issue.

## 🔥 Exception or Error

<pre><code>
ERROR: An error occurred during the fetch of repository 'rules_python':
   Traceback (most recent call last):
	File "/home/.cache/bazel/external/bazel_tools/tools/build_defs/repo/http.bzl", line 125, column 45, in _http_archive_impl
		download_info = ctx.download_and_extract(
[...]

PR #1032, opened 2023-01-30
Manually generate distribution archives

rickeylev:
## PR Checklist

## PR Type

- [x] Bugfix

Right now, we use the automatically generated GitHub source archives. These have had issues with GitHub changing how they're generated, which affects the checksums, which then breaks users.

Issue Number: #1031

## What is the new behavior?

A tar.gz is manually generated using `git export`. Some online sources indicate this is the underlying mechanism github uses for generate their archives. We also use this for computing the checksum, so that comports.

- [x] No

Though note that base filename changes

## Other information

This happened previously a year to so ago, too.


.github/workflows/create_archive_and_notes.sh
```diff
@@ -0,0 +1,91 @@
...
+
+
+set -o errexit -o nounset -o pipefail
+
+# Set by GH actions, see
+# https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables
+TAG=${GITHUB_REF_NAME}
+PREFIX="rules_python-${TAG}"
+ARCHIVE="rules_python-$TAG.tar.gz"
+git archive --format=tar --prefix=${PREFIX}/ ${TAG} | gzip > $ARCHIVE
```
linzhp, line 24: Adding a prefix here and asking users to add `strip_prefix` in `http_archive` seem unnecessary?

rickeylev: I'm guessing it was done that way to match the github-generated archives? I'm just leaving it as is for now so that it's less config people need to change.

linzhp: Because of today's incident, there is no point of matching the sha256 of github-generated archives. FYI: rules_go and Gazelle's release archives are independent from github-generated ones too:
https://github.com/bazelbuild/rules_go/releases/tag/v0.38.1
https://github.com/bazelbuild/bazel-gazelle/releases/tag/v0.29.0

rickeylev: Yes, I know. I'm not trying to match the sha256. I'm just keeping the structure the same, because it's a nicety to not have to change both url and the strip_prefix setting.

groodt (approved):
LGTM

.github/workflows/create_archive_and_notes.sh
alexeagle: +1 for keeping it the same

PR #1029, opened 2023-01-28
Resolving sibling modules with absolute imports

linzhp:
## PR Checklist

Please check if your PR fulfills the following requirements:

- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type

What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature (please, look at the "Scope of the project" section in the README.md file)
- [ ] Code style update (formatting, local variables)
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] Other... Please describe:

## What is the current behavior?

`py_test` and `py_binary` unconditionally depends on `py_library` in the same package. In order to do that, the Gazelle plugin introduce an virtual import to identify the `py_library` target.

## What is the new behavior?
With the capability of resolving sibling modules introduced in #1027, we can resolve `py_library` as needed in `py_test` and `py_binary`, and we no longer need any virtual imports or id for `py_library` targets. This simplifies the logic.

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

PR #1035, opened 2023-02-01
Revert "Resolving sibling modules with absolute imports"

f0rmiga:
Reverts bazelbuild/rules_python#1029

This broke Gazelle on a project. I'm reverting this for now while figuring out why. I'll also add a test case to cover the failure I got.


f0rmiga:
cc @linzhp

mattem (approved):

linzhp:
OK. Keep me posted

PR #1040, opened 2023-02-02
Pass cpp flags through in all cases to repository rule compilations

shs96c:
## PR Checklist

## PR Type

Bugfix

When using a layer of indirection to refer to a hermetic toolchain, python deps that compile local code fail to properly set the path to the local XCode install, causing compilation to fail (eg. when compiling `lxml`'s deps)

## What is the new behavior?

The CPP_FLAGS are set properly, and `lxml` will install from a repository rule cleanly

No

## Other information

I'm not entirely sure how to reproduce this with a test in this repo. It only showed up when we had some particularly gnarly setup with a repo at work.

### DIFF
--- a/python/pip_install/pip_repository.bzl
+++ b/python/pip_install/pip_repository.bzl
@@ -79,10 +79,6 @@ def _get_xcode_location_cflags(rctx):
     if not rctx.os.name.lower().startswith("mac os"):
         return []
 
-    # Only update the location when using a hermetic toolchain.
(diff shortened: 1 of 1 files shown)


f0rmiga:
I'm not particularly concerned with reproducing and testing the bug you faced here, as we have plans to move away from building wheels in repository rules. If you think a bit about it, this is a poor's man toolchain resolution to get flags, etc. It just doesn't play well with what Bazel expects overall.

f0rmiga (approved):

PR #1041, opened 2023-02-02
Collect all py file names in the same package

linzhp:
@f0rmiga I think this is a better fix for the bug you ran into. It would prevent similar bugs from happening when importing other Python modules (e.g., `__init__.py` etc) in the same package.


f0rmiga (approved):

f0rmiga:
Awesome, thanks!

PR #1046, opened 2023-02-07
fix: make py_proto_library respect PyInfo imports

aignas:
## PR Checklist

## PR Type

- [X] Bugfix
- [ ] Build related changes
- [ ] CI related changes
- [X] Documentation content changes

Currently the user needs to manually include the protobuf client runtime (e.g. via PyPI)   dependencies when consuming `py_proto_library` even though `py_proto_library` implicitly depends on them and they appear in runfiles.

## What is the new behavior?

`py_proto_library` now properly propagates the `imports` attribute the protobuf client runtime so that the libraries are added to `sys.path` correct.

Also adds an example for `bzlmod` and old way of using the `py_proto_library` as a test.

- [ ] Yes
- [X] No


aignas:
Right now the example is failing because of a Python import error `no module named 'google'` which I would not expect to hit here. What am I doing wrong here?

examples/py_proto_library/WORKSPACE
```diff
@@ -0,0 +1,24 @@
+workspace(name = "rules_python_py_proto_library_example")
```
rickeylev, line 1: Would this example be cleaner/simpler if we used bzlmod?

examples/py_proto_library/test.py
```diff
@@ -0,0 +1,19 @@
...
+import pricetag_pb2
+
+
+class TestCase(unittest.TestCase):
+    def test_pricetag(self):
+        got = pricetag_pb2.Pricetag(
+            name="dollar",
+            cost=5.00,
+        )
+        self.assertTrue(got is not None)
```
rickeylev, line 15: nit: assertIsNotNone(got)

alexeagle:
I agree that the current thing is broken, but I'm worried that this will be a subtle difficult breaking change for users. If they have rules_python before rules_go let's say, then this will change their protobuf version in a way they can't debug.
Of course, when bzlmod is enabled, rules_python already depends on those two modules https://github.com/bazelbuild/rules_python/blob/main/MODULE.bazel#L11-L12 so maybe we're just forced to reflect that same dependency graph in WORKSPACE

rickeylev:
re: adding rules_proto deps: Good point. How about putting it behind an arg, disabled by default, so users can easily enable the extra deps, but also easily turn that off? That at least gives them the ability to manually install what's needed without having to patch the rules.

With bzlmod, doesn't this whole problem go away? My understanding of bzlmod is it does...something...to better address this diamond dependency problem.

rickeylev (changes requested):

examples/py_proto_library/WORKSPACE
alexeagle: I think we should make every example "hybrid" with both workspace and bzlmod usage. we've been migrating all the Aspect rulesets to work this way. Keeps us honest that the two are feature-parity

python/repositories.bzl
```diff
@@ -53,6 +53,26 @@ def py_repositories():
         ],
     )
 
+    # below are dependencies if you want to use py_proto_library
```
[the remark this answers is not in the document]
alexeagle: this is definitely breaking users when done this way, since it changes the versions of these tools depending what order their WORKSPACE calls various `*_repositories` functions. Like imagine if they install python before Go. I think we need a separate `py_proto_repositories` function in this file instead.

examples/py_proto_library/WORKSPACE
alexeagle: https://github.com/aspect-build/rules_swc/commit/af52471ac451d647fd187b9e92162d6a6199e700 is an example

python/repositories.bzl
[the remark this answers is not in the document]
aignas: I've reverted the changes and I am limiting the scope of the PR to only add an example.

python/repositories.bzl
```diff
@@ -53,6 +53,26 @@ def py_repositories():
...
+    http_archive(
```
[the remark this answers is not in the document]
aignas: MODULE.bazel already has a comment on this.

examples/py_proto_library/WORKSPACE
```diff
@@ -0,0 +1,24 @@
...
+
+local_repository(
```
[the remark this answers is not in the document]
aignas: `local_repository` is how most of the examples are written, but I will use the `override_repository` here, which probably is a much cleaner approach.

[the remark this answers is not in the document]
aignas: Wanted to go with 2. but it seems that 1. is more pervasive in the repo, so did the first option. We usually use local overrides in all of the examples, so this is the desired state, I think.

examples/py_proto_library/WORKSPACE
aignas: Added bzlmod support for the example.

aignas:
Thanks @alexeagle and @rickeylev for comments. I changed the PR to be only about addition of an example as changing `py_repositories` may be suboptimal. The examples still not work on my local environment and I saw them fail on CI for the same reason as well.

examples/py_proto_library/WORKSPACE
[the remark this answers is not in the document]
rickeylev: Ah yes, I see. Thanks.

python/private/proto/py_proto_library.bzl
```diff
@@ -142,6 +142,9 @@ def _py_proto_library_rule(ctx):
         ),
         PyInfo(
             transitive_sources = default_outputs,
+            imports = depset(
+                direct = ["com_google_protobuf/python"],
```
aignas, line 146: @rickeylev, is this the right usage of the `imports` attribute here?

rickeylev: Not quite; this assumes the repo name and location of the client libraries in the other repo, which are decided elsewhere.

What we have to do is similar to how transitive_sources gets computed: have the aspect get info from the runtime and its deps, merge and return that, then have the rule finish propagating. So something like:

```
# aspect impl
  imports = depset(proto_lang_toolchain[PyInfo].imports)
  return [PyProtoInfo(imports=imports)]
# rule impl
  return [PyInfo(
    imports = depset([d[PyProtoInfo].imports for d in deps])
  ]
```

rickeylev (changes requested):

python/private/proto/py_proto_library.bzl
```diff
@@ -22,6 +22,9 @@ ProtoLangToolchainInfo = proto_common.ProtoLangToolchainInfo
 _PyProtoInfo = provider(
     doc = "Encapsulates information needed by the Python proto rules.",
     fields = {
+        "imports": """
+            (depset[str]) The field forwarding PyInfo.imports coming from
+            api_deps.""",
```
rickeylev, line 27:
```suggestion
            the proto language runtime dependency.""",
```

python/private/proto/py_proto_library.bzl
```diff
@@ -142,6 +148,7 @@ def _py_proto_library_rule(ctx):
         ),
         PyInfo(
             transitive_sources = default_outputs,
+            imports = depset(transitive = [d[_PyProtoInfo].imports for d in ctx.attr.deps]),
```
rickeylev, line 151:
```suggestion
            imports = depset(transitive = [info.imports for info in pyproto_infos]),
```

python/private/proto/py_proto_library.bzl
rickeylev: Yes. I suggested a couple minor edits, but otherwise LGTM

examples/py_proto_library/WORKSPACE
```diff
@@ -0,0 +1,66 @@
+workspace(name = "rules_python_py_proto_library_example")
+
+# You do not want to use the following command when you are using a WORKSPACE file
+# that is outside of rules_python repository.
+# This command allows targets from a local directory to be bound.
+# Which allows bazel to use targets defined in base rules_python directory.
+# If you are using this example outside of the rules_python git repo,
+# use the http_archive command that is commented out below.
+# https://bazel.build/reference/be/workspace#local_repository
```
rickeylev, line 9:
```suggestion
# The following local_path_override is only needed to run this example as part of our CI.
```

examples/py_proto_library/WORKSPACE
```diff
@@ -0,0 +1,66 @@
...
+#
+# The following StarLark would replace the `local_repository` rule mentioned above.
+#
+# load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+# http_archive(
+#     name = "rules_python",
+#     sha256 = "497ca47374f48c8b067d786b512ac10a276211810f4a580178ee9b9ad139323a",
+#     strip_prefix = "rules_python-0.16.1",
+#     url = "https://github.com/bazelbuild/rules_python/archive/refs/tags/0.16.1.tar.gz",
+# )
```
rickeylev, line 28:
```suggestion
# When not using this example in the rules_python git repo you would load the python
# rules using http_archive(), as documented in the release notes.
```

rickeylev (approved):

PR #1051, opened 2023-02-08
fix(release): wrong replacement for $(location) during code review

alexeagle:


f0rmiga (approved):

rickeylev:
Either is correct, with execpath being slightly better in theory, though effectively equivalent in practice.

rootpath basically means, "I expect this to be part of the binary's runfiles". This is somewhat more expensive as its an extra thing to materialize in runfiles and/or a packaged version of the binary (e.g subpar, zip, etc). It also means changing the file causes the binary to be rebuilt, which isn't actually necessary.

execpath basically means "I expect this to be provided as an input, but is not part of my binary itself". This is cheaper (just use the path to the already built file), and changing the file doesn't require rebuilding the binary itself.

(The Python rules ended up treating "data" as "binary-bundled-data", and now don't have a nice way of providing inputs to args _without_ causing it to be bundled into the binary. Which is kind of unfortunate).

PR #1053, opened 2023-02-08
fix: checked-in requirements imports generated requirements

f0rmiga:
When trying to add a `-r requirements.in` that is generated inside a checked-in requirements.txt, the Python script that calls pip-compile was going to the wrong location and having problems resolving the relative generated `requirements.in`.


.bazelci/presubmit.yml
```diff
@@ -237,21 +237,29 @@ tasks:
     name: compile_pip_requirements integration tests on Ubuntu
     working_directory: tests/compile_pip_requirements
     platform: ubuntu2004
+    run_targets:
+    - "bazel run //:requirements.update"
```
alexeagle: is there any added test coverage from running this on every platform? I think of the lost rainforests when we waste CI resources 😉

alexeagle (approved):
nice, thanks!

.bazelci/presubmit.yml
f0rmiga: The experience has taught me that when we don't add tests for all platforms, it breaks. ;)

There are only 2 dependencies in there, and they should be cached, so the rainforests are safe (for now).

python/pip_install/tools/dependency_resolver/dependency_resolver.py
```diff
@@ -99,27 +103,7 @@ def _fix_up_requirements_in_path(absolute_prefix, output_file):
...
-        err_msg = (
-            "Expected to find BUILD_WORKSPACE_DIRECTORY (running under `bazel run`) or "
-            "TEST_TMPDIR (running under `bazel test`) in environment."
-        )
-        print(
-            err_msg,
-            file=sys.stderr,
-        )
-        sys.exit(1)
+        shutil.copy(requirements_txt, requirements_out)
```
alexeagle, line 106: comment about how this actually copies back over the source file because we assume it's a symlink, and that's why we don't have to lookup the BUILD_WORKSPACE_DIRECTORY after all?

Also what happens if the target file didn't exist? I guess we fail in the analysis phase and don't even run this.

python/pip_install/requirements.bzl
```diff
@@ -103,6 +103,8 @@ def compile_pip_requirements(
 
     tags = tags or []
     tags.append("requires-network")
+    tags.append("no-remote")
```
alexeagle, line 106: just no-remote-exec, it should still be cached

alexeagle (approved):

PR #1061, opened 2023-02-11
Only set `py_runtime.coverage_tool` for Bazel 6 and higher.

### LINKED ISSUE #1056: 0.18.0 breaks compatibility for Bazel 5
# 🐞 bug report

### Affected Rule

 The issue is caused by the rule: python_register_toolchains

### Is this a regression?

 Yes, the previous version in which this bug was not present was: 0.17.3

### Description

I think https://github.com/bazelbuild/rules_python/pull/977 broke the compatibility for Bazel versions <6 as the `coverage_tool` attribute on `py_runtime` [was added](https://github.com/bazelbuild/bazel/commit/9d0163002ca63f4cdbaff9420380d72e2a6e38b3) in Bazel 6 . 

I could not find it stated anywhere that this is WAI.

## 🔬 Minimal Reproduction

```
python_register_toolchains(
  name = "local_config_python",
  python_version = "3.9",
)
```
```
py_binary(
    name = "foo",
    srcs = ["foo.py"],
)
```
`bazel-5.4.0 --bazelrc=/dev/null build //:foo`
## 🔥 Exception or Error
<pre><code>
ERROR: ~/.cache/bazel/_bazel_faximan/0920b5666e9ad52e0d97f6f97d994ed1/external/local_config_python_x86_64-unknown-linux-gnu/BUILD.bazel:62:11: @local_config_python_x86_64-unknown-linux-gnu//:py3_runtime: no such attribute 'coverage_tool' in 'py_runtime' rule
</code></pre>

## 🌍  Your Environment

**Operating System: Debian

<pre>
  <code>
Build label: 5.4.0
Build time: Thu Dec 15 16:14:25 2022 (1671120865)
Build timestamp: 1671120865
Build timestamp as int: 1671120865
  </code>
</pre>

**Rules_python version:**

<pre>
  <code>
0.18.0
  </code>
</pre>

rickeylev:
Avoid setting it in earlier version by checking `native.bazel_version` in the repository rule and disabling it if less than Bazel 6 is detected. A warning is also printed if coverage was requested, but the Bazel version check is ignoring it.

Fixes #1056


f0rmiga (approved):

alexeagle:
Could we add a Bazel 5 test in our matrix?

rickeylev:
Yeah, +1

python/repositories.bzl
```diff
@@ -472,6 +486,19 @@ def python_register_toolchains(
 
     toolchain_repo_name = "{name}_toolchains".format(name = name)
 
+    bazel_major = int(native.bazel_version.split(".")[0])
```
alexeagle, line 489: I think this is broken for bazel built from head.

FYI https://github.com/aspect-build/bazel-lib/blob/4edb5018e89cd56515fed269056429c353cd7c43/lib/private/utils.bzl#L146

rickeylev: Yes, that should be fixed by #1063 (let me know otherwise)

> aspect utils.bzl

Oh, nice find. That would probably also work with a custom Bazel 5.x build, too.

PR #1063, opened 2023-02-11
Allow building with unreleased Bazel versions.

rickeylev:
Unreleased Bazel versions don't set `native.bazel_version`, which was causing the version detection logic to error. To fix, just check it's non-empty. Treat a missing version as a sufficient version.

### DIFF
--- a/python/repositories.bzl
+++ b/python/repositories.bzl
@@ -486,18 +486,20 @@ def python_register_toolchains(
 
     toolchain_repo_name = "{name}_toolchains".format(name = name)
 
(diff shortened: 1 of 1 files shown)


f0rmiga (approved):

alexeagle:
https://docs.aspect.build/rules/aspect_bazel_lib/docs/utils#is_bazel_6_or_greater FYI

Maybe rules_python should take a dep on aspect_bazel_lib - or maybe you can get the bazel_skylib maintainers to allow broader additions.

PR #1064, opened 2023-02-12
fix: Propagate testonly et al for wheel `.dist` targets

### LINKED ISSUE #1057: Cannot build testonly py_wheel
# 🐞 bug report

### Affected Rule

 The issue is caused by the rule:  py_wheel

### Is this a regression?

 Yes, the previous version in which this bug was not present was: 0.16.2

### Description

With the addition of `.dist` meta targets in https://github.com/bazelbuild/rules_python/pull/1019, I can no longer build `//...` when including a `py_wheel` tagged `testonly`.

## 🔬 Minimal Reproduction

```
load("@rules_python//python:packaging.bzl", "py_wheel")
py_wheel(
    name = "foo_test_wheel",
    testonly = 1,
    distribution = "foo_test",
    version = "0.0.1",
)
```

`bazel --bazelrc=/dev/null build //...`

## 🔥 Exception or Error

<pre><code>
ERROR: ~/python/BUILD:240:9: in py_wheel_dist rule //foo_test_wheel.dist: non-test target '//:foo_test_wheel.dist' depends on testonly target '//:foo_test_wheel' and doesn't have testonly attribute set
</code></pre>

## 🌍  Your Environment

**Operating System:**

Debian

<pre>
  <code>
Build label: 5.4.0
Build time: Thu Dec 15 16:14:25 2022 (1671120865)
Build timestamp: 1671120865
Build timestamp as int: 1671120865
  </code>
</pre>

**Rules_python version:**

<pre>
  <code>
0.17.3
  </code>
</pre>

rickeylev:
The `.dist` target depends on the wheel, so it must copy the `testonly` setting as well as some others.

* Also adds a utility function to do this, since the multi-version rules also do this copying.

Fixes #1057

### DIFF
--- a/docs/BUILD.bazel
+++ b/docs/BUILD.bazel
@@ -84,6 +84,7 @@ bzl_library(
         "//python/private:py_package.bzl",
         "//python/private:py_wheel.bzl",
         "//python/private:stamp.bzl",
+        "//python/private:util.bzl",
     ],
 )
 
--- a/examples/wheel/BUILD.bazel
+++ b/examples/wheel/BUILD.bazel
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+load("@bazel_skylib//rules:build_test.bzl", "build_test")
 load("//examples/wheel/private:wheel_utils.bzl", "directory_writer")
 load("//python:defs.bzl", "py_library", "py_test")
 load("//python:packaging.bzl", "py_package", "py_wheel")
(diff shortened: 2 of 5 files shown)


aignas (approved):

aignas (approved):
LGTM, thanks.

f0rmiga (approved):

alexeagle:
Oops thanks - fyi we have that helper too https://docs.aspect.build/rules/aspect_bazel_lib/docs/utils#propagate_well_known_tags

PR #1067, opened 2023-02-14
feat: update `compile_pip_requirements` to support multiple input files

cj81499:
`pip-compile` can compile multiple input files into a single output file, but `rules_python`'s `compile_pip_requirements` doesn't currently support this.

With this change, the `requirements_in` argument to `compile_pip_requirements` can now accept a list of strings (in addition to the previously accepted argument types).

In order to support a variable number of input files, my coworker (@lpulley) and I updated `dependency_resolver.py` to use the `click` CLI library. We felt this was acceptable since `pip-compile` already requires `click` to run, so we're not adding a new dependency.

We also made changes to the script to avoid mutating `sys.argv`, instead opting to build a new list (`argv`) from scratch that'll be passed to the `pip-compile` CLI. While subjective, I feel this improves readability, since it's not immediately obvious what's in `sys.argv`, but it's clear that `argv` begins empty, and is added to over the course of the program's execution.


f0rmiga:
Thanks for this. How about we split this into 3 PRs? The first would be the changes to pre-commit, the second using click, and the third adding the new functionality.

aignas:
Right now this usecase might be supported by having three files:
```
# requirements.in
-r requirements_1.in
-r requirements_2.in

# requirements_1.in
numpy == 1.20.0

# requirements_2.in
boto3
```

And then use the `data` attribute to add dependencies to `requirements.in` compilation. This got implemented in #1026.

cj81499:
I split this PR into 3 as requested by @f0rmiga (#1070, #1071, and this one (which is blocked by #1071)). I expect it'll be easier to review now, so I really appreciate the suggestion!

cj81499:
@aignas I've actually tried doing exactly that! Unfortunately, I couldn't quite get it working right. We're using generated `requirements.in` files and I couldn't figure out the right path to put in the "root" `requirements.in` file (the one that is given directly to `compile_pip_requirements` as `requirements_in`). I even tried generating the "root" `requirements.in` file using `$(location //xxx:requirements.in)` (and the other path related make variable substitution options), but I couldn't get it to work correctly.

Regardless, I feel that accepting multiple input files is simpler than generating an additional "root" `requirements.in` file.

f0rmiga:
I believe my last fix on https://github.com/bazelbuild/rules_python/pull/1053 will cover your use case with what @aignas described. I merged it yesterday. I'm still open to reviewing your PR, though.

cj81499:
I still think this would be a nice feature to have (since I'm currently using https://github.com/cj81499/rules_python/commit/bc4569464ee62898b6eeaaa8bf7c98f42ef89e39 and it's working for my project), but I'll try out the latest commit from `main` to see if it's fixed later today.

cj81499:
@f0rmiga using 767b050e45c49e63cfe103e6b97f7e2fd9be5e2e and the `-r` approach, `//python:requirements_test` fails with:

```
Checking python/requirements.txt
  File "/home/<username>/.cache/bazel/_bazel_<username>/<hash>/sandbox/linux-sandbox/40/execroot/bazel-playground/bazel-out/k8-fastbuild/bin/python/requirements_test.runfiles/pypi__pip/pip/_internal/req/req_file.py", line 540, in get_file_content
    with open(url, "rb") as f:
         ^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: 'python/python/<name_of_project>/requirements.in'
```

Notice the repeated "python" in the last line of output.

The requirements file is being generated such that the line `-r $(rootpath //python/<name_of_project>:requirements.in)` will be included in the file (after make variable substitution of course). I've tried using `location`, `rlocationpath`, etc too, but none work (rootpath is closest to the expected path of `python/<name_of_project>/requirements.in`).

It seems that `compile_pip_requirements` is adding a prefix to the path for some reason?

My best guess is that this is b/c `compile_pip_requirements` is used from `python/BUILD.bazel`, rather than at `WORKSPACE` root (which seems to be the typical case based on the examples provided in this repo).

cj81499:
Just an update, getting the CLA signed is taking (far) longer than originally expected. I'm still working on it!

python/pip_install/tools/dependency_resolver/dependency_resolver.py
```diff
@@ -121,12 +124,17 @@ def _select_golden_requirements_file(
...
+    argv.append(f"--output-file={requirements_txt if UPDATE else requirements_out}")
+    argv.extend(
+        (req_in if req_in_path.exists() else resolved_req_in)
+        for req_in, req_in_path, resolved_req_in in zip(
+            requirements_ins, requirements_in_paths, resolved_requirements_ins
+        )
     )
+    argv.extend(extra_args)
+
+    argv.extend(extra_args)
```
cj81499, line 137: oops!

```suggestion
    argv.extend(extra_args)
```

tests/multiple_requirements_in/README.md
cj81499, line 1: note to self, this needs to be updated

daveboutcher (approved):

cj81499:
I believe the test failure is an unrelated (flaky) failure.

```
2023/11/02 23:20:29 could not download Bazel: could not copy from https://releases.bazel.build/6.4.0/release/bazel-6.4.0-linux-x86_64 to /var/lib/buildkite-agent/.cache/bazelisk/downloads/bazelbuild/bazel-6.4.0-linux-x86_64/bin/download461043671: stream error: stream ID 1; INTERNAL_ERROR
```

tests/multiple_requirements_in/WORKSPACE
rickeylev, line 1: You can delete this file; a separate workspace isn't required for testing compile_pip_requirements. We don't have sufficient CI capacity to add another job for a separate workspace, either.

python/pip_install/requirements.bzl
```diff
@@ -53,7 +53,7 @@ def compile_pip_requirements(
         generate_hashes: whether to put hashes in the requirements_txt file.
         py_binary: the py_binary rule to be used.
         py_test: the py_test rule to be used.
-        requirements_in: file expressing desired dependencies.
+        requirements_in: file(s) expressing desired dependencies.
```
rickeylev, line 56: A recent PR deprecated the `requirements_in` arg in favor of `src` to allow toml files as inputs, so you'll need to sync to head and update the code in light of that.

Does passing multiple toml files also work? If not, that's fine. Just curious.

tests/multiple_requirements_in/BUILD.bazel
```diff
@@ -0,0 +1,14 @@
+load("@rules_python//python:pip.bzl", "compile_pip_requirements")
+
+compile_pip_requirements(
+    name = "requirements",
+    extra_args = [
```
rickeylev, line 5: Please add a comment as to why special args are needed

rickeylev (changes requested):

cj81499:
Resolved the conflicts.

CI failure seems unrelated to my changes. My theory is it was caused by GitHub's partial outage today. Would appreciate a retry!

tests/multiple_requirements_in/WORKSPACE
rickeylev: Just delete the WORKSPACE file here for now.

Thanks for pointing out tests/compile_pip_requirements/WORKSPACE. I'm not sure why that's a separate workspace; maybe it's cruft left over from...something. In any case, let's just keep the multiple_requirements_in directory for now. I'll send a PR to cleanup tests/compile_pip_requirements and suss out why it might be the way it is.

aignas (commented):
Overall, LGTM, but we should make sure that the Changelog is not mentioning teh Src attribute and uses srcs instead.

Thanks for doing this improvement and using click for argument handling!

cj81499:
@aignas Good catch! I updated the changelog. Thanks for taking a look :)

cj81499:
Spent some time investigating the Windows-specific failures. I'm having a difficult time sorting it out independently and could use some help. I pushed a commit (make sure to read the commit message!) to a separate branch with my findings so far. https://github.com/cj81499/rules_python/commit/b843fd650f270a9309f5babd4ac22f7c7c6f852b

aignas:
I think these are just flakes, restarted the jobs.

lpulley:
(I work with @cj81499.) Any chance this will pass CI now? Maybe there have been some improvements? We've been using this PR as a patch for 14 months now since we're not sure what's wrong with the CI here.

We did take a crack at it on Windows a few months ago, as @cj81499 mentioned: https://github.com/cj81499/rules_python/commit/b843fd650f270a9309f5babd4ac22f7c7c6f852b

cj81499:
@rickeylev @aignas I (and my coworkers) would still quite like to make this contribution. Can you provide any guidance on the failures we saw?

CHANGELOG.md
```diff
@@ -624,7 +624,7 @@ A brief description of the categories of changes:
   default, which will cause `gazelle` to change third-party dependency labels
   from `@pip_foo//:pkg` to `@pip//foo` by default.
 
-* The `compile_pip_requirements` now defaults to `pyproject.toml` if the `src`
+* The `compile_pip_requirements` now defaults to `pyproject.toml` if the `srcs`
```
aignas, line 627: You should have a new changelog entry instead of modifying old entries. What you are doing is a breaking change here.

python/private/pypi/pip_compile.bzl
```diff
@@ -53,8 +53,8 @@ def pip_compile(
 
     Args:
         name: base name for generated targets, typically "requirements".
-        src: file containing inputs to dependency resolution. If not specified,
-            defaults to `pyproject.toml`. Supported formats are:
+        srcs: a list of files containing inputs to dependency resolution. If not specified,
```
aignas, line 56: If you want to support srcs list, you should add it as a new arg instead of changing the existing arg. As you've seen, you had to modify all of the examples to make this work, which means that all of the users would have to do this change as well.

python/private/pypi/pip_compile.bzl
```diff
@@ -88,16 +91,15 @@ def pip_compile(
...
 
     # Use the Label constructor so this is expanded in the context of the file
     # where it appears, which is to say, in @rules_python
     pip_compile = Label("//python/private/pypi/dependency_resolver:dependency_resolver.py")
 
     loc = "$(rlocationpath {})"
 
-    args = [
-        loc.format(src),
+    args = ["--src=%s" % loc.format(src) for src in srcs] + [
```
aignas, line 102: Here only having only this way of specifying the srcs is fine

tests/multiple_inputs/BUILD.bazel
```diff
@@ -0,0 +1,30 @@
...
+    srcs = [
+        "a/pyproject.toml",
+        "b/pyproject.toml",
+    ],
+    requirements_txt = "multiple_pyproject_toml.txt",
+)
+
+compile_pip_requirements(
+    name = "multiple_inputs",
+    srcs = [
```
aignas, line 23: In general this looks reasonable.

aignas (changes requested):
Not sure why windows is failing (and unfortunately I don't have a why to debug this). Disabling the failing test on windows would be one way to make CI happy.

That said, us supporting src and srcs at the same time is not great, but probably something we have to do here.

CHANGELOG.md
cj81499: Ugh, yep. When this was originally created, `src` hadn't been published 😭

cj81499:
> Disabling the failing test on windows would be one way to make CI happy.

Is that acceptable...? If so, how would I go about doing that?

CHANGELOG.md
cj81499: Should we deprecate `src` in favor of `srcs`?

python/private/pypi/pip_compile.bzl
```diff
@@ -63,7 +69,7 @@ def pip_compile(
         generate_hashes: whether to put hashes in the requirements_txt file.
         py_binary: the py_binary rule to be used.
         py_test: the py_test rule to be used.
-        requirements_in: file expressing desired dependencies. Deprecated, use src instead.
+        requirements_in: file expressing desired dependencies. Deprecated, use src or srcs instead.
```
cj81499, line 72: Should we deprecate `src` in favor of `srcs` (like how `requirements_in` was deprecated in favor of `src`)?

I figure this could be done in a follow-up PR, but it seems worth considering here.

cj81499:
Massive shout-out to my coworker @lpulley for pairing with me to root-cause and fix the failures on Windows!

We managed to implement a _bit_ of workaround/hack:

https://github.com/cj81499/rules_python/blob/9a5653ca7f32fbf698dfb1ba1b7c425dc521c7d5/python/private/pypi/pip_compile.bzl#L160-L164

I think the idea that ["Maybe rules_python should have it's own build-backend that does almost nothing?"](https://github.com/bazelbuild/rules_python/pull/1067#issuecomment-1813819649) _might_ be less hacky, but I suspect a lot of work to implement and maintain (and certainly not something I have bandwidth to contribute). Hopefully the workaround/hack is sufficient.

python/private/pypi/pip_compile.bzl
aignas: Yeah, I think deprecating `src` sounds like a good idea, but it can be done later.

python/private/pypi/pip_compile.bzl
```diff
@@ -147,6 +157,12 @@ def pip_compile(
     # cheap way to detect the bazel version
     _bazel_version_4_or_greater = "propeller_optimize" in dir(native)
 
+    # setuptools (the default python build tool) attempts to find user configuration in the user's home direcotory. This seems to work fine on linux and macOS, but fails on Windows. We provide a fake USERPROFILE env variable to allow setuptools to proceed without finding user-provided configuration.
```
aignas, line 160: Thank you very much for finding this, I think it is super good to have this fix, I definitely stumbled upon this a few times. This should be mentioned in the CHANGELOG.

nit: probably should split across multiple lines like the rest of the codebase.

python/private/pypi/pip_compile.bzl
```diff
@@ -147,6 +157,12 @@ def pip_compile(
...
+    kwargs["env"] = kwargs.pop("env", {}) | select({
+        "@@platforms//os:windows": {"USERPROFILE": "Z:\\FakeSetuptoolsHomeDirectoryHack"},
```
aignas, line 162: I think we should not override any `USERPROFILE` settings if the user defines them. However, I am not sure it is possible to do this if we are using a select.

On the other hand, maybe we don't need to allow user to customize it at all because how bazel is attempting to sandbox everything. Maybe we could sniff if the `tags` have `local` or `no-sandbox` and do not apply the `USERPROFILE` env here?

python/private/pypi/pip_compile.bzl
```diff
@@ -147,6 +157,12 @@ def pip_compile(
     # cheap way to detect the bazel version
     _bazel_version_4_or_greater = "propeller_optimize" in dir(native)
 
+    # setuptools (the default python build tool) attempts to find user configuration in the user's home direcotory. This seems to work fine on linux and macOS, but fails on Windows. We provide a fake USERPROFILE env variable to allow setuptools to proceed without finding user-provided configuration.
+    kwargs["env"] = kwargs.pop("env", {}) | select({
```
aignas, line 161: huh? I thought `env` is a non-configurable attribute (at least it was in some other rules), if the CI is happy, I'll trust it.

python/private/pypi/pip_compile.bzl
lpulley: Well, we could just do nothing if USERPROFILE is present, and apply our select if it isn't.

python/private/pypi/pip_compile.bzl
cj81499: Sure thing.

fwiw, I think this workaround is sufficient for merge, but I also think it's worth a stab at contributing a change to setuptools itself (to tolerate failures to locate the user's home directory) so we don't have to do this at all!

python/private/pypi/pip_compile.bzl
cj81499: I've gone with the approach @lpulley suggested. Let me know if you have a different idea of how this ought to work.

cj81499:
FYI: I've opened https://github.com/pypa/distutils/pull/278. I figure it's best to handle this upstream of `rules_python` if possible. I don't think the workaround is too nasty to merge in the meantime, but _hopefully_ it can be removed if/when that PR is merged to `distutils` (and vendored by `setuptools`).

python/private/pypi/pip_compile.bzl
```diff
@@ -147,6 +157,18 @@ def pip_compile(
...
+    # configuration in the user's home direcotory. This seems to work fine on
+    # linux and macOS, but fails on Windows, so we conditionally provide a fake
+    # USERPROFILE env variable to allow setuptools to proceed without finding
+    # user-provided configuration.
+    env = kwargs.get("env", {})
+    if "USERPROFILE" not in env:
+        kwargs["env"] = env | select({
+            "@@platforms//os:windows": {"USERPROFILE": "Z:\\FakeSetuptoolsHomeDirectoryHack"},
+            "//conditions:default": {},
+        })
```
cj81499, line 170: Pointed out by a coworker of mine (not @ lpulley this time!), `not in` will fail if `env` is a `select`.

We came up with the following to fix this, but the `dict` union operator (`|`) isn't supported pre bazel 6 and we weren't sure which approach was preferred.

```suggestion
        kwargs["env"] = select({
        "@@platforms//os:windows": {"USERPROFILE": "Z:\\FakeSetuptoolsHomeDirectoryHack"},
        "//conditions:default": {},
    }) | kwargs.get("env", {})
```

aignas: We only need to support bazel 6 and above with the latest version of `rules_python` so this is fine.

aignas:
> FYI: I've opened [pypa/distutils#278](https://github.com/pypa/distutils/pull/278). I figure it's best to handle this upstream of `rules_python` if possible. I don't think the workaround is too nasty to merge in the meantime, but _hopefully_ it can be removed if/when that PR is merged to `distutils` (and vendored by `setuptools`, and released, and `rules_python` upgrades to a newer version of `setuptools` that include the fix).

I am wondering if we should instead open a ticket in `setuptools` because `distutils` is deprecated ([PEP632]) and you may have a hard time getting response from them.

[PEP632]: https://peps.python.org/pep-0632/

I think I'll just raise a separate issue in rules_python to track this issue, let's continue conversation there.

aignas (approved):
LGTM

rickeylev (approved):

cj81499:
> I am wondering if we should instead open a ticket in `setuptools` because `distutils` is deprecated ([PEP632](https://peps.python.org/pep-0632/)) and you may have a hard time getting response from them.

Maybe? https://github.com/pypa/distutils _has_ been updated recently (within the last month or so), so I _someone_ must still be looking at it. You're more than welcome to open an issue on https://github.com/pypa/setuptools to try to draw attention to the issue/my PR https://github.com/pypa/distutils/pull/278 if you think that's worthwhile.

PR #1071, opened 2023-02-14
refactor: use click in dependency_resolver.py

cj81499:
Using click makes it easier to parse arguments. Many args are now named arguments
(options), and the need for using positional args with stub `"None"` values isn't
necessary anymore.

There is already a dependency on click via piptools, so this doesn't introduce a new
dependency.

Relates to #1067

### DIFF
--- a/python/pip_install/tools/dependency_resolver/dependency_resolver.py
+++ b/python/pip_install/tools/dependency_resolver/dependency_resolver.py
@@ -19,7 +19,9 @@ import os
 import shutil
 import sys
 from pathlib import Path
+from typing import Optional, Tuple
 
+import click
 import piptools.writer as piptools_writer
 from piptools.scripts.compile import cli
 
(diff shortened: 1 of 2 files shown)


cj81499:
@daveboutcher This is ready for your review

cj81499:
@hrfuller are you available to take a look at this? If not, is there someone else who might be able to?

hrfuller:
> are you available to take a look at this? If not, is there someone else who might be able to?
Yes, I can take a look. @rickeylev is a more active maintainer than me these days so I'd like him to take a look as well.

python/pip_install/tools/dependency_resolver/dependency_resolver.py
```diff
@@ -128,6 +131,8 @@ def _locate(bazel_runfiles, file):
     os.environ["LC_ALL"] = "C.UTF-8"
     os.environ["LANG"] = "C.UTF-8"
 
+    argv = []
```
hrfuller, line 134: It looks like we lose sys.argv[0] compared to before. I'm not sure if the piptools cli expects argv to have argv[0] be the name of the original script but maybe we can add sys.argv[0] here in case.

python/pip_install/tools/dependency_resolver/dependency_resolver.py
```diff
@@ -153,15 +157,14 @@ def _locate(bazel_runfiles, file):
...
-    sys.argv.append(requirements_file_relative if UPDATE else requirements_out)
-    sys.argv.append(
+    argv.append("--generate-hashes")
+    argv.append(f"--output-file={requirements_file_relative if UPDATE else requirements_out}")
+    argv.append(
         requirements_in_relative
         if Path(requirements_in_relative).exists()
         else resolved_requirements_in
     )
-    print(sys.argv)
```
hrfuller, line 164: Unclear why this was here. probably for debugging?

python/pip_install/tools/dependency_resolver/dependency_resolver.py
cj81499: Just to be clear, are you suggesting we change this to the following?

```suggestion
    argv = [sys.argv[0]]
```

python/pip_install/tools/dependency_resolver/dependency_resolver.py
cj81499: That was my assumption too. Didn't seem like something that ought to stick around (adding clutter to stdout!).

Happy to add it back if someone has a good reason for it to stay there.

cj81499 (commented):
> @rickeylev is a more active maintainer than me these days so I'd like him to take a look as well.

@hrfuller I take this to mean you'd like me to hold off on clicking the merge button until they review?

cj81499:
@hrfuller @rickeylev bump!

I would really like to get this (and the related #1067) merged.

cj81499:
@hrfuller @rickeylev I'd still like to get this (and #1067) merged.
Please take a look at your earliest convenience.

python/pip_install/tools/dependency_resolver/dependency_resolver.py
rickeylev: From what I can tell, omitting argv[0] when calling cli() is correct.

Under the hood, piptools is using click, and the implementation is obscured by a ton of decorators. Digging into it...
* `compile.cli` is an instance of the click `Command` class.
* `Command.__call__` is an alias for `Command.main`
* `Command.main` docs say, for the first positionl arg (named `args`): "the arguments that should be used for parsing. If not provided sys.argv[1:] is used"

So it doesn't expect args to contain argv[0]. The docs go on further to say the optional `prog_name` arg is initialized to sys.argv[0] if not specified.

As a side note, there is a `standalone_mode` arg, which controls whether the command tries to exit or not.

rickeylev (approved):

rickeylev:
Hi @cj81499,

Thanks for your PR. Sorry it took so long to be reviewed. I had let this languish because I thought it was introducing a new dependency on click, but we already have that dependency, so this isn't as bad as I thought.

cj81499:
@rickeylev thank you for getting around to this! ❤️

fwiw, I did say the following in the PR description ;)

> There is already a dependency on click via piptools, so this doesn't introduce a new dependency.

Excited that #1067 will be unblocked once this makes its way through the queue!

PR #1073, opened 2023-02-16
fix: missing flags and envs while building wheels

f0rmiga:
I found that in some cases, there are missing include directories or missing search paths for libraries. I tried to capture as much as possible for both clang and gcc. I'm not sure how breaking this could be for some people. If we decide to merge as is, we may need to do some tweaking before releasing the next version of rules_python.

This should be solved once we are able to do toolchain resolution for CC deps and inject into the wheel building using actions.

### DIFF
--- a/python/pip_install/pip_repository.bzl
+++ b/python/pip_install/pip_repository.bzl
@@ -20,6 +20,7 @@ load("//python/pip_install:requirements_parser.bzl", parse_requirements = "parse
 load("//python/pip_install/private:srcs.bzl", "PIP_INSTALL_PY_SRCS")
 
 CPPFLAGS = "CPPFLAGS"
+LDFLAGS = "LDFLAGS"
 
 COMMAND_LINE_TOOLS_PATH_SLUG = "commandlinetools"
 
@@ -110,7 +111,7 @@ def _get_xcode_location_cflags(rctx):
         "-isysroot {}/SDKs/MacOSX.sdk".format(xcode_root),
     ]
 
-def _get_toolchain_unix_cflags(rctx):
+def _get_toolchain_unix_cflags(rctx, c_compiler):
     """Gather cflags from a standalone toolchain for unix systems.
 
     Pip won't be able to compile c extensions from sdists with the pre built python distributions from indygreg
@@ -125,6 +126,25 @@ def _get_toolchain_unix_cflags(rctx):
     if not is_standalone_interpreter(rctx, rctx.attr.python_interpreter_target):
         return []
 
+    include_paths = []
+
+    is_clang = False
+    if c_compiler:
+        err = rctx.execute([c_compiler, "-v"])
+        out = err.stderr.strip()
+        is_clang = "clang" in out.split("\n")[0]
+
+        if is_clang:
+            args = ["-E", "-Wp,-v", "-xc", "-"]
+        else:
+            args = ["-E", "-Wp,-v", "-xc++", "-"]
+        err = rctx.execute([c_compiler] + args)
+        out = err.stderr.strip()
+        parsed = [line.strip() for line in out.split("\n") if line.startswith(" ")]
+        if is_clang:
+            include_paths.append(parsed[0] + "/../../../../include/c++/v1")
+        include_paths += parsed
+
     er = rctx.execute([
         rctx.path(rctx.attr.python_interpreter_target).realpath,
         "-c",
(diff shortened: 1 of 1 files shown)


jvolkman:
Do you have any example sdists that were problematic?

f0rmiga:
@jvolkman in my case it was https://github.com/kpu/kenlm.

python/pip_install/pip_repository.bzl
```diff
@@ -125,20 +126,65 @@ def _get_toolchain_unix_cflags(rctx):
...
+
+        if is_clang:
+            args = ["-E", "-Wp,-v", "-xc", "-"]
+        else:
+            args = ["-E", "-Wp,-v", "-xc++", "-"]
+        err = rctx.execute([c_compiler] + args)
+        out = err.stderr.strip()
+        parsed = [line.strip() for line in out.split("\n") if line.startswith(" ")]
+        if is_clang:
+            include_paths.append(parsed[0] + "/../../../../include/c++/v1")
```
alexeagle, line 145: could you comment why four `..` segments is the right number? seems magic

python/pip_install/pip_repository.bzl
```diff
@@ -125,20 +126,65 @@ def _get_toolchain_unix_cflags(rctx):
     if not is_standalone_interpreter(rctx, rctx.attr.python_interpreter_target):
         return []
 
+    include_paths = []
+
+    is_clang = False
+    if c_compiler:
```
alexeagle, line 132: could you add some commenting about the intent/approach here? it seems like it's "call the compiler to find out what includes it expects by default" or something

python/pip_install/pip_repository.bzl
```diff
@@ -125,20 +126,65 @@ def _get_toolchain_unix_cflags(rctx):
...
+        out = err.stderr.strip()
+        is_clang = "clang" in out.split("\n")[0]
+
+        if is_clang:
+            args = ["-E", "-Wp,-v", "-xc", "-"]
+        else:
+            args = ["-E", "-Wp,-v", "-xc++", "-"]
+        err = rctx.execute([c_compiler] + args)
+        out = err.stderr.strip()
+        parsed = [line.strip() for line in out.split("\n") if line.startswith(" ")]
```
alexeagle, line 143: is it okay that this throws away lines in `out` which don't start with a space character?

python/pip_install/pip_repository.bzl
```diff
@@ -125,20 +126,65 @@ def _get_toolchain_unix_cflags(rctx):
...
+    if c_compiler:
+        err = rctx.execute([c_compiler, "-v"])
+        out = err.stderr.strip()
+        is_clang = "clang" in out.split("\n")[0]
+
+        if is_clang:
+            args = ["-E", "-Wp,-v", "-xc", "-"]
+        else:
+            args = ["-E", "-Wp,-v", "-xc++", "-"]
+        err = rctx.execute([c_compiler] + args)
```
alexeagle, line 141: any of these executions potentially slow enough to merit an `rctx.progress` ?

alexeagle (approved):

rickeylev:
re: searching local system for compiler stuff: FWIW: I _think_ there's some part of the cc stuff built into bazel that searches for that sort of stuff. I ran into it getting confused when trying to set something up on macs last week. Looking around, there's a cc_configure repository rule, which might help, or give some inspiration? [cc_configure.bzl](https://github.com/bazelbuild/bazel/blob/master/tools/cpp/cc_configure.bzl)

hth

python/pip_install/pip_repository.bzl
thirtyseven, line 143: Testing this out on a Mac (ARM, macOS 13.2.1), I found that `clang -Wp,-v -xc -` returns this output:
```
clang -cc1 version 14.0.0 (clang-1400.0.29.202) default target arm64-apple-darwin22.3.0
ignoring nonexistent directory "/usr/local/include"
ignoring nonexistent directory "/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/local/include"
ignoring nonexistent directory "/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/Library/Frameworks"
#include "..." search starts here:
#include <...> search starts here:
 /Library/Developer/CommandLineTools/usr/lib/clang/14.0.0/include
 /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include
 /Library/Developer/CommandLineTools/usr/include
 /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/Frameworks (framework directory)
End of search list.
```
The last entry includes some extra junk after the path which ends up being passed verbatim to `cc -i` and breaking sdist compilation with errors like:
```
clang: error: no such file or directory: '(framework'; did you mean '-framework'?
```
This change seems to fix it:
```suggestion
        parsed = [line.strip().partition(" ")[0] for line in out.split("\n") if line.startswith(" ")]
```

chrislovecnm:
@f0rmiga is this ready to merge?  I am not sure why it is not in main.

chrislovecnm (changes requested):
If I read the comments correctly, I think @thirtyseven found some errors on OSX.  @f0rmiga and @thirtyseven thoughts?

thirtyseven:
> If I read the comments correctly, I think @thirtyseven found some errors on OSX. @f0rmiga and @thirtyseven thoughts?

Yeah, this PR as-is did not survive testing when rolled out to the ~200 Mac users on my team. My suggested change fixed one of the errors, but users reported other errors when building bdists. Unfortunately I've lost the build failure logs for the other errors, but my recollection was that they were on macOS 12.x.

chrislovecnm (approved):

python/pip_install/pip_repository.bzl
chrislovecnm: @f0rmiga can we get this change in?

chrislovecnm (changes requested):
@f0rmiga ptla https://github.com/bazelbuild/rules_python/pull/1073#discussion_r1148096464

PR #1074, opened 2023-02-16
Add some docs about how to configure coverage.

rickeylev:
This is to replace the docs on bazel.build that talk about coverage support for Python.

### DIFF
--- a/docs/coverage.md
+++ b/docs/coverage.md
@@ -8,7 +8,7 @@ when configuring toolchains.
 
 ## Enabling `rules_python` coverage support
 
-Enabling the coverage support bundled with `rules_path` just requires setting an
+Enabling the coverage support bundled with `rules_python` just requires setting an
 argument when registerting toolchains.
 
 For Bzlmod:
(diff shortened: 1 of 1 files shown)


aignas (approved):
Thank you for making these changes. Having docs here is really useful. I also think that we should start generating `repositories.bzl` documentation which should include the docstring for the toolchain registration functions. The doc generation could be done in a separate PR.

docs/coverage.md
```diff
@@ -0,0 +1,58 @@
...
+
+As of Bazel 6, the Python toolchains and bootstrap logic supports providing
+coverage information using the `coverage` library.
+
+As of `rules_python` version `0.18.1`, builtin coverage support can be enabled
+when configuring toolchains.
+
+## Enabling `rules_python` coverage support
+
+Enabling the coverage support bundled with `rules_path` just requires setting an
```
f0rmiga, line 11:
```suggestion
Enabling the coverage support bundled with `rules_python` just requires setting an
```

f0rmiga (approved):

PR #1091, opened 2023-02-24
py_proto_libray: Add api_deps to transitive_sources

comius:
## PR Type

- [x] Bugfix

Sources from runtime needed by py_proto_library are not added to PyInfo.

## What is the new behavior?

The sources are added.

- [ ] Yes
- [x] No


rickeylev:
LGTM, but...

Was there a failure you found from them missing? I'm not sure what, if anything, is actually using PyInfo.transitive_sources.

rickeylev (approved):

comius:
It seems, this is not needed.

PR #1095, opened 2023-02-24
Add pre-commit hook for running addlicense

rickeylev:
Integrate the addlicense.sh script with the pre-commit config so that it automatically runs when git committing.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,5 +1,0 @@
-- id: addlicense
-  name: Add copyright headers
-  entry: addlicense.sh
-  language: script
-  verbose: true
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,0 +1,5 @@
+- id: addlicense
+  name: Add copyright headers
+  entry: addlicense.sh
+  language: script
+  verbose: true
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -37,3 +37,7 @@ repos:
     rev: 22.8.0
     hooks:
       - id: black
+  - repo: .
+    rev: HEAD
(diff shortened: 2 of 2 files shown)


.pre-commit-hooks.yaml
```diff
@@ -0,0 +1,5 @@
+- id: addlicense
```
alexeagle, line 1: what is the function of this added file? what calls it?

alexeagle:
The standard way to do this is expose a pre-commit hook from the addlicense repo, so that everyone can add it, and you don't need a shell script or a new `.pre-commit-hooks.yaml` here

rickeylev:
Thanks for the pointer. I sent https://github.com/google/addlicense/pull/143. Once accepted (I don't see why it wouldn't be...), we'll have to wait for them to do a release with a stable tag to use in the configs. I have another PR ready for when that happens.

In the meantime, I found the docs that tell how to correctly setup a local hook and updated the code to do that. PTAL.

f0rmiga:
It bothers me that pre-commit doesn't use Bazel for the binaries. The best way to deal with this is:

- Pull the binaries using http_archives.
- Create an alias with selects to resolve the platform.
- Create a shell wrapper that cd into `BUILD_WORKSPACE_DIRECTORY` and execute the binary:
```shell
#!/usr/bin/env bash

set -o errexit -o nounset

cd "${BUILD_WORKSPACE_DIRECTORY}"

exec "${ADDLICENSE_BINARY}" "$@"
```
- Add the `entry` attribute to the pre-commit hook with `entry: "bazel run //<pre-commit wrappers package>:addlicense --"`.

chrislovecnm:
Can we get this into a test?  CI should fail on this.

chrislovecnm:
@rickeylev ping

PR #1099, opened 2023-02-25
docs: doc that the Conventional Commit style should be used for merged commits and PRs

rickeylev:
Having a standard format for commit messages makes it easier to understand what general behaviors and functionality are being affected. This makes it easier to easier to generate changelog information for releases and quickly skim PRs.

The default commitizen style is used, which is Conventional Commits. Docs are updated to give a brief description of the format and common types.

This isn't enforced using pre-commit hooks because PRs squash commits and use the PR description for the merged commit message, which makes linting intermediate commits rather moot.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -25,6 +25,13 @@ repos:
           - --warnings=all
       - id: buildifier-lint
         args: *args
+  # Enforce that commit messages allow for later changelog generation
+  - repo: https://github.com/commitizen-tools/commitizen
+    rev: v2.42.0
+    hooks:
+      # Requires that commitizen is already installed
+      - id: commitizen
+        stages: [commit-msg]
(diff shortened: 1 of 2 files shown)


rickeylev:
Actually, after playing around with this a bit more, the commitizen local hook is kind of annoying. I'm not sure it _really_ helps, either. The thing I *really* want is for the commits that get merged to have the structured format. Having it in PRs is a nice to have, but not a necessity. I don't care about what an author's local chain of commits looks like.

I'm going to update this to keep the docs, but remove the commitizen hook. Maybe I'll look into if there's a GH action that can do some basic checking of commit messages to enforce them upon merging.

f0rmiga:
It would indeed be easy to check for the PR title in a GH Action. Can be done separately.

f0rmiga (approved):

PR #1111, opened 2023-03-07
fix: expose ignore_root_user_error with the module extension api

stonier:
PR Instructions/requirements

Exposes one of the many args that used to be exposed via pre-module functions.

----

Q: Was there a reason that a limited argument set was made available with the module api?


rickeylev:
> Q: Was there a reason that a limited argument set was made available with the module api?

Not really. It's just a bit uncommon to need root, and the bzlmod support involved a decent amount of refactoring. It was probably just an oversight.

PR #1113, opened 2023-03-07
feat: make variable substitution for py_wheel abi, python_tag args

stonier:
As the title states - expands make variable substitution to abi and python tags.

See the example for usage.

NB: It allows variables to be passed in via toolchains (in which case it appears under `ctx.var`) or via `deps` (in which case it appears under `dep[].variables`)

### DIFF
--- a/examples/wheel/BUILD.bazel
+++ b/examples/wheel/BUILD.bazel
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 load("@bazel_skylib//rules:build_test.bzl", "build_test")
-load("//examples/wheel/private:wheel_utils.bzl", "directory_writer")
+load("//examples/wheel/private:wheel_utils.bzl", "directory_writer", "make_variable_tags")
 load("//python:defs.bzl", "py_library", "py_test")
 load("//python:packaging.bzl", "py_package", "py_wheel")
 load("//python:versions.bzl", "gen_python_config_settings")
@@ -56,6 +56,29 @@ py_wheel(
     distribution = "example_minimal_library",
     python_tag = "py3",
     version = "0.0.1",
+    deps = [
+        "//examples/wheel/lib:module_with_data",
+        "//examples/wheel/lib:simple_module",
+    ],
+)
+
+# Populate a rule with "Make Variable" arguments for
+# abi, python_tag and version. You might want to do this
+# for the following use cases:
+#  - abi, python_tag: introspect a toolchain to map to appropriate cpython tags
+#  - version: populate given this or a dependent module's version
+make_variable_tags(
+    name = "make_variable_tags",
+)
+
+py_wheel(
+    name = "minimal_with_py_library_with_make_variables",
+    testonly = True,
+    abi = "$(ABI)",
+    distribution = "example_minimal_library",
+    python_tag = "$(PYTHON_TAG)",
+    toolchains = ["//examples/wheel:make_variable_tags"],
+    version = "$(VERSION)",
     deps = [
         "//examples/wheel/lib:module_with_data",
         "//examples/wheel/lib:simple_module",
--- a/examples/wheel/private/wheel_utils.bzl
+++ b/examples/wheel/private/wheel_utils.bzl
@@ -56,6 +56,10 @@ directory_writer = rule(
 )
 
 def _make_variable_tags_impl(ctx):  # buildifier: disable=unused-variable
+    # This example is contrived. In a real usage, this rule would
+    # look at flags or dependencies to determine what values to use.
+    # If all you're doing is setting constant values, then you can simply
+    # set them in the py_wheel() call.
     vars = {}
     vars["ABI"] = "cp38"
     vars["PYTHON_TAG"] = "cp38"
--- a/examples/wheel/private/wheel_utils.bzl
+++ b/examples/wheel/private/wheel_utils.bzl
@@ -54,3 +54,16 @@ directory_writer = rule(
         ),
     },
 )
+
+def _make_variable_tags_impl(ctx):  # buildifier: disable=unused-variable
+    vars = {}
+    vars["ABI"] = "cp38"
+    vars["PYTHON_TAG"] = "cp38"
+    vars["VERSION"] = "0.99.0"
+    return [platform_common.TemplateVariableInfo(vars)]
+
+make_variable_tags = rule(
+    attrs = {},
+    doc = """Make variable tags to pass to a py_wheel rule.""",
+    implementation = _make_variable_tags_impl,
+)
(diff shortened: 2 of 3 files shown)


stonier:
@alexeagle as discussed in slack.

python/private/py_wheel.bzl
```diff
@@ -207,12 +207,23 @@ def _input_file_to_arg(input_file):
     return "%s;%s" % (py_package_lib.path_inside_wheel(input_file), input_file.path)
 
 def _py_wheel_impl(ctx):
-    version = _replace_make_variables(ctx.attr.version, ctx)
+    # Make Variable Substitutions
+    variable_index = {}
+    variable_index.update(ctx.var)
+    for dep in ctx.attr.deps:
```
rickeylev, line 213: I don't think you need special logic to mixin the TemplateVariableInfo values. Just pass the `toolchains` arg. See the "toolchains" attribute here: https://bazel.build/reference/be/common-definitions#common-attributes

rickeylev (changes requested):

python/private/py_wheel.bzl
stonier: Aye, I tried pretty hard to do that as that's what [Make Variables](https://bazel.build/reference/be/make-variables) and suggests should work, but just came away with a sore head. If you just try passing in a simple rule like `make_variable_tags` in the examples here to the toolchains argument, then it complains vociferously about it not being a `ToolchainType`.

Would be happy to find out I'm doing something unintelligible though.

stonier: Ach, just tested it on this simple example and it works. I'll go back and test it on the more complex examples we have in-house, see if I can bissect where it goes wrong.

stonier: Looks like it's working as a toolchain arg for this simple PR though. Let me go back and see if I can reproduce / bissect the problem on our more complex example.

stonier: Unable to reproduce. Updated the PR to use and exemplify from passing in a custom rule via the toolchains argument.

examples/wheel/BUILD.bazel
```diff
@@ -62,6 +62,29 @@ py_wheel(
...
+    name = "make_variable_tags",
+)
+
+py_wheel(
+    name = "minimal_with_py_library_with_make_variables",
+    testonly = True,
+    abi = "$(ABI)",
+    distribution = "example_minimal_library",
+    python_tag = "$(PYTHON_TAG)",
+    toolchains = ["//examples/wheel:make_variable_tags"],
```
alexeagle, line 80: nit: this seems like a strange example to me, would you actually expect users to write a custom rule and understand toolchains just to set these?

https://bazel.build/reference/be/make-variables#custom_variables
> Rule and [toolchain](https://bazel.build/docs/toolchains) writers can define...

I think that's referring to us, so if users need a thing spelled `$(ABI)` I'd expect something in rules_python to provide it.

rickeylev: Yeah, I agree it'd be best if the rules just told us this directly and users didn't have to write a custom rule. That said...

We could factor out a helper rule so a user doesn't need to write something a custom rule and use TemplateVariableInfo themselves, e.g.:
```
py_wheel(..., toolchains=[":vars"])
py_wheel_vars(name="vars", python_tag="py3", abi="abi3", platform="any")
```

But I'm not sure what value that really has -- it'd be easier to just set those values in the py_wheel() call. If you want to customize how it determines e.g. abi, then you're back to writing a custom rule.


Second, the more I learn about these tags, the more complicated it is to know the right value. We can get most of the way there for most users, though, I think.

PyInfo.uses_shared_libraries is probably the biggest indicator -- if there are no shared libraries, we can probably just set python_tag=py3, abi=none, platform=any as the defaults. This logic would have to live in py_wheel (more specifically, something that consumes the libraries to get that info).

If there are shared libraries, then it gets complicated.

For python tag, we need to use the major.minor python version. We can know this by look at the multi-python flags or the --python3_version flag.

For abi, we need the toolchain to tell us the abi value somehow. Normally this comes from sysconfig. A hack would be to call it in an action; this would work as long as host==target.

For platform tag, this one is complicated from what I'm told. I think a simple mapping of e.g. (linux, x86) -> "linux_x86_64" would go a long way (we even suggest such in the docs). I don't know how we'd know to use "manylinux", though -- from what I'm told, it's a specific docker image with specific versions of things used. It really sounds like it'd be modeled as a separate platform in Bazel-land with its own cc toolchain etc.


Some refs I found:
 * [pep 427](https://peps.python.org/pep-0427/#file-name-convention)
 * [pep 425](https://peps.python.org/pep-0425) (seems to talk about abi tag more)
 * [Py PA docs](https://packaging.python.org/en/latest/specifications/platform-compatibility-tags/)

stonier: I don't believe you can infer the abi/python_tag from the toolchain that works for all use cases. It depends on how and what you are packaging into the wheel.

For instance, we're using pybind11 to create bindings around c++ libraries. These bindings are major.minor locked since pybind11 uses features outside the stable C API. Build a wheel with python3.8 and you can't use it with python3.10. End result, we need to set `cp38` on wheels that package our bindings.

There is work going on to improve that state of affairs (e.g. https://github.com/python/cpython/pull/93012), so at some point, we will need to use, e.g. `cp38` for older versions and perhaps `abi3` when built with some later version of py3 where pybind11 can guarantee c python stability. In this case you have no way of being able to infer the correct abi tag from the toolchain without knowing what is going on with pybind11.

Internally, we're using a custom rule to introspect the currently resolved toolchain to determine an `abi` tag that is appropriate for our pybind11 wheels. That then gets fed into `py_wheel`. With this pipeline, we can build python versions for multiple python toolchains without having to hardcode any toolchain information into the bazel BUILD files. You just specify your py3XY toolchain on the command line (via `--extra-toolchains`) and the rest just works.

examples/wheel/private/wheel_utils.bzl
```diff
@@ -54,3 +54,16 @@ directory_writer = rule(
         ),
     },
 )
+
+def _make_variable_tags_impl(ctx):  # buildifier: disable=unused-variable
+    vars = {}
+    vars["ABI"] = "cp38"
```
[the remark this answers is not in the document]
stonier: See above.

examples/wheel/BUILD.bazel
stonier: NB: It'd be super useful if python toolchains exposed their major.minor.patch version somehow.

stonier: > Second, the more I learn about these tags, the more complicated it is to know the right value. We can get most of the way there for most users, though, I think.

Aye, I think the non-shared library case could get automagic handling. Problems in the shared library cases.

rickeylev (approved):

Issue #1069, opened 2023-02-14
Reimplement rules in Starlark in rules_python


rickeylev:
This is the general tracking issue for re-implementing the Bazel-bundled Python rules in Starlark in the rules_python repo.

The Bazel-side issue is https://github.com/bazelbuild/bazel/issues/15897

As of now (Feb 2023), the Bazel bundled rules (not providers or flags, though) have all been re-implemented in Starlark. They will used in Bazel 7. Hooks for low-level Bazel behaviors the Python rules need are exposed to the rules_python repo. So now this is mostly a matter of copying that code over here, cleaning it up, and flushing out any missing hooks.

TODO
* [x] Copy Bazel code over to rules_python
* [ ] Re-implement Bazel tests in rules_python
* [ ] Re-implement flags in Starlark
* [x] Setup a CI config that uses e.g. bazel=rolling to get access to pre-release bazel versions
* [x] Enable pystar by default

That's the high level gist, some lower level TODOs are:
* [x] Implement compatibility shim so rules_python and Bazel rules can interoperate to some degree
* [ ] Deprecate and remove Bazel-bundled flags
* [ ] Deprecate and remove Python-related parts in `@bazel_tools`

An optional bonus thing is to backport the low-level hooks to 5.x and/or 6.x. This makes them available sooner, so that we don't have to wait for Bazel 7. There isn't a _ton_ to backport (I intentionally made them to be amenable to this idea), but there's several small pieces. The two general pieces are exposing the flags as fragments and the `py_internal` object:
* `py_internal` object and it's backing pieces:
  * [PyBuiltins.java](https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/rules/python/PyBuiltins.java)
  * [BazelPyBuiltins.java](https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/bazel/rules/python/BazelPyBuiltins.java)
  * [`py_internal` name declaration in BazelRuleClassProvider.java](https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/bazel/rules/BazelRuleClassProvider.java#L468-L471)
  * [`py_internal` name override in exports.bzl](https://github.com/bazelbuild/bazel/blob/master/src/main/starlark/builtins_bzl/common/exports.bzl#L45)
[...]

PR #1122, opened 2023-03-09
feat: bzl file per rule/provider

rickeylev:
This is basically a performance optimization. Bazel can only see that a bzl file changed, not the particular contents. This means that any downstream bzl file loading it is invalidated, even if it doesn't load any of the affected code. As an example, if a package only loads `py_library.bzl`, then changing `py_test.bzl` doesn't need to invalidate all libraries.

* Also removes some more extraneous license comments

Work towards #1069

### DIFF
--- a/python/py_binary.bzl
+++ b/python/py_binary.bzl
@@ -1,0 +1,19 @@
+# Copyright 2023 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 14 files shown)


python/py_binary.bzl
```diff
@@ -0,0 +1,19 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Public entry point for py_binary."""
+
+load("//python/private:reexports.bzl", _py_binary = "py_binary")
```
Solumin, line 17: So this is a re-re-export?

Solumin (approved):

python/py_binary.bzl
rickeylev: Hah, yes. And if I change defs.bzl to export it it'll be a re-re-re-export...

Issue #1105, opened 2023-03-01
pip_parse does not respect requirement markers


arrdem:
# 🐞 bug report

### Affected Rule
The issue is caused by the rule: `pip_parse` (all versions)

### Is this a regression?

It appears to be. Previously, rules_python did evaluate platform markers see https://github.com/nikhaldi/rules_python/commit/323e84e699bdfac0fa785b03e869f7101c16acac

In 0.16, `python/pip_install/extract_wheels/parse_requirements_to_bzl.py` will gladly parse a requirements file which uses markers, but does nothing with them. Nor does the new Starlark parser introduced in ba69aec9.

It appears this functionality was simply never implemented in the new parser, and was deleted in 37e7e68e when the legacy `pip_import` implementation was removed.

### Description

`rules_python` makes no attempt to evaluate requirement marker expressions, which leads to false build failures due to attempting to install packages marked as not viable because `pip wheel` succeeds with a no-op and an error message when presented with a requirement which will be ignored due to markers.

## 🔬 Minimal Reproduction

When handed a `requirements.txt` file which uses platform markers to select a different version of an artifact depending on markers everything appears okay. Consider

```
matplotlib==3.4.3 ; sys_platform != "darwin" or platform_machine != "arm64"
matplotlib==3.4.3.post1+abnormal ; sys_platform == "darwin" and platform_machine == "arm64"
```

The correct behavior for this requirements file should be to choose between these two versions. Note that the boolean selector condition is disjoint. However the actual behavior is to unconditionally attempt to install the first requirement clause listed in the file. This breaks both `bazel fetch //...` and any rules referencing the `requirement("matplotlib")`.

The behavior should be to evaluate all the marker expressions and select matching versions/args.

It may be valid for no versions of a package to be selected. Consider `tensorflow-macos==2.7.0 ; sys_platform == "darwin"`. Arguably this should be achieved with platform-specific lockfiles and `requirements_darwin`, but it shouldn't be an error.

If multiple requirements are selected, that should likely be an error.

## 🔥 Exception or Error

```
ERROR: $REPO/BUILD:1:11: $TARGET depends on @pypi_py38_darwin_arm64_matplotlib//:pkg in repository @pypi_py38_darwin_arm64_matplotlib which failed to fetch. no such package '@pypi_py38_darwin_arm64_matplotlib//': whl_library pypi_py38_darwin_arm64_matplotlib failed:
[...]

PR #1123, opened 2023-03-10
fix: update correct requirements lock file when using os specific lock files

Rasrack:
Currently the dependency_resolver.py ignores that you give requirement lock files for different os's, except when checking if the golden file needs updating. This causes dependecy_resolver.py to update the wrong lock i.e the non platform specific one if ran in "update mode".


tests/compile_pip_requirements/BUILD.bazel
```diff
@@ -32,3 +32,33 @@ compile_pip_requirements(
...
+
+genrule(
+    name = "generate_os_specific_requirements_in",
+    srcs = [],
+    outs = ["requirements_os_specific.in"],
+    cmd = """
+cat > $@ <<EOF
+pip==22.3.0 ; sys_platform == "linux"
+pip==22.2.2 ; sys_platform == "darwin"
+pip==22.2.1 ; sys_platform == "win32"
```
aignas, line 44: nit: currently `pip_parse` does not support platform specifiers as described in #1105 and the output os specific requirements files will not be parseable by `pip_parse`.

However, since this is a test and not an example, it is fine.

python/pip_install/tools/dependency_resolver/dependency_resolver.py
```diff
@@ -88,6 +88,10 @@ def _select_golden_requirements_file(
     requirements_in_path = Path(requirements_in)
     resolved_requirements_in = str(requirements_in_path.resolve())
 
+    requirements_file = _select_golden_requirements_file(
```
aignas, line 91: Thanks for adding this functionality. This allows users to not need to run the `[[uname -r == foo ]] && cp requirements.txt requirements_foo.txt` themselves.

aignas (approved):

chrislovecnm:
@aignas can you review? You are more familiar with this part of the code base

chrislovecnm:
@Rasrack can you rebase?

python/pip_install/tools/dependency_resolver/dependency_resolver.py
```diff
@@ -88,6 +88,10 @@ def _select_golden_requirements_file(
...
+        requirements_txt, requirements_linux, requirements_darwin, requirements_windows
```
aignas, line 92: nit: I know that this function was created before this PR, but it would be great if we used keyword arguments here for readability and correctness.

It would also help if we return `pathlib.Path` type instance.

python/pip_install/tools/dependency_resolver/dependency_resolver.py
```diff
@@ -123,28 +127,28 @@ def _select_golden_requirements_file(
 
     sys.argv.append("--generate-hashes")
     sys.argv.append("--output-file")
-    sys.argv.append(requirements_txt if UPDATE else requirements_out)
+    sys.argv.append(requirements_file if UPDATE else requirements_out)
     sys.argv.append(
         requirements_in if requirements_in_path.exists() else resolved_requirements_in
     )
 
     if UPDATE:
```
aignas, line 135: Thinking about the broader picture, we could have the `UPDATE` being handled by a separate script and use the approach elsewhere where:
1. There is a rule that generates the pinned `requirements_file`.
2. There is a `diff_test` rule, which checks the source version with the output of 1.
3. There is a rule that updates the source version.

Right now we use the same script for everything, which allows us to not use `bazel-skylib` for this functionality, but now we depend on it more heavily and thus we could refactor this code as well.

Whilst this suggestion is likely to be outside the scope of the PR, I would love to understand if the platform resolution complexity could be pushed to starlark and a bunch of `select` statements in the right places.

aignas (approved):
LGTM in general.

python/pip_install/tools/dependency_resolver/dependency_resolver.py
Rasrack: It almost sound like you are aiming for how it is done in the pip_parse vendored example. To me it sounds like a sensible change.

python/pip_install/tools/dependency_resolver/dependency_resolver.py
Rasrack: I at least added the keyword arguments but will skip returning pathlib.Path for now.

aignas: LGTM

rickeylev (approved):

PR #1124, opened 2023-03-10
fix: compile_pip_requirements test from external repositories

Rasrack:
Previously when running the compile_pip_requirements test from an external repository the test failed. This was due to the fact that paths in the annotations of the lock file contained paths from the external repository. Another problem was that the requirement files could not be located.


examples/build_file_generation/gazelle_python.yaml
```diff
@@ -115,4 +115,4 @@ manifest:
   pip_repository:
     name: pip
     use_pip_repository_aliases: true
-integrity: 85f073e37e31339508aaaf5e0d5472adae5148fd5f054e9cc586343c026660e1
+integrity: 030d6d99b56c32d6577e616b617260d0a93588af791269162e43391a5a4fa576
```
aignas, line 118: This means that the `integrity` will change when users upgrade to a new `rules_python` version and recompile their requirements. @f0rmiga, what do you think about this? I am wondering if this is the right behaviour to need to regenerate the `gazelle_python.yaml` manifest when the `requirements.txt` changes but the contents themselves do not.

chrislovecnm: @Rasrack, I agree with @aignas. Can we do this without forcing an upgrade?  Are you able to follow his recommendation?

Rasrack: @chrislovecnm I don't think it's possible as the integrity will change as the annotations in the lock file change. However, my opinion is that the lock file should be generated without annotations but I think that is a separate issue.

f0rmiga: The integrity doesn't change on every rules_python upgrade. I made it so that only when the source code of the modules mapping, or the requirements.in change.

f0rmiga (approved):

PR #1131, opened 2023-03-14
fix: Fix compile_pip_requirements for our use case

blorente:
## PR Checklist

## PR Type

- [X] Bugfix
- [X] Build related changes

`compile_pip_requirements`:
- Forced the `no-remote-exec` tag on all its users, conflicting with our current toolchain + RBE setup.
- Use `shutil.copy` to copy files around. While this works for the current script, it broke in our remote execution enviornment, since it tries to open the destination file for writing.

## What is the new behavior?

- Make the tags optional. I chose to implement them as an overridable default because it more closer resembles the current implementation, but I think there is latitude to make some required (e.g. `requires-network`) and some optional (`no-remote-exec`).
- `shutil.move` can move files across namespaces, and preserves the semantics of `os.replace` better.

- [ ] Yes
- [X] No

### DIFF
--- a/python/pip_install/requirements.bzl
+++ b/python/pip_install/requirements.bzl
@@ -17,6 +17,8 @@
 load("//python:defs.bzl", _py_binary = "py_binary", _py_test = "py_test")
 load("//python/pip_install:repositories.bzl", "requirement")
 
+_DEFAULT_TAGS = ["no-sandbox", "no-remote-exec", "requires-network"]
+
 def compile_pip_requirements(
         name,
         extra_args = [],
@@ -29,7 +31,7 @@ def compile_pip_requirements(
         requirements_linux = None,
         requirements_windows = None,
         visibility = ["//visibility:private"],
-        tags = None,
+        tags = _DEFAULT_TAGS,
         **kwargs):
     """Generates targets for managing pip dependencies with pip-compile.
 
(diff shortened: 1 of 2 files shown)


aignas:
It looks like you still need to update the docs via `bazel run //docs:update` to fix the build failures.

hrfuller:
The change seems benign, but looking closer those tags seem to only apply to the generated `_test` target, unless the binary is used as a tool in an action elsewhere, so maybe those can be added to the test tags only. @f0rmiga do you remember if the no-sandbox and no-remote-exec tags were needed for the updater binary?

f0rmiga:
I'll have to dig the history to remember the details.

python/pip_install/tools/dependency_resolver/dependency_resolver.py
```diff
@@ -23,9 +23,9 @@
 import piptools.writer as piptools_writer
 from piptools.scripts.compile import cli
 
-# Replace the os.replace function with shutil.copy to work around os.replace not being able to
+# Replace the os.replace function with shutil.move to work around os.replace not being able to
 # replace or move files across filesystems.
-os.replace = shutil.copy
+os.replace = shutil.move
```
chrislovecnm, line 28: Dumb question.  Do we have a use-case where we would need the files you moved?

python/pip_install/requirements.bzl
```diff
@@ -29,7 +31,7 @@ def compile_pip_requirements(
         requirements_linux = None,
         requirements_windows = None,
         visibility = ["//visibility:private"],
-        tags = None,
+        tags = _DEFAULT_TAGS,
```
chrislovecnm, line 34: Adding defaults makes sense if we always need defaults, but I wonder if we always do. I would like to see documentation on our defaults and how to override them.  Is there a creative way to do this without adding defaults?

chrislovecnm (changes requested):

PR #1133, opened 2023-03-16
fix: Include filename when parsing imports for gazelle

jlaxson:
If the import scanner stumbles across a file with a syntax error, gazelle dies with a message that fails to specify what file it choked on.  Fix that.

Before: 
```
  File "<unknown>", line 28
    class FilesystemDatasetProvider:
IndentationError: expected an indented block
"""
```

After:
```
  File "libs/remote/python/filesystem_dataset.py", line 28
    class FilesystemDatasetProvider:
IndentationError: expected an indented block
"""
```

### DIFF
--- a/gazelle/python/parse.py
+++ b/gazelle/python/parse.py
@@ -27,7 +27,7 @@ from tokenize import COMMENT, tokenize
 
 def parse_import_statements(content, filepath):
     modules = list()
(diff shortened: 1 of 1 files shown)


f0rmiga (approved):
Awesome, thanks!

Issue #964, opened 2023-01-02
Allow multiple base URLs in python_register_toolchains


jiawen:
# 🚀 Allow multiple base URLs in python_register_toolchains

### Relevant Rules

`python_register_toolchains`

### Description

Currently, `python_register_toolchains` has a convenient `tool_versions` dictionary argument that the caller customize the mapping between Python versions and URLs from which to retrieve a build. However, the mapping is 1-to-1: a given Python version maps to a **unique** URL.

In my company's deployment, we have a monorepo that runs in multiple "environments" with strict firewalling rules. We are currently using a variant of [thundergolfer's solution](https://thundergolfer.com/bazel/python/2021/06/25/a-basic-python-bazel-toolchain/), passing a **list** of URLs for each Python version. The sha256 hash is the same but points to multiple URLs and `repository_ctx.download_and_extract` works "automatically": it tries to download from the first URL (say corp) and if it fails, tries to download from the next (say, public GitHub).

### Describe the solution you'd like

Change the type of `tool_versions` (or introduce a new attribute) from `str -> url` to `str -> list[url]`.

This is the simplest solution - Bazel will simply download each in turn and is a simple failure mechanism (to handle servers that are down).

### Describe alternatives you've considered

Introducing additional configurability at the `WORKSPACE` level seems too complex. One design could be a different Python "version" for each environment, e.g., `3.9-corp` and `3.9-github`. But this is difficult to manage as both need to be declared in `WORKSPACE` and then pushed down via a `select()` on the `requirements` at each leaf in `BUILD` files.

lummax:
related: https://github.com/bazelbuild/rules_python/pull/973

rickeylev:
Fixed by #1145

PR #1145, opened 2023-04-02
fix: Support specifying multiple download URLs in tool_versions.

quval:
The interface of `repository_ctx.download` and `repository_ctx.download_and_extract` supports string lists as well as strings as the value of the `url` argument. This is the ultimate destination of the `url` attribute in the `tool_versions` dictionary, so it makes sense for it to support lists as well.


rickeylev:
This looks reasonable. Can you describe the use case for this?

rickeylev (approved):

jiawen:
I think this fixes #964.

lpulley:
This change seems to have caused this DEBUG message for me:

```
DEBUG: Rule 'python_3_8_16_x86_64-unknown-linux-gnu' indicated that a canonical reproducible form can be obtained by modifying arguments url = ["<my python url>"] and dropping ["urls"]
```

which is fixed (after some digging and finding this PR) by changing my `tool_versions` `url` from a string to a singleton list.

However, this was a bit tricky to figure out, and though I don't exactly understand *why* this change would cause that DEBUG message, it seems that it must be the culprit. @quval @rickeylev is there an obvious fix that can be made here? I would think both a string URL and a singleton list would be equally canonical/reproducible.

PR #1077, opened 2023-02-17
feat: add bzlmod support for gazelle plugin

aignas:
## PR Checklist

Please check if your PR fulfills the following requirements:

- [X] Tests for the changes have been added (for bug fixes / features)
- [X] Docs have been added / updated (for bug fixes / features)

## PR Type

What kind of change does this PR introduce?

- [ ] Bugfix
- [X] Feature (please, look at the "Scope of the project" section in the README.md file)
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] Other... Please describe:

## What is the current behavior?

Fixes #965.

Gazelle plugin cannot work with `--experimental_enable_bzlmod` flag because of:
* missing `MODULE.bazel` file
* the labels that it generates in `BUILD.bazel` files point to repositories that are not visible due to repository remapping.

## What is the new behavior?

The new behaviour adds a flag to generate user friendly `alias` labels for third party libraries which follow a convention (where `<pip>` stands for the name attribute value passed to `pip_parse` and `<distro>` stands to the sanitised distribution name of the pip package, for example `python_multipart` for the `python-multipart` package on PyPI):
* `@<pip>//<distro>` -> `@<pip>_<distro>//:pkg`
* `@<pip>//<distro>:pkg` -> `@<pip>_<distro>//:pkg`
* `@<pip>//<distro>:whl` -> `@<pip>_<distro>//:whl`
* `@<pip>//<distro>:data` -> `@<pip>_<distro>//:data`
* `@<pip>//<distro>:dist_info` -> `@<pip>_<distro>//:dist_info`

This is gated by the `incompatible_generate_aliases` parameter to `pip_repository` rule and the `pip_parse` bzlmod extension.

In order to tell gazelle to use these aliases, one needs to pass `use_pip_repository_aliases = True` to `gazelle_python_manifest` rule, which will store the value in the `gazelle_python.yaml` manifest after regeneration. This will then instruct `gazelle` to use a different label scheme for third party dependencies.

Summary:
- feat: add optional pkg alias generation to pip_repository
- feat: support using aliases in gazelle
- doc: migrate gazelle example to use user friendly aliases
- feat: gazelle supports bzlmod
- chore: update gazelle plugin deps
- chore: bazel run //:gazelle
- fix: remove deps that are imported as bazel deps
- ci: add build_file_generation to show that we support bzlmod

[...]

PR #1146, opened 2023-04-03
fix: correct the labels returned by all_requirements lists

aignas:
This apparently was not working to begin with, but the CI running the
example did not catch it because we did not have an empty `WORKSPACE.bzlmod`
file.

Tested with the CI and with a clean cache on the local laptop.

### DIFF
--- a/examples/build_file_generation/WORKSPACE.bzlmod
+++ b/examples/build_file_generation/WORKSPACE.bzlmod
@@ -1,0 +1,2 @@
+# This file will be used when bzlmod is enabled, keep it empty
+# to ensure that all of the setup is done in MODULE.bazel
--- a/python/pip_install/pip_repository.bzl
+++ b/python/pip_install/pip_repository.bzl
@@ -370,11 +370,11 @@ def _pip_repository_bzlmod_impl(rctx):
     rctx.file("BUILD.bazel", build_contents)
     rctx.template("requirements.bzl", rctx.attr._template, substitutions = {
         "%%ALL_REQUIREMENTS%%": _format_repr_list([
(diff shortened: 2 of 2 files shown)


aignas:
cc @f0rmiga as he may be the right person to review this.

chrislovecnm:
@aignas can you run buildifier?

aignas:
@chrislovecnm, I have, the build is broken for all of the recent PRs. @rickeylev, do you have any idea why?

aignas:
I cannot reproduce the failure after reverting and it is passing even locally, this definitely needs more investigation

aignas:
OK, it seems that we **have** to have an empty `WORKSPACE.bzlmod` in all examples that support `bzlmod` or we are running a risk of leaking old style dependencies to `bzlmod` setups.

chrislovecnm:
@f0rmiga or @hrfuller PTAL . I have work on the examples that needs this PR

alexeagle:
@aignas you introduced the `@@` in https://github.com/bazelbuild/rules_python/pull/1077/files - was there a mistake in that PR, or has something changed in the meantiime?

In other rulesets I work on, we needed the `@@` to get the canonical rather than apparent module name under bzlmod. for example https://github.com/aspect-build/rules_js/blob/8dc6d52c1224775c75c5cba2524837ca66b3b33b/npm/private/npm_translate_lock_generate.bzl#L913-L916
So I'm suspicious if this is correct. I see this PR is green, so maybe I should trust that there's enough testing.

Also to detect the `--enable_bzlmod` flag you can start from https://github.com/aspect-build/bazel-lib/blob/main/docs/utils.md#is_bzlmod_enabled by the way

f0rmiga (approved):

PR #1147, opened 2023-04-03
refactor: Updating example to only use bzlmod and updates to example

chrislovecnm:
This commit updates the gazelle example to only use bzlmod.  The example was previously set up to support bzlmod and also pre-bzlmod configurations, but moving forward I recommend just using bzlmod.  We also missed a change in gazelle because bzlmod was picking up elements from the WORKSPACE file.

This commit removes the CI for the legacy bazel support as well.

I have updated one of the Python unit test names as well.  Gazelle supports unit tests named __test__.py and also unit tests that are named the same as the package name and prefixed or suffixed with the word "test".  This example now includes two of the 
three naming schemes.

Another commit in this PR contains the code in https://github.com/bazelbuild/rules_python/pull/1146. That commit will fall
off once that PR is merged.


aignas:
Please rebase on #1146 as it has the empty `WORKSPACE.bzlmod` file. Given that we have fixed the example to be correct and work under `bzlmod` as expected should we still remove the old setup? There could be users that are still using legacy `WORKSPACE` way of including deps and using `bzlmod` may not be feasible for them yet.

chrislovecnm:
@aignas i would like to remove the WORKSPACE file contents. The readme refers to GitHub tag that contains a working version not using bzlmod. I am using this code for a tutorial and leaving the WORKSPACE file will make it confusing.

aignas:
LGTM to

chrislovecnm:
@aignas that is a fair point about the testing.  I will modify the bzlmod example to include gazelle.  I don't want to create yet another example that we are maintaining.

aignas:
@chrislovecnm, SGTM. Having the `bzlmod` example contain gazelle support is probably the better alternative of the two.

chrislovecnm:
Closing because we have #1155.

PR #1155, opened 2023-04-05
feat(bzlmod): expose platform-agnostic repo target for toolchain interpreter

chrislovecnm:
This exposes a new repo and target, `@{name}_host_interpreter//:python`, created by `python.toolchain()`, that points to the host OS's Python interpreter for that particular toolchain.

This solves two problems:
 1. `pip.parse()` can now refer to the same interpreter used in the toolchains
 2. There is now a canonical, public, way to refer to the host OS Python interpreter
    for repository rules.

The above were _sort of_ possible for users to do already, but it required them to write much more configuration and extension code to do so. This moves that sort of boilerplate into our code so they have a simpler configuration.

Also:
- removing bzlmod support in the build_file_generation example; making examples work
  with both WORKSPACE and MODULE is a pain, hence splitting them.
- adding an example of bzlmod and gazelle
- improved documentation in the pip arguments

Closes: https://github.com/bazelbuild/rules_python/issues/1161


f0rmiga (dismissed):

examples/build_file_generation/MODULE.bazel
rickeylev, line 1: You also need to delete the WORKSPACE.bzlmod file, too.

examples/bzlmod/MODULE.bazel
```diff
@@ -1,36 +1,78 @@
...
     name = "example_bzlmod",
     version = "0.0.0",
     compatibility_level = 1,
 )
 
+# The following stanza defines the dependency rules_python.
+# For typical setups you set the version.
 bazel_dep(name = "rules_python", version = "0.0.0")
+
+# The following starlark loads rules_python from the file system.
```
rickeylev, line 16:
```suggestion
# The following loads rules_python from the file system.
```

examples/bzlmod/MODULE.bazel
```diff
@@ -1,36 +1,78 @@
...
+# These properties are either essential metadata of the module (such as the name and version),
+# or affect behavior of the current module and its dependents.
 module(
     name = "example_bzlmod",
     version = "0.0.0",
     compatibility_level = 1,
 )
 
+# The following stanza defines the dependency rules_python.
+# For typical setups you set the version.
```
rickeylev, line 13:
```suggestion
# For typical setups you set the version. See the releases page for available versions.
```

examples/bzlmod/MODULE.bazel
```diff
@@ -1,36 +1,78 @@
...
 bazel_dep(name = "rules_python", version = "0.0.0")
+
+# The following starlark loads rules_python from the file system.
+# For usual setups you should remove this local_path_override block.
 local_path_override(
     module_name = "rules_python",
     path = "../..",
 )
 
+# The following stanza defines the dependency rules_python.
```
rickeylev, line 23: s/rules_python/rule_python_gazelle_plugin

examples/bzlmod/MODULE.bazel
```diff
@@ -1,36 +1,78 @@
...
+bazel_dep(name = "rules_python_gazelle_plugin", version = "0.0.0")
+
+# The following starlark loads the gazelle plugingfrom the file system.
+# For usual setups you should remove this local_path_override block.
+local_path_override(
+    module_name = "rules_python_gazelle_plugin",
+    path = "../../gazelle",
+)
+
+# The following stanza defines the dependency rules_python.
```
rickeylev, line 34: rules_python -> rules python gazelle

examples/bzlmod/MODULE.bazel
```diff
@@ -1,36 +1,78 @@
...
+bazel_dep(name = "gazelle", version = "0.29.0", repo_name = "bazel_gazelle")
+
+
+# The following stanze Returns a proxy object representing a module extension;
```
rickeylev, line 38:
```suggestion
# The following stanze returns a proxy object representing a module extension;
```

examples/bzlmod/BUILD.bazel
```diff
@@ -11,6 +27,46 @@ compile_pip_requirements(
...
+# - //:gazelle_python_manifest.test is a test target ensuring that
+#   the manifest doesn't need to be updated
+gazelle_python_manifest(
+    name = "gazelle_python_manifest",
+    modules_mapping = ":modules_map",
+    pip_repository_name = "pip",
+    requirements = "//:requirements_lock.txt",
+    # NOTE: we can use this flag in order to make our setup compatible with
+    # bzlmod.
+    use_pip_repository_aliases = True,
```
aignas, line 56:
```suggestion
    use_pip_repository_aliases = True,
```

Since this is a `bzlmod` only example, we don't need this comment

examples/bzlmod/MODULE.bazel
```diff
@@ -1,36 +1,78 @@
...
+# Use the extension to call the `pip_repository` rule that invokes `pip`, with `incremental` set.
+# Accepts a locked/compiled requirements file and installs the dependencies listed within.
+# Those dependencies become available in a generated `requirements.bzl` file.
+# You can instead check this `requirements.bzl` file into your repo.
+# Because this project has different requirements for windows vs other
+# operating systems, we have requirements for each.
 pip.parse(
     name = "pip",
     requirements_lock = "//:requirements_lock.txt",
     requirements_windows = "//:requirements_windows.txt",
```
aignas, line 68: You need to add a flag here in order to generate labels compatible to gazelle:
```suggestion
    requirements_windows = "//:requirements_windows.txt",
    incompatible_generate_aliases = True,
```

examples/build_file_generation/WORKSPACE
```diff
@@ -17,22 +17,23 @@ load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
 # Define an http_archive rule that will download the below ruleset,
 # test the sha, and extract the ruleset to you local bazel cache.
+
 http_archive(
     name = "io_bazel_rules_go",
     sha256 = "099a9fb96a376ccbbb7d291ed4ecbdfd42f6bc822ab77ae6f1b5cb9e914e94fa",
```
aignas, line 23: sha256 needs to be updated

examples/bzlmod/runfiles/BUILD.bazel
```diff
@@ -1,5 +1,6 @@
 load("@rules_python//python:defs.bzl", "py_test")
 
+# gazelle:ignore
```
chrislovecnm, line 3: @aignas I am getting errors when I try to have gazelle create the target and resolve the dependencies for the `"@our_other_module//other_module/pkg:lib",`.

```
$ bazel run //:gazelle update
INFO: Analyzed target //:gazelle (39 packages loaded, 240 targets configured).
INFO: Found 1 target...
Target //:gazelle up-to-date:
  bazel-bin/gazelle-runner.bash
  bazel-bin/gazelle
INFO: Elapsed time: 0.381s, Critical Path: 0.02s
INFO: 1 process: 1 internal.
INFO: Running command line: bazel-bin/gazelle update
gazelle: ERROR: failed to validate dependencies for target "//runfiles:runfiles_test": "other_module.pkg.lib" at line 18 from "runfiles/runfiles_test.py" is an invalid dependency: possible solutions:
        3. Ignore it with a comment '# gazelle:ignore other_module.pkg.lib' in the Python file.

"other_module.pkg" at line 18 from "runfiles/runfiles_test.py" is an invalid dependency: possible solutions:
        3. Ignore it with a comment '# gazelle:ignore other_module.pkg' in the Python file.

"other_module" at line 18 from "runfiles/runfiles_test.py" is an invalid dependency: possible solutions:
        3. Ignore it with a comment '# gazelle:ignore other_module' in the Python file.
 ```

I tried using the resolve directive, but I could not get that working in the manner I assumed would work.  I'm guessing that sub-modules, at this point, are not supported.  Is that correct?

Comment (chrislovecnm) [reply examples/bzlmod/runfiles/BUILD.bazel:3]: @f0rmiga any ideas around the question above ^

Comment (chrislovecnm) [reply examples/bzlmod/runfiles/BUILD.bazel:3]: @aignas this example does not seem to download python when it is running.  It is using the hosts python binary.  If I am correct we have bazel download python by setting the `interpreter`  in pip_parse.

From the WORKSPACE file in the pre bzlmod example

```
pip_parse(
    name = "pip",
    # Generate user friendly alias labels for each dependency that we have.
    incompatible_generate_aliases = True,
    # (Optional) You can provide a python_interpreter (path) or a python_interpreter_target (a Bazel target, that
    # acts as an executable). The latter can be anything that could be used as Python interpreter. E.g.:
    # 1. Python interpreter that you compile in the build file.
    # 2. Pre-compiled python interpreter included with http_archive.
    # 3. Wrapper script, like in the autodetecting python toolchain.
    #
    # Here, we use the interpreter constant that resolves to the host interpreter from the default Python toolchain.
    # Set the location of the lock file.
)
```

What is the syntax that I use in a MODULES.bzl file to load the interpreter?  I noticed this by running this example on windows and it is complaining that it cannot find python in the path, even though python exists on the host.

**Comment by chrislovecnm** [comment]
We are having a few flakes, but the building on Windows OS is not working.  See https://github.com/bazelbuild/rules_python/issues/1163.  I have recreated the problem locally.

File: examples/bzlmod/runfiles/BUILD.bazel (hunk 9 above)
Comment (chrislovecnm) [reply examples/bzlmod/runfiles/BUILD.bazel:3]: I figured this out. The current parameters will only accept hard coded values and won’t work cross platform

File: examples/bzlmod/MODULE.bazel [hunk 10]
```diff
@@ -1,36 +1,122 @@
...
+
+# The following stanze returns a proxy object representing a module extension;
+# its methods can be invoked to create module extension tags.
 python = use_extension("@rules_python//python:extensions.bzl", "python")
+
+# This version is used in various places to load the correct
+# version of Python.
+PYTHON_VERSION = "3.9"
+
+# We next initaze the python toolchain using the extension.
```
rickeylev, line 51: typo: initialize

examples/bzlmod/MODULE.bazel
```diff
@@ -1,36 +1,122 @@
...
+    python_version = PYTHON_VERSION,
 )
-use_repo(python, "python3_9")
-use_repo(python, "python3_9_toolchains")
 
+# Import the python repositories generated by the given module extension into the scope of the current module.
+use_repo(python, "python3_toolchains")
+use_repo(python, "python3")
+
+# In bazel 6.2 you will be able to load these dynamically.
```
rickeylev, line 66: What does this mean, "load dynamically"? Is there a description of this 6.2 feature I can read?

examples/bzlmod/MODULE.bazel
```diff
@@ -1,36 +1,122 @@
...
+# We need each of these repos load in order for setting the pyton_interpreter_target.
```
rickeylev, line 67: nit: the "in order" sorta makes it sound like the order of statements matter, but I don't think that's the case, right? rephase to "so that" to avoid implying statement order matters

examples/bzlmod/MODULE.bazel
```diff
@@ -1,36 +1,122 @@
...
+# You only need the python repo for the operating systems that you are supporting.
+use_repo(python, "python3_x86_64-unknown-linux-gnu")
```
rickeylev, line 69: I think I understand why these use_repo() calls are necessary, but can we hide this from users somehow? Because its only necessary for the pip setup, can we fold it into that extension somehow?

examples/bzlmod/extensions.bzl
```diff
@@ -0,0 +1,39 @@
+"""
+This module is for setting the pyhton_interpreter_target argument
```
rickeylev, line 2: typo: Python

rickeylev, line 2: Can we move this into rules_python instead? I don't think users should have to write their own extension for this. Having to call an extra extension between getting the interpreters and setting up pip is annoying, but far better than having to write your own extension.

examples/bzlmod/extensions.bzl
```diff
@@ -0,0 +1,39 @@
...
+when we use pip.parse in MODULE.bazel.  This extension includes a
+build rule that creates a file named python.cfg.  The rules_python
```
rickeylev, line 4: Hm, writing a little config like like this seems like it might give us some other options? Because we could drop some extra info into it for the later pip repo rule to read.

But, also: why do we have to write a file? Why can't we pass in the target directly? e.g.,

```
use_repo(python, "python3_x86_linux")
pip.parse(interpreter = "@python3_x86_linux//:interpreter")
```

examples/bzlmod/extensions.bzl
```diff
@@ -0,0 +1,39 @@
...
+_symlink_python_interpreter = repository_rule(
+    implementation = _symlink_python_interpreter_impl,
+    local = True,
+)
+
+# buildifier: disable=unused-variable
+def _python_interpreter(module_ctx):
+    _symlink_python_interpreter(name = "python_interpreter")
+
+python_interpreter = module_extension(
```
rickeylev, line 34: I have to go no, but a few ideas I had from reading the bzlmod docs (I mentioned this in chat, so posterity)

1. module_ctx has a path() function that says it accepts a label. So presumably we can pass a label in and get a file path back?
2. module_tx has a modules attribute, which says it gives the whole dependency graph. Maybe we could search this for the repo-level-interpreter instead?

The bzlmod docs also give an example of collecting things to call an external dependency resolving. Which is basically what pip is doing. This seems relevant, but our usage seems fairly different. I just get the feeling like there might be some adaptation of that example we're missing.

rickeylev (changes requested):
I only got about half way through; i'll finish later today.

examples/bzlmod/extensions.bzl
rickeylev: another idea: repo mapping. Something like:

pip always looks for a "@pip_python_interpreter" repo. In the module, it's up to the user to map the appropriate repo to that name so that pip can find it. e.g. `use_repo(..., pip_python_interpreter="python3_x86_linux")`

rickeylev (changes requested):

examples/bzlmod/extensions.bzl
chrislovecnm: you can, but then you are hardcoding it for Linux

examples/bzlmod/MODULE.bazel
chrislovecnm: I don't think so.

examples/bzlmod/extensions.bzl
chrislovecnm: I don't know how to.  We depend on the python repo that we loaded with use_repo.  That is providing the interpreter.  So with the current code base, the answer is no.  If I understood the hub and spoke model more, maybe ...

examples/bzlmod/extensions.bzl
chrislovecnm: How do we get the label for it? That my question.

examples/bzlmod/extensions.bzl
```diff
@@ -0,0 +1,39 @@
+"""
+This module is for setting the pyhton_interpreter_target argument
+when we use pip.parse in MODULE.bazel.  This extension includes a
+build rule that creates a file named python.cfg.  The rules_python
+pip extension can then use that file to set the python_interpreter_target
+label dynamically.  If you do not use something like this extension
+the python_interpreter_target is hardcoded to an OS specific interpreter.
+"""
+
+load("@python3//:defs.bzl", "interpreter")
```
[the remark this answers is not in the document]
chrislovecnm: used https://github.com/bazelbuild/rules_python/blob/6ce2dbea2e05a962c393e5b3efa196630c0ae019/examples/bzlmod/extensions.bzl#L23

I don't know how to move that call into our extension.

rickeylev:
Posterity/inform other maintainers:

Alright, this is gonna be a long one. I'll try to keep it as brief as I can. Me and Chris have been talking and I spent the last couple days trying out some ideas. There's a couple ways to do this with some variations, but only one of them is appealing.

The tl;dr of that appealing option is a module config that looks like this:

```
python = use_extension("@rules_python:extension.bzl", "python")
python.toolchain(...)

use_repo(python, "python3_9_linux_x86", "python3_9_mac_arm", ...etc...)
host_python = use_extension("@rules_python:extension.bzl", "host_python")
host_python.register(interpreters=["@python3_9_linux_x86", "@python3_9_mac_x86", ...etc..])

use_repo(host_python, "host_python")
pip = use_extension("@rules_python:extension.bzl", "pip")
pip.parse(..., python_interpreter_target="@host_python//:python")
```

The new bit is the middle part, the "host_python" parts. Users will have to use this extension to "register" which interpreters to use for repository rule commands (aside: maybe rename this to "repo_python", as "host" carries meaning with host/exec config/actions).

It's a bit annoying that users will have to manually list out all the host interpreters they might want to use, but I don't see a way around that. This is because bzlmod restricts when repositories are visible to who. Specifically:
 * Within a `module_extension` execution, repos are not visible to one another. e.g. if
   an extension function does `repo(name="foo"); repo(name="bar", tool="@foo")`, then
   an error about "foo" not being defined will occur. This means it's required that
   a separate extension consume the interpreter repos generated by the "python" extension.
 * use_repo() is required to make a repository visible to the current module (Chris said
   this already, just reminding). This means it's required that all the platform-specific
   interpreter repos are in a use_repo() call.

In Bazel 6.2, there's supposed to be a feature coming to eliminate a bit of the use_repo() repetition.

Anyways, back to the host_python extension.

This extension is basically a hub to pick out the proper platform-specific interpreter. It's logic is pretty simple: loop through the interpreters list, figure out which one matches `repo_ctx.os`, and then create a symlink to the underlying interpreter (this last step must be done by a helper repository rule). All the possible choices have to be provided as input labels because otherwise it doesn't have access to the underlying files (and will give an error when trying to convert the label to its file location).

That's the high level overview. Now onto some alternatives, variations, and tricks.

----

The main alternative is to duplicate the underlying repo that matches the host. e.g., in the "python" extension, make an additional call to `python_repository()`:

```
# python module extension impl
def _python_impl(module_ctx):
  <loop that calls python_register_toolchains()>
  python_repository(
    name = "python_host_interpreter",
    <duplicate the call that python_register_toolchains made that matches ctx.os>
  )
```

The downside of this is we're duplicating the whole repo. We _have_ to duplicate it because we can't reference the repos created within the extension function. Note I haven't tried this, but see no reason why it wouldn't work.

-----

Trick: The labels don't _entirely_ matter. Note in the tldr, I only wrote the repo name. This is because the label just forms a dependency on the repo, not any specific "target" in the repo. I put target in quotes because, as far as I can tell, the labels are just syntactic; they don't have to exist or actually be backed by any BUILD file during this phase. Within the repo rule and module extension, you can arbitrarily manipulate them to get arbitrary files in the repo they point to. **Why does this matter?** It just lets us eliminate some typing in the config is all. e.g., you can give "@python3_9_linux_arm//:whatever", and then the repo rule can root around the directory looking at other files to figure out if it matches `repo_ctx.os`. In my prototype, I had each interpreter write a json file of platform info that got read back.

----

Variation: The host_python extension isn't strictly necessary. Instead, `pip.parse()` could take a list of the platform-specific interpreters, then figure out which one it should use. This works, but further complicates the pip.parse() interface (it already has 2 ways of specifying an interpreter). I also just think having an explicit way to configure a Python to use for repo-rule stuff is just going to be helpful in general. It's not like repo rules are going away, so having a clear way to register and use a Python interpreter that works on the host sounds like a good idea.

Variation: The above is _also_ true of the python extension _as long as the repo is passed in as an argument_. e.g. this appears to work:
```
python = use_extension(...)
python.toolchain(...)
use_repo(python, ["python_linux_arm", ...])
python.host_interpreter(interpreters=["@python_linux_arm", ...])
use_repo(python, "python_host_interpreter")
```

----

Variation: In my prototype, I made `python_repository()` write a config file of platform information. This could be moved to the extension calls, e.g.

```
host_python.register(os="linux", arch="arm", interpreter="@python3_9_linux_arm")
host_python.register(os="mac", arch="arm", interpreter="@python3_9_mac_arm")
```

It just turns into more typing is all. The module and repo can read files, so having the information available as args seems of limited value.

----

Future consideration: Something the bzlmod docs say is that module extensions can access the transitive usages of themselves in the external dependency graph. I _think_ what this might translate to a separate module (specified by bazel_dep) can call the extension to register interprters, and then a higher level call to the extension can see those (by iterating over the modules), and mix them into what it does. Not really sure; haven't tried this, but it's an interesting idea. It seems like the only option that actually eliminates having to repeat all the repo names, though.

groodt:
Thanks for the thorough explanation of options. 

I align with this sentiment:

```
It's not like repo rules are going away, so having a clear way to register and use a Python interpreter that works on the host sounds like a good idea.

```

chrislovecnm:
This pretty much sums it up: 

> Within a module_extension execution, repos are not visible to one another.  e.g., if
an extension function does repo(name="foo"); repo(name="bar", tool="@foo"), then
an error about "foo" not being defined will occur.  This means it's required that
a separate extension consume the interpreter repos generated by the "python" extension.

To add to what @rickeylev is saving, the challenge is that we cannot pass the interpreter value into the rules Python extension.  Users can write an extension like the following (pseudo code) in their repo.

```python
# load interpreter
load("@python3//:defs.bzl", "interpreter")

# inside a repository rule, we do this
# when the extension is located inside a users repository
def _my_repo_rule_impl(repository_ctx):
     python_bin_name = "python.exe" if not on windows
     repo_ctx.symlink(get_path_from_label(interpreter), python_bin)
```

If you write an extension inside of your project, you can get access to the `interpreter` label, but you cannot pass that label directly into the rules_python extension.   So the interpreter is already registered, but the extension cannot access it.  @rickeylev, please correct me if I am incorrect.

I will get this PR working across multiple platforms cause we have been messing with this for weeks, but we can optimize this code more internally.  I don't think we need to write JSON files to the different repositories; we have the information internally and can pass in the various Python binary labels instead.

Also, this is only needed when a module/project needs to support running on different operating systems.  If the project only runs on Linux, you set the Python interpreter target to a value like: `@python3_x86_64-unknown-linux-gnu//:bin/python3`.  When you need CI to build this program on Linux, Mac, and Windows, you have to jump through all of these hoops.

python/extensions.bzl
```diff
@@ -87,6 +89,12 @@ def _pip_impl(module_ctx):
                 incompatible_generate_aliases = attr.incompatible_generate_aliases,
             )
 
+            # if python_interpreter_cfg is set we need to read the label
```
chrislovecnm, line 92: TODO: remove

python/versions.bzl
```diff
@@ -20,6 +20,10 @@ MACOS_NAME = "mac os"
 LINUX_NAME = "linux"
 WINDOWS_NAME = "windows"
 
+# Values need to normalize os name.
+APPLE_NAME = "apple"
+UNKNOWN = "UNKNOWN"
```
chrislovecnm, line 25: TODO: I do not think I am using this anymore

chrislovecnm:
CI is passing 🎆 - @rickeylev @f0rmiga PTAL

chrislovecnm:
TODO: update commit message and also the PR documentation.

rickeylev:
> So the interpreter is already registered, but the rules_python extension cannot directly access it. This is odd IMO.

Maybe? It seems congruent with other behaviors we see (i.e. an extension can't see repos it creates within its own extension function). I get the feeling it might stem from the interaction of different behaviors. Something I noticed is if an extension calls `repo(name="foo")`, then, within the `repo()` rule, printing `repo_ctx.attr.name` prints not "foo", but e.g. "rules_python~override~foo~foo" (or something like that). It almost seems like extensions intercept the creation of repos (probably to do repo mapping or something).

----

FYI, re: repo visibility in an extensions: I discovered that an extension can see repos it creates, _but_ it requires them to be passed as inputs. e.g., this works:

```
# extension.bzl
def extension_impl(...)
  repo(name = "foo")
  repo(name = "bar", other=ctx.tag.bar.other)

python = extension(
  foo=tag_class(),
  bar=tag_class(attrs={other=attr.label()})
)

python = use_extension(...)
python.foo(...)
use_repo(python, "foo")
python.bar(other="@foo")
```

This doesn't change much, though. It just just allows changing `host_interpreter = use_extension(...); host_interpreter.register(...)` to simply `python.register(...)`.

python/pip_install/pip_repository.bzl
```diff
@@ -550,7 +550,7 @@ of a binary found on the host's `PATH` environment variable. If no value is set
 If you are using a custom python interpreter built by another repository rule,
 use this attribute to specify its BUILD target. This allows pip_repository to invoke
 pip using the same interpreter as your toolchain. If set, takes precedence over
-python_interpreter.
+python_interpreter. An example value: "@python3_x86_64-unknown-linux-gnu//:bin/python3".
```
rickeylev, line 553: nit: Just `//:python` can be used. All our python_repository rules create a symlink for that name pointing to the bin/python3 file.

examples/bzlmod/.bazelrc
```diff
@@ -1,3 +1,9 @@
+test --test_output=errors --enable_runfiles
```
rickeylev, line 1: What is --enable_runfiles needed for?

rickeylev (changes requested):

examples/bzlmod/MODULE.bazel
```diff
@@ -1,36 +1,129 @@
...
+use_repo(python, "python3_x86_64-pc-windows-msvc")
+
+# This extension is used to look up the host system interpreter.
+# A label that is created by this extension is used for the
+# python_interpreter_target.
+host_python = use_extension("@rules_python//python:extensions.bzl", "host_python")
+host_python.register(
+    # You only need the interpreters that you are supporting. In your project.
+    interpreters = [
+        "@python3_x86_64-unknown-linux-gnu//:host.cfgX",
```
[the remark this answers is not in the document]
chrislovecnm: PTAL now.  We have interpreters and repo.  Let me know what you want to name them.

python/repositories.bzl
```diff
@@ -348,6 +350,13 @@ py_runtime_pair(
     rctx.file(STANDALONE_INTERPRETER_FILENAME, "# File intentionally left blank. Indicates that this is an interpreter repo created by rules_python.")
     rctx.file("BUILD.bazel", build_content)
 
+    # Write out a host config file that allows us to match the interpreter
+    # when using bzlmod.
+    rctx.file("host.cfg", json.encode({
```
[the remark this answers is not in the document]
chrislovecnm: Removed that code.

python/extensions.bzl
```diff
@@ -127,3 +135,71 @@ pip = module_extension(
...
+
+    # We should only find one compatible interpreter. If we
+    # find more than one, or don't find it, then it is an error.
+    if len(compatible) == 1:
+        chosen_config = compatible[0]
+
+        # we always need to use python, even if we are on windows
+        repo_ctx.symlink(chosen_config["path"], "python")
+    elif len(compatible) > 1:
+        fail("more than one compatible interpreter found for host os: {}, arch: {}".format(os, arch))
```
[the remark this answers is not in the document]
chrislovecnm: Removed that code.

python/extensions.bzl
```diff
@@ -127,3 +135,71 @@ pip = module_extension(
...
+            doc = """\
+This rule find the compatible interpreter for the host and creates a repo that includes a 
+symlink to the interpreters Python binary.
+""",
+        ),
+    }),
+)
+
+def host_python_impl(modules_ctx):
+    for mod in modules_ctx.modules:
```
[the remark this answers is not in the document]
chrislovecnm: Removed it. Not applicable.

python/pip_install/pip_repository.bzl
chrislovecnm: TODO

examples/bzlmod/.bazelrc
chrislovecnm: Windows.   It is disabled by default

python/extensions.bzl
```diff
@@ -32,10 +33,28 @@ def _python_impl(module_ctx):
...
+                    mandatory = True,
+                    doc = """\
+List of labels matching the various python interperters.
+""",
+                ),
+                "repo": attr.label(
+                    mandatory = True,
+                    doc = """\
+The name of the repo name passed into the python.toolchain method. For instance "python3".
+""",
```
rickeylev, line 56: Looks unused?

python/extensions.bzl
```diff
@@ -32,10 +33,28 @@ def _python_impl(module_ctx):
...
 
 python = module_extension(
     implementation = _python_impl,
     tag_classes = {
+        "host_python_interpreter": tag_class(
+            attrs = dict({
+                "interpreters": attr.label_list(
+                    mandatory = True,
+                    doc = """\
+List of labels matching the various python interperters.
```
rickeylev, line 49:
```suggestion
List of labels matching the various python interperters. e.g. 
@python3_x86_64-unknown-linux-gnu//:python

```

python/extensions.bzl
```diff
@@ -127,3 +135,71 @@ pip = module_extension(
...
+            mandatory = True,
+            doc = """\
+This rule find the compatible interpreter for the host and creates a repo that includes a 
+symlink to the interpreters Python binary.
+""",
+        ),
+    }),
+)
+
+def host_python_impl(modules_ctx):
```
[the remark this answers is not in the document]
chrislovecnm: Done

python/extensions.bzl
```diff
@@ -127,3 +152,59 @@ pip = module_extension(
         "parse": tag_class(attrs = _pip_parse_ext_attrs()),
     },
 )
+
+def _host_hub_impl(repo_ctx):
+    repo_ctx.file("BUILD.bazel", """\
```
rickeylev, line 153: I don't think this BUILD file actually needs anything within it?

python/extensions.bzl
```diff
@@ -127,3 +135,71 @@ pip = module_extension(
...
+        chosen_config = compatible[0]
+
+        # we always need to use python, even if we are on windows
+        repo_ctx.symlink(chosen_config["path"], "python")
+    elif len(compatible) > 1:
+        fail("more than one compatible interpreter found for host os: {}, arch: {}".format(os, arch))
+    else:
+        fail("no compatible interpreter found for host norm os: {},norm arch: {},ctx os: {},ctx arch: {}".format(os, arch, repo_ctx.os.name, repo_ctx.os.arch))
+
+host_hub = repository_rule(
```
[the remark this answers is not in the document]
chrislovecnm: Done

python/extensions.bzl
```diff
@@ -127,3 +152,59 @@ pip = module_extension(
...
+
+    found = False
+    for config_label in repo_ctx.attr.interpreters:
+        if host_platform in str(config_label):
+            # We found the interpreter, so we create the symlink
+            # Regardless of the OS, we need a symlink named "Python"
+            # as we have code in the pip rules that is hardcoded on the
+            # name.
+            repo_ctx.symlink(config_label, "python")
+            found = True
```
rickeylev, line 171: Instead of taking the first one, please find all matches and check that only one was found.

python/extensions.bzl
```diff
@@ -127,3 +152,59 @@ pip = module_extension(
...
+    # the Label "interpreter".
+    (os, arch) = get_host_os_arch(repo_ctx)
+    host_platform = get_host_platform(os, arch)
+
+    found = False
+    for config_label in repo_ctx.attr.interpreters:
+        if host_platform in str(config_label):
+            # We found the interpreter, so we create the symlink
+            # Regardless of the OS, we need a symlink named "Python"
+            # as we have code in the pip rules that is hardcoded on the
```
rickeylev, line 168: What code in pip is hard coded to look for "python"

python/extensions.bzl
```diff
@@ -127,3 +152,59 @@ pip = module_extension(
...
+    # create the toolchains with, that create
+    # the Label "interpreter".
+    (os, arch) = get_host_os_arch(repo_ctx)
+    host_platform = get_host_platform(os, arch)
+
+    found = False
+    for config_label in repo_ctx.attr.interpreters:
+        if host_platform in str(config_label):
+            # We found the interpreter, so we create the symlink
+            # Regardless of the OS, we need a symlink named "Python"
```
rickeylev, line 167:
```suggestion
            # Regardless of the OS, we need a symlink named "python"
```

python/extensions.bzl
```diff
@@ -127,3 +152,59 @@ pip = module_extension(
...
+    local = True,
+    attrs = dict({
+        "interpreters": attr.label_list(
+            mandatory = True,
+            doc = """\
+This rule find the compatible interpreter for the host and creates a repo that includes a 
+symlink to the interpreters Python binary.
+""",
+        ),
+        "repo": attr.label(
```
rickeylev, line 191: Looks unused?

python/pip_install/pip_repository.bzl
chrislovecnm: Done

python/repositories.bzl
```diff
@@ -530,6 +544,18 @@ def python_register_toolchains(
                 install = not bzlmod,
             )
 
+        if host_os:
```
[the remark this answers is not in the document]
chrislovecnm: Done

python/extensions.bzl
```diff
@@ -127,3 +135,71 @@ pip = module_extension(
...
+This rule find the compatible interpreter for the host and creates a repo that includes a 
+symlink to the interpreters Python binary.
+""",
+        ),
+    }),
+)
+
+def host_python_impl(modules_ctx):
+    for mod in modules_ctx.modules:
+        for attr in mod.tags.register:
```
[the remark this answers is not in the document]
chrislovecnm: It is "attr" everywhere. Do you want me to change `_pip_impl`?

rickeylev (changes requested):

python/extensions.bzl
```diff
@@ -127,3 +135,71 @@ pip = module_extension(
...
+            str(config_label).replace(":host.cfg", ":python"),
+        )
+        content = repo_ctx.read(alt_label)
+
+    # We should only find one compatible interpreter. If we
+    # find more than one, or don't find it, then it is an error.
+    if len(compatible) == 1:
+        chosen_config = compatible[0]
+
+        # we always need to use python, even if we are on windows
```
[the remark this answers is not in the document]
chrislovecnm: We can file and issue.  I forget where it blows up.  Someone would have to track it down.

python/extensions.bzl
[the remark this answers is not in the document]
rickeylev: Just changing it locally is fine.

python/extensions.bzl
chrislovecnm: I forget :(  I know it was blowing up when I was testing using windows.

python/extensions.bzl
chrislovecnm: Yah. I used it previously.  Will remove.

python/extensions.bzl
```diff
@@ -32,10 +33,30 @@ def _python_impl(module_ctx):
                 register_coverage_tool = attr.configure_coverage_tool,
                 ignore_root_user_error = attr.ignore_root_user_error,
             )
+        for attr in mod.tags.host_python_interpreter:
+            _host_hub(name = "host_python_interpreter", interpreters = attr.interpreters, repo = attr.repo)
```
rickeylev, line 37: I think we need to add a prefix to this name after all to support multiple Python versions. Today, with workspace, you can do e.g.

```
load("@python38//:defs.bzl", py38_interpreter = "interpreter")
load("@python310/:defs.bzl", py310_interpreter = "interpreter")
pip_parse(..., interpreter = py38_interpreter)
pip_parse(..., interpreter = py310_interpreter)
```

To get the equivalent, we'll need separately addressable repos for each python version.

This might also be a point in favor of moving it all into the toolchain() tag, because that call has the name attribute with the prefix that is traditionally used.

examples/bzlmod/MODULE.bazel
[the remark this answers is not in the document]
rickeylev: If all we need is the //:python label, then lets stick with "interpreters". Typing the extra "//:python" is NBD.

My only thought about "maybe we should just pass @<platformrepo>" stems from "maybe we need to root around in the platform repo for a bunch of things?". I came across our toolchains_alias repo rule, which code gens a bunch of stuff, but it's not entirely clear to me why -- if all that code gen is used in _repo_ rules, then we need it. But if it's done later, by _regular_ rules, then we don't.

examples/bzlmod/MODULE.bazel
```diff
@@ -1,36 +1,136 @@
...
+# latest version.
+bazel_dep(name = "gazelle", version = "0.30.0", repo_name = "bazel_gazelle")
+
+# The following stanze returns a proxy object representing a module extension;
+# its methods can be invoked to create module extension tags.
 python = use_extension("@rules_python//python:extensions.bzl", "python")
+
+# This version is used in various places to load the correct
+# version of Python.
+PYTHON_VERSION = "3.9"
```
rickeylev, line 49: nit: it's actually only used once. So I'd recommend just inlining it.

python/extensions.bzl
chrislovecnm: Don't we need the export?

examples/bzlmod/MODULE.bazel
chrislovecnm: Done

python/extensions.bzl
chrislovecnm: I am setting the name.  This gives us a unique name for the interpreter.

python/extensions.bzl
chrislovecnm: Done

python/extensions.bzl
chrislovecnm: Done

python/extensions.bzl
chrislovecnm: Done

python/extensions.bzl
chrislovecnm: Removed

examples/bzlmod/MODULE.bazel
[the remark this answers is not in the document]
chrislovecnm: It is updated

examples/bzlmod/MODULE.bazel
```diff
@@ -1,36 +1,143 @@
...
+    # You only need the interpreters that you are supporting. In your project.
+    # These labels are referencing the above repos, and they all have a :python
+    # exported.
+    interpreters = [
+        "@python3_x86_64-unknown-linux-gnu//:python",
+        "@python3_aarch64-apple-darwin//:python",
+        "@python3_aarch64-unknown-linux-gnu//:python",
+        "@python3_x86_64-apple-darwin//:python",
+        "@python3_x86_64-pc-windows-msvc//:python",
+    ],
```
rickeylev, line 98: I think we can get rid of having to pass these in. I was talking to a Xudong about this problem, and he said it sounded like a bug and asked for a simple repro. [When I tried][1]...I couldn't repro it and didn't need to pass these labels in for them to resolve. I'm going to take another stab at figuring out why.

[1]: https://github.com/bazelbuild/rules_python/compare/main...rickeylev:rules_python:repo-resolve-bug?expand=1

python/extensions.bzl
rickeylev: No, because it's only used by repo operations, which don't understand the content within BUILD files.

python/pip_install/pip_repository.bzl
rickeylev: I still see :bin/python3 here?

python/extensions.bzl
```diff
@@ -19,23 +19,48 @@ load("@rules_python//python/pip_install:pip_repository.bzl", "locked_requirement
...
+                interpreters = interpret_attr.interpreters,
             )
 
 python = module_extension(
     implementation = _python_impl,
     tag_classes = {
+        # This extension finds the Python version that works on the host, the
+        # Python version was defined by the toolchain.
+        "host_python_interpreter": tag_class(
+            attrs = dict({
```
rickeylev, line 48: The "dict" call is unnecessary here

python/extensions.bzl
```diff
@@ -127,3 +152,48 @@ pip = module_extension(
...
+    for config_label in repo_ctx.attr.interpreters:
+        if host_platform in str(config_label):
+            interpreters.append(config_label)
+
+    num = len(interpreters)
+    if num == 1:
+        # We found the interpreter, so we create the symlink
+        # Regardless of the OS, we need a symlink named "python"
+        # as we have code in the pip rules that is hardcoded on the
+        # name.
```
rickeylev, line 177: Does this work with windows, or do we need to create "python.exe" ? Or is that handled by pip?

python/extensions.bzl
```diff
@@ -127,3 +152,48 @@ pip = module_extension(
...
+        repo_ctx.symlink(interpreters[0], "python")
+    elif num == 0:
+        fail("Unable to find interpreter, that matches the host os and architecture, in the argument interpreters")
```
rickeylev, line 180: Please include the ones that were considered in the error message. "None of these were deemed compatible: <list of the labels>".

python/extensions.bzl
```diff
@@ -127,3 +152,48 @@ pip = module_extension(
...
+    else:
+        fail("Found more than one interpreter, that matches the host os and architecture, in the argument interpreters")
+
+# This repository rule finds the host's Python interpreter and creates a symlink to it.
+# We use this rule to set the pips interpreter target when using different Operating
+# Systems with the same project.
+_host_hub = repository_rule(
+    implementation = _host_hub_impl,
+    local = True,
+    attrs = dict({
```
rickeylev, line 190: Remove extraneous dict() call

rickeylev (changes requested):

python/extensions.bzl
chrislovecnm: Works fine on windows

python/extensions.bzl
chrislovecnm: >  BUILD file not found in directory '' of external repository @rules_python~override~python~python3_host_interpreter. Add a BUILD file to a directory to mark it as a package.

So we need it :)

examples/bzlmod/MODULE.bazel
chrislovecnm: How do we know what to check if we don't pass them in?

aignas:
I am a little worried that this PR includes multiple changes:
* Remove `bzlmod` from `build_file_generation`.
* Include `gazelle` in `bzlmod` example.
* Dynamic interpreter target.
* Improving `bzlmod` example to make it easier to understand/follow to newcomers.

When uploading to BCR, we ought to have a test module and for `rules_python` it was `examples/bzlmod`, which should not include `gazelle` in it, because it checks that `rules_python` can be used without it. In https://github.com/bazelbuild/bazel-central-registry/pull/583 I used `build_file_generation` example, which is the only test module that we have with `gazelle` and `rules_python`. Having separate test modules for both would be go good to retain.

Even though it may be easier for users to understand how to use `rules_python` when they look at a single example, we may be complicating the release process. @rickeylev, what's your opinion on this?

python/extensions.bzl
chrislovecnm: Also, I think we need the same name for the target in Windows and Linux-like operating systems to have this dynamic across operating systems.  Unless we pass in the repo name and not the actual label name to the symlink.  We would have a label ending in `:python` vs. a label ending in `:python.exe`.  I think this is why we have just `python`.

rickeylev:
@aignas has some good points.

Let's do this, then: 

Take gazelle out of the bzlmod example because, yeah, it's important that our BCR test can verify gazelle isn't required.

Ah oops -- tbh, it wasn't obvious to me that the build_file_generation example was about gazelle integration -- I just figured it was a regression test for one of the many parts of the code that generates BUILD/bzl files. But the readme in there makes it pretty clear it's about gazelle. I'd be +1 for another PR to rename it to "gazelle" to make it more clear. In any case, BCR is using it to verify bzlmod+gazelle work together, so lets respect that: keep bzlmod+gazelle in build_file_generation for now.

@chrislovecnm Xudong sent me some ideas on another way to do this. I'm still trying to digest it. The key parts seem to be having rules_python's MODULE file call use_repo(), and then generating a .bzl file with Labels embedded in it that a separate (still within rules_python) extensions file loads.

rickeylev:
re: other ideas for another way to do this: we'll stick with this "manually construct the canonical repo name" implementation for now. The other solution has some appeal, but requires moving the pip extension into another file in order to work; moving the extension to another file means a breaking change. I'd rather we fix this bug with pip.parse using the wrong interpreter first; we can move it to a separate file and release that separately from the fix.

examples/bzlmod/MODULE.bazel
```diff
@@ -1,36 +1,128 @@
...
+# into the scope of the current module.
+# All of the python3 repositories use the PYTHON_NAME as there prefix.  They
+# are not catenated for ease of reading.
+use_repo(python, PYTHON_NAME)
+use_repo(python, "python3_toolchains")
+
+# This extension is used to look up the host system interpreter.
+# A label that is created by this extension is used for the
+# python_interpreter_target.
+python.host_python_interpreter(
```
rickeylev, line 75: I don't think we need this separate tag class anymore. The toolchain() call can just create the {name}_host_interpreter repo.

examples/bzlmod/MODULE.bazel
```diff
@@ -1,36 +1,128 @@
...
+    # This value is a label of a symlink that is generated by the
+    # host_python_interpreter extension above.  The extension chooses the
+    # correct interpreter and creates a symlink to the Python binary.
+    #
+    # Instead of using the extension you can also set the
+    # python_interpreter_target to a label like
+    # "python3_x86_64-unknown-linux-gnu//:python",
+    # if you are only supporting Linux.  If you are supporting
+    # multiple Operating Systems you need to use the host_python_interpreter as
+    # shown below.
```
rickeylev, line 115:
```suggestion
    # This attribute points to the interpreter to use for running
    # pip commands to download the packages in the requirements file.
    # As a best practice, we use the same interpreter as the toolchain
    # that was configured above; this ensures the same Python version
    # is used for both resolving dependencies and running tests/binaries.
    # If this isn't specified, then you'll get whatever is locally installed
    # on your system.
```

examples/bzlmod/README.md
rickeylev, line 1: A readme! Thank you, brilliant.

examples/bzlmod/README.md
```diff
@@ -0,0 +1,28 @@
+# Bzlmod example
+
+This example demostrates how to use `rules_python` and gazelle with `bzlmod` enabled.
```
rickeylev, line 3: Per discussion, please remove gazelle from the bzlmod example; keep that in build_file_generation instead.

python/extensions.bzl
```diff
@@ -19,23 +19,48 @@ load("@rules_python//python/pip_install:pip_repository.bzl", "locked_requirement
...
                 register_toolchains = False,
-                register_coverage_tool = attr.configure_coverage_tool,
-                ignore_root_user_error = attr.ignore_root_user_error,
+                register_coverage_tool = toolchain_attr.configure_coverage_tool,
+                ignore_root_user_error = toolchain_attr.ignore_root_user_error,
+            )
+        for interpret_attr in mod.tags.host_python_interpreter:
+            _host_hub(
+                name = interpret_attr.name,
+                extension_name = interpret_attr.extension_name,
```
rickeylev, line 39: The extension name doesn't need to be passed in. It's determined by the `python = module_extension()` line (like a rule name is determined). Just hardcode the extension name to "python"

python/extensions.bzl
```diff
@@ -19,23 +19,48 @@ load("@rules_python//python/pip_install:pip_repository.bzl", "locked_requirement
...
                 # Toolchain registration in bzlmod is done in MODULE file
                 register_toolchains = False,
-                register_coverage_tool = attr.configure_coverage_tool,
-                ignore_root_user_error = attr.ignore_root_user_error,
+                register_coverage_tool = toolchain_attr.configure_coverage_tool,
+                ignore_root_user_error = toolchain_attr.ignore_root_user_error,
+            )
+        for interpret_attr in mod.tags.host_python_interpreter:
+            _host_hub(
+                name = interpret_attr.name,
```
rickeylev, line 38: Reuse the name attribute from toolchain_attr.name (i.e move this into the loop above where the toolchain tag is processed). This avoids any issue of the two being out of sync. The two are closely coupled anyways.

python/extensions.bzl
```diff
@@ -127,3 +152,74 @@ pip = module_extension(
         "parse": tag_class(attrs = _pip_parse_ext_attrs()),
     },
 )
+
+# This function allows us to build the label name of a label
+# that is not passed into the current context.
+# The module_label is the key element that is passed in.
+# This value provides the root location of the labels
+def _repo_mapped_label(module_label, sub_module, apparent):
+    return Label("@@{module}~{sub_module}~{apparent}".format(
```
rickeylev, line 161: Rename "sub_module" to "extension name"

python/extensions.bzl
```diff
@@ -127,3 +152,74 @@ pip = module_extension(
         "parse": tag_class(attrs = _pip_parse_ext_attrs()),
     },
 )
+
+# This function allows us to build the label name of a label
+# that is not passed into the current context.
+# The module_label is the key element that is passed in.
+# This value provides the root location of the labels
+def _repo_mapped_label(module_label, sub_module, apparent):
```
rickeylev, line 160:
```suggestion
def _repo_mapped_label(module_label, sub_module, apparent):
    """Construct a canonical repo label accounting for repo mapping.
    
    Args:
        module_label: Label object of the module hosting the extension; see
          "_module" implicit attribute.
        ext_name: str, name of the extension that created the repo in `apparent`.
        apparent: str, a repo-qualified target string, but without the "@". e.g.
          "python38_x86_linux//:python". The repo name should use the apparent
          name used by the extension named by `ext_name` (i.e. the value of the
          `name` arg the extension passes to repository rules)
    """
```

python/extensions.bzl
rickeylev, line 161:
```suggestion
    # See https://bazel.build/external/extension#repository_names_and_visibility
    # for the format of canonical repo strings.
    return Label("@@{module}~{sub_module}~{apparent}".format(
```

python/extensions.bzl
```diff
@@ -127,3 +152,74 @@ pip = module_extension(
...
+# We are doing some bazel stuff here that could use an explanation.
+# The basis of this function is that we need to create a symlink to
+# the python binary that exists in a different repo that we know is
+# setup by rules_python.
+#
+# We are building a Label like
+# @@rules_python~override~python~python3_x86_64-unknown-linux-gnu//:python
+# and then the function creates a symlink named python to that Label.
+#
+# We are building the Label name dynamically,  and can do this even
```
rickeylev, line 176: typo: two spaces after comma

python/extensions.bzl
```diff
@@ -127,3 +152,74 @@ pip = module_extension(
...
+    ))
+
+# We are doing some bazel stuff here that could use an explanation.
+# The basis of this function is that we need to create a symlink to
+# the python binary that exists in a different repo that we know is
+# setup by rules_python.
+#
+# We are building a Label like
+# @@rules_python~override~python~python3_x86_64-unknown-linux-gnu//:python
+# and then the function creates a symlink named python to that Label.
```
rickeylev, line 174:
```suggestion
# and then the function creates a symlink named python to that Label.
# The tricky part is the "~override~" part can't be known in advance
# and will change depending on how and what version of rules_python
# is used. To figure that part out, an implicit attribute is used to
# resolve the module's current name (see "_module" attribute)
```

python/extensions.bzl
```diff
@@ -127,3 +152,74 @@ pip = module_extension(
...
+    # the repositories for the different interpreters.
+    (os, arch) = get_host_os_arch(repo_ctx)
+    host_platform = "{}_{}//:python".format(
+        repo_ctx.attr.extension_name,
+        get_host_platform(os, arch),
+    )
+
+    # the attribute is set to attr.label(default = "//:_"), which
+    # provides us the root module name.
+    # The sub_module name "python" is created by the rules_python
```
rickeylev, line 198: s/submodule/extension name

python/extensions.bzl
```diff
@@ -127,3 +152,74 @@ pip = module_extension(
...
+    # The two get_ functions we use are also utilized when building
+    # the repositories for the different interpreters.
+    (os, arch) = get_host_os_arch(repo_ctx)
+    host_platform = "{}_{}//:python".format(
+        repo_ctx.attr.extension_name,
+        get_host_platform(os, arch),
+    )
+
+    # the attribute is set to attr.label(default = "//:_"), which
+    # provides us the root module name.
```
rickeylev, line 197:
```suggestion
    # provides us the resolved, canonical, prefix for the module's repos
```

python/extensions.bzl
```diff
@@ -127,3 +152,74 @@ pip = module_extension(
...
+    label = _repo_mapped_label(repo_ctx.attr._module, "python", host_platform)
+
+    # create the symlink in order to set the interpreter for pip.
+    repo_ctx.symlink(label, "python")
+
+# This repository rule finds the host's Python interpreter and creates a symlink to it.
+# We use this rule to set the pips interpreter target when using different Operating
+# Systems with the same project.
+# This code exists because of changes bzlmod does not allow the use
+# of native.load_toolchain.
```
rickeylev, line 211: Please remove; this isn't because of native.register_toolchain, but because of repo remapping and visibility

rickeylev (changes requested):

examples/bzlmod/MODULE.bazel
chrislovecnm: Will that give us a unquie name when we have multiple tool chains?

examples/bzlmod/README.md
chrislovecnm: I am gonna create a different PR for the changes in examples.

python/extensions.bzl
chrislovecnm: The name is providing us a unique name that we use when we have multiple tool chains.

rickeylev: The repos for the toolchains are already unique. Calling python.toolchain() creates 3 types of repos:
  * {name}_<platform> these are the downloaded interpreters
  * {name}_toolchains this is a hub of just toolchain() declration calls, one for each platform, that point back to the the <platform> repos
  * {name} this is a hub like {name}_toolchains, but for some other stuff (I can't remember what)

examples/bzlmod/MODULE.bazel
rickeylev: Yes, each toolchain() call gets a name, and those names must be unique because eventually they create repos based on that name.

python/extensions.bzl
chrislovecnm: > {name}_<platform> these are the downloaded interpreters 

But they are not downloaded by the toolchain call.  They are only downloaded when we put in the host_python_interpreter call.  I could only see the new messages they downloaded when I added the code for the host_python_interpreter extension.

examples/bzlmod/README.md
chrislovecnm: We will only test the code if I have an example that uses pip and the host interpreter.  I am going to copy the current bzlmod to a new example.  I chatted with the bzlmod folks, and they still need to create a test harness to test the extension.

python/extensions.bzl
chrislovecnm: Please example more because I started trying to do that call before all this work.

examples/bzlmod/MODULE.bazel
chrislovecnm: @rickeylev addressed

examples/bzlmod/MODULE.bazel
chrislovecnm: @rickeylev addressed

python/extensions.bzl
chrislovecnm: @rickeylev addressed

python/extensions.bzl
chrislovecnm: @rickeylev addressed

python/extensions.bzl
chrislovecnm: @rickeylev addressed

python/extensions.bzl
chrislovecnm: @rickeylev addressed

python/extensions.bzl
chrislovecnm: @rickeylev addressed

examples/bzlmod/README.md
chrislovecnm: @rickeylev addressed

chrislovecnm:
@aignas I added a new example that includes bzlmod and gazelle.  We have some issues with the sub-module, which I will file.  I am not changing the build_file_generation example, and I will back out bzlmod support out of that example in another PR.

examples/bzlmod_build_file_generation/BUILD.bazel
```diff
@@ -0,0 +1,106 @@
...
+# - bazel run //:gazelle fix
+# See: https://github.com/bazelbuild/bazel-gazelle#fix-and-update
+gazelle(
+    name = "gazelle",
+    data = GAZELLE_PYTHON_RUNTIME_DEPS,
+    gazelle = "@rules_python_gazelle_plugin//python:gazelle_binary",
+)
+
+# Python gazelle does not support transition tests yet
+# so we add the following directive for gazelle to ignore
```
rickeylev, line 76: This comment says a directive is added, but I don't see it? I'm expecting to see e.g. "# gazelle: something" or some other special-looking comment.

examples/bzlmod_build_file_generation/runfiles/runfiles_test.py
```diff
@@ -0,0 +1,64 @@
...
+
+import os
+import unittest
+
+from other_module.pkg import lib
+
+from python.runfiles import runfiles
+
+
+class RunfilesTest(unittest.TestCase):
```
rickeylev, line 23: A whole testing of runfiles seems out of scope for an example...? Is this part of the example, or really just for test coverage?

python/extensions.bzl
```diff
@@ -19,18 +19,24 @@ load("@rules_python//python/pip_install:pip_repository.bzl", "locked_requirement
...
                 register_toolchains = False,
-                register_coverage_tool = attr.configure_coverage_tool,
-                ignore_root_user_error = attr.ignore_root_user_error,
+                register_coverage_tool = toolchain_attr.configure_coverage_tool,
+                ignore_root_user_error = toolchain_attr.ignore_root_user_error,
+            )
+            host_hub_name = toolchain_attr.name + "_host_interpreter"
+            _host_hub(
+                name = host_hub_name,
+                extension_name = toolchain_attr.name,
```
rickeylev, line 39: You can remove this because the extension name is always "python" (the name bound to the `module_extension()` return value), not the name passed to the tag class.

python/extensions.bzl
```diff
@@ -127,3 +133,88 @@ pip = module_extension(
...
+# though the Label is not passed into this function.  If we choose
+# not do this a user would have to write another 16 lines
+# of configuration code, but we are able to save them that work
+# because we know how rules_python works internally.  We are using
+# functions from private:toolchains_repo.bzl which is where the repo
+# is being built. The repo name differs between host OS and platforms
+# and the functions from toolchains_repo gives us this functions that
+# information.
+def _host_hub_impl(repo_ctx):
+    repo_ctx.file("BUILD.bazel", "")
```
rickeylev, line 183:
```suggestion
    # Intentionally empty; this is only intended to be used by repository
    # rules, which don't process build file contents.
    repo_ctx.file("BUILD.bazel", "")
```

python/extensions.bzl
```diff
@@ -127,3 +133,88 @@ pip = module_extension(
...
+    (os, arch) = get_host_os_arch(repo_ctx)
+    host_platform = "{}_{}//:python".format(
+        repo_ctx.attr.extension_name,
+        get_host_platform(os, arch),
+    )
+
+    # the attribute is set to attr.label(default = "//:_"), which
+    # provides us the resolved, canonical, prefix for the module's repos.
+    # The extension_name "python" is created by the rules_python
+    # folder structure.  Ergo rules_python/python.
```
rickeylev, line 196: This bit of the comment looks out of date? The repo_mapped_label function docs this pretty well, so remove.

```suggestion
```

python/extensions.bzl
```diff
@@ -127,3 +133,88 @@ pip = module_extension(
...
+    # We then have the OS and platform specific name of the python
+    # interpreter.
+    label = _repo_mapped_label(repo_ctx.attr._module, "python", host_platform)
+
+    # create the symlink in order to set the interpreter for pip.
+    repo_ctx.symlink(label, "python")
+
+# This repository rule finds the host's Python interpreter and creates a symlink to it.
+# We use this rule to set the pips interpreter target when using different Operating
+# Systems with the same project.
```
rickeylev, line 206:
```suggestion
# We use this rule to set the pips interpreter target when using different Operating
# Systems with the same project.
```
```suggestion
# We use this rule to set the pip interpreter target when using different operating
# systems with the same project.
```

python/extensions.bzl
```diff
@@ -127,3 +133,88 @@ pip = module_extension(
...
+    implementation = _host_hub_impl,
+    local = True,
+    attrs = {
+        "extension_name": attr.string(
+            mandatory = True,
+            doc = """\
+The name of the extension in the repository.  Usually the name you used when you created the 
+Python toolchain.
+""",
+        ),
```
rickeylev, line 217:
```suggestion
```

rickeylev (changes requested):
Looking pretty good, mostly small stuff

examples/bzlmod_build_file_generation/BUILD.bazel
chrislovecnm: It ignores it already.  I removed the comment.

examples/bzlmod_build_file_generation/runfiles/runfiles_test.py
chrislovecnm: It is part of the other example that I copied.  I think it is testing the sub module, but I do not know.

python/extensions.bzl
chrislovecnm: I'm trying to explain why we have "python" hardcoded.  It is because of the folder structure in rules_python. What do you recommend?

```
label = _repo_mapped_label(repo_ctx.attr._module, "python", host_platform)
```

We are hardcoding "python" in the line above.

python/extensions.bzl
```diff
@@ -127,3 +133,89 @@ pip = module_extension(
...
+    Args:
+        module_label: Label object of the module hosting the extension; see
+          "_module" implicit attribute.
+        extension_name: str, name of the extension that created the repo in `apparent`.
+        apparent: str, a repo-qualified target string, but without the "@". e.g.
+          "python38_x86_linux//:python". The repo name should use the apparent
+          name used by the extension named by `ext_name` (i.e. the value of the
+          `name` arg the extension passes to repository rules)
+    """
+    return Label("@@{module}~{extension_name}~{apparent}".format(
```
linzhp, line 154: You may be able to use `str(Label(apparent))` to get the canonical name. It's probably not a good idea to hardcode the format of a canonical name.

cc @fmeum

rickeylev: Label(apparent) won't work in this context because the repo being referenced (the platform-specific one) isn't visible (nobody called use_repo() on it).

While I agree that manually creating the canonical name isn't ideal, the other option, a hub-spoke model, requires Chris to essentially start over, which I'm not going to ask of him after a week of effort and now that we have something that works.

fmeum: @rickylev Repos generated by the same extension always see each other under the name given in the extension impl function, so the hub repo introduced in this PR should be able to just do `str(Label(apparent))` without extra complexity or `use_repo`s.

chrislovecnm:
@linzhp, maybe I am not understanding what you want, but this is what I did as a test:

```diff
diff --git a/python/extensions.bzl b/python/extensions.bzl
index 3bcbb50..3c3d582 100644
--- a/python/extensions.bzl
+++ b/python/extensions.bzl
@@ -151,11 +151,12 @@ def _repo_mapped_label(module_label, extension_name, apparent):
           name used by the extension named by `ext_name` (i.e. the value of the
           `name` arg the extension passes to repository rules)
     """
-    return Label("@@{module}~{extension_name}~{apparent}".format(
-        module = module_label.workspace_name,
-        extension_name = extension_name,
-        apparent = apparent,
-    ))
+    return Label("@python3_x86_64-unknown-linux-gnu")
+    #return Label("@@{module}~{extension_name}~{apparent}".format(
+    #    module = module_label.workspace_name,
+    #    extension_name = extension_name,
+    #    apparent = apparent,
+    #))

 # We are doing some bazel stuff here that could use an explanation.
 # The basis of this function is that we need to create a symlink to
```

I used `return Label("@python3_x86_64-unknown-linux-gnu")` on Linux instead of building the label in the fashion that is working.

We then get the error:

```
$ bazel clean --expunge && bazel build //...
INFO: Starting clean (this may take a while). Consider using --async if the clean takes more than several minutes.
DEBUG: gazelle@0.27.0/MODULE.bazel:7:6: WARNING: The bazel_gazelle Bazel module is still highly experimental and subject to change at any time. Only use it to try out bzlmod for now.
DEBUG: rules_go@0.37.0/MODULE.bazel:8:6: WARNING: The rules_go Bazel module is still highly experimental and subject to change at any time. Only use it to try out bzlmod for now.
DEBUG: rules_go@0.33.0/MODULE.bazel:7:6: WARNING: The rules_go Bazel module is still highly experimental and subject to change at any time. Only use it to try out bzlmod for now.
INFO: Repository rules_python~override~python~python3_host_interpreter instantiated at:
  callstack not available
Repository rule _host_hub defined at:
  /home/clove/.cache/bazel/_bazel_clove/e19dab5c65f1ab29c26be4ab0039d0eb/external/rules_python~override/python/extensions.bzl:209:28: in <toplevel>
ERROR: An error occurred during the fetch of repository 'rules_python~override~python~python3_host_interpreter':
        File "/home/clove/.cache/bazel/_bazel_clove/e19dab5c65f1ab29c26be4ab0039d0eb/external/rules_python~override/python/extensions.bzl", line 205, column 21, in _host_hub_impl
                repo_ctx.symlink(label, "python")
Error in symlink: Unable to load package for @[unknown repo 'python3_x86_64-unknown-linux-gnu' requested from @rules_python~override]//:python3_x86_64-unknown-linux-gnu: The repository '@[unknown repo 'python3_x86_64-unknown-linux-gnu' requested from @rules_python~override]' could not be resolved: No repository visible as '@python3_x86_64-unknown-linux-gnu' from repository '@rules_python~override'
ERROR: <builtin>: fetching _host_hub rule @rules_python~override//python:rules_python~override~python~python3_host_interpreter: Traceback (most recent call last):
        File "/home/clove/.cache/bazel/_bazel_clove/e19dab5c65f1ab29c26be4ab0039d0eb/external/rules_python~override/python/extensions.bzl", line 205, column 21, in _host_hub_impl
                repo_ctx.symlink(label, "python")
Error in symlink: Unable to load package for @[unknown repo 'python3_x86_64-unknown-linux-gnu' requested from @rules_python~override]//:python3_x86_64-unknown-linux-gnu: The repository '@[unknown repo 'python3_x86_64-unknown-linux-gnu' requested from @rules_python~override]' could not be resolved: No repository visible as '@python3_x86_64-unknown-linux-gnu' from repository '@rules_python~override'
INFO: Repository bazel_tools~remote_coverage_tools_extension~remote_coverage_tools instantiated at:
  callstack not available
Repository rule http_archive defined at:
  /home/clove/.cache/bazel/_bazel_clove/e19dab5c65f1ab29c26be4ab0039d0eb/external/bazel_tools/tools/build_defs/repo/http.bzl:372:31: in <toplevel>
INFO: Repository rules_cc~0.0.2 instantiated at:
  callstack not available
Repository rule http_archive defined at:
  /home/clove/.cache/bazel/_bazel_clove/e19dab5c65f1ab29c26be4ab0039d0eb/external/bazel_tools/tools/build_defs/repo/http.bzl:372:31: in <toplevel>
INFO: Repository rules_java~5.3.5 instantiated at:
  callstack not available
Repository rule http_archive defined at:
  /home/clove/.cache/bazel/_bazel_clove/e19dab5c65f1ab29c26be4ab0039d0eb/external/bazel_tools/tools/build_defs/repo/http.bzl:372:31: in <toplevel>
ERROR: /home/clove/.cache/bazel/_bazel_clove/e19dab5c65f1ab29c26be4ab0039d0eb/external/rules_python~override~pip~pip/tabulate/BUILD.bazel:3:6: no such package '@rules_python~override~pip~pip_tabulate//': no such package '@rules_python~override~python~python3_host_interpreter//': Unable to load package for @[unknown repo 'python3_x86_64-unknown-linux-gnu' requested from @rules_python~override]//:python3_x86_64-unknown-linux-gnu: The repository '@[unknown repo 'python3_x86_64-unknown-linux-gnu' requested from @rules_python~override]' could not be resolved: No repository visible as '@python3_x86_64-unknown-linux-gnu' from repository '@rules_python~override' and referenced by '@rules_python~override~pip~pip//tabulate:tabulate'
ERROR: Analysis of target '//:bzlmod_build_file_generation' failed; build aborted:
INFO: Elapsed time: 3.694s
INFO: 0 processes.
```

Let me know if there is a different approach.  We can always iterate to improve this, but we have already spent many hours of dev time on this.

rickeylev (approved):

python/extensions.bzl
rickeylev: Thanks for trying, Chris.

@fmeum @linzhp I agree with your reasoning, but it just doesn't seem to actually be working like that. This is queued for merging, so should land in a few minutes, so you'll be able to play with it and maybe be able tell us what's going on -- both Chris and I have hit a will trying to figure out what's going on.

rickeylev:
I think I remember why calling Label() there doesn't work: It's being called in the repo _implementation_ function, not as part of _evaluating_ the code generated by the repo.

When Label() is called in the host_hub_impl implementation function in extensions.bzl, it's current context is the rules_python module. Since the rules_python module itself never called use_repo() on anything (it can't; the repo names are user selected), it's repo visibility mapping is empty.

If host_hub generated a .bzl file with e.g. `Label("@python3_x86_64-linux")`, that _would_ work, because when someone loads the host hub repo, that Label call will be evaluated in the context of the host hub _repo_, which _does_ have access to the other repos that the extension generated.

fmeum:
@rickeylev That's correct. This situation is especially tricky since the resolved label is needed in the implementation function of the repo rule. This may require introducing a helper extension.

@Wyverald and I were discussing a `module_ctx.label_in_repo` function that would drastically simplify such situations. If you would like to see something like this realized, could you file an issue for it?

PR #1158, opened 2023-04-07
fix: respect kind mapping

OniOni:
When using the kind `gazelle:map_kind` directive, `gazelle` will correctly generate the buildfile on the first pass (or if no target of that type / name are present). 

However, when running gazelle a second time (or if a target of the mapped kind with the same name is present), `gazelle` will error out saying that it kind create a target of the original kind because a target of mapped kind is present and has the same name. 

Ex:
Given the directive `# gazelle:map_kind py_test py_pytest_test //src/bazel/rules/python:py_pytest_test.bzl`, `gazelle` will correctly generate a `py_pytest_test` target where it would have generated a `py_test` target. 

But on a second invocation of `gazelle` (and subsequent invocations) it will error our with:
```
gazelle: ERROR: failed to generate target "//test/python/common:common_test" of kind "py_test": a target of kind "py_pytest_test" with the same name already exists. Use the '# gazelle:python_test_naming_convention' directive to change the naming convention.
```

### DIFF
--- a/gazelle/python/generate.go
+++ b/gazelle/python/generate.go
@@ -46,6 +46,13 @@ var (
 	buildFilenames = []string{"BUILD", "BUILD.bazel"}
 )
 
+func GetActualKindName(kind string, args language.GenerateArgs) string {
+	if kindOverride, ok := args.Config.KindMap[kind]; ok {
+		return kindOverride.KindName
+	}
+	return kind
+}
+
 // GenerateRules extracts build metadata from source files in a directory.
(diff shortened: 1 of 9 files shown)


OniOni:
We can see the test correctly fails without the fix: https://buildkite.com/bazel/rules-python-python/builds/4567#01875d27-a8ef-4642-bea1-cba19d930bde/339-344

f0rmiga (approved):
Awesome, thanks!

PR #1175, opened 2023-04-20
Update bzlmod version for Gazelle plugin

linzhp:
Update the bzlmod version for Gazelle plugin to the next release so we can use it as bzlmod.

@aignas FYI

### DIFF
--- a/gazelle/MODULE.bazel
+++ b/gazelle/MODULE.bazel
@@ -1,6 +1,6 @@
 module(
     name = "rules_python_gazelle_plugin",
-    version = "0.0.0",
+    version = "0.21.0",
     compatibility_level = 1,
 )
 


gazelle/MODULE.bazel
```diff
@@ -1,6 +1,6 @@
 module(
     name = "rules_python_gazelle_plugin",
-    version = "0.0.0",
+    version = "0.21.0",
```
chrislovecnm, line 3: So BCR does not let us publish two modules from the same git repo yet. https://github.com/bazelbuild/rules_python/issues/1177

And since this works now, I think the version here does not matter.  Because it is living in the same repo, I am thinking this value is getting overridden.  Someone who is a better bzlmod guru than I can comment.

aignas:
Should we close this as the the gazelle plugin is part of the BCR now?

PR #1176, opened 2023-04-20
fix: Using canonical name in requirements.bzl

linzhp:
When `incompatible_generate_aliases = False`, `pip.parse` doesn't work in bzlmod, because requirements.bzl has:

```
all_requirements = ["@rules_python~0.21.0~pip~pip_yapf//:pkg"]

all_whl_requirements = ["@rules_python~0.21.0~pip~pip_yapf//:whl"]
```

Starting Bazel 6, canonical names should be referred with double "@".

The reason why `incompatible_generate_aliases = True` worked is because it uses apparent name by parsing the canonical label with `repo_name = rctx.attr.name.split("~")[-1]`. This is dangerous because Bazel may change its canonical name pattern in future versions.

This PR adds a new attribute "repo_name" to `pip_repository_bzlmod`, so we have access to the human-readable name in the implementation.


aignas (approved):
LGTM, thanks for fixing this and giving context.

chrislovecnm:
How do we add a test for this?  I have noticed that bzlmod testing is a little light.  Happy to see a follow-up PR for testing.

python/pip_install/pip_repository.bzl
```diff
@@ -386,6 +386,10 @@ pip_repository_bzlmod_attrs = {
         default = False,
         doc = "Allow generating aliases in '@pip//:<pkg>' -> '@pip_<pkg>//:pkg'. This replaces the aliases generated by the `bzlmod` tooling.",
     ),
+    "repo_name": attr.string(
+        mandatory = True,
```
chrislovecnm, line 390: If I am reading this correctly, putting this value as mandatory introduces a breaking change for our API.  Should we create a way not to break the API so that an upgrade path is more straightforward?  I am curious.

chrislovecnm (changes requested):
I think we are introducing an API change that will break the API when you upgrade.  It is only needed for bzlmod.  Please let me know if I am wrong, but I recommend a nonbreaking change.  Test for bzlmod and throw an error if we are using bzlmod.

python/pip_install/pip_repository.bzl
linzhp: According to this:

https://github.com/bazelbuild/rules_python/blob/c72c7bcf4e0899c275042328e8233e3124ccae86/python/pip_install/pip_repository.bzl#L414-L418

`pip_repository_bzlmod` is not part of API.

chrislovecnm: 👍 Thanks!  I am always learning.  I was wondering if it was private.  Thoughts on a follow-up PR for testing?

linzhp: I think that's a good idea. Let me figure it out later.

linzhp:
[This set of tests](https://github.com/bazelbuild/rules_python/blob/73aec8f29b12b1e0f69f2cb7fc8e7fa2c34ca591/.bazelci/presubmit.yml#L150-L174) should have failed without this PR, because it would build `//:modules_map`. With bzlmod, the target would be expanded to:

```
modules_mapping(
  name = "modules_map",
  exclude_patterns = ["^_|(\\._)+", "(\\.tests)+"],
  wheels = ["@rules_python~override~pip~pip//click:whl", "@rules_python~override~pip~pip//flask:whl", "@rules_python~override~pip~pip//importlib_metadata:whl", "@rules_python~override~pip~pip//itsdangerous:whl", "@rules_python~override~pip~pip//jinja2:whl", "@rules_python~override~pip~pip//markupsafe:whl", "@rules_python~override~pip~pip//werkzeug:whl", "@rules_python~override~pip~pip//zipp:whl"],
)
```

Labels like `@rules_python~override~pip~pip//click:whl` are invalid. @chrislovecnm any idea why it didn't fail?

chrislovecnm:
@linzhp so that is interesting; I am uncertain why those are not failing.  We are removing bzlmod from the `build_file_generation` project in #1155.  But they are not failing on the bzlmod project either. Do you have any ideas on how to track this down?  We need more testing.

aignas:
> modules_mapping(
>   name = "modules_map",
>   exclude_patterns = ["^_|(\\._)+", "(\\.tests)+"],
>   wheels = ["@rules_python~override~pip~pip//click:whl", "@rules_python~override~pip~pip//flask:whl", "@rules_python~override~pip~pip//importlib_metadata:whl", "@rules_python~override~pip~pip//itsdangerous:whl", "@rules_python~override~pip~pip//jinja2:whl", "@rules_python~override~pip~pip//markupsafe:whl", "@rules_python~override~pip~pip//werkzeug:whl", "@rules_python~override~pip~pip//zipp:whl"],
> )

The tests are passing because the `incompatible_generate_aliases` is used in the mentioned tests and the wheel labels are different. Thanks for pointing out that the `else` clause on [this line](https://github.com/bazelbuild/rules_python/blob/main/python/pip_install/pip_repository.bzl#L373) has a bug. I don't think there are any users of the `whl` labels within the `rules_python` other than the `modules_mapping`.

linzhp:
Good catch. I added two tests to cover this situation. @chrislovecnm Please take another look

aignas (approved):

chrislovecnm (approved):
LGTM - can @rickeylev or @f0rmiga PTAL?

chrislovecnm (changes requested):
Is CI working locally?  You may need to rebase.

chrislovecnm:
@f0rmiga or @rickeylev can you PTAL?

linzhp:
@chrislovecnm Like I commented on #1155, the way you deduce from apparent name to canonical name assumes some Bazel implementation details, which is not a good idea. In this PR, I don't see a way to deduce apparent name from canonical name without assuming that same impl details either.

python/pip_install/pip_repository.bzl
```diff
@@ -379,11 +379,11 @@ def _pip_repository_bzlmod_impl(rctx):
     rctx.file("BUILD.bazel", build_contents)
     rctx.template("requirements.bzl", rctx.attr._template, substitutions = {
         "%%ALL_REQUIREMENTS%%": _format_repr_list([
-            "@{}//{}".format(repo_name, p) if rctx.attr.incompatible_generate_aliases else "@{}_{}//:pkg".format(rctx.attr.name, p)
+            "@{}//{}".format(repo_name, p) if rctx.attr.incompatible_generate_aliases else "@{}//:{}_pkg".format(repo_name, p)
```
rickeylev, line 382: Why not just use the canonical name? i.e. rctx.name. We recently changed the requirement() macro to use that under the hood (see a few lines above). Actually, it'd probably be good to change this code to use the macro_tmpl variable from above. Both that and these lines should be generating the same strings, right?

Using the canonical name isn't bad per-se; using rctx.name is fine. The difference here vs #1155 is, in #1155 we had to construct canonical names to _other_ repos; this just wants to reference itself.

These strings are in the same situation as the requirements macro, right? i.e. they are loaded into another context and passed along, and then actually resolved to labels elsewhere?

aignas (approved):
Thanks, this looks better than before.

Left a single comment that a simplification is possible, but overall, LGTM.

linzhp:
@aignas I didn't see your comment?

linzhp:
@chrislovecnm Done

PR #1181, opened 2023-04-26
fix: manually ignore bazel-* directories to make using custom Bazel builds easier

rickeylev:
Normally, Bazel will ignore its convenience symlinks, so putting them in the .bazelignore file isn't necessary. However, when `--output_user_root` is set, which is beneficial to set when using different Bazel versions (it preserves the analysis cache between versions), the symlinks aren't ignored. Putting them in the bazelignore file fixes this.


Solumin (approved):

chrislovecnm:
@rickeylev can you nudge CI?

PR #1184, opened 2023-04-27
tests: Upgrade rules_testing to 0.0.5

rickeylev:
This just keeps it at a recent version, which makes errors easier to grok. rules_testing recently underwent a large refactor, so it's confusing when errors reference lines that no longer exist in recent versions.


groodt (approved):
LGTM

PR #1187, opened 2023-04-29
tests: Force analysis test labels to resolve within @rules_python context

rickeylev:
When a string label is passed to the `@rules_testing` analysis_test functions, the strings are evaluated within the context of @rules_testing because that is where the actual rule invocation happens. Without bzlmod, this just requires qualifying the labels with the repo name (which is what was being done) because there's just a flat global namespace of repos.

With bzlmod enabled, repo mapping happens, so rules_testing tries to resolve those repo names using its repo mapping, which doesn't work because rules_testing's mapping doesn't include every repo using it.

### DIFF
--- a/tools/build_defs/python/tests/py_test/py_test_tests.bzl
+++ b/tools/build_defs/python/tests/py_test/py_test_tests.bzl
@@ -22,6 +22,12 @@ load(
 )
 load("//tools/build_defs/python/tests:util.bzl", pt_util = "util")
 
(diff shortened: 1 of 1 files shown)


Solumin (approved):

chrislovecnm (changes requested):
CI is failing

chrislovecnm:
@rickeylev can we nudge CI and see if this is green?

PR #1189, opened 2023-04-30
feat: expose python package entrypoints via an opaque struct

### LINKED ISSUE #958: [bzlmod bug/regression] `entry_point` no longer working with blzmod
# 🐞 bug report

### Affected Rule

`entry_point` when using with blzmod

### Is this a regression?

Yes

### Description

`entry_point` rule no longer working when used with blzmod

## 🔬 Minimal Reproduction

You can find a reproducible example at this repo: https://github.com/xinbinhuang/rules_python_bug

## 🔥 Exception or Error

```
~ $ bazel run "//:flake8"    
ERROR: /workspace/BUILD.bazel:3:6: no such package '@[unknown repo 'pip_flake8' requested from @]//': The repository '@[unknown repo 'pip_flake8' requested from @]' could not be resolved: No repository visible as '@pip_flake8' from main repository and referenced by '//:flake8'
ERROR: Analysis of target '//:flake8' failed; build aborted: Analysis failed
INFO: Elapsed time: 4.436s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (1 packages loaded, 0 targets configured)
```
## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
mac/linux
  </code>
</pre>

<pre>
  <code>
6.0.0
  </code>
</pre>

**Rules_python version:**

<pre>
  <code>
0.16.1
  </code>
</pre>

**Anything else relevant?**

aignas:
Fixes #958.

For bzlmod users we did not have a working alternative for the `entry_point` macro for the non-bzlmod users. The reason it is tricky is that in order to know the name of the entry_point available to the user one needs to download the Python package and parse the metadata. This works around the limitation by exposing the list of entrypoints back to the hub repo as an opaque struct, which the users can use in their code.

Summary:
* Added support for users setting `incompatible_generate_aliases` to `True`. They can use the `bin` struct exposed via `@<hub_rep>//<dep>:bin.bzl`.
* Added support for bzlmod users not setting `incompatible_generate_aliases`. They can use it via `use_repo("<hub_repo>_<dep")` and then importing the same struct via `@<hub_repo>_<dep>//:bin.bzl`.
* The legacy behaviour of the `entry_point` macro is unchanged for the remaining users for now.

Design notes:
* Expose the struct in separate `.bzl` files in order to have no eager fetches.
* Exposing it via a struct will give users an error message if the target with the specified name does not exist and it will tell the available struct attributes.
* The inspiration comes from `rules_js` which @alexeagle has pointed to.

Open questions:
* I know that @groodt was toying with an idea of removing the repository macros (`requirement`, et. al) and I am wondering if this may be the last piece that we need in order to give our users an alternative.
* If the design looks good, I am wondering about the documentation we should have. Is the current addition to the `bzlmod` example sufficient?


aignas:
Depends on #1190 being merged to `main`. For now I have merged the contents of that PR into this branch to ensure that CI is happy.

python/pip_install/pip_repository.bzl
```diff
@@ -475,6 +495,7 @@ def _pip_repository_impl(rctx):
         ]),
         "%%ANNOTATIONS%%": _format_dict(_repr_dict(annotations)),
         "%%CONFIG%%": _format_dict(_repr_dict(config)),
+        "%%ENTRYPOINT_REPO%%": "//{pkg}" if rctx.attr.incompatible_generate_aliases else "_{pkg}//",
```
rickeylev, line 498: Where is this %ENTRYPOINT_REPO% used? I'm not seeing it.

rickeylev, line 498: Something about this looks weird.

The variable is named "repo", but "//{pkg}" isn't a repo-name.

The "_{pkg}//" value: I dont't repos can start with underscore? I got an error when I tried to do that recently. Shouldn't it also have "@" in it?

python/pip_install/tools/wheel_installer/wheel_installer.py
```diff
@@ -289,6 +298,17 @@ def _generate_build_file_contents(
     )
 
 
+def _generate_bin_bzl_contents(entrypoints: dict[str, str]) -> str:
+    return textwrap.dedent(
+        """\
+        # Autogenerated by wheel_installer.py
```
rickeylev, line 304:
```suggestion
        # Autogenerated by wheel_installer.py#_generate_bin_bzl_contents
```

Thanks for adding the "generated by" text. Those comments have been very helpful as I've tried to trace code around. I just a small suggestion to point directly to the part doing it. I'm also +1 on including any other extra info in the comment that makes it easier for maintainers.

python/pip_install/tools/wheel_installer/wheel_installer.py
```diff
@@ -289,6 +298,17 @@ def _generate_build_file_contents(
...
+        bin = struct(**{})
+        """.format(
+            repr({name: Label(target=target) for name, target in entrypoints.items()})
```
rickeylev, line 307: Please use a dedicate formatting function instead of relying on repr(). The output of Python's repr isn't guaranteed to be stable between Python releases, nor is it guaranteed to be compatible with Starlark.

rickeylev, line 307: Please sort the values on output. This just helps avoid any determinism issues from the dict ordering. While Python dicts now have guaranteed insertion order, it's easy for some other code to change the insertion order of the dict (e.g. someone calls set on some values earlier). We can avoid that problem by just sorting the output.

examples/bzlmod/BUILD.bazel
```diff
@@ -43,3 +44,11 @@ py_test_with_transition(
     main = "test.py",
     deps = [":lib"],
 )
+
+alias(
+    name = "yamllint",
+    # This is using the struct defined by the spoke repo 'yamllint' and it is
+    # re-exported by the hub repo named 'pip'. This allows bzlmod and
+    # non-bzlmod users to access the entry point targets.
+    actual = yamllint_bin.yamllint,
```
rickeylev, line 53: I don't really follow this part -- why does it have to be put into a bzl file and loaded like this? This is pointing to a target, right? So why can't it just be "@pip//yamllint:bin"?

(I'm guessing the answer has something to do with bzlmod, repo mapping, and repo visibility, but I'd like to understand more).

python/pip_install/pip_repository.bzl
```diff
@@ -307,6 +307,22 @@ alias(
         )
         rctx.file("{}/BUILD.bazel".format(name), build_content)
 
+        # Re-export the opaque struct that contains labels to all of the
+        # entrypoint labels available in the package repository.
+        #
+        # Prefix the re-export with a dep name in order to avoid collisions
+        # during loading of multiple bin structs at the same time.
+        bin_content = """\
+\"\"\"Autogenerated by 'pip_parse'.\"\"\"
```
rickeylev, line 316: nit: be a bit more specific in what generated it. "pip_parse" is a bit vague and there's several pieces to it.

python/pip_install/tools/wheel_installer/wheel_installer.py
```diff
@@ -32,6 +33,14 @@
 from python.pip_install.tools.wheel_installer import namespace_pkgs, wheel
 
 
+@dataclass
+class Label:
```
rickeylev, line 37: Can we call this LabelWrapper or something? I just want it to be easier to tell when it's Python code calling the dataclass named Label vs generated bzlcode in the Python code calling the bzl Label. The behavior of Label() is confusing enough; adding "Which label object is this?" is going to make it very confusing.

Actually, can we remove this dataclass entirely? It looks like it's just used once?

python/pip_install/tools/wheel_installer/wheel_installer.py
```diff
@@ -32,6 +33,14 @@
...
+    target: str
```
rickeylev, line 38: nit: "target_name" instead of just "target". A Target is a 3-tuple of (workspace, package, and name), and this is only the name portion. That, or add a comment to indicate it's only the name portion of a target.

rickeylev (changes requested):

rickeylev:
> I am wondering about the documentation we should have. 

Probably also update the "entry_point" function docs. If I'm trying to use an entry_point in a certain way, I'll consult its reference documentation to see how to use it. I'll look at examples when I'm first using a library, but after I use it a few times, I'm unlikely to look at the examples (they'll just tell me most of what I know).

aignas:
@rickylev, thanks for the great comments. It seems that the documentation for the bzlmod extensions is non-existent right now so I added minimal text to the right place even though it is not generated right now.

I also noticed that there are plenty of `bzl_library` re-definitions in the `docs` package even though we should have most of the `bzl_library` definitions in the appropriate places already. Would you prefer me to:
1. Add `extensions.bzl` to the `bzl` filegroup to keep the consistentcy?
1. Or to re-use the `bzl_library` definitions from the appropriate places and keep the `docs` package only about docs generation?
1. Or should we use the gazelle plugin as a dev dependency for generating `bzl_library` targets automatically?

I feel like the documentation for extensions could be a separate PR as there are extra decisions to be made here. I'll see how far I get with adding `extensions.bzl` to the `//python:bzl` filegroup and if it is small enough, I'll add it to this PR.

aignas:
So I have generated some docs for bzlmod extensions specifically, but I am not sure if how this is turning out is the best way it can be. See the link [here](https://github.com/bazelbuild/rules_python/blob/e16e3a944838a67345c94f611e161c6287845e12/docs/extensions.md).

It may be a good start, but without `stardoc` generating docs automatically for `module_extension`, there is just no good way to have consistent docs for this. Putting the docs to the non-bzlmod macros may be as, if not more, confusing.

aignas:
Depends on #1190, but marking as ready for review as I have addressed the first round of comments.

aignas:
Whilst trying to add tests for this I have noticed that it would be better to wait for a few PRs already in the queue, so that I can add better tests. What is more, maybe splitting the documentation for the `extensions.bzl` into its own PR would be a good idea.

aignas:
I am wondering if this is really the best way to solve this. The reason for it is that if the user would like to use a select statement to select the entrypoint based on a configuration option, it would inevitably eagerly fetch all of the repos used in the select statement. If we had an entry point macro that is in rules_python, it might be a different story.

aignas:
I am wondering if this is really the best way to solve this. The reason for it is that if the user would like to use a select statement to select the entrypoint based on a configuration option, it would inevitably eagerly fetch all of the repos used in the select statement. If we had an entry point macro that is in rules_python, it might be a different story.

aignas:
I am going to close this as #1220 is the approach I would recommend for now.

PR #1190, opened 2023-04-30
fix(bzlmod): correctly template repository macros for requirements, etc

aignas:
It seems that the macros for specifying the requirements break when the
user starts using `incompatible_generate_aliases=True`. This PR fixes
this.

Testing done:
1. Modify the example:
   ```
   $ git diff
   diff --git a/examples/bzlmod/MODULE.bazel b/examples/bzlmod/MODULE.bazel
   index ce91228..1750210 100644
   --- a/examples/bzlmod/MODULE.bazel
   +++ b/examples/bzlmod/MODULE.bazel
   @@ -26,6 +26,7 @@ register_toolchains(
    pip = use_extension("@rules_python//python:extensions.bzl", "pip")
    pip.parse(
        name = "pip",
   +    incompatible_generate_aliases=True,
    )
   ```
2. Run `bazel build ...` and check that it is still working.

I noticed this when working on #1189 and creating a separate PR for easier cherry-picking if we wanted to make a patch release which includes this. I am not sure how I could make an automated test for this other than creating a separate example.

### DIFF
--- a/python/pip_install/pip_repository.bzl
+++ b/python/pip_install/pip_repository.bzl
@@ -367,9 +367,14 @@ def _pip_repository_bzlmod_impl(rctx):
     else:
         build_contents += _bzlmod_pkg_aliases(repo_name, bzl_packages)
 
-    macro_tmpl = "@@{name}//:{{}}_{{}}".format(name = rctx.attr.name)
+    # NOTE: we are using the canonical name with the double '@' in order to
+    # always uniquely identify a repository, as the labels are being passed as
+    # a string and the resolution of the label happens at the call-site of the
+    # `requirement`, et al. macros.
     if rctx.attr.incompatible_generate_aliases:
         macro_tmpl = "@@{name}//{{}}:{{}}".format(name = rctx.attr.name)
+    else:
+        macro_tmpl = "@@{name}//:{{}}_{{}}".format(name = rctx.attr.name)
 
     rctx.file("BUILD.bazel", build_contents)
     rctx.template("requirements.bzl", rctx.attr._template, substitutions = {
@@ -381,8 +386,8 @@ def _pip_repository_bzlmod_impl(rctx):
             "@{}//{}:whl".format(repo_name, p) if rctx.attr.incompatible_generate_aliases else "@{}_{}//:whl".format(rctx.attr.name, p)
             for p in bzl_packages
         ]),
+        "%%MACRO_TMPL%%": macro_tmpl,
         "%%REQUIREMENTS_LOCK%%": str(requirements_txt),
-        "%%TMPL%%": macro_tmpl,
     })
 
 pip_repository_bzlmod_attrs = {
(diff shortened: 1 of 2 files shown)


python/pip_install/pip_repository.bzl
```diff
@@ -367,6 +367,10 @@ def _pip_repository_bzlmod_impl(rctx):
     else:
         build_contents += _bzlmod_pkg_aliases(repo_name, bzl_packages)
 
+    macro_tmpl = "@@{name}//:{{}}_{{}}".format(name = rctx.attr.name)
+    if rctx.attr.incompatible_generate_aliases:
+        macro_tmpl = "@@{name}//{{}}:{{}}".format(name = rctx.attr.name)
+
```
rickeylev, line 373: Please use an if-else to ensure there is only one code path that sets the variable. This prevents bugs in the optional if-body that the non-optional statements above might hide.

python/pip_install/pip_repository.bzl
```diff
@@ -367,6 +367,10 @@ def _pip_repository_bzlmod_impl(rctx):
     else:
         build_contents += _bzlmod_pkg_aliases(repo_name, bzl_packages)
 
+    macro_tmpl = "@@{name}//:{{}}_{{}}".format(name = rctx.attr.name)
```
rickeylev, line 370: Would you mind putting a brief comment about why we use "@@"? When and why to use the apparent name vs canonical name isn't always clear and is a bit confusing.

python/pip_install/pip_repository.bzl
```diff
@@ -377,8 +381,8 @@ def _pip_repository_bzlmod_impl(rctx):
             "@{}//{}:whl".format(repo_name, p) if rctx.attr.incompatible_generate_aliases else "@{}_{}//:whl".format(rctx.attr.name, p)
             for p in bzl_packages
         ]),
-        "%%NAME%%": rctx.attr.name,
         "%%REQUIREMENTS_LOCK%%": str(requirements_txt),
+        "%%TMPL%%": macro_tmpl,
```
rickeylev, line 385: Lets use a more descriptive name than "TMPL" as the substituted name. A template that templates in a a variable named template is a bit ~~meta-templately~~ confusing :).

rickeylev (changes requested):

rickeylev (approved):
>  I am not sure how I could make an automated test for this other than creating a separate example.

Yeah, easily testing modules is hard. I'm +1 on creating another "example" to test this, but OK with that being a separate PR. I put example in quotes because it doesn't have to be a user-friendly example in the "examples" directory. Maybe we should have a directory for various MODULE/WORKSPACE variations we need to test, so we know they don't have to be user-friendly examples with extensive docs etc.

rickeylev:
CI is failing, but I think it's unrelated and due to https://github.com/bazelbuild/continuous-integration/issues/1605, so I'm going to merge

PR #1191, opened 2023-05-01
fix: Allow passing a tuple to the `tags` attribute.

rickeylev:
Starlark rules allow giving the tags as a tuple. The helper function that added the special migration tag assumed tags was always a list, resulting in an error when it tried to concatenate a list and tuple.

To fix, check if tags is a tuple and concatenate a tuple if so. The input type of the tags attribute is preserved so that a test verifying tags can be passed to the underlying rule can be implemented (this test is to verify there isn't a regression during the rewrite to Starlark).

### DIFF
--- a/docs/BUILD.bazel
+++ b/docs/BUILD.bazel
@@ -80,6 +80,9 @@ bzl_library(
         "//python/private:stamp.bzl",
         "//python/private:util.bzl",
     ],
+    deps = [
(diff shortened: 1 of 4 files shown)


Solumin (approved):

rickeylev:
CI failures are due to some experiments going awry in Bazel CI. Merging anyways.

Issue #1194, opened 2023-05-02
URL fragments generate invalid "wheel" `--hash` arguments


mattoberle:
# 🐞 bug report

### Affected Rule

```
python/pip_install/requirements_parser.bzl:parse
```

### Is this a regression?

This change in behavior was introduced in `0.19.0`.

### Description

The `requirements_parser` produces bad output when it encounters the `#` symbol in a dependency URL.
It appends the `--hash=` directive that follows to `extra_pip_args`, which subsequently breaks all `pip wheel` commands.

The [reference for the requirements file format](https://pip.pypa.io/en/stable/reference/requirements-file-format/#comments) defines comments as:

1. Lines that begin with `#`.
2. Content preceded by ` #` (a `space` followed by `#`).

Example breaking build file:

```
#requirements_lock.txt
requests @ https://github.com/psf/requests/releases/download/v2.29.0/requests-2.29.0.tar.gz#sha1=3897c249b51a1a405d615a8c9cb92e5fdbf0dd49 \
    --hash=sha256:eca58eb564b134e4ff521a02aa6f566c653835753e1fc8a50a20cb6bee4673cd
    # via -r requirements.txt
setuptools==67.4.0 \
    --hash=sha256:f106dee1b506dee5102cc3f3e9e68137bbad6d47b616be7991714b0c62204251
    # via -r requirements.txt
```

The problem ultimately resides in this line of code:
- https://github.com/bazelbuild/rules_python/blob/main/python/pip_install/requirements_parser.bzl#L130-L131

I started work on a patch, but realized that [_handleConsumeSpace](https://github.com/bazelbuild/rules_python/blob/main/python/pip_install/requirements_parser.bzl#L78-L88) discards spaces, making it difficult to determine whether the previous consumed character was a space.

~To conform with the spec, I think we'd need to either modify the way we manage the buffer or track whether the last action was `_handleConsumeSpace`.~

~When evaluating whether the current parser is "correct enough", I suppose some questions to answer are:~

- ~Do `#shaX=...` in URLs even provide value given the explicit `--hash` arguments? (My guess: no).~
- ~Is it still worth fixing given that `#` in a URL is parse-able by `pip`?~

> **Edit:** I was mistaken about the buffer. This is easier to implement than I thought. I can open a supporting PR.

## 🔬 Minimal Reproduction

This patch creates a test for the case.

```diff
[...]

PR #1195, opened 2023-05-02
fix: allow url fragments in requirements file

mattoberle:
This commit addresses issue #1194 (see issue for details).

It brings the `comment` detection of `requirements_parser.bzl` closer to the spec described here:
- https://pip.pypa.io/en/stable/reference/requirements-file-format/#comments

1. Lines that begin with `#` are comments.
2. Content after (and including) ` #` is a comment.

Prior to this commit, a dependency like this would result in invalid `pip wheel` arguments:

```
requests @ https://github.com/psf/requests/releases/download/v2.29.0/requests-2.29.0.tar.gz#sha1=3897c249b51a1a405d615a8c9cb92e5fdbf0dd49
```

### DIFF
--- a/python/pip_install/private/test/requirements_parser_tests.bzl
+++ b/python/pip_install/private/test/requirements_parser_tests.bzl
@@ -63,6 +63,9 @@ FooProject==1.0.0
 # Comment
 BarProject==2.0.0 #Comment
 """).requirements)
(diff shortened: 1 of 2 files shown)


aignas (approved):
LGTM, thanks for adding a test!

chrislovecnm (approved):
LGTM

PR #1197, opened 2023-05-02
fix: coverage: rename html.py file to htmlx.py to avoid a clash while loading html deps

supercairos:
This is to avoid a clash when loading the html python std library;

See issue #1196


rickeylev:
See my comment on the issue -- the likely fix is to make sure sys.path is correct. If coverage's directly itself is on sys.path, that means all the files within coverage can potentially shadow a stdlib module, not just html.

aignas:
@chrislovecnm, this can be closed as #1196 was resolved.

PR #1204, opened 2023-05-04
test(bzlmod): explicitly enable bzlmod in the test harness

aignas:
Previously we would depend on the value of .bazelrc and this
change ensures that we are explicitly enable bzlmod via CLI
args. It seems that the `py_proto_library` integration tests
defined in the `//examples:BUILD.bazel` file were not running
using `bzlmod` before hand, however, they were correctly
executed in the CI.

Work towards #958.

PR #1205, opened 2023-05-04
test(pip_parse): add integration tests for macros in requirements.bzl

aignas:
The PR #1190 fixed incompatible_generate_aliases breaking the
`requirement` macro under `bzlmod`. This PR is a followup in
establishing the test infrastructure using the classic `bazel` only for
now.

Work towards #958.


aignas:
After #1155, #1176, #1204, we can add an integration test target which
would exercise the same `BUILD.bazel` files under `bzlmod` and we could
potentially remove `build_test` definitions in the bzlmod example added
in #1176.

chrislovecnm:
Ping

PR #1206, opened 2023-05-04
fix(update_deleted_packages.sh): allow to run from anywhere in the repo

aignas:
It seems that the tool was correctly trying to modify the `.bazelrc` at the
root of the repo, but the `find` execution would run in the `$PWD`. This
change ensures that the `find` is consistent with the file we are trying
to modify and allows the user to execute the script from anywhere in the
repo.

Tested:
1. Update the deleted packages with the version of the script at HEAD
1. Fix the bug
1. Rerun the script from the 'tests' sub-folder in the repo to ensure that
   running the script is noop.
1. Revert the changes to '.bazelrc' to reduce conflicts as other PRs are
   modifying it.

Work towards #958.


chrislovecnm (commented):
LGTM.  I have been fighting this for awhile.  Thanks!

rickeylev (approved):
I was also thinking a .bazelignore file might be another way to more easily manage the directories to ignore

aignas:
@rickeylev, yeah, the only reason why we are using this instead of `.bazelignore` is to support the `bazel_integration_test` rule, which allows one to include the example or the integration test directory in the test run.

That said, I noticed, that we are using the legacy version of the ruleset and we'd need to invest some time to port it to the current version that exists in bazel-contrib. We do not use the integration test rules in the CI and I personally find not using them on my local dev env due to how long it takes to execute the tests.

PR #1207, opened 2023-05-04
fix(update_deleted_packages.sh): wheels example should not be included in .bazelrc

### LINKED ISSUE #919: tools/bazel_integration_test/update_deleted_packages.sh is adding too much
# 🐞 bug report

This is a bug on an internal script.

### Description

Inside of the .bazelrc file it states: To update these lines, run tools/bazel_integration_test/update_deleted_packages.sh

## 🔬 Minimal Reproduction

Run the script and the wheel examples are being added.

## 🔥 Exception or Error

CI fails

## 🌍  Your Environment

**Operating System:**

WSL2

```console
$ bazelisk version
Bazelisk version: v1.14.0
Build label: 5.3.2
Build time: Wed Oct 19 18:22:12 2022 (1666203732)
Build timestamp: 1666203732
Build timestamp as int: 1666203732
```

**Rules_python version:**

master

**Anything else relevant?**

aignas:
This correctly handles the integration tests and examples that are a part of
the `rules_python` workspace and should not be included in the deleted packages
list.

This brings the changes made to the `.bazelrc` very close to what is in `main`,
but I would like to update this later once #1155 and #1205 are merged.

Fixes #919.

### DIFF
--- a/tools/bazel_integration_test/update_deleted_packages.sh
+++ b/tools/bazel_integration_test/update_deleted_packages.sh
@@ -15,11 +15,23 @@
 
 # For integration tests, we want to be able to glob() up the sources inside a nested package
 # See explanation in .bazelrc
+#
+# This script ensures that we only delete subtrees that have something a file
+# signifying a new bazel workspace, whether it be bzlmod or classic. Generic
+# algorithm:
+#   1. Get all directories where a WORKSPACE or MODULE.bazel exists.
+#   2. For each of the directories, get all directories that contains a BUILD.bazel file.
+#   3. Sort and remove duplicates.
 
 set -eux
 
 DIR="$(dirname $0)/../.."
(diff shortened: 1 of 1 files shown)


aignas:
Marking as a draft as it depends on #1206.

aignas:
Changed #1206 and this PR to not depend on each other for easier merging and review. Right now this PR can be rebased on #1206 without any conflicts, so it should be safe to merge independently.

chrislovecnm:
@dumb question, but can we run this using bazel and then find the directories more cleanly?

rickeylev (approved):

rickeylev:
It could be run using bazel, but it's such a simple shell script I don't think it would gain much. Bazel running for one workspace is likely to get an error if it tries to process the build, workspace, or module files of another workspace embedded workspace.

PR #1212, opened 2023-05-05
fix: Strip trailing newline from python output

illicitonion:
This is used to generate a path, which shouldn't have a trailing newline.


hrfuller (approved):
LGTM

hrfuller:
The test failure looks like a network flake, but I can't restart the build so maybe try force pushing to the branch.

chrislovecnm:
@hrfuller yes, I have an issue filed about the flake that CI hit.  @rickeylev can you nudge CI?

PR #1216, opened 2023-05-09
feat(bzlmod): Adding base_url attr to python.toolchain

### LINKED ISSUE #1172: Unable to specify base_url in python.toolchain() when using bzlmod
# 🐞 bug report

### Affected Rule

 The issue is caused by the rule: python:extensions.bzl

### Is this a regression?

 Yes, the previous version in which this bug was not present was: works fine without bzlmod

### Description

 While trying to prepare my test repo for bzlmod switch, I've noticed that it seems to be no longer possible to override base_url location for python toolchains. This is a super useful feature for someone working behind a corporate firewall and maintaining a mirror of python-build-standalone artefacts.

## 🔬 Minimal Reproduction

Following example is perfectly valid when using WORKSPACE.bazel:

```
python_register_toolchains(
    name = "python3_8_13",
    python_version = "3.8",
    base_url = "https://github.com/indygreg/python-build-standalone/releases/download",
)
```

but when defining a toolchain in bzlmod, there is simply no parameter to pass base_url value:

```
python = use_extension("@rules_python//python:extensions.bzl", "python")
python.toolchain(
    name = "python3_8_13",
    python_version = "3.8",
)
```

## 🔥 Exception or Error

<pre><code>

None

</code></pre>

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
Linux
  </code>
</pre>

<pre>
  <code>
Bazelisk version: v1.16.0
Build label: 6.1.1
Build time: Wed Mar 15 15:44:56 2023 (1678895096)
Build timestamp: 1678895096
Build timestamp as int: 1678895096
  </code>
</pre>

**Rules_python version:**

<pre>
  <code>
bazel_dep(name = "rules_python", version = "0.20.0")
  </code>
</pre>

**Anything else relevant?**

chrislovecnm:
The bazel_url attribute allows the user to set a custom location (URL) that hosts the Python binaries.

The base_url attribute allows users to override the URL passed into python_register_toolchains, via kwargs. This URL is then passed to python_repository, which enables the download of a Python installation package from a custom URL. 

Closes #1172

### DIFF
--- a/python/extensions.bzl
+++ b/python/extensions.bzl
@@ -24,15 +24,27 @@ load("@rules_python//python/private:toolchains_repo.bzl", "get_host_os_arch", "g
 def _python_impl(module_ctx):
     for mod in module_ctx.modules:
         for toolchain_attr in mod.tags.toolchain:
-            python_register_toolchains(
-                name = toolchain_attr.name,
-                python_version = toolchain_attr.python_version,
-                bzlmod = True,
-                # Toolchain registration in bzlmod is done in MODULE file
-                register_toolchains = False,
-                register_coverage_tool = toolchain_attr.configure_coverage_tool,
-                ignore_root_user_error = toolchain_attr.ignore_root_user_error,
-            )
+            if toolchain_attr.base_url:
+                python_register_toolchains(
+                    name = toolchain_attr.name,
+                    python_version = toolchain_attr.python_version,
+                    bzlmod = True,
+                    # Toolchain registration in bzlmod is done in MODULE file
+                    register_toolchains = False,
+                    register_coverage_tool = toolchain_attr.configure_coverage_tool,
+                    ignore_root_user_error = toolchain_attr.ignore_root_user_error,
+                    base_url = toolchain_attr.base_url,
+                )
+            else:
+                python_register_toolchains(
+                    name = toolchain_attr.name,
+                    python_version = toolchain_attr.python_version,
+                    bzlmod = True,
+                    # Toolchain registration in bzlmod is done in MODULE file
+                    register_toolchains = False,
+                    register_coverage_tool = toolchain_attr.configure_coverage_tool,
+                    ignore_root_user_error = toolchain_attr.ignore_root_user_error,
+                )
             host_hub_name = toolchain_attr.name + "_host_interpreter"
             _host_hub(
                 name = host_hub_name,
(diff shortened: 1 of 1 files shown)


python/extensions.bzl
```diff
@@ -24,15 +24,27 @@ load("@rules_python//python/private:toolchains_repo.bzl", "get_host_os_arch", "g
...
-            python_register_toolchains(
-                name = toolchain_attr.name,
-                python_version = toolchain_attr.python_version,
-                bzlmod = True,
-                # Toolchain registration in bzlmod is done in MODULE file
-                register_toolchains = False,
-                register_coverage_tool = toolchain_attr.configure_coverage_tool,
-                ignore_root_user_error = toolchain_attr.ignore_root_user_error,
-            )
+            if toolchain_attr.base_url:
```
chrislovecnm, line 27: @rickeylev not sure if there is a cleaner way to do this.  If we don't set the base_url when we call python.toolchain, we end up with an empty string passed in.  Then python_register_toolchains passes in an empty value to python_repository.  We may have a weird use case where an empty URL is allowed; if not, I can test for an empty URL in python_register_toolchains, which is fewer lines of code.

python/extensions.bzl
```diff
@@ -44,6 +56,14 @@ python = module_extension(
     tag_classes = {
         "toolchain": tag_class(
             attrs = {
+                "base_url": attr.string(
+                    mandatory = False,
+                    doc = """\
+The URL that is used to override rules_python DEFAULT_RELEASE_BASE_URL. This rule set downloads
+the different Python binaries from this website. Override this URL if you want to have rules_python
+download the hermetic versions of Python from a different website.
+""",
```
aignas, line 65: Should we use the bazel downloader for this instead?

See https://blog.aspect.dev/configuring-bazels-downloader

chrislovecnm: Great question! I asked the folks on #1172 to get their opinion. I think folks are using base_url, so if we are not going to support it people will have to migrate to this as well.

chrislovecnm:
https://github.com/bazelbuild/rules_python/issues/1172#issuecomment-1542320897

Folks are requesting feature parity

python/extensions.bzl
chrislovecnm: See https://github.com/bazelbuild/rules_python/issues/1172#issuecomment-1542320897

rickeylev:
Just to recap my post on the issue: An arg on the toolchain() tag class doesn't seem like the way to solve this. A separate tag class, probably only respected by the root module, to customize it would make more sense. Let's continue any design discussion on the issue and leave the PR for code review comments.

chrislovecnm:
Closing as we don't want to add an argument.

PR #1220, opened 2023-05-11
feat(bzlmod): support entry_point macro

### LINKED ISSUE #958: [bzlmod bug/regression] `entry_point` no longer working with blzmod

aignas:
Add `entry_point` macro to the repo generated by the `pip.parse` extension.
This works by using the canonical label literal, so should work without users
needing to add the spoke repos to the `use_repo` statement. We test this by
having an extra folder in the `bzlmod` example.

Fixes #958.


aignas:
@rickeylev, I like this approach better than #1189 at least for now, because it does not require users to eagerly fetch the repo just to access the entrypoint struct as the label for the entrypoint in this PR is constructed from the hub repo macro without any imports from the spoke repos.

This means that the users wanting to use the `entry_point` have to add an extra `use_repo` statement, but it may be a good enough trade-off here.

The rough edges that I found when porting a test from the `//examples/pip_parse` to `bzlmod`:
1. I could not import `runfiles` library if I depend on the `@rules_python//python/runfiles` under `bzlmod`. It seems to me that it could be because in the runfiles directory the `rules_python` exists as `rules_python~override` and our python import machinery is not repo mapping aware.
2. The `py_test_with_transition` does not work with `$(location :yamllint)` and I am not sure exactly why.

Should we raise a ticket for 1. in the `bazel` repo and 2. is a non-issue if 1. works and we are using the `runfiles` library?

aignas:
Switched to `$(rlocationpath` and everything started working, ready for the final review.

examples/bzlmod/MODULE.bazel
```diff
@@ -26,10 +27,15 @@ register_toolchains(
...
     name = "pip",
+    python_interpreter_target = "@python3_9_host_interpreter//:python",
     requirements_lock = "//:requirements_lock.txt",
     requirements_windows = "//:requirements_windows.txt",
 )
-use_repo(pip, "pip")
+use_repo(
+    pip,
+    "pip",
+    "pip_yamllint",
```
alexeagle, line 37:
```suggestion
    # The 'entry_point' macro will expand to reference this repository in "userspace",
    # so we have to declare that usage for bzlmod's strict repository visibility.
    "pip_yamllint",
```

alexeagle (commented):
This seems reasonable to me. FYI @fmeum as a subject matter expert here, in case there's some way to avoid the `use_repo` I'm not thinking of.

examples/bzlmod/entry_point/test_entry_point.py
```diff
@@ -0,0 +1,39 @@
...
+import subprocess
+import unittest
+
+
+class ExampleTest(unittest.TestCase):
+    def test_entry_point(self):
+        rlocation_path = os.environ.get("YAMLLINT_ENTRY_POINT")
+        assert rlocation_path is not None
+
+        entry_point = pathlib.Path("..") / rlocation_path
```
fmeum, line 26: This would be a great place to use `//python/runfiles` instead of just accessing the file directly as the latter probably doesn't work on Windows.

examples/bzlmod/MODULE.bazel
fmeum: `bazel test //...` in `examples/bzlmod` passes for me if I just delete it. The only reference is a canonical label literal, which doesn't require a `use_repo`.

python/pip_install/pip_repository_requirements_bzlmod.bzl.tmpl
```diff
@@ -22,3 +22,18 @@ def data_requirement(name):
...
+    ```
+    use_repo(
+        "pypi",
+        "pypi_yamllint",
+    )
+    ```
+    """
+    if not script:
+        script = pkg
+    return "@@%%NAME%%_{}//:rules_python_wheel_entry_point_{}".format(_clean_name(pkg), script)
```
fmeum, line 39: You should be able to use `@%%FRIENDLY_NAME%%` instead, where `FRIENDLY_NAME` is what the user passed in as `name`. You can get it by making the repository rule a macro that sets both the actual `name` attr and a `friendly_name` `attr.string()` to the value passed in as `name`.

aignas:
As @fmeum suggested, it seems that we can avoid the `use_repo` here and the example just works without it. Learned something new about how `bzlmod` works.

@fmeum, your last comment about the `%%FRIENDLY_NAME%%` works only if we have the `use_repo` statement, right? If the instructions to users are that they should not need to add additional `use_repo` statements, then the code should be good as is, right?

fmeum:
> @fmeum, your last comment about the `%%FRIENDLY_NAME%%` works only if we have the `use_repo` statement, right? If the instructions to users are that they should not need to add additional `use_repo` statements, then the code should be good as is, right?

Every repo created by a module extension can see all other repositories created by the extension without any `use_repo`s. If you return `Label("%%FRIENDLY_NAME%%...)`, that should also work. The currently approach looks alright too, so there is not need to change it.

examples/bzlmod/MODULE.bazel
```diff
@@ -18,6 +18,7 @@ python.toolchain(
 )
 use_repo(python, "python3_9")
 use_repo(python, "python3_9_toolchains")
+use_repo(python, "python3_9_host_interpreter")
```
rickeylev, line 21: Chris's PR has been merged. To get the toolchain interpreter as a repo-available interpreter, do:

```
interpreter = use_extension("@rules_python//python/extensions:interpreter.bzl", "interpreter")
interpreter.install(
  name = "python3_9_interpreter",
  python_name = "python3_9",  # the name from the toolchain() call
)
use_repo(interpreter, "python3_9_interpreter")

pip.parse(..., python_interpreter_target = "@python3_9_interpreter//:python")
```

examples/bzlmod/entry_point/test_entry_point.py
```diff
@@ -0,0 +1,41 @@
...
+import subprocess
+import unittest
+
+from python.runfiles import runfiles
+
+
+class ExampleTest(unittest.TestCase):
+    def test_entry_point(self):
+        rlocation_path = os.environ.get("YAMLLINT_ENTRY_POINT")
+        assert rlocation_path is not None
```
rickeylev, line 26: Please add a descriptive message if the assert fails

examples/bzlmod/entry_point/BUILD.bazel
```diff
@@ -0,0 +1,22 @@
...
+)
+
+py_test(
+    name = "entry_point_test",
+    srcs = ["test_entry_point.py"],
+    data = [
+        ":yamllint",
+    ],
+    env = {
+        "YAMLLINT_ENTRY_POINT": "$(rlocationpath :yamllint)",
```
rickeylev, line 16: Oh neat. TIL about `$(rlocationpath)`.

alexeagle (approved):

rickeylev (changes requested):
Mostly LGTM. Just need to update to head to reflect Chris's recent changes to how to setup bzlmod stuff.

aignas:
@rickeylev, the build is fixed, so should be ready to be merged.

rickeylev (approved):

aignas:
Squashed everything into a single commit and rebased on the current `main`. Tested the example with:
```
$ bazel test ...
INFO: Analyzed 62 targets (121 packages loaded, 5685 targets configured).
INFO: Found 55 targets and 7 test targets...
WARNING: cleared convenience symlink(s) bazel-bin, bazel-testlogs because their destinations would be ambiguous
INFO: Elapsed time: 67.466s, Critical Path: 10.95s
INFO: 90 processes: 26 internal, 62 darwin-sandbox, 2 local.
INFO: Build completed successfully, 90 total actions
//:all_requirements                                                      PASSED in 0.8s
//:all_wheels                                                            PASSED in 0.3s
//:requirements_test                                                     PASSED in 10.1s
//:test                                                                  PASSED in 1.1s
//:test_with_transition                                                  PASSED in 1.4s
//entry_point:entry_point_test                                           PASSED in 2.5s
//runfiles:runfiles_test                                                 PASSED in 1.4s

Executed 7 out of 7 tests: 7 tests pass.
There were tests whose specified size is too big. Use the --test_verbose_timeout_warnings command line option to see which ones these are.
```

PR #1227, opened 2023-05-15
fix: Use Bazel's built-in zstd support in _python_repository_impl.

jiawen:
Bazel has built-in support for zstd [since 5.1](https://github.com/bazelbuild/bazel/pull/15087). This simplifies the logic and makes python_repository work on systems where building from source fails.

### DIFF
--- a/python/repositories.bzl
+++ b/python/repositories.bzl
@@ -108,55 +108,11 @@ def _python_repository_impl(rctx):
     release_filename = rctx.attr.release_filename
     url = rctx.attr.urls or [rctx.attr.url]
 
-    if release_filename.endswith(".zst"):
-        rctx.download(
-            url = url,
-            sha256 = rctx.attr.sha256,
-            output = release_filename,
-        )
-        unzstd = rctx.which("unzstd")
-        if not unzstd:
-            url = rctx.attr.zstd_url.format(version = rctx.attr.zstd_version)
-            rctx.download_and_extract(
-                url = url,
-                sha256 = rctx.attr.zstd_sha256,
-            )
-            working_directory = "zstd-{version}".format(version = rctx.attr.zstd_version)
-            make_result = rctx.execute(
-                ["make", "--jobs=4"],
-                timeout = 600,
-                quiet = True,
-                working_directory = working_directory,
-            )
-            if make_result.return_code:
-                fail_msg = (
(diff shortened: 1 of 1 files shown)


f0rmiga (approved):

jiawen:
Thanks @f0rmiga! I'll obviously make CI pass before asking to merge.

rickeylev:
Can you please sync to head and resolve the merge conflicts?

CI also reports there are failures, something relating to setting file permissions for a directory that doesn't exist? They failed even after retrying, so they're likely legit.

jiawen:
My apologies - my day job got in the way. Draft sounds good - looks like I have some fun chmod issues to resolve.

jiawen:
Argh, this PR uncovered some grungy indygreg packaging details. The existing code works because:

* The `.tar.gz` files are all `-install_only`, so the contents are `["python/bin", "python/include", "python/lib", and "python/share"]`.
* But the `zst` versions are "full". The contents are:
  * `["python/build/lib", "python/build/Modules",  ...]` (a couple of others)
  * `["python/install/bin", "python/install/include", "python/install/lib", and "python/install/share"]`
  * (A few other files and dirs)

So the existing code happens to work because:
* It uses `rctx.download_and_extract` in the `.tar.gz` case, which unpacks `python/lib` into the right place.
* The `zstd` path uses `tar --strip-components=2`, which also unpacks the contents two levels deep into the right place. But interestingly, it relies on tar's behavior to *merge* the two `lib` directories together. It's obviously possible to do this with `rctx.execute(["cp -R ...", ])` but this is somewhat irritating.

@rickeylev Any suggestions?

aignas:
FYI, the `.tar.gz` files were created due to https://github.com/indygreg/python-build-standalone/issues/79 if I remember correctly.

rickeylev:
Perhaps the `strip_prefix` arg or `rename_files` arg of `downloaded_and_extract`? I don't really have ideas otherwise.

If a separate command has to be run after download_and_extract to merge the two directories, I'm not sure it'll be a net win. Maybe `cp --link`?

> This ... makes python_repository work on systems where building from source fails.

I don't follow -- how does this PR fix building from source? This PR basically removes the need to have the zstd and tar tools locally available. What does that have to do with building from source?

Issue #1229, opened 2023-05-16
bzlmod Multiple Python Versions


chrislovecnm:
Currently, rules_python does not have bzlmod support to use multiple Python versions.  The following example demonstrates the current "WORKSPACE" support: https://github.com/bazelbuild/rules_python/tree/main/examples/multi_python_versions

### Dependencies

In order to have support we need extensions that create the following data relationships:

First, we have 2 or more versions of Python; say Python 3.9, 3.10, and 3.11.
Second, each of those Python versions has a host-specific interpreter.  You can see the `interpreter` extension used in the following code:

https://github.com/bazelbuild/rules_python/blob/16126d0ebfee074a3a8fe216b20cc19e1b3603c1/examples/bzlmod_build_file_generation/MODULE.bazel#L77 

This extension creates a symlink to the host-specific interpreter so that pip can utilize the correct interpreter. 

So we have 

- multiple Python versions
- multiple Python host (OS and platform) specific Interpreter binaries
- symlinks to each of those binaries

The above is a one to one to one relationship.

Lastly, we have requirements files.   Currently _ONLY_ one requirement lock file per Python version is supported.  This is a limitation since Windows or Linux could need different requirements lock f files.

chrislovecnm:
## Current multiple_pip_parse call

The next part of the workspace file is:

https://github.com/bazelbuild/rules_python/blob/16126d0ebfee074a3a8fe216b20cc19e1b3603c1/examples/multi_python_versions/WORKSPACE#L36

This macro is dynamically generated by the `python_register_multi_toolchains` call.

I understand the design of how to pass in the labels to the interpreters and the labels for the requirements files.  That is quite doable.

The main challenge that I am trying to get my head around is can we use 

https://github.com/bazelbuild/rules_python/blob/16126d0ebfee074a3a8fe216b20cc19e1b3603c1/python/pip.bzl#L252

Directly.   I do not understand the differences between 

https://github.com/bazelbuild/rules_python/blob/16126d0ebfee074a3a8fe216b20cc19e1b3603c1/python/pip_install/pip_repository.bzl#L352

and

How https://github.com/bazelbuild/rules_python/blob/16126d0ebfee074a3a8fe216b20cc19e1b3603c1/python/pip_install/pip_repository.bzl#L429 works

Vs how

https://github.com/bazelbuild/rules_python/blob/16126d0ebfee074a3a8fe216b20cc19e1b3603c1/python/pip.bzl#L252

Works.

[...]

PR #1230, opened 2023-05-17
feat(bzlmod): Allowing multiple python.toolchain extension calls

chrislovecnm:
We do this work for two reasons.

First, we must support Module dependencies and sub-modules using `python.toolchain`. There are already two known instances of sub-modules setting up a Python toolchain and colliding with another module (nanobind and rules_testing both run into this).

Second, the upcoming multi-version support is going to work by having each `python.toolchain()` call register its particular version with the extra toolchain constraint. This also helps unify the version-aware and non-version-aware code paths (the non-version aware paths are just version-aware with a single version registered as the default)

This commit implements various business logic in the toolchain class.

Toolchains in Sub Modules

It will create a toolchain in a sub-module if the toolchain
of the same name does not exist in the root module.  The extension stops name
clashing between toolchains in the root module and sub-modules.
You cannot configure more than one toolchain as the default toolchain.

Toolchain set as the default version.

This extension will not create a toolchain in a sub-module
if the sub-module toolchain is marked as the default version.  If you have
more than one toolchain in your root module, you need to set one of the
toolchains as the default version.  If there is only one toolchain, it
is set as the default toolchain.

See #1229 for more information


python/extensions/python.bzl
```diff
@@ -17,40 +17,132 @@
...
+        fail("""Unable to find a default toolchain in the root module.  
+Please define a toolchain that has default_version set to True.""")
+
+    # Create the toolchains in the submodule(s).
+    for name, toolchain_attr in sub_toolchains_map:
+        # A sub module cannot have a toolchain that is marked as the
+        # default version. TODO: should we create the toolchain anyways,
+        # but set the default version to False?
+        if toolchain_attr.default_version:
+            fail("""Not able to create toolchain named: {}.  This toolchain exists
```
rickeylev, line 86: Per my main comment: create the repo, just force default=False

rickeylev (changes requested):

rickeylev:
Mostly LGTM.

> This extension will not create a toolchain in a sub-module if the sub-module toolchain is marked as the default version.
> TODO: should we create the toolchain anyways,
> but set the default version to False?

Yeah, I'm not sure about this. My main concern is, because the repo will sometimes exist and sometimes not, it'll cause headaches. There are two cases I have in mind:

Case 1: Leaky development-only dependencies. I'm doing development. I add `python.toolchain(X, default=True); use_repo(X)`. I run my tests, things work. I deploy. My users break, because my `use_repo(X)` call fails -- X was never created. I _should_ have marked something dev_only=True, but forgot; instead of some extra work being done, things have broken.

Case 2: Some non-dev-only reason you want the repo, but don't need it to be registered as a toolchain? I don't know what this reason would be....Aha -- won't pip.parse stop working? For example, my module might have a tool, and my tool has dependencies, which it gets through pip.parse() and python_interpreter_target, so the toolchain repo has to be available.

Ok, yeah, I think we need the repo to be created. Just don't let it be registered as a toolchain when we move the register_toolchains() call into rules_python itself. So in this PR, create the repo. In the PR when we refactor where register_toolchains() is called, update the logic to respect this.

This does expose another issue, though: the submodule will use pip.parse() with one interpreter, but its py_binary targets might end up using a different one through toolchain resolution. I think the best we can do here is provide a "default interpreter" repo, so that the sub-module has a way to pass the same interpreter the default toolchain will use into its pip.parse calls. (Something for another PR, though). A default-interpreter repo might also simplify how much config the user has to do in their MODULE file? Haven't thought deeply here.

cc @fmeum as a second pair of eyes on my thinking here.

> If you have more than one toolchain in your root module, you need to set one of the
> toolchains as the default version.

For posterity, recap of the convo Chris and I had:

Having a dependency allowed to change your Python version is bad because that dependency might be far away. A consumer might unexpectedly change their version updating a seemingly unrelated dependency. A producer might inadvertently break someone changing their python version. Neither has a good insight into the transitive effects.

We talked about the idea of taking the highest version of any toolchains marked default -- we rejected this idea for two reasons. The first is the same rationale as above. The second is: if you're introducing a new python version to a large project, it's beneficial to get the toolchain setup and installed, but leave it in an unused-by-default state. This allows work to progress to test and validate things incrementally. When you're finally ready, you can switch the default.

python/repositories.bzl
```diff
@@ -486,9 +485,15 @@ def python_register_toolchains(
         set_python_version_constraint: When set to true, target_compatible_with for the toolchains will include a version constraint.
         tool_versions: a dict containing a mapping of version with SHASUM and platform info. If not supplied, the defaults
             in python/versions.bzl will be used.
-        bzlmod: Whether this rule is being run under a bzlmod module extension.
         **kwargs: passed to each python_repositories call.
     """
+
+    # If we have @@ we have bzlmod
+    bzlmod = str(Label("//:distribution")).startswith("@@")
```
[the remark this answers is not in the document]
chrislovecnm: Cool! What is the difference?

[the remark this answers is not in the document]
rickeylev: Just to make it more clear the target name doesn't matter. Label() just parses the string, it doesn't validate the target exists.

python/extensions/python.bzl
```diff
@@ -17,40 +17,132 @@
...
+        # a toolchain in the root module. This will cause name clashing.
+        if name in toolchain_names:
+            _print_warn("""Not creating the toolchain from sub module, with the name {}. The root
+ modhas a toolchain of the same name.""".format(toolchain_attr.name))
+            continue
+        toolchain_names.append(name)
+        _python_register_toolchains(toolchain_attr, True)
+
+    # We register the default toolchain last.
+    _python_register_toolchains(default_toolchain, False)
```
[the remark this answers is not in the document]
rickeylev: There's two aspects to this: ordering, and compatiblity.

Usually the order targets are defined in doesn't matter, but toolchains are an exception. Basically, all the `native.toolchain` targets that are found via `register_toolchains()` calls are searched in the order they're encountered.

For example, say we had this build file:

```
toolchain(name="foo", ...)
toolchain(name="bar", ...)
```

And this module:

```
register_toolchains("//:all")
```

This will record a list of `[foo, bar]`. When someone asks for a toolchain, it'll first check `foo`, and, if it's compatible, will use it. Otherwise it moves onto the next item in the list.

If we change the BUILD file to put "bar" before "foo", bar will be checked first. If we change the register_toolchains call to explicitly name the targets, e.g. `register_toolchains("//:bar", "//:foo")`, that will also make bar checked first.

Next, compatibility: when iterating over the toolchains, the first compatible toolchain is used, where "compatible" means "are all of the toolchain's constraints satisfied?". For example, say we had this build file:

```
toolchain(name="default_linux", compatible_with=["//os:linux"])
toolchain(name="py38_linux", compatible_with=["//os:linux", "//py:py38"])
```

And we do `register_toolchains("//:all")`, so the ordering is `[default_linux, py38_linux]`. Note how the first toolchain has fewer constraints than the second -- it just cares about matching linux.

When a multi-version aware rule is used, it's constraints are `(os=linux, py=py38)`. We want it to use the `py38_linux` toolchain. But, `default_linux` is checked first and will be matched because all its constraints (it just needs linux) are satisfied, and thus it uses the wrong toolchain.

The toolchain docs explain the algorithm in detail: https://bazel.build/extending/toolchains#toolchain-resolution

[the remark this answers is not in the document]
fmeum: I think that targets within a package are ordered lexicographically, not in declaration order. That's why rules_go appends a prefix such as `0001_` to all toolchain definitions.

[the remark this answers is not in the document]
rickeylev: You're right! Is that documented somewhere? I couldn't find a mention of `:all` or the lexicographic ordering it uses.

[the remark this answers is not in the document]
fmeum: I don't think so, @Wyverald dug this out of the code when I asked him about it.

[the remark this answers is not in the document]
Wyverald: https://cs.opensource.google/bazel/bazel/+/refs/heads/master:src/main/java/com/google/devtools/build/lib/packages/Package.java;drc=665a3241f4acf2291d157c76a413d872ae59241e;l=704

All the target pattern resolution stuff eventually go to `Package.getTargets()`, and it's an `ImmutableSortedSet` on the target name.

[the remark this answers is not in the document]
rickeylev: @chrislovecnm No, we don't need to address all of this in this PR because there isn't a hub for rules_python to call register_toolchains on.

The changes for this PR are:
 1. Remove L99, the python_register_toolchains() call for the default toolchain.
 2. Remove the skip-if-default logic up around L56 (so that the repo for the default version is still created)

When we do the PR to automatically register toolchains, then we'll need to worry about what this comment thread is about.

examples/bzlmod_build_file_generation/MODULE.bazel
```diff
@@ -46,40 +46,81 @@ python = use_extension("@rules_python//python/extensions:python.bzl", "python")
 
 # This name is passed into python.toolchain and it's use_repo statement.
 # We also use the same name for python.host_python_interpreter.
-PYTHON_NAME = "python3"
+PYTHON_NAME_39 = "python_39"
+
+PYTHON_TOOLCHAINS_39 = PYTHON_NAME_39 + "_toolchains"
```
chrislovecnm, line 51: I need to move this into the bzlmod example, not here. I doubt gazelle works with multiple toolchains.

chrislovecnm (commented):
Need to update the example.

rickeylev (approved):
Mostly LGTM; a couple doc nits. Per chat: one of those CI failures looks legit. I've retried it 4 times now and it keeps failing. It's odd that its only failing on windows, though. Our guess is due to the multiple versions in the example; lets just pull that out for now.

I'll give approval so you can merge when ready.

Two closing thoughts; these don't need to be addressed in this PR. I'm just posting them for posterity.

First: I think the creation of the default repo is sort of wrong, but things end up working out anyways. Basically, because it's not being created with version_constraint=True, there's no toolchain that is multi-version aware. However, all the others are, and the default ends up last. The net effect is looking for the same version as the default will fail to match all the versioned toolchains, then land on the default, and end up with the proper toolchain. This is fine for now -- no need to address it in this PR; when we create the toolchain repo, we can address it in that PR.


Second: I'm think we should have a way for the default interpreter to be referenced by the interpreter extension. Because: lets say you're a sub module and you're OK with using the default python version. How do you configure your pip.parse() call to use it? You can't -- you don't know what the root module is using for the default. So we should address that somehow.

examples/bzlmod/MODULE.bazel
```diff
@@ -11,17 +11,21 @@ local_path_override(
...
     python_version = "3.9",
 )
-use_repo(python, "python3_9")
-use_repo(python, "python3_9_toolchains")
+use_repo(python, PYTHON_NAME)
+use_repo(python, PYTHON_TOOLCHAINS)
 
 register_toolchains(
-    "@python3_9_toolchains//:all",
+    "@{}//:all".format(PYTHON_TOOLCHAINS),
```
aignas, line 28: nit: I find the usage of constants here making it harder to read. Even though the definition and the usage of the constants is nearby, my eyes need to jump to lines 14 and 16 when reading line 28. Having the text verbatim would be better at the expense of very little duplication.

examples/bzlmod_build_file_generation/MODULE.bazel
aignas: Yeah, it won't work, so better to keep the `bzlmod` `gazelle` example smaller.

python/extensions/private/interpreter_hub.bzl
```diff
@@ -21,6 +21,7 @@ _build_file_for_hub_template = """
 INTERPRETER_LABELS = {{
 {lines}
```
aignas, line 22: nit: `{lines}` is not as descriptive as `{interpreter_labels}`.

python/extensions/python.bzl
```diff
@@ -17,35 +17,117 @@
...
-                ignore_root_user_error = toolchain_attr.ignore_root_user_error,
-            )
-
-            # We collect all of the toolchain names to create
-            # the INTERPRETER_LABELS map.  This is used
-            # by interpreter_extensions.bzl
-            toolchains.append(toolchain_attr.name)
+            # If we are in the root module we always register the toolchain.
+            # We wait to register the default toolchain till the end.
+            if mod.is_root:
```
aignas, line 51: nit: consider inverting the if statement and using `continue` statement to reduce the indentation for the code that matters.

python/extensions/python.bzl
```diff
@@ -17,35 +17,117 @@
 load("@rules_python//python:repositories.bzl", "python_register_toolchains")
 load("@rules_python//python/extensions/private:interpreter_hub.bzl", "hub_repo")
```
aignas, line 18: minor: having `_interpreter_hub_repo = "hub_repo"` here would make the code below more readable.

python/extensions/python.bzl
```diff
@@ -17,35 +17,117 @@
...
+            if mod.is_root:
+                toolchain_names.append(toolchain_attr.name)
+
+                # If we have the default version or we only have one toolchain
+                # in the root module we set the toolchain as the default toolchain.
+                if toolchain_attr.is_default or len(mod.tags.toolchain) == 1:
+                    # We have already found one default toolchain, and we can
+                    # only have one.
+                    if default_toolchain_name != None:
+                        fail("""We found more than one toolchain that is marked
```
aignas, line 60: Consider printing to the user the toolchains that are set default to make fixing of the error quicker.

aignas (approved):
Left some comments on readability. Thanks for the PR, in general LGTM other than the `gazelle` example modifications which you commented on yourself.

examples/bzlmod/MODULE.bazel
chrislovecnm: If you look at the example now, we now have multiple versions, and I find that having variables causes me less grief because of typos.  Do you have a compromise?

python/extensions/python.bzl
chrislovecnm: Please provide more context.

chrislovecnm:
So this PR is passing CI now, besides a known OSX flake.  For some reason, I am getting an error locally on Windows that I saw in a flake earlier.  Moreover, the example/bzlmod/.bazelrc did not enable Windows symlinks, so it should not have passed CI.  I am confused. cc: @rickeylev @aignas

examples/bzlmod_build_file_generation/MODULE.bazel
chrislovecnm: I added the code to the examples/bzlmod example and found a bug because I started using submodules.

python/extensions/python.bzl
rickeylev: I think "interpreter hub" and "toolchain hub" are sufficiently clear.

"interpreter hub" isn't great, but we couldn't think of a better name for it. The interpreter hub name is internal, so we can rename it if we think of a better name.

PR #1231, opened 2023-05-17
fix(toolchain): set correct return attrs to remove non-hermeticity warning

aignas:
Fixes the incorrect debug statement when downloading the toolchain for the first
time asking users to replace "urls" with "url" in the toolchain definition which
we maintain:
```
$ bazel build ...
... indicated that a canonical reproducible form can be obtained by modifying ...
```

Even though this is seen only once by the user, it may confuse them.

Summary of changes:
- refactor(toolchain): rename a local variable url -> urls
- fix(toolchain): set url return attrs correctly

### DIFF
--- a/python/repositories.bzl
+++ b/python/repositories.bzl
@@ -348,7 +348,7 @@ py_runtime_pair(
     rctx.file(STANDALONE_INTERPRETER_FILENAME, "# File intentionally left blank. Indicates that this is an interpreter repo created by rules_python.")
     rctx.file("BUILD.bazel", build_content)
 
-    return {
+    attrs = {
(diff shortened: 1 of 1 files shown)


chrislovecnm:
Looks like https://buildkite.com/bazel/rules-python-python/builds/4855#0188271d-8328-4162-865b-df228a71e1ce is a flake that I have not seem before.

chrislovecnm (approved):
LGTM.  If you can push again so that we can see if CI is happy?

aignas:
The CI failed with a flake in the `pip_annotations` test, which I think may be unrelated.

PR #1235, opened 2023-05-18
feat(gazelle): allow passing multiple files to the manifest generator

aignas:
For certain workflows it is useful to calculate the integrity hash of
the manifest file based on a number of requirements files. The
requirements locking is usually done by executing a script on each
platform and having gazelle manifest generator be aware that more than
one requirements file may affect the outcome (e.g. the wheels that get
passed to modules map may come from `multi_pip_parse` rule) is generally
useful.

This change modifies the `generator` to support multiple
`--requirements` flag specification and uses all files that got passed
to generate the final integrity hash.

NOTE: This may require users to regenerate the `gazelle_python.yaml` files due
to a change in integrity, otherwise this is a backwards compatible change.

### DIFF
--- a/examples/build_file_generation/BUILD.bazel
+++ b/examples/build_file_generation/BUILD.bazel
@@ -42,7 +42,10 @@ gazelle_python_manifest(
     name = "gazelle_python_manifest",
     modules_mapping = ":modules_map",
     pip_repository_name = "pip",
-    requirements = "//:requirements_lock.txt",
+    requirements_files = [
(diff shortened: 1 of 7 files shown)


aignas:
This is probably not the best way, because we can use multiple requirements file and a `genrule` to generate a single requirements file that can be used in this function. That way the `hash` will change whenever any of the input requirements files changes, but we will be still correct. We can also strip away the comments for them to not affect the hash value.

PR #1246, opened 2023-06-02
cleanup: Set toolchain target_setting directly instead of via inline ternary

rickeylev:
The generated toolchain BUILD file is confusing to read because it relies on a ternary expression in the BUILD file to set the `target_settings` attribute. This makes debugging harder because, upon first look, all the toolchains appear to have the version constraint set. It's only upon closer inspection that you can see the 1-character difference of "False" vs "True" embedded into the middle of a line amongst other similar looking lines.

Also:
  * Adds a bit of validation logic for the `set_python_version_constraint` argument
    because it's conceptually a boolean, but is passed as a string, so is prone to
    having an incorrect value passed.
  * Documents the `set_python_version_constraint` arg, since it has a particular range
    of values it accepts.

### DIFF
--- a/python/private/toolchains_repo.bzl
+++ b/python/private/toolchains_repo.bzl
@@ -46,18 +46,30 @@ def python_toolchain_build_file_content(
     Args:
         prefix: Python toolchain name prefixes
         python_version: Python versions for the toolchains
-        set_python_version_constraint: string "True" or "False"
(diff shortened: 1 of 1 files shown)


chrislovecnm (approved):
LGTM - good thing I left the comment about why we have a string instead of a boolean.  I already forgot why I wrote the darn code.

PR #1248, opened 2023-06-02
chore(pip): Code simplification

chrislovecnm:
Working on doing some cleanup of the pip_repository
  and pip extension code. This refactor merges the two
  seperate code paths we had for bzlmod vs. non-bzlmod
  implementation.
  
This commit also includes refactoring so that we do not
have multiple for loops over the same lists and other
code cleanup.


python/pip_install/pip_repository.bzl
```diff
@@ -474,18 +478,26 @@ def _pip_repository_impl(rctx):
     if rctx.attr.python_interpreter_target:
         config["python_interpreter_target"] = str(rctx.attr.python_interpreter_target)
 
+    # TODO can we get the bzlmod maco_tmpl and the pip
```
arrdem, line 481: Instant merge conflict with the cycle machinery although I like the cleanup and would love to see this get down to one template. How do you want to sequence these changes - and apologies that I haven't had time to respond to the change requests on that patch.

chrislovecnm: I am not certain how to get this down to one template. I think we can simplify it some. I’m working on figuring some changes for bzlmod and through those changes I would like to figure out how to get the two code paths closer.

chrislovecnm: > How do you want to sequence these changes - and apologies that I haven't had time to respond to the change requests on that patch.

No worries.  We will have to figure out the merging process.  I don't think I will have more code, and I can merge yours to mine if needs be.  I am working with others on refactoring bzlmod pip, which is why I did this work.  Getting my head around this code base more.

python/pip_install/pip_repository.bzl
```diff
@@ -349,46 +349,56 @@ alias(
...
+    macro_tmpl = ""
+    incompat_aliases = rctx.attr.incompatible_generate_aliases
 
     # NOTE: we are using the canonical name with the double '@' in order to
     # always uniquely identify a repository, as the labels are being passed as
     # a string and the resolution of the label happens at the call-site of the
     # `requirement`, et al. macros.
-    if rctx.attr.incompatible_generate_aliases:
+    if incompat_aliases:
+        _pkg_aliases(rctx, repo_name, pkgs.bzl_packages)
```
chrislovecnm, line 383: @rickeylev can we get the bzlmod and non-bzlmod templates the same?

python/pip_install/pip_repository.bzl
```diff
@@ -474,18 +478,25 @@ def _pip_repository_impl(rctx):
     if rctx.attr.python_interpreter_target:
         config["python_interpreter_target"] = str(rctx.attr.python_interpreter_target)
 
+    # TODO can we get the bzlmod maco_tmpl and the pip
+    # macro_tmpl the same?
     if rctx.attr.incompatible_generate_aliases:
-        _pkg_aliases(rctx, rctx.attr.name, bzl_packages)
+        all_reqs_tmpl = "@{name}//{{}}".format(name = rctx.attr.name)
```
chrislovecnm, line 484: This is the other format @rickeylev

rickeylev:
re: unifying the two styles of addressing the target: I suspect this has to do with repo visibility and the non-aliased code path.

From what I can see, it looks like, in WORKSPACE style, doing `requirement("foo")` with aliases disabled would up with `@pypi_foo//foo:pkg`. A direct translation of that to bzlmod would mean `@pypi_foo` would have to be `use_repo()`d, which doesn't generalize well. My point is: The `@pypi_foo`-style labels can't really be supported in bzlmod, so that style is workspace-specific.

I think the only thing left do is make both impls use the same template, and that template is basically just the bzlmod template, i.e., it basically just does `"%MACRO_TMPL%".format(name, "pkg")`, and have the repo rule determine what `%MACRO_TMPL%` is.

As an aside, I don't immediately see why the bzlmod code has a non-alias based code path at all? It's always put into one repo, the only difference seems to be e.g. `@pypi//:foo_pkg` vs `@pypi//foo:pkg`; I don't grok why that difference matters.

rickeylev (approved):

python/pip_install/pip_repository.bzl
chrislovecnm: @arrdem I think we have even more of a merge conflict now.

aignas:
The foo_pkg convention was the initial convention, which meant that we had one big file with all of the aliases. foo:pkg convention was the best thing we could think of to support gazelle and bzlmod / non-bzlmod codebases at the same time and lastly, have a nice human readable label.

python/pip_install/pip_repository_requirements.bzl.tmpl
```diff
@@ -18,28 +18,34 @@ def _clean_name(name):
...
+
+    return ep
 
 def _get_annotation(requirement):
     # This expects to parse `setuptools==58.2.0     --hash=sha256:2551203ae6955b9876741a26ab3e767bb3242dafe86a32a749ea0d78b6792f11`
     # down to `setuptools`.
     name = requirement.split(" ")[0].split("=")[0].split("[")[0]
     return _annotations.get(name)
 
+# TODO name needs to change
```
aignas, line 48: Could you elaborate here, please?

python/pip_install/pip_repository.bzl
```diff
@@ -349,100 +349,74 @@ alias(
...
-of 'requirements' no resolve will take place and pip_repository will create individual repositories for each of your dependencies so that
-wheels are fetched/built only for the targets specified by 'build/run/test'.
-""",
-    ),
-    "requirements_windows": attr.label(
-        allow_single_file = True,
-        doc = "Override the requirements_lock attribute when the host platform is Windows",
-    ),
-    "_template": attr.label(
-        default = ":pip_repository_requirements_bzlmod.bzl.tmpl",
```
aignas, line 424: What is more, since there are subtle differences, it is easy to break one or the other during refactorings. Having separate templates ensures that we are explicit with what we support.

python/pip_install/pip_repository_requirements.bzl.tmpl
chrislovecnm: I have it working now.  It was a note while I migrated.

chrislovecnm:
@aignas, per your advice I choose to not remove the second template.  
@rickeylev CI is failing with one of its usual fakes

rickeylev:
Sorry, looking at it now, I'm not really keen on this refactor overall. The reason I asked for separate bzlmod and workspace repo rules in the original PR is that they are almost entirely different implementations. They share a superficial structure, but have different attributes and internal logic. The net result is the implementation is basically a big if-else, and the rule definition itself carries several extra attributes. For example:

* `macro_tmpl` is used in both cases, but has a unique value for each
* The `all_whl_reqs_tmpl` variable is workspace-specific
* `all_requirements` and `all_whl_requirements` are used in both cases, but each case
  uses a separate algorithm to compute the value.
* Both expand a template, but each has a different template, stored in a different attribute
  that has different template variable names.

Of the ~100 lines of code, they actually only share about three lines of code: calling `_pip_get_packages()`, checking the incompatible flag (which I'm still not convinced bzlmod even needs to do), and the call to write the build file; these are all very trivial lines.

+1 to factoring out `_pip_get_packages`, but not so much on the rest.

chrislovecnm:
Closing as this refactor does not make complete sense with the pip changes

PR #1254, opened 2023-06-06
feat(bzlmod)!: Calling pip multiple times allowing for multiple Python versions

chrislovecnm:
Implementing the capability to call pip bzlmod extension multiple times.  We can now call the pip extension with the same hub name multiple times.  This allows a user to have multiple different requirement files based on the Python version.

With workspace, we need incompatible aliases because you get @pip//tabulate or @pip_tabulate//:pkg.
The requirement() macro hides this, but changing the flag becomes a breaking change if you've manually referenced things. With bzlmod, though, the @pip_tabulate style isn't a realistic option (you'd have to use_repo everything, which is a huge pain).

So we have chosen to have `@pip//tabulate`.

This commit implements that capability for pip.parse to determine the Python version from programmatically
the provided interpreter.

See 

https://github.com/bazelbuild/rules_python/blob/76aab0f91bd614ee1b2ade310baf28c38695c522/python/extensions/pip.bzl#L88

For more in-depth implementation details.

INTERFACE CHANGE::

- pip.parse argument python_version is introduced but not required. When possible, it is
  inferred.

BREAKING CHANGE:
  * `pip.parse()` extension: the `name` attribute is renamed `hub_name`. This is to reflect
    that the user-provided name isn't unique to each `pip.parse()` call.   We now have a hub that is created, and each
   pip.parse call creates a spoke.
  * `pip.parse()` extension: the `incompatible_generate_aliases` arg is removed; behavior
     has changed to assume it is always True.
  *  `pip.parse()` calls are collected under the same `hub_name` to support multiple Python versions under the same resulting repo name (the hub name0.

Close: #1229


rickeylev:
couple quick comments (only skimmed half of it thus far):

We need to keep a non-versioned compile_pip_requirements() call around. It's testing that the non-versioned variant of the rule, the one that uses the default python version, is working. Similarly, we'll want a non-versioned usage in the module config.

I see the pip extension is loading the hub? Originally we decided to avoid doing that, but I'm not strongly tied to that idea. Loading it will probably make life easier and we could mostly avoid having the interpreter extension as an intermediary if do, I think.

chrislovecnm:
@rickeylev we have a non versioned requirement call here: https://github.com/bazelbuild/rules_python/blob/28e15c2092a9512064e4b2a97a0fa3d1a83bc5ae/examples/bzlmod_build_file_generation/BUILD.bazel#L18

In regards to the hub dependency. We need to pass through the python default version, so I put that in the hub.  I can create a different repo rule and put it in there if you like.

Also, any idea for getting the python version from the interpreter? I could not figure out how to map that.

rickeylev:
> In regards to the hub dependency. We need to pass through the python default version, so I put that in the hub. I can create a different repo rule and put it in there if you like.

I think we can leave it in the hub? I don't see a downside. The hub being loaded is an internal detail, so it doesn't change our API surface. The hub is small, so loading it is cheap. While the hub has some label references to the platform repos, loading the hub doesn't cause the platform repos to load, so its still cheap.

> Also, any idea for getting the python version from the interpreter? I could not figure out how to map that.

Why does that need to be looked up? It'd be simple to put a label->version mapping in the hub, but I don't see why we'd use it.

python/extensions/pip.bzl
```diff
@@ -14,63 +14,140 @@
...
+
+    return whl_map
+
 def _pip_impl(module_ctx):
+    root_hubs = {}
+    submodules_hubs = {}
+    hub_whl_map = {}
+
+    # TODO we need to check for duplicate python versions in the same hub
+    # TODO we need to check for duplicate python hub names in submodules
```
rickeylev, line 91: To identify name conflicts, keep a mapping of global hub names to their defining module.

python/extensions/pip.bzl
```diff
@@ -14,63 +14,140 @@
...
+            continue
+        for pip in pip_attr:
+            hub_whl_map = _create_pip(module_ctx, pip, hub_name, hub_whl_map)
+
+    for hub_name, whl_map in hub_whl_map.items():
+        for whl_name, version_map in whl_map.items():
+            whl_library_alias(
+                name = hub_name + "_" + whl_name,
+                wheel_name = whl_name,
+                default_version = str(DEFAULT_PYTHON_VERSION),
```
rickeylev, line 124: nit: str() call isn't necessary; it's already a string.

rickeylev (changes requested):
Something I didn't see (it must be hidden in the pip repo rules somewhere?) is where the python-version-specific-ness of the `whl_library` repos are. How is that being done?

I was expecting to see something that generated a repo named e.g. "pip_tabulate_3_10", and it would have all the tabulate files in it for Python 3.10. And then "pip_tabuleate_3_9" for 3.9 etc.

But i don't see code that puts the version in any repo name. But the pip alias repo is using select() to dispatch to other repos (right?). Any idea how things are wired together in this regard?

rickeylev:
> [rickeylev] how is the python version-ness of whl_library working?

Ah I think I see it now: in extensions/pip.bzl, the version gets put into the `pip_name` variable.

python/extensions/pip.bzl
chrislovecnm: It is. I need to see if I can put a string in the hub, rather than an int

python/extensions/pip.bzl
chrislovecnm: I could use a walkthrough of how your refactored toolchains. The pattern you used is sweet.

we need to check for unique names and also need to check that we don’t have duplicate pip python versions in the same hub.

python/extensions/pip.bzl
```diff
@@ -14,63 +14,140 @@
...
+    for hub_name, pip_attr in root_hubs.items():
+        for pip in pip_attr:
+            # TODO we need to check to see if we have the same python version
+            # in a different pip call.
+            hub_whl_map = _create_pip(module_ctx, pip, hub_name, hub_whl_map)
+
+    for hub_name, pip_attr in submodules_hubs.items():
+        if hub_name in root_hubs:
+            # TODO clean up the warn statement like we are doing in toolchains
+            _print_warn("Not creating pip with the hub_name {}, same hub name found in the root module.".format(hub_name))
```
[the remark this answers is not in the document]
chrislovecnm: So then the user will have to modify sub modules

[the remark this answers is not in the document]
rickeylev: Unfortunately, yes, but there isn't much we can do about it. The only options are:
1. The user manually picks a globally unique hub name
2. We automatically put the module name in the hub name to force it to be globally unique, and users have the option of using repo aliases.
3. The user writes their own extension that does what pip.parse does (so that the repos end up in their module's namespace, not rules_python), probably based on helpers we provided (this could be as simple as a user extension just loading our extension and re-exporting the name, since the use_extension() args define identity).
4. We wait for the `is_isolated` feature that allows a particular extension usage to have its own namespace.

(1) is today. I'm mixed about (2) and (3). (4) is ideal, but doesn't exist yet.

python/pip.bzl
```diff
@@ -278,6 +278,9 @@ def _whl_library_render_alias_target(
         rules_python,
         version_map,
         wheel_name):
+    rules = rules_python
+    if str(Label("//:unused")).startswith("@@"):
```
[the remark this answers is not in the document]
rickeylev: Oh, is the value here e.g. `@rules_python~override`? That would be a canonical label already. Is the caller of this manually constraucting the "@rules_python" value, perhaps?

In anycase, this is pretty innocous, so just put a comment for now.

python/pip_install/multiple_pip_repository_requirements_bzlmod.bzl.tmpl
```diff
@@ -0,0 +1,29 @@
+"""Starlark representation of locked requirements.
```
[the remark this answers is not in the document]
chrislovecnm: I created an issue to track this: #1262

python/extensions/pip.bzl
```diff
@@ -14,63 +14,135 @@
...
+        )
+
+        if whl_name not in whl_map[hub_name]:
+            whl_map[hub_name][whl_name] = {}
+
+        whl_map[hub_name][whl_name][pip_attr.python_version] = pip_name + "_"
+
 def _pip_impl(module_ctx):
+    pip_hub_map = {}
+    hub_whl_map = {}
```
rickeylev, line 80:
```suggestion
    # Keeps track of all the hub's whl repos across the different versions.
    # dict[hub, dict[whl, dict[version, str pip]]]
    # Where hub, whl, and pip are the repo names
    hub_whl_map = {}
```

(tbh, this structure strikes me as a bit strange, but I don't fully understand the interaction between things here, so won't press the point).

python/extensions/pip.bzl
```diff
@@ -14,63 +14,135 @@
...
-        "name": attr.string(mandatory = True),
+        "hub_name": attr.string(mandatory = True),
+        "python_version": attr.string(mandatory = True),
     }, **pip_repository_attrs)
 
     # Like the pip_repository rule, we end up setting this manually so
     # don't allow users to override it.
     attrs.pop("repo_prefix")
 
+    # We do not use incompatible_generate_aliases in bzlmod.
```
rickeylev, line 143:
```suggestion
    # incompatible_generate_aliases is always True in bzlmod
```

python/pip.bzl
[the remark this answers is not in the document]
rickeylev: Ah, I see. Up in `_whl_library_alis_impl`, it uses the implicit label trick to get the canonical name, however, it then extract outs _just_ the name by accessing `.workspace_name`. Below, the template has "@" in it already, so we have to add the extra "@" for bzlmod, since the name is canonical. I've suggested an edit to this effect.

python/pip.bzl
```diff
@@ -278,6 +284,8 @@ def _whl_library_render_alias_target(
         rules_python,
         version_map,
         wheel_name):
+    if str(Label("//:unused")).startswith("@@"):
```
rickeylev, line 287:
```suggestion
    # The template below adds one @, but under bzlmod, the name
    # is canonical, so we have to add a second @.
    if str(Label("//:unused")).startswith("@@"):
```

examples/bzlmod/entry_point/BUILD.bazel
```diff
@@ -1,4 +1,4 @@
-load("@pip//:requirements.bzl", "entry_point")
+load("@pip_39//:requirements.bzl", "entry_point")
```
rickeylev, line 1: I don't think this is correct

This loads the version-specific macro, which means anything it resolves will resolve to a 3.9 dependency. In comparison, the rules will using the default Python, which may not be 3.9.

But, I'm guessing this was done because `entry_point` isn't implemented in the multi-pip variant of the rules?

rickeylev (changes requested):
Hm, I wonder if part of the solution to the multi-pip issue is to not load e.g. `@pip_39//:requirements.bzl`, but instead load `@pip//3.9:requirements.bzl`. Similar to what is done with `@python_aliases`.

examples/bzlmod/entry_point/BUILD.bazel
chrislovecnm: Correct this is because we don’t have the pip hub working yet. As well as you can still access the python specific version

rickeylev:
I was going to ask to remove the other bzlmod template because it's unused...but no, I don't think it actually is?

So, just to make sure I understand what's going on here. Please correct me where I'm wrong.

* `@pip` is the bzlmod pip hub. It has aliases to the whl alias hubs (`@pip_pylint`).
* `@pip_{38,39}` are the regular pip "aliases" hubs (i.e. the incompatible_generate_aliases=True thing). Each points to their particular `@pip_{ver}_pylint` repo. I think these are only of interest because they have a working entry point function? 
* `@pip_pylint` is the whl alias hub. It has select-based aliases pointing to the `@pip_{ver}_pylint` repos.
* `@pip_38_pylint` is a whl_library. Finally, something concrete.

Is the whl alias intermediary really necessary? I don't see why the bzlmod pip hub can't point directly to the underlying wheels. I'm fine with addressing this in a separate PR.

I think the regular pip hubs, e.g. `@pip_38`, are only around because they have a working entry point function, right? That's fine if so; just making sure I understand why they're here.

chrislovecnm:
> Something I didn't see (it must be hidden in the pip repo rules somewhere?) is where the python-version-specific-ness of the `whl_library` repos are. How is that being done?
> 
> I was expecting to see something that generated a repo named e.g. "pip_tabulate_3_10", and it would have all the tabulate files in it for Python 3.10. And then "pip_tabuleate_3_9" for 3.9 etc.
> 
> But i don't see code that puts the version in any repo name. But the pip alias repo is using select() to dispatch to other repos (right?). Any idea how things are wired together in this regard?

So because I am reusing how pip and multiple work, we are getting those repositories.  We have a pip-hub -> pip-whl-alias (which has the selects) -> pip versioned whl.

chrislovecnm:
> I was going to ask to remove the other bzlmod template because it's unused...but no, I don't think it actually is?
> 
> So, just to make sure I understand what's going on here. Please correct me where I'm wrong.
> 
> * `@pip` is the bzlmod pip hub. It has aliases to the whl alias hubs (`@pip_pylint`).
> * `@pip_{38,39}` are the regular pip "aliases" hubs (i.e. the incompatible_generate_aliases=True thing). Each points to their particular `@pip_{ver}_pylint` repo. I think these are only of interest because they have a working entry point function?
> * `@pip_pylint` is the whl alias hub. It has select-based aliases pointing to the `@pip_{ver}_pylint` repos.
> * `@pip_38_pylint` is a whl_library. Finally, something concrete.
> 
> Is the whl alias intermediary really necessary? I don't see why the bzlmod pip hub can't point directly to the underlying wheels. I'm fine with addressing this in a separate PR.
> 
> I think the regular pip hubs, e.g. `@pip_38`, are only around because they have a working entry point function, right? That's fine if so; just making sure I understand why they're here

We can probably improve the code more, but I would like to do that by iterating on the code.  This is how, currently, multiple Python versions work, with the addition of a hub.  So I wanted to keep the code as close as possible.

We need a hub -> alias (with the selects) -> Python versioned pip.  I'm guessing that we could combine these some to reduce the number of repos, but I wanted to get something out that I can guarantee works.  Just getting this working took a bunch of time.

python/extensions/pip.bzl
```diff
@@ -14,69 +14,301 @@
...
+            environment = pip_attr.environment,
+        )
+
+        if whl_name not in whl_map[hub_name]:
+            whl_map[hub_name][whl_name] = {}
+
+        whl_map[hub_name][whl_name][python_version] = pip_name + "_"
+
 def _pip_impl(module_ctx):
+    """Implmentation of a class tag that creates the pip hub(s) and corresponding pip spoke, alias and whl repositories.
```
chrislovecnm, line 89: @rickeylev cause this code is so complex, I wrote a small dissertation explaining how this code actually works.  Let me know what you think.

examples/bzlmod_build_file_generation/MODULE.bazel
```diff
@@ -90,6 +87,7 @@ pip.parse(
     # If this isn't specified, then you'll get whatever is locally installed
     # on your system.
     python_interpreter_target = "@{}//:python".format(INTERPRETER_NAME),
+    python_version = "3.9",
```
chrislovecnm, line 90: I technically do not need this here, but it follows a different code path so I will leave it here.

python/pip_install/multiple_pip_repository_requirements_bzlmod.bzl.tmpl
```diff
@@ -0,0 +1,29 @@
...
+def data_requirement(name):
+    return "%%MACRO_TMPL%%".format(_clean_name(name), "data")
+
+def dist_info_requirement(name):
+    return "%%MACRO_TMPL%%".format(_clean_name(name), "dist_info")
+
+def entry_point(pkg, script = None):
+    """entry_point returns the target of the canonical label of the package entrypoints.
+    """
+    print("not implemented")
```
chrislovecnm, line 29: TODO: add a comment why we have a different template and that entry_point is not yet implemented in both bzlmod and non-bzlmod.

chrislovecnm:
> Hm, I wonder if part of the solution to the multi-pip issue is to not load e.g. `@pip_39//:requirements.bzl`, but instead load `@pip//3.9:requirements.bzl`. Similar to what is done with `@python_aliases`.

For what @rickeylev? For entrypoint?

chrislovecnm:
@aignas do you mind taking a look?  This is another big PR and I would love another set of eyes.

python/pip_install/pip_hub_repository_requirements_bzlmod.bzl.tmpl
```diff
@@ -0,0 +1,32 @@
...
+def data_requirement(name):
+    return "%%MACRO_TMPL%%".format(_clean_name(name), "data")
+
+def dist_info_requirement(name):
+    return "%%MACRO_TMPL%%".format(_clean_name(name), "dist_info")
+
+def entry_point(pkg, script = None):
+    """entry_point returns the target of the canonical label of the package entrypoints.
+    """
+    print("not implemented")
```
aignas, line 32: Nit, this could be in theory easily supported if we attempt to return a select statement pointing to the spoke repos. Probably out of scope of this PR.

aignas (approved):
From my first pass, everything looks reasonable, will take a look later once more but feel free to merge before I do so.

python/pip_install/pip_hub_repository_requirements_bzlmod.bzl.tmpl
rickeylev: ref: https://github.com/bazelbuild/rules_python/issues/1262

python/extensions/pip.bzl
```diff
@@ -14,69 +14,301 @@
...
+    Both of these pip spokes contain requirements files that includes websocket
+    and its dependencies.
+
+    Two different repositories are created for the two spokes:
+
+    - @@rules_python~override~pip~pip_39
+    - @@rules_python~override~pip~pip_310
+
+    The different spoke names are a combination of the hub_name and the Python version.
+    In the future we may remove this repository, but we do not support endpoints
```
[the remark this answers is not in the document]
chrislovecnm: https://github.com/bazelbuild/rules_python/blob/main/examples/bzlmod/entry_point/test_entry_point.py See that.  I am not real familiar with them.

python/extensions/pip.bzl
```diff
@@ -14,69 +14,301 @@
...
-            # Create the repository where users load the `requirement` macro. Under bzlmod
-            # this does not create the install_deps() macro.
-            pip_repository_bzlmod(
-                name = attr.name,
-                repo_name = attr.name,
-                requirements_lock = attr.requirements_lock,
-                incompatible_generate_aliases = attr.incompatible_generate_aliases,
-            )
+        for pip_attr in mod.tags.parse:
+            # We can automatically dertermine the Python version of an
```
rickeylev, line 181: I'm not against inferring the version from the name per se (we're matching our internal names, so its fine), and I'm OK with this right now.

That said, I'm leaning more and more towards the idea of using `python_version` to act as the "driver" of things.
  * (Ignas's idea) We can easily map the version to the backing repository and interpreter. Now users don't need to specify the interpreter target or use the interpreter extension.
  * With Phil's work to uncouple the pip dependency processing from the interpreter used, we would no longer need the interpreter, but would still need the python version.
  * Allowing multi-version requirements files using env markers requires us to know the intended python version; knowing the interpreter doesn't really help.

Anyways, fine for now. I filed https://github.com/bazelbuild/rules_python/issues/1268 for more thoughts here.

python/extensions/pip.bzl
```diff
@@ -14,69 +14,301 @@
...
+                # We get the full name of the label, and this resolves symlinks
+                python_binary_path = module_ctx.path(pip_attr.python_interpreter_target).realpath
+                python_binary_path = str(python_binary_path)
+
+                # Iterate through the different Python versions supported and
+                # match on the correct directory structure.
+                for version in MINOR_MAPPING.keys():
+                    # This will create a value like
+                    # _3_10_x86_64-unknown-linux-gnu
+                    python_platform_version = "_{version}_{platform}".format(
```
rickeylev, line 200: We control the repo name, so lets just use its full apparent name when matching.

```suggestion
                    python_platform_version = "python_{version}_{platform}".format(
```

rickeylev (changes requested):
impl mostly looks good, but a couple questions still

python/extensions/pip.bzl
chrislovecnm: > I'm not against inferring the version from the name per se (we're matching our internal names, so its fine), and I'm OK with this right now.
> 
> That said, I'm leaning more and more towards the idea of using `python_version` to act as the "driver" of things.
> 
> * (Ignas's idea) We can easily map the version to the backing repository and interpreter. Now users don't need to specify the interpreter target or use the interpreter extension.
> * With Phil's work to uncouple the pip dependency processing from the interpreter used, we would no longer need the interpreter, but would still need the python version.
> * Allowing multi-version requirements files using env markers requires us to know the intended python version; knowing the interpreter doesn't really help.
> 
> Anyways, fine for now. I filed #1268 for more thoughts here.

Not having the interpreter repo may cause some challenges with the dependencies. We would need to test it. We could also write out the sym link in the python_aliases repo. 

There are use cases where the user won’t want an interpreter, so I am wondering how we do that.

python/pip.bzl
```diff
@@ -259,6 +259,12 @@ _multi_pip_parse = repository_rule(
 
 def _whl_library_alias_impl(rctx):
     rules_python = rctx.attr._rules_python_workspace.workspace_name
+    if rctx.attr.default_version not in rctx.attr.version_map:
+        fail(
+            """
+Unable to find '{}' in your version map, you may need to update your requirement files.
+        """.format(rctx.attr.version_map),
+        )
```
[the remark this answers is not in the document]
chrislovecnm: I can remove

python/extensions/pip.bzl
rickeylev: > Not having the interpreter repo may cause some challenges with the dependencies.

How so?

> There are use cases where the user won’t want an interpreter, so I am wondering how we do that.

I'm guessing you mean a user wanting to use a local system python? The solution for that is for the user to specify the interpreter using the pip.parse `python_interpreter` arg (this accepts a path to an arbitrary interpreter).

If you mean something else, I'm not sure what you mean -- an interpreter is required in some fashion because we have to eventually run various commands using it.

rickeylev: I don't follow, but lets continue discussion on #1268

python/pip.bzl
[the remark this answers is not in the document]
rickeylev: Oh silly me -- the line below does a dict lookup and would fail if this wasn't true. Sorry, didn't look beyond the diff. It's fine to keep.

python/extensions/pip.bzl
```diff
@@ -14,69 +14,301 @@
...
-
-            # Create the repository where users load the `requirement` macro. Under bzlmod
-            # this does not create the install_deps() macro.
-            pip_repository_bzlmod(
-                name = attr.name,
-                repo_name = attr.name,
-                requirements_lock = attr.requirements_lock,
-                incompatible_generate_aliases = attr.incompatible_generate_aliases,
-            )
+        for pip_attr in mod.tags.parse:
```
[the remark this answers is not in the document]
aignas: OK, another reason for splitting could be the indentation level of the code below, as it is getting a quite left-padded, but it's fine to merge as is.

python/pip_install/pip_hub_repository_requirements_bzlmod.bzl.tmpl
aignas: Yup, I found that after commenting here, LGTM.

rickeylev (approved):

Issue #1232, opened 2023-05-17
transitive bzlmod toolchains can cause name collisions


rickeylev:
# 🐞 bug report

### Affected Rule

The `python.toolchain()` extension

### Is this a regression?

Yes and no. Bzlmod support is beta and it's been present from the start. This was clearly possible with WORKSPACE, but better known as a potential issue there. Bzlmod just makes it easier and less obvious.

### Description

The repos created by `python.toolchain()` are all created in the `rules_python~python` namespace (module="rules_python", extension="python"). Because the user is able to _directly_ specify the name of the created repo, if two modules both specify the same name, a conflict occurs.

This is partially because in the python.toolchain implementation, we walk the module graph for all the toolchain calls and create any requested toolchains. This is WAI, as modules may need different Python versions, so must be able to request their particular Python version be made available.

A partial workaround is to use a unique name, e.g include the module name in the toolchain call: `python.toolchain(name="my_module_python3_10")`. I say partial because, while this makes the toolchain available, another module can register a different python version, and then the usual `register_toolchains()` registration ordering applies. i.e. a sub-module doesn't have a guarantee the toolchain they asked for will be used (this will be fixed as part of making the multi-version toolchains work with bzlmod).

## 🔬 Minimal Reproduction

A simple reproduction is to have a sub-module call `python.toolchain()` with the same name as the root module.

rickeylev:
cc @chrislovecnm

chrislovecnm:
@rickeylev I think we can close this now?

rickeylev:
Yeah, between 
 * https://github.com/bazelbuild/rules_python/commit/148622aa92fdd5afcaf9f153d7bb6afce713e553
 * https://github.com/bazelbuild/rules_python/commit/afdbedd3a58cb5d65e659e1dcc77108cfd22715e

Unique names are enforced, so collisions aren't possible. I also have an wip change that will remove the name arg entirely.

PR #1257, opened 2023-06-07
fix(bzlmod)!: Remove ability to specify toolchain repo name.

rickeylev:
The main reasons this is removed is because if modules choose different names for the same toolchain, only one of the two toolchains (which are, hopefully, identical) will be used. Which toolchain is used depends on the module graph dependency ordering.

Furthermore, as of #1238, only one repo per version is created; others are ignored. This means a module doesn't know if the name it chooses will result in a repo being created with that name.

Instead, the toolchain repos are named by rules_python: `python_{major}_{minor}`.

BREAKING CHANGES

This removes the `name` arg from `python.toolchain()`. Users will need to remove such usages from their `MODULE.bazel` and update their `use_repo()` statements. If keeping the custom repo name is necessary, then repo mappings can be used. See #1232 for additional migration steps, commands, and information.

### DIFF
--- a/.bazelignore
+++ b/.bazelignore
@@ -6,3 +6,5 @@ bazel-rules_python
 bazel-bin
 bazel-out
 bazel-testlogs
+examples/bzlmod/bazel-bzlmod
+examples/bzlmod_build_file_generation/bazel-bzlmod_build_file_generation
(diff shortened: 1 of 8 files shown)


rickeylev:
Something is wrong with this PR; it's not getting updates that i'm pushing. I'm going to close it and reopen another.

PR #1258, opened 2023-06-07
fix(bzlmod)!: Remove ability to specify toolchain repo name.

rickeylev:
The main reasons this is removed is because if modules choose different names for the same toolchain, only one of the two toolchains (which are, hopefully, identical) will be used. Which toolchain is used depends on the module graph dependency ordering.

Furthermore, as of #1238, only one repo per version is created; others are ignored. This means a module doesn't know if the name it chooses will result in a repo being created with that name.

Instead, the toolchain repos are named by rules_python: `python_{major}_{minor}`. These repo names are currently part of the public API, since they end up referenced in MODULE config (to wire the toolchain interpreter to pip).

BREAKING CHANGES

This removes the `name` arg from `python.toolchain()`. Users will need to remove such usages from their `MODULE.bazel` and update their `use_repo()` statements. If keeping the custom repo name is necessary, then repo mappings can be used. See #1232 for additional migration steps, commands, and information.


.bazelignore
```diff
@@ -6,3 +6,5 @@ bazel-rules_python
 bazel-bin
 bazel-out
 bazel-testlogs
+examples/bzlmod/bazel-bzlmod
```
chrislovecnm, line 9: @rickeylev any reason these are here?

examples/py_proto_library/MODULE.bazel
```diff
@@ -14,11 +14,10 @@ local_path_override(
 
 python = use_extension("@rules_python//python/extensions:python.bzl", "python")
 python.toolchain(
-    name = "python3_9",
     configure_coverage_tool = True,
     python_version = "3.9",
 )
-use_repo(python, "python3_9")
+use_repo(python, "python_3_9")
```
chrislovecnm, line 20: So if we register 3.9.11, what is the name? I assume it is 3_9_11.  We need to document this pattern somewhere.   Update the documentation in the tag class and also update the documentation in the MODULE.bazel files.  How the names generated gave me heartburn when I had less bazel experience.

chrislovecnm (changes requested):
Nit picks, but we need more documentation please :)

.bazelignore
rickeylev: Yes, they make `bazel test ...` (from the rules_python workspace) ignore the generated symlinks in the example sub-directory workspaces. Otherwise it'll recurse into them and find the various BUILD files and try to interpret them.

examples/py_proto_library/MODULE.bazel
rickeylev: We don't support registering patch versions, just major.minor.

I added doc to the tag class.

Which module file are you referring to? I'm guessing the bzlmod example? I've updated the comment there about the name.

chrislovecnm:
CI go boom 😂

rickeylev:
CI fixed

chrislovecnm (approved):

Issue #1233, opened 2023-05-17
rules_python bzlmod should register a default toolchain


rickeylev:
# 🚀 feature request

Well, sort of a feature-parity / bug report, too; see below.

### Relevant Rules

bzlmod module

### Description

Today, users must manually specify a toolchain and manually register it. This adds friction and causes problems for a few reasons:

1. User configs are more verbose. They end up having to specify a version and toolchain to use.
2. What they specify above is somewhat misleading and can lead to surprising behavior because another module can specify a different toolchain and version, and it might be used instead (it depends on the order of register_toolchain calls)
3. It's just a general barrier to entry; configuring the toolchain is extra steps that get in the way to just getting started. I, myself, have put off trying out various ideas and bug fixes because setting up a MODULE with all the boiler plate is a chore.

It's also worth noting that WORKSPACE builds already register a default toolchain -- whatever is found on the local system (`@bazel_tools//tools/python:autodetecting_toolchain`). This toolchain is problematic for various reasons, but that is the existing WORKSPACE-based behavior.

### Describe the solution you'd like

Instead, rules_python should register a toolchain by default. This avoids the three issues above.

I propose that it register the most recent version of Python available. We should be clear that this is for convenience -- the default provided shouldn't be relied upon.

If a module needs a specific version, then it should use the multi-version rules.

chrislovecnm:
@rickeylev and I have discussed this a bunch, we are uncertain if we are going to do this. We need to have the rules register a toolchain, and usually not the user. But autoregistering may cause issues.

Clement-Jean:
This could also avoid problems like the following: [[Bug]: rules_python@0.21.0 'python/extensions' is not a package](https://github.com/bazelbuild/bazel-central-registry/issues/659#issue-1722994278). Because the rules_python repo changed faster than the bzlmod registry it seems we can't use the last bzlmod dep. Less config in the MODULE.bazel, the less problems we will have for updating rules_python.

rickeylev:
I came back to this issues today when thinking about what happens if the root module doesn't use Python, but a submodule does. For example, if I'm a JavaScript project and don't use Python myself, I shouldn't have to setup Python because a module I depend on happens to use Python.

My basic conclusion is rules_python should register a default toolchain. The root module can, of course, still override it.

[...]

PR #1259, opened 2023-06-07
feat(bzlmod): Register a default toolchain

rickeylev:
This makes rules_python always provide a default toolchain when using
bzlmod.

Note that, unlike workspace builds, the default is not the local
system Python (`@bazel_tools//tools/python:autodetecting_toolchain`).

Instead, the default is a hermetic runtime, but no guarantees are made
about the particular version used. In practice, it will be the latest
available Python version.

Work towards #1233


rickeylev:
Note this is stacked atop #1258, so if you look at the diff, select the commits after the `feat(bzlmod): Remove ability to specify toolchain repo name` commit. It'll give a cleaner diff. Or just wait until 1258 is merged and that diff base goes away

chrislovecnm:
LGTM, but waiting on #1233.

chrislovecnm:
@rickeylev I put the PR into draft so it does not get accidentally merged.

chrislovecnm (changes requested):
We need the README and BZLMOD files updated please.

rickeylev:
Oh, I see the section in the readme. I'll tweak that text a bit

rickeylev (commented):
PTAL

BZLMOD_SUPPORT.md
```diff
@@ -35,3 +35,22 @@ This rule set does not have full feature partity with the older `WORKSPACE` type
...
+
+### Default toolchain is not the local system Python
+
+Under bzlmod, the default toolchain is no longer based on the locally installed
+system toolchain. Instead, a recent Python version using the pre-built,
+standalone runtimes are used.
+
+If you need to use the local system python, then it's suggested that you
+setup and configure your own toolchain and register it. Note that using the
+local system's Python is not advised because will vary between users and
```
chrislovecnm, line 49: So it sounds like we don’t need a local python installed, but we still do. Please add a note that we need Python installed locally to shim the use of the hermetic toolchain.

chrislovecnm (commented):
Another nit pick in the docs.

BZLMOD_SUPPORT.md
rickeylev: Well, OK, I added a note about that. However, this section is just about the toolchain, which is separate from the bootstrapping process. Regardless of what toolchain you use or how (local, hermetic, your own, etc), the bootstrap is the same: a local python is necessary to do it.

I'd like to reference the issue about this, but I can't find it. If you know it, LMK.

rickeylev:
PTA

rickeylev:
PTAL, added the reference to the issue about the bootstrap needing a local python

README.md
```diff
@@ -45,31 +45,52 @@ the older way of configuring bazel with a `WORKSPACE` file.
...
+python.toolchain(
+    python_version = "3.9",
+)
+```
+
+### Using pip with bzlmod
 
+NOTE: bzlmod support is still experimental; APIs subject to change.
+
+To use dependencies from PyPi, the `pip.parse()` extension is used to
```
aignas, line 81: nit: it's Python Package Index and PyPI is the usual spelling.

aignas (approved):
LGTM, thanks for the really extensive docs!

README.md
rickeylev: Fixed

chrislovecnm (approved):
LGTM - I want to improve our docs.

PR #1264, opened 2023-06-13
fix(bzlmod+gazelle): update BCR release presubmit to use correct example

rickeylev:
The bzlmod-compatible build_file_generation example was moved to the bzlmod_build_file_generation example.

This should fix the automatic build/release of the gazelle BCR module.


chrislovecnm (approved):
LGTM

PR #1265, opened 2023-06-13
fix(bzlmod): Fixing Windows Python Interpreter symlink issues

chrislovecnm:
When using Windows you cannot run the symlink directly.

Because of how symlinks work in Windows we need to path realpath resolve the link.
Unlike Linux and OSX we cannot execute the Python symlink directly. Windows throws an error "-1073741515", when we execute the symlink directory.  This error means that the Python binary cannot find dlls. This is because the symlink that we create is not in the same folder as the dlls.

### DIFF
--- a/python/pip_install/pip_repository.bzl
+++ b/python/pip_install/pip_repository.bzl
@@ -80,8 +80,8 @@ def _resolve_python_interpreter(rctx):
         if str(Label("//:unused")).startswith("@@"):
             (os, _) = get_host_os_arch(rctx)
 
-            # If we have Windows the symlink will not work directly and we need
-            # to resolve the realpath.
+            # On Windows, the symlink doesn't work because Windows attempts to find
+            # Python DLLs where the symlink is, not where the symlink points.
             if os == WINDOWS_NAME:
                 python_interpreter = python_interpreter.realpath
     elif "/" not in python_interpreter:
--- a/python/pip_install/pip_repository.bzl
+++ b/python/pip_install/pip_repository.bzl
@@ -15,9 +15,11 @@
 ""
 
 load("//python:repositories.bzl", "get_interpreter_dirname", "is_standalone_interpreter")
+load("//python:versions.bzl", "WINDOWS_NAME")
 load("//python/pip_install:repositories.bzl", "all_requirements")
 load("//python/pip_install:requirements_parser.bzl", parse_requirements = "parse")
 load("//python/pip_install/private:srcs.bzl", "PIP_INSTALL_PY_SRCS")
+load("//python/private:toolchains_repo.bzl", "get_host_os_arch")
 
 CPPFLAGS = "CPPFLAGS"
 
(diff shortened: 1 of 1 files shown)


python/pip_install/pip_repository.bzl
```diff
@@ -72,8 +74,38 @@ def _resolve_python_interpreter(rctx):
...
-        python_interpreter = rctx.path(target)
+        # If we have @@ we have bzlmod so we need to hand Windows differently.
+        if str(Label("//:unused")).startswith("@@"):
+            (os, _) = get_host_os_arch(rctx)
+            # If we have Windows the symlink will not work directly and we need
+            # to run a powershell script.
+            # The powershell script will resolve the Python target and
+            # return that string.
+            if os == WINDOWS_NAME:
+                path = rctx.path(rctx.attr.python_interpreter_target).dirname
```
rickeylev, line 85: Did you try the `.realpath` I mentioned in chat yet?

`path = rctx.path(rctx.attr.python_interpreter_target).realpath`

This should resolve any symlinks without having to call powershell.

python/pip_install/pip_repository.bzl
```diff
@@ -72,8 +74,14 @@ def _resolve_python_interpreter(rctx):
...
 
     if rctx.attr.python_interpreter_target != None:
-        target = rctx.attr.python_interpreter_target
-        python_interpreter = rctx.path(target)
+        python_interpreter = rctx.path(rctx.attr.python_interpreter_target)
+        # If we have @@ we have bzlmod so we need to hand Windows differently.
+        if str(Label("//:unused")).startswith("@@"):
+            (os, _) = get_host_os_arch(rctx)
+            # If we have Windows the symlink will not work directly and we need
+            # to resolve the realpath.
```
rickeylev, line 84:
```suggestion
            # On Windows, the symlink doesn't work because Windows attempts to find
            # Python DLLs where the symlink is, not where the symlink points.
```

python/pip_install/pip_repository.bzl
```diff
@@ -699,7 +708,7 @@ def _whl_library_impl(rctx):
     )
 
     if result.return_code:
-        fail("whl_library %s failed: %s (%s)" % (rctx.attr.name, result.stdout, result.stderr))
+        fail("whl_library %s failed: %s (%s) error code: '%s'" % (rctx.attr.name, result.stdout, result.stderr, result.return_code))
```
rickeylev, line 711: Since we're here...

```suggestion
        fail(("whl_library '{name}' failed: wheel_installer failed\n" +
              "command: {args}\n" +
              "error code: {error_code}\n" +
              "===== stdout ====={stdout}\n" +
              "===== stderr ====={stderr}\n=====").format(
              name = rctx.attr.name,
              command = args,
              code=result.return_code,
              stdout=result.stdout,
              stderr=result.stderr
            ))
```

rickeylev (approved):
Just doc and ux improvements, otherwise lgtm

python/pip_install/pip_repository.bzl
chrislovecnm: Did you test this code?

rickeylev: no. i'm fine if you want to skip it for now.

PR #1272, opened 2023-06-15
feat!(docs): Updating bzlmod docs

chrislovecnm:
Fixing the bzlmod docs to match latest pip changes.  Also, adding better docs for the release.

### DIFF
--- a/.github/workflows/create_archive_and_notes.sh
+++ b/.github/workflows/create_archive_and_notes.sh
@@ -31,14 +31,20 @@ Add to your \`MODULE.bazel\` file:
 
 \`\`\`starlark
 bazel_dep(name = "rules_python", version = "${TAG}")
+python = use_extension("@rules_python//python:extensions.bzl", "python")
+
+python.toolchain(
+    python_version = "3.10",
+)
+use_repo(python, "python_3_10", "python_aliases")
 
 pip = use_extension("@rules_python//python:extensions.bzl", "pip")
-
 pip.parse(
-    name = "pip",
-    requirements_lock = "//:requirements_lock.txt",
+    hub_name = "pip",
+    python_version = "3.10",
+    requirements_lock = "//:requirements_lock_3_10.txt",
+    requirements_windows = "//:requirements_windows_3_10.txt",
 )
-
 use_repo(pip, "pip")
 \`\`\`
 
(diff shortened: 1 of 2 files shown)


.github/workflows/create_archive_and_notes.sh
```diff
@@ -31,14 +31,20 @@ Add to your \`MODULE.bazel\` file:
 
 \`\`\`starlark
 bazel_dep(name = "rules_python", version = "${TAG}")
+python = use_extension("@rules_python//python:extensions.bzl", "python")
 
-pip = use_extension("@rules_python//python:extensions.bzl", "pip")
+python.toolchain(
+    python_version = "3.10",
+)
+use_repo(python, "python_3_10", "python_aliases")
```
rickeylev, line 39: I'd rather omit the toolchain part and python_aliases part and put those into a dedicated doc about using bzlmod. The reason is if and how you should do it depends on whether you're the root module or a submodule (which is further complicated by every submodule is the root module for itself during development).

.github/workflows/create_archive_and_notes.sh
```diff
@@ -31,14 +31,20 @@ Add to your \`MODULE.bazel\` file:
...
 
+pip = use_extension("@rules_python//python:extensions.bzl", "pip")
 pip.parse(
-    name = "pip",
-    requirements_lock = "//:requirements_lock.txt",
+    hub_name = "pip",
```
rickeylev, line 43:
```suggestion
    hub_name = "pip",  # This name must be unique across MODULEs
```

rickeylev (changes requested):

.github/workflows/create_archive_and_notes.sh
rickeylev: I'd be fine with parring this down to just the `bazel_dep()` call. That's about the minimum you need, and it contains the release-specific information (the version).

chrislovecnm: I am on the fence.  Do many people use the bare minimum?

rickeylev:
Is this PR defunct? We've update the readme quite a bit now

chrislovecnm:
We have already done this work. Closing

PR #1275, opened 2023-06-15
cleanup: Typos, doc, and formatting cleanup in pip extension et al

rickeylev:
* Corrects some typos in docs
* Expands the user-facing documentation
* Fixes errors having newlines in the middle of them
* Renames some internal functions to be more self-describing.


python/extensions/pip.bzl
```diff
@@ -209,17 +223,19 @@ def _pip_impl(module_ctx):
...
 
-            _create_pip(module_ctx, pip_attr, hub_whl_map)
+            _create_versioned_pip_and_whl_repos(module_ctx, pip_attr, hub_whl_map)
 
     for hub_name, whl_map in hub_whl_map.items():
         for whl_name, version_map in whl_map.items():
             if DEFAULT_PYTHON_VERSION not in version_map:
-                fail(
-                    """
-Unable to find the default python version in the version map, please update your requirements files
```
chrislovecnm, line 219: > version map

Can we create a warning message that tells the user how to fix it?  "update" your requirements files?  How?

python/extensions/pip.bzl
```diff
@@ -209,17 +223,19 @@ def _pip_impl(module_ctx):
...
-to include Python '{}'.
-""".format(DEFAULT_PYTHON_VERSION),
-                )
+                fail((
+                    "Default python version '{version}' missing in pip " +
+                    "hub '{hub}': update your pip.parse() call to include " +
```
chrislovecnm, line 233: How about something like:

Update one of your pip.parse() calls to include default_version = True

python/extensions/pip.bzl
```diff
@@ -247,14 +263,27 @@ def _pip_parse_ext_attrs():
         "hub_name": attr.string(
             mandatory = True,
             doc = """
-The unique hub name.  Mulitple pip.parse calls that contain the same hub name, 
-create spokes for specific Python versions.                                
+The repo name pip dependencies will be made available within.
```
chrislovecnm, line 266: "within" what?

python/extensions/pip.bzl
```diff
@@ -247,14 +263,27 @@ def _pip_parse_ext_attrs():
...
+
+This name must be unique between modules, but should be the same within
```
chrislovecnm, line 268: You can have more than one hub name in a Module.  I could see it used in larger projects that have different requirements files.  So "should be the same", can we do something like "typically the same"?

python/extensions/pip.bzl
```diff
@@ -247,14 +263,27 @@ def _pip_parse_ext_attrs():
...
+a module. Unless your module is guaranteed to always be the root module, it's
+highly recommended to include your module name in the hub name. Repo mapping,
+`use_repo(..., pip="my_modules_pip_deps")`, can be used for shorter local names
+within your module.
+
+Multiple `pip.parse()` calls are grouped under this name to provide support
```
chrislovecnm, line 274:
```suggestion
Multiple `pip.parse()` calls are grouped under the same `hub_name` to provide support
```

chrislovecnm (changes requested):
More word smithing requested.

python/extensions/pip.bzl
rickeylev: Rephrased a bit.

python/extensions/pip.bzl
rickeylev: Done; there's no default_version arg. I think you mean python_version = <default version>? Updated the text to say as much.

python/extensions/pip.bzl
rickeylev: Ah yes, good catch. Updated the text to clarify that better.

python/extensions/pip.bzl
rickeylev: Rewrote this section; n/a

rickeylev (commented):
ptal

chrislovecnm (approved):
LGTM

PR #1276, opened 2023-06-17
feat(py_wheel): Add support for specifying Project-URL in METADATA

sfc-gh-wzhao:
`Project-URL` is a field available in core metadata since version 1.2, which allows specifying additional URLs and display as Project Links in PyPI package web page.

https://packaging.python.org/en/latest/specifications/core-metadata/#project-url-multiple-use

This change adds the support to specify that.


sfc-gh-wzhao:
@sfc-gh-sdas FYI.

python/private/py_wheel.bzl
```diff
@@ -301,6 +305,9 @@ def _py_wheel_impl(ctx):
     if ctx.attr.summary:
         metadata_contents.append("Summary: %s" % ctx.attr.summary)
 
+    for label, url in sorted(ctx.attr.project_urls.items()):
```
chrislovecnm, line 308: Please check for the max length of the label. The label is limited to 32 characters.

python/private/py_wheel.bzl
```diff
@@ -176,6 +176,10 @@ _other_attrs = {
         doc = "A string specifying the license of the package.",
         default = "",
     ),
+    "project_urls": attr.string_dict(
+        doc = ("A string dict specifying additional browsable URLs for the project and corresponding labels. " +
```
chrislovecnm, line 180: Nit pick. Mention the key first and the label second. You have them reversed.

chrislovecnm (commented):
Thanks for the PR! See comments.

python/private/py_wheel.bzl
sfc-gh-wzhao: Thank you for pointing out. A check has been added, and will fail out if the length is over limit. Corresponding test is also added.

python/private/py_wheel.bzl
sfc-gh-wzhao: Updated with better explanation about the way to specify.

chrislovecnm (approved):
LGTM!

PR #1280, opened 2023-06-20
fix: plugin_output in py_proto_library rule

comius:
plugin_output was wrong in case multiple repositories are involved and/or _virtual_imports. 

The code is taken from `cc_proto_library` and has been verified in practice.

### DIFF
--- a/python/private/proto/py_proto_library.bzl
+++ b/python/private/proto/py_proto_library.bzl
@@ -75,6 +75,7 @@ def _py_proto_aspect_impl(target, ctx):
             name_mapper = lambda name: name.replace("-", "_").replace(".", "/"),
         )
 
+        # Handles multiple repository and virtual import cases
         proto_root = proto_info.proto_source_root
         if proto_root.startswith(ctx.bin_dir.path):
             plugin_output = proto_root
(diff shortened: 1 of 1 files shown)


python/private/proto/py_proto_library.bzl
```diff
@@ -75,12 +75,21 @@ def _py_proto_aspect_impl(target, ctx):
             name_mapper = lambda name: name.replace("-", "_").replace(".", "/"),
         )
 
+        proto_root = proto_info.proto_source_root
```
rickeylev, line 78: nit: add a comment that explains this is to handle when multiple repositories are involved and virtual imports. It's easy to forget when different repos are involved and sort of a pain to test it, so lets at least comment it.

rickeylev (approved):
Is there an issue somewhere describing the bug in more detail? Or did you just come across this accidentally?

Otherwise lgtm

python/private/proto/py_proto_library.bzl
comius: Done.

comius:
> Is there an issue somewhere describing the bug in more detail? Or did you just come across this accidentally?
> 
> Otherwise lgtm

Accidentally. We had the same problem internally, reported on b/277335175.

And I spent a month fixing _virtual_imports and otherwise generally cleaning up/optimising proto_library implementation.

PR #1284, opened 2023-06-21
feat!: using Gazelle's lifecycle manager to manage external processes

linzhp:
Gazelle v0.30.0 introduced a lifecycle manager. We can use that to start and shutdown parser and stdmodule processes. So we don't need to use `init` function or creating `context.WithTimeout`.

BREAKING CHANGES:
This requires the users of this Gazelle extension to upgrade to Gazelle v0.30.0 or above.


gazelle/python/std_modules.go
```diff
@@ -25,19 +25,18 @@ import (
...
 	"strings"
 	"sync"
-	"time"
 
 	"github.com/bazelbuild/rules_go/go/tools/bazel"
 )
 
 var (
-	stdModulesStdin  io.Writer
+	stdModulesStdin  io.WriteCloser
```
f0rmiga, line 33: Why this? You should not close stdio.

linzhp: According to the official doc of [StdinPipe](https://pkg.go.dev/os/exec#Cmd.StdinPipe):

> A caller need only call Close to force the pipe to close sooner. For example, if the command being run will not exit until standard input is closed, the caller must close the pipe.

Both parser.py and std_modules.py loop through their stdin until EOF. So the proper way to shutdown those two processes is to send them EOF by closing the pipe:

https://github.com/bazelbuild/rules_python/blob/0d59fcf561f6d2c4705924bc17c151fb4b998841/gazelle/python/parse.py#L84

https://github.com/bazelbuild/rules_python/blob/0d59fcf561f6d2c4705924bc17c151fb4b998841/gazelle/python/std_modules.py#L39

f0rmiga: I see, thanks.

f0rmiga (approved):

PR #1285, opened 2023-06-26
fix: add `format()` calls to `glob_exclude` templates

lpulley:
Currently, Python toolchain `:files` and `:py3_runtime` include some unnecessary files. This is because these globs result in literally `lib/libpython{python_version}.so` etc., which do not match anything. The formatting needs to be applied here since it will not be applied later.

I believe this was introduced by a47c6cd681b34b1ad990ed40dcc01ab5f024406a.

### DIFF
--- a/python/repositories.bzl
+++ b/python/repositories.bzl
@@ -204,12 +204,12 @@ def _python_repository_impl(rctx):
         "**/* *",  # Bazel does not support spaces in file names.
         # Unused shared libraries. `python` executable and the `:libpython` target
         # depend on `libpython{python_version}.so.1.0`.
-        "lib/libpython{python_version}.so",
+        "lib/libpython{python_version}.so".format(python_version = python_short_version),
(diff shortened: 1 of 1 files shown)


lpulley:
Build fails without 61ec702f, though it's unrelated. Should that be separated?

daveboutcher (approved):

rickeylev (approved):

lpulley:
@rickeylev do you know why the build is failing on Windows?

lpulley:
@rickeylev I think this is ready to go now. I rebased onto main and that seems to have helped.

chrislovecnm (approved):
LGTM

chrislovecnm:
@lpulley do you know if we are pulling in too many files with windows? I’m guessing we need to add a pattern for windows as well. If you don’t work with windows can you please file an issue so we investigate it?

PR #1286, opened 2023-06-27
fix(bzlmod)!: Bump compatibility level

jsharpe:
https://github.com/bazelbuild/rules_python/commit/1c581242c25925a1bd6764608877ec170724f11d introduces a change to the extension API provided by rules_python that are not backwards compatible so bump compatibility_level before the next release.

BREAKING CHANGES
The extension API has changed in this release so the compatibility level has been bumped.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -1,7 +1,7 @@
 module(
     name = "rules_python",
     version = "0.0.0",
-    compatibility_level = 1,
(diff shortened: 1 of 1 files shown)


jsharpe:
Note this may not be the correct solution. See discussion on slack: https://bazelbuild.slack.com/archives/C014RARENH0/p1687895823784159

PR #1289, opened 2023-06-29
Adding kwargs to gazelle_python_manifest

linzhp:
Adding kwargs to gazelle_python_manifest, so we can set the tags and size to the test. This is similar to:

https://github.com/bazelbuild/rules_python/blob/5b8fa22a2f22501b18b4aea97c5dbfe3a6913a0c/python/pip_install/requirements.bzl#L20-L62

Alternatively, we can add individual `go_test` attributes as needed. Please advice which way is preferred.


chrislovecnm (changes requested):
Thanks for the PR!  Where can we show users how to use this functionality?  Can we add it to one of the examples?  I have two goals:

1. Get it tested.
2. Show a user how to use this parameter.

linzhp:
@chrislovecnm I updated the doc and added an example to use the new feature. Can you take another look?

chrislovecnm (approved):
Lgtm

Issue #894, opened 2022-11-22
compile_pip_requirements: a way to disable hashes (for e.g. constraints files)


vonschultz:
# 🚀 feature request

### Relevant Rules

`compile_pip_requirements`

### Description

As it is, `compile_pip_requirements`, through `python/pip_install/pip_compile.py`, always adds the `--generate-hashes` flag.
```
    sys.argv.append("--generate-hashes")
```
I would like an option to disable that.

Specifically, I'm using `compile_pip_requirements` both to generate `requirement.txt` files, where I do want the hashes, and also to generate `constraints.txt` files, where the hashes don't really make sense. It's the `requirements.txt` files that control what gets installed, so the hashes make a lot of sense there, but in the constraints file (used with `-c constraints.txt` in a `requirements.in` file) hashes make less sense. I'm not sure they are even obeyed by `pip` (see https://github.com/pypa/pip/issues/8792), so in the constraints file it's confusing noise at best.

### Describe the solution you'd like

A parameter to `compile_pip_requirements` that says if we want hashes in the output file.

Topher-the-Geek:
+1

As we migrate from our old build system to Bazel, we are straddling the two. This flag would really help&mdash;in the meantime, we are going to patch `rules_python` in `http_archive`.

PR #1290, opened 2023-07-02
feat: Add setting generate_hashes for requirements

vonschultz:
Add the new parameter `generate_hashes` (default True) to `compile_pip_requirements()`, letting the user control whether to put `--hash` entries in the requirements lock file generated. In particular if the generated file is supposed to be used as a constraints file the hashes don't make much sense.

Fixes bazelbuild/rules_python#894.


aignas:
+1 to this change as right now I have a `genrule` that is stripping the hashes away in one of our projects.

aignas (approved):

chrislovecnm (approved):
LGTM

PR #1301, opened 2023-07-10
feat(gazelle): support multiple requirements files in manifest generation

aignas:
For certain workflows it is useful to calculate the integrity hash of
the manifest file based on a number of requirements files. The
requirements locking is usually done by executing a script on each
platform and having gazelle manifest generator be aware that more than
one requirements file may affect the outcome (e.g. the wheels that get
passed to modules map may come from multi_pip_parse rule) is generally
useful.

This change modifies the generation macro to concatenate the requirements
files into one before passing it to the manifest generator.


aignas:
Supersedes #1235.

f0rmiga (approved):

Issue #1295, opened 2023-07-05
Use a common constant for detecting bzlmod being enabled


rickeylev:
Various parts of the codebase detect whether bzlmod is enabled or not. Most of them copy/paste the same `if @ in Label(..)` trick, and use a comment to explain what they're doing.

Rather that copy/paste that everywhere, lets just have a constant defined that does this once, then code can simply check `if BZLMOD_ENABLED`.

PR #1287 is adding such a constant; this issue is for cleaning up the code to use it everywhere.

PR #1302, opened 2023-07-10
feat(bzlmod): Use a common constant for detecting bzlmod being enabled

chrislovecnm:
Various parts of the codebase detect whether bzlmod is enabled or not. Most of them copy/paste the same if @ in Label(..) trick and use a comment to explain what they're doing.

Rather than copy/paste that everywhere, this commit uses a constant defined that does this once and reuses the constant value to determine if bzlmod is enabled.

Closes: https://github.com/bazelbuild/rules_python/issues/1295

### DIFF
--- a/python/cc/BUILD.bazel
+++ b/python/cc/BUILD.bazel
@@ -1,8 +1,8 @@
 # Package for C/C++ specific functionality of the Python rules.
 
 load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
+load("//python/private:bzlmod_enabled.bzl", "BZLMOD_ENABLED")
 load("//python/private:current_py_cc_headers.bzl", "current_py_cc_headers")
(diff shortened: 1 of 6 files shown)


chrislovecnm:
@rickeylev Thoughts on me breaking the `BZLMOD_ENABLED` to a separate file?  Otherwise, we have to depend on `bazel_skylib` in a bunch of new places.

rickeylev:
> @rickeylev Thoughts on me breaking the BZLMOD_ENABLED to a separate file? Otherwise, we have to depend on bazel_skylib in a bunch of new places.

SGTM. From the error you shared, I think the issue is the skylib dependency becomes newly introduced during the workspace/module-extension phase. Which yeah, if we don't actually need it during that phase, +1 to refactor things such that it doesn't need to be loaded.

rickeylev (approved):

PR #1307, opened 2023-07-11
cleanup(tests): Use new APIs in rules_testing 0.4.0

rickeylev:
* Use public APIs for DictSubject and StrSubject.
* Use rules_testing's StructSubject instead of our own.

Work towards 1297


chrislovecnm (approved):
LGTM

PR #1308, opened 2023-07-11
Updating requirements files and cleanup

chrislovecnm:
Updated the requirements files under examples, except for bzlmod. I excluded bzlmod because I found a bug.

This commit also removes examples/bzlmod/gazelle_python.yaml, which we do not need.  This is there because the bzlmod example previously included gazelle.


chrislovecnm:
Closing, duplicate PR

PR #1310, opened 2023-07-11
docs: Use correct pip extension path in generated release notes

rickeylev:
Also add a note that bzlmod support is still beta.


aignas (approved):
LGTM

Issue #1262, opened 2023-06-08
Entrypoint is not implemented for multiple pips


chrislovecnm:
So entry point does not work in multiple pips. It is not supported. So I have to have two different templates or modify code.

The pip hub, whl_alias both need to be modified.  The key is that we have this:

```
def entry_point(pkg, script = None):
    """entry_point returns the target of the canonical label of the package entrypoints.
    """
    if not script:
        script = pkg
    return "@@rules_python~override~pip~pip_39_{}//:rules_python_wheel_entry_point_{}".format(_clean_name(pkg), script)
```

Created by a repo rules like `@pip_39_pylint`.  But that is one of the spokes that has "39" for the Python version.

In the pip hub we need to have the capability to have:

```
def entry_point(pkg, script = None):
    """entry_point returns the target of the canonical label of the package entrypoints.
    """
    if not script:
        script = pkg
    return "@@rules_python~override~pip~pip_{}_{}//:rules_python_wheel_entry_point_{}".format(_clean_name(pkg),python_version, script)
``` 

We need the Python version in the pip hub via a select statement.

_Originally posted by @chrislovecnm in https://github.com/bazelbuild/rules_python/pull/1254#discussion_r1221659188_

rickeylev:
I've tagged this bzlmod because I consider it bzlmod specific -- we're using the multi-pip impl under the hood for bzlmod, but need it to support the same features as non-multi-pip. Workspace multi-pip not supporting entry_point is fine and just continues what it does today.

chrislovecnm:
@rickeylev ahat about the user providing the configuran information, or how about  put this into gazelle?

aignas:
I have created a PR on solving this in a slightly different way than @rickeylev proposed and did some cleanup after that.

rickeylev:
> I have created a PR on solving this in a slightly different way 

To summarize: This is PR #1294. The net differences to what I posted are:
  * whl_library_alias isn't used in the bzlmod impl anymore. Instead, the hub, using select(),
    directly points to the backing whl_library targets. (This makes sense; the hub is
    doing what whl_library_alias was doing, so we don't need two levels of indirection).
  * The entry_point() function computes and returns the select() expression directly (instead of returning a label to an alias with that expression).

[...]

PR #1316, opened 2023-07-14
fix(multi-versions): correctly default 'main' arg for transition rules

aignas:
This fixes a bug where the version-aware rules required `main` to always be explicitly
specified. This was necessary because the main file is named after the outer target
(e.g. "foo"), but usage of the main file is done by the inner target ("_foo"). The net
effect is the inner target looks for "_foo.py", while only "foo.py" is in srcs.

To fix, the wrappers set main, if it isn't already set, to their name + ".py"

Work towards #1262

### DIFF
--- a/python/config_settings/private/py_args.bzl
+++ b/python/config_settings/private/py_args.bzl
@@ -1,0 +1,42 @@
+# Copyright 2023 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""A helper to extract default args for the transition rule."""
+
+def py_args(name, kwargs):
+    """A helper to extract common py_binary and py_test args
+
+    See https://bazel.build/reference/be/python#py_binary and
+    https://bazel.build/reference/be/python#py_test for the list
+    that should be returned
+
+    Args:
+        name: The name of the target.
+        kwargs: The kwargs to be extracted from; MODIFIED IN-PLACE.
+
+    Returns:
+        A dict with the extracted arguments
+    """
+    return dict(
+        args = kwargs.pop("args", None),
+        data = kwargs.pop("data", None),
+        env = kwargs.pop("env", None),
+        srcs = kwargs.pop("srcs", None),
+        deps = kwargs.pop("deps", None),
+        # See https://bazel.build/reference/be/python#py_binary.main
+        # for default logic.
+        # NOTE: This doesn't match the exact way a regular py_binary searches for
+        # it's main amongst the srcs, but is close enough for most cases.
+        main = kwargs.pop("main", name + ".py"),
+    )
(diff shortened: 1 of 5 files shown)


python/extensions/private/defs.bzl.tmpl
rickeylev, line 1: I don't see this used anywhere -- I'm guessing this leaked over from your other PR? Please remove

python/config_settings/private/py_args.bzl
```diff
@@ -0,0 +1,40 @@
...
+    """
+    return dict(
+        args = kwargs.pop("args", None),
+        data = kwargs.pop("data", None),
+        env = kwargs.pop("env", None),
+        srcs = kwargs.pop("srcs", None),
+        deps = kwargs.pop("deps", None),
+        # See https://bazel.build/reference/be/python#py_binary.main
+        # for default logic.
+        main = kwargs.pop("main", name),
```
rickeylev, line 39:
```suggestion
        # See https://bazel.build/reference/be/python#py_binary.main
        # for default logic.
        # NOTE: This doesn't match the exact way a regular py_binary searches for
        # it's main amongst the srcs, but is close enough for most cases.
        main = kwargs.pop("main", name),
```

python/config_settings/private/py_args.bzl
```diff
@@ -0,0 +1,40 @@
...
+def py_args(name, kwargs):
+    """A helper to extract common py_binary and py_test args
+
+    See https://bazel.build/reference/be/python#py_binary and
+    https://bazel.build/reference/be/python#py_test for the list
+    that should be returned
+
+    Args:
+        name: The name of the target.
+        kwargs: The kwargs to be extracted from.
```
rickeylev, line 26:
```suggestion
        kwargs: The kwargs to be extracted from; MODIFIED IN-PLACE
```

tests/config_settings/transition/py_args_tests.bzl
```diff
@@ -0,0 +1,68 @@
...
+
+def _test_py_args_default(env):
+    actual = py_args("foo", {})
+
+    want = {
+        "args": None,
+        "data": None,
+        "deps": None,
+        "env": None,
+        "main": "foo",
```
rickeylev, line 30: Main should be a .py file, i.e. name + ".py" -- please fix.

I'm surprised this didn't didn't cause a failure somewhere. When given an explicit main, the py_binary should have failed when it tried to look for a non-existent file with the same name as itself

tests/config_settings/transition/py_args_tests.bzl
```diff
@@ -0,0 +1,68 @@
...
+    want = {
+        "args": ["some", "args"],
+        "data": ["data"],
+        "deps": ["deps"],
+        "env": {"key": "value"},
+        "main": "__main__.py",
+        "srcs": ["__main__.py"],
+    }
+    env.expect.that_dict(actual).contains_exactly(want)
+    env.expect.that_dict(kwargs).keys().contains_exactly(["visibility"])
```
rickeylev, line 58: Also check that kwargs doesn't contain one of the keys, to verify that it is modified in place.

rickeylev:
Some fun trivia about how py_binary finds main: it's basically a suffix search. e.g.

```
search_for = name + ".py"
main = None
for src in srcs:
  if src.endswith(search_for):
    main = None
    break
```

rickeylev (changes requested):
Mostly lgtm, but one bug to fix and some nits

aignas:
@rickeylev, yeah, the fact that `main` parameter does not have to be a string ending with `.py` suffix also surprised me.

As for the comment about modification in place test, it is there already because the `contains_exactly` is used for assertions. Tested this by adding an extra `kwarg` (e.g. `tags`) into the input and the expectation started failing.

rickeylev (approved):

PR #1321, opened 2023-07-16
feat: add a tool to update internal dependencies

aignas:
Before this change the updates to the dependencies would happen very
seldomly, with this script, I propose we do it before each minor version
release. Adding a shell script and adding a reminder to the release
process may help with that.


tools/print_pip_deps.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash
+#
+# A script to manage internal dependencies
+
```
chrislovecnm, line 4: Please 
```shell
set -euxo pipefail
```

tools/print_pip_deps.sh
```diff
@@ -0,0 +1,53 @@
...
+internal_deps=(
```
chrislovecnm, line 5: Please add instructions on where these come from. Do we have a requirements file? Not sure if that would work

aignas:
@chrislovecnm, thanks for the initial comments. I have decided to rewrite the `jq` based approach with a little less magical Python implementation. This means that users don't have to have any extra dependencies on their machines.

chrislovecnm:
@aignas I'm going to let @rickeylev or some other Python guru do a code review on this.  My only comment is that you don't have public function docs.

DEVELOPING.md
```diff
@@ -8,6 +8,16 @@ Before running through the release it's good to run the build and the tests loca
 also test-drive the commit in an existing Bazel workspace to sanity check functionality.
 
 #### Steps
+1. (Optional) modify the `./python/pip_install/tools/requirements.txt` file and run `./tools/update_pip_deps.sh` script:
+   1. Using system python:
```
rickeylev, line 12: How does one know whether they should use the system Python or toolchain Python?

Is there a certain case the system python has to be used for? I'd rather not recommend it at all because it just sounds like a source of confusion -- we don't know what Python version it is, so different people might get different results, and then we churn the deps unknowingly.

DEVELOPING.md
```diff
@@ -8,6 +8,16 @@ Before running through the release it's good to run the build and the tests loca
 also test-drive the commit in an existing Bazel workspace to sanity check functionality.
 
 #### Steps
+1. (Optional) modify the `./python/pip_install/tools/requirements.txt` file and run `./tools/update_pip_deps.sh` script:
```
rickeylev, line 11: Lets move this to another section that is focused on  updating internal tool dependencies. These aren't really release steps; it's a chore that should be done regularly but also needs verification.

DEVELOPING.md
```diff
@@ -8,6 +8,16 @@ Before running through the release it's good to run the build and the tests loca
...
+   1. Using system python:
+   ```bash
+   ./tools/update_pip_deps.py
+   ```
+   1. Using bazel toolchain:
+   ```bash
+   bazel run --enable_bzlmod @python_3_10_9//:python3 -- ./tools/update_pip_deps.py
```
rickeylev, line 18: Ah, er, hm. Out of scope for this PR, but, isn't 3.10.9 both correct _and_ incorrect?

IIRC, the tools get run with the Python version from e.g. `pip.parse(python_version=...)` (and similar for workspace builds).

So it's correct if someone is using 3.10, but not _really_ correct for someone using e.g. 3.9.

Please tell me I'm wrong? :)

Otherwise -- ugh gah >.<

tools/update_file.py
```diff
@@ -0,0 +1,137 @@
+#!/usr/bin/env python3
```
rickeylev, line 1: This doesn't look like a directly invokable file? Above, it's used as a library. I don't see the code that directly invokes it (i.e. where name is `__main__`).

As such, please remove the shebang, if-main and main() from this file.

tools/update_file.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+def update_file(
+    path: pathlib.Path,
+    snippet: str,
+    start_marker: str,
+    end_marker: str,
+    dry_run: bool = True,
+):
+    """update a file on disk to replace text in a file between two markers.
```
rickeylev, line 59:
```suggestion
    """Update a file on disk to replace text in a file between two markers.
```

tools/update_file.py
```diff
@@ -0,0 +1,137 @@
...
+    Args:
+        path: pathlib.Path, the path to the file to be modified.
+        snippet: str, the snippet of code to insert between the markers.
+        start_marker: str, the text that marks the start of the region to be replaced.
+        end_markr: str, the text that marks the end of the region to be replaced.
+        dry_run: bool, if set to True, then the file will not be written and instead we are going to print a diff to
+            stdout.
+    """
+    with open(path) as f:
+        input = f.readlines()
```
rickeylev, line 70: Since path is a pathlib.Path, this can simply be: `path.read_text()`

rickeylev (changes requested):

DEVELOPING.md
aignas: Good point, did not want a case where the deps may break the bazel setup (e.g. incorrect coverage deps are updated). However, once the PR is merged we will have a known to work setup and we can do `git checkout` upon failures to update the dependencies.

tools/update_file.py
aignas: Good catch, it was a remnant of the times when the `update_pip_deps.py` was a shell script. Removed.

aignas (commented):
Added extra tests.

rickeylev (approved):

PR #1328, opened 2023-07-19
refactor: add a version label function for consistent labels

aignas:
Before this PR there would be at least a few places where we would be
converting a `X.Y.Z` version string to a shortened `X_Y` or `XY` string segment
to be used in repository rule labels. This PR adds a small utility function
that helps making things consistent.

Work towards #1262, split from #1294.


chrislovecnm (approved):
LGTM - are we doing this anywhere else?  Looking to see if we can replace more code with the function call.

Issue #1132, opened 2023-03-16
Incorrect wheel filename for distribution names with hyphen


alokpr:
# 🐞 bug report

### Affected Rule

py_wheel

### Is this a regression?

Yes, 0.10

### Description

py_wheel_rule creates wheel with escaped strings. py_wheel_dist however creates wheel with unescaped strings.

## 🔬 Minimal Reproduction

See https://github.com/bazelbuild/rules_python/issues/517

alokpr:
The minimal change to fix this would be:
```
diff --git a/tools/wheelmaker.py b/tools/wheelmaker.py
index 6138c93..d05158e 100644
--- a/tools/wheelmaker.py
+++ b/tools/wheelmaker.py
@@ -85,7 +85,7 @@ class WheelMaker(object):
         if self._build_tag:
             components.append(self._build_tag)
         components += [self._python_tag, self._abi, self._platform]
-        return "-".join(components) + ".whl"
+        return "-".join([escape_filename_segment(c) for c in components]) + ".whl"
 
     def filename(self) -> str:
         if self._outfile:
-- 

But I wonder if py_wheel_rule can be simplified. It seems like the same logic of escaping filename components is being duplicated in multiple places.

Issue #883, opened 2022-11-16
Support for PEP 440 local version identifiers is broken


vonschultz:
# 🐞 bug report

### Affected Rule

The issue is caused by the rule: `py_wheel`.

### Is this a regression?

 Yes, the previous version in which this bug was not present was: 0.3.0

### Description

Support for [PEP 440 local version identifiers](https://peps.python.org/pep-0440/#local-version-identifiers) is broken. While a version number such as `1.0.0+local` is correctly reflected in the `*.dist-info/METADATA` file, it is not correctly reflected in commands such as `pip list`, `pip show` or `pip freeze`, which show this version number as `1.0.0-local`, with `-` instead of `+`.

## 🔬 Minimal Reproduction

<pre><code>load("@rules_python//python:packaging.bzl", "py_wheel")

py_wheel(
    name = "testwheel",
    distribution = "test",
    version = "1.0.0+local",
    visibility = ["//visibility:public"],
) </code></pre>

and a `WORKSPACE.bazel` file according to [version 0.3.0](https://github.com/bazelbuild/rules_python/releases/tag/0.3.0) (works), [version 0.4.0](https://github.com/bazelbuild/rules_python/releases/tag/0.4.0) (exhibits problem) or [version 0.13.0](https://github.com/bazelbuild/rules_python/releases/tag/0.13.0) (exhibits problem).

Build the wheel, install in a venv, run `pip list`, see the correct version `1.0.0+local` (in case of rules_python 0.3.0) or incorrect version `1.0.0-local` (in newer versions).

## 🔥 Exception or Error

If used with a package index and a `requirements.txt` file, you'll find that the `-` version number raises `pip._vendor.packaging.version.InvalidVersion` or `has inconsistent version` (depending on `pip` version), and the `+` version number produces `Could not find a version that satisfies the requirement`.

Using `pip 22.3.1` (you may need to `pip install --upgrade pip`):
<pre><code>$ pip install test==1.0.0+local -f rules_python-0.13.0/bazel-bin/
Looking in links: rules_python-0.13.0/bazel-bin/
ERROR: Could not find a version that satisfies the requirement test==1.0.0+local (from versions: 1.0.0-local)
ERROR: No matching distribution found for test==1.0.0+local</code></pre>
<pre><code>$ pip install test==1.0.0-local -f rules_python-0.13.0/bazel-bin/
Looking in links: rules_python-0.13.0/bazel-bin/
[...]

PR #1331, opened 2023-07-20
feat(py_wheel): Normalize name and version

vonschultz:
Added the `incompatible_normalize_name` feature flag to normalize the package distribution name according to latest Python packaging standards. Defaults to `False` for the time being.

Added the `incompatible_normalize_version` feature flag to normalize the package version according to PEP440 standard. This also adds support for local version specifiers (versions with a `+` in them), in accordance with PEP440. Defaults to `False` for the time being.

Instead of following the obsolete PEP 427 escaping procedure for distribution names and versions, we should use the rules specified by https://packaging.python.org/en/latest/specifications (sections "Package name normalization" and "Binary distribution format"). For the versions, this means normalizing them according to PEP 440.

Added as feature flags to avoid forcing the user to deal with breaking changes when upgrading `rules_python`:

- Distribution names have stronger requirements now: "A valid name consists only of ASCII letters and numbers, period, underscore and hyphen. It must start and end with a letter or number." https://packaging.python.org/en/latest/specifications/name-normalization/

- Versions must be valid PEP 440 version identifiers. Previously versions such as "0.1-2-3" would have been accepted; that is no longer the case.

- The file name of generated wheels may have changed, if the distribution name or the version identifier wasn't in normalized form.

- The wheelmaker now depends on `packaging.version`, which means the `py_wheel` user now needs to load pip dependencies in their `WORKSPACE.bazel` file:
  ```
  load("@rules_python//python/pip_install:repositories.bzl", "pip_install_dependencies")

  pip_install_dependencies()
  ```

Fixes bazelbuild/rules_python#883. Fixes bazelbuild/rules_python#1132.


vonschultz:
Could I get some help with the test failures? They seem to be related to this line:
https://github.com/bazelbuild/rules_python/blob/5c5ab5bd9577a284784d1c8b27bf58336de06010/python/runfiles/BUILD.bazel#L55
where, of course, `{BUILD_EMBED_LABEL}` is not a valid version according to PEP 440. Now, I have been assuming that once we get to `wheelmaker.py` any stamping should already have taken place, taking care of all the placeholders, and the version we get should be a valid PEP 440 version. Is this mistaken? Should `wheelmaker.py` be forced to handle invalid versions with placeholders in them? Or should the `{BUILD_EMBED_LABEL}` be replaced with a valid version identifier?

Looks like @alexeagle and @rickeylev were involved in the patch (https://github.com/bazelbuild/rules_python/commit/1722988c)  that introduced `{BUILD_EMBED_LABEL}`.  Any insights or opinions?

aignas:
I think by default we are not stamping the builds within the CI, so the `embed-label` is not substituted. This would mean that a restrictive version name validation may need to be conditional based on whether the build is being stamped.

chrislovecnm:
@vonschultz does this change fix this issue: https://github.com/bazelbuild/rules_python/issues/1132

alexeagle:
Even in an unstamped build, version placeholders should be replaced by a constant value like `0.0.0-PLACEHOLDER`. IMO it was always incorrect that we left literal `{BUILD_EMBED_LABEL}` segments in output filenames.

vonschultz:
Yes, @chrislovecnm, this fixes #1132. I updated the pull request message accordingly.

chrislovecnm:
@aignas or @rickeylev PTAL

rickeylev (commented):
re: requiring pep 440 compliant names: I don't think we should try to perform such strict validation. Doing the basic transforms to make a string more pep 440 compatible (e.g. `-` to `.` or w/e) makes sense because we want to produce output that other tools can consume. But we don't really care if a user does e.g. `.after` instead of `.post`, or does `.post0.dev4.rc1`.

> The wheelmaker now depends on packaging.version, which means the py_wheel user now needs to load pip dependencies in their WORKSPACE.bazel file

I'm pretty sure we call pip_install_dependencies() internally, so users don't need to worry about this. You can remove this from the description.

vonschultz:
> re: requiring pep 440 compliant names: I don't think we should try to perform such strict validation. Doing the basic transforms to make a string more pep 440 compatible (e.g. `-` to `.` or w/e) makes sense because we want to produce output that other tools can consume. But we don't really care if a user does e.g. `.after` instead of `.post`, or does `.post0.dev4.rc1`.

I don't think we in `rules_python` should be making up our own rules for how Python versions work. The authoritative source is the specification for the binary distribution format at https://packaging.python.org/en/latest/specifications/binary-distribution-format/, and it says "Version numbers should be normalised according to [PEP 440](https://peps.python.org/pep-0440/)." We might also look at how influential projects such as `setuptools` handle it, and they require PEP 440-conforming versions. https://setuptools.pypa.io/en/latest/history.html#v66-0-0

In my mind one of two approaches makes sense. Either we take it upon us to do the normalization, in which case we should do it according to the specification, or we put the responsibility on the user entirely, documenting that the `version` attribute of a `py_wheel` needs to be a version normalized according to PEP 440, and skip any validation and sanitation whatsoever. That would certainly make for a less complex implementation, but I'm not sure that's a good idea.

> > The wheelmaker now depends on packaging.version, which means the py_wheel user now needs to load pip dependencies in their WORKSPACE.bazel file
> 
> I'm pretty sure we call pip_install_dependencies() internally, so users don't need to worry about this. You can remove this from the description.

I tested this. If I just do a minimal `WORKSPACE.bazel` file,
```

http_archive(
    name = "rules_python",
    sha256 = "0a8003b044294d7840ac7d9d73eef05d6ceb682d7516781a4ec62eeb34702578",
    strip_prefix = "rules_python-0.24.0",
    url = "https://github.com/bazelbuild/rules_python/releases/download/0.24.0/rules_python-0.24.0.tar.gz",
)

py_repositories()
```
in a toy project, as documented on https://github.com/bazelbuild/rules_python/releases/tag/0.24.0, and then try to use a `py_wheel` it will work with the 0.24.0 release, and it won't work when running with `bazel build --override_repository=rules_python=/home/von/src/github.com/vonschultz/rules_python ...`, unless we change the toy project to include a call to `pip_install_dependencies()`.

It's true that we call `pip_install_dependencies()` internally in `rules_python`, but the end user doesn't benefit from that, so from the point of view of the end user it's a breaking change.

python/private/py_wheel.bzl
```diff
@@ -203,6 +203,427 @@ _DESCRIPTION_FILE_EXTENSION_TO_TYPE = {
...
+                "Name '%s' has bad character '%s'." % (name, character),
+            )
+    if escaped.endswith("_"):
+        fail(
+            "A valid name must end with a letter or number.",
+            "Name '%s' does not." % name,
+        )
+    return escaped
+
+def normalize_pep440(version):
```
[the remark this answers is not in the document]
vonschultz: It's a parser, yes. If you have ideas on how to make it more approachable, I'm all for that.

My preference would have been to use a regular expression instead — in particular because PEP 440 includes a regular expression that we could take more or less as it is. But regular expressions aren't supported in Starlark, so here I am writing a parser instead. (Though whether a regular expression is more approachable than a parser is perhaps debatable, unless you're really into regular expressions.)

My apologies for the complexity, but PEP 440 is hundreds of lines of English, so some of the complexity seems unavoidable. On the other hand, PEP 440 is a fixed document. Once we have a `normalize_pep440()` function that works to our satisfaction, there will be few (if any) reasons to go back and change it. That's good for maintainability, isn't it, functions that have no conceivable reason to change as the product evolves.

tests/py_wheel/py_wheel_tests.bzl
```diff
@@ -16,7 +16,12 @@
 load("@rules_testing//lib:analysis_test.bzl", "analysis_test", "test_suite")
 load("@rules_testing//lib:util.bzl", rt_util = "util")
 load("//python:packaging.bzl", "py_wheel")
+load(
```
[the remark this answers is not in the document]
vonschultz: What's the recommended max line length for `.bzl` files in the `rules_python` project? In the Python world, PEP 8 says 79 characters, and `black` defaults to 88 characters. This line would be 95 characters, if written on one line.

alexmirrington:
For wheel builds that don't require specific semantic versioning (e.g. for internal tools that are installed in multiple other projects at a company), I think it's pretty common to append a `+sha` suffix to the wheel version. Currently the validation doesn't support this, and it would be very helpful if it did.

vonschultz:
> For wheel builds that don't require specific semantic versioning (e.g. for internal tools that are installed in multiple other projects at a company), I think it's pretty common to append a `+sha` suffix to the wheel version. Currently the validation doesn't support this, and it would be very helpful if it did as proposed in #883

Not sure I follow @alexmirrington. What you are talking about is a local version identifier. That is not currently supported on the main branch, which is #883. Fixing #883 that is the purpose of this pull request, and the validation that's done here does support local version identifiers: 
https://github.com/bazelbuild/rules_python/pull/1331/files#diff-e5abf313ad5bebb5c80d6984829326c40288f734daecb0217f5d13ef75adad23R600-R610

python/private/py_wheel.bzl
aignas, line 242: Please move this function into a separate file so that it is easier to maintain.

python/private/py_wheel.bzl
```diff
@@ -203,6 +203,427 @@ _DESCRIPTION_FILE_EXTENSION_TO_TYPE = {
...
+        contexts.append({"norm": "", "start": start})
+        return contexts[-1]
+
+    def close_context():
+        """Close the current context successfully and merge the results."""
+        finished = contexts.pop()
+        contexts[-1]["norm"] += finished["norm"]
+        contexts[-1]["start"] = finished["start"]
+
+    def is_(reference):
```
aignas, line 278: Functions that do not depend on `contexts` or `version` should be defined as regular functions outside the `normalize_pep440`. This should be easy when the whole code is in a separate file. They should be all unexported and prefixed with `_` to make sure they stay private.

So far I only see `is_`, `is_not`, `in-`, `isdigit` and `isalnum`. They can all be small function defined separately.

python/private/py_wheel.bzl
```diff
@@ -203,6 +203,427 @@ _DESCRIPTION_FILE_EXTENSION_TO_TYPE = {
...
+            parsing step starts.
+        """
+        contexts.append({"norm": "", "start": start})
+        return contexts[-1]
+
+    def close_context():
+        """Close the current context successfully and merge the results."""
+        finished = contexts.pop()
+        contexts[-1]["norm"] += finished["norm"]
+        contexts[-1]["start"] = finished["start"]
```
aignas, line 276: You could create an object like struct with the core parser fields that are private and pass it around. See for an example here: https://github.com/bazelbuild/rules_testing/blob/f90632532f3bb6c33ac4d27e5f89b55474cdb925/lib/private/action_subject.bzl#L29

I checked that all of the `accept*` methods should not be part of the struct because we cannot call `self.bar` inside of `_foo` that is accessible via `self.foo`.

python/private/py_wheel.bzl
```diff
@@ -203,6 +203,427 @@ _DESCRIPTION_FILE_EXTENSION_TO_TYPE = {
...
+            saying if we want to accept the token.
+          value: the string to add if there's a match, or, if `value`
+            is a function, the function to apply to the current token
+            to get the string to add.
+
+        Returns:
+          whether a token was accepted.
+        """
+
+        context = contexts[-1]
```
aignas, line 309: consider renaming the variable to `ctx` and making a `parser.context()` function to access the last context without accessing a private `contexts` array. Right now `contexts` implements a stack, but it could also in theory be a linked list and the interface of the parser should be agnostic to the underlying data structure.

python/private/py_wheel.bzl
```diff
@@ -203,6 +203,427 @@ _DESCRIPTION_FILE_EXTENSION_TO_TYPE = {
...
+
+        if context["start"] >= len(version):
+            return False
+
+        token = version[context["start"]]
+
+        if predicate(token):
+            if type(value) in ["function", "builtin_function_or_method"]:
```
aignas, line 317: Since this is called many times, consider optimizing and doing 
```suggestion
            t = type(value)
            if t == "function" or t == "builtin_function_or_method":
```

That way we are not creating a list just for comparison.

python/private/py_wheel.bzl
```diff
@@ -203,6 +203,427 @@ _DESCRIPTION_FILE_EXTENSION_TO_TYPE = {
...
+        sense where they appear. Before the stamping has happened, a
+        resulting wheel file name containing a placeholder will not
+        actually be valid.
+
+        """
+        context = open_context(contexts[-1]["start"])
+
+        if not accept(is_("{"), str):
+            contexts.pop()
+            return False
```
aignas, line 342: Consider making a `parser.discard()` method which returns False

python/private/py_wheel.bzl
```diff
@@ -203,6 +203,427 @@ _DESCRIPTION_FILE_EXTENSION_TO_TYPE = {
...
+        start = context["start"]
+
+        for i in range(start, len(version) + 1):
+            if not accept(isdigit, str) and not accept_placeholder():
+                if i - start >= 1:
+                    if context["norm"].isdigit():
+                        # PEP 440: Integer Normalization
+                        context["norm"] = str(int(context["norm"]))
+                    close_context()
+                    return True
```
aignas, line 372: Consider making a `parser.accept()` method that returns True. This will simplify the code quite a bit.

python/private/py_wheel.bzl
```diff
@@ -203,6 +203,427 @@ _DESCRIPTION_FILE_EXTENSION_TO_TYPE = {
...
+
+        if accept(is_("+"), "+") and accept_alnum():
+            accept_separator_alnum_sequence()
+            close_context()
+            return True
+
+        contexts.pop()
+        return False
+
+    open_context(0)
```
aignas, line 612: It would be better to initialize the `contexts` list with the right first entry to begin with, because `contexts = []` is not a valid list when used in the parsing functions.

tools/wheelmaker.py
```diff
@@ -32,9 +34,54 @@ def commonpath(path1, path2):
...
+
+    """
+
+    try:
+        return str(packaging.version.Version(version))
+    except packaging.version.InvalidVersion:
+        pass
+
+    sanitized = re.sub(r'[^a-z0-9]+', '.', version.lower()).strip('.')
+    substituted = re.sub(r'\{\w+\}', '0', version)
```
aignas, line 77: I assume this has been taken from somewhere, could you paste a link to the source please?

aignas (commented):
In general the parsing approach LGTM, but the tests could be improved to unit test parts of the parser and then sanity check that the parts work together. Right now we only have integration tests.

I have made some experiments in my branch and created a PR to this branch in https://github.com/vonschultz/rules_python/pull/1 so that it would be easy for you to review and merge.

@rickeylev, I agree that this looks a little bit daunting at first and may require a few iterations of cleanup whilst we get to understand the code, however, once the nested `def`s are removed and we have a `stack`-like parser with very few methods it suddenly becomes much easier to grok in my opinion.

python/private/py_wheel.bzl
vonschultz: Done.

python/private/py_wheel.bzl
vonschultz: Done.

python/private/py_wheel.bzl
vonschultz: Done.

python/private/py_wheel.bzl
vonschultz: Done.

python/private/py_wheel.bzl
vonschultz: Are you sure? I'm not measuring any noticeable speed-up. Measuring with
```
time bazelisk test //tests/py_wheel:test_pep440_normalization
```
(with `bazelisk clean` in between), and sometimes the first approach is faster, sometimes the second approach is faster.

If it's speed we're after, it was noticeably faster before the readability improvements you suggested. (Sorry, I do like the readability improvements.) If we're going for readability over speed, I like the list comprehension.

python/private/py_wheel.bzl
vonschultz: Done.

python/private/py_wheel.bzl
vonschultz: Done.

tools/wheelmaker.py
vonschultz: It wasn't copy-pasted, if that's what you're thinking. 

Reference for `version.lower()`: https://peps.python.org/pep-0440/#case-sensitivity
Reference for `re.sub(..., '.', ...)`: https://peps.python.org/pep-0440/#local-version-segments
Reference for only allowing a-z, 0-9 and period: https://peps.python.org/pep-0440/#local-version-identifiers
Reference for removing periods from beginning and end, and collapsing sequences of multiple periods to one: https://peps.python.org/pep-0440/#appendix-b-parsing-version-strings-with-regular-expressions

Substituting Bazel placeholders with "0" is essentially arbitrary, but I took inspiration from https://github.com/bazelbuild/rules_python/pull/1331#issuecomment-1668422845

See also the docstring for the purpose of this: this is Bazel-specific functionality to coerce a non-PEP440 version with a placeholder into a PEP440-compliant version, in the case the build is performed without stamping. We can do whatever we want here, as long as the result is a normalized PEP440 version.

tests/py_wheel/py_wheel_tests.bzl
```diff
@@ -92,8 +97,109 @@ def _test_content_type_from_description_impl(env, target):
...
+        for input_dev in input_devs
+    ]
+    locals = [
+        [normalized_local, input_local]
+        for normalized_local, input_locals in locals.items()
+        for input_local in input_locals
+    ]
+    postfixes = ["", "  ", " \t\r\n"]
+    i = 0
+    for nepoch, iepoch in epochs:
```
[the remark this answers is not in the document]
vonschultz: I agree that we should be able to remove some of the unit tests that overlap if we're smart about it. I'm not sold on testing the individual helper functions (e.g. epoch parsing) and then having a single integration test.

What is our purpose with having tests here? Arguably, one purpose is to find bugs, and then the main consideration is what bugs we expect to find. This is then best informed by what sort of bugs I did find when developing this, and apart from the trivial it was what you might call integration issues between different parts of the parser, as exemplified by the `0bdev` issue, where the pre-release parser was broken in the presence of a dev release.

Another purpose that we shouldn't discount is to facilitate future changes. What future changes do we anticipate? Not changes to the requirements of PEP 440, but, as you point out, we may refactor the code as new techniques on writing good starlark code are developed. A refactoring is helped a lot by a reasonably comprehensive test suite _that you don't have to change_ when you refactor the code. Unit tests that guard the current detailed design aren't useful; tests that guard the unchanging requirements are useful. If we replace the parser with regular expressions, there won't be an epoch parser as a distinct thing. Likewise if we replace the recursive descent parser with a table-based parser generated by a parser generator using a grammar definition. The person refactoring this won't thank the test writer if he has to totally rewrite the tests as well. Therefore, I think all the tests should be testing `normalize_pep440()`, not any helper functions it uses, and if we want tests to be smaller and more specific, we should move towards testing the individual requirements of PEP 440.

tests/py_wheel/py_wheel_tests.bzl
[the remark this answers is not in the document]
vonschultz: `buildifier` doesn't do line breaks. At some point, a human needs to step in and say "enough is enough". If you really want 120, I guess that's up to you.

tests/py_wheel/py_wheel_tests.bzl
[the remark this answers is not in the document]
aignas: Thanks for the extensive write up, I agree with the points made here, mostly. I think it is important for someone who is seeing the code for the first time to understand what to change if something starts failing. Tests should be also DAMP (https://enterprisecraftsmanship.com/posts/dry-damp-unit-tests/) because it makes them easier to read and understand.

I am OK with resolving this for now.

python/private/py_wheel.bzl
aignas: Thanks for being rigorous and checking it. LGTM.

tools/wheelmaker.py
aignas: Thanks for explaining. I am wondering if some of this should be documented in the code.

aignas: I see the docstring now. I would prefer code comments in here, but I am fine with the current state as well.

aignas (approved):
Thanks for the PR, @rickeylev, with the new structure of having fewer things being passed via the closure and nested defs, does this look better?

python/private/py_wheel.bzl
```diff
@@ -238,8 +275,8 @@ def _py_wheel_impl(ctx):
     version = _replace_make_variables(ctx.attr.version, ctx)
 
     outfile = ctx.actions.declare_file("-".join([
-        _escape_filename_segment(ctx.attr.distribution),
-        _escape_filename_segment(version),
+        _escape_filename_distribution_name(ctx.attr.distribution),
+        normalize_pep440(version),
```
aignas, line 279: It would be great if we could make this new behaviour optional and gated under a boolean flag so that users that depend on the broken behaviour have more time to migrate.

The plan could be:
* Add a flag named `normalize_version_pep440` which defaults to `False` for now.
* If the flag is set to `True` then the new codepath is used.
* In a future release the `incompatible_normalize_version` is set to `True` but users have ability to set it to False if required.

That way the change is no longer breaking.

Consider adding an extra dictionary with args named:
```starlark
_feature_flags = {
    "incompatible_normalize_version": attr.bool(
        default = False,
        doc = "Normalize the package version according to PEP440 standard",
    ),
    "incompatible_explicit_distribution_stamp_vars": attr.bool(
        default = False,
        doc = "Change the format of how the stamp variables can be used to " +
              "in the distribution name. With this option set to True, the " +
              "user has to pass stamp variables enclosed in '{}', e.g. " +
              "'{BUILD_TIMESTAMP}'."
    ),
}
```

examples/wheel/wheel_test.py
```diff
@@ -384,7 +384,7 @@ def test_rule_expands_workspace_status_keys_in_wheel_metadata(self):
             "rules_python",
             "examples",
             "wheel",
-            "example_minimal_library_BUILD_USER_-0.1._BUILD_TIMESTAMP_-py3-none-any.whl",
+            "example_minimal_library{build_user}-0.1.{BUILD_TIMESTAMP}-py3-none-any.whl",
```
aignas, line 387: This probably needs a separate feature flag as well, so that users can depend on the old behaviour.

aignas (changes requested):

python/private/py_wheel.bzl
rickeylev: To add some context (I was out and about on my phone when Ignas asked me about this, so I didn't have time to respond in full):

The key point is to decouple upgrading the rules_python version from having to address incompatible changes. This allows a more incremental and manageable upgrade process. A user can upgrade rules_python, opt into behavior where they can, and followup later where they can't. The next version (which has the incompatible behavior by default) blocks them from upgrading, but they're still able to make their current version forward compatible.

examples/wheel/wheel_test.py
vonschultz: I made two feature flags, `incompatible_normalize_name` and `incompatible_normalize_version`. If both of them are `False`, you will get the old behavior here.

python/private/py_wheel.bzl
```diff
@@ -117,6 +118,21 @@ See [`py_wheel_dist`](/docs/packaging.md#py_wheel_dist) for more info.
     ),
 }
 
+_feature_flags = {
+    "incompatible_normalize_name": attr.bool(
+        default = False,
+        doc = "Normalize the package distribution name according to current " +
+              "standards.",
```
aignas, line 125: nit:
```suggestion
        doc = """\
Normalize the package distribution name according to latest
Python packaging standards.

See https://packaging.python.org/en/latest/specifications/binary-distribution-format/#escaping-and-unicode
and https://packaging.python.org/en/latest/specifications/name-normalization/.

Apart from the valid names according to the above, we also accept
'{' and '}', which may be used as placeholders for stamping.
""",
```

Later this doc string could be moved to the documentation of the `attr.distribution` documentation. Having the links to the official Python documentation may be nice.

aignas (commented):
Thank you very much for doing the extra work to add the feature flags. I think the only thing remaining before merging is to add a line in the `CHANGELOG.md` file.

vonschultz:
Should I update the pull request description? Does this still count as a breaking change? Does this still count as fixing https://github.com/bazelbuild/rules_python/issues/883 and https://github.com/bazelbuild/rules_python/issues/1132, or should that wait for a pull request that changes the default value of the feature flags to "True"?

aignas:
Yes, the PR title and description should be updated and I would be of an opinion that it still fixes #883 and #1132 as users have a way to opt-into the new behavior.

aignas (approved):

PR #1335, opened 2023-07-21
feat: stop generating imports when not necessary

linzhp:
When gazelle:python_root is not set or is at the root of the repo, we don't need to set imports for python rules, because that's the Bazel's default. This would reduce unnecessary verbosity.

### DIFF
--- a/gazelle/python/target.go
+++ b/gazelle/python/target.go
@@ -122,6 +122,11 @@ func (t *targetBuilder) setTestonly() *targetBuilder {
 // case, the value we add is on Bazel sub-packages to be able to perform imports
 // relative to the root project package.
 func (t *targetBuilder) generateImportsAttribute() *targetBuilder {
+	if t.pythonProjectRoot == "" {
+		// When gazelle:python_root is not set or is at the root of the repo, we don't need
+		// to set imports, because that's the Bazel's default.
+		return t
+	}
 	p, _ := filepath.Rel(t.bzlPackage, t.pythonProjectRoot)
 	p = filepath.Clean(p)
 	if p == "." {
--- a/gazelle/python/testdata/dependency_resolution_order/bar/BUILD.out
+++ b/gazelle/python/testdata/dependency_resolution_order/bar/BUILD.out
@@ -3,6 +3,5 @@ load("@rules_python//python:defs.bzl", "py_library")
 py_library(
     name = "bar",
     srcs = ["__init__.py"],
-    imports = [".."],
     visibility = ["//:__subpackages__"],
 )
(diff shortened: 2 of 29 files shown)


f0rmiga:
I don't think the logic is correct, based on the changed tests. Could you elaborate, please? I may be missing something.

linzhp:
Consider this workspace, where `imports` are not set in any of the py rules:

```
-- MODULE.bazel --
module(name = "py_imports")

rules_python_version = "0.24.0"

bazel_dep(name = "rules_python", version = rules_python_version)
bazel_dep(name = "rules_python_gazelle_plugin", version = rules_python_version)
bazel_dep(name = "gazelle", version = "0.31.0", repo_name = "bazel_gazelle")

python.toolchain(
    is_default = True,
    python_version = "3.9",
)
-- WORKSPACE --
-- pkg1/BUILD.bazel --
load("@rules_python//python:defs.bzl", "py_binary")

py_binary(
    name = "pkg1",
    srcs = ["__init__.py", "main.py"],
    main = "main.py",
    deps = ["//pkg2"],
)
-- pkg1/__init__.py --
-- pkg1/main.py --
from pkg2 import lib

print(lib.hello())
-- pkg2/BUILD.bazel --
load("@rules_python//python:defs.bzl", "py_library")

py_library(
    name = "pkg2",
    srcs = ["__init__.py", "lib.py"],
)
-- pkg2/__init__.py --
-- pkg2/lib.py --
def hello():
    return "hello world"
```

It's perfectly fine to run:

```
bazel run --enable_bzlmod //pkg1
INFO: Analyzed target //pkg1:pkg1 (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
Target //pkg1:pkg1 up-to-date:
  bazel-bin/pkg1/pkg1
INFO: Elapsed time: 0.125s, Critical Path: 0.01s
INFO: 1 process: 1 internal.
INFO: Running command line: bazel-bin/pkg1/pkg1
hello world
```

This is because Bazel use the root of the workspace as PYTHONPATH by default, so the module pkg2/lib.py is imported as `from pkg2 import lib`. From Gazelle's perspective, this means when `# gazelle:python_root` is not set or is set at the root of the repo, we don't need to assign any value to `imports` attribute. 

The tests are updated because `# gazelle:python_root` is not set in those tests, and no `import` attribute is populated for the py rules in those tests, reflecting the new behavior.

Note that the new behavior only affects new targets generated by Gazelle. Existing `imports` attribute will be preserved, whether they are written by hand or populated by Gazelle in the past.

Does this make sense?

f0rmiga (approved):
It makes sense! Thanks!

Issue #1330, opened 2023-07-19
Consider removing duplicate code in PyPI package name normalization


aignas:
This is a followup to @chrislovecnm comment in #1329.

If #1329 lands, then there are the following parts where we have name sanitisation and normalization.

First are the `requirement`, `whl_requirement`, `data_requirement`, `dist_info_requirement` and `entry_point` macros are using a vendored copy of the function named `_clean_name`. We could just use the function introduced in #1329 most of the time, but we have to expose it either in `//python/pip_install:normalize_name.bzl` or somewhere else because as it is right now, it's invisible from the `pip_parse` hub repos.

Second is the Python implementation in `//python/pip_install/tools/lib:bazel.py` has a copy and we cannot remove it that easily. The only (semi) sensible way would be to change the implementation of the `whl_library`:
* Starlark implementation declares a `metadata.json` output file that gets written by the `wheel_installer.py` script.
* The script extracts or builds the wheel and outputs the metadata to the `metadata.json` file as opposed to creating a `BUILD.bazel` file.
* Starlark implementation reads the `metadata.json` file and then constructs the `BUILD.bazel` file.

The benefit of the `whl_library` refactor would be that it would make the wheel installer a single purpose tool and we could pass the extra `BUILD.bazel` content or do patching within the starlark context and it would make annotation handling easier. The whole argument parsing for the `wheel_installer.py` is likely to also become simpler.

aignas:
The only remaining part for this is to use the `normalize_name` in the hub macros, but I'll leave this for a later time because there is a discussion in #979 on the potential replacement for the repository macros.

PR #1336, opened 2023-07-22
refactor(whl_library): move bazel file generation to Starlark

aignas:
Before this PR, the `wheel_installer` was doing three things:
1. Downloading the right wheel.
2. Extracting it into the output directory.
3. Generating BUILD.bazel files based on the extracted contents.

This PR is moving the third part into the `whl_library` repository rule
and it has the following benefits:
* We can reduce code duplication and label sanitization functions in
  rules_python.
* There are many things that the `wheel_installer` does not care anymore
  and we don't need to change less code when extending `whl_library` as
  we can now do many things in starlark directly.
* It becomes easier to change the API of how we expose the generated
  BUILD.bazel patching because we only need to change the Starlark
  functions.

Work towards #1330.


aignas:
FYI @chrislovecnm, I decided to see how this idea of generating BUILD.bazel and other extra files worked out in reality and I think I am happy with the solution. It makes working on the #1262 and extending the API easier, hence me doing this refactor before extending the `whl_library` more.

python/pip_install/private/generate_whl_library_build_bazel.bzl
```diff
@@ -0,0 +1,225 @@
...
+    srcs = glob(["*.whl"], allow_empty = True),
+    data = {whl_file_deps},
+)
+
+py_library(
+    name = "{name}",
+    srcs = glob(
+        ["site-packages/**/*.py"],
+        exclude={srcs_exclude},
+        allow_empty = True,
```
rickeylev, line 72:
```suggestion
        # Empty sources are allowed to support wheels that don't have any
        # pure-Python code, e.g. pymssql, which is written in Cython.
        allow_empty = True,
```

python/pip_install/private/generate_whl_library_build_bazel.bzl
```diff
@@ -0,0 +1,225 @@
...
+        data_exclude: more patterns to exclude from the data attribute of generated py_library rules.
+        tags: list of tags to apply to generated py_library rules.
+        entry_points: A dict of entry points to add py_binary rules for.
+        annotation: The annotation for the build file.
+
+    Returns:
+        A complete BUILD file as a string
+
+    We allow for empty Python sources as for Wheels containing only compiled C code
+    there may be no Python sources whatsoever (e.g. packages written in Cython: like `pymssql`).
```
rickeylev, line 107: nit: move the comment inline where the allow_empty setting is actually used (see the other suggested edit)

```suggestion
```

python/pip_install/private/generate_whl_library_build_bazel.bzl
```diff
@@ -0,0 +1,225 @@
...
+    additional_content = []
+    data = []
+    srcs_exclude = []
+    data_exclude = [] + data_exclude
+    dependencies = sorted(dependencies)
+    tags = sorted(tags)
+
+    for entry_point, entry_point_script_name in entry_points.items():
+        additional_content.append(
+            _generate_entry_point_rule(
```
rickeylev, line 119: nit: use named args for this call.

python/pip_install/private/generate_whl_library_build_bazel.bzl
```diff
@@ -0,0 +1,225 @@
...
+        script (str): The path to the entry point's python file.
+        pkg (str): The package owning the entry point. This is expected to
+            match up with the `py_library` defined for each repository.
+
+    Returns:
+        str: A `py_binary` instantiation.
+    """
+    return _ENTRY_POINT_RULE_TEMPLATE.format(
+        name = name,
+        src = str(script).replace("\\", "/"),
```
rickeylev, line 223: This str() call looks unnecessary? The input is already a string, no?

rickeylev (changes requested):
Mostly lgtm

rickeylev (approved):

PR #1341, opened 2023-07-24
revert(bzlmod)!: allow bzlmod pip.parse to implicitly use default python version

chrislovecnm:
Reverts bazelbuild/rules_python#1303

The main issue is that `pip.parse()` accepts a locked requirements file -- this means
the requirements are specific to a particular Python version[1]. Because the default Python
version can arbitrarily change, the lock file may not be valid for the Python version
that is used at runtime. The net result is a module will use dependencies for e.g. Python
3.8, but will use 3.9 at runtime. Additionally, the dependencies resolved for 3.8 will
be created under names such as `@foo_39` (because that's the python_version pip.parse sees),
which is just more confusing.

BREAKING CHANGE:
  * pip.parse() must have `python_version` explicitly set. Set it to the Python version
    used to resolve the requirements file.

[1] Lock files aren't necessarily version specific, but we don't currently support the
environment markers in lock files to make them cross-python-version compatible.

### DIFF
--- a/examples/bzlmod/MODULE.bazel
+++ b/examples/bzlmod/MODULE.bazel
@@ -11,10 +11,6 @@ local_path_override(
     path = "../..",
 )
 
-# Setting python.toolchain is optional as rules_python
-# sets a toolchain for you, using the latest supported version
-# of Python.  We do recomend that you set a version here.
-
 # We next initialize the python toolchain using the extension.
 # You can set different Python versions in this block.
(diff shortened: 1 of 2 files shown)


rickeylev:
Mostly LGTM, but I want to rewrite the description to give more context. "causes bugs" and "we decided" doesn't give enough information for future readers. It's my EOD now, though, so I don't have the energy to the description and add to the merge queue.

rickeylev (approved):

PR #1344, opened 2023-07-24
fix: Don't require default Python version for pip hubs

### LINKED ISSUE #1320: Don't require a pip.parse call for the default python version
There's a check in pip.bzl that requires that each pip hub have a call with the default python version. This should be removed because a child doesn't necessarily need anything setup for the default python version. If it uses the version-aware rules, then it won't be affected by what the default is.

While it's likely the child module isn't using the version-aware rules, we don't know one-way or the other during extension execution. So we'll just have to assume the MODULE author knows what they're doing. The failure mode in this case is dependencies will be resolved for e.g. Python 3.9, but at runtime, the targets will use e.g. Python 3.10.

The check is around here: https://github.com/bazelbuild/rules_python/blob/main/python/extensions/pip.bzl#L294

cc @chrislovecnm

rickeylev:
This fixes the issue where a sub-module was required to always have a pip.parse() call configured for the default Python version if it used any pip.parse() call. Such a
requirement puts sub-modules in an impossible situation:
  * If they don't have the default version, they'll get an error.
  * If they register the default version, but also register a specific version,
    they'll potentially cause an error if a root module changes the default to
    match their specific version (because two pip.parse() calls for the same version
    are made, which is an error).

The requirement to have the default version registered for a pip hub was only present to satisfy the `whl_library_alias` repository rule, which needed a Python version to map `//conditions:default` to.

To fix, the `whl_library_alias` rule's `default_version` arg is made optional. When None is passed, the `//conditions:default` condition is replaced with a `no_match_error` setting. This prevents the pip hub from being used with the version-unaware rules, but that makes sense: no wheels were setup for that version, so it's not like there is something that
can be used anyways.

Fixes #1320


rickeylev:
OK, CI is happy now. PTAL

chrislovecnm (approved):
Lgtm

PR #1346, opened 2023-07-25
refactor: support rendering pkg aliases without whl_library_alias

aignas:
Before this PR the only way to render aliases for PyPI package repos
using the version-aware toolchain was to use the `whl_library_alias` repo.
However, we have code that is creating aliases for packages within the hub repo
and it is natural to merge the two approaches to keep the number of layers of
indirection to minimum.

- feat: support alias rendering for python aware toolchain targets.
- refactor: use render_pkg_aliases everywhere.
- refactor: move the function to a private `.bzl` file.
- test: add unit tests for rendering of the aliases.

Split from #1294 and work towards #1262 with ideas taken from #1320.

### DIFF
--- a/tests/pip_hub_repository/render_pkg_aliases/render_pkg_aliases_test.bzl
+++ b/tests/pip_hub_repository/render_pkg_aliases/render_pkg_aliases_test.bzl
@@ -1,0 +1,220 @@
+# Copyright 2023 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"render_pkg_aliases tests"
+
+load("@rules_testing//lib:test_suite.bzl", "test_suite")
+load("//python/private:render_pkg_aliases.bzl", "render_pkg_aliases")  # buildifier: disable=bzl-visibility
+
+_tests = []
+
+def _test_legacy_aliases(env):
+    actual = render_pkg_aliases(
+        bzl_packages = ["foo"],
+        repo_name = "pypi",
+    )
+
+    want = {
+        "foo/BUILD.bazel": """\
+package(default_visibility = ["//visibility:public"])
+
+alias(
+    name = "foo",
+    actual = ":pkg",
+)
+
+alias(
+    name = "pkg",
+    actual = "@pypi_foo//:pkg",
+)
+
+alias(
+    name = "whl",
+    actual = "@pypi_foo//:whl",
+)
+
+alias(
+    name = "data",
+    actual = "@pypi_foo//:data",
+)
+
+alias(
+    name = "dist_info",
+    actual = "@pypi_foo//:dist_info",
+)""",
+    }
+
+    env.expect.that_dict(actual).contains_exactly(want)
+
+_tests.append(_test_legacy_aliases)
+
+def _test_all_legacy_aliases_are_created(env):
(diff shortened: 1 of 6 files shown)


rickeylev (changes requested):
> feat: support alias rendering for python aware toolchain targets.

What does this mean? I don't see the behavior change that is new functionality?

Otherwise, mostly LGTM

rickeylev:
Also, does this overlap with https://github.com/bazelbuild/rules_python/pull/1344 ? The files don't, but I see `no_match_error` as part of the logic, but don't see `no_match_error` in the removed code, nor recall that in other existing code.

aignas:
> Also, does this overlap with #1344 ? The files don't, but I see `no_match_error` as part of the logic, but don't see `no_match_error` in the removed code, nor recall that in other existing code.

@rickeylev, with regards to overlap with #1344, I wanted to ensure that the new alias rendering function can fail if the appropriate version is not found. Let me know if I should move addition of the error message to a separate PR.

chrislovecnm:
@aignas can you provide more details to @rickeylev about the error message so that we can get that error handling in this PR?

aignas:
After having a look at #1344 I see that the PRs overlap with a great extent. To keep this PR scoped I would like to not add more things here but leave it for a followup PR. My plan would be to:
* Move of the `pip_hub_repository` from `python/pip_install` to `python/extensions/private.bzl` and remove the usage of `whl_library_alias` and use `render_pkg_aliases` for everything. This would not include anything to do with the `entry_point` macro usage from the main hub repo yet because I think it is a bit complex and I'd like to first do the refactor of getting rid of some of the complexity in the `pip.bzl` extension.
* Fix the usage of the `pip.parse` extension with various toolchain versions, which is the #1340 PR (or a different one entirely). I am afraid that #1328 may have regressed the functionality in cases where we the users have the specific `X.Y.Z` Python toolchain versions specified in their `MODULE.bazel`. Previously it would work but we would have weird and undocumented `pip_XYZ_` prefixes and weird `@python_versions//X.Y.Z:defs.bzl` imports. If you would like to revert #1328 in the mean time, feel free to do so.

rickeylev:
FYI 1344 has been merged. If this PR obviates the no_match_error logic I put in that PR, that's mostly fine -- just retain the detailed error message. When select() fails to match, it's a very confusing state to figure out unless you're well acquainted with bazel query, bazel info, configurations, etc

> I wanted to ensure that the new alias rendering function can fail if the appropriate version is not found.

Makes sense, SGTM.

> My plan would be to

Mostly SGTM, I think? I'd have to see the code to get a better sense of what this means and looks like. I'm convinced that whl_library_alias is extraneous. My gut says that something else is, too, but I'm not sure to what, especially with entry_point looming around.

> If you would like to revert https://github.com/bazelbuild/rules_python/pull/1328 in the mean time, feel free to do so.

I don't think that's necessary. We've only ever gave examples for and documented usage of Major.Minor formats. If anything, we can just fail or warn if a non-Major.Minor value is given until we figure out how to properly support that.

aignas:
@rickeylev, could you have a look at this once more? I have rebased on `main` and have made minor changes to reuse the same `no_match_error` between the `whl_library_alias` and this new utilitiy. I think this is ready to be merged as is without any extra work that I mentioned in my previous comment.

tests/pip_hub_repository/render_pkg_aliases/render_pkg_aliases_test.bzl
```diff
@@ -0,0 +1,251 @@
...
+        whl_map = {
+            "bar-baz": ["3.2.3", "3.1.3"],
+        },
+    )
+
+    want_key = "bar_baz/BUILD.bazel"
+    want_content = """\
+package(default_visibility = ["//visibility:public"])
+
+_NO_MATCH_ERROR = \"\"\"\\
```
rickeylev, line 152: Ah, I guess I better get on https://github.com/bazelbuild/rules_testing/issues/65 so you can do e.g. `that_str(actual).matches("_NO_MATCH_ERROR*<everything else>")`. The long error message is nice, but embedding it in the test is brittle and distracting.

(I'm not asking to change this, it's fine as is; it's easy to update if we change the error message)

rickeylev (approved):

PR #1349, opened 2023-07-27
docs: Better explain when and how to use toolchains for bzlmod

rickeylev:
This explains the different ways to register toolchains and how to use them.

Also fixes python_aliases -> python_versions repo name

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,36 +41,97 @@ the older way of configuring bazel with a `WORKSPACE` file.
 
 **IMPORTANT: bzlmod support is still in Beta; APIs are subject to change.**
 
+The first step to using rules_python with bzlmod is to add the dependency to
+your MODULE.bazel file:
+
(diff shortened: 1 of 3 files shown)


aignas (approved):
Thanks for this explanation, I think it becomes much clearer of what is possible and how the toolchains can be used.

That said, I wonder if if we should advise users to only use `X.Y` instead of `X.Y.Z` in the python_version field. All of the docs only have examples in that case and I am wondering if saying `python_version = "3.11.1"` even makes sense here.

rickeylev:
Yeah good point, we should be explicit about that. Only major.minor is
supported. If major.minor.patch works, it's just incidental.

rickeylev:
Updated the attribute docs for the python_version attributes.

Solumin (approved):

PR #1352, opened 2023-08-01
feat: Add s390x release

namrata-ibm:
Include s390x in release and update python-build-standalone to 3.9.17, 3.10.12, 3.11.4.
[Latest python-build-standalone release](https://github.com/indygreg/python-build-standalone/releases/tag/20230726) has s390x support added.

These changes are needed to build TensorFlow on s390x, which is currently blocked due to missing support.

### DIFF
--- a/python/versions.bzl
+++ b/python/versions.bzl
@@ -153,6 +153,19 @@ TOOL_VERSIONS = {
         },
         "strip_prefix": "python",
     },
+    "3.9.17": {
+        "url": "20230726/cpython-{python_version}+20230726-{platform}-{build}.tar.gz",
+        "sha256": {
+            "aarch64-apple-darwin": "73dbe2d702210b566221da9265acc274ba15275c5d0d1fa327f44ad86cde9aa1",
+            "aarch64-unknown-linux-gnu": "b77012ddaf7e0673e4aa4b1c5085275a06eee2d66f33442b5c54a12b62b96cbe",
+            "ppc64le-unknown-linux-gnu": "c591a28d943dce5cf9833e916125fdfbeb3120270c4866ee214493ccb5b83c3c",
+            "s390x-unknown-linux-gnu": "01454d7cc7c9c2fccde42ba868c4f372eaaafa48049d49dd94c9cf2875f497e6",
+            "x86_64-apple-darwin": "dfe1bea92c94b9cb779288b0b06e39157c5ff7e465cdd24032ac147c2af485c0",
+            "x86_64-pc-windows-msvc": "9b9a1e21eff29dcf043cea38180cf8ca3604b90117d00062a7b31605d4157714",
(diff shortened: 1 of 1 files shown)


aignas:
@namrata-ibm, #1340 should fix what you mention in your message, then we would only have `3.11:defs.bzl` instead of `3.11.1:defs.bzl`. That said, it would be fixing it only for `bzlmod` and for traditional workspaces we would likely need to do additional fixes.

namrata-ibm:
Thank you @aignas for pointing out. 
Will check #1340 for further updates. If it is ongoing work and might take longer, then I can revert the MINOR_MAPPING changes from this PR. 
However it would be good to have, to avoid explicit python version setting(eg in TensorFlow) during build on s390x.

aignas:
I think the minor mapping changes are good and they should stay in this PR,
if any other code is broken, it should be fixed.

chrislovecnm:
@namrata-ibm please take a look at CI/CD failures: https://buildkite.com/bazel/rules-python-python/builds/5380#0189af9c-468d-4b55-ac85-9954d0197118

namrata-ibm:
@chrislovecnm Thank you for checking, all the CI failures are same and addressed in above [comments](https://github.com/bazelbuild/rules_python/pull/1352#issuecomment-1659721946). It would be great to hear comments about the queries listed up.

aignas:
@namrata-ibm, could you see if you can fix the WORKSPACE with the expected command please? Replacing `3.11.1` with `3.11.4` everywhere would be good enough.

namrata-ibm:
@aignas I added the WORKSPACE change and all earlier failures have passed in latest run. However I see `Buildifier` failed with 1 format issue. I tried running `buildifier python/versions.bzl` with v6.1.0 , however I do not see any error. Could you please help?

aignas:
@namrata-ibm, this looks like a CI flake.

chrislovecnm:
> @namrata-ibm, this looks like a CI flake.

@aignas did we hit a bug in buildifier?

namrata-ibm:
@aignas @chrislovecnm @rickeylev could you please provide your inputs on how to proceed further?
Thank you.

python/versions.bzl
```diff
@@ -153,6 +153,19 @@ TOOL_VERSIONS = {
         },
         "strip_prefix": "python",
     },
+    "3.9.17": {
```
chrislovecnm, line 156: @rickeylev @aignas any other Python versions that we should have `s390x` support for?

chrislovecnm (changes requested):
Thanks for the PR. Please run `buildifier` to get CI passing.  I think we should have two PRs here.

1. include support for the new Python versions and support for `s390x`. 
2. after that PR bump the mappings to increase the default version numbers

namrata-ibm:
@chrislovecnm Thank you, I have updated this PR to keep only s390x and python version updates.
Will raise PR for Minor mapping changes later.

Could you please check and help merge this?

Issue #1326, opened 2023-07-17
Error with ppc versions


chrislovecnm:
We are getting

e.g.:
```
bazel query "deps(@python_3_10//:python_headers)"
INFO: Repository rules_python~0.24.0~python~python_3_10_x86_64-unknown-linux-gnu instantiated at:
  callstack not available
Repository rule python_repository defined at:
  /private/var/tmp/_bazel_melliot/84bf3287741d1703ed98e552f5dae756/external/rules_python~0.24.0/python/repositories.bzl:381:36: in <toplevel>
INFO: Repository rules_python~0.24.0~python~python_3_10_x86_64-pc-windows-msvc instantiated at:
  callstack not available
Repository rule python_repository defined at:
  /private/var/tmp/_bazel_melliot/84bf3287741d1703ed98e552f5dae756/external/rules_python~0.24.0/python/repositories.bzl:381:36: in <toplevel>
INFO: Repository rules_python~0.24.0~python~python_3_10_aarch64-apple-darwin instantiated at:
  callstack not available
Repository rule python_repository defined at:
  /private/var/tmp/_bazel_melliot/84bf3287741d1703ed98e552f5dae756/external/rules_python~0.24.0/python/repositories.bzl:381:36: in <toplevel>
INFO: Repository rules_python~0.24.0~python~python_3_10_aarch64-unknown-linux-gnu instantiated at:
  callstack not available
Repository rule python_repository defined at:
  /private/var/tmp/_bazel_melliot/84bf3287741d1703ed98e552f5dae756/external/rules_python~0.24.0/python/repositories.bzl:381:36: in <toplevel>
ERROR: Evaluation of query "deps(@python_3_10//:python_headers)" failed: preloading transitive closure failed: no such package '@[unknown repo 'python_3_10_ppc64le-unknown-linux-gnu' requested from @rules_python~0.24.0~python~python_3_10]//': The repository '@[unknown repo 'python_3_10_ppc64le-unknown-linux-gnu' requested from @rules_python~0.24.0~python~python_3_10]' could not be resolved: No repository visible as '@python_3_10_ppc64le-unknown-linux-gnu' from repository '@rules_python~0.24.0~python~python_3_10'

```

The problem is that we don’t have ppc in every version of each python version. We are using PLATFORMS.keys() and hitting cases were we don’t have ppc.

chrislovecnm:
[...]

PR #1364, opened 2023-08-05
fix: Fixing select when we don't have specific platforms

chrislovecnm:
In certain releases of Python we do not have various platforms like ppc.  When using the call PLATFORMS.keys() we get all of the platforms instead of the platforms that are included with the Python version.

We get the error "no such package" because, for instance, python_3_10_ppc64le-unknown-linux-gnu does not exist in Python 3.10.9 (see versions.bzl).

You can recreate this error by running a bazel query command inside the examples/bzlmod directory.  If you run:

bazel query "deps(@python_3_10//:python_headers)"

With a version of the rules previous to this commit, you will get an error.  With this commit, you get a printout of all of the python_headers for all of the platforms that are supported in the Python 3.10 version used by the repository.

Instead of using PLATFORMS.keys(), we use the sha256 keys in TOOL_VERSIONS dict based on the full Python version. This change now returns only the platforms supported in the Python version.

Fixes: #1326


chrislovecnm:
@rickeylev any idea why we are doing:

```starlark
alias(name = "python_headers",  actual = select({":" + item: "@python_3_10_" + item + "//:python_headers" for item in TOOL_VERSIONS['3.10.9']['sha256'].keys()}))
```

Instead of:

```starlark
alias(name = "python_headers",  actual = "@python_3_10_x86_64-unknown-linux-gnu//:python_headers")
```

I'm uncertain how to test that this change does not break anything.  Here is most of the BUILD.bazel file that is created by this rule:

```starlark
 alias(name = "files",           actual = select({":" + item: "@python_3_10_" + item + "//:files" for item in TOOL_VERSIONS['3.10.9']['sha256'].keys()}))
 alias(name = "includes",        actual = select({":" + item: "@python_3_10_" + item + "//:includes" for item in TOOL_VERSIONS['3.10.9']['sha256'].keys()}))
 alias(name = "libpython",       actual = select({":" + item: "@python_3_10_" + item + "//:libpython" for item in TOOL_VERSIONS['3.10.9']['sha256'].keys()}))
 alias(name = "py3_runtime",     actual = select({":" + item: "@python_3_10_" + item + "//:py3_runtime" for item in TOOL_VERSIONS['3.10.9']['sha256'].keys()}))
 #alias(name = "python_headers",  actual = select({":" + item: "@python_3_10_" + item + "//:python_headers" for item in TOOL_VERSIONS['3.10.9']['sha256'].keys()}))
 alias(name = "python_headers",  actual = "@python_3_10_x86_64-unknown-linux-gnu//:python_headers")
 alias(name = "python_runtimes", actual = select({":" + item: "@python_3_10_" + item + "//:python_runtimes" for item in TOOL_VERSIONS['3.10.9']['sha256'].keys()}))
 alias(name = "python3",         actual = select({":" + item: "@python_3_10_" + item + "//:" + ("python.exe" if "windows" in item else "bin/python3") for item in TOOL_VERSIONS['3.10.9']
```

Running:

```
bazel query "deps(@python_3_10//:python_headers)"
```

Works but only gives you the headers for specific host platform with the change.  With the select statement, we get the headers for every platform.

chrislovecnm:
@rickeylev I tried it out in https://github.com/bazelbuild/rules_python/pull/1365 and it is passing CI

chrislovecnm:
We may want #1365 instead

chrislovecnm:
Based on a conversation with @rickeylev we need to have a select with all of the PLATFORMS that are supported in it.  This PR wins ;)

python/private/toolchains_repo.bzl
```diff
@@ -150,27 +152,41 @@ def _toolchain_aliases_impl(rctx):
     is_windows = (os_name == WINDOWS_NAME)
     python3_binary_path = "python.exe" if is_windows else "bin/python3"
 
+    # uncertain if we need to do this check or not
+    full_python_version = rctx.attr.python_version
+    if TOOL_VERSIONS.get(full_python_version) == None:
+        full_python_version = MINOR_MAPPING[full_python_version]
```
aignas, line 158: I had code in #1294 that was similar to this, it was called `full_version`, which would always ensure that the python version is in `X.Y.Z` format. I think we could cherry pick it from there. I did the cherry pick in #1340, but that has not been reviewed yet.

python/private/toolchains_repo.bzl
```diff
@@ -150,27 +152,41 @@ def _toolchain_aliases_impl(rctx):
...
+
+    # In previous versions of this code we used PLATFORMS.each()
+    # for the select statements below.  This gave use a list of all of
+    # the platforms and not just the platforms supported by the
+    # Python version.  For instance not all Python versions include
+    # support for Power8 (ppc).
+    # We now get the supported platforms from the TOOL_VERSIONS dict.
```
aignas, line 165: Comments should explain about the current code and not previous code because then the question is how much context about previous code we should leave?
```suggestion
```

python/private/toolchains_repo.bzl
aignas: And a nit: this could be changed to be `full_python_version = MINOR_MAPPING.get(rctx.attr.python_version, rctx.attr.python_version)` as we should have an entry in `MINOR_MAPPING` for each entry in `TOOL_VERSIONS`.

chrislovecnm: I will wait for #1340 to get in and use that.

python/private/toolchains_repo.bzl
chrislovecnm: Agreed, and done

chrislovecnm:
I will wait for https://github.com/bazelbuild/rules_python/pull/1340 to get in and use the code to find the full version.

aignas:
Implemented by #1473 and I have verified that the bug no longer exists. Verified that
```
bazel query "deps(@python_3_10//:python_headers)"
```
succeeds in the `examples/bzlmod`.

PR #1365, opened 2023-08-05
fix: Fixing select when we don't have specific platforms

chrislovecnm:
In certain releases of Python we do not have various platforms like
ppc. When using the call PLATFORMS.keys() we get all of the
platforms instead of the platforms that are included with the Python
version.

We get the error "no such package" because, for instance,
python_3_10_ppc64le-unknown-linux-gnu does not exist in Python
3.10.9 (see versions.bzl).

You can recreate this error by running a bazel query command inside
the examples/bzlmod directory. If you run:

```
bazel query "deps(@python_3_10//:python_headers)"
```

With a version of the rules previous to this commit, you will get
an error. With this commit, you get a printout of all of the
python_headers for all of the platforms that are supported in the
Python 3.10 version used by the repository.

Instead of using PLATFORM.keys() we are only listing the actual
repository based on the host.  For instance:

```starlark
alias(
    name = "python_headers",
    actual = "@python_3_10_x86_64-unknown-linux-gnu//:python_headers",
)
```

Instead of:

```starlark
alias(
    name = "python_headers",
    actual = select( {":" + item: "@python_3_10_" + item + "//:python_headers" for item in PLATFORM.keys()})),
)
```

Fixes: #1326


chrislovecnm:
TODO: update comments

PR #1395, opened 2023-09-01
refactor(bzlmod)!: simplify pip.parse repository layout

### LINKED ISSUE #1255: Refactor for loops in pip repository
We are not doing the same for loop multiple times. See:

https://github.com/bazelbuild/rules_python/blob/28e15c2092a9512064e4b2a97a0fa3d1a83bc5ae/python/pip_install/pip_repository.bzl#L484

aignas:
Before this PR we would generate extra `alias` repos and the extra `hub` repo
for the `entry_point` macro usage. This PR removes the extras and delegates the
creation of version-aware aliases to the `render_pkg_aliases` internal
function. This reduces the number of repositories created by the `pip.parse`
extension.

Fixes #1255.

BREAKING CHANGE:

Note that this only affects bzlmod support, which is still beta.

* Bzlmod `pip.parse` no longer generates `{hub_name}_{py_version}` hub repos.
* Bzlmod `pip.parse` no longer generates `{hub_name}_{distribution}` hub repos.

These repos aren't part of a public API, but were typically used for the `entry_point`
macros. Instead, use `py_console_script_binary`, which is the supported replacement
for entry points under bzlmod. Directly referencing the underlying distribution
repos remains unsupported.


aignas:
@rickeylev or @chrislovecnm, could you please take a look?

aignas:
Depends on #1423.

rickeylev (approved):

PR #1397, opened 2023-09-01
refactor: upgrade certifi

cflewis:
Older versions of certifi allow for revoked HTTPS certificates. This
change updates usages of certifi to the first known-good version.

See https://security.snyk.io/vuln/SNYK-PYTHON-CERTIFI-5805047 and https://nvd.nist.gov/vuln/detail/CVE-2023-37920


aignas (approved):
Thanks for bumping the dependencies!

cflewis:
Can someone help me understand what's wrong with the Windows Bazel integration test? It's passing on *nix but not Windows, and I only have access to a Debian machine :(

rickeylev:
Chris and I tried to debug what was going on with Windows, but neither of us has a Windows machine, so it was slow going. We're still not entirely sure what's going on.

On the surface, the error is the requirements diff test failing on Windows. That isn't too unexpected (locked files are somewhat platform specific). However, the diff is about the certifi version (it wants 2021, while it gets 2023), which is exactly what is being changed. Ok, a bit weird, but whatever, lets just use `requirements_windows` and create windows-specific requirements file.

So Chris copy/pasted the locked requirements the tests wants into requirements_windows.txt. This then fails because `@pip_parsed_wheel` doesn't exist. Why? Because requirements_windows.txt doesn't have the `wheel` entry. Which is weird because `wheel` is specified in requirements.in.

So I'm not really sure what's going on. It almost seems like windows is ignoring requirements.in? Or partially ignoring it?

In any case, I'm in favor of just disabling the requirements_test diff checking for Windows. I'm pretty sure this isn't the first time Windows has given weird/incorrect behavior like this.

cflewis:
I'll wait for whatever consensus y'all need to have the test disabled on Windows. Then just tell me what to do and I'll update the PR.

groodt:
I'll try push what I think is a fix to your branch. I think you've updated the wrong file. The repo is a bit of a mess, but I think the file you need to change is: `tests/pip_repository_entry_points/requirements_windows.txt`

groodt (approved):
LGTM

PR #1406, opened 2023-09-08
fix: allowing to import code generated from proto with strip_import_prefix

linzhp:
When the `proto_library` has `strip_import_prefix`, the py files from proto are generated into a directory like `bazel-bin/tests/py_proto_library/proto/_virtual_imports` and symlinked in the runfiles as `<runfilesroot>/<workspace name>/tests/py_proto_library/proto/_virtual_imports`.
We need to add `<workspace name>/tests/py_proto_library/proto/_virtual_imports` to the `imports` of `_PyProtoInfo`, so it will be appended to `PYTHONPATH`.

Modified an existing example to demonstrate the scenario and verify the fix.


examples/py_proto_library/example.com/proto/BUILD.bazel
```diff
@@ -0,0 +1,14 @@
...
+py_proto_library(
+    name = "pricetag_proto_py_pb2",
+    visibility = ["//visibility:public"],
+    deps = [":pricetag_proto"],
+)
+
+proto_library(
+    name = "pricetag_proto",
+    srcs = ["pricetag.proto"],
+    strip_import_prefix = "/example.com",
```
aignas, line 13: nit: Could you add a comment pointing the users to the docs for `strip_import_prefix` please? E.g. I personally am not sure why we need `/example.com` instead of `example.com`.

python/private/proto/py_proto_library.bzl
```diff
@@ -109,6 +107,7 @@ def _py_proto_aspect_impl(target, ctx):
     return [
         _PyProtoInfo(
             imports = depset(
+                [proto_root],
```
aignas, line 110: Could you explain in a comment why this is needed, please. Do we have a test for this?

.bazelci/presubmit.yml
```diff
@@ -28,6 +28,8 @@ buildifier:
   build_targets:
     - "--"
     - "..."
+    # This has its own workspace and needs to run from there
+    - "-//examples/py_proto_library/..."
```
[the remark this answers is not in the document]
linzhp: The targets in `//examples/py_proto_library/...` are still matched by `...`, failing the CI because the proto_library tries to strip `example.com`, but the path starts with `examples/py_proto_library`

python/private/proto/py_proto_library.bzl
linzhp: Yes, that's why I changed the examples/py_proto_library

.bazelci/presubmit.yml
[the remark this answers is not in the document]
aignas: I would expect that changes proposed in https://github.com/linzhp/rules_python/pull/1 would remove the need for this change.

aignas:
Could you also please a line in the `CHANGELOG.md`?

aignas (approved):
Thanks

PR #1407, opened 2023-09-11
fix: Skip printing unneccesary warning.

matts1:
If the python version is explicitly provided by the root module, they should not be warned for choosing the same version that rules_python provides as default.

### DIFF
--- a/python/extensions/python.bzl
+++ b/python/extensions/python.bzl
@@ -86,16 +86,22 @@ def _python_impl(module_ctx):
             # much else that can be done. Modules don't know and can't control
             # what other modules do, so the first in the dependency graph wins.
             if toolchain_version in global_toolchain_versions:
-                _warn_duplicate_global_toolchain_version(
-                    toolchain_version,
-                    first = global_toolchain_versions[toolchain_version],
-                    second_toolchain_name = toolchain_name,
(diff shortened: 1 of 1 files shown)


aignas:
@matts1, in order to better understand the behaviour, could you post a snippet of a `MODULE.bazel` file that prints the warning, that with your PR would disappear?

matts1:
This is the relevant part of the file.
```
python.toolchain(
    is_default = True,
    python_version = "3.11",
)
```

Since rules_python declares the same version [here](https://github.com/bazelbuild/rules_python/blob/main/MODULE.bazel#L40-L48), it gives out a warning.

aignas:
I was confused about it and it seems that our examples have
```
```

and it means that the warning is not triggered because by default `rules_python` configures the toolchains without coverage.

If I add
```
python.toolchain(
    python_version = "3.11",
)
```
then the warning is generated on `main`, but not with this patch.

I'd be in favor of merging this patch as it seems weird that `rules_python` would raise a warning when the user passes `3.11` but not when `3.11.1` is passed.

aignas (approved):

rickeylev:
Yes, this is fine. It's expected that rules_python be overridden by the root module; we shouldn't warn when people do so.

PR #1412, opened 2023-09-12
fix: don't set distribs in version transitioning rule

comius:
This makes the version-aware transition rule compatible with an upcoming Bazel change that
disallows setting unknown attributes to None (the `distribs` attribute, in this case).

The `distribs` attribute was common to all rules, but it has been long deprecated and
it won't be part of every rule in upcoming Bazel versions. The previous implementation
resulted in setting `distribs = None` on the target. Bazel won't support setting
undefined attributes to None.

Addresses: https://github.com/bazelbuild/bazel/issues/19403

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,10 @@ A brief description of the categories of changes:
 
 ## Unreleased
 
+### Changed
+* (multi-version) The `distribs` attribute is no longer propagated. This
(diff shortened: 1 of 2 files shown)


rickeylev:
I'm assuming this means, in Bazel at head, that the `distribs` attribute is no longer part of the common attributes every rule has?

rickeylev (approved):
LGTM. I pushed a change to update the changelog.

PR #1418, opened 2023-09-18
internal: copy Starlark rule implementation from Bazel

rickeylev:
This is a copy of the Starlark implementation of the Python rules from Bazel. This code isn't
loaded and won't work as-is. Modifications to make it work will be made in
subsequent changes. It's almost pristine; changes are made to satisfy the
buildifier check.

Work towards #1069


Solumin (approved):

groodt (approved):
LGTM
🎉

PR #1423, opened 2023-09-20
refactor(toolchain): use a helper method to convert an X.Y version to X.Y.Z

aignas:
Before this PR in numerous places we would check the MINOR_MAPPING dict.
This PR adds a function that also fails if the X.Y format is not in the
MINOR_MAPPING dict making the code more robust.

Split from #1340 to unblock #1364.

### DIFF
--- a/python/private/full_version.bzl
+++ b/python/private/full_version.bzl
@@ -1,4 +1,4 @@
-# Copyright 2022 The Bazel Authors. All rights reserved.
+# Copyright 2023 The Bazel Authors. All rights reserved.
 #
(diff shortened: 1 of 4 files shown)


python/private/full_version.bzl
```diff
@@ -0,0 +1,43 @@
+# Copyright 2022 The Bazel Authors. All rights reserved.
```
f0rmiga, line 1:
```suggestion
# Copyright 2023 The Bazel Authors. All rights reserved.
```

f0rmiga (approved):

PR #1415, opened 2023-09-14
fix: gazelle: Fix non-hermetic runfiles lookup

fmeum:
`bazel.Runfiles` is a deprecated way to look up runfiles that can result in non-hermetic lookups. `github.com/bazelbuild/rules_go/go/runfiles` is the recommended package for this.

PR #1429, opened 2023-09-22
fix(gazelle): runfiles discovery

### LINKED ISSUE #1426: AssertionError: Cannot find .runfiles directory
https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/3319#018ab077-c764-4cc2-8b32-9fb08619ef90

rules_python: rules_python (example bzlmod build file running gazelle and pip integration test on :ubuntu: 20.04 LTS (OpenJDK 11, gcc 9.4.0))
Logs:
```
 File "/var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/3b10eb146ec4242f66f1b569edf0069e/execroot/_main/bazel-out/k8-opt-exec-ST-e846b08c7501/bin/external/rules_python_gazelle_plugin~override/python/std_modules", line 172, in FindModuleSpace
    raise AssertionError('Cannot find .runfiles directory for %s' % sys.argv[0])
AssertionError: Cannot find .runfiles directory for /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/3b10eb146ec4242f66f1b569edf0069e/execroot/_main/bazel-out/k8-opt-exec-ST-e846b08c7501/bin/external/rules_python_gazelle_plugin~override/python/std_modules
gazelle: failed to wait for std_modules: exit status 1
```
Culprit Found: [2f9c7d60ea676f9f8e929b20a3ff74e786995c91](https://github.com/bazelbuild/bazel/commit/2f9c7d60ea676f9f8e929b20a3ff74e786995c91)

CC Greenteam @comius

aignas:
Pass the environment generated by the `runfiles` helper so that the
Python launcher can find the runfiles correctly as implemented in
bazelbuild/bazel#14740.

Fixes #1426

### DIFF
--- a/gazelle/python/parser.go
+++ b/gazelle/python/parser.go
@@ -38,13 +38,20 @@ var (
 )
 
 func startParserProcess(ctx context.Context) {
-	parseScriptRunfile, err := runfiles.Rlocation("rules_python_gazelle_plugin/python/parse")
+	rfiles, err := runfiles.New()
+	if err != nil {
+		log.Printf("failed to create a runfiles object: %v\n", err)
+		os.Exit(1)
+	}
+
+	parseScriptRunfile, err := rfiles.Rlocation("rules_python_gazelle_plugin/python/parse")
 	if err != nil {
 		log.Printf("failed to initialize parser: %v\n", err)
 		os.Exit(1)
 	}
 
 	cmd := exec.CommandContext(ctx, parseScriptRunfile)
+	cmd.Env = rfiles.Env()
 
 	cmd.Stderr = os.Stderr
 
--- a/gazelle/python/parser.go
+++ b/gazelle/python/parser.go
@@ -51,7 +51,7 @@ func startParserProcess(ctx context.Context) {
 	}
 
 	cmd := exec.CommandContext(ctx, parseScriptRunfile)
-	cmd.Env = rfiles.Env()
+	cmd.Env = append(os.Environ(), rfiles.Env()...)
 
 	cmd.Stderr = os.Stderr
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,8 @@ A brief description of the categories of changes:
   dependencies improving initial build times involving external dependency
   fetching.
 
+* (gazelle) Improve runfiles lookup hermeticity.
+
 ## [0.25.0] - 2023-08-22
 
 ### Changed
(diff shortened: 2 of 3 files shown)


aignas:
@fmeum, could you please take a look? This is an additional change needed on top of your work in #1415.

fmeum (approved):

fmeum:
Just FYI: With https://github.com/bazelbuild/bazel-gazelle/pull/1632, Gazelle will be setting these variables and this change should only matter for tests at that point.

rickeylev:
I'm not familiar with Go, so its not clear to me why this works. All it seems to be doing is passing Env() to the subprocess? I don't see Env being modified. Do subprocesses in Go not inherit the current env? I sort of expected to see the PYTHON_RUNFILES envvar being set.

gazelle/python/parser.go
```diff
@@ -38,13 +38,20 @@ var (
...
+	}
+
+	parseScriptRunfile, err := rfiles.Rlocation("rules_python_gazelle_plugin/python/parse")
 	if err != nil {
 		log.Printf("failed to initialize parser: %v\n", err)
 		os.Exit(1)
 	}
 
 	cmd := exec.CommandContext(ctx, parseScriptRunfile)
+	cmd.Env = rfiles.Env()
```
fmeum, line 54: @aignas I think that this needs to be:

```suggestion
	cmd.Env = append(os.Env(), rfiles.Env()...)
```

fmeum (changes requested):

fmeum:
> I'm not familiar with Go, so its not clear to me why this works. All it seems to be doing is passing Env() to the subprocess? I don't see Env being modified. Do subprocesses in Go not inherit the current env? I sort of expected to see the PYTHON_RUNFILES envvar being set.

`rfiles.Env()` discovers runfiles (potentially by looking next to `argv[0]`) and then returns the environment variables that subprocesses need to discover runfiles. If no runfiles env variables are set for the outer process, the inner process wouldn't be able to discover runfiles without this line (which has a bug, see my comment).

rickeylev (approved):

rickeylev:
Thanks. LGTM after Fabian's comments are addressed.

f0rmiga (approved):

gazelle/python/parser.go
aignas: Good catch, I wonder if we should modify to include the `os.Env` in the `std_module.go` handling or not. For now I'll skip this to retain the current behaviour.

aignas:
I am going to leave this like this due to my observation in https://github.com/bazelbuild/rules_python/pull/1429#discussion_r1335314004. @rickeylev, @f0rmiga, feel free to merge it during your working hours if you agree to postpone including `os.Env()` in the `std_module.go`.

PR #1430, opened 2023-09-23
feat: removed uses of `_py_builtins.get_current_os_name()`

UebelAndre:
This change supports cross-platform remote execution where the Bazel client's platform can be different from the remote executors.

### DIFF
--- a/python/private/common/py_executable.bzl
+++ b/python/private/common/py_executable.bzl
@@ -47,6 +47,7 @@ load(
     "ALLOWED_MAIN_EXTENSIONS",
     "BUILD_DATA_SYMLINK_PATH",
     "IS_BAZEL",
+    "PLATFORMS_LOCATION",
     "PY_RUNTIME_ATTR_NAME",
 )
 
@@ -91,6 +92,17 @@ filename in `srcs`, `main` must be specified.
     allow_none = True,
 )
 
+def is_target_platform_windows(ctx):
+    """Determine if the target platform is a windows platform.
+
+    Args:
+        ctx (ctx): The rule's context object
+
+    Returns:
+        bool: True if the target platform is Windows.
+    """
+    return ctx.target_platform_has_constraint(Label("{}/os:windows".format(PLATFORMS_LOCATION)))
+
 def py_executable_base_impl(ctx, *, semantics, is_test, inherited_environment = []):
     """Base rule implementation for a Python executable.
 
@@ -114,7 +126,8 @@ def py_executable_base_impl(ctx, *, semantics, is_test, inherited_environment = 
     direct_sources = filter_to_py_srcs(ctx.files.srcs)
     output_sources = semantics.maybe_precompile(ctx, direct_sources)
     imports = collect_imports(ctx, semantics)
-    executable, files_to_build = _compute_outputs(ctx, output_sources)
+    is_windows = is_target_platform_windows(ctx)
+    executable, files_to_build = _compute_outputs(ctx, output_sources, is_windows)
 
     runtime_details = _get_runtime_details(ctx, semantics)
     if ctx.configuration.coverage_enabled:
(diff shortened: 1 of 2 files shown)


UebelAndre:
@rickeylev this is a followup to https://github.com/bazelbuild/bazel/issues/19587#issuecomment-1732206317 (related to https://github.com/bazelbuild/rules_python/issues/1069). I figured I'd sketch out a potential change that can remove the use of `_py_builtins.get_current_os_name()` to support cross-platform remote execution. Does something like this seem reasonable?

rickeylev:
Instead of adding a new rule and going through select(), use `ctx.target_platform_has_constraint` and add an implicit dependency on the various platform constraints. See attributes.bzl and `_appl_constraints` for an example.

Hm, I wonder if that's actually sufficient. If you look at the darwin support, it also has to add `ExecutionInfo({"requires-darwin": ""})` (common.bzl) so that tests/binaries _run_ on darwin, too. Is there a `requires-windows` magic tag, too?

UebelAndre:
> Instead of adding a new rule and going through select(), use `ctx.target_platform_has_constraint` and add an implicit dependency on the various platform constraints. See attributes.bzl and `_appl_constraints` for an example.
> 
> Hm, I wonder if that's actually sufficient. If you look at the darwin support, it also has to add `ExecutionInfo({"requires-darwin": ""})` (common.bzl) so that tests/binaries _run_ on darwin, too. Is there a `requires-windows` magic tag, too?

I didn't know about [ctx.target_platform_has_constraint](https://bazel.build/rules/lib/builtins/ctx#target_platform_has_constraint), neat!

I also don't expect to need more than a way to check the target platform. Execution constraints don't seem relevant to the code that was changed. Perhaps I'm mistaken?

python/private/common/py_executable.bzl
```diff
@@ -91,6 +92,17 @@ filename in `srcs`, `main` must be specified.
...
+def is_target_platform_windows(ctx):
+    """Determine if the target platform is a windows platform.
+
+    Args:
+        ctx (ctx): The rule's context object
+
+    Returns:
+        bool: True if the target platform is Windows.
+    """
+    return ctx.target_platform_has_constraint(Label("{}/os:windows".format(PLATFORMS_LOCATION)))
```
rickeylev, line 104: This works? I thought they had to be attr.label() dependencies.

I think this is incomplete, too. IIRC, there are multiple windows OS targets for the different variants (32, 64, etc). All of them should be listed and checked (just as done for the "is darwin" logic).

rickeylev (changes requested):

rickeylev:
btw, you can now enable the starlark rule implementation by setting `RULES_PYTHON_ENABLE_PYSTAR=1` and `USE_BAZEL_VERSION=last_green`.

python/private/common/py_executable.bzl
UebelAndre: Does passing CI not demonstrate this works? I made these changes mostly without running locally as I have no windows machine. Is this code path tested on all platforms in PR builds?

rickeylev: CI only runs Ubuntu with the Starlark rules, and only a subset of tests (none of the integration tests or examples). Windows + the Starlark impl is actually broken today; I ran into it while fixing another bug. Once https://github.com/bazelbuild/bazel/commit/899e928bf2af79a60e4c6fd2595dae2b90cb02ed is available in last_green (might be already by now), I can merge in #1443 and then cleanup #1452 (which removes the get_current_os_name() calls and also adds tests for Windows).

And yes, I can with certainty say that passing a label is incorrect: target_platfrom_has_constraint expects a ConstraintValueInfo object.

UebelAndre:
Is it possible to add CI for windows and macOS that regression tests the starlark rules?

rickeylev:
> Is it possible to add CI for windows and macOS that regression tests the starlark rules?

You know, I think we have exactly 2 slots left for CI jobs, so I think we can, actually.

PR #1452 (merged) switches to using platform detection, so I'll close this PR.

PR #1433, opened 2023-09-27
feat(bzlmod): mark pip extension as os/arch dependent

aignas:
This ensures that under bzlmod with `--lockfile-mode=update` we
would generate an entry per os/arch, which is needed because the
hermetic toolchain interpreter path is os/arch dependent.

Summary:
- add bazel_features dep
- mark the pip extension as arch/os dependent

Related: bazelbuild/bazel#19154

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,7 +37,7 @@ A brief description of the categories of changes:
   cases. An error about `@rules_python_internal` means the `py_repositories()`
   call is missing in `WORKSPACE`.
 
-* (bzlmod): the `pip.parse` extension will generate os/arch specific lock
+* (bzlmod) The `pip.parse` extension will generate os/arch specific lock
   file entries on `bazel>=6.4`.
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,9 @@ A brief description of the categories of changes:
   cases. An error about `@rules_python_internal` means the `py_repositories()`
   call is missing in `WORKSPACE`.
 
+* (bzlmod): the `pip.parse` extension will generate os/arch specific lock
+  file entries on `bazel>=6.4`.
+
 
 ### Added
 
--- a/python/extensions/pip.bzl
+++ b/python/extensions/pip.bzl
@@ -14,6 +14,7 @@
 
 "pip module extension for use with bzlmod"
 
+load("@bazel_features//:features.bzl", "bazel_features")
 load("@pythons_hub//:interpreters.bzl", "DEFAULT_PYTHON_VERSION", "INTERPRETER_LABELS")
 load(
     "//python/pip_install:pip_repository.bzl",
@@ -380,6 +381,17 @@ cannot have a child module that uses the same `hub_name`.
     }
     return attrs
 
+def _extension_extra_args():
+    args = {}
+
+    if bazel_features.external_deps.module_extension_has_os_arch_dependent:
+        args = args | {
+            "arch_dependent": True,
+            "os_dependent": True,
+        }
+
+    return args
+
 pip = module_extension(
     doc = """\
 This extension is used to make dependencies from pip available.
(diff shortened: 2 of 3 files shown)


python/extensions/pip.bzl
```diff
@@ -428,6 +429,18 @@ cannot have a child module that uses the same `hub_name`.
...
 
+def _extension_extra_args():
+    args = {}
+
+    if bazel_features.external_deps.module_extension_has_os_arch_dependent:
+        args = args | {
+            "arch_dependent": True,
+            "os_dependent": True,
+        }
+        fail("This does not work right now")
```
aignas, line 440: @fmeum, could you help me understand why the `bazel_features` detection does not work? I added a `fail` statement here, but
```
$ USE_BAZEL_VERSION=last_green bazel test --lockfile_mode=update ...
```

is completing successfully and the lock file does not have the new structure.

fmeum: Where did you run this? If I run it in `examples/bzlmod`, I get the expected failure. When I comment out the `fail`, I do get the expected lockfile structure.

aignas: It seems that I needed to do `bazel clean` in order to get it to work. I have seen similar behaviour in the past where the updated module code would not necessarily work as intended in the `rules_python` code when the external repository content changes.

CHANGELOG.md
```diff
@@ -37,6 +37,9 @@ A brief description of the categories of changes:
   cases. An error about `@rules_python_internal` means the `py_repositories()`
   call is missing in `WORKSPACE`.
 
+* (bzlmod): the `pip.parse` extension will generate os/arch specific lock
```
rickeylev, line 40:
```suggestion
* (bzlmod) The `pip.parse` extension will generate os/arch specific lock
```

For consistency (other entries don't use ":")

rickeylev (approved):

python/extensions/pip.bzl
fmeum: Could you try to provide instructions to reproduce this behavior? This would be a very concerning bug that we would very much like to fix before Bazel 7.

aignas: I can repro it on my local machine with:
```console
git checkout main
git log -n 3 --oneline
git checkout f2a4dd5
cd examples/bzlmod
USE_BAZEL_VERSION=latest bazel test --lockfile_mode=update ...
# press Ctrl-C to abbort requirements_test
git log -n 3 --oneline main
git checkout cdb5902
USE_BAZEL_VERSION=last_green bazel test --lockfile_mode=update ...
# press Ctrl-C to abbort requirements_test
```

aignas: And then later I can reproduce it with just:
```console
git checkout f2a4dd5
USE_BAZEL_VERSION=last_green bazel test --lockfile_mode=update ...
# press Ctrl-C to abort requirements_test
git checkout main
USE_BAZEL_VERSION=last_green bazel test --lockfile_mode=update ...
```

meteorcloudy:
Thanks for addressing this issue in rules_python. But I'm just wondering is there any specific reason why rules_python couldn't generate repos for all supported platforms and select the correct repo during build time? This is what the java toolchains do and could avoid using use_os/use_arch.

One problem I found is that there is a bit of a usability issue of use_os/use_arch: if we enforce `--lockfile_mode=error` on CI build, a user who updates rules_python has to update the lockfile on each supported platforms to pass check. This is sometimes hard because some users may not have access to specific machines.

aignas:
Hi, I agree that the ergonomics of this feature is not ideal because of this reason. The way rules_python is implemented today it needs to resolve the Python interpreter for the host platform and pass that to the repository rule that then uses that to generate the BUILD.bazel content and/or to build a python wheel. I think it would be great to write an overview doc and outline possible ways to move forward as there are many moving parts and getting rid of this platform dependence is not trivial.

Issue #1360, opened 2023-08-03
Pull rules_pycross into rules_python


rickeylev:
A recurring source of issues in rules_python is the underlying support for how wheels are handled, mainly that support for multiple platforms and versions doesn't work very well because a lot of activity happens at repo-rule time. Several of us have been looking at different parts of this problem, and we've generally come to the conclusion that a more idealized implementation is more akin to having simple http_file download's of the pypi artifacts, and moving any subsequent logic out of repo-rule time and into the regular build phase. This is a pretty big change in how things are done.

Luckily for us, @jvolkman's [rules_pycross](https://github.com/jvolkman/rules_pycross) has done much of this already, and he's said its OK for us to take that code.

[Per Google's rules](https://opensource.google/documentation/reference/releasing/preparing#third-party_components), there are two ways we can do this:
  1. jvolkman@ signs the Google CLA (done already, can check internal signcla page for jvolkman) and then sends a pull request contributing his code. This is treated like any other pull request, and rules_python then owns the code.
  2. Copy the rules_pycross code in a top-level third_party directory. Any existing LICENSE and copyright headers must be preserved. If we edit a file, we must modify the copyright header to include Google LLC.

(1) is preferred because it's just less overhead in the long run.

I suggest we ask Jeremy to simply do a one-time code-dump it into `//python/private/pycross_staging`, then we can start picking it apart as necessary. This avoids the extra overhead of license and copyright management from (2), and having to ask Jeremy to send a PR for some new piece of pycross we want to reuse.

After its ingested, we can rename things to drop the "pycross" prefixes, figure out appropriate APIs we wish to expose, how to integrate with our existing code, etc

* [ ] jvolkman@ sends PR contributing code
* [ ] Rename to remove "pycross" prefixes
* [ ] Figure out public API surface
* [ ] Integrate with existing workspace/bzlmod build styles

chrislovecnm:
Who is working on this?

groodt:
Option 1 LGTM if LGTY @jvolkman

jvolkman:
I assume it's me working on this, but we're camping this week so I won't be able to get to it until after.

martis42:
@rickeylev 

Since the robots are gonna close this soonish, are we still aiming to do this?
While I can't contribute on this, I would be interested in this being available in rules_python eventually.

groodt:
Code was vendored here https://github.com/bazelbuild/rules_python/tree/main/third_party/rules_pycross

[...]

PR #1436, opened 2023-09-28
pycross: Add patching support to py_wheel_library

philsc:
This patch adds a few arguments to `py_wheel_library` to simulate how
`http_archive` accepts patch-related arguments.

I also amended the existing test to validate the behaviour at a very
high level.

References: #1360


third_party/rules_pycross/pycross/private/tools/wheel_installer.py
```diff
@@ -97,6 +98,24 @@ def main(args: Any) -> None:
 
     setup_namespace_pkg_compatibility(lib_dir)
```
aignas, line 99: I would have expected patching to happen before this line.

third_party/rules_pycross/pycross/private/wheel_library.bzl
```diff
@@ -119,6 +123,28 @@ and py_test targets must specify either `legacy_create_init=False` or the global
...
+        "patches": attr.label_list(
+            allow_files = True,
+            default = [],
+            doc =
+                "A list of files that are to be applied as patches after " +
+                "extracting the archive. This will use the patch command line tool.",
+        ),
         "python_version": attr.string(
             doc = "The python version required for this wheel ('PY2' or 'PY3')",
             values = ["PY2", "PY3", ""],
```
aignas, line 150: nit: we should probably remove support for `PY2`, which is probably outside the scope of this PR.

third_party/rules_pycross/pycross/private/wheel_library.bzl
```diff
@@ -119,6 +123,28 @@ and py_test targets must specify either `legacy_create_init=False` or the global
...
+            doc =
+                "The arguments given to the patch tool. Defaults to -p0, " +
+                "however -p1 will usually be needed for patches generated by " +
+                "git. If multiple -p arguments are specified, the last one will take effect.",
+        ),
+        "patch_dir": attr.string(
+            default = "",
+            doc = "The directory from which to invoke the patch_tool.",
+        ),
+        "patch_tool": attr.string(
```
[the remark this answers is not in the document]
aignas: I am wondering if we could reuse the same repository util that I have written in #1393.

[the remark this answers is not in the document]
philsc: So I'm trying out GNU patch! I have no idea if it's worth it or not. Please ignore updates on this PR until I report back.

It's possible that it's too much of a hassle to get this working properly.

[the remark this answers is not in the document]
philsc: I've given up on this and just used a wrapper script to skip patching on Windows. At least for the test. When used in production, the user has to provide a patcher. Could be a simple wrapper script.

third_party/rules_pycross/pycross/private/tools/wheel_installer.py
philsc: Hmm. In case the user wants to set up their own `__init__.py` file or something? My thought was to let the user patch as much of the internal code gen as possible for the most flexibility.

third_party/rules_pycross/pycross/private/wheel_library.bzl
philsc: Yeah, I can submit a follow-up PR.

third_party/rules_pycross/pycross/private/wheel_library.bzl
```diff
@@ -119,6 +129,34 @@ and py_test targets must specify either `legacy_create_init=False` or the global
...
+            default = ["-p0"],
+            doc =
+                "The arguments given to the patch tool. Defaults to -p0, " +
+                "however -p1 will usually be needed for patches generated by " +
+                "git. If multiple -p arguments are specified, the last one will take effect.",
+        ),
+        "patch_dir": attr.string(
+            default = "",
+            doc = "The directory from which to invoke the patch_tool.",
+        ),
```
aignas, line 142: Since the user is supplying patches and they can supply `patch_args`, is this `patch_dir` necessary?

third_party/rules_pycross/pycross/private/wheel_library.bzl
```diff
@@ -119,6 +129,34 @@ and py_test targets must specify either `legacy_create_init=False` or the global
...
+        "patch_tool": attr.string(
+            doc = "The patch(1) utility from the host to use. " +
+                  "If set, overrides `patch_tool_target`.",
+        ),
+        "patch_tool_target": attr.label(
+            executable = True,
+            cfg = "exec",
+            doc = "The label of the patch(1) utility to use. " +
+                  "Only used if `patch_tool` is not set.",
```
aignas, line 151: Do we restrict users to only use `patch(1)` tool?

third_party/rules_pycross/pycross/private/wheel_library.bzl
```diff
@@ -119,6 +129,34 @@ and py_test targets must specify either `legacy_create_init=False` or the global
...
+                "The arguments given to the patch tool. Defaults to -p0, " +
+                "however -p1 will usually be needed for patches generated by " +
+                "git. If multiple -p arguments are specified, the last one will take effect.",
+        ),
+        "patch_dir": attr.string(
+            default = "",
+            doc = "The directory from which to invoke the patch_tool.",
+        ),
+        "patch_tool": attr.string(
+            doc = "The patch(1) utility from the host to use. " +
```
aignas, line 144: We probably should print a warning or add a warning in the docs saying that setting `patch_tool` will mean that builds are not completely hermetic.

I think I am fine with them being that way, but we should do a good job to communicate to the users that.

aignas (commented):
Thanks for the update!

third_party/rules_pycross/pycross/private/tools/wheel_installer.py
```diff
@@ -97,6 +98,30 @@ def main(args: Any) -> None:
...
+    if args.patch:
+        if not args.patch_tool and not args.patch_tool_target:
+            raise ValueError("Specify one of 'patch_tool' or 'patch_tool_target'.")
+
+        patch_args = [
+            args.patch_tool or Path.cwd() / args.patch_tool_target
+        ] + args.patch_arg
+        patch_dir = args.patch_dir or "."
+        for patch in args.patch:
+            with patch.open("r") as stdin:
```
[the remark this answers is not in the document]
philsc: I cannot comment on whether it's universal, but at least GNU patch (and presumably BSD patch) is pretty much hard-coded to expect the patch on standard input. It's actually kind of annoying to make `patch` accept the patch as an argument instead of via stdin.

Unless I'm missing something simple? At least it appears not to be as simple as specifying the patch as an argument.

third_party/rules_pycross/pycross/private/wheel_library.bzl
```diff
@@ -33,19 +33,29 @@ def _py_wheel_library_impl(ctx):
...
         args.add("--wheel-name-file", name_file)
 
+    if ctx.attr.patch_tool_target:
+        args.add("--patch-tool-target", ctx.attr.patch_tool_target.files_to_run.executable)
+        tools.append(ctx.executable.patch_tool_target)
+
     if ctx.attr.enable_implicit_namespace_pkgs:
         args.add("--enable-implicit-namespace-pkgs")
 
     ctx.actions.run(
```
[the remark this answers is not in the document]
philsc: The advantage is that you only have one copy of the extracted wheel lying around/cached etc.  If you split it into two actions, then you end up with ~2x the disk/cache usage.   That's really the only reason. Some wheels are huge.

third_party/rules_pycross/pycross/private/wheel_library.bzl
philsc: It is something that I found myself wanting several times when using `http_archive`'s patching functionality. I added it here because I figured it may come up again or someone else might want to use it.

third_party/rules_pycross/pycross/private/wheel_library.bzl
philsc: No. This is simply copied from the `http_archive` implementation.  I'm happy to change the wording, but keeping it similar to `http_archive` seemed reasonable at the time at least. I think part of the reason for writing it as `patch(1)` explicitly is to imply the API (i.e. patch on stdin).
WDYT?

https://github.com/bazelbuild/bazel/blob/0226327c0ac03d818c211a718494c7c7cdc9bafc/tools/build_defs/repo/http.bzl#L327-L330

third_party/rules_pycross/pycross/private/wheel_library.bzl
philsc: Done

third_party/rules_pycross/pycross/private/tools/wheel_installer.py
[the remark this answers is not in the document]
aignas: Thanks for explaining, LGTM, I am dealing with patching on a daily basis.

third_party/rules_pycross/pycross/private/wheel_library.bzl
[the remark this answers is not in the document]
aignas: Ah yes, I've seen some wheels being about 800MB unextracted (I am looking at you, `pytorch`). I think explaining this reason in a comment may help maintainers understand why we have a single action.

third_party/rules_pycross/pycross/private/wheel_library.bzl
aignas: I am a fan of extending things when needed, so if we don't have a usecase yet, we should not add it as it increases the API surface we need to maintain. What is more, `http_archive` patching is very limited, so the experience from there might not exactly transfer.

third_party/rules_pycross/pycross/private/wheel_library.bzl
```diff
@@ -135,3 +174,17 @@ This option is required to support some packages which cannot handle the convers
...
     },
 )
+
+def py_wheel_library(patches = None, patch_tool = None, patch_tool_target = None, **kwargs):
+    # If the user has specified patches, but hasn't specified how to apply
+    # those patches, default to our internally-provided patch binary.
+    if patches and not patch_tool and not patch_tool_target:
+        fail("Patches specified, but missing `patch_tool` or `patch_tool_target` attributes. " +
+             "I.e. unsure how to apply patches.")
+
```
aignas, line 184: Thinking out loud, we set it here so that we get the error at load time?

I think just doing it at the top of the `_py_wheel_library` may be just as good. That way we can avoid the macro. On the other hand, if this is idiomatic and something that is commonly done in the bazel world, I am OK with leaving it as is.

aignas (commented):
I am not sure that having patching as part of the `build` action is the best way. Patches in general can modify the `METADATA` that is used to generate dependencies and as you mentioned, if we split patching into a separate action, then we may end up with double the space in the build action cache.

If we introduce patching at this layer, then we will have to have a different type of patching for the dependencies. What is more, making this hermetic requires the user to go through extra hoops and just leaving it in the repository rule world and non-hermetic by construction segregates the responsibilities slightly better, IMHO.

That said, let's discuss as I am happy to be convinced that my view is wrong. :)

third_party/rules_pycross/pycross/private/tools/wheel_installer.py
```diff
@@ -97,6 +98,30 @@ def main(args: Any) -> None:
 
     setup_namespace_pkg_compatibility(lib_dir)
 
+    if args.patch:
+        if not args.patch_tool and not args.patch_tool_target:
+            raise ValueError("Specify one of 'patch_tool' or 'patch_tool_target'.")
```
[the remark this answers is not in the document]
philsc: Done

third_party/rules_pycross/pycross/private/wheel_library.bzl
philsc: Removed.

third_party/rules_pycross/pycross/private/wheel_library.bzl
philsc: Yeah, putting the error at load time makes it more likely for the user to notice before uploading for code review. But I don't think it's _really_ buying us anything. The tool itself will fail during the action phase.

philsc:
> I am not sure that having patching as part of the `build` action is the best way. Patches in general can modify the `METADATA` that is used to generate dependencies and as you mentioned, if we split patching into a separate action, then we may end up with double the space in the build action cache.
> 
> If we introduce patching at this layer, then we will have to have a different type of patching for the dependencies. 

You mean patching what the dependencies themselves are? If so, you are correct.  That can be done by modifying the intermediate file itself. The same place where you introduce what patches to apply.
https://github.com/philsc/rules_python/blob/unreviewed/phil/pypi_install/examples/pypi_install/intermediate_file_patcher.py

> What is more, making this hermetic requires the user to go through extra hoops and just leaving it in the repository rule world and non-hermetic by construction segregates the responsibilities slightly better, IMHO.

That does sound nice, but that's not really what my goal is. My goal is to cut the logic in repository rules to an absolute minimum. Everything else should go into actions.  I.e. I really just want to download the wheel, set up some minimal BUILD files that specify the dependencies and such, and leave everything else to actions.

Having a non-hermetic patch tool in an action is really no worse than having a non-hermetic patch tool in repository context. Making the patching hermetic is easier in action context because you can depend on the output of other actions (e.g. a bazel-compiled `patch` binary).

aignas (approved):
> > I am not sure that having patching as part of the `build` action is the best way. Patches in general can modify the `METADATA` that is used to generate dependencies and as you mentioned, if we split patching into a separate action, then we may end up with double the space in the build action cache.
> > If we introduce patching at this layer, then we will have to have a different type of patching for the dependencies.
> 
> You mean patching what the dependencies themselves are? If so, you are correct. That can be done by modifying the intermediate file itself. The same place where you introduce what patches to apply. https://github.com/philsc/rules_python/blob/unreviewed/phil/pypi_install/examples/pypi_install/intermediate_file_patcher.py

Thanks for the link, this will give me a little bit of more context.

> > What is more, making this hermetic requires the user to go through extra hoops and just leaving it in the repository rule world and non-hermetic by construction segregates the responsibilities slightly better, IMHO.
> 
> That does sound nice, but that's not really what my goal is. My goal is to cut the logic in repository rules to an absolute minimum. Everything else should go into actions. I.e. I really just want to download the wheel, set up some minimal BUILD files that specify the dependencies and such, and leave everything else to actions.
>
> Having a non-hermetic patch tool in an action is really no worse than having a non-hermetic patch tool in repository context. Making the patching hermetic is easier in action context because you can depend on the output of other actions (e.g. a bazel-compiled `patch` binary).

Having non-hermetic tools within the action leads to the results of that action go into the action cache, which may make the cached entries harder to trust. That said, we can mark the actions with tags like `no-cache` to ensure that they don't get cached, whereas in the `repository_ctx` we don't have this choice.

Anyway, I am pro having this merged and discussing more as we go.

PR #1440, opened 2023-09-29
tests: Move base rule tests under tests instead of //tools/build_defs/python

rickeylev:
The tools/build_defs/python path is an artifact of some Google-internal
naming that isn't applicable to the rules_python source tree layout.
With the Starlark implementation running in CI and working in some
capacity, it's time to move them to a less mysterious location.

* Also moves the py_wheel tests out of the base rule tests. Not sure
  how they ended up there.

Work towards #1069


groodt (approved):
LGTM

Oh thank you for starting to clean up the repo structure. It has been doing my head in!

PR #1441, opened 2023-09-30
tests(pystar): py_runtime_pair and py_runtime analysis tests

rickeylev:
These analysis tests verify that `py_runtime` and `py_runtime_pair` are working as intended
for both the native Bazel and Starlark implementations.

Work towards #1069

### DIFF
--- a/tests/py_runtime/py_runtime_tests.bzl
+++ b/tests/py_runtime/py_runtime_tests.bzl
@@ -1,0 +1,262 @@
+# Copyright 2023 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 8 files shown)


tests/py_runtime/py_runtime_tests.bzl
```diff
@@ -0,0 +1,262 @@
...
+def _test_bootstrap_template_impl(env, target):
+    env.expect.that_target(target).provider(
+        PyRuntimeInfo,
+        factory = py_runtime_info_subject,
+    ).bootstrap_template().path().contains("bootstrap.txt")
+
+_tests.append(_test_bootstrap_template)
+
+def _test_cannot_have_both_inbuild_and_system_interpreter(name):
+    if br_util.is_bazel_6_or_higher():
```
aignas, line 65: nit: Should we extend https://github.com/bazel-contrib/bazel_features/blob/main/features.bzl and use that instead of `br_util` in the long run?

aignas (approved):

tests/py_runtime/py_runtime_tests.bzl
rickeylev: I'm not sure. bazel_features seems to be based on version numbers, but, iirc, our internal Bazel builds don't have a version number; it looks like rolling does, though, at least.

PR #1452, opened 2023-10-03
tests(pystar): add analysis tests to cover basic windows building

rickeylev:
The CI currently only runs on Ubuntu, so it assumes that is the target platform. This ends
up missing some notable Windows code paths, though.

Since its analysis-phase logic, we can force the platform to be Windows for the analysis
tests, and then the rules follow the code paths that should be taken under Windows.
This allows testing Windows logic under Ubuntu.


Solumin (approved):

UebelAndre:
For the paper trail https://github.com/bazelbuild/bazel/issues/19587 relates to this PR.

PR #1453, opened 2023-10-03
fix(py_wheel): produce deterministic wheel files

dizzy57:
Current implementation does not produce deterministic output because:
- `ZipFile.writestr()` leaks current date and time
- `ZipFile.write()` leaks the source file's mtime and mode bits (permissions) into the resulting zip archive.

By manually creating our own `ZipInfo` objects we can explicitly set date and time fields to `Jan 1, 1980, 00:00` (minimum value allowed by the zip file standard), and ensure that other file attributes are uniform across all entries in a zip file.

### DIFF
--- a/tools/wheelmaker.py
+++ b/tools/wheelmaker.py
@@ -246,6 +246,8 @@ class WheelMaker(object):
         arcname = filename.lstrip(separators)
 
         zinfo = zipfile.ZipInfo(filename=arcname, date_time=_ZIP_EPOCH)
+        zinfo.create_system = 3  # ZipInfo entry created on a unix-y system
+        zinfo.external_attr = 0o777 << 16  # permissions: rwxrwxrwx
         zinfo.compress_type = self._zipfile.compression
         return zinfo
 
--- a/tools/wheelmaker.py
+++ b/tools/wheelmaker.py
@@ -14,13 +14,14 @@
 
 import argparse
 import base64
-import collections
 import hashlib
 import os
 import re
 import sys
 import zipfile
 from pathlib import Path
+
+_ZIP_EPOCH = (1980, 1, 1, 0, 0, 0)
 
 
 def commonpath(path1, path2):
@@ -189,7 +190,8 @@ class WheelMaker(object):
         """Add given 'contents' as filename to the distribution."""
         if sys.version_info[0] > 2 and isinstance(contents, str):
             contents = contents.encode("utf-8", "surrogateescape")
-        self._zipfile.writestr(filename, contents)
+        zinfo = self._zipinfo(filename)
+        self._zipfile.writestr(zinfo, contents)
         hash = hashlib.sha256()
         hash.update(contents)
         self._add_to_record(filename, self._serialize_digest(hash), len(contents))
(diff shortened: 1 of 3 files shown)


rickeylev:
I'm not a wheel expert, but dropping the permissions might be problematic? My basic thinking is if a +x bit gets dropped, that might be a problem. Can't wheel contains arbitrary data, and someone might intend for a data file to have the executable bit?

Maybe mask such that we keep the executable bit if set, but normalize to read only?

tools/wheelmaker.py
```diff
@@ -139,20 +141,34 @@ def arcname_from(name):
...
+    def _zipinfo(self, filename):
+        """Construct deterministic ZipInfo entry for a file named filename"""
+        # Strip leading path separators to mirror ZipInfo.from_file behavior
+        separators = os.path.sep
+        if os.path.altsep is not None:
+            separators += os.path.altsep
+        arcname = filename.lstrip(separators)
+
+        zinfo = zipfile.ZipInfo(filename=arcname, date_time=_ZIP_EPOCH)
+        zinfo.compress_type = self._zipfile.compression
```
philsc, line 169: Can we maybe put something like
```python
info.external_attr = 0777 << 16
```
here? Just to make sure executable things are actually executable. I guess we don't have a test for this?
https://stackoverflow.com/a/434689

jvolkman:
If it helps, I've got code that canonicalizes wheels [here](https://github.com/jvolkman/repairwheel/blob/main/src/repairwheel/wheel.py#L78).

Another thing to watch out for is `ZipInfo.create_system` which differs based on whether you're running on windows or linux/unix.

rickeylev:
I talked to a Bazel dev about how permissions are handled

1. Bazel doesn't make any guarantees about file permissions
1. It ignores permissions when considering if the file changed
1. Action outputs are set to rxrxrx
1. Source files don't have permission changed
1. They pointed out symlinks might be weird (no one was sure off-hand how symlinks + permissions would behave)

Pretty much any of those are a good reason for us to not preserve permissions. I find the second and third one particularly compelling.

So I think its really just a question of: do we set rx or rwx?

My vote is for rwx; it seems what is commonly done, and it "preserves" the writability of any files.

aignas:
Whilst playing with the code, I have realized that the proposed solution does not fully solve determinism within wheels. Depending on the order that the files get passed to the `wheelmaker` we may end up with a different `zip` from time to time.

nit: I think we should follow the convention of sorting the `RECORD` file entries in a particular way - first all of the files, then all of the dist-info files and then the RECORD file last in order to ensure that it is somewhat consistent with what `repairwheel` is doing.

rickeylev:
(Ignas posted this, but it only shows up in my email not on the pr for some reason)

> Whilst playing with the code, I have realized that the proposed solution does not fully solve determinism within wheels. Depending on the order that the files get passed to the wheelmaker we may end up with a different zip from time to time.
>
> nit: I think we should follow the convention of sorting the RECORD file entries in a particular way - first all of the files, then all of the dist-info files and then the RECORD file last in order to ensure that it is somewhat consistent with what repairwheel is doing.

Is the order coming from depsets? Or is it coming from e.g. a Python dict/set re-ordering things?

If it's coming from a depset, then, insofar as Bazel is concerned, this is fine. Depsets (and input lists) have ordering and it's reflected in how bazel computes what is affected/changed (e.g. the argv of the action's command line). I don't think that can be avoided. The zip itself is still deterministic, just not "predictable".

If it's coming from Python doing a set/dict call somewhere, then we should fix that -- that's a source of non-determinism Bazel isn't aware of. Having that in a separate PR is fine; this one has been out for awhile.

But yes, I agree that sorting the zip entries is a good idea regardless.

dizzy57:
> Is the order coming from depsets? Or is it coming from e.g. a Python dict/set re-ordering things?

We're sorting file lists in `main()` [here](https://github.com/bazelbuild/rules_python/blob/382b6785a57ee428fc0ec367bcb380c6266cab7b/tools/wheelmaker.py#L442) and [here](https://github.com/bazelbuild/rules_python/blob/382b6785a57ee428fc0ec367bcb380c6266cab7b/tools/wheelmaker.py#L516).
The files end up getting added to the zip archive in deterministic order: 
 - sorted files from `all_files`
 - `WHEEL`
 - `METADATA`
 - optional `entry_points.txt`
 - sorted files from `extra_distinfo_file`
 - `RECORD`

aignas:
Yeah, my previous comment (that I have deleted) was all wrong, I think all is good with the sorting of the contents within the wheelmaker, so the only things that I would love to see in this PR would be:
* A test asserting that the sha256 of the wheel remains the same.
* The `rwx` permission setting as @philsc has suggested.

tools/wheelmaker.py
jvolkman, line 169: I'd suggest `zinfo.create_system = 3` here to work around [this logic](https://github.com/python/cpython/blob/main/Lib/zipfile/__init__.py#L418-L422) in zipfile.py.

aignas:
Actually now I realized what had confused me in the beginning - when writing the `RECORD` file we change the order of the entries in [this line](https://github.com/bazelbuild/rules_python/pull/1453/files#diff-b29be3845ac48e11fd0c22ac678e780fb376ae59cc4e894bf38cdcada2dea59bL186) and given that the order of files going into the zip file is deterministic and the order of the slice is also deterministic, I would vote for removing the `entries.sort()` on that line because it may be unnecessary. I am not sure if it is harmless. Maybe someone else with more knowledge into how things should be can comment.

groodt (approved):
LGTM

dizzy57:
Rebased on top of main.

aignas (approved):
LGTM, thanks

Issue #1458, opened 2023-10-04
Make //docs project private visibility


rickeylev:
The //docs package currently defaults to public visibility, however, the targets within it are just for rules_python to generate its own docs. Other projects shouldn't be relying on the details of how we generate our documentation (especially as I'd like us to move to a generated doc solution like Sphinx so we can have Real Documentation).

From what I can tell, the docs targets are only public so that other projects are able to get the necessary bzl sources for the rules_python bzl code (e.g. `//python:defs.bzl`) (usually for their own documentation generation), and various filegroups and bzl_libraries were defined in //docs to avoid a dependency on skylib.

We depend on skylib profusely now, so we don't need to worry about that concern.

But, the bzl_library targets in //docs were public targets for a long time, and a Slack thread indicates a couple projects might be pointing to them, so I don't want to just yank them out. I'll put some aliases to make it a bit easier to handle an upgrade.

groodt:
Yes, this will be great. I found the docs setup quite awkward.

PR #1459, opened 2023-10-04
docs: move dependency management into respective bzl packages

rickeylev:
This moves the dependency management of what transitive files are needed to generate docs for a .bzl file out of the docs directory and into the respective bzl file's directory. This ensures that the bzl_library targets we expose to users contain all the necessary dependencies.

Because there are some projects using the bzl_library targets in //docs, some compatiblity aliases are added to make their migration path easier. Those targets only public for historical reasons and shouldn't be used.

Work towards #1458

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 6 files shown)


groodt (approved):
LGTM

Issue #1361, opened 2023-08-03
Release version 1.0


rickeylev:
This issue is for tracking the effort to release a version 1.0 of rules_python.

There are several reasons for wanting to have a major version release:
  * Better communicates when breaking changes are made
  * Easier support for multiple versions and backporting of functionality
  * Major versions are an indicator of fitness for use.

Anyways, there's a few large items to complete before a 1.0 release can be considered:

* [x] Implementation is all in rules_python Starlark, not Bazel itself.
* [x] Written policies that cover breaking changes, backwards compatibility, support for older rules_python versions, and supported Bazel versions
  * Bazel rules are rather leaky: it's easy for implementation details to unintentionally be
    visible and accessible to users, even if weren't intended so. This is especially true for
    rules_python, which carries various legacy and historical behaviors.
  * Requiring e.g. the latest patch or minor version release of Bazel allows us to take advantage
    of newer functionality and reduce the size of the test/support matrix
  * Steps necessary to introduce a breaking change (e.g. docs about how to migrate, issues to file, etc).
* [x] Human friendly changelog, i.e. CHANGELOG.md in keepachangelog.org-style format.
* [x] API review: reduce our API footprint to what makes sense (there's a variety of things exposed publicly that probably shouldn't be public)

EDIT: additional scope

- [ ] ~Rename `pip_parse` to something else.~
- [x] Mark `gazelle` as experimental or splitting the versioning scheme so that it is clear that it does not have similar semver guarantees.
- [x] Mark `pip.parse` APIs as experimental and to stabilize later.
- [x] We should have a category of public but unstable API. Certain things should be called out: gazelle, sphinxdocs, toolchain implementation details, requirements locking?
- [ ] Should we explicitly say that we would implement `bzlmod` first and `WORKSPACE` is something that needs to be supported until bazel 9 is the lowest supported version. It's OK to have new features in `bzlmod` only, but `WORKSPACE` has to be present. PRs would be welcome for `WORKSPACE` additions.
- [x] Allow overriding `TOOL_VERSIONS` for the `python` bzlmod extension to have API parity.

chrislovecnm:
Do we want https://github.com/bazelbuild/rules_python/issues/1360 as well?

This is a big change to have after a 1.0 release. It is more a 1.1.

rickeylev:
> Do we want #1360 as well?

[...]

PR #1460, opened 2023-10-05
docs: document a breaking changes policy

### LINKED ISSUE #1424: Breaking changes policy
As part of establishing a 1.0 release (https://github.com/bazelbuild/rules_python/issues/1361), I want to write down our policy and process for breaking changes.

The basic policy and procedure I'm leaning towards is something like:

> In order to introduce a breaking change, the breaking behavior must (when reasonably possible)
> be opt-in in the first release (`N`) it's available for.

The basic idea is: it's easier to handle breaking changes when you can incrementally upgrade a code base. e.g., upgrade the dependency version N, update apis to the new version's requirements, then upgrade the dependency to N+1.

I'm not keen on a date-based scheme (e.g. an incompatible change must retain a compatible option for 6 months); I'd rather just use version releases.

I'm torn about what happens after that, though. I see two approaches:
* The next release forces the new behavior on; the toggle is a no-op or removed. This has the advantage of being easier on us maintainers; defunct code sticks around for less longer.
* The next release enables the new behavior by default, but users are able to opt-out. The release after that one forces the new behavior on. This has the advantage of being easier for users (you can opt-out problem cases reactively to more clearly identify what needs to be fixed). But, I suspect most people won't notice an incompatible change until it's the default behavior, so I'm not sure how much value the extra step really has. This is a bit more of a burden on us maintainers, though.

rickeylev:
This largely covers how to introduce a breaking change.

It also attempts to clarify what is or isn't a breaking change.

Closes #1424
Work towards #1361

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -4,6 +4,7 @@ PR Instructions/requirements
   * Update `CHANGELOG.md` as applicable
 * Breaking changes include "!" after the type and a "BREAKING CHANGES:"
   section at the bottom.
+  See CONTRIBUTING.md for our breaking changes process.
 * Body text describes:
   * Why this change is being made, briefly.
   * Before and after behavior, as applicable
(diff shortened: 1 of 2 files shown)


aignas (approved):

rickeylev:
Updated pull request template, per Ignas's request on slack

aignas (approved):

PR #1473, opened 2023-10-07
fix(repo setup): Skip aliases for unloaded toolchains

### LINKED ISSUE #1472: Missing dependencies when not using latest Python version
# 🐞 bug report

### Affected Rule

 The issue is caused by the function:
`python_register_toolchains` 

### Is this a regression?

 Yes. The root cause has existed for a long time but this got more problematic in version 0.25.0 with the introduction of platforms 
`ppc64le-unknown-linux-gnu` and `s390x-unknown-linux-gnu`.

### Description

When using a WORKSPACE file and not using the latest Python versions, then there are dangling repo references to the new above-mentioned platforms. This leads to errors when doing Bazel queries on reverse dependencies.

## 🔬 Minimal Reproduction

Create a Bazel workspace with this WORKSPACE file:
```
VERSION = "0.26.0"
http_archive(
    name = "rules_python",
    strip_prefix = "rules_python-{}".format(VERSION),
    url = "https://github.com/bazelbuild/rules_python/releases/download/{}/rules_python-{}.tar.gz".format(VERSION, VERSION),
)
py_repositories()
python_register_toolchains(
    name = "python_3_8",
    python_version = "3.8",                                                                                                                                                   
)
```
then run the command
```
bazel query 'rdeps(@python_3_8//...,@python_3_8//:python3)'
```

## 🔥 Exception or Error

<pre><code>
ERROR: Evaluation of query "rdeps(@python_3_8//..., @python_3_8//:python3)" failed: preloading transitive closure failed: no such package '@python_3_8_ppc64le-unknown-linux-gnu//': The repository '@python_3_8_ppc64le-unknown-linux-gnu' could not be resolved: Repository '@python_3_8_ppc64le-unknown-linux-gnu' is not defined
</code></pre>

In a real-world workspace making a lot of use of Python, this error also shows on much more relevant and less obvious queries:
```
bazel cquery 'rdeps(//..., //<any_target_in_a_workspace_with_python_dependencies>)'
``` 

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
Ubuntu Linux 20.04
  </code>
</pre>

<pre>
  <code>
bazel 6.3.2
  </code>
</pre>

**Rules_python version:**

<pre>
  <code>
0.26.0
  </code>
</pre>
The issue was introduced in 0.25.0 when adding two new platforms.

**Anything else relevant?**
The root cause of this seems to be the fact that loading a Python interpreter for some platforms and version combinations is skipped, but the aliases are always created.

kilian-funk:
Some platforms don't contain every version, e.g. s390x doesn't have 3.8, which is
indicated by a missing sha256 value. When this happens, no repository for the runtime
is created (`python_repository` rule).

Similar logic needs to be in the toolchains setup logic because otherwise a reference
to an undefined repository exists in the select() expression of the aliases. Because those
references are lazily evaluated, they don't always cause a problem, but do mean that
query operations (e.g., `rdeps()`) can't work and the order of entries is important
(which is surprising).

Closes #1472


python/private/toolchains_repo.bzl
```diff
@@ -195,6 +196,9 @@ load("{rules_python}//python:pip.bzl", _compile_pip_requirements = "compile_pip_
 
 host_platform = "{host_platform}"
 interpreter = "@{py_repository}_{host_platform}//:{python3_binary_path}"
+PLATFORMS = [
```
rickeylev, line 199: Please take this out of defs.bzl -- defs.bzl is a public entry point, and this isn't a detail we want to make part of our public API.

Instead, it can be inlined in build_contents directly:
```
_plaforms = [
  {loaded_platforms}
]
```

CHANGELOG.md
```diff
@@ -19,6 +19,13 @@ A brief description of the categories of changes:
 
 ## Unreleased
 
+### Fixed
+
+* Skip aliases for unloaded toolchains. Toolchains without a checksum are skipped
+  in the repo load phase. These skipped repos now don't have aliases in the toolchains repo.
+  This is relevant when working with older python versions that don't have full platform 
+  support.
```
rickeylev, line 27:
```suggestion
* Skip aliases for unloaded toolchains. Some Python versions that don't have full
  platform support, and referencing their undefined repositories can break operations
  like `bazel query rdeps(...)`.
```

rickeylev (changes requested):
mostly lgtm, just move the platforms constant out of the generated defs.bzl

rickeylev:
CI is unhappy, but it's not immediately apparent why. It gives me the feeling there might be a syntax error somewhere? Or maybe what's happening is the BUILD file is trying to load defs.bzl before something defs.bzl needs is ready? Maybe removing that `load(defs.bzl)` as I commented will fix it.

python/private/toolchains_repo.bzl
kilian-funk: IMO the selection of platforms is valuable information, but I am not sure if this is the right form to make that available. So, I am fine with removing it.

CHANGELOG.md
kilian-funk: Done.

python/private/toolchains_repo.bzl
kilian-funk: Changed as suggested.

kilian-funk:
@rickeylev I have addressed your concerns. CI is happy now, even without knowing the root cause of the original failures.

Could you please re-review?

aignas (approved):
LGTM

Issue #1469, opened 2023-10-06
Migrate to Bzlmod for managing external dependencies


meteorcloudy:
With future Bazel LTS versions, Bzlmod will become the default system for external dependencies management.

Please consider migrating all your external dependencies in WORKSPACE to MODULE.bazel

For more information, please check https://github.com/bazelbuild/bazel/issues/18958

meteorcloudy:
Some tests are still failing with Bzlmod enabled:
https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/3352#018af628-22bd-468d-8813-0356ebe78c86

rickeylev:
posterity: From a cursory look, I get the impression the cause is some tests aren't going through the runfiles apis, so end up not being aware of repo-remapping.

rickeylev:
PR #1482 out to fix this. It's as I thought -- the runfiles directory changing to `_main` broke various assumptions.

rickeylev:
1482 is merged, so this should be fixed now

PR #1482, opened 2023-10-10
test(bzlmod): Make some tests bzlmod compatible with Bazel@HEAD

rickeylev:
A few tests weren't compatible with bzlmod, so would fail when it was enabled. The various causes and fixes are:

* Under bzlmod, `runfiles.CurrentRepository()` returns the empty string for the main repository. To fix, an environment variable is used to tell the test whether bzlmod is enabled or not.

* Accessing data files through `TEST_SRCDIR` directly is error-prone under bzlmod because the directory name within runfiles changes from the workspace name to `_main`. To fix, use the runfiles libraries, which know how to map apparent repo names to the actual directory name. In the integration tests, the runfiles library isn't available, so just check for the `_main` directory instead.

Work towards #1469

### DIFF
--- a/examples/wheel/wheel_test.py
+++ b/examples/wheel/wheel_test.py
@@ -29,9 +29,16 @@ class WheelTest(unittest.TestCase):
         self.runfiles = runfiles.Create()
 
     def _get_path(self, filename):
-        return self.runfiles.Rlocation(
-            os.path.join("rules_python/examples/wheel", filename)
-        )
+        runfiles_path = os.path.join("rules_python/examples/wheel", filename)
+        path = self.runfiles.Rlocation(runfiles_path)
+        if not path:
+            raise AssertionError(f"Runfiles failed to find {runfiles_path}")
+        elif not os.path.exists(path):
+            raise AssertionError(
+                f"Path {path} does not exist (from runfiles " + f"path {runfiles_path}"
+            )
+        else:
+            return path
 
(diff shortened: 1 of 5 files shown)


examples/wheel/wheel_test.py
```diff
@@ -18,18 +18,23 @@
...
-            "wheel",
-            "example_minimal_library-0.0.1-py3-none-any.whl",
+    def setUp(self):
+        super().setUp()
+        self.runfiles = runfiles.Create()
+
+    def _get_path(self, filename):
+        return self.runfiles.Rlocation(
+            os.path.join("rules_python/examples/wheel", filename)
         )
```
aignas, line 34: nit: we could improve this by passing `pathlib.Path` around.

```suggestion
        path = pathlib.Path(
            self.runfiles.Rlocation(
                os.path.join("rules_python/examples/wheel", filename)
            )
        )
        self.assertTrue(path.exists(), f"The path for {filename} does not exist")
        return path
        
```

aignas (approved):

examples/wheel/wheel_test.py
rickeylev: I tried this, but then realized the path is just immediately passed into ZipFile, so any of the additional functionality of pathlib isn't really used. I did update this to check for existence, though, which should help hint any any file not found errors in the future.

PR #1485, opened 2023-10-11
build(gazelle): embed Python zip file

### LINKED ISSUE #1455: Unable to compile standalone gazelle binary with rules_python plugin
Context:
I work within a repo where we require building a static gazelle binary per platform that is checked into git LFS. The precompiled binaries are then used for pre-commit hooks and can be run in CI environments without the need to build from source with bazel.

Issue:
I'm having issues compiling gazelle with the rules_python package because it makes use of runfiles.

The runfiles however is not portable and short of patching the [parser.go](https://github.com/bazelbuild/rules_python/blob/9eccb7943f012b4b20c57107bc40e7cdeccbc145/gazelle/python/parser.go#L47) I don't see a clear path to compiling this gazelle plugin to be a portable go binary.

Is this intentional or could there be an alternative that allows for this?

linzhp:
The runtime dependencies of Gazelle Python extension makes it hard to distribute Gazelle binaries: we have to preserve the runfiles structure and distribute it with Gazelle binaries.

Instead, we can build a single Python zip file that comes a built-in interpreter, and embed the zip file into the Go binary in compile time and avoid the runtime dependency.

Fixes #1455


gazelle/python/parser.go
```diff
@@ -35,23 +36,19 @@ var (
...
-
-	parseScriptRunfile, err := rfiles.Rlocation("rules_python_gazelle_plugin/python/parse")
-	if err != nil {
-		log.Printf("failed to initialize parser: %v\n", err)
-		os.Exit(1)
-	}
-
-	cmd := exec.CommandContext(ctx, parseScriptRunfile)
-	cmd.Env = append(os.Environ(), rfiles.Env()...)
+	cmd := exec.CommandContext(ctx, "python3", parserPath)
```
aignas, line 51: This is using system Python, right? What would be the way to use the hermetic toolchain?

linzhp: > This is using system Python, right?
Yeah, this only preserves the status quo. The old script from `rfiles.Rlocation("rules_python_gazelle_plugin/python/parse")` has this line at the beginning:
```
#!/usr/bin/env python3
```
So it's calling system Python too, although rules_python may let people feel it's using the hermetic Python toolchain.

> What would be the way to use the hermetic toolchain?
I have some idea to do it, but it's probably better to separate that into a different PR.

aignas: The shebang there calls the system python, but then it launches the actual Python file `parse.py` using the toolchain included via the `py_binary`. I think some of this confusion stems from #691.

This PR makes the system Python dependency more present, because now the `std_module.py` will use the Python version that is on the system as opposed to the one configured by the `py_binary` if I understand things correctly.

gazelle/python/BUILD.bazel
```diff
@@ -16,9 +15,9 @@ go_library(
         "std_modules.go",
         "target.go",
     ],
-    data = [
-        ":parse",
-        ":std_modules",
+    embedsrcs = [
+        "parse.py",
+        "std_modules.py",
```
aignas, line 20: To me this change would be fine if we would have previously had
```
data = [
    ":parse.py",
    ":std_modules.py",
]
```

The `py_binary` usage in here makes use of the default Python toolchain that is registered in the root module/workspace.

Hence the change made here is not equivalent. It would be if we had a `current_py_toolchain` in runtime gazelle deps and if we located that via the runfiles API.

aignas (changes requested):

fmeum:
> 1. It assumes that the Gazelle plugin repo is called `rules_python_gazelle_plugin". This may not be the case when the Gazelle plugin is managed as a go dependency, and by convention named as "com_github_bazelbuild_rules_python_gazelle".

This could also be fixed by baking the `$(rlocationpath ...)` into `x_defs` of the `go_library`.

> 2. It makes Gazelle only runnable under `bazel run`. So we won't be able to debug Gazelle using regular Go IDEs without special Bazel plugins, or distribute pre-built Gazelle binary

Could you explain why that is the case? Runfiles discovery should work equally well when the binary is executed directly as long as its `.runfiles` directory is present beside it. If that doesn't work, then there is another bug to fix in the setup.

gazelle/python/parser.go
linzhp: Oh, interesting. I didn't know that. Thanks for pointing out. Now I embed the python zip file instead, which comes with a Python interpreter. So it only relies in system python to locate that interpreter.

linzhp:
> Could you explain why that is the case? Runfiles discovery should work equally well when the binary is executed directly as long as its .runfiles directory is present beside it. If that doesn't work, then there is another bug to fix in the setup.

The concept of ".runfiles" doesn't exist outside Bazel world. If we want to distribute pre-built Gazelle binary, we have to also distribute the runfiles in a certain layout; if people want to debug Gazelle using IDEs that use "go build" under the hood, they won't be able to find the runfiles.

In addition, I think runtime dependency makes sense when we need to change it at runtime, e.g., a config file for prod, and another config file for staging. For such files, the location should be configurable through command line arguments or environment variables. For static content like html template or Python scripts like this, we never want to change them in runtime, it's better to embed them into the binary, so we have one less thing to worry about when deploying/distributing the binary.

fmeum:
Makes sense. In the Go case, Gazelle with Bzlmod reads files generated by Bazel (the list of repos), so it would be difficult to distribute it as a single binary anyway. But if it works for Python, that's good.

gazelle/python/python_test.go
```diff
@@ -21,18 +21,15 @@ package python_test
 
 import (
 	"bytes"
-	"context"
 	"errors"
+	"github.com/bazelbuild/bazel-gazelle/testtools"
+	"github.com/bazelbuild/rules_go/go/tools/bazel"
+	"github.com/ghodss/yaml"
```
aignas, line 27: nit: is this expected behaviour of `goimports`?

gazelle/python/parser.go
```diff
@@ -38,20 +38,7 @@ var (
...
-
-	parseScriptRunfile, err := rfiles.Rlocation("rules_python_gazelle_plugin/python/parse")
-	if err != nil {
-		log.Printf("failed to initialize parser: %v\n", err)
-		os.Exit(1)
-	}
-
-	cmd := exec.CommandContext(ctx, parseScriptRunfile)
-	cmd.Env = append(os.Environ(), rfiles.Env()...)
+	cmd := exec.CommandContext(ctx, "python3", pyzPath, "parse")
```
aignas, line 41: nit: it may be good to add a comment stating why this is system python and mention #691 as the ticket which may change this need.

aignas (approved):
Overall, the change looks good to me (once the imports are tidied up), thanks for the improvements! Could you please change the description of the PR so that it more suitable as a commit message (after the commit is merged, the PR description is going to be used for it), add a CHANGELOG.md entry and mention the #1455 ticket in the description. I personally think that it may fix the issue reported in #1455, but it would be good to double check with the issue author.

@rickeylev, I am not too familiar with the `python_zip` functionality that is used here and I am wondering if there are any gotchas here which we should be careful about before merging?

In the long term, I think maight probably need include `zip` files for each `X.Y` interpreter version we support and add additional support in `gazelle` which would allow us to tell it to use a particular Python version to use. That way we could correctly handle edge cases where `tomllib` could be detected as a `std_module` in `Python 3.11` and above but not in earlier versions. That is definitely outside the scope of this PR.

gazelle/python/python_test.go
```diff
@@ -152,9 +149,7 @@ func testPath(t *testing.T, name string, files []bazel.RunfileEntry) {
 
 		args := []string{"-build_file_name=BUILD,BUILD.bazel"}
 
-		ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
-		t.Cleanup(cancel)
-		cmd := exec.CommandContext(ctx, gazellePath, args...)
+		cmd := exec.Command(gazellePath, args...)
```
aignas, line 152: nit: why are we removing the 2 second timeout here? Does running `gazelle` become really expensive because of the needing to write the `embedded` file? What is roughly the impact?

gazelle/def.bzl
```diff
@@ -16,6 +16,4 @@
 """
 
 GAZELLE_PYTHON_RUNTIME_DEPS = [
```
aignas, line 18: Please update the `example` docs and the `gazelle/README.md` to not use this import.

gazelle/python/python_test.go
linzhp: yeah, writing and extracting dozens of zip files take time. I introduced a new variable so tests can use the runfiles directly

gazelle/python/python_test.go
linzhp: oops

linzhp:
> In the long term, I think maight probably need include zip files for each X.Y interpreter version we support and add additional support in gazelle which would allow us to tell it to use a particular Python version to use. That way we could correctly handle edge cases where tomllib could be detected as a std_module in Python 3.11 and above but not in earlier versions. That is definitely outside the scope of this PR.

We can just document that for those who are prebuilding and distributing Gazelle binaries, the registered Python interpreter in the repo where Gazelle is built needs to be the same as the target repo. It's the distributor's responsibility to build a Gazelle binary for each combination of OS+arch+python interpreter version.

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ A brief description of the categories of changes:
 * Make `//python/pip_install:pip_repository_bzl` `bzl_library` target internal
   as all of the publicly available symbols (etc. `package_annotation`) are
   re-exported via `//python:pip_bzl` `bzl_library`.
+* Gazelle Python extension no longer have runtime dependencies.
```
aignas, line 27:
```suggestion
* Gazelle Python extension no longer has runtime dependencies.
```

aignas:
> > In the long term, I think maight probably need include zip files for each X.Y interpreter version we support and add additional support in gazelle which would allow us to tell it to use a particular Python version to use. That way we could correctly handle edge cases where tomllib could be detected as a std_module in Python 3.11 and above but not in earlier versions. That is definitely outside the scope of this PR.
> 
> We can just document that for those who are prebuilding and distributing Gazelle binaries, the registered Python interpreter in the repo where Gazelle is built needs to be the same as the target repo. It's the distributor's responsibility to build a Gazelle binary for each combination of OS+arch+python interpreter version.

+1 for documenting this for now. The `gazelle_python.yaml` needs to be version aware for this to fully work, right now the `gazelle` Python plugin does not work well in cases where repos are using version aware rules and have multiple Python toolchains being used at the same time. However this is not something we can solve right now with the current structure.

linzhp:
Updated the doc, but somehow caused integration tests on Windows to fail

gazelle/README.md
```diff
@@ -7,7 +7,7 @@ Gazelle may be run by Bazel using the gazelle rule, or it may be installed and r
 
 This directory contains a plugin for
 [Gazelle](https://github.com/bazelbuild/bazel-gazelle)
-that generates BUILD files content for Python code.
+that generates BUILD files content for Python code. When Gazelle is run as a command line tool with this plugin, it embeds with a Python interpreter, which is the one Bazel resolves to when it is building the plugin. The behavior of the plugin is different with different version of the interpreter. Distributors of Gazelle binaries need to build a Gazelle binary for each OS+CPU architecture+Python version combination they are targeting.
```
aignas, line 10:
```suggestion
that generates BUILD files content for Python code. When Gazelle is run as a command line tool with this plugin, it embeds a Python interpreter resolved during the plugin build.
The behavior of the plugin is slightly different with different version of the interpreter as the Python `stdlib` changes with every minor version release.
Distributors of Gazelle binaries should, therefore, build a Gazelle binary for each OS+CPU architecture+Minor Python version combination they are targeting.
```

Issue #1357, opened 2023-08-02
Use bazel's downloader for downloading pip packages


cameron-martin:
# 🚀 feature request

### Relevant Rules

The relevant rule is `whl_library`.

### Description

Currently [whl_library](https://github.com/bazelbuild/rules_python/blob/e355becc30275939d87116a4ec83dad4bb50d9e1/python/pip_install/pip_repository.bzl#L658) downloads pip packages by executing a python script. However this bypasses bazel's downloader, causing us to lose a few nice features such as:

* The ability to rewrite these urls to a mirror by using `--experimental_downloader_config`.
* Sharing the downloader cache between machines to speed up downloads.

### Describe the solution you'd like

It would be nice to use bazel's downloader for these, by using `repository_ctx.download`.

### Describe alternatives you've considered

Not doing this and keeping it like it is, I suppose.

PR #1487, opened 2023-10-12
refactor(whl_library): split wheel downloading and extraction into separate executions

aignas:
Before the PR the downloading/building of the wheel and the extraction
would be done as a single step, which meant that for patching of the
wheel to happen, we would need to do it within the python script. In
order to have more flexibility in the approach, this PR splits the
process to two separate invocations of the wheel_installer, which
incidentally also helps in a case where the downloading of the wheel
file can happen separately via http_file.

Related issues #1076, #1357

### DIFF
--- a/python/pip_install/private/generate_whl_library_build_bazel.bzl
+++ b/python/pip_install/private/generate_whl_library_build_bazel.bzl
@@ -60,7 +60,7 @@ filegroup(
 
 filegroup(
     name = "{whl_file_label}",
-    srcs = glob(["*.whl"], allow_empty = True),
+    srcs = ["{whl_name}"],
     data = {whl_file_deps},
 )
 
@@ -86,7 +86,9 @@ py_library(
 """
 
 def generate_whl_library_build_bazel(
+        *,
         repo_prefix,
+        whl_name,
         dependencies,
         data_exclude,
         tags,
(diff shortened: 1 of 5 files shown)


aignas:
Note, that with this we could in theory set up a cyclic dependency test with:
* Creating a repository rule that creates two wheels via the `py_wheel`.
* Creating two `whl_library` rules that instead of downloading a wheel from requirements use the wheel targets provided by the `py_wheel`.

python/pip_install/private/generate_whl_library_build_bazel.bzl
```diff
@@ -86,7 +86,9 @@ py_library(
 """
 
 def generate_whl_library_build_bazel(
+        *,
```
groodt, line 89: nit: can we call it `*args`?

I'm not entirely following why we need this? Can you elaborate?

groodt (approved):
LGTM

python/pip_install/private/generate_whl_library_build_bazel.bzl
aignas: This is to force all of the arguments to be specified as keyword arguments. this function has lots of arguments and supporting positional arguments may be a little bit error prone.

groodt: ah, right. of course.

PR #1488, opened 2023-10-12
feat(whlmaker): introduce an internal _WhlFile class and stop sorting RECORD

aignas:
This class is for being able to more easily recreate a wheel file after
extracting it. This is not intended for usage outside the rules_python
project. Also stop sorting the entries when writing a RECORD file making
the order of the RECORD file to be the same as the order the files to
the zip file are added.

Towards #1076


groodt (approved):
LGTM

I'm not entirely sure what our preference is in terms of: https://github.com/bazelbuild/rules_python/pull/1453

aignas:
I would like to merge #1453  and then rebase and merge this.

groodt:
> I would like to merge #1453 and then rebase and merge this.

SGTM

Issue #1484, opened 2023-10-11
Setup ReadTheDocs


aignas:
See [this comment](https://github.com/bazelbuild/rules_python/pull/1478#issuecomment-1756834416) by @rickeylev 

The main idea would be to setup a ReadTheDocs instance for better documentation and this could be a way to solve #1332. There is a reserved URL for this project and the setup could be split into multiple PRs as follows:
1. Look at http://github.com/bazelbuild/rules_testing and mimic the setup of the generated docs.
2. Setup a way to locally start a docs server.
3. Add integration to push docs for the latest `main` version.
4. Add integration to push docs for each tagged release version.

rickeylev:
The RTD project is: https://readthedocs.org/projects/rules-python/

> Setup a way to locally start a docs server.

With Sphinx, all this requires is a local webserver pointing to a directory. This is because Sphinx generates all the necessary files. In rules_testing, I just have a simple sh_binary that runs `python -m http.server` after being given the generated files as data dependencies.

https://github.com/bazelbuild/rules_testing/blob/master/docs/run_sphinx_build.sh

It could be spruced up a bit, but its simple overall.

> Add integration to push docs for the latest main version.
> Add integration to push docs for each tagged release version.

I think the docs for this are around here: https://docs.readthedocs.io/en/stable/guides/setup/index.html

From what I remember, this requires setting up a webhook json token thing on the github project and/or the RTD project. I can't remember exactly. I tried to add your email @aignas to the rules_python RTD project, but it said it didn't exist; you might have to create an account on RTD?

Building for different versions pretty much Just Works right ouf the box. RTD has logic to detect version-looking tags and considers them releases. It builds them automatically.

rickeylev:
> From what I remember, this requires setting up a webhook ...

Oh nice, I think this was all automatically setup when I created the RTD project. All we need in the `.readthedocs.yaml` and a build target to generate docs

rickeylev:
Neat, it worked! https://rules-python.readthedocs.io/en/latest/index.html

I mean, it's a bit funny looking right now because of some styling and markdown oddities, but we got something basic going now.

Notes to self:

* Crossrefs are a bit annoying because myst doesn't have an easy way to include link labels in a reusable way. I hacked around this in rules_testing by just appending a list of link names to the generated files. Worked well enough. The nice thing with this approach is you could just write [\`foo\`] and it wold link to the named label. As far as I can tell, there isn't any way to intercept how myst parses and resolves that.
[...]

Issue #1332, opened 2023-07-21
Refactor documentation


chrislovecnm:
The README is getting very long.  We may want to consider pulling the WORKSPACE and bzlmod instructions into separate files under the docs folder.

rules_go is doing this here https://github.com/bazelbuild/rules_go/blob/master/docs/go/core/bzlmod.md

The examples are getting quite complex for testing purposes, so having a first-time user dig through the examples is probably not best.  We would refer to the examples as more advanced usage.

PR #1489, opened 2023-10-13
docs: initial doc generation using Sphinx

rickeylev:
This lays the groundwork for using Sphinx to generate user-facing documentation and
having it published on readthedocs. It integrates with Bazel Stardoc to generate
MyST-flavored Markdown that Sphinx can process.

There are 4 basic pieces that are glued together:
  1. `sphinx_docs`: This rule invokes Sphinx to generate e.g. html, latex, etc
  2. `sphinx_stardoc`: This rule invokes Stardoc to generate MyST-flavored Markdown
     that Sphinx can process
  3. `sphinx_build_binary`: This rule defines the Sphinx executable with any necessary
     dependencies (e.g. Sphinx extensions, like MyST) to process the docs in (1)
  4. `readthedocs_install`: This rule does the necessary steps to build the docs and
     put them into the location the readthedocs build process expects. This is basically
     just `cp -r`, but its cleaner to hide it behind a `bazel run` command than have
     to put various shell in the readthedocs yaml config.

* Bump Bazel 6 requirement: 6.0.0 -> 6.20. This is necessary to support
  bzlmod and Stardoc.

Work towards #1332, #1484


docs/BUILD.bazel
```diff
@@ -196,3 +196,114 @@ sh_binary(
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Generated documentation for the docs/ directory
+
+load(":docgen.bzl", "sphinx_stardocs")
+
+sphinx_stardocs(
```
aignas, line 220: IMHO this could be a separate workspace within this `git` repo. Since Sphinx is a Python project, having `sphinxdocs` next to `python` at the root of the repo would be an interesting way to share code with others.

docs/docgen.bzl
```diff
@@ -0,0 +1,104 @@
...
+def sphinx_docs(name, **kwargs):
+    _sphinx_docs(name = name, **kwargs)
+
+def _sphinx_docs_impl(ctx):
+    name = ctx.label.name
+
+    print(ctx.files.srcs)
+    source_dir = paths.dirname(ctx.files.srcs[0].path)
+    source_dir = "docs"
+    print("sd=", source_dir)
```
aignas, line 77: just a note that this should be cleaned up later.

docs/index.md
```diff
@@ -0,0 +1 @@
+Hello, world
```
aignas, line 1: nit: Should this be something else?

aignas (commented):
Really interesting, thanks for sharing this!

docs/BUILD.bazel
phlax: fwiw we (envoy) have quite a bit of code wrapping bazel/sphinx, which we reuse in a few places

i had been thinking to factor out related code, push more to bazel and shift to our tooling repo - i think a bazel sphinx rule would be generally useful (putting aside ongoing maintenance issue wrt sphinx and bazel support specifically)

WORKSPACE
```diff
@@ -103,6 +103,21 @@ load("@publish_deps//:requirements.bzl", "install_deps")
 
 install_deps()
 
+#####################
+# Install sphinx for doc generation.
+load("@python//3.11.6:defs.bzl", "interpreter")
+load("@rules_python//python:pip.bzl", "pip_parse")
+
+pip_parse(
```
aignas, line 111: nit: we could add the equivalent to `MODULE.bazel` and mark the `sphinx` deps as `dev_dependencies` and support this under `bzlmod`.

However, I am not sure if we have support for `module_ctx.is_dev_dependency` in the `pip.parse`. If not, then it is fine to leave as is.

docs/sphinx/BUILD.bazel
```diff
@@ -0,0 +1,74 @@
...
+    ],
+)
+
+# Run bazel run //docs/sphinx:requirements.update
+compile_pip_requirements(
+    name = "requirements",
+    requirements_in = "requirements.in",
+    requirements_txt = "requirements.txt",
+    # The requirements output differs on Windows, so just restrict it to Linux.
+    # The build process is only run on, and only works for, Linux anyways.
```
aignas, line 72: nit: what are the reasons to not support Mac?

docs/BUILD.bazel
rickeylev: Yeah, it's definitely the sort of functionality that should be more generally usable. Putting it in the stardoc rules was floated as an idea, too, when I initially did this for rules_testing. I wrote the code to be mostly re-usable, but I think it will need a couple more iterations to entirely tease out how to make it a stand-alone project. There are various issues like

I moved the sphinx code to //sphinxdocs in that vein to better conceptually separate rules_python's docs from sphinx stuff. I don't want to create a separate workspace, though. Partially because we don't have sufficient CI slots to test it, but also because I think it's premature to do so (the stardoc templates, theme, css, myst extension, and markdown design decisions are all somewhat intertwined, and each are project-specific decisions; untangling that is more work than I can put in right now), and it involves having to wire more things together, which just seems like extra work with little gain at this point.

WORKSPACE
rickeylev: For some weird reason, bzlmod isn't working with this.

```
ERROR: Analysis of target '//docs/sphinx:docs' failed; build aborted: module extension "toolchains" from "@rules_java~6.3.0//java:extensions.bzl" does not generate repository "remote_java_tools_darwin", yet it is imported as "remote_java_tools_darwin" in the usage at bazel_tools@_/MODULE.bazel:20:32
```

Which is weird, because didn't we just enable bzlmod + stardoc a few weeks ago? I'll have to poke this a bit more.

docs/sphinx/BUILD.bazel
rickeylev: Mostly because the docs will only be built on Ubuntu. Mac should be similar enough it shouldn't be an issue; added requirements_darwin

rickeylev:
Ok, I've cleaned this up enough to be ready for an initial review. This PR just lays the groundwork -- I'm not writing the complete porting of docs over to RTD in this PR.

What I'm going to do is:
1. Land this PR
2. Reconfigure RTD to read from a new branch, "rtd-testing". This is where I'll push commits to the RTD build process and get it working
3. Land another PR on main with the changes from (2). At this point, RTD will be building and publish main and our tagged versions.
4. Enable RTD building for PRs
5. Then we can port our plain stardoc md/html docs over to Sphinx/RTD. This will require various changes to docstrings etc so that things group and render nicely.

WORKSPACE
aignas: I think the reason is that `.bazelversion` is `6.0.0` which is super old. Changing it to `6.3.2` allows me to run `:docs.serve`.

MODULE.bazel
```diff
@@ -54,3 +54,11 @@ use_repo(python, "pythons_hub")
 
 # This call registers the Python toolchains.
 register_toolchains("@pythons_hub//:all")
+
+# ===== DEV ONLY SETUP =====
+docs_pip = use_extension("@rules_python//python/extensions:pip.bzl", "pip")
```
aignas, line 59: nit: should we use something like this?
```suggestion
docs_pip = use_extension("//python/extensions:pip.bzl", "pip")
```

sphinxdocs/sphinx.bzl
```diff
@@ -0,0 +1,191 @@
...
+    output_dir = ctx.actions.declare_directory("{}/{}".format(output_prefix, format))
+
+    args = ctx.actions.args()
+    args.add("-T")  # Full tracebacks on error
+    args.add("-b", format)
+    args.add("-c", paths.dirname(ctx.file.config.path))
+    args.add("-q")  # Suppress stdout informational text
+    args.add("-j", "auto")  # Build in parallel, if possible
+    args.add("-E")  # Don't try to use cache files. Bazel can't make use of them.
+    args.add("-a")  # Write all files; don't try to detect "changed" files
```
aignas, line 179: nit: it would be great to use parameters with full names (instead of `-T`) we can use something that does not require inline comments. However, since we have inline comments here, it is good.

sphinxdocs/sphinx.bzl
```diff
@@ -0,0 +1,191 @@
...
+    source_marker = ctx.actions.declare_file(source_prefix + "/__marker")
+    ctx.actions.write(source_marker, "")
+    sphinx_source_dir_path = paths.dirname(source_marker.path)
+    sphinx_source_files = []
+    for orig in ctx.files.srcs:
+        source_rel_path = orig.short_path
+        if source_rel_path.startswith(ctx.attr.strip_prefix):
+            source_rel_path = source_rel_path[len(ctx.attr.strip_prefix):]
+
+        sphinx_source = ctx.actions.declare_file(source_prefix + "/" + source_rel_path)
```
aignas, line 159: nit: what about using `bazel-skylib` `paths.join` here and elsewhere in the PR?

sphinxdocs/readthedocs_install.py
```diff
@@ -0,0 +1,28 @@
...
+
+  rf = runfiles.Create()
+  for doc_dir_runfiles_path in args:
+    doc_dir_path = pathlib.Path(rf.Rlocation(doc_dir_runfiles_path))
+    dest = install_to / doc_dir_path.name
+    print(f"Copying {doc_dir_path} to {dest}")
+    shutil.copytree(src=doc_dir_path, dst=dest, dirs_exist_ok=True)
+
+if __name__ == "__main__":
+  sys.exit(main(list(sys.argv)))
```
aignas, line 28: nit: passing `sys.argv[1:]` here would not necessitate the `del argv`.

aignas (approved):
In general this looks great! looking forward to better docs. Maybe linking #1332 should be mentioned in the PR description?

sphinxdocs/sphinx.bzl
rickeylev: I agree, but as far as I can tell, Sphinx doesn't have long names for them.

MODULE.bazel
rickeylev: Done

sphinxdocs/readthedocs_install.py
rickeylev: Done

sphinxdocs/sphinx.bzl
rickeylev: Done

WORKSPACE
rickeylev: Thanks. Yes, that was it. I updated it to 6.2.0, since that's the lowest version that things seem happy with.

rickeylev:
> linking #1332 should be mentioned in the PR description?

Done

PR #1490, opened 2023-10-13
refactor(visibility)!: limit visibility of an internal library

aignas:
This was arguably made too-visible previously and since all of the consumable
symbols by the end users are re-exported via the //python:pip_bzl, we can keep
everything that is in pip_install internal.

This could be a breaking change for people who are depending on internal
symbols or if they are using `whl_library` to create their own implementation
of the `pip_repository` rule and are generating documentation internally.
However, in that case they can apply a small patch to change the visibility of
the `pip_repository_bzl` target.


aignas:
I have created this to start a discussion around visibility. I am OK to close
this without merging if others think that restricting the visibility does not
bring enough value.

rickeylev:
Yeah, we want to make //docs private, see https://github.com/bazelbuild/rules_python/issues/1458

I made //python/pip_install:pip_repository_bzl public because, ... hm why did I do that. I think I was just trying to provide some alternative to the prior public bzl_library targets in //docs that covered pip_repository.bzl. It's very unclear to me whether pip_repository.bzl is supposed to be a publicly loadable target or not, and some of the old targets in //docs were very broad. The top of the docs, for example, says `package_annotation` is public. So it wasn't entirely clear how to split and restrict things.

groodt (approved):
LGTM

chrislovecnm:
@groodt when you do a lgtm could you push the merge button?

PR #1494, opened 2023-10-16
test: remove usage of deprecated method `TestCase.assertEquals`

dizzy57:
`TestCase.assertEquals` is an alias for `TestCase.assertEqual`.
[This method is deprecated since Python 3.2 and was removed in Python 3.12](https://docs.python.org/3/whatsnew/3.12.html#id3).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,7 +33,6 @@ A brief description of the categories of changes:
   like `bazel query rdeps(...)`.
 
 * Python code generated from `proto_library` with `strip_import_prefix` can be imported now.
-* Removed usage of a deprecated method `TestCase.assertEquals` in tests.
 
 * (py_wheel) Produce deterministic wheel files and make `RECORD` file entries
   follow the order of files written to the `.whl` archive.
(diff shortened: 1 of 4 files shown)


dizzy57:
Meta-question: should we have a linter for this?
I'd like to suggest using [ruff](https://docs.astral.sh/ruff/). It is fast, configurable, and integrates well with existing pre-commit setup in `rules_python`.

aignas:
`ruff` could be a good addition as a separate PR. Feel free to add one and add me as a reviewer.

I would be of an opinion to keep this change out of `CHANGELOG.md` as this is not visible from the stakeholder point of view.

groodt (approved):
LGTM

Agree with removing the CHANGELOG entry.

dizzy57:
> `ruff` could be a good addition as a separate PR. Feel free to add one and add me as a reviewer.

Will do.


> I would be of an opinion to keep this change out of `CHANGELOG.md` as this is not visible from the stakeholder point of view.

I was thinking the same, but decided to go by letter of PR checklist.
Reverted changes to `CHANGELOG.md`.

dizzy57:
Rebased to resolve merge conflicts.

PR #1511, opened 2023-10-19
docs: make readthedocs render a bit nicer and port some docs over to Sphinx

rickeylev:
This makes the Sphinx-based docs hosted on readthedocs render a bit more nicely, fixes a few issues, and adds some features to //sphinxdocs

Doc fixes/improvements:
* Ports various docs over to Sphinx pages. They're split out from the readme file.
* Version RTD is building is reflected in the docs
* Fixes some references to github files
* Includes the custom CSS file that styled the api docs
* Removes `-Q` from doc building; all warnings should be fixed now
* Added Bazel inventory file. Bazel doesn't provide one, but we can manually provide on
  and still use intersphinx functionality.
* Added `gh-path` custom role. This is a shortcut for writing the whole github URL.
* Sets the primary domain to None. The default is py, which we don't use much of, so it
  just results in confusing crossref errors.
* Enable nitpicky mode to catch more errors.
* Remove the `starlark` marker from codeblocks; that name isn't recognized by Sphinx.
  The highlighting is still sufficient.
* Adds a glossary

Sphinxdocs improvements:
* Added a flag to pass along arbitrary `-D` args to the Sphinx invocations. This allows
  e.g., the `version` setting of the docs to be set on the command line from the
  `READTHEDOCS_VERSION` environment variable
* Added inventory file generation. These are files that allow referencing external
  projects using intersphinx.
* `sphinx_stardocs` have their public load path set as their page title. This groups the
  API docs more naturally by file. The path can be customized.
* `sphinx_stardocs` can have a footer specified for generated pages. This allows easily
  added a list of link labels for easy re-use.
* `readthedocs_install` now tries harder to find an open port
* The conf.py file is moved into the generated sources directly. This was done because some
  config settings are relative to the conf.py file, which was being placed one directory
  above the regular sources.


rickeylev:
This can be previewed at: https://rules-python.readthedocs.io/en/latest/

docs/sphinx/pypi-dependencies.md
```diff
@@ -0,0 +1,153 @@
...
+        requirement("some_pip_dep"),
+        requirement("another_pip_dep"),
+    ]
+)
+```
+
+The reason `requirement()` exists is that the pattern for the labels,
+while not expected to change frequently, is not guaranteed to be
+stable. Using `requirement()` ensures you do not have to refactor
+your `BUILD` files if the pattern changes.
```
aignas, line 105: nit: this is an interesting statement. not sure if we can actually rely on this in the future when making changes to it though.

docs/sphinx/pypi-dependencies.md
```diff
@@ -0,0 +1,153 @@
...
+buildozer command:
+
+```shell
+buildozer 'substitute deps @old_([^/]+)//:pkg @new_${1}//:pkg' //...:*
+```
+
+For `pip_install`, the labels are instead of the form:
+
+```
+@{name}//pypi__{package}
```
aignas, line 129: nit: it would be good to remove this, because no code is producing such labels anymore, I think.

python/private/py_console_script_binary.bzl
```diff
@@ -49,19 +49,19 @@ def py_console_script_binary(
     """Generate a py_binary for a console_script entry_point.
 
     Args:
-        name: str, The name of the resulting target.
-        pkg: target, the package for which to generate the script.
-        entry_points_txt: optional target, the entry_points.txt file to parse
+        name: [`target-name`] The name of the resulting target.
```
aignas, line 52: nit: the related `docs/*.md` file has more useful user documentation that we may want to also move to the sphinx docs.

sphinxdocs/private/sphinx_server.py
```diff
@@ -28,5 +31,19 @@ def __init__(self, *args, **kwargs):
     return 0
 
 
+@contextlib.contextmanager
+def _start_server(handler, ip, start_port):
+    for port in range(start_port, start_port + 10):
```
aignas, line 36: nit: an alternative would be to open a socket on a random port and then close it and then use the previously allocated port. It's a little racy, but could work most of the time. That said, this is probably better.

docs/sphinx/conf.py
```diff
@@ -5,37 +5,35 @@
...
+    "bazel": ("https://bazel.build/", "bazel_inventory.inv"),
+}
 
-intersphinx_disabled_domains = ["std"]
-
-# Prevent local refs from inadvertently linking elsewhere, per
-# https://docs.readthedocs.io/en/stable/guides/intersphinx.html#using-intersphinx
-intersphinx_disabled_reftypes = ["*"]
+extlinks = {
+    "gh-path": ("https://github.com/bazelbuild/rules_python/tree/main/%s", "%s"),
```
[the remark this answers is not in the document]
rickeylev: Hm. Maybe? The extlinks plugin only allows a single `%s`, so we'd have to compute the URL in the conf file. I don't see a way to defer lookup, either. Unfortunately, it looks like the `-D` setting is somewhat special and only works with "known" config settings. It looks like there are html-output specific settings that carry the github url...somewhere.

Something for another time?

python/private/py_console_script_binary.bzl
rickeylev: Oh right, you added the merging functionality. This can be done using an include pretty easily.

docs/sphinx/pypi-dependencies.md
rickeylev: Done

.readthedocs.yml
```diff
@@ -6,5 +6,6 @@ build:
   tools:
     nodejs: "19"
   commands:
+    - env
     - npm install -g @bazel/bazelisk
-    - bazel run //docs/sphinx:readthedocs_install
+    - bazel run --//sphinxdocs:extra_defines=version=$READTHEDOCS_VERSION //docs/sphinx:readthedocs_install
```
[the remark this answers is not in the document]
rickeylev: Done

README.md
```diff
@@ -8,9 +8,7 @@ This repository is the home of the core Python rules -- `py_library`,
 `py_binary`, `py_test`, `py_proto_library`, and related symbols that provide the basis for Python
 support in Bazel. It also contains package installation rules for integrating with PyPI and other indices. 
 
-Documentation for rules_python  lives in the
-[`docs/`](https://github.com/bazelbuild/rules_python/tree/main/docs)
-directory and in the
+Documentation for rules_python is at rules-python.readthedocs.io and in the
```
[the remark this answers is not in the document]
rickeylev: Done

docs/sphinx/getting-started.md
```diff
@@ -0,0 +1,181 @@
...
+
+1. For submodules to ensure they run with the appropriate Python version
+2. To allow incremental, per-target, upgrading to newer Python versions,
+   typically in a mono-repo situation.
+
+To configure a submodule with the version-aware rules, request the particular
+version you need, then use the `@python_versions` repo to use the rules that
+force specific versions:
+
+```
```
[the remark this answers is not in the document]
rickeylev: Done

docs/sphinx/pypi-dependencies.md
```diff
@@ -0,0 +1,153 @@
+## Using dependencies from PyPI
```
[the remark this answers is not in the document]
rickeylev: Done

rickeylev:
Hm, you know, I think i'll just delete the old stuff in docs/ in this PR, too. That'll avoid any odd rendering issues with myst-markdown in the bzl code, and fix the issue with bazel's downstream testing. I'll make sure to copy over any of the manual merging like what py_console_script_binary was doing

Issue #1496, opened 2023-10-17
Start using bazel@HEAD again for PYSTAR tests


aignas:
They broke as described in https://github.com/bazelbuild/bazel/issues/19838. This ticket is to re-enable them and in the mean time we can run them against the last known-to-work version.

EDIT: it seems that the tests are the only ones that are using bazel-in-bazel and we probably need to migrate them to properly support bzlmod.

PR #1514, opened 2023-10-19
refactor!: do not use a wrapper macro for pip_parse

aignas:
This brings back the generated documentation for the pip_parse
attributes making switches to default values more prominent.

Towards #1496.


aignas:
Verified that the docs for `pip_parse` are in https://rules-python--1514.org.readthedocs.build/en/1514/api/pip.html

groodt (approved):
LGTM

PR #1519, opened 2023-10-27
feat: support pyproject.toml in compile_pip_requirements

aignas:
With this PR we can also use `pyproject.toml` in addition
to `requirements.in` which helps in making the requirements
in a more structured form. For example, we could parse the
toml itself and create aliases in the hub repos only for the
packages outlined in the `pyproject.toml` file. The same for
`gazelle`, we could restrict `gazelle_python.yaml` contents
to only the dependencies listed in `pyproject.toml`.

Examples can be migrated once we agree on the interface.

Summary:
- feat: support pyproject.toml in compile_pip_requirements
- chore: use pyproject.toml for sphinx doc requirements

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,10 @@ A brief description of the categories of changes:
   default, which will cause `gazelle` to change third-party dependency labels
   from `@pip_foo//:pkg` to `@pip//foo` by default.
 
+* The `compile_pip_requirements` now defaults to `pyproject.toml` if the `src`
+  or `requirements_in` attributes are unspecified, matching the upstream
+  `pip-compile` behaviour more closely.
+
 Breaking changes:
 
 * (pip) `pip_install` repository rule in this release has been disabled and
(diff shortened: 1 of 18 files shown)


groodt (approved):
LGTM

I think in the description, could be worth clarifying that it's support PEP-621 `project.dependencies` https://packaging.python.org/en/latest/specifications/declaring-project-metadata/#declaring-project-metadata only. It shouldn't imply support of any other TOML tables that may exist, or tool specific tables such as `tool.black` or `tool.poetry` etc

rickeylev (approved):

rickeylev:
LGTM. I pushed a small update to the `src` arg doc.

The only comment, more question, I have is: should src default to `requirements.in` or `pyproject.toml`? If I understand correctly, toml files are the newer, better way?

aignas:
I have changed the code to default to `pyproject.toml` as I agree that people who are new to `rules_python` should use by default new standards, but that may cause issues for older users who were relying on default behaviour.

If you are OK with the changes, explanation, feel free to merge it during your working hours.

rickeylev (approved):

Issue #1520, opened 2023-10-30
rules_python build_file_generation integration tests on Ubuntu using minimum supported Bazel version is failing with "requirements_test on Bazel@HEAD


sgowroji:
https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/3423#018b7e52-e06d-4e38-8577-97686d10d343

Platform : [rules_python (build_file_generation integration tests on Ubuntu)](https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/3423#018b7e52-e06d-4e38-8577-97686d10d343)

Logs:
```

(02:05:20) FAIL: //:requirements_test (see /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/6a1011ee018bdbf38289288c626fc42e/execroot/_main/bazel-out/k8-fastbuild/testlogs/requirements_test/test.log)
 
FAILED: //:requirements_test (Summary)
```

Steps:
```
git clone -v https://github.com/bazelbuild/rules_python.git
git reset 5b082bb02b048a4b630843b6b582bd6f40d38b9e --hard
export USE_BAZEL_VERSION=8cab6aa21fa765985a962e59264e12251e4d479f
bazel test -- ...
```
Autosheriff culprit : https://github.com/bazelbuild/bazel/commit/32cc97e0a9a5810ab5469708f768666022c27642

CC Greenteam @comius

rickeylev:
qq: did the system python version change on the build hosts? i.e. from python 3.8 to python 3.9? That looks to be a common diff showing up. (I'm not asking to undo that, it would just explain why this failure shows up).

PR 1522 it out to partially address this -- many of those tests are trying to run bazel at head for the configs that should only be running Bazel 5.4. That will, at the least, skip those configs where it doesn't need to run.

I'm still looking into the root cause.

rickeylev:
Ah, I think this is because of switching bzlmod on by default. Setting `--noenable_bzlmod` when running some of the failing tests fixes it.

rickeylev:
Ok, I think I have PRs out for all the failures shown in that link. They're all fairly simple fixes. Most just need a MODULE.bazel file, a couple needed to change how they were finding files in runfiles.

rickeylev:
All the PRs are merged. I think that fixes all the failures. @sgowroji let us know if more failures show up in subsequent downstream bazel runs

sgowroji:
[...]

PR #1524, opened 2023-10-30
tests: make pip_parse example work with bzlmod enabled

rickeylev:
Bazel is enabling bzlmod by default, which means the examples need to be updated to be bzlmod compatible.

Work towards #1520

### DIFF
--- a/examples/pip_parse/BUILD.bazel
+++ b/examples/pip_parse/BUILD.bazel
@@ -47,7 +47,7 @@ py_test(
 
 py_console_script_binary(
     name = "yamllint",
-    pkg = "@pypi//yamllint:pkg",
+    pkg = "@pypi//yamllint",
 )
 
 # This rule adds a convenient way to update the requirements file.
--- a/examples/pip_parse/BUILD.bazel
+++ b/examples/pip_parse/BUILD.bazel
@@ -1,11 +1,6 @@
-load(
-    "@pypi//:requirements.bzl",
-    "data_requirement",
-    "dist_info_requirement",
-    "entry_point",
-)
 load("@rules_python//python:defs.bzl", "py_binary", "py_test")
 load("@rules_python//python:pip.bzl", "compile_pip_requirements")
+load("@rules_python//python/entry_points:py_console_script_binary.bzl", "py_console_script_binary")
 
 # Toolchain setup, this is optional.
 # Demonstrate that we can use the same python interpreter for the toolchain and executing pip in pip install (see WORKSPACE).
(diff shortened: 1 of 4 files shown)


examples/pip_parse/BUILD.bazel
```diff
@@ -50,9 +45,9 @@ py_test(
 # For pip dependencies which have entry points, the `entry_point` macro can be
 # used from the generated `pip_parse` repository to access a runnable binary.
 
-alias(
+py_console_script_binary(
     name = "yamllint",
-    actual = entry_point("yamllint"),
+    pkg = "@pypi//yamllint:pkg",
```
aignas, line 50: nit:
```suggestion
    pkg = "@pypi//yamllint",
```

examples/pip_parse/WORKSPACE
```diff
@@ -24,6 +24,7 @@ pip_parse(
     # can be passed
     # environment = {"HTTPS_PROXY": "http://my.proxy.fun/"},
     name = "pypi",
+    incompatible_generate_aliases = True,
```
aignas, line 27: nit: Please wait for #1525 and remove this flag.

examples/pip_parse/pip_parse_test.py
```diff
@@ -19,20 +19,28 @@
...
 
 class PipInstallTest(unittest.TestCase):
     maxDiff = None
 
+    def _remove_leading_dirs(self, paths):
+        # Removes the first two directories (external/<reponame>)
+        # to normalize what workspace and bzlmod produce.
+        #return [str(Path(*Path(v).parts[2:])) for v in paths]
+        return [
+            '/'.join(v.split('/')[2:])
```
aignas, line 33: FYI, I just wanted to comment that we should use an os specific path separator, but realized that the values come from `rlocationpaths` via an env var which will always use `/` separators, right?

aignas (approved):

examples/pip_parse/WORKSPACE
rickeylev: Done

examples/pip_parse/BUILD.bazel
rickeylev: Done

examples/pip_parse/pip_parse_test.py
rickeylev: Correct. Originally I used pathlib.Path to do this, but on windows, it changed things to `\` and then the paths didn't match the expected values.

PR #1525, opened 2023-10-31
chore(pip_parse, gazelle): generate/use hub repo aliases by default

### LINKED ISSUE #1498: Flip feature flags for 0.27.0
This is not necessarily to release 0.27.0 very soon, but to rather collect any extra info/feature flag toggles that we should do before the release is out.

The queued up changes are: https://github.com/bazelbuild/rules_python/compare/0.26.0...main

This is a list of flags that we may want to flip for this release and notify users that the previous will be removed in 0.28.0:
- [ ] `incompatible_generate_aliases` should default to `True` to make the third party repo structure very similar in `bzlmod` and non-bzlmod.
- [X] Start failing if users use `pip_install` instead of `pip_parse` we should introduce a message for the users to start using `pip_parse` instead. (See #1510)
- [x] Set default `py_wheel` related feature flags to True, see https://github.com/bazelbuild/rules_python/blob/main/python/private/py_wheel.bzl#L121 (#1331, #1513)

aignas:
This makes `pip_parse.incompatible_generate_aliases = True` the default. This only affects
workspace builds; it is already the default for bzlmod.

Summary:
- Enable the generation of aliases in `pip_repository`.
- Flip usage of aliases in `gazelle`.
- Remove usage of the old flags from the example code.
- Update the `gazelle` manifest generator to leave the
  `gazelle_python.yaml` manifest unchanged for people who have
  `use_pip_repository_aliases = True` in their `BUILD.bazel` files.
  Once they remove the flag, the `gazelle_python.yaml` will be updated.
- Update `multi_pip_parse` to handle sub-hub repositories where
  the `all_requirements` returns aliased targets.

Fixes #1498


rickeylev:
What's the incompatible part again? I can't remember what the potential incompatibility is. The existing `@pip_foo` targets will still exist and can still be directly referenced, right? The only edge case I can think of is maybe pip_parse.annotations? e.g. changes to the BUILD file would be reflected in the `@pip_foo` repo, but not `@pip//foo`.

Let's just make sure to note this in the changelog/docs.

Otherwise lgtm.

rickeylev (approved):

aignas:
I think for `non-bzlmod` it is a backwards incompatible change, however, in the early days it was an incompatible change for `bzlmod` users and the name just stuck.

PR #1531, opened 2023-11-01
cleanup(pystar): inline @bazel_tools and @platforms references

rickeylev:
The location of the `@bazel_tools` and `@platforms` repositories were originally part of the semantics.bzl config because performing rewrites on the code as part of the Bazel code export process was too difficult.

With the direction being reversed (imported instead of exported), and the scope of the codebase being reduced (just rules_python instead of the entire Bazel codebase), it's easier to perform copybara rewrites.

In particular, the `"//` strings are problematic to rewrite because they look like intra-repo references instead of parts of a larger expression.

### DIFF
--- a/python/private/common/attributes.bzl
+++ b/python/private/common/attributes.bzl
@@ -19,9 +19,7 @@ load(":py_internal.bzl", "py_internal")
 load(
     ":semantics.bzl",
     "DEPS_ATTR_ALLOW_RULES",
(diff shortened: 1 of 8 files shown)


groodt (approved):
LGTM

groodt (approved):
LGTM

aignas (approved):

PR #1532, opened 2023-11-01
feat(pip): provide pypi -> whl target mapping in requirements.bzl

alexeagle:
Currently a BUILD file can load `all_whl_requirements` but then can't determine which one is associated with a given package installed by the user.
This makes it impossible to build rules where the user can choose a given package and then override the wheel used, for example. @mattem and I are working at a client where we need this ability.

This PR makes a small, non-breaking refactoring to the generated `requirements.bzl` file so that this information is available in a new `all_whl_requirements_by_package` symbol.

Users can then do something like this:

```
load("@pip//:requirements.bzl", "all_whl_requirements_by_package")

some_rule(
  wheels = dict(all_whl_requirements_by_package, **{
    "charset-normalizer": "@charset_1_2_3//:file"
  }),
)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,10 @@ A brief description of the categories of changes:
 * (pip_parse) The installation of `pip_parse` repository rule toolchain
   dependencies is now done as part of `py_repositories` call.
 
+* (pip_parse) The generated `requirements.bzl` file now has an additional symbol
+  `all_whl_requirements_by_package` which provides a map from the original package name
+  (as it appears in requirements.txt) to the target that provides the built wheel file.
+
 * (pip_parse) The flag `incompatible_generate_aliases` has been flipped to
   `True` by default on `non-bzlmod` setups allowing users to use the same label
   strings during the transition period. For example, instead of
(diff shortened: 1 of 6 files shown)


rickeylev (changes requested):
Please also update the bzlmod implementation:
* python/private/bzlmod/pip_repository.bzl
* python/private/bzlmod/requirements.bzl.tmpl

rickeylev:
Can you also update the changelog, please?

alexeagle:
@rickeylev attempting to update the bzlmod implementation, but I'm not seeing the code used. Even in `examples/pip_parse` with `--enable_bzlmod`, I still get the `requirements.bzl` produced by the implementation I modified. Is the new bzlmod implementation not wired up yet?

rickeylev:
Yes, it's wired up.

The dependency goes:

python/extension/pip.bzl -> python/private/bzlmod/pip.bzl -> python/private/bzlmod/pip_repository.bzl -> python/private/bzlmod/requirements.bzl.tmpl

alexeagle:
Thanks, updated.

I found one issue that I don't think I want to try to solve here: in this code
https://github.com/bazelbuild/rules_python/blob/main/python/private/bzlmod/pip.bzl#L124-L128
we overwrite `whl_name` with the normalized name, but only after a comment about how users shouldn't need to guess our sanitization scheme. I agree with that comment - but FWICT we then use the normalized `whl_name` in places that ARE user-affordances like the `whl_overrides`.

As a consequence, under the bzlmod implementation, my feature has the normalized names appearing as the keys in `all_whl_requirements_by_package`. I think that's not desirable, but it seems like a substantial refactoring to instead take more care about the normalized `whl_name` vs. the original distribution name.

aignas (approved):

rickeylev (approved):

PR #1536, opened 2023-11-04
fix: Update MODULE.bazel comment

geaden:
Fix typo in MODULE.bazel commet


rickeylev (approved):
LGTM, but the checks are failing with some strange errors. The RTD build, for example, says the config file doesn't exist. Try updating to head?

PR #1537, opened 2023-11-06
test(gazelle): have a go_test target for each gazelle test

aignas:
This provides better caching and it allows us to have better
developer velocity whilst iterating on a single test case and
it also solves some of the `timeout` errors I was seeing locally
because now each `gazelle` invocation is run at a lower
parallelism that `bazel` decides itself.

### DIFF
--- a/gazelle/python/BUILD.bazel
+++ b/gazelle/python/BUILD.bazel
@@ -65,6 +65,8 @@ gazelle_test(
         # Use this so that we don't need to manually maintain the list.
         ["testdata/*"],
         exclude = ["testdata/*.md"],
+        # The directories aren't inputs themselves; we just want their
+        # names.
         exclude_directories = 0,
     ),
     deps = [
--- a/gazelle/python/BUILD.bazel
+++ b/gazelle/python/BUILD.bazel
@@ -1,6 +1,7 @@
 load("@bazel_gazelle//:def.bzl", "gazelle_binary")
-load("@io_bazel_rules_go//go:def.bzl", "go_library", "go_test")
+load("@io_bazel_rules_go//go:def.bzl", "go_library")
 load("@rules_python//python:defs.bzl", "py_binary")
+load(":gazelle_test.bzl", "gazelle_test")
 
 go_library(
     name = "python",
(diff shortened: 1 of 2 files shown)


gazelle/python/BUILD.bazel
```diff
@@ -53,13 +54,19 @@ filegroup(
...
         ":gazelle_binary",
         ":helper",
-    ] + glob(["testdata/**"]),
+    ],
+    testdata = glob(
+        # Use this so that we don't need to manually maintain the list.
+        ["testdata/*"],
+        exclude = ["testdata/*.md"],
+        exclude_directories = 0,
+    ),
```
aignas, line 69: @rickeylev, is this an appropriate use for a glob?

gazelle/python/gazelle_test.bzl
```diff
@@ -0,0 +1,49 @@
...
+
+    for testdatum in testdata:
+        _testdata, _, t = testdatum.rpartition("/")
+
+        test = "%s_%s" % (name, t)
+        tests.append(test)
+
+        go_test(
+            name = test,
+            data = native.glob(["%s/%s/**" % (_testdata, t)]) + data,
```
rickeylev, line 42: nit: use format() instead of %

gazelle/python/gazelle_test.bzl
```diff
@@ -0,0 +1,49 @@
...
+        testdata (list[str]): The list of folders in the 'testdata' directory
+            that we should create separate 'go_test' cases for. Each of them
+            will be prefixed with '{name}'.
+        **kwargs: extra arguments passed to 'go_test'.
+    """
+    tests = []
+
+    data = kwargs.pop("data", [])
+
+    for testdatum in testdata:
```
rickeylev, line 34: nit: "test_dir" instead of test_datum. That better reflects what the values are.

gazelle/python/gazelle_test.bzl
```diff
@@ -0,0 +1,49 @@
...
+
+load("@io_bazel_rules_go//go:def.bzl", "go_test")
+
+def gazelle_test(*, name, testdata, **kwargs):
+    """A simple macro to better cache gazelle integration tests
+
+    Args:
+        name (str): The name of the test suite target to be created and
+            the prefix to all of the individual test targets.
+        testdata (list[str]): The list of folders in the 'testdata' directory
```
rickeylev, line 25: nit: name "test_dirs" instead, as they aren't actually data, just directory names

gazelle/python/BUILD.bazel
rickeylev: Oh I see. The directories are being included because they're then used to feed another glob to create per-directory targets. I was going to comment that directories shouldn't be direct inputs, but that isn't the case here. Just a small comment to clarify. Renaming testdata to test_dirs would also help clarify.

```suggestion
    testdata = glob(
        # Use this so that we don't need to manually maintain the list.
        ["testdata/*"],
        exclude = ["testdata/*.md"],
        # The directories aren't inputs themselves; we just want their names.
        exclude_directories = 0,
    ),
```

rickeylev (approved):

gazelle/python/gazelle_test.bzl
aignas: Named it a `dir`

PR #1539, opened 2023-11-06
tests: disable bzlmod for workspace-only build_file_generation example

rickeylev:
The build_file_generation example only supports workspace style builds. The bzlmod equivalent is in `examples/bzlmod_build_file_generation`

Work towards #1520


aignas (approved):
I wonder if we should just merge the examples and add a warning to the Workspace file that someone is looking at the legacy example and that bzlmod is the preferred way. This obviously can be discussed separately

rickeylev:
Merging them is probably feasible, especially now that the bzlmod work has settled down. There are some features that are workspace only (e.g. entry points), that might make a simple consolidation harder, but we can probably figure something out there.

PR #1540, opened 2023-11-06
tests: disable bzlmod for workspace-only pip_repository_annotations example

rickeylev:
The pip_repository_annoations example is workspace-only. Its equivalent functionality is demonstrated in examples/bzlmod with the whl_mods feature.

Work towards #1520


aignas:
In order to save CI exec slots, we could merge this with the pip_parse example.

aignas (approved):

PR #1541, opened 2023-11-06
fix: always ignore `.pyc.NNNN` files from the hermetic runtime tree

### LINKED ISSUE #1261: flaky test: bzlmod integration test on MacOS
This is a tracking bug to collect info on the "bzlmod integration test on MacOS" to hopefully help figure out why it's failing. They usually pass after a couple retries

The error is typically something about the "Middleman" missing pyc files.

Example error:
```
ERROR: /Users/<snip>/rules-python-python/examples/<snip>: Middleman _middlemen/requirements_Utest-runfiles failed: missing input file '<snip>/big5freq.cpython-39.pyc.4316963824'
```

Current theories:
  * Issue with the build machines (wouldn't be the first time)

* hosts with **failures**:
  * bk-imacpro-2 (x2)
  * bk-imacpro-3
  * mk-imacpro-9
  * bk-imacpro-11 (x3)
  * bk-imacpro-12
  * bk-imacpro-18

* hosts with **successes**:
  * bk-imacpro-2
  * mk-imacpro-6
  * bk-imacpro-15 x2
  * bk-imacpro-18

**Failing** builds

https://buildkite.com/bazel/rules-python-python/builds/5036#01889b48-56c1-448c-bbac-cf37dad80b84

https://buildkite.com/bazel/rules-python-python/builds/5037#01889b60-d830-4827-ae25-a9c37e7c8ce8

https://buildkite.com/bazel/rules-python-python/builds/5029#

https://buildkite.com/bazel/rules-python-python/builds/5047#01889c11-eae2-4402-9ba5-81620511ea18

https://buildkite.com/bazel/rules-python-python/builds/5050#01889c3a-1c95-4f53-a538-0c7434773a5b

https://buildkite.com/bazel/rules-python-python/builds/5047#01889c11-eae2-4402-9ba5-81620511ea18

https://buildkite.com/bazel/rules-python-python/builds/5066#0188a254-986a-4bdc-a972-220424b5349a

https://buildkite.com/bazel/rules-python-python/builds/5067#0188a258-ed2a-455a-b5fe-3148f93d1231

https://buildkite.com/bazel/rules-python-python/builds/5084#0188b31f-dbab-4e48-9726-fe872304785c

**Successful** builds

https://buildkite.com/bazel/rules-python-python/builds/5039#01889b82-d0cc-4133-8712-cc11392043d0

https://buildkite.com/bazel/rules-python-python/builds/5029#01889bb5-4ed4-42c9-9816-4be86d9da129

https://buildkite.com/bazel/rules-python-python/builds/5067#_

https://buildkite.com/bazel/rules-python-python/builds/5084#0188b327-c85e-4f27-ac14-db9697815997

rickeylev:
Part of the pyc compilation process is to create a temporary file named `<name>.pyc.NNNN`, where `NNNN` is a timestamp. Once the pyc is entirely written, this file is renamed to the regular pyc file name. These files only exist for brief periods of time, but its possible for different threads/processes to see the temporary files when computing the glob() values. Later, since the file is gone, an error is raised about the file missing.

PR #1266 mostly fixed this issue, except that the exclude for the `.pyc.NNNN` files for an
interpreter runtime's files was behind the `ignore_root_user_error` flag, which meant it
wasn't always applied. This changes it to always be applied, which should eliminate the
failures due to the missing NNNN files.

Fixes #1261

Work towards #1520

### DIFF
--- a/python/repositories.bzl
+++ b/python/repositories.bzl
@@ -234,6 +234,7 @@ def _python_repository_impl(rctx):
         # tests for the standard libraries.
         "lib/python{python_version}/**/test/**".format(python_version = python_short_version),
         "lib/python{python_version}/**/tests/**".format(python_version = python_short_version),
+        "**/__pycache__/*.pyc.*",  # During pyc creation, temp files named *.pyc.NNN are created
     ]
 
     if rctx.attr.ignore_root_user_error:
@@ -243,7 +244,6 @@ def _python_repository_impl(rctx):
             # the definition of this filegroup will change, and depending rules will get invalidated."
             # See https://github.com/bazelbuild/rules_python/issues/1008 for unconditionally adding these to toolchains so we can stop ignoring them."
             "**/__pycache__/*.pyc",
-            "**/__pycache__/*.pyc.*",  # During pyc creation, temp files named *.pyc.NNN are created
             "**/__pycache__/*.pyo",
         ]
 


aignas:
Not sure why we are moving some of the files in the ignore_root_user_error but not all. Could you please add more context to the description of the PR please?

rickeylev:
PR #1266 only added the `*.pyc.*` ignore to the glob. The other excludes, such as `*.pyc` (no trailing star) were already present. So this is just moving that one line affected. I updated the description to better describe its just the NNNN pattern from #1266 being moved.

I'm not entirely sure as to why the pycs are only excluded as part of the ignore_root_user_error condition. It looks like it originates in https://github.com/bazelbuild/rules_python/pull/907. I can't tell why it is behind that guard, though. The conversation seems to go in the direction that it _shouldn't_ be behind that guard, but then it ended up being commited behind the guard (which a follow up https://github.com/bazelbuild/rules_python/pull/1038 fixing a typo in the original PR).

Perhaps python-build-standalone changed something so the pyc files aren't invalidated anymore? Though I guess if that was true, there wouldn't be any NNNN files being created. Also, this pyc.NNNN failure seems to only show up on windows now; so perhaps this is a no-op for linux, but something about python-built-standalone on Windows still needs it.

aignas (approved):

aignas:
I think the reason it shows up on windows is that on linux when the runners are running as non root, the folders containing the interpreter are made read-only and then there is no way for extrafiles to appear. That is different for Windows, so maybe the actual fix should be different, because I would suspect pyc files ending up in the glob on Windows builds if pyc.NNN is a problem.

PR #1549, opened 2023-11-09
fix(bzlmod pip): ensure that sub-modules do not have invalid repos

### LINKED ISSUE #1548: `bzlmod` `pip.parse` default is leaking across to the `other_module`
In the [bzlmod example](https://github.com/bazelbuild/rules_python/blob/main/examples/bzlmod/MODULE.bazel), if you examine the `@other_module_pip//absl_py:BUILD.bazel` contents, you will see:
```console
$ cat bazel-bzlmod/external/rules_python~override~pip~other_module_pip/absl_py/BUILD.bazel

alias(
    name = "absl_py",
    actual = ":pkg",
)

alias(
    name = "pkg",
    actual = select(
        {
            "@@rules_python~override//python/config_settings:is_python_3.11.6": "@other_module_pip_311_absl_py//:pkg",
            "//conditions:default": "@other_module_pip_39_absl_py//:pkg",
        },
    ),
)

alias(
    name = "whl",
    actual = select(
        {
            "@@rules_python~override//python/config_settings:is_python_3.11.6": "@other_module_pip_311_absl_py//:whl",
            "//conditions:default": "@other_module_pip_39_absl_py//:whl",
        },
    ),
)

alias(
    name = "data",
    actual = select(
        {
            "@@rules_python~override//python/config_settings:is_python_3.11.6": "@other_module_pip_311_absl_py//:data",
            "//conditions:default": "@other_module_pip_39_absl_py//:data",
        },
    ),
)

alias(
    name = "dist_info",
    actual = select(
        {
            "@@rules_python~override//python/config_settings:is_python_3.11.6": "@other_module_pip_311_absl_py//:dist_info",
            "//conditions:default": "@other_module_pip_39_absl_py//:dist_info",
        },
    ),
)%
```

However, the `other_module_pip` does not have anything set up for `3.9` and is setting `3.11` as the default toolchain version for the sub-module. It seems that [this line](https://github.com/bazelbuild/rules_python/blob/main/python/private/bzlmod/pip.bzl#L305) may be wrong and we need to get the default version per module that the extension is used in.

aignas:
This fixes the cases where the 'default_version' is passed to the
'render_pkg_aliases' utility but the 'default_version' is not present
for the wheels. This usually happens when a sub-module is using the
'pip.parse' extension and the default_version can only be set by the
root module.

Previously, such a case would generate a `select()` expression that mapped
the default condition to a non-existent target (because the sub-module didn't
call `pip.parse()` with that version). This would either result in errors
due the target not existing, or silently using a target intended for a
different Python version (which may work, but isn't correct to so).

Now, it results in a error via `select.no_match_error`.

Fixes #1548.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,6 +78,10 @@ Breaking changes:
 * (gazelle) Generate a single `py_test` target when `gazelle:python_generation_mode project`
   is used.
 
+* (bzlmod) sub-modules now don't have the `//conditions:default` clause in the
+  hub repos created by `pip.parse`. This should fix confusing error messages
+  in case there is a misconfiguration of toolchains or a bug in `rules_python`.
+
 ### Added
 
 * (bzlmod) Added `.whl` patching support via `patches` and `patch_strip`
--- a/python/private/render_pkg_aliases.bzl
+++ b/python/private/render_pkg_aliases.bzl
@@ -81,7 +81,7 @@ def _render_whl_library_alias(
         )
         selects[condition] = actual
 
-    if default_version and default_version in versions:
+    if default_version:
         no_match_error = None
         default_actual = "@{repo_name}_{version}_{dep}//:{target}".format(
             repo_name = repo_name,
@@ -109,7 +109,11 @@ def _render_common_aliases(repo_name, name, versions = None, default_version = N
     if versions:
         versions = sorted(versions)
 
-    if versions and (not default_version or default_version not in versions):
+    if not versions:
+        pass
+    elif default_version in versions:
+        pass
+    else:
         error_msg = NO_MATCH_ERROR_MESSAGE_TEMPLATE.format(
             supported_versions = ", ".join(versions),
             rules_python = rules_python,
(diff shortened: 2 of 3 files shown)


rickeylev:
An edge case I don't think is being handled: when the default python version is the sub-module's version. Recall that the version-unaware rules (//python:py_binary.bzl) won't match the `:is_py_X.Y` config conditions. So given something like:

```
# root/MODULE.bazel
python.toolchain(python_version="3.9", is_default=True)
# submodule/MODULE.bazel
pip.parse(python_version="3.9", ...)
# submodule/BUILD.bazel
load("@rules_python//python:py_binary.bzl", "py_binary")
py_binary(name="subbin", ...) 

# generate foo/BUILD.bzl

alias(name = "foo", actual = select({
  ":is_py_3.9": "pip_39_foo//:pkg",
  no_match = "error message"
}))
```

Then it'll result in an error when it should work.

So the logic needs to be something like...

```
if default in python_version_to_pip_target:
  actuals["//conditions:default"] = python_version_to_pip_target[default]
```

aignas:
@rickeylev, I think the scenario you are thinking about is covered by https://github.com/bazelbuild/rules_python/pull/1549/files#diff-6b599a1257ef3562418f2372fe871a2a18bbd09517414d6085c87f7f9d994327L75 and it is as you would expect.

rickeylev (approved):

PR #1553, opened 2023-11-10
fix: py_proto_library: append to PYTHONPATH less

### LINKED ISSUE #1551: py_proto_library prepends too aggressively to the PYTHONPATH
# 🐞 bug report

### Affected Rule

py_proto_library

### Is this a regression?

Yes, this is a regression due to #1406. (The failing provided below passes if you check out `rules_python` from the previous commit.)

### Description

#1406 added support for generating Python code for proto libraries with `strip_import_prefix` attributes. This is great! Our project (http://pigweed.dev) uses such attributes (to my chargin, but they're here to stay), so I'm excited for it.

But the solution adopted in #1406 involves quite aggressive prepending to the PYTHONPATH. In particular, I find that if a generated Python proto library is added to the transitive dependencies of a target, then the repository root, and the roots of external repositories like `com_google_protobuf`, get added close to the top of the PYTHONPATH, ahead of paths coming from `import` attributes. This can lead to module name collisions at import time.

The description above is a little hard to parse, so I created a small repro (see below).

## 🔬 Minimal Reproduction

https://github.com/tpudlik/rules_python_path/

## 🔥 Exception or Error

Depending on whether you're importing a module or an object from a module, you'll hit a `ModuleNotFoundError` or an `ImportError`.

## 🌍  Your Environment

**Operating System:**

Linux

<pre>
  <code>
Build label: 6.4.0
Build time: Thu Oct 19 17:07:43 2023 (1697735263)
Build timestamp: 1697735263
Build timestamp as int: 1697735263
  </code>
</pre>

**Rules_python version:**

<pre>
  <code>
bee35ef2abb0a1b59123500528c7d4ca0cd8a688
  </code>
</pre>

**Anything else relevant?**

https://issues.pigweed.dev/issues/266950138 tracks our migration to the `rules_python` version of `py_proto_library`.

tpudlik:
Only append the `[proto_root]` to PYTHONPATH when it's actually necessary, i.e. when the generated Python modules are under _virtual_imports. Do no append it in the general case, when `[proto_root]` is just the repository root.

This makes it much less likely that adding a transitive dep on a `py_proto_library` will reorder the `PYTHONPATH`. Such reordering is undesirable because it may lead to import errors.

Fixes #1551

### DIFF
--- a/python/private/proto/py_proto_library.bzl
+++ b/python/private/proto/py_proto_library.bzl
@@ -107,10 +107,13 @@ def _py_proto_aspect_impl(target, ctx):
     return [
         _PyProtoInfo(
             imports = depset(
-                # Adding to PYTHONPATH so the generated modules can be imported.
-                # This is necessary when there is strip_import_prefix, the Python
-                # modules are generated under _virtual_imports.
-                [proto_root],
+                # Adding to PYTHONPATH so the generated modules can be
+                # imported.  This is necessary when there is
+                # strip_import_prefix, the Python modules are generated under
+                # _virtual_imports. But it's undesirable otherwise, because it
+                # will put the repo root at the top of the PYTHONPATH, ahead of
+                # directories added through `imports` attributes.
+                [proto_root] if "_virtual_imports" in proto_root else [],
                 transitive = [dep[PyInfo].imports for dep in api_deps],
             ),
             runfiles_from_proto_deps = runfiles_from_proto_deps,


aignas (commented):
LGTM, but I would wait for @rickeylev and/or @linzhp to also chime in.

Could you also please add a `CHANGELOG.md` entry and make the PR description be more like a commit message (which it will become)?

linzhp (approved):

tpudlik:
I updated the PR description.

Re `CHANGELOG.md`: the bug this fixes was never released, so is a changelog entry still desirable?

aignas (approved):

PR #1555, opened 2023-11-13
feat(config_settings): allow matching minor version of python_version flag

dizzy57:
Currently a user has to specify a full (x.y.z) version of Python when setting the `//python/config_settings:python_version` flag. When they upgrade `rules_python` or change `MINOR_MAPPING` in some other way, user has to update the flag's value to keep it in sync with `MINOR_MAPPING`.

This adds micro-version agnostic config settings to allow matching the minor version.
For example e.g. `//python/config_settings:is_python_3.8` will match any of
`3.8.1, 3.8.2, ...' (or whatever other versions are listed in `TOOL_VERSIONS`)


python/config_settings/config_settings.bzl
```diff
@@ -15,26 +15,52 @@
...
+        minor_mapping: mapping from minor (x.y) versions to the corresponding full (x.y.z) versions.
+        python_versions: list of full (x.y.z) Python versions supported by rules_python.
     """
+
+    minor_versions = list(minor_mapping.keys())
+    allowed_flag_values = python_versions + minor_versions
+
     string_flag(
         name = "python_version",
         build_setting_default = python_versions[0],
```
dizzy57, line 35: Using `build_setting_default = python_versions[0]` here looks like a minor bug, BTW.

It enables a config setting, and if user manages to configure a toolchain that matches this full version, this toolchain will be selected by default, instead of the one marked by the `is_default` attr in `python.toolchain(...)`.

I think a better approach would be to have something like `build_setting_default = "default"` and `values = ["default"] + allowed_flag_values`.

dizzy57:
>### Could you please explain how this would be used?

Currently users can switch python toolchains by passing `--@rules_python//python/config_settings:python_version=3.9.18` flag. See https://github.com/bazelbuild/rules_python/pull/846#issuecomment-1302476859 for the design.
This flag allows users, for example, to run existing `py_test`s with a non-default toolchain to help evaluating interpreter version upgrade. Another use for this flag might be running a [matrix testing strategy](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs) in CI (testing on a multitude of OS + interpreter version combinations).
Anyway, the flag is public, so by Hyrum's law, some users (including me) are using it for toolchain selection.

Allowed values for this flag are `TOOL_VERSIONS.keys()`, e.g. full (x.y.z) versions. But `rules_python` uses minor (x.y) versions in most places, e.g. when calling `python.toolchain` and `pip.parse`: 
https://github.com/bazelbuild/rules_python/blob/a9032d2c241d3e5ff104e1d66f56148107b33af1/examples/multi_python_versions/MODULE.bazel#L13-L16
https://github.com/bazelbuild/rules_python/blob/a9032d2c241d3e5ff104e1d66f56148107b33af1/examples/multi_python_versions/MODULE.bazel#L38-L42
The latter `pip.parse` call leads to generating a select for every wheel:
https://github.com/bazelbuild/rules_python/blob/a9032d2c241d3e5ff104e1d66f56148107b33af1/python/pip.bzl#L188-L198
To land in a `3.9` branch in this select, you need to specify `3.9.18` as the flag's value, and when `MINOR_MAPPING` inevitably get updated, `pip.parse` call will renter the template with a different `full_python_version`, and after that moment passing `--@rules_python//python/config_settings:python_version=3.9.18` will lead to selecting the **default** toolchain (that might not even be 3.9), leading to confusion.

This PR allows users to specify `--@rules_python//python/config_settings:python_version=3.9`, and this flag value will select the correct 3.9.x toolchain behind the scenes.

>### What implications for users?
Flag starts to accept a new set of values in addition to previously accepted ones.
New values make flag-based toolchain selection more user-friendly: users that start using new values will not need to fix the flag value in their scripts when `rules_python` maintainers change `MINOR_MAPPING`.

>### I see all of the new targets are `public`, do they have to be?
I'm not expert here, but I got an error when I made `:python_version_flag_equals_Y` visible to `__pkg__` only.
Looks like since `config_setting_group(name = X, match_any = [Y])` is expanded to `alias(name = X, actual = Y)` in case of one-element `match_any`, if we want a public `:is_python_X` to alias `:python_version_flag_equals_Y` the latter should be public as well.

aignas:
Thanks for the explanation, this is the first time I've heard you can switch the registered python like this and this is really interesting. And the Hyrum's law mention make sense in this case, just wanted to understand the Y in the [XY](https://xyproblem.info/).

If I understand correctly, what you are after is being able to select the version using `--@rules_python//python/config_settings:python_version=3.9.X` and then have the `select` statement in the `pip` rule repository work correctly and select a wheel for some `3.9.X` version that is registered. Please correct me if I got it wrong. If I haven't I do think that we may need to look at the `pip` repository select statements and how they work with the registered toolchains and this usecase and maybe the fix should be done in a different way.

rickeylev:
Thanks for this PR. This topic has come up in the back of my mind a few times, but I haven't had time to think on it. I, too, want to allow specifying and matching the version to work with both minor and patch levels specified. Libraries matching every patch version doesn't scale, for example.

> This flag allows ... evaluating interpreter version upgrade. ... matrix testing strategy in CI

Yes! This is _exactly_ the intent of these flags :smile:. If you have remote execution, you can even take it a step further by putting a transition around the targets to force them to particular platforms or versions and thus test _everything_ in a single build invocation. It's pretty nice when it's all put together.

rickeylev (changes requested):

aignas:
@dizzy57, gentle ping. Is this something you would like to get merged? Having a config setting for Python `3.8` which is compatible would be something that would be very useful!

dizzy57:
@aignas Thanks for the ping, I plan to get back to this diff next week

python/config_settings/config_settings.bzl
rickeylev: Added a comment to note this

python/config_settings/config_settings.bzl
```diff
@@ -15,26 +15,52 @@
...
-        values = python_versions,
+        values = allowed_flag_values,
         visibility = ["//visibility:public"],
     )
 
-    for python_version in python_versions:
-        python_version_constraint_setting = "is_python_" + python_version
+    for flag_value in allowed_flag_values:
+        flag_value_constraint_setting = "python_version_flag_equals_" + flag_value
```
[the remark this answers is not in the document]
rickeylev: Done

python/config_settings/config_settings.bzl
```diff
@@ -15,26 +15,52 @@
...
+    flag_values_that_enable_version = {
+        full_version: [full_version]
+        for full_version in python_versions
+    }
+
+    for minor_version, full_version in minor_mapping.items():
+        flag_values_that_enable_version[full_version].append(minor_version)
+
+    for full_version, flag_values in flag_values_that_enable_version.items():
+        python_version_constraint_setting = "is_python_" + full_version
```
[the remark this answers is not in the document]
rickeylev: Done

rickeylev (commented):
I had some time, so implemented the missing parts

rickeylev:
I modified this a bit from the original implementation, but I think the intent is the same. The flag value still requires a micro-level version (3.10.1), but an additional config setting is exposed to match a minor-version (3.10) against any of the (known) micro versions. This allows omitting the micro-version is most cases and switching between e.g. `--//python/config_settings:python_version=3.8.1` to 3.8.2 won't require updating the select conditions. For example, one can write this:

```
alias(actual = select({
  "//python/config_settings:is_python_3.8": ...
  })
)
```

Instead of having to enumerate all the 3.8.x values explicitly, or having to map the minor version back to the full version (as we do in the pip build file generation code).

rickeylev (approved):

python/config_settings/config_settings.bzl
```diff
@@ -15,26 +15,58 @@
...
         visibility = ["//visibility:public"],
     )
 
-    for python_version in python_versions:
-        python_version_constraint_setting = "is_python_" + python_version
+    for minor_version, micro_versions in minor_to_micro_versions.items():
+        # This matches the raw flag value, e.g. --///python/config_settings:python_version=3.8
+        # It's private because matching the concept of e.g. "3.8" value is done
+        # using the `is_python_X.Y` config setting group, which is aware of the
+        # minor versions that could match instead.
```
aignas, line 50: Thank you for the explanation.

python/config_settings/config_settings.bzl
```diff
@@ -15,26 +15,58 @@
...
+            is_micro_version_name = "is_python_" + micro_version
+            native.config_setting(
+                name = is_micro_version_name,
+                flag_values = {":python_version": micro_version},
+                visibility = ["//visibility:public"],
+            )
+            matches_minor_version_names.append(is_micro_version_name)
+
+        selects.config_setting_group(
+            name = "is_python_" + minor_version,
```
aignas, line 69: @rickeylev, this will create the following targets:
```
//python/config_settings:is_python_3.10
//python/config_settings:is_python_3.10_2
//python/config_settings:is_python_3.10_3
//python/config_settings:is_python_3.10_4
//python/config_settings:is_python_3.10_5
//python/config_settings:is_python_3.10_6
//python/config_settings:is_python_3.10_7
//python/config_settings:is_python_3.10_8
```

and it is super confusing when looking at them through `bazel query`. I know that not many people might look, but having this config_setting_group with `_` prefix and then having a `native.alias` with `is_python_` prefix would work around this issue.

aignas (commented):
Adding a few comments on implementation for @rickeylev.

python/config_settings/config_settings.bzl
rickeylev: Good catch. Yes, that is confusing, especially since the micro-version names are an `.N` prefix. Added the aliased config group, as suggested.

Issue #735, opened 2022-06-23
Allow pip_repository (and pip_parse) to "select" requirements based on both OS and architecture


jiawen:
# 🚀 feature request

### Relevant Rules

`pip_repository` and by extension `pip_parse`.

### Description

Because `pip_repository` is a repository rule, it somewhat awkwardly uses several attributes to override the default requirements file based exclusively on a limited selection of operating systems (linux, macos, windows).

Now that ARM is a more popular architecture and available on all three operating systems, we need a way to select among that axis as well. Otherwise, some wheels that have native dependencies may not be available on certain architectures. We may want to consider an "os version" as a third axis as well. For example, [rawpy](https://pypi.org/project/rawpy/0.17.0/#files) is available on MacOS but only on Intel, and possibly only MacOS 10.9 (I don't have a machine to test).

### Describe the solution you'd like

Because it's awkward to add all possible combinations as new attributes, perhaps we should consider passing in a dictionary of tuples to match against fields in [repository_os](https://bazel.build/rules/lib/repository_os). The [current implementation](https://github.com/bazelbuild/rules_python/blob/ae7a2677b3003b13d45bc9bfc25f1425bed5b407/python/pip_install/pip_repository.bzl#L125) uses only `.name`. It's easy to add `.arch` to match against arch. It might be a little tricky to add "OS version" as it requires either adding a new field to `repository_os` (probably possible), or invoking a platform-dependent shell script.

### Describe alternatives you've considered

Passing in a struct? It's more or less the same as a dictionary.

jiawen:
@likesum FYI.

yk:
bump. generated requirements differ between architectures, thus making version control of e.g. lock files very difficult

arrdem:
In order to solve this problem we settled on using a workspace rule we call `python_config`, which is able to use the `rctx` to inspect the building host and template out a `config.bzl` file containing values such as a computed path to a `requirements_lock.txt` chosen based on the building host. This allows us to check in many platform+virtualenv combination lockfiles and pick "the right one" transparently within Bazel without modification to `rules_python`.

Our `WORKSPACE` then looks something like this -

```
python_version = "3.8"

load("//tools/rules:external_python.bzl", "external_python")

external_python(
    name = "local_py",
    version = python_version,
)

register_toolchains("@local_py//:python_toolchain")

# Refer to this workspace rule and associated template for how we're choosing
# the Python configuration.
load("//tools/rules:python_config.bzl", "python_config")

[...]

PR #1564, opened 2023-11-16
feat(whl_library): set target_compatible_with based on whl name

aignas:
This adds a pure starlark function to further parse the platform_tag from the
wheel and set the correct attributes for the target generated by the whl_library
rule. This should make errors much more easy to understand when users try to
build a docker image using incompatible libraries.

This is currently hidden under an experimental flag described in the changelog.

Related #735, #260


aignas:
This is not the direction we want to follow in the future, because `target_compatible_with` does not help with toolchain resolution and we would be looking to something like #1625 instead.

PR #1567, opened 2023-11-19
chore!: remove incompatible_generate_aliases and related flags

aignas:
According to our breaking change policy, we are removing the flag
since in the previous release it has been flipped by default and
enabled to all users in 0.27.0.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -120,7 +120,6 @@ pip_parse(
             "sphinxcontrib-applehelp",
         ],
     },
-    incompatible_generate_aliases = True,
     python_interpreter_target = interpreter,
     requirements_lock = "//docs/sphinx:requirements.txt",
 )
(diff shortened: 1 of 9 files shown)


python/private/py_console_script_binary.bzl
```diff
@@ -27,8 +27,7 @@ def _dist_info(pkg):
     rules_python does not know anything about the hub repos that the user has
     available.
 
-    NOTE: Works with `incompatible_generate_aliases` and without by assuming the
-    following formats:
+    NOTE: Works with assuming the following label formats:
```
rickeylev, line 30: delete L31: `@pypi_pylint//:pkg` (nothing generates that format anymore, correct?)

rickeylev (approved):

WORKSPACE
```diff
@@ -120,7 +120,7 @@ pip_parse(
             "sphinxcontrib-applehelp",
         ],
     },
-    incompatible_generate_aliases = True,
+    name = "docs_deps",
```
aignas, line 123:
```suggestion

```

PR #1569, opened 2023-11-20
chore/fix!: remove incompatible whl packaging flags

aignas:
Remove incompatible `whl` packaging flags that had been flipped
in `0.27.0` release.

Also fixes a bug where the the build stamp values (e.g. build username) were lowercased
when they shouldn't be.

### DIFF
--- a/examples/wheel/wheel_test.py
+++ b/examples/wheel/wheel_test.py
@@ -192,44 +192,6 @@ second = second.main:s""",
             filename, "48eed93258bba0bb366c879b77917d947267d89e7e60005d1766d844fb909118"
         )
 
-    def test_legacy_filename_escaping(self):
-        filename = self._get_path(
-            "file_name_escaping-0.0.1_r7-py3-none-any.whl",
-        )
-        with zipfile.ZipFile(filename) as zf:
-            self.assertAllEntriesHasReproducibleMetadata(zf)
-            self.assertEqual(
-                zf.namelist(),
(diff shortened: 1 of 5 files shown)


aignas:
FYI @vonschultz

examples/wheel/wheel_test.py
```diff
@@ -450,15 +412,15 @@ def test_rule_creates_directory_and_is_included_in_wheel(self):
...
+            "example_minimal_library{BUILD_USER}-0.1.{BUILD_TIMESTAMP}-py3-none-any.whl"
         )
 
         with zipfile.ZipFile(filename) as zf:
             self.assertAllEntriesHasReproducibleMetadata(zf)
             metadata_file = None
             for f in zf.namelist():
-                self.assertNotIn("_BUILD_TIMESTAMP_", f)
-                self.assertNotIn("_BUILD_USER_", f)
+                self.assertNotIn("{BUILD_TIMESTAMP}", f)
```
rickeylev, line 422: I'm guessing the lowercase bug went undetected because this just looked for "not in" vs a more positive check about the value? Can we somehow check the values to verify the bug is gone?

aignas: The filename in the line above is probably the assertion you seek for. The bug went undetected because we did not have a copy of this test with the new feature enabled.

rickeylev (approved):

PR #1572, opened 2023-11-20
feat(twine): support 'bzlmod' users out of the box

### LINKED ISSUE #1369: py_wheel support in bzlmod
We don't currently have any tests verifying py_wheel works under bzlmod. I don't think any of us have tried it out ourselves, either.

It should just work. I don't recall it doing anything particularly special, but you never know.

So, at the least, manually verify it works and post the result here. Ideally, add some tests to verify it actually works.

aignas:
Implements a test that starts a [`pypiserver`] and checks
that the publishing with the new machinery still works.

Fixes #1369

[pypiserver]: https://github.com/pypiserver/pypiserver


python/packaging.bzl
```diff
@@ -158,14 +162,28 @@ def py_wheel(name, twine = None, publish_args = [], **kwargs):
...
-        if not twine.endswith(":pkg"):
-            fail("twine label should look like @my_twine_repo//:pkg")
-        twine_main = twine.replace(":pkg", ":rules_python_wheel_entry_point_twine.py")
+    twine_args = []
+    if twine or BZLMOD_ENABLED:
         twine_args = ["upload"]
         twine_args.extend(publish_args)
         twine_args.append("$(rootpath :{})/*".format(_dist_target))
 
+    if BZLMOD_ENABLED or str(Label(twine)) == str(Label("//tools/publish:twine")):
```
rickeylev, line 171: Maybe it should be a separate arg? `twine_binary` instead of `twine`. This would allow passing an arbitrary binary, so long as it supported the CLI args.

I guess in retrospect, `twine_lib` would have been a better name than `twine`. Ah well.

tools/publish/BUILD.bazel
```diff
@@ -1,8 +1,15 @@
...
+load("//python/entry_points:py_console_script_binary.bzl", "py_console_script_binary")
 
 compile_pip_requirements(
     name = "requirements",
     src = "requirements.in",
     requirements_darwin = "requirements_darwin.txt",
     requirements_windows = "requirements_windows.txt",
 )
+
+py_console_script_binary(
```
rickeylev, line 11: We should use the version-aware rule to make sure this uses the python version the deps were built for.

rickeylev:
> If this is something we may want to do

Yes, +1 on making it work out of the box. As a user, I don't really want to care about twine. I just want it to be easy to publish to pypi.

> , I could add a test checking that we can:
> 
> 1. Start a simple PyPI-like registry.
> 2. Publish our `runfiles` package there.
> 3. (optional) Download it from there.

Is this something we could generalize? We have quite a few tests that rely on external packages having particular structures (e.g. circular deps, native deps, extra, etc). If we had a "local" pypi registry, we'd be able to more easily define and test the relevant scenarios.

python/packaging.bzl
```diff
@@ -158,14 +162,28 @@ def py_wheel(name, twine = None, publish_args = [], **kwargs):
...
+        native_binary(
+            name = "{}.publish".format(name),
+            src = Label("//tools/publish:twine"),
+            out = "{}.publish",
```
rbtcollins, line 175:
```suggestion
            out = "{}.publish-script".format(name),
```

perhaps?

aignas:
> Is this something we could generalize? We have quite a few tests that rely on external packages having particular structures (e.g. circular deps, native deps, extra, etc). If we had a "local" pypi registry, we'd be able to more easily define and test the relevant scenarios.

It depends on how we want the testing to be done. If we want manual tests, then starting a script to launch a server and then running a test script manually would be doable.

If we want to integrate with the rest of the tooling so that it is automatic, it may be harder, but these are some ideas:
1. We could have an integration test with a custom runner for bazel-in-bazel tests where the runner is written in python and starts a `pypiserver` just before running the bazel commands.
2. Local iterations could be facilitated with running the server manually.
3. We could potentially create a build action that creates the correct layout with the correct wheels made with py_wheel rule.

I am not sure I am going to pursue this academic exercise, so if anyone finds this message and is interested, feel free :)

aignas:
Not sure what to do with the failing Windows tests. Any ideas?

rickeylev (approved):

PR #1573, opened 2023-11-21
pystar: support builtin providers for compatibility

rickeylev:
This makes the rules_python Starlark implementation accept and return the builtin providers. This allows depending on, and being depended on by, the builtin rules, which enables the two rule sets to interoperate better.

Work towards #1069

### DIFF
--- a/python/private/common/common.bzl
+++ b/python/private/common/common.bzl
@@ -13,9 +13,9 @@
 # limitations under the License.
 """Various things common to Bazel and Google rule implementations."""
 
+load("//python/private:reexports.bzl", "BuiltinPyInfo")
 load(":cc_helper.bzl", "cc_helper")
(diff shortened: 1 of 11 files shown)


python/private/common/common.bzl
```diff
@@ -15,6 +15,7 @@
 
 load(":cc_helper.bzl", "cc_helper")
 load(":providers.bzl", "PyInfo")
+load("//python/private:reexports.bzl", "BuiltinPyInfo")
```
Solumin, line 18: Looks like the tests are failing due to this line.

rickeylev: Fixed. The exports.bzl dep was missing.

Solumin (approved):

PR #1574, opened 2023-11-21
feat: use rules_python implemented py_runtime, py_runtime_pair, PyRuntimeInfo

rickeylev:
This switches over to using the rules_python implementation of `py_runtime`,
`py_runtime_pair`, and `PyRuntimeInfo` for Bazel 6 and higher. Bazel 5 lacks features
(specifically provider ctors) to allow enabling it on that version.

This is possible because the rules don't directly use the PyRuntimeInfo provider
(mostly, see below), they only care about the structure of it as exposed from the
ToolchainInfo provider.

Switching the toolchain providers and rules over early allows some development of
the toolchain prior to Bazel 7 and the rest of the rules_python Starlark implementation
being enabled.

The builtin PyRuntimeInfo is still returned and accepted for two reasons:
* Better compatibility with the builtin rules to make transitioning easier
* `py_binary` has an old, possibly defunct (not sure) code path that will look up the
  the PyRuntimeInfo from a flag/implicit attribute.

### DIFF
--- a/python/private/common/providers.bzl
+++ b/python/private/common/providers.bzl
@@ -13,14 +13,15 @@
 # limitations under the License.
 """Providers for Python rules."""
 
-load("@rules_python_internal//:rules_python_config.bzl", "config")
+load("//python/private:util.bzl", "IS_BAZEL_6_OR_HIGHER")
 
 # TODO: load CcInfo from rules_cc
 _CcInfo = CcInfo
 
 DEFAULT_STUB_SHEBANG = "#!/usr/bin/env python3"
 
-DEFAULT_BOOTSTRAP_TEMPLATE = "@bazel_tools//tools/python:python_bootstrap_template.txt"
+DEFAULT_BOOTSTRAP_TEMPLATE = Label("//python/private:python_bootstrap_template.txt")
+
 _PYTHON_VERSION_VALUES = ["PY2", "PY3"]
 
 # Helper to make the provider definitions not crash under Bazel 5.4:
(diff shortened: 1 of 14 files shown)


python/private/common/providers.bzl
```diff
@@ -13,14 +13,15 @@
 # limitations under the License.
 """Providers for Python rules."""
 
-load("@rules_python_internal//:rules_python_config.bzl", "config")
+load("//python/private:util.bzl", "IS_BAZEL_6_OR_HIGHER")
```
aignas, line 16: Nit, what about using bazel_features package for this? I know that this adds an extra dep, but we are already using that on with bzlmod setups.

python/private/common/providers.bzl
```diff
@@ -13,14 +13,15 @@
...
 
 # TODO: load CcInfo from rules_cc
 _CcInfo = CcInfo
 
 DEFAULT_STUB_SHEBANG = "#!/usr/bin/env python3"
 
-DEFAULT_BOOTSTRAP_TEMPLATE = "@bazel_tools//tools/python:python_bootstrap_template.txt"
+DEFAULT_BOOTSTRAP_TEMPLATE = Label("//python/private:python_bootstrap_template.txt")
```
aignas, line 23: Does this mean we can have a custom template on bazel 6? I think that no, but I am not sure if this was just a fixup of previously broken code due to migration.

python/private/common/py_runtime_rule.bzl
```diff
@@ -189,8 +195,8 @@ For a platform runtime, this is the absolute path of a Python interpreter on
 the target platform. For an in-build runtime this attribute must not be set.
 """),
         "python_version": attr.string(
-            default = "_INTERNAL_SENTINEL",
-            values = ["PY2", "PY3", "_INTERNAL_SENTINEL"],
+            default = "PY3",
+            values = ["PY2", "PY3"],
```
aignas, line 199: Nit: I thought we are not supporting PY2?

aignas (approved):

PR #1586, opened 2023-11-30
docs: update docs deps to latest versions

rickeylev:
The latest Sphinx version has several circular dependencies, but those can be handled with the requirements_cycles feature.


aignas (commented):
I think the Module.bazel and othe platform dependent requirements files need updating.

rickeylev:
Module.bazel and requirements darwin updated.

Macs are still disabled in CI, so I hope the requirements are correct. I don't have a mac myself, but had a coworker regenerate it.

aignas (approved):

PR #1590, opened 2023-12-02
fix(gazelle): ensure that gazelle helper modules are on PYTHONPATH

### LINKED ISSUE #1589: gazelle plugin is not working with Python 3.11 in 0.27.0 release
# 🐞 bug report

### Affected Rule

`@rules_python_gazelle_plugin//python` is broken when used in `gazelle_binary` with the latest release when using Python 3.11 as the default toolchain.

### Is this a regression?

Yes, introduced in #1455.

### Description

See https://bazelbuild.slack.com/archives/CA306CEV6/p1700496408567889

## 🔬 Minimal Reproduction

See https://github.com/aignas/rules_python_repro_gazelle_binary_rule

## 🔥 Exception or Error

```
  File "/tmp/Bazel.runfiles_tugoo1fs/runfiles/rules_python_gazelle_plugin/python/__main__.py", line 19, in <module>
    import parse
ModuleNotFoundError: No module named 'parse'
  File "/var/folders/rl/v9dqv4mx1jsgwx81qjy74scc0000gn/T/Bazel.runfiles_o1h0_q7c/runfiles/rules_python_gazelle_plugin/python/__main__.py", line 19, in <module>
    import parse
ModuleNotFoundError: No module named 'parse'
gazelle: ERROR: failed to parse: EOF
```

aignas:
Before this change there was a bug in how the parsing helpers were being
used in case we were using Python 3.11 toolchain, which is using a more
strict version of the entrypoint template. This change adds `imports = ["."]`
to ensure that the gazelle helper components are on PYTHONPATH and updates
the non-bzlmod tests to run under 3.11.

We also:
* Change `.bazelrc` to use explicit `__init__.py` definition to avoid
  non-reproducible errors in the future.
* Add a dedicated `gazelle_binary` that uses `DEFAULT_LANGUAGES` *and*
  `//python`.

Fixes #1589


aignas:
@rickeylev, what do you think about cherry-picking this into a `0.27.1` release?

gazelle/WORKSPACE
```diff
@@ -13,10 +13,10 @@ http_archive(
 
 http_archive(
     name = "bazel_gazelle",
-    sha256 = "29d5dafc2a5582995488c6735115d1d366fcd6a0fc2e2a153f02988706349825",
+    sha256 = "d3fa66a39028e97d76f9e2db8f1b0c11c099e8e01bf363a923074784e451f809",
     urls = [
-        "https://mirror.bazel.build/github.com/bazelbuild/bazel-gazelle/releases/download/v0.31.0/bazel-gazelle-v0.31.0.tar.gz",
-        "https://github.com/bazelbuild/bazel-gazelle/releases/download/v0.31.0/bazel-gazelle-v0.31.0.tar.gz",
+        "https://mirror.bazel.build/github.com/bazelbuild/bazel-gazelle/releases/download/v0.33.0/bazel-gazelle-v0.33.0.tar.gz",
+        "https://github.com/bazelbuild/bazel-gazelle/releases/download/v0.33.0/bazel-gazelle-v0.33.0.tar.gz",
```
rickeylev, line 19: Is .33 a new minimum version requirement? If so, please note that in the changelog, too

gazelle/.bazelrc
```diff
@@ -11,3 +11,11 @@ build --incompatible_default_to_explicit_init_py
...
 build --enable_runfiles
 startup --windows_enable_symlinks
+
+# Do NOT implicitly create empty __init__.py files in the runfiles tree.
+# By default, these are created in every directory containing Python source code
+# or shared libraries, and every parent directory of those directories,
+# excluding the repo root directory. With this flag set, we are responsible for
+# creating (possibly empty) __init__.py files and adding them to the srcs of
+# Python targets as required.
+build --incompatible_default_to_explicit_init_py
```
rickeylev, line 21: Shouldn't there be an `__init__.py` file in the rules_python_gazelle_helper directory?

gazelle/rules_python_gazelle_helper/__main__.py
```diff
@@ -16,10 +16,16 @@
 # STDIN receives parse requests, one per line. It outputs the parsed modules and
 # comments from all the files from each request.
 
-import parse
-import std_modules
 import sys
 
+# NOTE @aignas 2023-12-02: Use absolute imports with respect to WORKSPACE root.
```
rickeylev, line 21: It would be better to not rely on the repo root being added to sys.path -- that's a behavior I plan to disable once the Starlark impl can be enabled.

But, I can understand this is just a bugfix CL, so this is fine for now. Is Gazelle able to under PyInfo.imports, or have some equivalent?

rickeylev (approved):
>  cherry-picking this into a 0.27.1 release?

+1, I agree

gazelle/WORKSPACE
aignas: reverting, as it is not needed for this bugfix.

gazelle/.bazelrc
aignas: I have tested with and without and it did not change the behaviour, I think with recent Python versions the `__init__.py` files are optional (see [here](https://docs.python.org/3/whatsnew/3.3.html#pep-420-implicit-namespace-packages)), however, this flag allows us to have better tests.

gazelle/rules_python_gazelle_helper/__main__.py
aignas: This made me realize that I just needed to use `imports = ["."]` on the `py_binary` that is being used here to fix the error and no moving of files is actually needed. Thanks!

PR #1593, opened 2023-12-04
feat(whl_library): generate platform-specific dependency closures

### LINKED ISSUE #1591: Dependencies are not when using wheels on a different platform
# 🐞 bug report

### Affected Rule

`pip_parse`, `whl_library`.

### Is this a regression?

No.

### Description

Fetching a whl using `download_only = True` one would expect to get the
dependencies for the target platform as specified by the extra variable that
one has to pass as described in
[this bazelcon talk](https://www.youtube.com/watch?v=Bjaiu8tZZhs).
However, for `pytorch 2.0`, `linux` wheels depend on an extra wheel that is a
linux only wheel and it never gets into the dependency closure when the
`whl_library` gets setup due to
[this line](https://github.com/bazelbuild/rules_python/blob/main/python/pip_install/tools/wheel_installer/wheel.py#L80).
This means that the output of the `whl_library` will be host platform specific,
which may break in situations when building a docker image with `rules_oci` or
similar.

This behaviour is related to #260, but much smaller in scope.

## 🔬 Minimal Reproduction

Setup a `whl_library` with `click` and observe, that there is no conditional
dependency for the Windows platform.

aignas:
Before this change, the dependency closures would be influenced by the
host-python interpreter, this removes the influence by detecting the
platforms against which the `Requires-Dist` wheel metadata is evaluated.
This functionality can be enabled via `experimental_target_platforms`
attribute to the `pip.parse` extension and is showcased in the `bzlmod`
example. The same attribute is also supported on the legacy `pip_parse`
repository rule.

The detection works in the following way:
- Check if the python wheel is platform specific or cross-platform
  (i.e., ends with `any.whl`), if it is then platform-specific
  dependencies are generated, which will go through a `select`
  statement.
- If it is platform specific, then parse the platform_tag and evaluate
  the `Requires-Dist` markers assuming the target platform rather than
  the host platform.

NOTE: The `whl` `METADATA` is now being parsed using the `packaging`
Python package instead of `pkg_resources` from `setuptools`.

Fixes #1591

### DIFF
--- a/examples/bzlmod/MODULE.bazel
+++ b/examples/bzlmod/MODULE.bazel
@@ -102,7 +102,13 @@ pip.parse(
     python_version = "3.9",
     requirements_lock = "//:requirements_lock_3_9.txt",
     requirements_windows = "//:requirements_windows_3_9.txt",
-    target_platforms = ["all"],
+    # You can use one of the values below to specify the target platform
+    # to generate the dependency graph for.
+    target_platforms = [
+        "all",
+        "linux_*",
+        "host",
+    ],
     # These modifications were created above and we
     # are providing pip.parse with the label of the mod
     # and the name of the wheel.
@@ -126,7 +132,13 @@ pip.parse(
     python_version = "3.10",
     requirements_lock = "//:requirements_lock_3_10.txt",
     requirements_windows = "//:requirements_windows_3_10.txt",
-    target_platforms = ["all"],
+    # You can use one of the values below to specify the target platform
+    # to generate the dependency graph for.
+    target_platforms = [
+        "all",
+        "linux_*",
+        "host",
+    ],
     # These modifications were created above and we
     # are providing pip.parse with the label of the mod
     # and the name of the wheel.
--- a/examples/bzlmod/MODULE.bazel
+++ b/examples/bzlmod/MODULE.bazel
@@ -98,17 +98,17 @@ pip.parse(
             "sphinxcontrib-serializinghtml",
         ],
     },
-    hub_name = "pip",
-    python_version = "3.9",
-    requirements_lock = "//:requirements_lock_3_9.txt",
-    requirements_windows = "//:requirements_windows_3_9.txt",
     # You can use one of the values below to specify the target platform
     # to generate the dependency graph for.
-    target_platforms = [
+    experimental_target_platforms = [
         "all",
         "linux_*",
         "host",
     ],
+    hub_name = "pip",
+    python_version = "3.9",
+    requirements_lock = "//:requirements_lock_3_9.txt",
+    requirements_windows = "//:requirements_windows_3_9.txt",
     # These modifications were created above and we
     # are providing pip.parse with the label of the mod
     # and the name of the wheel.
@@ -128,17 +128,17 @@ pip.parse(
             "sphinxcontrib-serializinghtml",
         ],
     },
-    hub_name = "pip",
-    python_version = "3.10",
-    requirements_lock = "//:requirements_lock_3_10.txt",
-    requirements_windows = "//:requirements_windows_3_10.txt",
     # You can use one of the values below to specify the target platform
     # to generate the dependency graph for.
-    target_platforms = [
+    experimental_target_platforms = [
         "all",
         "linux_*",
         "host",
     ],
+    hub_name = "pip",
+    python_version = "3.10",
+    requirements_lock = "//:requirements_lock_3_10.txt",
+    requirements_windows = "//:requirements_windows_3_10.txt",
     # These modifications were created above and we
     # are providing pip.parse with the label of the mod
     # and the name of the wheel.
--- a/examples/bzlmod/MODULE.bazel
+++ b/examples/bzlmod/MODULE.bazel
@@ -102,6 +102,7 @@ pip.parse(
     python_version = "3.9",
     requirements_lock = "//:requirements_lock_3_9.txt",
     requirements_windows = "//:requirements_windows_3_9.txt",
+    target_platforms = ["all"],
     # These modifications were created above and we
     # are providing pip.parse with the label of the mod
     # and the name of the wheel.
@@ -125,6 +126,7 @@ pip.parse(
     python_version = "3.10",
     requirements_lock = "//:requirements_lock_3_10.txt",
     requirements_windows = "//:requirements_windows_3_10.txt",
+    target_platforms = ["all"],
     # These modifications were created above and we
     # are providing pip.parse with the label of the mod
     # and the name of the wheel.
--- a/python/pip_install/pip_repository.bzl
+++ b/python/pip_install/pip_repository.bzl
@@ -538,7 +538,7 @@ cross platform wheels by parsing the wheel metadata. This will generate the
 correct dependencies for packages like `sphinx` or `pylint`, which include
 `colorama` when installed and used on Windows platforms.
 
-An empty list means falling back to the legacy behaviour when we use the host
+An empty list means falling back to the legacy behaviour where the host
 platform is the target platform.
 
 WARNING: It may not work as expected in cases where the python interpreter
--- a/python/pip_install/pip_repository.bzl
+++ b/python/pip_install/pip_repository.bzl
@@ -35,6 +35,19 @@ COMMAND_LINE_TOOLS_PATH_SLUG = "commandlinetools"
 
 _WHEEL_ENTRY_POINT_PREFIX = "rules_python_wheel_entry_point"
 
+_ALL_PLATFORMS = [
+    "linux_aarch64",
+    "linux_ppc64le",
+    "linux_s390x",
+    "linux_x86_64",
+    "linux_x86_32",
+    "osx_aarch64",
+    "osx_x86_64",
+    "windows_x86_64",
+    "windows_x86_32",
+    "windows_aarch64",
+]
+
 def _construct_pypath(rctx):
     """Helper function to construct a PYTHONPATH.
 
@@ -345,6 +358,8 @@ def _pip_repository_impl(rctx):
 
     if rctx.attr.python_interpreter_target:
         config["python_interpreter_target"] = str(rctx.attr.python_interpreter_target)
+    if rctx.attr.target_platforms:
+        config["target_platforms"] = rctx.attr.target_platforms
 
     if rctx.attr.incompatible_generate_aliases:
         macro_tmpl = "@%s//{}:{}" % rctx.attr.name
@@ -515,6 +530,25 @@ python_interpreter. An example value: "@python3_x86_64-unknown-linux-gnu//:pytho
 Prefix for the generated packages will be of the form `@<prefix><sanitized-package-name>//...`
 """,
     ),
+    "target_platforms": attr.string_list(
+        # TODO @aignas 2023-12-04: this is not the best way right now
+        # so if this approach is generally good, this can be refactored, by
+        # using the names of the config_setting labels.
+        default = [],
+        doc = """\
+A list of platforms that we will generate the conditional dependency graph for
+cross platform wheels by parsing the wheel metadata. This will generate the
+correct dependencies for packages like `sphinx` or `pylint`, which include
+`colorama` when installed and used on Windows platforms.
+
+An empty list means falling back to the legacy behaviour when we use the host
+platform is the target platform.
+
+WARNING: It may not work as expected in cases where the python interpreter
+implementation that is being used at runtime is different between different platforms.
+This has been tested for CPython only.
+""",
+    ),
     # 600 is documented as default here: https://docs.bazel.build/versions/master/skylark/lib/repository_ctx.html#execute
     "timeout": attr.int(
         default = 600,
@@ -669,6 +703,15 @@ See the example in rules_python/examples/pip_parse_vendored.
 )
 
 def _whl_library_impl(rctx):
+    target_platforms = []
+    if len(rctx.attr.target_platforms) == 1 and rctx.attr.target_platforms[0] == "all":
+        target_platforms = _ALL_PLATFORMS
+    elif rctx.attr.target_platforms:
+        target_platforms = rctx.attr.target_platforms
+        for p in target_platforms:
+            if p not in _ALL_PLATFORMS:
+                fail("target_platforms must be a subset of {} but got {}".format(["all"] + _ALL_PLATFORMS, target_platforms))
+
     python_interpreter = _resolve_python_interpreter(rctx)
     args = [
         python_interpreter,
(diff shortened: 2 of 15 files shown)


python/private/text_util.bzl
```diff
@@ -28,18 +28,18 @@ def _render_alias(name, actual):
         ")",
     ])
 
-def _render_dict(d):
+def _render_dict(d, *, value = repr):
```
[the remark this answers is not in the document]
aignas: Thanks, using `value_repr`.

python/pip_install/private/generate_whl_library_build_bazel.bzl
```diff
@@ -101,11 +102,40 @@ alias(
 )
 """
 
+def _render_list_and_select(deps, deps_by_platform, tmpl, group_deps):
+    deps = render.list(
+        [
+            tmpl % normalize_name(d)
```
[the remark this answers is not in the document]
aignas: I remember it being slightly slower in starlark than `%`, but we can worry about the speed when that becomes a problem.

python/pip_install/pip_repository.bzl
```diff
@@ -713,7 +713,20 @@ def _whl_library_impl(rctx):
...
 
     result = rctx.execute(
-        args + ["--whl-file", whl_path],
+        args + [
+            "--whl-file",
+            whl_path,
+            # TODO @aignas 2023-12-04: this is not the best way right now
+            # so if this approach is generally good, this can be refactored, by
+            # using the names of the config_setting labels.
+            "--platform=linux_x86_64",
```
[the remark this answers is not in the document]
aignas: Added extra platforms and moved the definition to an `attr` on the `repository_rule`. It may still need us updating the `wheel.py` file for any unseen platforms, but I made it a little bit more robust.

python/pip_install/tools/wheel_installer/wheel.py
```diff
@@ -13,18 +13,186 @@
...
+    def from_string(cls, platform: str):
+        os, _, arch = platform.partition("_")
+
+        return cls(
+            os=OS[os],
+            arch=Arch[arch],
+        )
+
+    @property
+    def os_name(self) -> str:
```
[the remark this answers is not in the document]
aignas: Added a link to the PEP that defined this.

python/pip_install/tools/wheel_installer/wheel.py
```diff
@@ -13,18 +13,186 @@
...
+            return "Windows"
+
+
+def _default_select():
+    return defaultdict(set)
+
+
+@dataclass
+class Deps:
+    deps: Set[str] = field(default_factory=set)
```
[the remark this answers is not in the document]
aignas: Moved the sorting to this class to make it more evident.

aignas:
I think the resultant BUILD.bazel files are good enough for it to be considered for merging. @arrdem, cc-ing yau as you've spent quite some time in this recently and would love to hear your opinion/feedback on the change.

python/pip_install/tools/wheel_installer/wheel.py
```diff
@@ -13,18 +13,335 @@
...
+        if not self._select:
+            return FrozenDeps(
+                deps=sorted(self._deps),
+                deps_select={},
+            )
+
+        select_by_os = {
+            p: deps for p, deps in self._select.items() if deps and p.arch is None
+        }
+        # Order the most specific platforms first
```
rickeylev, line 326: The order of the select entries doesn't matter. [Only one can match, where specializations have precedence](https://bazel.build/docs/configurable-attributes#:~:text=It%27s%20an%20error%20for%20multiple%20conditions%20to%20match%20unless%20one%20is%20an%20unambiguous%20%22specialization%22%20of%20the%20others%20or%20they%20all%20resolve%20to%20the%20same%20value.%20See%20here%20for%20details.)

python/pip_install/pip_repository.bzl
```diff
@@ -513,6 +528,22 @@ python_interpreter. An example value: "@python3_x86_64-unknown-linux-gnu//:pytho
     "repo_prefix": attr.string(
         doc = """
 Prefix for the generated packages will be of the form `@<prefix><sanitized-package-name>//...`
+""",
+    ),
+    "target_platforms": attr.string_list(
```
rickeylev, line 533: Please add the list of supported values into the doc somewhere. Or have it set as the `values` arg (which enforces the value is in the given list). Otherwise you have to fail first to see the values. Much nicer to have it seen in the docs.

python/pip_install/pip_repository.bzl
```diff
@@ -513,6 +528,22 @@ python_interpreter. An example value: "@python3_x86_64-unknown-linux-gnu//:pytho
...
+        default = [],
+        doc = """\
+A list of platforms that we will generate the conditional dependency graph for
+cross platform wheels by parsing the wheel metadata. This will generate the
+correct dependencies for packages like `sphinx` or `pylint`, which include
+`colorama` when installed and used on Windows platforms.
+
+An empty list means falling back to the legacy behaviour when we use the host
```
rickeylev, line 541: Should we have a special "host" value?

python/pip_install/pip_repository.bzl
```diff
@@ -513,6 +528,22 @@ python_interpreter. An example value: "@python3_x86_64-unknown-linux-gnu//:pytho
     "repo_prefix": attr.string(
         doc = """
 Prefix for the generated packages will be of the form `@<prefix><sanitized-package-name>//...`
+""",
+    ),
+    "target_platforms": attr.string_list(
+        default = [],
+        doc = """\
+A list of platforms that we will generate the conditional dependency graph for
```
rickeylev, line 536: Should we have a `*` matching? e.g. `linux_*` to match all linux archs?

python/pip_install/pip_repository.bzl
rickeylev, line 533: Long term, should we just plan on making target_platforms=all the default behavior?

python/pip_install/pip_repository.bzl
```diff
@@ -669,6 +700,15 @@ See the example in rules_python/examples/pip_parse_vendored.
 )
 
 def _whl_library_impl(rctx):
+    target_platforms = []
+    if len(rctx.attr.target_platforms) == 1 and rctx.attr.target_platforms[0] == "all":
+        target_platforms = _ALL_PLATFORMS
+    elif rctx.attr.target_platforms:
+        target_platforms = rctx.attr.target_platforms
+        for p in target_platforms:
+            if p not in _ALL_PLATFORMS:
```
rickeylev, line 709: What defines the set of known/valid platforms?

I guess it's somewhat limited by what we can generate constraints for. But, ignoring that aspect.

It looks like it's technically unbounded? After going through a few peps, I found this: https://packaging.python.org/en/latest/specifications/platform-compatibility-tags/#platform-compatibility-tags

We probably want to link to that somewhere in the comments? Probably over in the wheel.py, since that's where the "parsing" of the whl platform tag is converted to the strings starlark understands?

python/pip_install/pip_repository.bzl
aignas: Maybe just `linux` would be sufficient as well.

python/pip_install/pip_repository.bzl
aignas: Correct.

python/pip_install/pip_repository.bzl
aignas: Yeah, technically it is unbouded and in theory, we could just do all of the combinations of `OS` and `Arch` enums.

python/pip_install/tools/wheel_installer/wheel.py
```diff
@@ -13,18 +13,335 @@
...
+            # we assume that the following are the same as the interpreter used to setup the deps:
+            # "implementation_version": "X.Y.Z",
+            # "implementation_name": "cpython"
+            # "python_version": "X.Y",
+            # "python_full_version": "X.Y.Z",
+            # "platform_python_implementation: "CPython",
+        }
+
+
+FrozenDeps = namedtuple("FrozenDeps", "deps deps_select")
```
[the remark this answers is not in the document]
aignas: OK, will change to a frozen dataclass.

python/pip_install/tools/wheel_installer/wheel.py
```diff
@@ -13,18 +13,335 @@
...
+    def __init__(
+        self,
+        name: str,
+        extras: Optional[Set[str]] = None,
+        platforms: Optional[Set[Platform]] = None,
+    ):
+        self.name: str = Deps._normalize(name)
+        self._deps: Set[str] = set()
+        self._select: Dict[Platform, Set[str]] = defaultdict(set)
+        self._want_extras: Set[str] = extras or {""}
```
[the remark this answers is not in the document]
aignas: Empty string means no extras.

python/pip_install/tools/wheel_installer/wheel.py
```diff
@@ -13,18 +13,335 @@
...
+                req.marker.evaluate(plat.env_markers(extra)) for extra in extras
+            ):
+                continue
+
+            if "platform_machine" in marker_str:
+                self._add(req.name, plat)
+            else:
+                self._add(req.name, Platform(plat.os))
+
+    def build(self, m=None) -> FrozenDeps:
```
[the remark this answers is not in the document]
aignas: Yeah, this has to be removed.

python/pip_install/pip_repository.bzl
aignas: I have removed the starlark validation and left only the validation in Python.

python/pip_install/tools/wheel_installer/wheel.py
```diff
@@ -94,13 +129,27 @@ def from_tag(cls, tag: str) -> "Platform":
         )
 
     @classmethod
-    def from_string(cls, platform: str) -> "Platform":
-        os, _, arch = platform.partition("_")
+    def from_string(cls, platform: Union[str, List[str]]) -> List["Platform"]:
```
rickeylev, line 139: Small simplification: normalize the value to a list, then operate with the list value uniformly. Since we have to potentially loop over the values anyways to match the `*` cases, we can just always loop:

```
values = [value] if isinstance(str, value) else value

for value in values:
  if value == "host": result.update(...)
  elif value == "all": result.update(...)
  elif <star case> : ...
  return result
``

File: python/pip_install/tools/wheel_installer/wheel.py [hunk 15]
```diff
@@ -80,6 +82,39 @@ class Platform:
...
+            if not want_os or want_os == os
+        ]
+
+    @classmethod
+    def host(cls) -> "Platform":
+        return [
+            cls(os=OS[sys.platform.lower()], arch=Arch[platform.machine()])
+            for os in OS
+            for arch in Arch
+        ]
```
rickeylev: This looks wrong? "For every OS and arch combination, return the host os/arch" ? I think you meant to not have a list comp here?

examples/bzlmod/MODULE.bazel
```diff
@@ -102,7 +102,13 @@ pip.parse(
     python_version = "3.9",
     requirements_lock = "//:requirements_lock_3_9.txt",
     requirements_windows = "//:requirements_windows_3_9.txt",
-    target_platforms = ["all"],
+    # You can use one of the values below to specify the target platform
+    # to generate the dependency graph for.
+    target_platforms = [
```
rickeylev, line 107: (Out of scope for this PR, just thinking out loud)

Hm, if we had a label_keyed_string_dict, then I think we could create an arbitrary mapping of conditions to platform strings? So if you had some exotic platform, you could do...

```
target_platform_mapping = {
  "@//myplatforms:chromeos_arm": "chromeos_arm"
}
```

And then, if we parse out "chromeos_arm" from the wheel name, we're able to put the corresponding label for the select.

Oh hm. Or maybe have a register() tag part of a bzlmod extension? e.g. `pip.register_platform(name="chromeos_arm", label="@//myplatforms:chromeos_arm", os="chromeos", arch="arm")`  ? The latter sounds like it would be useful for...something.

examples/bzlmod/MODULE.bazel
```diff
@@ -126,7 +132,13 @@ pip.parse(
     python_version = "3.10",
     requirements_lock = "//:requirements_lock_3_10.txt",
     requirements_windows = "//:requirements_windows_3_10.txt",
-    target_platforms = ["all"],
+    # You can use one of the values below to specify the target platform
+    # to generate the dependency graph for.
+    target_platforms = [
```
rickeylev, line 137: At the maintainers meeting, you sounded uncertain how well this will interact with the dependency grouping. Are you more certain now?

I'm just thinking we might want to name this `experimental_target_platforms` or `incompatible_target_platforms` if we're not sure on its stability and interoperability.

python/pip_install/tools/wheel_installer/wheel.py
```diff
@@ -80,6 +82,39 @@ class Platform:
...
+    def all(cls, want_os: Optional[OS] = None) -> List["Platform"]:
+        return [
+            cls(os=os, arch=arch)
+            for os in OS
+            for arch in Arch
+            if not want_os or want_os == os
+        ]
+
+    @classmethod
+    def host(cls) -> "Platform":
```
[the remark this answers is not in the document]
aignas: Fixed the signature.

python/pip_install/tools/wheel_installer/wheel.py
```diff
@@ -94,13 +129,27 @@ def from_tag(cls, tag: str) -> "Platform":
         )
 
     @classmethod
-    def from_string(cls, platform: str) -> "Platform":
-        os, _, arch = platform.partition("_")
+    def from_string(cls, platform: Union[str, List[str]]) -> List["Platform"]:
+        if platform == "host":
+            return [cls.host()]
```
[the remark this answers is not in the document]
aignas: Good catch, this simplifies the code elsewhere.

python/pip_install/tools/wheel_installer/wheel.py
aignas: Yeah, this got fixed very recently.

examples/bzlmod/MODULE.bazel
aignas: I'll create a followup ticket.

examples/bzlmod/MODULE.bazel
aignas: After a little bit more testing and tinkering with the code, I am more certain that it should work. I was a little bit uncertain on how the dependency groups and this would work together, but now I think it should work well.

That said, I am happy to rename the parameter to signal that the API is not fully stable.

rickeylev (approved):

PR #1613, opened 2023-12-14
chore: Update Bazel support: drop 5, minimum 6.2, current 7.0

rickeylev:
With the release of Bazel 7, support for 5 is no longer necessary.

Also deletes all the integration tests of the examples. They were broken when
run from the main repo, and CI was only building them, not running them, as part
of the main repo due to having `--test_tag_filters=-integration-test`. CI runs them
separately in their own workspace.

Work towards #1069


groodt (approved):
LGTM

rickeylev:
Huh. Apparently all those bazel_integration_test targets in the examples directory have been broken and CI has been skipping them as part of the main repo tests (CI runs them as separate jobs in separate workspace directories). I just went and deleted them.

Issue #1618, opened 2023-12-15
pip.parse doesn't support python_interpreter (string) argument


rickeylev:
# 🐞 bug report

### Affected Rule

`pip.parse`

### Is this a regression?

No

### Description

The pip.parse() bzl mod extension accepts a `python_interpreter` arg, which is a string path
to some interpreter (basically an equiv to `py_runtime.interpreter_path`). This is an alternative to `python_interpreter_target` for specifying the interpreter to use.

This arg is, essentially, ignored. I think it is inherited from the workspace pip_parse attributes and got overlooked when implementing the bzlmod extension.

Looking through the code a bit, I think we just need to check `.python_interpreter` wherever we're checking for `.python_interpreter_target`.

## 🔬 Minimal Reproduction

Use `pip.parse` with the `python_interpreter` arg set. Specify a version of Python that isn't the default (e.g. 3.9), and *don't* register a toolchain for that version using `python.toolchain`. This will cause an error about being unable to find the hermetic runtimes, which shouldn't occur because they shouldn't be used.

```
load("@rules_python//python/extensions:pip.bzl", "pip")

pip.parse(
    hub_name = "syspy",
    python_interpreter = "/usr/bin/python3",
    python_version = "3.9",
)
```

## 🔥 Exception or Error

<pre><code>
File "/.../rules_python~0.27.1/python/private/bzlmod/pip.bzl", line 298, column 30, in _pip_impl
                _create_whl_repos(module_ctx, pip_attr, hub_whl_map, whl_overrides)
        File "/.../rules_python~0.27.1/python/private/bzlmod/pip.bzl", line 91, column 17, in _create_whl_repos
                fail((
Error in fail: Unable to find interpreter for pip hub 'pip_deps' for python_version=3.9: Make sure a corresponding `python.toolchain(python_version="3.9")` call exists

</code></pre>

rickeylev:
Alright, the immediate fix is fairly easy (there were 2 small bugs).

I'm not sure I can create a test fit for CI, though. By definition, this relies on local system state, which is a huge headache. I think the best bet is to do a bazel-in-bazel test where the local environment bazel sees can be directly controlled. I think that will have to wait until we switch to the newer bazel_integration_test rules, though, as our current ones are all broken and disabled.

-----

I also uncovered an annoyance: the default python version can only be specified by registering a hermetic toolchain. This is normally fine, but, in this case, our basic goal is to _not_ use those, and instead the local runtime for everything.

[...]

PR #1619, opened 2023-12-15
fix(pip.parse): allow absolute path for python_interpreter; skip hermetic toolchain lookup when used

rickeylev:
The `python_interpreter` arg wasn't being properly handled in the bzlmod code in two ways.

1. Lookup of a hermetic runtime wasn't being skipped when it was set. The net effect was a hermetic toolchain of a matching version had to be registered, even if it wasn't used. To fix, add a check for python_interpreter to the guard of the lookup.

2. Specifying an absolute path for the value wasn't being converted to a `path` object, which meant a plain string eventually made its way to some code expecting a `path` object. To fix, call `repository_ctx.path()` on the path to convert it to a path object.

Fixes https://github.com/bazelbuild/rules_python/issues/1618

### DIFF
--- a/python/pip_install/pip_repository.bzl
+++ b/python/pip_install/pip_repository.bzl
@@ -76,7 +76,9 @@ def _resolve_python_interpreter(rctx):
 
     Args:
         rctx: Handle to the rule repository context.
-    Returns: Python interpreter path.
+
+    Returns:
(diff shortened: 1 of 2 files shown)


aignas (approved):
LGTM

Issue #1615, opened 2023-12-15
pip.parse does not accept lock files with duplicated dependencies


AndrewGuenther:
# 🐞 bug report

### Affected Rule

`pip.parse`

### Is this a regression?

Yes, this behavior is only present when using `bzlmod`

### Description

When using `pip.parse`, if a dependency is repeated in your requirements lock, you'll get an error. This is handled just fine by pip and is necessary for tools like `poetry` to export this way because it's possible for the same dependency to have different constraints in different parts of the dependency tree.

Ref: https://github.com/python-poetry/poetry/pull/5688

## 🔬 Minimal Reproduction

Repo with repro here: https://github.com/AndrewGuenther/rules_python_1615_repro

The `main` branch uses `bzlmod` and `workspace` uses `WORKSPACE`. The `workspace` branch builds fine, but `main` does not. Same requirements on both branches, the only difference is `bzlmod`.

## 🔥 Exception or Error

If you try and build the example above, you'll get an error to the following effect:

<pre><code>
ERROR: Traceback (most recent call last):
        File "/home/andrew/.cache/bazel/_bazel_andrew/9ca33225fba41f1be5682708dd655aad/external/rules_python~0.27.1/python/private/bzlmod/pip.bzl", line 298, column 30, in _pip_impl
                _create_whl_repos(module_ctx, pip_attr, hub_whl_map, whl_overrides)
        File "/home/andrew/.cache/bazel/_bazel_andrew/9ca33225fba41f1be5682708dd655aad/external/rules_python~0.27.1/python/private/bzlmod/pip.bzl", line 129, column 20, in _create_whl_repos
                whl_library(
Error in repository_rule: A repo named pip_38_pyjwt is already generated by this module extension at /home/andrew/.cache/bazel/_bazel_andrew/9ca33225fba41f1be5682708dd655aad/external/rules_python~0.27.1/python/private/bzlmod/pip.bzl:129:20
ERROR: error evaluating module extension pip in @@rules_python~0.27.1//python/extensions:pip.bzl
INFO: Elapsed time: 1.071s
INFO: 0 processes.
ERROR: Build did NOT complete successfully
FAILED: 
    Fetching module extension pip in @@rules_python~0.27.1//python/extensions:pip.bzl; starting
</code></pre>

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
Ubuntu
  </code>
</pre>

**Output of `bazel version`:**

<pre>
  <code>
6.4.0
[...]

PR #1620, opened 2023-12-16
fix(bzlmod pip.parse): allow requirements with duplicate package lines

rickeylev:
Requirements files are permitted to have duplicate lines for the same package. An example
of this is having separate lines for a package and its extras. When we parse requirements,
the parser discards the "extra" portion of the entry and returns a list of all the packages
as-is. When a repository is created for each entry, this means the same name is used for
multiple repositories.

Under WORKSPACE builds, duplicate repository names aren't an error. It appears that last
defined repo takes affect. Under bzlmod, duplicate repo names are an error.

To fix, mimic the last-defined-wins behavior in bzlmod by using a map to dedupe the package names.

Fixes https://github.com/bazelbuild/rules_python/issues/1615


python/private/bzlmod/pip.bzl
```diff
@@ -109,7 +109,19 @@ def _create_whl_repos(module_ctx, pip_attr, whl_map, whl_overrides):
...
+    # Defining a repo with the same name multiple times, but only the last
+    # definition is respected.
+    # The requirement lines might have duplicate names because lines for extras
+    # are returned as just the base package name. e.g., `foo[bar]` results
+    # in an entry like `("foo", "foo[bar] == 1.0 ...")`.
+    requirements = {
+        normalize_name(entry[0]): entry
+        # The WORKSPACE pip_parse sorted entries, so mimic that ordering.
+        for entry in sorted(parse_result.requirements)
+    }.values()
```
aignas, line 123: I was wondering if the fix should be in the `parse_requirements` function instead, but I do like it here better even though this needs an extra comment here.

aignas (approved):
Thanks for adding a test for this!

aignas:
The test failures look like the `pre-commit` hook may need to be run or something similar?

AndrewGuenther:
@rickeylev I patched this to our local environment and can confirm it works!

Issue #1612, opened 2023-12-14
Cannot use `py_runtime` with the output of a `sh_binary`


snakethatlovesstaticlibs:
# 🐞 bug report

### Affected Rule

The issue is caused by the rule: `py_runtime`

### Is this a regression?

 No

### Description

We're not able to use `py_runtime` with the output of a `sh_binary`. This is useful because we want to customize how python searches for dynamic libraries at runtime.

## 🔬 Minimal Reproduction

https://github.com/snakethatlovesstaticlibs/bazel-py-runtime-repro

## 🔥 Exception or Error

<pre><code>
BUILD:11:11: in interpreter attribute of py_runtime rule //:python3_runtime: '//:wrapped_python' must produce a single file. Since this rule was created by the macro 'py_runtime', the error might have been caused by the macro implementation
</code></pre>

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
  Linux 6.5.11-linuxkit #1 SMP PREEMPT Mon Dec  4 11:30:00 UTC 2023 aarch64 aarch64 aarch64 GNU/Linux and
  Darwin 22.6.0 Darwin Kernel Version 22.6.0
  </code>
</pre>

**Output of `bazel version`:**

<pre>
  <code>
bazel 7.0.0
  </code>
</pre>

**Rules_python version:**

<pre>
  <code>
latest main / b6ba317b8227c6d182d5e41d2571a217ecdf11a9d3ebc2cdef3bc5503fca2e76
  </code>
</pre>

**Anything else relevant?** 

Related to https://github.com/bazelbuild/bazel/issues/4286

snakethatlovesstaticlibs:
Thanks for the quick response. I'm still pretty new to bazel, so I could potentially make the changes with guidance. I had some questions about this approach:

* Do you have any examples for other rules that modify the logic based on whether the input is executable?
* Which changes would be required to the consumers of the `py_runtime`?
* Would we need to make changes to any providers to get this working?

rickeylev:
> Do you have any examples for other rules that modify the logic based on whether the input is executable?

You can look at the `py_runtime.coverage_tool` attribute for an example of how a label attribute can be treated as an executable or a singleton file.

> Which changes would be required to the consumers of the py_runtime?

None. A single-file target is expressed the same way as a multiple-file target (they're both just `//foo:bar`.

> Would we need to make changes to any providers to get this working?

Maybe? That depends on what the target being passed as the interpreter needs at runtime.

[...]

PR #1621, opened 2023-12-16
feat(py_runtime): Allow `py_runtime` to take an executable target as the interpreter

mishazharov:
This PR allows `py_runtime` to accept an executable (e.g. `sh_binary`).

This makes it easier to customize the interpreter binary used, as it allows
intercepting invocation of the interpreter. For example, it can be used to
change how the interpreter searches for dynamic libraries.

Related to https://github.com/bazelbuild/rules_python/issues/1612

### DIFF
--- a/python/private/common/py_runtime_rule.bzl
+++ b/python/private/common/py_runtime_rule.bzl
@@ -45,9 +45,7 @@ def _py_runtime_impl(ctx):
     else:
         interpreter_di = interpreter[DefaultInfo]
 
-        if _is_singleton_depset(interpreter_di.files):
-            interpreter = interpreter_di.files.to_list()[0]
-        elif interpreter_di.files_to_run and interpreter_di.files_to_run.executable:
+        if interpreter_di.files_to_run and interpreter_di.files_to_run.executable:
             interpreter = interpreter_di.files_to_run.executable
             runfiles = runfiles.merge(interpreter_di.default_runfiles)
 
@@ -56,6 +54,8 @@ def _py_runtime_impl(ctx):
                 interpreter_di.default_runfiles.files,
                 runtime_files,
             ])
+        elif _is_singleton_depset(interpreter_di.files):
+            interpreter = interpreter_di.files.to_list()[0]
         else:
             fail("interpreter must be an executable target or must produce exactly one file.")
 
--- a/python/private/common/py_runtime_rule.bzl
+++ b/python/private/common/py_runtime_rule.bzl
@@ -25,7 +25,7 @@ _py_builtins = py_internal
 
 def _py_runtime_impl(ctx):
     interpreter_path = ctx.attr.interpreter_path or None  # Convert empty string to None
-    interpreter = ctx.file.interpreter
+    interpreter = ctx.attr.interpreter or None
     if (interpreter_path and interpreter) or (not interpreter_path and not interpreter):
         fail("exactly one of the 'interpreter' or 'interpreter_path' attributes must be specified")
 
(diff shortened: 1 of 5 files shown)


python/private/common/py_runtime_rule.bzl
```diff
@@ -25,7 +25,7 @@ _py_builtins = py_internal
 
 def _py_runtime_impl(ctx):
     interpreter_path = ctx.attr.interpreter_path or None  # Convert empty string to None
-    interpreter = ctx.file.interpreter
+    interpreter = ctx.attr.interpreter or None
```
rickeylev, line 28: You can omit the `or None`. Optional labels are None when not given a value

python/private/common/py_runtime_rule.bzl
```diff
@@ -34,12 +34,33 @@ def _py_runtime_impl(ctx):
...
+    runfiles = ctx.runfiles()
+
     hermetic = bool(interpreter)
     if not hermetic:
         if runtime_files:
             fail("if 'interpreter_path' is given then 'files' must be empty")
         if not paths.is_absolute(interpreter_path):
             fail("interpreter_path must be an absolute path")
+    else:
+        interpreter_di = ctx.attr.interpreter[DefaultInfo]
```
rickeylev, line 46:
```suggestion
        interpreter_di = interpreter[DefaultInfo]
```

python/private/common/py_runtime_rule.bzl
```diff
@@ -34,12 +34,33 @@ def _py_runtime_impl(ctx):
...
+
+        if _is_singleton_depset(interpreter_di.files):
+            interpreter = interpreter_di.files.to_list()[0]
+        elif interpreter_di.files_to_run and interpreter_di.files_to_run.executable:
+            interpreter = interpreter_di.files_to_run.executable
+            runfiles.merge(interpreter_di.default_runfiles)
```
rickeylev, line 52: This line is a basically a no-op: runfiles.merge doesn't modify in-place, it returns an updated object.

I think you meant to `runfiles = runfiles.merge(...)` ?

python/private/common/py_runtime_rule.bzl
```diff
@@ -34,12 +34,33 @@ def _py_runtime_impl(ctx):
...
+
+            runtime_files = depset(
+                direct = [interpreter_di.files_to_run.runfiles_manifest, interpreter_di.files_to_run.repo_mapping_manifest],
```
rickeylev, line 55: files_to_run.runfiles_manifest can be None if certain flags are set, so this needs a guard, at the least.

However, I don't think either the manifest or repo_mapping need to be manually propagated; you can remove that part.

python/private/common/py_runtime_rule.bzl
```diff
@@ -186,10 +207,11 @@ runtime. For a platform runtime this attribute must not be set.
 """,
         ),
         "interpreter": attr.label(
-            allow_single_file = True,
+            allow_files = True,
```
rickeylev, line 210: Please put a comment about why this is allow_files instead of allow_single_file, as normally we'd expect an attribute that is supposed to be a single file to be marked as such.

rickeylev (changes requested):

mishazharov:
Thanks for the review, I've also added a bare bones unit test

python/private/common/py_runtime_rule.bzl
```diff
@@ -34,12 +34,30 @@ def _py_runtime_impl(ctx):
...
     if not hermetic:
         if runtime_files:
             fail("if 'interpreter_path' is given then 'files' must be empty")
         if not paths.is_absolute(interpreter_path):
             fail("interpreter_path must be an absolute path")
+    else:
+        interpreter_di = interpreter[DefaultInfo]
+
+        if _is_singleton_depset(interpreter_di.files):
+            interpreter = interpreter_di.files.to_list()[0]
```
rickeylev, line 49: Hm, does this logic incorrectly ignore a well-behaved executable rule? i.e. one that has just a single file (the executable) its default outputs? We might need to flip the order of checking here.

tests/py_runtime/pretend_binary
rickeylev, line 1: A neat thing about analysis tests is they don't actually require the files to exist, so this can be deleted. Analysis tests don't actual build the code; they run rule logic, but stop short of doing the subsequent steps where things need to actually read files.

tests/py_runtime/py_runtime_tests.bzl
```diff
@@ -320,6 +320,32 @@ def _test_system_interpreter_must_be_absolute_impl(env, target):
 
 _tests.append(_test_system_interpreter_must_be_absolute)
 
+def _test_interpreter_sh_binary_target(name):
+    native.sh_binary(
```
rickeylev, line 324: This needs to use rt_util.helper_target. It just hides the target from `...` expansion so that it isn't built when it shouldn't be (it should only be built as part of the analysis test).

rickeylev, line 324: Rather than use sh_binary itself, lets create an executable rule that isn't well behaved (has multiple outputs).

I don't want to use sh_binary itself here because (1) we're interested in the behavior of multiple outputs, not sh_binary itself, and (2) I spent months cleaning up usages of py_* rules in tests because they incidentally had some behavior and they were convenient to use, but woo, that was a lot of work, so I don't want to propagate that bad practice.

rickeylev (commented):
This looks pretty good; almost there. Most is just cosmetic changes.  Only real potential bug is the order of checks for the single-file vs executable cases.

I got a bit of time now, so I think I can resolve most of the requested changes and push them into this PR.

tests/py_runtime/pretend_binary
rickeylev: Deleted

tests/py_runtime/py_runtime_tests.bzl
rickeylev: Done

rickeylev: Done

rickeylev (approved):
I think this is in good shape and can be merged.

The only thing I can think to add is an integration test to verify that using an sh_binary as described actually works, but that can be done in a separate PR if necessary.

mishazharov:
Awesome, thank you for getting this over the line!

PR #1623, opened 2023-12-17
fix: Handle the empty `whl_map` correctly

kshramt:


rickeylev (changes requested):
Ah, it looks like the bug is bzl_packages can be None, which then the for-loop tries to iterate over? Thanks for finding this.

Can you update the description to describe the bug and fix, please? A test can be added to tests/pip_hub_repository/render_pkg_aliases/render_pkg_aliases_test.bzl

PR #1628, opened 2023-12-18
test(bzlmod): refactor tests to not depend on implementation details

aignas:
This refactors the `whl_mods` tests to not rely on the layout of the
repositories, which I found to be needed whilst prototyping on #1625.
Whilst doing this I realized that in general it would be great to
support `Path` instances in the `runfiles` library, but that should be
done next time.


aignas:
The change can be smaller, so I'll revert the path usage within the tests and will keep the helper methods returnitg strings. The implementation of those should return Path.name as opposed to just Path.

rickeylev (approved):

PR #1629, opened 2023-12-18
docs: use stardoc proto output to generate markdown docs

rickeylev:
The template language Stardoc uses (Velocity) is niche and fairly
esoteric, and requires a lot of experimenting to understand how to
make it produce the desired output. In particular, it largely assumes
whitespace doesn't matter, which makes it a poor fit for generating
Markdown, where whitespace often does matter.

Instead, a small Python program is used to consume the binary proto
output of Stardoc, which converts it to Markdown. This also makes it
easier to customize the overall output and re-use code for the different
types of objects rendered.

The visible changes to the docs are:
  * Module extensions are now documented
  * Repository rules follow the style of the other generated docs
  * Fixed the rendering of pip_repository docs -- it had an h2
    section which broke the section grouping of the API objects.
  * Puts some padding between the border and content for text in
    params/attrs/fields listings.

Other notable changes:
  * Make RTD builds use bzlmod. This is necessary so that the pip
    extension can be documented. It loads
    `@pythons_hub//:interpreters.bzl`, but that repo is only created
    when bzlmod is enabled)


rickeylev:
Ah doh, need to fix a few test failures before review.

rickeylev:
Ok, tests fixed. Just windows flakes left.

docs/sphinx/BUILD.bazel
```diff
@@ -83,7 +85,13 @@ sphinx_stardocs(
         "api/entry_points/py_console_script_binary.md": "//python/entry_points:py_console_script_binary_bzl",
         "api/packaging.md": "//python:packaging_bzl",
         "api/pip.md": "//python:pip_bzl",
-    },
+    } | ({
+        # Bazel 6 + Stardoc isn't able to parse something about the python bzlmod extension
+        "api/extensions/python.md": "//python/extensions:python_bzl",
+    } if IS_BAZEL_7_OR_HIGHER else {}) | ({
+        # This depends on @pythons_hub, which is only created under bzlmod
+        "api/extensions/pip.md": "//python/extensions:pip_bzl",
```
aignas, line 93: nit: I don't see the docs in the PR preview here: https://rules-python--1629.org.readthedocs.build/en/1629/api/extensions/python.html#

sphinxdocs/private/proto_to_markdown.py
```diff
@@ -0,0 +1,429 @@
...
+    with pathlib.Path(output_path).open("wt", encoding="utf8") as out_stream:
+        MySTRenderer(module, out_stream, public_load_path).render()
+        out_stream.write(footer_content)
+
+
+def _create_parser():
+    parser = argparse.ArgumentParser(fromfile_prefix_chars="@")
+    parser.add_argument("--footer", dest="footer")
+    parser.add_argument("--proto", dest="proto")
+    parser.add_argument("--output", dest="output")
```
aignas, line 412: having these as Paths may be nicer in other functions.

```suggestion
    parser.add_argument("--footer", dest="footer", type=pathlib.Path)
    parser.add_argument("--proto", dest="proto", type=pathlib.Path)
    parser.add_argument("--output", dest="output", type=pathlib.Path)
```

sphinxdocs/private/proto_to_markdown.py
```diff
@@ -0,0 +1,429 @@
...
+                # The text has to be indented to be associated with the block correctly.
+                "".join(body_pieces).strip().replace("\n", "\n  "),
+                "\n",
+            )
+
+    def _write(self, *lines):
+        self._out_stream.writelines(lines)
+
+
+def _convert(*, proto_path, output_path, footer_path, public_load_path):
```
aignas, line 396: nit: having type annotations in all functions may help with readability because most of the functions don't have docstrings.

sphinxdocs/private/proto_to_markdown.py
```diff
@@ -0,0 +1,429 @@
...
+                [
+                    anchor,
+                    attr.name,
+                    f"_({required} {attr_type}{default})_\n",
+                    attr.doc_string,
+                ]
+            )
+        self._render_field_list(entries)
+
+    def _render_signature(
```
aignas, line 366: I would love to have at list a single test for each of the rendering functions to illustrate what we are doing here. That said, the tests may be too verbose, so I am OK with not having them. Especially that we are reading protos to generate the docs.

sphinxdocs/private/proto_to_markdown.py
```diff
@@ -0,0 +1,429 @@
...
+            entries = []
+            for param in parameters:
+                entries.append(
+                    [
+                        f"{func_anchor}_{param.name}",
+                        param.name,
+                        f"(_default `{param.default_value}`_) "
+                        if param.default_value
+                        else "",
+                        param.doc_string if param.doc_string else "_undocumente_",
```
aignas, line 286:
```suggestion
                        param.doc_string if param.doc_string else "_undocumented_",
```

docs/sphinx/BUILD.bazel
rickeylev: Good catch! The RTD builds weren't using --config=rtd. Fixed.

sphinxdocs/private/proto_to_markdown.py
rickeylev: Done

sphinxdocs/private/proto_to_markdown.py
rickeylev: I added a few tests to at least exercise a decent chunk of the code paths.

Testing the markdown output is a bit tough without being really brittle.

I don't really want to put too much into testing the output because after this, I'm going to try making a Sphinx plugin. That way, the markdown that is generated can be simpler (a more raw api definitions, ala what you'd do for defining a python API), and sphinx can take care of transforming it to the UI format (sphinx has a document model plus direct integration into its other facilities like cross refs etc).

aignas: SGTM, thanks for this work!

docs/sphinx/BUILD.bazel
```diff
@@ -12,8 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-load("@docs_deps//:requirements.bzl", "requirement")
-load("@rules_python//python:pip.bzl", "compile_pip_requirements")
+load("@dev_pip//:requirements.bzl", "requirement")
```
aignas, line 15: nit: technically I think we don't need this macro anymore as the same alias labels can be accessed from WORKSPACE and MODULE.bazel setups.

docs/sphinx/_stardoc_footer.md
```diff
@@ -7,6 +7,8 @@
 [`Label`]: https://bazel.build/rules/lib/Label
 [`list`]: https://bazel.build/rules/lib/list
 [`str`]: https://bazel.build/rules/lib/string
+[str]: https://bazel.build/rules/lib/string
```
aignas, line 10: Why do we have both? `str` and `str`?

aignas (approved):
LGTM and merging since the questions are very minor.

docs/sphinx/_stardoc_footer.md
rickeylev: The "ref" style links require an exact match of their contents. For whatever reason (I'll go with "whimsy"), I made it output

```
list of [str]
# instead of
list of [`str`]
```

(I don't really like either style; `list[str]` is more natural IMHO, but I'm not actually sure how to make it render with code font, clickable type links, and look OK)

PR #1630, opened 2023-12-18
fix: missing dependencies of py_binary

linzhp:
When multiple main modules import the same package, the package is only added to one of the main module's deps, this is because the `allDeps` set returned from `parse` removed the duplicates based on the import's name.

This PR have every main module maintains a list of deps it imports to avoid this issue.


linzhp:
@aignas in case you missed this PR

aignas (approved):

PR #1634, opened 2023-12-19
fix: support `runfiles.CurrentRepository` with non-hermetic `sys.path`

### LINKED ISSUE #1631: bazel-runfiles module always throws ValueError
# 🐞 bug report

### Affected Rule

`bazel-runfiles` ackage

### Is this a regression?

Appears to be, we were using `bazel-runfiles` on v0.12.0 with no issue

### Description

When running a `py_binary` target, the `main` file is executed from within the runfiles tree, but everything imported points back to the original workspace. As a result of this, if you try to use `bazel-runfiles`, you're always met with this `ValueError`:

```
ValueError: /home/andrew/workspace/rules_python_runfiles_repro/my_module/__init__.py does not lie under the runfiles root /home/andrew/.cache/bazel/_bazel_andrew/43e798d9071ddd30271aac70168abdc4/execroot/_main/bazel-out/k8-fastbuild/bin/hello_world.runfiles
```

## 🔬 Minimal Reproduction

https://github.com/AndrewGuenther/rules_python_1631_repro

Minimal reproduction repository linked above.

## 🔥 Exception or Error

<pre><code>
ValueError: /home/andrew/workspace/rules_python_runfiles_repro/my_module/__init__.py does not lie under the runfiles root /home/andrew/.cache/bazel/_bazel_andrew/43e798d9071ddd30271aac70168abdc4/execroot/_main/bazel-out/k8-fastbuild/bin/hello_world.runfiles
</code></pre>

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
Ubuntu
  </code>
</pre>

<pre>
  <code>
6.4.0
  </code>
</pre>

**Rules_python version:**

<pre>
  <code>
0.27.1
  </code>
</pre>

**Anything else relevant?**

fmeum:
When using Python 3.10 or earlier, the first `sys.path` entry is the directory containing the script. This can result in modules being loaded from the source root rather than the runfiles root, which the runfiles library previously didn't support.

Fixes #1631


fmeum:
@rickeylev Not sure what the CI failures are about.

python/runfiles/runfiles.py
```diff
@@ -212,6 +212,15 @@ def CurrentRepository(self, frame=1):
...
         caller_runfiles_path = os.path.relpath(caller_path, self._python_runfiles_root)
         if caller_runfiles_path.startswith(".." + os.path.sep):
+            # With Python 3.10 and earlier, sys.path contains the directory
+            # of the script, which can result in a module being loaded from
+            # outside the runfiles tree. In this case, assume that the module is
+            # located in the main repository.
+            # TODO: This doesn't cover the case of a script being run from an
+            #       external repository, which could be heuristically detected
+            #       by parsing the script's path.
+            if sys.path[0] != self._python_runfiles_root:
```
aignas, line 222: What about checking that we are running on python 3.10 and older and applying this behaviour only then? That may be easier to understand later that we can remove it once the lowest supported Python version is 3.11 (which will probably be a while), but it would also give validation errors on Python 3.11 and above.

aignas:
The CI should be happy again if you rebase. We had some issues when this PR was created, if I remember correctly.

python/runfiles/runfiles.py
fmeum: I added a minor version check. Why would we see a validation error on Python 3.11?

aignas: I think with python 3.11, the hermeticity is improved and sys.path entries within the script are better isolated from the the system python. The comment in the code mentions that this affects only 3.10 and below, so I thought that this should not change anything for 3.11 and above, or should it?

fmeum: It shouldn't, I was just referring to your comment, which said "but it would also give validation errors on Python 3.11 and above.". I wasn't sure what that was about.

rickeylev:
re: just assume the file is in the main repo if its loaded from outside the runfiles

Yeah, that's probably fine. This is to support older Python versions that were already relying on non-hermetic behavior.

I think this problem would only happen for files that are in the same directory (or a subdirectory of) as the main .py file? I don't see how an import would be able to find such files otherwise. So I suppose the `sys.modules["__main__"].__file__` could be used in lieu of the caller's file in this situation. A main entry point that is in another repo _and_ is relying on `sys.path[0]` being main's directory is likely for an ad-hoc tool in the main repo, but seems unlikely for any sort of library code.

python/runfiles/runfiles.py
```diff
@@ -247,6 +247,15 @@ def CurrentRepository(self, frame: int = 1) -> str:
...
         caller_runfiles_path = os.path.relpath(caller_path, self._python_runfiles_root)
         if caller_runfiles_path.startswith(".." + os.path.sep):
+            # With Python 3.10 and earlier, sys.path contains the directory
+            # of the script, which can result in a module being loaded from
+            # outside the runfiles tree. In this case, assume that the module is
+            # located in the main repository.
+            # TODO: This doesn't cover the case of a script being run from an
+            #       external repository, which could be heuristically detected
+            #       by parsing the script's path.
+            if sys.version.minor <= 10 and sys.path[0] != self._python_runfiles_root:
```
rickeylev, line 257: CI called out a bug here: I think you meant sys.version_info

python/runfiles/runfiles.py
```diff
@@ -247,6 +247,15 @@ def CurrentRepository(self, frame: int = 1) -> str:
             raise ValueError("failed to determine caller's file path") from exc
         caller_runfiles_path = os.path.relpath(caller_path, self._python_runfiles_root)
         if caller_runfiles_path.startswith(".." + os.path.sep):
+            # With Python 3.10 and earlier, sys.path contains the directory
```
rickeylev, line 250: nit: mention the PYTHONSAFEPATH envvar. Python 3.11+ is a necessary but not sufficient condition to trigger this situation.

rickeylev (changes requested):

aignas (commented):
I think this got lost, @rickeylev, any extra comments you have here?

limdor:
Any updates on this PR? I think I'm hitting the same issue

aignas:
I merged `main` into this PR, let's see how it goes.

alexander-scott:
The problem indeed does only surface in edge-cases where:
  - Bazel 7 is used, with or without bzlmod
  - Python 3.9 or earlier is used
  - Runfiles rlocation is called from a file which is in the same directory as the main.py file (or test entry point)

But still I think this looks like the best solution for now, so it would be great to get this integrated if possible. Gentle ping @rickeylev

rickeylev (approved):
I updated this to main and added changelog text. LGTM.

PR #1635, opened 2023-12-19
doc: Updating doc on py_binary generation

linzhp:
Updating doc on the generation of `py_binary` targets to include the new feature that creates a py_binary target per main module.

### DIFF
--- a/gazelle/README.md
+++ b/gazelle/README.md
@@ -218,7 +218,7 @@ dependencies are added to the `deps` attribute.
 A `py_test` target is added to the BUILD file when gazelle encounters
 a file named `__test__.py`.
 Often, Python unit test files are named with the suffix `_test`.
-For example, if we had a folder that is a package named "foo" we could have a Python file named `foo_test.py` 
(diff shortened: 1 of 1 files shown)


gazelle/README.md
```diff
@@ -234,28 +234,35 @@ py_test(
...
 
-A `py_binary` target will be created, named `[package]_bin`.
+When no such entry point exists, Gazelle will look for a line like this in the top level in every module:
+
+```python
+if __name == "__main__":
+```
+
+Gazelle will create `py_binary` target will be created for every module with such line, with the target name
+being the same as module name.
```
aignas, line 252: @linzhp, do you think it would be a good idea to add a directive to disable this behaviour? Some people may want to avoid creating such binaries in certain directories and I am wondering if enabling `py_binary` generation per `module` by default and then giving an option to turn it off would be a better way to introduce this feature.

linzhp: Maybe we can wait for such feature request comes up?

aignas (approved):

PR #1637, opened 2023-12-20
fix(whl_library): actually apply patches and warn if RECORD file patch is needed

aignas:
Before this PR there was a typo, that was actually causing the
patching function to not use the provided patches. With this change
we are finally correctly doing it. After fixing this bug I noticed
that `repository_ctx.patch` results in files with `CRLF` on Windows,
which made me make the `RECORD` mismatch to be a warning rather than
a hard failure to make the CI happy and allow users on Windows to
patch wheels but see a warning if they have a multi-platform bazel
setup.

The `CRLF` endings on Windows issue is fixed in
bazelbuild/bazel@07e0d316a345a3cb2593f98525320590bbc56e30

Related #1631, #1639.

### DIFF
--- a/python/private/patch_whl.bzl
+++ b/python/private/patch_whl.bzl
@@ -73,11 +73,15 @@ def patch_whl(rctx, *, python_interpreter, whl_path, patches, **kwargs):
         parsed_whl.platform_tag,
     ]))
 
+    record_patch = rctx.path("RECORD.patch")
+
     result = rctx.execute(
         [
             python_interpreter,
             "-m",
             "python.private.repack_whl",
+            "--record-patch",
+            record_patch,
             whl_input,
             whl_patched,
         ],
@@ -96,5 +100,20 @@ def patch_whl(rctx, *, python_interpreter, whl_path, patches, **kwargs):
                 return_code = result.return_code,
             ),
         )
+
+    if record_patch.exists:
+        record_patch_contents = rctx.read(record_patch)
+        warning_msg = """WARNING: the resultant RECORD file of the patch wheel is different
+
+    If you are patching on Windows, you may see this warning because of
+    a known issue (bazelbuild/rules_python#1639) with file endings.
+
+    If you would like to silence the warning, you can apply the patch that is stored in
+      {record_patch}. The contents of the file are below:
+{record_patch_contents}""".format(
+            record_patch = record_patch,
+            record_patch_contents = record_patch_contents,
+        )
+        print(warning_msg)  # buildifier: disable=print
 
     return rctx.path(whl_patched)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,11 @@ A brief description of the categories of changes:
 * (bzlmod pip.parse) Requirements files with duplicate entries for the same
   package (e.g. one for the package, one for an extra) now work.
 * (whl_library) Actually use the provided patches to patch the whl_library.
+  On Windows the patching may result in files with CRLF line endings, as a result
+  the RECORD file consistency requirement is lifted and now a warning is emitted
+  instead with a location to the patch that could be used to silence the warning.
+  Copy the patch to your workspace and add it to the list if patches for the wheel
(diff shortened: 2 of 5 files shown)


aignas:
It seems that the tests started failing on windows, because `bazel`'s `rctx.patch` has different behaviour on Windows as it is replacing the line endings with `CRLF` as opposed to leaving them `LF`.

I could reproduce this by changing the repackaging code with:
```
--- a/python/private/repack_whl.py
+++ b/python/private/repack_whl.py
@@ -148,6 +148,8 @@ def main(sys_argv):

         with _WhlFile(args.output, mode="w", distinfo_dir=distinfo_dir) as out:
             for p in _files_to_pack(patched_wheel_dir, record_contents):
+                if p.name == "METADATA":
+                    p.write_text(p.read_text().replace("\n", "\r\n"))
                 rel_path = p.relative_to(patched_wheel_dir)
                 out.add_file(str(rel_path), p)
```

And the test started failing with:
```
repack_whl: ERROR: Please also patch the RECORD file with:
--- a/requests-2.25.1.dist-info/RECORD
+++ b/requests-2.25.1.dist-info/RECORD
@@ -17,7 +17,7 @@
 requests/structures.py,sha256=msAtr9mq1JxHd-JRyiILfdFlpbJwvvFuP3rfUQT_QxE,3005
 requests/utils.py,sha256=_K9AgkN6efPe-a-zgZurXzds5PBC0CzDkyjAE2oCQFQ,30529
 requests-2.25.1.dist-info/LICENSE,sha256=CeipvOyAZxBGUsFoaFqwkx54aPnIKEtm9a5u2uXxEws,10142
-requests-2.25.1.dist-info/METADATA,sha256=fRSAA0u0Bi0heD4zYq91wdNUTJlbzhK6_iDOcRRNDx4,4177
+requests-2.25.1.dist-info/METADATA,sha256=FIRoD6JzkS0gW2tgRJemIR-Dj9FVZPLY7ALIWbib-bg,4280
 requests-2.25.1.dist-info/WHEEL,sha256=Z-nyYpwrcSqxfdux5Mbn_DQ525iP7J2DG3JgGvOYyTQ,110
 requests-2.25.1.dist-info/top_level.txt,sha256=fMSVmHfb5rbGOo6xv-O_tUX6j-WyixssE-SnwcDRxNQ,9
 requests-2.25.1.dist-info/RECORD,,
```
which is exactly the same suggested sha value as on failing Windows tests.

rickeylev (approved):

aignas:
@rickeylev, feel free to merge if you agree with the warning refactor.

python/private/patch_whl.bzl
```diff
@@ -73,11 +73,15 @@ def patch_whl(rctx, *, python_interpreter, whl_path, patches, **kwargs):
         parsed_whl.platform_tag,
     ]))
 
+    record_patch = rctx.path("RECORD.patch")
```
rickeylev, line 76: Does this work if RECORD.patch doesn't exist? In chat the other day, Jeremy linked to an issue where it sounded like the initial path() call would fail. Or maybe that's only if a label is passed? relative paths are OK?

python/private/patch_whl.bzl
```diff
@@ -97,4 +101,19 @@ def patch_whl(rctx, *, python_interpreter, whl_path, patches, **kwargs):
             ),
         )
 
+    if record_patch.exists:
```
rickeylev, line 104: Record patching is effectively required, right, because it has the hashes? I'm just thinking warning about this might be spammy.

rickeylev (approved):

python/private/patch_whl.bzl
rickeylev: resolve in chat. I didn't realize .exists is a function call to stat the path, not a simple attribute lookup from the time path() was called.

python/private/patch_whl.bzl
rickeylev: Resolved in chat

PR #1640, opened 2023-12-20
docs: document support policies

rickeylev:
This just writes down our support policies and puts them in a single location in the hosted
docs. Summarized:

* Bazel version support is as discussed from the maintainers meeting: upcoming,
  current, and last versions
* Reference the Bazel rule compatibility guidelines (always having an incremental
  path to upgrade)
* Described what experimental features mean.
* Only support the latest rules_python version; older ones are best effort.
* Only support platforms CI can run.

Work towards #1361


aignas (approved):
This is great, thank you for bringing clarity to this.

Issue #1638, opened 2023-12-20
Intermediate modules that register the same python version as rules_python's default prevent identifying default toolchain


rickeylev:
# 🐞 bug report

### Affected Rule

`python.toolchain`

### Is this a regression?

Exists as far back as 0.25.0, not sure before that.

### Description

If an intermediate module registers the same python version as rules_python's default, then it prevents a default toolchain from being identified.

## 🔬 Minimal Reproduction

The basic repro is to have an earlier module register the same version as what rules_python sets as the default.

```
# root MODULE
bazel_dep(name="rules_python")

python = use_extension("@rules_python//python/extensions:python.bzl", "python")
# Two must be registered; one is always treated as is_default=True
python.toolchain(python_version="3.11", is_default=False)
python.toolchain(python_version="3.10", is_default=False)

# rules_python MODULE
python = use_extension("@rules_python//python/extensions:python.bzl", "python")
python.toolchain(python_version="3.11", is_default=True)
```

The error/output is:

<pre><code>
... python/extensions/python.bzl:44:10: WARNING: Ignoring toolchain 'python_3_11' from module 'rules_python': Toolchain 'python_3_11' from module 'middle' already registered Python version 3.11 and has precedence
ERROR: Traceback (most recent call last):
	File ".../python/extensions/python.bzl", line 140, column 13, in _python_impl
		fail("No default Python toolchain configured. Is rules_python missing `is_default=True`?")
</code></pre>

**Anything else relevant?**

Blocking rules_proto_grpc getting onto bcr, but also would affect other intermediate modules

rickeylev:
It's late so I gotta pack it in. I think to fix this, the logic needs to do two things:
1. Track what version (instead of toolchain struct) was marked as default. Then later look up the toolchain struct for that python version
2. Before skipping a lower-precedence toolchain entirely, check its default-ness so (1) can work.

PR #1642, opened 2023-12-20
fix(bzlmod): allow modules to register the same toolchain as rules_python's default

rickeylev:
This fixes a bug where, if a module tries to register a non-default toolchain with the
same version as rules_python's default toolchain, an error would occur. This happened
because the earlier (non-default) toolchain caused the later (default) toolchain to
be entirely skipped, and then no default toolchain would be seen. This most affects
intermediary modules that need to register a toolchain, but can't specify a default one.

To fix, just skip creating and registering the duplicate toolchain, but still check
its default-ness to determine if it's the default toolchain.

Fixes https://github.com/bazelbuild/rules_python/issues/1638

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,9 @@ A brief description of the categories of changes:
   specifying a local system interpreter.
 * (bzlmod pip.parse) Requirements files with duplicate entries for the same
   package (e.g. one for the package, one for an extra) now work.
+* (bzlmod python.toolchain) Submodules can now (re)register the Python version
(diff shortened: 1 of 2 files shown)


aignas (approved):

cpsauer:
Hit this bug. Thank you so much for fixing. Is there a release that would include it on the horizon?

rickeylev:
Yes. The PR updating the changlog for 0.28.0 is in the merge queue now, a new release will be made once that's merged

cpsauer:
Amazing. Thank you!

PR #1648, opened 2023-12-21
chore: fix sphinx requirements

aignas:
Before this PR the requirements files for `osx` and `linux` would be
split, but in practise they would be always the same. This means that
`linux` users would have to ask an `osx` user to run a command where
we could actually just reuse the same files.

Summary:
- chore: add absl-py to `osx` requirements
- chore: remove the `osx` requirements and rename `requirements_linux.txt`
  to just be `requirements.txt` as both of the files are the same.


groodt (approved):
LGTM

Issue #1600, opened 2023-12-08
Bazel coverage report file is empty when loading py_binary with Python version specified


tanan:
# 🐞 bug report

### Affected Rule

The issue is caused by the `py_binary` and `py_test` when running `bazel coverage`.

### Is this a regression?

No.

### Description

When I load py_binary and py_test with Python version specified like below and run `bazel coverage` command, a coverage report file is empty.

```
load("@python//3.10:defs.bzl", "py_binary", "py_test")
```

When I load these rules with the default Python version, the correct coverage report is created.

```
load("@rules_python//python:defs.bzl", "py_binary", "py_test")
```

## 🔬 Minimal Reproduction

Here is an example repository.
https://github.com/tanan/bazel-python-coverage/tree/python-coverage-empty

I run the command:

```
bazel coverage --combined_report=lcov //apps/demo/...
```

On the other hand, `bazel coverage` creates a correct file when I load the default version of `py_binary` and `py_test`. The sample repository is here:
https://github.com/tanan/bazel-python-coverage/tree/python-coverage-success

The difference between the two codes is the URL below.
https://github.com/tanan/bazel-python-coverage/compare/python-coverage-success...python-coverage-empty

## 🔥 Exception or Error

coverage file is empty with the log `WARNING: There was no coverage found`.

<pre><code>
$ bazel coverage --combined_report=lcov //apps/demo/...
Starting local Bazel server and connecting to it...
INFO: Using default value for --instrumentation_filter: "^//apps/demo[/:]".
INFO: Override the above default with --instrumentation_filter
INFO: Analyzed 3 targets (54 packages loaded, 3852 targets configured).
INFO: Found 2 targets and 1 test target...
INFO: LCOV coverage report is located at /home/foo/.cache/bazel/_bazel_foo/eb400fbb2bc2347b07965f1284d95f71/execroot/__main__/bazel-out/_coverage/_coverage_report.dat
 and execpath is bazel-out/_coverage/_coverage_report.dat
INFO: From Coverage report generation:
Dec 08, 2023 6:12:14 PM com.google.devtools.coverageoutputgenerator.Main getTracefiles
INFO: Found 1 tracefiles.
Dec 08, 2023 6:12:14 PM com.google.devtools.coverageoutputgenerator.Main parseFilesSequentially
INFO: Parsing file bazel-out/k8-fastbuild-ST-eff9b1c97e51/testlogs/apps/demo/lib_test/coverage.dat
[...]

PR #1649, opened 2023-12-21
fix(coverage): add test attributes in transition module

tanan:
This PR is to fix an issue that coverage report is empty when using transition module.
This is due to the absence of the `_lcov_merger` and `_collect_cc_coverage` attributes.
Coverage reports will be created adding these attributes.

Fixes https://github.com/bazelbuild/rules_python/issues/1600


CHANGELOG.md
```diff
@@ -46,6 +46,8 @@ A brief description of the categories of changes:
   specifying a local system interpreter.
 * (bzlmod pip.parse) Requirements files with duplicate entries for the same
   package (e.g. one for the package, one for an extra) now work.
+* (coverage): `_lcov_merger` attr added in `_transition_py_test` and
+  coverage reports are now created when using the transition module.
```
```suggestion
* (coverage): coverage reports are now created when the version-aware
  rules are used.
  ([#1600](https://github.com/bazelbuild/rules_python/issues/1600))
```

python/config_settings/transition.bzl
```diff
@@ -142,16 +142,29 @@ _COMMON_ATTRS = {
...
+        default = "@bazel_tools//tools/test:collect_cc_coverage",
+        executable = True,
+        cfg = "exec",
+    ),
+    "_lcov_merger": attr.label(
+        default = configuration_field(fragment = "coverage", name = "output_generator"),
+        executable = True,
+        cfg = "exec",
+    ),
+}
```
rickeylev, line 156: Yay magic attributes. I'm guessing you copied this from the files in python/private/common? Lets also copy the comment over so we can give what little context we have about them. Based on your analysis in the issue, I updated the text a bit.

```suggestion
_PY_TEST_ATTRS = {
    # Magic attribute to help C++ coverage work. There's no
    # docs about this; see TestActionBuilder.java
    "_collect_cc_coverage": attr.label(
        default = "@bazel_tools//tools/test:collect_cc_coverage",
        executable = True,
        cfg = "exec",
    ),
    # Magic attribute to make coverage work. There's no
    # docs about this; see TestActionBuilder.java
    "_lcov_merger": attr.label(
        default = configuration_field(fragment = "coverage", name = "output_generator"),
        executable = True,
        cfg = "exec",
    ),
}
```

rickeylev (changes requested):
Just some textual updates, otherwise LGTM.

I'd ask for a test, but I don't think we have any infrastructure setup to verify the coverage command is working.

python/config_settings/transition.bzl
```diff
@@ -142,16 +142,29 @@ _COMMON_ATTRS = {
...
+
 _transition_py_binary = rule(
     _transition_py_impl,
-    attrs = _COMMON_ATTRS,
+    attrs = _COMMON_ATTRS | _PY_TEST_ATTRS,
```
aignas, line 160: Why is it needed in both, `py_binary` and `py_test` as opposed just `py_test`?

tanan: @aignas My understanding is that code coverage is collected from `py_binary` targets as well.
One of the reference is in https://bazel.build/reference/be/python#py_runtime.coverage_tool

These attributes are defined in common `py_binary` rule in this repo.
https://github.com/bazelbuild/rules_python/blob/main/python/private/common/py_binary_rule_bazel.bzl

python/config_settings/transition.bzl
tanan: Thank you for your suggestion.
These attributes are also written In [python/private/common/py_binary_rule_bazel.bzl](https://github.com/bazelbuild/rules_python/blob/main/python/private/common/py_binary_rule_bazel.bzl), but there is no comment.
So, I'll copy the comment in it as well.

python/config_settings/transition.bzl
aignas: Ah, thank you, makes sense, just wanted to double check.

aignas (approved):

python/config_settings/transition.bzl
rickeylev: Yeah, if a test runs a binary, then the binary needs to be instrumented so it can produce the coverage rules (which the outer test can collect and process)

rickeylev (approved):

PR #1654, opened 2023-12-24
feat(runfiles): Added type hints to `@rules_python//python/runfiles`

UebelAndre:
This change adds mypy types directly to the `rules_python.python.runfiles` library as well as regression testing to CI.

### DIFF
--- a/.github/workflows/mypy.yaml
+++ b/.github/workflows/mypy.yaml
@@ -1,0 +1,32 @@
+name: mypy
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    types:
+      - opened
+      - synchronize
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  ci:
+    runs-on: ubuntu-20.04
+    steps:
+      # Checkout the code
+      - uses: actions/checkout@v2
+      - uses: jpetrucciani/mypy-check@master
+        with:
+          requirements: 1.6.0
+          python_version: 3.11
+          path: 'python/runfiles'
+      - uses: jpetrucciani/mypy-check@master
+        with:
+          requirements: 1.6.0
+          python_version: 3.11
+          path: 'tests/runfiles'
+      
--- a/.github/workflows/mypy.yaml
+++ b/.github/workflows/mypy.yaml
@@ -22,11 +22,11 @@ jobs:
       - uses: jpetrucciani/mypy-check@master
         with:
           requirements: 1.6.0
-          python_version: 3.11
+          python_version: 3.8
           path: 'python/runfiles'
       - uses: jpetrucciani/mypy-check@master
         with:
           requirements: 1.6.0
-          python_version: 3.11
+          python_version: 3.8
           path: 'tests/runfiles'
       
--- a/python/runfiles/runfiles.py
+++ b/python/runfiles/runfiles.py
@@ -14,82 +14,116 @@
 
 """Runfiles lookup library for Bazel-built Python binaries and tests.
 
-See README.md for usage instructions.
+See @rules_python//python/runfiles/README.rst for usage instructions.
 """
 import inspect
 import os
 import posixpath
 import sys
-
-if False:
-    # Mypy needs these symbols imported, but since they only exist in python 3.5+,
-    # this import may fail at runtime. Luckily mypy can follow this conditional import.
-    from typing import Callable, Dict, Optional, Tuple, Union
+from typing import Dict, Optional, Tuple, Union
 
 
-def CreateManifestBased(manifest_path):
-    # type: (str) -> _Runfiles
-    return _Runfiles(_ManifestBased(manifest_path))
+class _ManifestBased:
+    """`Runfiles` strategy that parses a runfiles-manifest to look up runfiles."""
+
+    def __init__(self, path: str) -> None:
+        if not path:
+            raise ValueError()
+        if not isinstance(path, str):
+            raise TypeError()
+        self._path = path
+        self._runfiles = _ManifestBased._LoadRunfiles(path)
+
+    def RlocationChecked(self, path: str) -> Optional[str]:
+        """Returns the runtime path of a runfile."""
+        exact_match = self._runfiles.get(path)
+        if exact_match:
+            return exact_match
+        # If path references a runfile that lies under a directory that
+        # itself is a runfile, then only the directory is listed in the
+        # manifest. Look up all prefixes of path in the manifest and append
+        # the relative path from the prefix to the looked up path.
+        prefix_end = len(path)
+        while True:
+            prefix_end = path.rfind("/", 0, prefix_end - 1)
+            if prefix_end == -1:
+                return None
+            prefix_match = self._runfiles.get(path[0:prefix_end])
+            if prefix_match:
+                return prefix_match + "/" + path[prefix_end + 1 :]
+
+    @staticmethod
+    def _LoadRunfiles(path: str) -> Dict[str, str]:
+        """Loads the runfiles manifest."""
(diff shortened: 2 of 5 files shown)


python/runfiles/runfiles.py
```diff
@@ -14,82 +14,117 @@
...
-    # Mypy needs these symbols imported, but since they only exist in python 3.5+,
-    # this import may fail at runtime. Luckily mypy can follow this conditional import.
-    from typing import Callable, Dict, Optional, Tuple, Union
-
-
-def CreateManifestBased(manifest_path):
-    # type: (str) -> _Runfiles
-    return _Runfiles(_ManifestBased(manifest_path))
 
+class _ManifestBased:
```
UebelAndre, line 26: These must be defined earlier to be able to use them as types

python/runfiles/runfiles.py
```diff
@@ -14,82 +14,117 @@
...
+            # TODO(laszlocsomor): remove JAVA_RUNFILES once the Java launcher can
+            # pick up RUNFILES_DIR.
+            "JAVA_RUNFILES": directory,
+        }
 
-    Otherwise, if `env` contains "RUNFILES_DIR" with non-empty value (checked in
-    this priority order), this method returns a directory-based implementation.
 
-    If neither cases apply, this method returns null.
+class _DirectoryBased:
```
UebelAndre, line 89: Same as https://github.com/bazelbuild/rules_python/pull/1654/files#r1435762546

python/runfiles/runfiles.py
aignas: nit: I thought we could just use a string for the type annotation and I would have assumed that this would have worked irrespectively of where in the file they are written.

python/runfiles/runfiles.py
```diff
@@ -14,82 +14,116 @@
...
-class _Runfiles(object):
+class Runfiles:
     """Returns the runtime location of runfiles.
 
     Runfiles are data-dependencies of Bazel-built binaries and tests.
     """
 
-    def __init__(self, strategy):
-        # type: (Union[_ManifestBased, _DirectoryBased]) -> None
+    def __init__(self, strategy: Union[_ManifestBased, _DirectoryBased]) -> None:
```
aignas, line 119: Related to my comment above, does the following work if the `_ManifestBased` is below this line?
```suggestion
    def __init__(self, strategy: Union["_ManifestBased", "_DirectoryBased"]) -> None:
```

.github/workflows/mypy.yaml
```diff
@@ -0,0 +1,32 @@
...
+jobs:
+  ci:
+    runs-on: ubuntu-20.04
+    steps:
+      # Checkout the code
+      - uses: actions/checkout@v2
+      - uses: jpetrucciani/mypy-check@master
+        with:
+          requirements: 1.6.0
+          python_version: 3.11
```
aignas, line 25: it would be great to test against Python `3.8`, which does not have the new style annotations to ensure backwards compatibility with previous Python versions. WDYT?

python/runfiles/runfiles.py
UebelAndre: Isn’t it simpler to reorder them in the long run? I agree it’s annoyingly noisy but I think the outcome will benefit future maintainers and contributors.

python/runfiles/runfiles.py
UebelAndre: https://github.com/bazelbuild/rules_python/pull/1654#discussion_r1435943418

If this is a nit I’d push back since I feel this is the right thing to do. If this blocks merging then I can reluctantly try to put this back. Just let me know what the position is

.github/workflows/mypy.yaml
UebelAndre: Yeah makes sense. I'll move this back to 3.8

python/runfiles/runfiles.py
aignas: Having to move the code around just to add type annotations is surprising and for ease of review it would be good to split such a change into two - one to add type annotations (but maybe use strings in places as a workaround for Python). That said, I am OK to review this as one PR, I just need more time to do it :)

If you think that the end goal should be the place where we don't use strings, I don't have any objections to this.

UebelAndre: I think the end goal is not to use strings. If the review seems too much then I can try to reduce splash damage but it's still mostly code moves and then tweaks to function signatures. Let me know though!

python/runfiles/runfiles.py
```diff
@@ -156,11 +190,13 @@ def Rlocation(self, path, source_repo=None):
         # target_repo is an apparent repository name. Look up the corresponding
         # canonical repository name with respect to the current repository,
         # identified by its canonical name.
-        target_canonical = self._repo_mapping[(source_repo, target_repo)]
-        return self._strategy.RlocationChecked(target_canonical + "/" + remainder)
+        if source_repo is not None:
```
aignas, line 193: This `if` statement should always be `True` and it would be better to add an assert there. On line `180` there is
`if not remainder or (source_repo, target_repo) not in self._repo_mapping`. This means that if there is no remainder, then the code will return there, but if there is, then the `(None, target_repo) not in some_dict` should be always true because I would expect that value to never be present.

tests/runfiles/runfiles_test.py
```diff
@@ -126,23 +131,25 @@ def testManifestBasedRunfilesEnvVars(self):
...
 
-    def testCreatesDirectoryBasedRunfiles(self):
+    def testCreatesDirectoryBasedRunfiles(self) -> None:
         r = runfiles.Create(
             {
                 "RUNFILES_DIR": "runfiles/dir",
                 "TEST_SRCDIR": "always ignored",
             }
         )
+        assert r is not None
```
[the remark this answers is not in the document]
UebelAndre: I originally had this but it mypy would immediately complain since it doesn't seem to understand the API. I could do this and then use a `cast` but I felt this was the lesser evil. I don't really care which it is though so with that said, which would you prefer? or is there a better path?

python/runfiles/runfiles.py
UebelAndre: I can promote this to an assertion

tests/runfiles/runfiles_test.py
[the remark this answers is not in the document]
aignas: SGTM to leave the `assert r is not None` but it may be nice to add a comment as to why we are not using the `unittest` API.

python/runfiles/runfiles.py
```diff
@@ -153,14 +187,15 @@ def Rlocation(self, path, source_repo=None):
             #   which also should not be mapped.
             return self._strategy.RlocationChecked(path)
 
+        assert source_repo is not None
```
aignas, line 190: consider adding an explanation to the `assert`. If this assert fails, having no message may require the user to read the code. Maybe something like the following could work.

```suggestion
        assert source_repo is not None, f"BUG: if the `source_repo` is None, we should never go past the `if` statement above"
```

aignas (commented):
For me this looks good, but I'd like to wait for @rickeylev to have a look at this since this would constrain the `runfiles` library to Python `3.7` and I don't know if this is something we want to do now.

UebelAndre:
For what it’s worth, 3.7 was chosen because it’s EOL and I thought it would be a sufficiently low bar.

tests/runfiles/runfiles_test.py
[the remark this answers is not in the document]
UebelAndre: Done!

aignas (approved):
LGTM, thanks

PR #1656, opened 2023-12-27
feat(runfiles): Add static methods to `Runfiles` class to simplify interface

UebelAndre:
With this change we allow for more ergonomic creation of the runfiles
object and using type annotations:
```python
from python.runfiles import Runfiles

def main() -> None:
    runfiles: Runfiles | None = Runfiles.Create()
```

Furthermore, the docs have been updated to describe pulling `runfiles`
library directly from `rules_python` to avoid the complexity of needing
to setup `pip_parse` to use this behavior.


UebelAndre:
cc @aignas is this one ok as well?

aignas (commented):
I'll have a look at it later early next week.

python/runfiles/README.md
```diff
@@ -0,0 +1,64 @@
+# bazel-runfiles library
+
+This is a Bazel Runfiles lookup library for Bazel-built Python binaries and tests.
+
+Learn about runfiles: read [Runfiles guide](https://bazel.build/extending/rules#runfiles)
+or watch [Fabian's BazelCon talk](https://www.youtube.com/watch?v=5NbgUMH1OGo).
+
+## Typical Usage
+
+1. Depend on this runfiles library from your build rule, like you would other third-party libraries:
```
aignas, line 10: I think making a not that projects can consume the PyPI parkage would be good. @rickeylev, do you remember the reasons why getting the runfiles lib from PyPI was the preferred approach?

aignas (commented):
I personally don't like the Create method, because it smells of bad polymorphism design, but at the same time i think that allowing easy type annotations is a not win. @rickeylev, what do you think?

python/runfiles/README.md
rickeylev: It's hard to say if one is strictly better than the other.

Pulling from pypi means the runfiles library version is decoupled from the rules version. This might be good -- you could use a newer version of runfiles to pick up a bug fix. Or it could be bad -- if we do something to couple the runfiles code to the rules logic (not sure what that would be, though). Both options could do more advanced things like generate pyc files, use c extensions, etc, so its just a question as to when those build steps happen (pre-built or on-demand).

That pip_parse needs to be setup seems like NBD. Most programs have _some_ external dependency, so you're probably going to have to do that no matter what.

Either way, I think we should document both. They're both public APIs we support, so might as well write them down.

UebelAndre: Updated the readme! Let me know what you think

python/runfiles/README.md
```diff
@@ -0,0 +1,90 @@
...
+r2 = Runfiles.CreateDirectoryBased("path/to/foo.runfiles/")
+```
+
+If you want to start subprocesses, and the subprocess can't automatically
+find the correct runfiles directory, you can explicitly set the right
+environment variables for them:
+
+```python
+import subprocess
+from rules_python.python.runfiles import Runfiles
```
aignas, line 79: FYI, within bzlmod this import isn't working. If we strip the rules_python segment it should work in both.

aignas (changes requested):

python/runfiles/README.md
UebelAndre: Why doesn’t that work? That’s gonna be a pretty big change to make :(

aignas: That is because the name of the directory that contains the library in bzlmod is not rules_python, but something else. Documenting that the import path should be `python.runfiles` for now should be good enough for now. 

If we want a specific import path to be `bazel_runfiles` instead or make `rules_python.python.runfiles` work in bzlmod, we should probably address this in a separate PR. Relying on the repository name in python imports tends to be brittle over time and standardising on a single import that is consistent with the PyPI library would be better, but I would argue that can be done as a separate PR - adding a folder bazel_runfiles under the python/runfiles dir and adding the necessary re-exports and aliases to it is fairly straightforward if that's the path we want nto go with.

aignas: It looks like this is still the same? However the PR LGTM in general.

UebelAndre: I changed the import path in docs. If that’s not what you meant can you be more specific?

aignas: I thought I still saw the import with rules_python in the subprocess example. I see that tho commit changed the docs elsewhere?

UebelAndre:
If there's no blockers is this something that could be merged soon? I think it'd be good to consolidate the import paths for pypi releases and rules_python releases but that should be a separate change.

cc @aignas @rickeylev

python/runfiles/README.md
aignas:
```suggestion
from python.runfiles import Runfiles
```

aignas (approved):

aignas:
@UebelAndre, just came back from holiday.

Sorry for being pedantic, but as per the PR template todo checklist, could you please add a `CHANGELOG.md` and make the PR description more like a commit message? Otherwise the change looks good to me.

UebelAndre:
> @UebelAndre, just came back from holiday.
> 
> Sorry for being pedantic, but as per the PR template todo checklist, could you please add a `CHANGELOG.md` and make the PR description more like a commit message? Otherwise the change looks good to me.

What do you mean "more like a commit message"?

aignas:
Condensed the message slightly and used passive voice and reworded to sound more like what is being added/changed with the commit rather than what is proposed. Thanks a lot for the improvements to `rules_python`!

aignas (approved):

PR #1657, opened 2023-12-28
refactor(pip_install): make the API surface of deps parsing smaller

aignas:
With this change we also fix a minor bug where the self edge dependencies
within the wheels are correctly handled as per unit tests.

Also refactor the code to use the fact that the selects are chosen
based on the specificity of the condition. This can make the addition
of deps incrementally correct without the need for extra processing in
the `build` method. This should hopefully make the implementation more
future-proof as it makes use of these generic concepts and is likely to
help to extend this code to work with selecting on multiple Python versions.

I have added a test to ensure coverage of the new code.


aignas:
When you have time, please take a look @rickeylev. My end-goal would be to support parsing the metadata to support select statements with version aware python toolchain config settings, which should hopefully land in another PR.

rickeylev (commented):
Sorry, ran out of time tonight to review in full; only got like 10 lines into it.

internal_deps.bzl
```diff
@@ -213,3 +213,13 @@ def rules_python_internal_deps():
         ],
         sha256 = "4531deccb913639c30e5c7512a054d5d875698daeb75d8cf90f284375fe7c360",
     )
+
+    # Needed for tests for parsing whl METADATA
```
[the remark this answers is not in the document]
aignas: It was noted as an example with having self references which can change the requested extras. I guess we can retain a comment to this example, but  we can use just hardcoded testcases.

python/pip_install/tools/wheel_installer/wheel.py
```diff
@@ -228,14 +234,27 @@ class Deps:
...
         name: str,
+        *,
+        requires_dist: Optional[List[str]],
         extras: Optional[Set[str]] = None,
         platforms: Optional[Set[Platform]] = None,
     ):
         self.name: str = Deps._normalize(name)
+        self._platforms: Set[Platform] = platforms or set()
+
+        reqs = sorted(
```
rickeylev, line 245: nit: Is this sorted call for correctness, determinism, or just being defensive? Please drop a brief comment as to why.

python/pip_install/tools/wheel_installer/wheel.py
```diff
@@ -228,14 +234,27 @@ class Deps:
...
+            [Requirement(wheel_req) for wheel_req in requires_dist],
```
rickeylev, line 246: nit: the inner list comp can be dropped; a generator can be sorted

python/pip_install/tools/wheel_installer/wheel.py
```diff
@@ -228,14 +234,27 @@ class Deps:
...
+            key=lambda x: f"{x.name}:{sorted(x.extras)}",
+        )
+
+        # Resolve any extra extras due to self-edges, empty string means not extras
```
rickeylev, line 250: typo
```suggestion
        # Resolve any extra extras due to self-edges, empty string means no extras
```

rickeylev, line 250: I think I commented on the empty string thing in an earlier PR; this empty string thing is confusing.

Does the presence of the empty string mean no extras? i.e. are `{"", "foo"}` Or is it assumed that an empty string element is mutually exclusive with other elements? Why is "no extras" communicated using set-with-only-empty-string as opposed to empty-set ?

(I've authored and had to maintain code where None vs empty set vs non-empty set have 3 different meanings, and it was always a cognitive burden)

python/pip_install/tools/wheel_installer/wheel.py
```diff
@@ -245,26 +264,33 @@ def _add(self, dep: str, platform: Optional[Platform]):
...
-    def add(self, *wheel_reqs: str) -> None:
-        reqs = [Requirement(wheel_req) for wheel_req in wheel_reqs]
+        # Add the dep to more specific constraints
+        for p in platform:
+            if p not in self._select:
+                continue
 
-        # Resolve any extra extras due to self-edges
-        self._want_extras = self._resolve_extras(reqs)
+            self._select[p].add(dep)
```
rickeylev, line 279: The `if p not in self._select` part of this block of code is confusing me.

The iteration will produce one of two list of values:
1. Exactly one element: self (may or may not have arch set)
2. Two or more elements: self (without an arch) and one or more with arch set.

In case one, L279 duplicates what L272 did: the loop is one element long, p is present, and L279 executes, and we're done.

In case two, the same thing happens: the loop is 1 + N long. The first element does L279 (duplicating L272), and then subsequent iterations do nothing (the arch-bound platform isn't present).

What am I missing? That some earlier add_req() call added an entry? But then wouldn't that mean some of these were ignored when they shouldn't have been...? or maybe the loop below has something to do with this? I think if I could see the requirements lines being processed it'd be easier for me to reason about.

rickeylev (approved):
I'm guessing the loop I'm confused by (and looks like a bug?) is just me not understanding the flow. So I'll approve. But i would like to wrap my head around it a bit.

Rest of my comments are mostly nits/doc requests.

python/pip_install/tools/wheel_installer/wheel.py
aignas: Added a unit-test and more comments to hopefully make the code more self-explanatory.

python/pip_install/tools/wheel_installer/wheel.py
aignas: Right now the case of `{"", "foo"}` and `{"foo"}` has the same meaning. I moved this detail into the method implementation to better isolate it from the rest of the code as it is an implementation detail. The return value of `_resolve_extras` should be always a set with at least one element in it and it just makes the matching with `any(...)` easier to write.

python/pip_install/tools/wheel_installer/wheel.py
```diff
@@ -86,6 +86,12 @@ def host(cls) -> List["Platform"]:
             )
         ]
 
+    def __iter__(self) -> Iterator["Platform"]:
```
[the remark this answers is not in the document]
aignas: Thanks, I liked the last one as well.

PR #1662, opened 2024-01-02
fix: ignore_root_user_error should be taken from root module only

### LINKED ISSUE #1658: setting ignore_root_user_error to True in root module is not enough to avoid the root user error (bzlmod)
# 🐞 bug report

### Affected Rule

 The issue is caused by the rule: rule_python, python extension.

### Description

 
Even when setting `ignore_root_user_error = True` in my root module, I still get the:
`fail("The current user is root, please run as non-root when using the hermetic Python interpreter. See https://github.com/bazelbuild/rules_python/pull/713.")` when using a submodule that does not have the ignore_root_user_error attr set to True. 

Shouldn't it be such that the root module decides that for the entire dep graph ? Similar to the `is_default` attribute ?

**Rules_python version:**
`0.27.1`

shabanzd:
Only the root module should be able to decide `ignore_root_user_error`. Modules being depended upon don't know the final environment, so they aren't in the right position to know or decide what the correct setting is. This PR implements a solution to take the `ignore_root_user_error` value from the root module only.

Fixes #1658


python/private/bzlmod/python.bzl
```diff
@@ -104,22 +104,28 @@ def _python_impl(module_ctx):
...
+                # * It prevents submodules from specifying different defaults and only
+                #   one of them winning.
+                # * rules_python needs to set a soft default in case the root module doesn't,
+                #   e.g. if the root module doesn't use Python itself.
+                # * The root module is allowed to override the rules_python default.
                 # A single toolchain is treated as the default because it's unambiguous.
                 is_default = toolchain_attr.is_default or len(mod.tags.toolchain) == 1
+                # Only the root module should have a say on whether or not to ignore the root user 
+                # error. This can be achieved by taking the ignore_root_user_error for the root 
+                # module. For all other modules, the ignore_root_user_error attr is set to True.
```
aignas, line 119: Should it be set to the same value as the root module? This comment suggests that only the root module can toggle the behaviour, but the non-root modules will have the toggle set to True. I am wondering, what would be the benefit for setting the non-root module root warning to True by default.

aignas (commented):
Thanks for the PR!

python/private/bzlmod/python.bzl
ziad-fernride: as far as I understand, [this](https://github.com/bazelbuild/rules_python/blob/main/python/private/bzlmod/python.bzl#L132) will be called over and over again for every module. Meaning that the check will be run for every toolchain in every module. 
Since there is a [guarantee](https://bazel.build/rules/lib/builtins/module_ctx#modules) that the root module will be checked first, the only way the check will even reach submodules, is if all toolchains in the root module have  `ignore_root_user_error` set to `True`. Which is why it is safe to assume that setting `ignore_root_user_error` to `True` is rendering the submodules redundant in the decision.

ziad-fernride: The reason why I did not simply set it equal to the root module, is that the root module can have multiple toolchains, meaning that the flag can be set to different values for different toolchains, and one `False` value is enough to hit the root_user error.
I could have done something like:
ignore_root_user_error = ignore_root_user_error_toolchain1 && ignore_root_user_error_toolchain2 && ignore_root_user_error_toolchainN I guess, but the result is the same as above.

rickeylev: > I am wondering, what would be the benefit for setting the non-root module root warning to True by default.

To clarify (double negatives at play): you mean set `ignore_root_user_error=True` by default? The current default is False.

Following the code comments and PR comments, setting it to True makes the runtime directory non-deterministic because of pyc file creation. This "ignore_root_user_error" name is pretty misleading. A more accurate name would be something like "allow_runtime_to_create_nondeterministic_stdlib_pyc_files_upon_first_use".

This setting controls whether the runtime is made read-only. If it's made readonly, then the runtime can't create pyc files for the stdlib files, and there isn't a nondeterministic pyc problem. It probably also prevents the issue we see on the windows CI machines where the set of files in the runtime is changed mid-execution and/or intermittently picking up the temp pyc files mid-execution.

So I would say making `ignore_root_user_error=True` the default would be a bad idea.

The name of this setting and the logic it affects is kind of strange. I would think it would just change the fail() calls to a print() calls? I wonder if something got lost in some other fix or refactoring...

rickeylev:
Also, remember the root might might not explicitly setup any python toolchains; it might only be getting one because of some transitive module that needs python.

And yes, only the root module should be able to set `ignore_root_user_error`. A module being used as a library doesn't know what the final runtime environment is, so it can't know what the correct decision is.

python/private/bzlmod/python.bzl
ziad-fernride: I was not trying to make the` ignore_root_user_error=True` by default, I was simply setting the `ignore_root_user=True` in the submodules only ... Assuming there is at least one toolchain registered in the root module,  setting ` ignore_root_user_error=True`  in all toolchains in the submodules is equivalent to only allowing the root module to set the` ignore_root_user_error`, since one tool chain with ` ignore_root_user_error=False` in the root module is enough to trigger the root user error. 

What I did not think of tho, is the comment you have below 😆 it is of course true that the root module may not register any toolchains, and then this logic will default to True (which is not correct). I will update the code to handle this case and come back 😄 

Thanks for the feedback ! 🙏

aignas (approved):
LGTM, but it needs a CHANGELOG snippet.

aignas:
@shabanzd could you think of a way to add an integration test for this? The `main` branch has support for `bazel-in-bazel` integration tests now and you could add one in `tests/integration` that would exercise the modified code.

rickeylev:
I originally requested tests when Ignas asked me about this PR, but I'm not so sure we should block on that now. The only idea I have for testing this is to add some print statements and check their output in the integration test. The prints would be guarded by an environment variable. However, doing that requires setting up a new test runner for the integration tests (one that can capture output and assert on it), which is maybe a bit too much to ask a new contributor in this PR.

Any alternative testing ideas that might be easier?

ziad-fernride:
I would agree with @aignas that maybe the simplest way would be to have `bazel-in-bazel` integration tests with different scenarios and check for failures/successes based on the scenario ?

rickeylev:
Ah, that gives me an idea: write an extra file somewhere with some info about what the extension did. Guard this behind an environment variable. Then we can write a test inside the integration test that reads that file and verifies its content. Maybe have the python repository repo rule write out what its settings were? Then it should be easy to use_repo() it and locate the file within it.

rickeylev (approved):
I added an integration test to verify behavior. It works as I described last: the extension checks an environment variable and writes a file with some debug info. A test reads it to verify what happened. Overall it works pretty well and isn't very invasive; we can probably use this same technique to test other subtle cases with the bzlmod extension code.

shabanzd:
Sorry was offline since Thursday, thanks for taking care of the testing @rickeylev 🙏🏼🙏🏼

PR #1671, opened 2024-01-07
feat: add interpreter_version_info to py_runtime

mattem:
Adds an `interpreter_version_info` attribute to the `py_runtime` and associated provider that maps to the `sys.version_info` values. This allows the version of the interpreter to be known statically, which can be useful for rule sets that depend on the interpreter, and need to build environments / pathing that contain version info (virtualenvs for example).


python/private/common/py_runtime_rule.bzl
```diff
@@ -61,6 +61,10 @@ def _py_runtime_impl(ctx):
 
     python_version = ctx.attr.python_version
 
+    interpreter_version_info = ctx.attr.interpreter_version_info
+    if interpreter_version_info and len(interpreter_version_info) < 2:
```
mattem, line 65: We can expand this to include micro if we want

python/private/common/providers.bzl
```diff
@@ -136,6 +138,11 @@ the same conventions as the standard CPython interpreter.
             "filesystem path to the interpreter on the target platform. " +
             "Otherwise, this is `None`."
         ),
+        "interpreter_version_info": (
```
mattem, line 141: Let the bikeshedding on the name commence...

python/private/common/py_runtime_rule.bzl
```diff
@@ -196,6 +206,14 @@ platform runtime this attribute must not be set.
 For a platform runtime, this is the absolute path of a Python interpreter on
 the target platform. For an in-build runtime this attribute must not be set.
 """),
+        "interpreter_version_info": attr.string_list(
+            doc = """
+This is the version of the interpreter that this runtime provides.
+The list can contain the five components of the version number: major, minor, micro, releaselevel, and serial.
+This should match the format given by `sys.version_info`, however for simplicity the micro, releaselevel and serial values may be omitted.
```
f0rmiga, line 213:
```suggestion
This should match the format given by `sys.version_info`, however for simplicity, the micro, releaselevel and serial values may be omitted.
```

f0rmiga (approved):
I'll let other folks bikeshed if they want.

python/private/common/providers.bzl
rickeylev: I'm OK with `interpreter_version_info`, as that is what it's representing. `python_version_info` or `version_info` would also be OK IMHO.

rickeylev (changes requested):
My main comment is to use a dict/struct instead of list. This makes it easier to know which parts are check for the relevant pieces. Plus we can stash extra implementation-specific parts in there if necessary.

i.e. accept a attr.string_dict on the rule and and convert it to a struct() for the provider. Convert particular keys to their appropriate types (e.g. major/minor/patch to ints, similar to sys.version_info)

Please add tests. You can add them to //tests/py_runtime

mattem:
@rickeylev Can I have another pass on this please?

rickeylev:
Looks pretty good! Thank you for adding tests.

Can you please add an entry to CHANGELOG.md? Then it's ready to merge.

rickeylev (approved):

PR #1680, opened 2024-01-10
chore: remove MODULE.bazel.lock and ignore it

aignas:
When we use the latest `bazel` release in the `rules_python`
module, then the lock file gets created, but I am not sure if
we should commit it in. Since the lock file is very useful
for local development as it speeds up dependency refetching,
this PR disables it as advised in the issue below. Whilst at
it, the `bzlmod` example also removes it for smaller diffs
when we develop extensions.

See bazelbuild/bazel#20369


rickeylev (approved):
Can you update the change description to better explain why? Is it just to avoid the potential for the merge conflits? That's fine; I can get on board with that (such conflicts are really annoying), even though regenerating after a conflict should be pretty easy and cheap since our project is small.

PR #1681, opened 2024-01-11
fix(gazelle): Cease write narrow visibilities

aryamccarthy:
This PR causes Gazelle to not write `visibility` lines for generated rules.

# Context

This is necessary in monorepo situations where there are multiple `# gazelle:python_root` directives used, e.g. in the structure below:

```
ROOT
    proj_1
        src  # is a python root and depends on proj_2
            foo
    pro_2
        src  # is a python root
            bar  ← The visibility here will be incorrect; proj_1 cannot see this.
```

As it stands, we write a narrowly scoped `visibility` attribute for every generated rule; the visibility is `//{python-root-package}:__subpackages__`. Let's not!

---

# Alternatives

Two considered alternatives were:
* Have the Gazelle Python extension respect the `# gazelle:default_visibility` directive. This is probably correct in the long run, but I'm not Go-literate.
* Set all visibilities to `//visibility:public`. This achieves correct behavior but gives less control than the current PR. Instead of respecting any default visibility stanzas, it forces the entirety to be public.

---

# Preserving current behavior

If someone wants the current behavior, it can still be achieved by setting the following in Python root's BUILD file:
```python
package(
    default_visibility = ["//" + package_name() + ":__subpackages__"],
)
```

### DIFF
--- a/gazelle/python/generate.go
+++ b/gazelle/python/generate.go
@@ -212,7 +212,7 @@ func (py *Python) GenerateRules(args language.GenerateArgs) language.GenerateRes
 	}
 
 	parser := newPython3Parser(args.Config.RepoRoot, args.Rel, cfg.IgnoresDependency)
-	visibility := fmt.Sprintf("//%s:__subpackages__", pythonProjectRoot)
+	visibility := "//visibility:public"
 
 	var result language.GenerateResult
 	result.Gen = make([]*rule.Rule, 0)


f0rmiga:
This is very tragic in many ways. Do you want to have less working setting visibility and then set everything to public? You are defeating the purpose of the visibility attribute and wanting everyone to defeat it with you!

Open an issue first to discuss the alternatives, but setting everything to public is most likely not going to happen.

aryamccarthy:
I absolutely don't want to set everything to `//visibility:public`! But in the base implementation, visibilities are written per-rule, using a hardcoded value in Go. There's no way to inherit defaults from the package or up the tree.

If I have a rule like this in `proj_2`, it should be respected:

```
package(
    default_visibility = ["proj_1:__subpackages__", "proj_2:__subpackages__"],
)
```
This isn't public: `proj_3` and `proj_4` can't see this.

I think that not writing a `visibility` attribute is safer here than setting to public.

aryamccarthy:
Okay, I see our problem. I had uploaded an older commit. The latest version does what the explanation above describes.

Issue #1682, opened 2024-01-11
gazelle: Visibilities are incorrect for projects with multiple `# gazelle:python_root` directives


aryamccarthy:
# 🐞 bug report

### Affected Rule

 The issue is caused by the rule: https://github.com/bazelbuild/rules_python/blob/b106f91c9da7e31d73a7293e88ac78eedcad2057/gazelle/python/generate.go#L215

### Is this a regression?

 No

### Description

This is necessary in monorepo situations where there are multiple `# gazelle:python_root` directives used, e.g. in the structure below:

```
ROOT
    proj_1
        src  # is a python root and depends on proj_2
            foo
    proj_2
        src  # is a python root
            bar  ← The visibility here will be incorrect; proj_1 cannot see this.
```

As it stands, we write a narrowly scoped `visibility` attribute for every generated rule; the visibility is `//{python-root-package}:__subpackages__`. Bazel will __fail__ to build targets in `proj_1` because of this.

## 🔥 Exception or Error

<pre><code>

</code></pre>

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
macOS 14.2
  </code>
</pre>

**Output of `bazel version`:**

<pre>
  <code>
Build label: 7.0.0
Build target: @@//src/main/java/com/google/devtools/build/lib/bazel:BazelServer
Build time: Mon Dec 11 16:52:42 2023 (1702313562)
Build timestamp: 1702313562
Build timestamp as int: 1702313562
  </code>
</pre>

**Rules_python version:**

<pre>
  <code>
0.27.1
  </code>
</pre>

**Anything else relevant?**

The desired implementation is not to set everything to `//visibility:public`, rather, it's to resolve cross-project dependencies in a monorepo the same way as they are within one directory.

Consider a working example here: https://github.com/bazelbuild/rules_python/compare/main...aryamccarthy:rules_python:main

aignas:
I think this is something that you may want only if the gazelle roots are using the same requirements files, so this ask is not necessarily something that is useful for all scenarios and I am not sure it should be the default.

eaplatanios:
@aignas I posted a reply here: https://github.com/bazelbuild/rules_python/pull/1683#discussion_r1449064691. Regarding the requirements files, that's true regarding external dependencies. Our main issue has to do with internal dependencies between the different Python projects in our Bazel workspace themselves.

aignas:
Continuing the discussion her.

[...]

PR #1683, opened 2024-01-11
fix(gazelle): _actually_ don't write narrow dependencies.

aryamccarthy:
Sorry about #1681 - I had uploaded an older version. This one contains the correct behavior, in line with what I had written.

The goal is to inherit visibility from the BUILD file's default_visibility (or whatever defaults exist higher in the tree).

I'm happy to discuss the issue further at #1682 if you have apprehensions about this change.


aryamccarthy:
Note that every failing test right now is for the same reason. They look like this: 

              	py_binary(
              	    name = "main",
              	    srcs = ["main.py"],
            - 	    visibility = ["//:__subpackages__"],
              	    deps = ["@pip//:pandas"],
              	)

This is because the auto-generated visibility lines aren't produced by this PR. That's the intended behavior, but `grep`-ing for the test names didn't turn anything up. Otherwise, I'd update the tests myself.

gazelle/python/generate.go
```diff
@@ -212,7 +212,6 @@ func (py *Python) GenerateRules(args language.GenerateArgs) language.GenerateRes
 	}
 
 	parser := newPython3Parser(args.Config.RepoRoot, args.Rel, cfg.IgnoresDependency)
-	visibility := fmt.Sprintf("//%s:__subpackages__", pythonProjectRoot)
```
aignas, line 215: Actually, this is desirable in usecases where you have a monorepo, where you have many requirements.txt files that you don't want to mix. This change would be a regression in that front.

The `go` plugin for `gazelle` has directives to control the visibility as is documented in https://github.com/bazelbuild/bazel-gazelle, maybe introducing a directive would be useful here as well?

aignas (changes requested):
The tests that are failing are in `gazelle/python/testdata` if I remember correctly.

gazelle/python/generate.go
eaplatanios: With this PR you can still get the old behavior by adding this to your Python root BUILD file:
```
package(default_visibility = ["//" + package_name() + ":__subpackages__"])
```
There's some more context in [this previous PR](https://github.com/bazelbuild/rules_python/issues/1682) that accidentally included the wrong change but the right PR description.

aignas:
Let me know if we can close this since #1784 got merged. i think that should address your use-case.

Issue #1643, opened 2023-12-21
Make MODULE.bazel.lock file have the same contents no matter the platform bazel is executed on


aignas:
This is a ticket to tidy up and do feature work to ensure that the `pip` bzlmod extension can generate a lock file that is `os,arch` independent. The rough list of steps that we need to take are:
* Create a `host` toolchain that can be passed to the `python_interpreter_target`. We need to ensure that the label that is passed to the `whl_library` repo rule is the same no matter the platform.
* Introduce a way to handle multiple requirements files and generate the wheel repositories similar to how it is done in #1625.
* Remove the requirement file selection based on the host platform and instead generate the repos for all available requirement files. Because the platform specific requirements files may have different versions of packages, we have to go all the way through and go with the approach outlined in #1625 in the general case to be fully correct.

To make the `pip` extension more maintainable we might also need to have:
* config_setting values for (`pyXY`, `os`, `cpu`) triples so that we can more easily create the required aliases in the hub repository.
* Making the dependency resolution aware of the python abi.

The above would make the number of layers of indirection fewer and hopefully make it easier to extend the code in general.

~This general work may be related fix some aspects of #310 for `bzlmod`.~

Original issue that spurred a feature on `bazel` itself: https://github.com/bazelbuild/bazel/issues/19154. The motivation for restarting the effort is the fact that maintaining a lock file that has `os,arch` specific sections is hard and requires updating the lock file by running `bazel` on a number of platforms and this just does not scale in the long term.

aignas:
Original issue that spurred a feature on `bazel` itself: https://github.com/bazelbuild/bazel/issues/19154

PR #1692, opened 2024-01-14
wip: feat: support resolving dependencies for a different python version

aignas:
With this PR we prepare the dependency resolution machinery to support other
than host python versions. This may make it possible to generate select
statements based on the version but we target this usecase as it requires
little code to be changed and once wired to the starlark code could be of use.
Creating this PR for discussion so that we can decide on a suitable interface.

If this change was wired to `whl_library`, we could generate correct
`BUILD.bazel` files for `whl` distributions when passing `/usr/bin/python3` as
the interpreter to the `pip_parse` repository rule or its `bzlmod` equivalent.
This means that users would only need to fetch a single `_host` toolchain repo
or none at all if they are using the system interpreter for the `whl_library`
generation.

I also see this useful for eventual `select` statements involving the Python
toolchain ABI and the platform and selecting the right dependencies based on
that, which would be useful in downloading the `tomli` library only for Python
versions that don't have the native `toml` support. This could allow us to
download a pure python package once for all python interpreter versions making
the repository fetching faster.

Work towards #1643.

TODO:
- [ ] Discuss on how we could expose the desired version of the python interpreter.
  Should we add an optional `python_version` argument to the `whl_library` rules? Should
  it be a `string_list`? Or should we have the `experimental_target_platforms` extended
  to accept strings like `cp_3_9_0_linux_aarch64`?

### DIFF
--- a/python/pip_install/tools/wheel_installer/wheel.py
+++ b/python/pip_install/tools/wheel_installer/wheel.py
@@ -84,10 +84,20 @@ def _as_int(value: Optional[Union[OS, Arch]]) -> int:
     return int(value.value)
 
 
+class MinorVersion(int):
+    @classmethod
+    def host(cls) -> "MinorVersion":
+        version = platform.python_version()
+        _, _, tail = version.partition(".")
+        minor, _, _ = tail.partition(".")
+        return cls(minor)
+
+
 @dataclass(frozen=True)
 class Platform:
     os: OS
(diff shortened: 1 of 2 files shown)


aignas:
Tried doing a stacked PR and this got auto-closed when the upstream PR's branch got merged in #1691. I'll reopen this as a separate PR, because I don't seem to be able to reopen this one.

PR #1693, opened 2024-01-14
feat: support parsing whl METADATA on any python version

aignas:
With this PR we can deterministically parse the METADATA and generate a
`BUILD.bazel` file using the config settings introduced in #1555. Let's
imagine we had a `requirements.txt` file that used only wheels, we could
use the host interpreter to parse the wheel metadata for all the target
platforms and use the version aware toolchain at runtime. This potentially
unlocks more clever layouts of the `bzlmod` hub repos explored in #1625
where we could have a single `whl_library` instance for all versions within
a single hub repo.

Work towards #1643.


python/pip_install/private/generate_whl_library_build_bazel.bzl
```diff
@@ -232,19 +232,28 @@ def generate_whl_library_build_bazel(
...
+            # The first item is not an abi
+            abi, os, cpu = "", abi, os
+
+        constraint_values = [
+            "@platforms//cpu:{}".format(cpu),
+            "@platforms//os:{}".format(os),
+        ]
+        if abi:
+            minor_version = int(abi[len("cp3"):])
+            constraint_values.append("@rules_python//python/config_settings:is_python_3.{}".format(minor_version))
```
rickeylev, line 247: I think the hard-coded `@rules_python` part of this string is incorrect if repo remapping is in use?

python/pip_install/private/generate_whl_library_build_bazel.bzl
```diff
@@ -232,19 +232,28 @@ def generate_whl_library_build_bazel(
         if p.startswith("@"):
             continue
 
-        os, _, cpu = p.partition("_")
+        abi, _, tail = p.partition("_")
```
rickeylev, line 235: Sorry, I forget, please remind me: what's the format of these strings being parsed? Above there is a guard against "@" -- where is that coming from? The string format looks like the wheel packaging spec, but not quite?

python/pip_install/tools/wheel_installer/wheel.py
```diff
@@ -84,16 +84,30 @@ def _as_int(value: Optional[Union[OS, Arch]]) -> int:
     return int(value.value)
 
 
+class MinorVersion(int):
+    @classmethod
+    def host(cls) -> "MinorVersion":
+        version = platform.python_version()
```
rickeylev, line 90: This can be simplified: `sys.version_info.minor`

python/pip_install/tools/wheel_installer/wheel.py
```diff
@@ -84,16 +84,30 @@ def _as_int(value: Optional[Union[OS, Arch]]) -> int:
     return int(value.value)
 
 
+class MinorVersion(int):
```
rickeylev, line 87: I'm guessing you're subclassing because you want a type-safe int? This will be confusing because it's easy to lose the type when operations are performed. If you want a type safe int, use NewType instead or manually wrap it in a distinct class (that doesn't inherit from int).

python/pip_install/tools/wheel_installer/wheel.py
```diff
@@ -235,11 +285,14 @@ def env_markers(self, extra: str) -> Dict[str, str]:
             "platform_system": self.platform_system,
             "platform_release": "",  # unset
             "platform_version": "",  # unset
+            "python_version": f"3.{minor_version}",
+            # FIXME @aignas 2024-01-14: is putting zero last a good idea? Maybe we should
+            # use `20` or something else to avoid having wheird issues where the full version is used for
```
rickeylev, line 290: typo: weird

rickeylev (approved):
Mostly just small things. Otherwise LGTM.

python/pip_install/private/generate_whl_library_build_bazel.bzl
aignas: Added a comment, thanks for pointing this out.

python/pip_install/tools/wheel_installer/wheel.py
aignas: I was subclassing because I was not sure if this should be an ABI `enum` or just an `int` and you are right, `int` here works better.

examples/bzlmod/MODULE.bazel
```diff
@@ -137,8 +139,13 @@ pip.parse(
     # You can use one of the values below to specify the target platform
     # to generate the dependency graph for.
     experimental_target_platforms = [
-        "all",
+        # Using host python version
```
aignas, line 142:
```suggestion
        # Using implicit python version
```

examples/bzlmod/.bazelrc
```diff
@@ -8,3 +8,6 @@ test --test_output=errors --enable_runfiles
 build --enable_runfiles
 
 startup --windows_enable_symlinks
+
+# Set the default version of python_version config
+common --@rules_python//python/config_settings:python_version=3.9.18
```
aignas, line 13: I got this this example to work by setting the default toolchain/python version using this and then version specific selects start working in practice with regular `py_binary` and `py_test` rules.

@rickeylev , what do you think about how we should set the default `python_version` value? I think there are several improvements (outside this PR) that we could make:
1. Allow users to set `3.9` as the value to `python_version`. This would requires us to change the `toolchain` definition so that it uses `select.match_any(3.9, 3.9.18)` in the `target_settings` and update the `is_python` generation code.
2. Move the definition of the `python_version` to the `pythons_hub` under bzlmod with setting the default value the same as the default toolchain. We can make an alias if we are using `bzlmod` to the `pythons_hub` string value. That way I can keep the `default` settings in the `MODULE.bazel` and would not need this line here.

python/pip_install/private/generate_whl_library_build_bazel.bzl
```diff
@@ -126,6 +145,87 @@ def _render_list_and_select(deps, deps_by_platform, tmpl):
...
 
+def _render_config_settings(dependencies_by_platform):
+    py_version_by_os_arch = {}
+    for p in dependencies_by_platform:
+        # p can be one of the following formats:
+        # * @platforms//os:{value}
+        # * @platforms//cpu:{value}
+        # * @//python/config_settings:is_python_3.{minor_version}
+        # * {os}_{cpu}
+        # * cp3{minor_version}_{os}_{cpu}
```
aignas, line 156: FYI, I had to move the rendering of the config settings outside the main loop because it was getting complex. However, this PR can generate the right select for `pylint`:
```starlark
    deps = [
        "@pip_39_astroid//:pkg",
        "@pip_39_dill//:pkg",
        "@pip_39_isort//:pkg",
        "@pip_39_mccabe//:pkg",
        "@pip_39_platformdirs//:pkg",
        "@pip_39_tomlkit//:pkg",
    ] + select(
        {
            "@@rules_python~override//python/config_settings:is_python_3.10": ["@pip_39_tomli//:pkg"],
            "@@rules_python~override//python/config_settings:is_python_3.9": [
                "@pip_39_tomli//:pkg",
                "@pip_39_typing_extensions//:pkg",
            ],
            ":is_cp310_windows_anyarch": [
                "@pip_39_colorama//:pkg",
                "@pip_39_dill//:pkg",
                "@pip_39_tomli//:pkg",
            ],
            ":is_cp311_windows_anyarch": [
                "@pip_39_colorama//:pkg",
                "@pip_39_dill//:pkg",
            ],
            ":is_cp39_windows_anyarch": [
                "@pip_39_colorama//:pkg",
                "@pip_39_dill//:pkg",
                "@pip_39_tomli//:pkg",
                "@pip_39_typing_extensions//:pkg",
            ],
            "//conditions:default": [],
        },
    ),
```

Disregard the fact that the conditions and the target repos should match - this can be done as a followup item. I am not sure yet what design would be best. This versioned select is only enabled if the user specifies multiple target python versions.

python/pip_install/tools/wheel_installer/wheel.py
```diff
@@ -301,18 +383,39 @@ def _add(self, dep: str, platform: Optional[Platform]):
...
+            # existing dependencies from less specialized platforms are propagated
+            # to the newly added dependency set.
+            for p, deps in self._select.items():
+                # Check if the existing platform overlaps with the given platform
+                if p == platform or platform not in p.all_specializations():
+                    continue
+
+                self._select[platform].update(self._select[p])
+
+    def _merge_to_common(self, dep):
```
aignas, line 397: I can't think of a better name right now, any suggestions? It merges deps into the `self._deps` if it is present for all python versions that we target.

examples/bzlmod/.bazelrc
rickeylev: > Allow users to set 3.9 as the value to python_version. This would requires us to change the toolchain definition so that it uses select.match_any(3.9, 3.9.18) in the target_settings and update the is_python generation code.

Yeah, I think that's fine. I was going to suggest a slightly different structure: two separate toolchain registrations, with 3.9 after any 3.9.X toolchains. This would allow multiple micro-versions. But, the rest of the everything doesn't really support a micro-version, so perhaps that is premature.

I'm mixed on allowing "3.9" as a flag value. It does seem very natural, so I do like that part. But with how flags work, "3.9" has no relation to "3.9.18", so I worry that's going to rear its head somehow. I guess our public interface here are the "is_XX" settings, though, so we have some freedom about the flag input format (e.g. "latest" as a possible value sounds convenient).

Anyways, yeah, i guess allowing 3.9 as valid value is fine.

I've heard rumor that config_common.FeatureFlagInfo can be used to allow a flag be implemented as a rule, and have the resulting value seen by config_setting. I haven't tried it myself, though, and was told this was "deprecated", so I'm not sure about that. In any case, my point in mentioning this is that would allow taking 3.9 as the input value and then transform it to 3.8.18 (by looking at MINOR_MAPPING), and everything should just work as before.

> Move the definition of the python_version to the pythons_hub under bzlmod with setting the default value the same as the default toolchain.

What about just loading the default and setting it instead? e.g.

```
# python/config_settings/BUILD
load("@pythons_hub//:interpreters.bzl", "DEFAULT_PYTHON_VERSION")

define_python_version_flag(default=DEFAULT_PYTHON_VERSION)
```

Regardless, something that isn't clear to me is how to do the same for workspace builds. Is that even possible? I think `python_register_multi_toolchains` would need to generate something with the default value in it. Except that function isn't always called, so we can't guarantee it is called. Yeah, I dunno.

python/pip_install/tools/wheel_installer/wheel.py
rickeylev: `_maybe_add_common_dep()` ?

I like to use the "maybe" term when the condition isn't brief to describe. The term encourages readers pause and go look at the doc to understand the possible post-conditions.

examples/bzlmod/.bazelrc
jvolkman: FWIW I'm currently using `python_version` directly since I can mix it with constraints in a single `config_setting`, like
```
config_setting(
  name = "python_3.10.12_aarch64-apple-darwin_config",
  flag_values = {"@rules_python//python/config_settings:python_version": "3.10.12"},
  constraint_values = ["@platforms//os:macos", "@platforms//cpu:aarch64"],
)
```

Otherwise, to use the `is_XX` settings, I guess I'd have to use  skylib's `selects.config_setting_group`?

aignas: Yeah, that is what I noticed and using `config_setting_group` is a little unwieldy. I have created #1720 as a POC to potentially allow having `"@rules_python//python/config_settings:python_version": "3.10"` instead, but I am missing a few things there to get it fully working.

rickeylev:
Sorry, removed from queue -- just want to double check something about the unit tests.

tests/pip_install/whl_library/generate_build_bazel_tests.bzl
```diff
@@ -44,7 +120,13 @@ filegroup(
...
     ] + select(
         {
-            "@platforms//os:windows": ["@pypi_colorama//:whl"],
+            "@//python/config_settings:is_python_3.9": ["@pypi_py39_dep//:whl"],
+            "@platforms//cpu:aarch64": ["@pypi_arm_dep//:whl"],
+            "@platforms//os:windows": ["@pypi_win_dep//:whl"],
+            ":is_cp310_linux_ppc": ["@pypi_py310_linux_ppc_dep//:whl"],
+            ":is_cp39_anyos_aarch64": ["@pypi_py39_arm_dep//:whl"],
+            ":is_cp39_linux_anyarch": ["@pypi_py39_linux_dep//:whl"],
+            ":is_linux_x86_64": ["@pypi_linux_intel_dep//:whl"],
```
rickeylev, line 129: This doesn't look like a correct set of conditions?

The first 3 could all be true simultaneously and aren't specializations of each other, which would be an error.

rickeylev: resolved on chat. This is just an artifact of the unittest manually constructing values to verify the select generation. the exlusivity logic is part of wheel.py

PR #1696, opened 2024-01-17
fix(bzlmod): use X.Y select for the hub repo

aignas:
This allows better interoperability between toolchains and the wheels
built by the `whl_library`, as wheels are usually compatible across all
of the python patch versions.

### DIFF
--- a/python/config_settings/config_settings.bzl
+++ b/python/config_settings/config_settings.bzl
@@ -39,10 +39,12 @@ def construct_config_settings(name, python_versions):
 
     string_flag(
         name = "python_version",
-        # Default to the legacy value that non-version aware toolchains use
-        # when setting the `python_version` attribute on `py_*` rules.
-        build_setting_default = allowed_flag_values[0],
-        values = sorted(allowed_flag_values),
+        # TODO: The default here should somehow match the MODULE config. Until
+        # then, use the empty string to indicate an unknown version. This
+        # also prevents version-unaware targets from inadvertently matching
+        # a select condition when they shouldn't.
+        build_setting_default = "",
+        values = [""] + sorted(allowed_flag_values),
         visibility = ["//visibility:public"],
     )
 
--- a/python/config_settings/config_settings.bzl
+++ b/python/config_settings/config_settings.bzl
@@ -38,8 +38,10 @@ def construct_config_settings(name, python_versions):
     string_flag(
         name = "python_version",
         # TODO: The default here should somehow match the MODULE config
-        build_setting_default = python_versions[0],
-        values = sorted(allowed_flag_values),
+        # NOTE @aignas 2024-01-17: python_versions[0] as default makes the
+        # default wheel selection on bzlmod not work.
+        build_setting_default = "",
+        values = [""] + sorted(allowed_flag_values),
         visibility = ["//visibility:public"],
     )
 
@@ -53,7 +55,6 @@ def construct_config_settings(name, python_versions):
             name = equals_minor_version_name,
             flag_values = {":python_version": minor_version},
         )
-
         matches_minor_version_names = [equals_minor_version_name]
 
         for micro_version in micro_versions:
--- a/python/config_settings/config_settings.bzl
+++ b/python/config_settings/config_settings.bzl
@@ -29,7 +29,9 @@ def construct_config_settings(name, python_versions):
 
     # Maps e.g. "3.8" -> ["3.8.1", "3.8.2", etc]
     minor_to_micro_versions = {}
-    allowed_flag_values = []
+
+    # Add the legacy value that is used in non-version aware toolchains
+    allowed_flag_values = ["PY3"]
     for micro_version in python_versions:
         minor, _, _ = micro_version.rpartition(".")
         minor_to_micro_versions.setdefault(minor, []).append(micro_version)
@@ -37,11 +39,10 @@ def construct_config_settings(name, python_versions):
 
     string_flag(
         name = "python_version",
-        # TODO: The default here should somehow match the MODULE config
-        # NOTE @aignas 2024-01-17: python_versions[0] as default makes the
-        # default wheel selection on bzlmod not work.
-        build_setting_default = "",
-        values = [""] + sorted(allowed_flag_values),
+        # Default to the legacy value that non-version aware toolchains use
+        # when setting the `python_version` attribute on `py_*` rules.
+        build_setting_default = allowed_flag_values[0],
+        values = sorted(allowed_flag_values),
         visibility = ["//visibility:public"],
     )
 
--- a/python/config_settings/config_settings.bzl
+++ b/python/config_settings/config_settings.bzl
@@ -30,8 +30,7 @@ def construct_config_settings(name, python_versions):
     # Maps e.g. "3.8" -> ["3.8.1", "3.8.2", etc]
     minor_to_micro_versions = {}
 
-    # Add the legacy value that is used in non-version aware toolchains
-    allowed_flag_values = ["PY3"]
+    allowed_flag_values = []
     for micro_version in python_versions:
         minor, _, _ = micro_version.rpartition(".")
         minor_to_micro_versions.setdefault(minor, []).append(micro_version)
--- a/python/private/bzlmod/pip.bzl
+++ b/python/private/bzlmod/pip.bzl
@@ -31,6 +31,11 @@ load("//python/private:parse_whl_name.bzl", "parse_whl_name")
 load("//python/private:version_label.bzl", "version_label")
 load(":pip_repository.bzl", "pip_repository")
 
+def _major_minor_version(version):
+    version = full_version(version)
+    major_minor, _, _ = version.rpartition(".")
+    return major_minor
+
 def _whl_mods_impl(mctx):
     """Implementation of the pip.whl_mods tag class.
 
@@ -190,7 +195,7 @@ def _create_whl_repos(module_ctx, pip_attr, whl_map, whl_overrides):
         if whl_name not in whl_map[hub_name]:
             whl_map[hub_name][whl_name] = {}
 
-        whl_map[hub_name][whl_name][full_version(pip_attr.python_version)] = pip_name + "_"
+        whl_map[hub_name][whl_name][_major_minor_version(pip_attr.python_version)] = pip_name + "_"
 
 def _pip_impl(module_ctx):
     """Implementation of a class tag that creates the pip hub and corresponding pip spoke whl repositories.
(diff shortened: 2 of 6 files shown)


python/private/bzlmod/pip.bzl
```diff
@@ -31,6 +31,11 @@ load("//python/private:parse_whl_name.bzl", "parse_whl_name")
 load("//python/private:version_label.bzl", "version_label")
 load(":pip_repository.bzl", "pip_repository")
 
+def _major_minor_version(version):
+    version = full_version(version)
```
rickeylev, line 35: Is the full_version() call just kept for the error checking? The logic it incurs is basically:
```
version = "3.10"
full_version = mapping[version] # "3.10.1" or error if unrecognized version
major_minor, _, _ = fulL_version.rpartion(full_version) # back to 3.10
```

If so, I'm fine with that for now. Lets just drop a comment about that.

A version-parsing function would be better TBH. I had to do this same "parse the dots" game in the PR to generate the config settings. But that can go in a separate PR.

rickeylev (approved):

rickeylev:
CI failure might be legit? it's a bit weird because it says there's a failure, but I don't see any test failure logs. Seems related to...the coverage command...?

It looks like env markers _could_ have a micro-version level of specificity, but I don't think I've seen that in practice. Does a wheel's metadata have that detail?

I'm just musing out loud if this will ever, practically speaking, need to regain the ability to have a more specific version (e.g. `select({":3.10": ..., ":3.10.1": ...})`

python/config_settings/config_settings.bzl
```diff
@@ -38,8 +38,10 @@ def construct_config_settings(name, python_versions):
     string_flag(
         name = "python_version",
         # TODO: The default here should somehow match the MODULE config
-        build_setting_default = python_versions[0],
-        values = sorted(allowed_flag_values),
+        # NOTE @aignas 2024-01-17: python_versions[0] as default makes the
+        # default wheel selection on bzlmod not work.
+        build_setting_default = "",
```
aignas, line 43: @rickeylev, I have found that without this line the `multi_python_version` test for default PyPI binary selection would fail unless I have this. Not sure why this is the case. I think it is because `python_versions[0]` is `3.8` so I guess we should have a special value here, which an empty string is.

rickeylev: That's weird.

python_versions is a list of full version strings, so it would be 3.8.X, not 3.8.

Is it because...do the multi-version code paths special case a falsey default value maybe? Maybe they're actually generating a no_match_error instead of //conditions:defualt entry, or something?

python/private/bzlmod/pip.bzl
aignas: Yeah, just wanted to get something functional before I spend cleaning up the API.

aignas:
> CI failure might be legit? it's a bit weird because it says there's a failure, but I don't see any test failure logs. Seems related to...the coverage command...?
> 
> It looks like env markers _could_ have a micro-version level of specificity, but I don't think I've seen that in practice. Does a wheel's metadata have that detail?
> 
> I'm just musing out loud if this will ever, practically speaking, need to regain the ability to have a more specific version (e.g. `select({":3.10": ..., ":3.10.1": ...})`

The wheels are built for py3 (pure python), abi3 (subset of cpython abi, which is binary compatible with all cpython 3 versions), and cpxy, where there is no micro specificity. I have not seen any micro level specificity out there. That said, I think I once saw a python metadata file that excluded a particular micro version as a supported version, but I cannot remember where it was and if my brain is not making stuff up.

python/config_settings/config_settings.bzl
```diff
@@ -29,16 +29,19 @@ def construct_config_settings(name, python_versions):
 
     # Maps e.g. "3.8" -> ["3.8.1", "3.8.2", etc]
     minor_to_micro_versions = {}
-    allowed_flag_values = []
+
+    # Add the legacy value that is used in non-version aware toolchains
+    allowed_flag_values = ["PY3"]
```
aignas, line 34: @rickeylev can you think of any unintended consequences here? If it looks good to you feel free to press merge.

python/private/bzlmod/pip.bzl
aignas: Changed the function to do actual version parsing and if this is something that looks OK, we can move it to `python/private/semver.bzl#parse_semver` or something similar.

python/config_settings/config_settings.bzl
aignas: I've no idea about this to be honest, if I set this to `python_versions[0]`, which is `3.8.x`, then the `websockets` version that we get in the test is for Python 3.8 and not 3.9 as it should be. Is there a way to somehow print the trace of how the select happens?

python/config_settings/config_settings.bzl
rickeylev: Is the idea to support a more drop-in replacement of `py_binary(python_version="PY3")` with a version-aware `py_binary` symbol?

I think that's a good idea, but it makes more sense to drop the value in the transition rather than accept the value as flag state. i.e. in the transition:

```
def transition_ipml(_, attr):
  pv = attr.python_version
  if pv == in ("PY3", "PY3ONLY", ...):
    pv = ""  # default, or some such
  return {"...:python_version": pv}
```

Then nothing ever sees "PY3" as value and we don't have to worry about people trying to match PY3 with config_setting.flag_values

That's the main unintended consequence I see here -- people will try to match on that flag value as some indicator of "wasn't set" or "unspecified version" or something

python/private/bzlmod/pip.bzl
```diff
@@ -25,12 +25,29 @@ load(
     "whl_library",
 )
 load("//python/pip_install:requirements_parser.bzl", parse_requirements = "parse")
-load("//python/private:full_version.bzl", "full_version")
 load("//python/private:normalize_name.bzl", "normalize_name")
 load("//python/private:parse_whl_name.bzl", "parse_whl_name")
 load("//python/private:version_label.bzl", "version_label")
 load(":pip_repository.bzl", "pip_repository")
 
+def _parse_version(version):
```
rickeylev, line 33: Oh nice! This is clever and clean.

python/private/bzlmod/pip.bzl
```diff
@@ -25,12 +25,29 @@ load(
...
+    major, _, version = version.partition(".")
+    minor, _, version = version.partition(".")
+    patch, _, version = version.partition(".")
+    build, _, version = version.partition(".")
```
rickeylev, line 37: (no action required)

Is the 4-dot component notation pip or semver?

IIRC, the python sys.version format is "3.10.5a4" (3 dots, letter, serial).

I don't think it matters much for now as we only ever have "x.y.z" for the python versions

python/config_settings/config_settings.bzl
```diff
@@ -29,16 +29,19 @@ def construct_config_settings(name, python_versions):
...
     for micro_version in python_versions:
         minor, _, _ = micro_version.rpartition(".")
         minor_to_micro_versions.setdefault(minor, []).append(micro_version)
         allowed_flag_values.append(micro_version)
 
     string_flag(
         name = "python_version",
-        # TODO: The default here should somehow match the MODULE config
-        build_setting_default = python_versions[0],
+        # Default to the legacy value that non-version aware toolchains use
```
rickeylev, line 42: I don't think this comment is accurate? I thought the value here was arbitrary; it's just picking the first value from TOOLS.keys(), which I don't think corresponds to any thing in particular?

Ideally this value would be more intentional. Either the value to in rules_python (3.11 right now think) or whatever was configured as the default python version

python/config_settings/config_settings.bzl
rickeylev: I don't think so...

To make sure I understand...
* Setting it so 3.8.x failed
* Setting it to empty string passed
* Setting it to "PY3" passed

That seems even weirder. It doesn't make much sense, since this is the default for the flag, and only handful of things directly read that flag...

The only idea i have is something in the multi-pip-parse example has something like `select({":3.8.x": ..., "//conditions:default": ...})`. So an empty string and PY3 go down the default case.

What if there was a bug here all along? Like "3.8.1" was being set, but the select was looking for "3.8.2", and then somehow //conditions:default just ended up somewhere happy enough for tests to pass? And by changing the select to effectively match _both_ 3.8.1 and 3.8.2, it's now doing the _intended_ thing and showing some misconfiguration?

rickeylev (commented):
I'll try to find some time tonight to poke that weird failure. If I can't find that time, then we'll merge anyways; we can fix the "they might match PY3 in a config setting" separately if necessary.

rickeylev:
Ok, I think I know what's going on. This exposed an interesting misconfiguration -> surprising behavior -> bug. The basic issue is the `:is_3.8` condition is matching the default flag value (3.8.10), which causes the 3.8 select path to be followed, which picks websockets 3.8.

Why was this passing before but failing now? Because all of these were true:
1. The flag's default was 3.8.10
2. The version-unaware rules will inherit the flag value (the default in this case)
3. The WORKSPACE/MODULE config set the default python version as 3.8.18
4. the select condition was looking for _exactly_ 3.8.18.

The 3.8.10 value comes from `TOOL_VERSIONS.keys()[0]`. The 3.8.18 value comes from `MINOR_MAPPING["3.8"]`.

Unrolling things, the code looked like this:
```
# config.bzl
string_flag(default="3.8.10", ...) # from TOOL_VERSIONS.keys()[0]
# MODULE
python.toolchain(python_version="3.8") # turns into 3.8.18 from MINOR_MAPPING
python.toolchain(python_version="3.9", is_default=True)

# bazel query --output=build @pypi//websockets:pkg
alias(..., select({
  "@rules_python//python/config_settings:is_python_3.8.18": "...websockets 38..."
  "//conditions:default": "..websockets 39..."
})
```

When you build a non-version aware target, it'll use `--python_version=3.8.10` (the default). When you build version aware target, it'll turn into 3.8.18.

In the new code, those select conditions match all 3.8.x.

Each piece here is working as intended. The bug is really that the default flag value is set to a value that isn't the _actual_ default python. Fixing that is out of scope for this PR (I'm also not sure we can fix this without bzlmod). I think the best we can do is just set the default to some value that won't match. Empty string seems fine? It's that or a special value like "default"; not sure which is better. Mild preference for empty string (seems closer to "undefined")

rickeylev:
Thinking about it more, I think that bug is just in this PR. In order to trigger it prior to this PR, you would need to have somehow have a version-unaware build generate the 3.8.10 select condition -- but the MINOR_MAPPING would prevent that. Another way would be to have a 3.8.10 config condition, but being as how activating that version is rather hard (you'd have to patch I think), that seems unlikely. Plus its an older version of Python that is no longer supported.

rickeylev (approved):

PR #1699, opened 2024-01-17
feat: enable rules_python Starlark implemention by default for Bazel 7+

rickeylev:
This makes the base rules use the Starlark code in rules_python
("pystar") instead of what is built into Bazel 7.

The pystar implementation can be disabled by setting
`RULES_PYTHON_ENABLE_PYSTAR=0` in your environment or using
`--action_env`.

Earlier versions of Bazel continue to use the rules built into Bazel
regardless (the new implementation requires special APIs that only Bazel
7 has).

Work towards https://github.com/bazelbuild/rules_python/issues/1069


rickeylev:
Doh, forgot the changelog entry

aignas (approved):

rickeylev:
posterity, from maintainer's meeting:

The plan for this PR is to make it its own release. So immediately after the next release, I'll merge this and then do a subsequent release. This is to help decouple it from the other fixes/features/breaking-changes that have been going in. I'm not _too_ worried about regressions (Bazel 7 uses the new starlark impl this is based upon; Google has been using the starlark impl for many months and the rules_python variant for several weeks), but I'd rather be more cautious in this particular case.

PR #1700, opened 2024-01-17
fix: use shutil.copyfile instead of shutil.copy

SpencerC:
Encountering this error when running on a CI VM: https://stackoverflow.com/questions/11835833/why-would-shutil-copy-raise-a-permission-exception-when-cp-doesnt

Per the accepted solution, this solves the problem and doesn't seem to impact functionality. Is this consistent with the intent here?

### DIFF
--- a/python/pip_install/tools/dependency_resolver/dependency_resolver.py
+++ b/python/pip_install/tools/dependency_resolver/dependency_resolver.py
@@ -29,7 +29,7 @@ from python.runfiles import runfiles
 
 # Replace the os.replace function with shutil.copy to work around os.replace not being able to
 # replace or move files across filesystems.
-os.replace = shutil.copy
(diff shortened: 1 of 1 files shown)


aignas:
See https://github.com/bazelbuild/rules_python/pull/1053/commits/c6eb1af9ab8a0b2aca5c99e5d22a39ed8cf55451 from #1053, which is using `shutil.copyfile` instead of `shutil.copy`.

If you would still like to contribute in this area, changing the PR to add a comment why it is `shutil.copy` instead of `shutil.copyfile` would be really appreciated!

aignas:
On the other hand, let's close this PR, but please feel free to create a separate PR to improve docs or provide a better solution than the existing shutil.copy given this context.

PR #1704, opened 2024-01-18
refactor: broaden visibility and use list() instead of keys()

rickeylev:
This upstreams a couple trivial Google patches to make rules_python more compatible with some Google-internal changes.

* Use list(dict) instead of dict.keys(). This is because some bzl files are run through a Python-based testing framework, and dict.keys() can't be concatenated to a list in Python 3.
* Expand visibility to all of rules_python instead of just the python subdirectory. This makes some patches that use internals easier to maintain, but also makes the visiblity more in line with the rest of the project (where `//:__subpackages__` is used for convenience, as it avoids having to change visibility frequently, but still prevents public dependencies).

### DIFF
--- a/python/private/common/BUILD.bazel
+++ b/python/private/common/BUILD.bazel
@@ -15,7 +15,7 @@
 load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
 
 package(
-    default_visibility = ["//python:__subpackages__"],
+    default_visibility = ["//:__subpackages__"],
 )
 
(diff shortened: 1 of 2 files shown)


Solumin (approved):

rickeylev:
It looks like CI is unhappy because of some issue with 7.0.1 and the pystar implementation. Ugh. This is gonna be annoying to fix...

I'm just going to force-merge past this. The rest of CI OK.

PR #1709, opened 2024-01-20
fix: fix launching Python 3.11 from <3.8 on macOS

armandomontanez:
On macOS, some older versions of Python leak __PYVENV_LAUNCHER__ into
the interpreter's environment variables. Starting with Python 3.11, this
causes problems with Python being able to start correctly when launched
from the Python bootstrap template. This change introduces a workaround
to prevent the problematic environment variable from leaking into
launched subprocesses.

I've verified this problem exists with the version of Python 3.7
included with Xcode, and there are reports of it also being an issue
with the Python 3.8 included with Xcode. Later versions of Python have
fixed this.

Context:
- https://issues.pigweed.dev/310293060
- https://github.com/xonsh/xonsh/issues/3576#issuecomment-629013991
- https://github.com/pypa/virtualenv/issues/1458


rickeylev (approved):
per chat, I think this just needs to be updated to head, then CI will be happy. Otherwise LGTM

PR #1712, opened 2024-01-21
docs: Update gazelle/README.md to fix `package_name` formatting

michael-christen:
Latex output was shown, should just be escaped

image


aignas:
Thanks for the contribution! Could you please sign the CLA?

michael-christen:
Of course, signed and updated!

Looks like there's an unrelated failure at "examples/multi_python_versions: Windows, workspace on :windows: (OpenJDK 11, VS2022)", anything I should do about that?

PR #1718, opened 2024-01-24
fix(bzlmod): do not fail pip.parse with empty requirements.txt

aignas:
With this change we handle correctly the case where the requirements.txt
might be empty (e.g. during onboarding to `rules_python`). I have added
an integration test because this usecase seems important enough to be covered
to ensure that we have no regressions. It also creates a place to add more
`pip.parse` and `pip_parse` tests.

Closes #1623.


tests/integration/pip_parse/.bazelrc
```diff
@@ -0,0 +1,11 @@
...
+
+build --enable_runfiles
+startup --windows_enable_symlinks
+
+# https://docs.bazel.build/versions/main/best-practices.html#using-the-bazelrc-file
+try-import %workspace%/user.bazelrc
+
+# The requirements.bzl entry_point functions aren't supported under bzlmod.
+# They are replaced by py_console_script_binary, which already has tests
+build --noexperimental_enable_bzlmod
```
rickeylev, line 11: Lets leave bzlmod enabled by default, to match the default config

tests/integration/pip_parse/.gitignore
```diff
@@ -0,0 +1,4 @@
+# git ignore patterns
```
rickeylev, line 1: This shouldn't be necessary? The top-level gitignore should also be ignoring these

rickeylev (approved):
just some small nits, otherwise lgtm

tests/integration/pip_parse/.bazelrc
aignas:
```suggestion
# The requirements.bzl entry_point functions aren't supported under bzlmod.
# They are replaced by py_console_script_binary, which already has tests
build --enable_bzlmod
```

PR #1719, opened 2024-01-24
fix(py_wheel): Fix parsing errors with `requires_file` attribute.

UebelAndre:
The `requires_file` and `extra_requires_files` attributes added by https://github.com/bazelbuild/rules_python/pull/1710 break wheels by leaving trailing `;` and currently do not support `requirements.in` files with comments following a constraint on the same line. This PR fixes these issues.

### DIFF
--- a/examples/wheel/BUILD.bazel
+++ b/examples/wheel/BUILD.bazel
@@ -278,6 +278,7 @@ write_file(
 --index-url https://pypi.com
 
 tomli>=2.0.0
+starlark  # Example comment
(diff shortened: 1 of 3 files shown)


UebelAndre:
@aignas Do you think this needs a note in the CHANGELOG since the original feature has yet to be released?

rickeylev (approved):
Nah, no changelog entry is needed for a fix to something unreleased.

aignas (commented):
LGTM

PR #1720, opened 2024-01-24
feat(toolchain): support specifying x.y versions in transitions

aignas:
This is inspired by how rules_go is registering their toolchains.

Their toolchains have multiple `target_settings` values. This
allows for a simpler passing of `X.Y` version to the `py_binary` and
`py_test` rules and does not strictly require us to provide the APIs
that pass the full python version value as the closure. This is only
possible because #1555 introduced working aliases and now we can also
have this.

Summary:
- refactor: move the toolchain_def to starlark as opposed to templating
- refactor: move the version setting as well
- feat: support matching on X.Y versions
- feat: X.Y.Z will match if X.Y is used as python_version flag and the
  MINOR_MAPPING has `"X.Y": "X.Y.Z"`.
- test: add tests checking the generated config settings.
- doc: add an example of how we could use the transition files directly

See https://github.com/bazelbuild/rules_go/blob/master/go/private/go_toolchain.bzl#L181


aignas:
Hah, the build is unhappy, but not in all situations, however, wanted to submit this as it was in my local branch and #1555 added the missing pieces that I needed to get this almost working.

examples/bzlmod/tests/BUILD.bazel
```diff
@@ -104,6 +128,19 @@ py_test_3_10(
     main = "cross_version_test.py",
 )
 
+cross_py_test(
+    name = "version_3_10_takes_3_9_subprocess_test_2",
```
rickeylev, line 132: Oh good idea testing this case :+1:

python/private/toolchain_defs.bzl
```diff
@@ -0,0 +1,81 @@
...
+# limitations under the License.
+
+"""Create the toolchain defs in a BUILD.bazel file."""
+
+load("@bazel_skylib//lib:selects.bzl", "selects")
+
+_py_toolchain_type = Label("@bazel_tools//tools/python:toolchain_type")
+_py_cc_toolchain_type = Label("//python/cc:toolchain_type")
+
+def toolchain_defs(*, prefix, user_repository_name, python_version, set_python_version_constraint, **kwargs):
```
rickeylev, line 22: Perhaps `define_toolchain_suite()` instead? That a more verby-name and "suite" refers to how there's more than just a singe toolchain_type that is defined.

python/private/toolchains_repo.bzl
```diff
@@ -51,58 +50,32 @@ def python_toolchain_build_file_content(
...
     return "".join([
         """
-toolchain(
-    name = "{prefix}{platform}_toolchain",
-    target_compatible_with = {compatible_with},
-    target_settings = {target_settings},
-    toolchain = "@{user_repository_name}_{platform}//:python_runtimes",
-    toolchain_type = "@bazel_tools//tools/python:toolchain_type",
-)
+load("@rules_python//python/private:toolchain_defs.bzl", "toolchain_defs")
```
rickeylev, line 63: Won't this cause the load to be repeated?

examples/bzlmod/tests/BUILD.bazel
```diff
@@ -27,6 +28,13 @@ py_binary_3_11(
     main = "version.py",
 )
 
+cross_py_binary(
```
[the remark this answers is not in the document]
aignas: Yeah, maybe for now we should have `py_versioned_test` at the same time I was inspired by https://github.com/bazelbuild/rules_go/blob/master/docs/go/core/rules.md#go_cross_binary for the naming. In the long run I think we may want to have pycross-like functionality, but we can cross that bridge later.

python/private/toolchain_defs.bzl
aignas: Chose `py_toolchain_suite` as there was `cc_toolchain_suite` already.

python/config_settings/config_settings.bzl
```diff
@@ -59,14 +62,39 @@ def construct_config_settings(name, python_versions):
         )
         matches_minor_version_names = [equals_minor_version_name]
 
+        default_micro_version = MINOR_MAPPING.get(minor_version)
+
         for micro_version in micro_versions:
             is_micro_version_name = "is_python_" + micro_version
+            if not default_micro_version:
```
rickeylev, line 69: What case is this handling? AFAIK, the minor version is always in the MINOR_MAPPING dict. This if statement implies that is not true?

Is this supposed to be `if default_micro_version != micro_version` ?

python/private/BUILD.bazel
```diff
@@ -206,6 +206,14 @@ bzl_library(
     srcs = ["text_util.bzl"],
 )
 
+bzl_library(
+    name = "py_toolchain_suite_bzl",
```
rickeylev, line 210: nit: move this to a more alphabetical location

CHANGELOG.md
```diff
@@ -47,6 +47,13 @@ A brief description of the categories of changes:
 * New Python versions available: `3.11.7`, `3.12.1` using
   https://github.com/indygreg/python-build-standalone/releases/tag/20240107.
 
+* (toolchain) Allow setting `x.y` as the `python_version` parameter in
+  `py_binary` and `py_test` with transitions. This allows the users to
+  use the same rule import for testing with specific Python versions and
+  rely on toolchain configuration and how the latest version takes precedence
+  if e.g. `3.8` is selected. That also simplifies `.bazelrc` for any users
+  that set the default `python_version` string flag in that way.
+
```
rickeylev, line 56:
```suggestion
* (toolchain) Allow setting `x.y` as the `python_version` parameter in
  the version-aware `py_binary` and `py_test` rules. This allows users to
  use the same rule import for testing with specific Python versions and
  rely on toolchain configuration and how the latest version takes precedence
  if e.g. `3.8` is selected. That also simplifies `.bazelrc` for any users
  that set the default `python_version` string flag in that way.
```

python/config_settings/config_settings.bzl
aignas: If the user is using this macro, which is not private, then we will still work. Do you think we don't need it?

python/config_settings/config_settings.bzl
```diff
@@ -59,14 +62,39 @@ def construct_config_settings(name, python_versions):
...
+
+            # Ensure that is_python_3.9.8 is matched if python_version is set
+            # to 3.9 if MINOR_MAPPING points to 3.9.8
+            equals_micro_name = "_python_version_flag_equals_" + micro_version
             native.config_setting(
-                name = is_micro_version_name,
+                name = equals_micro_name,
                 flag_values = {":python_version": micro_version},
+            )
+            selects.config_setting_group(
```
rickeylev, line 85:
```suggestion
            # An alias pointing to an underscore-prefixed config_setting_group
            # is used because config_setting_group creates
            # `is_{minor}_N` targets, which are easily confused with the
            # `is_{minor}.{micro}` (dot) targets.
            selects.config_setting_group(
```

rickeylev (approved):
LGTM. Just either remove the optional MINOR_MAPPING check, or put a comment about the case it is serving.

aignas:
I'm gonna merge this, but if you have a comment on how to make the test for the generated code better, let me know. It seemed like visual inspection was the easiest way for now.

rickeylev:
For testing, analysis_test can be used with selects and a small custom rule to hold values.

```

subject = rule(
  implementation = lambda ctx: None,
  attrs = {"match_a": attr.bool()}
)

def _test_config_matching(name):
  subject(
    name="subject",
    match_a = select({"is_python_3.4.5": True}, "//conditions:default": False})
  )
  analysis_test(
    ...,
    config_settings = {
      "@rules_python//python/config_settings:python_version": "3.4"
    }
  )

def _test_config_matching_impl(env, target):
  env.expect.that_target(target).attr("match_a", factory=subjects.bool).equals(True)
```

Oh, actually, it looks like there are some tests that do basically this in tests/config_settings/construct_config_settings_test.

PR #1728, opened 2024-01-29
feat: (WIP) Add pypi_install() rule

philsc:
Adds an `example/pypi_install` directory to show-case multi-arch py binaries.

Caveats:
* Doesn't use bzlmod yet.
* Doesn't support multiple Python versions.
* Need to re-lock requirements.txt files with the hermetic toolchain.
* Probably more that I can't think of.


aignas:
Closing this as we don't intend to merge it.

PR #1734, opened 2024-01-30
fix(coverage): generating lcov was causing issues

trentontrees:
* examples/bzlmod - bazel coverage //tests:version_3_10_takes_3_9_subprocess_test was always failing due to .coveragerc generated file not being unique
* generating the locv report resulted in messages going to stdout/stderr that resulted in test failures.  To fix this, we run with --quiet.  If VERBOSE_COVERAGE is defined we will output to stderr.

Reproduction steps:
```bash
git co main
cd examples/bzlmod
bazel coverage //tests/...
# failures occur
```


python/private/python_bootstrap_template.txt
```diff
@@ -389,8 +394,8 @@ def _RunForCoverage(python_program, main_filename, args, env,
         runfiles directory if set.
   """
   # We need for coveragepy to use relative paths.  This can only be configured
-  # via an rc file, so we need to make one.
-  rcfile_name = os.path.join(os.environ['COVERAGE_DIR'], '.coveragerc')
+  unique_id = uuid.uuid4()
+  rcfile_name = os.path.join(os.environ['COVERAGE_DIR'], ".coveragerc_{}".format(unique_id))
```
aignas, line 398: Is this something that is visible only if you have multiple `py_test` targets in the same directory?

aignas (commented):
If I understand correctly, this will work only with `bazel 7.0`, could you add an item in the `CHANGELOG.md` in the `fixed` section so that we communicate this to users?

aignas:
Could you also update https://github.com/bazelbuild/rules_python/blob/main/.bazelci/presubmit.yml#L62 and related so that we can reproduce the failures in the CI?

python/private/python_bootstrap_template.txt
aignas: I think that the template file that you are modifying will only be used in bazel 7.0 and above. @rickeylev, could you remind please if this template usage was backported to 6.x or not?

trentontrees:
I updated the presubmit.yml to make the failing tests run.  I'm having troubles reproducing the Ubuntu 20.04 issues locally so I can debug, iterate, and fix it.

I created a separate PR to show the issue without any changes in this PR. https://github.com/bazelbuild/rules_python/pull/1740

python/private/python_bootstrap_template.txt
trentontrees: with more testing, I think you are right. I'll make those changes tomorrow.

aignas:
Let me know when this becomes ready to review again.

trentontrees:
@aignas , this should be ready for review.

python/private/python_bootstrap_template.txt
```diff
@@ -415,18 +420,18 @@ relative_files = True
 
   PrintVerboseCoverage('Converting coveragepy database to lcov:', output_filename)
   # Run coveragepy again to convert its .coverage database file into lcov.
+  params = [python_program, coverage_entrypoint, "lcov", "--rcfile=" + rcfile_name, "-o", output_filename, "--quiet"]
+  kparams = {"env": env, "cwd": workspace, "stdout": subprocess.DEVNULL, "stderr": subprocess.DEVNULL}
+  if IsVerboseCoverage():
+    params.remove("--quiet")
```
aignas, line 426: I think adding a comment into why this behaviour is as is would be great.

trentontrees:
Added the comments, hopefully they bring value ;)

Rewrote the CHANGELOG.md blurb while I was at it.

trentontrees:
@aignas This should be ready for review again.

aignas (approved):

PR #1735, opened 2024-01-30
feat: add debug logging of repo sub processes

rickeylev:
Debugging what the repo rules are up to can be difficult because Bazel doesn't provide many facilities to inspect what they're up to. This adds an environment variable, `RULES_PYTHON_REPO_DEBUG`, that, when set, will make our repo rules print out detailed information about the subprocesses they are running.

This also makes failed commands dump much more comprehensive information.

This was driven by the recent report of failures on Windows during a repo rule.


python/pip_install/pip_repository.bzl
```diff
@@ -719,28 +726,14 @@ def _whl_library_impl(rctx):
     # Manually construct the PYTHONPATH since we cannot use the toolchain here
     environment = _create_repository_execution_environment(rctx, python_interpreter)
 
-    result = rctx.execute(
-        args,
+    repo_utils.execute_checked(
+        rctx,
+        op = "whl_library.ResolveRequirement({}, {})".format(rctx.attr.name, rctx.attr.requirement),
```
aignas, line 731: nit: we could also have this semantics for the op, where the first item is the function name and the remaining are parameters which will be formated to: `op[0] + "(" + ",".join(op[1:]) + ")"`
```suggestion
        op = ["whl_library.ResolveRequirement", rctx.attr.name, rctx.attr.requirement],
```

aignas (approved):
Thanks, this is really great!

python/pip_install/pip_repository.bzl
rickeylev: I like the idea. It also gave me another idea: automatically add the "whl_library" part (the repo rule name) by looking for a `_repo_name` attribute we set on all our repo rules.

But, I'm strapped for time today, so I'm going to merge this as-is.

rickeylev (commented):
I tried writing some tests for this but it started to spiral out of control pretty quickly (I tried mocking rctx, fail, and print 😵‍💫) and didn't test very well. I manually ran `bazel build ...` with the env set and forcing failures instead.

PR #1743, opened 2024-02-05
feat(config_settings): add extra settings for toolchain platforms

aignas:
Prior to this PR `//python/config_settings` would contain only the
`config_setting` targets for the Python versions from `//python:versions.bzl`.
We would only match the `python_version` configuration and if someone needed
matching against the `constraint_values` they would have to redefine the
`config_setting` values themselves as `match_all` helper from bazel skylib
cannot be used in this case do to how specialization priority of
`config_setting` in a `select` works - `match_all` would create `select`
statement entries that are not specialized enough, i.e. they would not have the
python version config setting and the target platform constraint values in a single
`config_setting`.

This PR creates additional `config_setting` values that are useful for rulesets
implementing handling of platform-specific wheels in a multi-python version scenario
and it lays a good foundation for supporting `download_only = True` in a better way
within the `pip.parse` extension itself.

NOTE: `config_setting` values are only created for the Python toolchain target
platforms.


python/pip_install/private/generate_whl_library_build_bazel.bzl
```diff
@@ -141,85 +141,49 @@ def _render_list_and_select(deps, deps_by_platform, tmpl):
...
-    py_version_by_os_arch = {}
+    plats = []
     for p in dependencies_by_platform:
         # p can be one of the following formats:
         # * @platforms//os:{value}
         # * @platforms//cpu:{value}
         # * @//python/config_settings:is_python_3.{minor_version}
+        # * @//python/config_settings:is_python_3.{minor_version}_{os}_{cpu}
+        # * @//python/config_settings:is_python_3.{minor_version}_{os}_any
+        # * @//python/config_settings:is_python_3.{minor_version}_any_{cpu}
```
aignas, line 152:
```suggestion
        # * @//python/config_settings:is_python_3.{minor_version}_{os}_{cpu}
        # * @//python/config_settings:is_python_3.{minor_version}_{os}
        # * @//python/config_settings:is_python_3.{minor_version}_{cpu}
```

python/pip_install/tools/wheel_installer/wheel_test.py
```diff
@@ -236,13 +236,24 @@ def test_can_get_version_select(self):
         )
         got = deps.build()
 
+        self.maxDiff = None
+
```
aignas, line 240:
```suggestion
```

python/pip_install/private/generate_whl_library_build_bazel.bzl
```diff
@@ -141,85 +141,49 @@ def _render_list_and_select(deps, deps_by_platform, tmpl):
...
-            abi = ""
-        os, _, arch = tail.partition("_")
-        os = "" if os == "anyos" else os
-        arch = "" if arch == "anyarch" else arch
+        os, _, arch = p.partition("_")
+        os = "" if os == "any" else os
+        arch = "" if arch == "any" else arch
 
-        py_version_by_os_arch.setdefault((os, arch), []).append(abi)
+        # TODO @aignas 2024-02-03: simplify
```
aignas, line 161:
```suggestion
```

aignas:
Thinking about it a little more, I am wondering if it is OK to have this code in here. It may be better to not expose extra `config` settings. If the `os/arch-specific` selects are needed, maybe they should be defined/passed in the `pip.parse` instead. This is so that the `rules_python` API surface is kept to a minimum and so that we don't make things too coupled.

That said, having the mapping between a python version string and the expected flag values that are passed to `match_any` would be useful, because that is a property of the `python_version` string flag.

Feel free to comment if you disagree with my statement above.

Issue #1732, opened 2024-01-30
bug: PyRuntimeInfo provided by py_binary doesn't match PyRuntimeInfo symbol from rules_python


axbycc-mark:
# 🐞 bug report

### Affected Rule

 The issue is caused by the rule: 
py_binary rule

### Is this a regression?

Yes. Well behaved code loading everything from rules_python (as recommended and desired) stopped working.

A workaround is possible if you control the rule that consumes the target that should provide PyRuntimeInfo: accept both the rules_python PyRuntimeInfo symbol and builtin PyRuntimeInfo provider. When accessing it in rule code, prefer the rules_python one over the bulitin one. When forwarding, forward both if you need compatibility. The gist of the code would look like this:

```
load("@rules_python//python:py_runtime_info.bzl", RulesPythonPyRuntimeInfo = "PyRuntimeInfo")

BuiltinPyRuntimeInfo = PyRuntimeInfo

foo_binary = rule(..., attrs = {
  # Note its a 2-element list of 1-element lists.
  "bin": attr.label(provides = [[RulesPythonPyRuntimeInfo], [BuiltinPyRuntimeInfo]])
})

def foo_binary_impl(ctx):
  bin = ctx.attr.bin
  if RulesPythonPyRuntimeInfo in bin:
    rp_info = bin[RulesPythonPyRuntimeInfo]
    providers.append(rp_info)
  else:
    rp_info = None
  if BuiltinPyRuntimeInfo in bin:
    builtin_info = bin[BuiltinPyRuntimeInfo]
    providers.append(builtin_info)
  else:
    builtin_info = None
  return providers
```

### Description

 
Using py_binary rule triggers fail() called with error string "target {} does not have rules_python PyRuntimeInfo or builtin PyRuntimeInfo)" on Windows 10

## 🔬 Minimal Reproduction

Load PyRuntimeInfo and py_binary from rules_python, then try to read the loaded PyRuntimeInfo from the py_binary target.

```
# BUILD
load(":defs.bzl", "foo")
load("@rules_python//python:py_binary.bzl", "py_binary")

foo(name="foo", bin=":bin")
py_binary(name="bin", srcs=["bin.py"])

# defs.bzl

load("@rules_python//python:py_runtime_info.bzl", "PyRuntimeInfo")

foo = rule(attrs = {
  "bin": attr.label(provides=[PyRuntimeInfo])
}, ...
)
```

## 🔥 Exception or Error

[...]

PR #1748, opened 2024-02-09
fix(PyRuntimeInfo): use builtin PyRuntimeInfo unless pystar is enabled.

rickeylev:
This fixes the bug where the PyRuntimeInfo symbol rules_python exported wasn't matching the provider identity that `py_binary` actually provided.

The basic problem was, when pystar is disabled:
 * PyRuntimeInfo is the rules_python defined provider
 * py_binary is `native.py_binary`, which provides only the builtin PyRuntimeInfo provider.

Thus, when users loaded the rules_python PyRuntimeInfo symbol, it was referring to a provider that the underlying py_binary didn't actually provide. Pystar is always disabled on Bazel 6.4,
and enabling it for Bazel 7 will happen eminently.

This typically showed up when users had a custom rule with an attribute definition that used the rules_python PyRuntimeInfo.

To fix, only use the rules_python define PyRuntimeInfo if pystar is enabled. This ensures the providers the underlying rules are providing match the symbols that rules_python is exported.

* Also fixes `py_binary` and `py_test` to also return the builtin PyRuntimeInfo. This
  should make the transition from the builtin symbols to the rules_python symbols a bit
  easier.

Fixes https://github.com/bazelbuild/rules_python/issues/1732

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,13 @@ A brief description of the categories of changes:
 * (toolchain) Changed the `host_toolchain` to symlink all files to support
   Windows host environments without symlink support.
 
+* (PyRuntimeInfo) Switch back to builtin PyRuntimeInfo for Bazel 6.4 and when pystar
+  is disabled. This (for now) fixes an error about `target ... does not have
+  ... PyRuntimeInfo`. This error is likely to recur in a subsequent release when
+  the rules_python implementation of the rules and providers ("pystar") is
+  enabled -- see associated issue for how to handle this case when it occurs.
+  ([#1732](https://github.com/bazelbuild/rules_python/issues/1732))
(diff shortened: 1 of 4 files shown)


rickeylev:
Ah hm, looks like the pystar config isn't happy with this. I'll put this back into draft and figure out what's going on.

rickeylev:
Tests fix. PTAL.

It uncovered another gap in compatibility: pystar py_binary wasn't returning the builtin PyRuntimeInfo. It does now, which should allow it to (1) work with a builtin py_runtime defined toolchain, and (2) allow it to work with code that is expecting the builtin PyRuntimeInfo (i.e. users that aren't loading from rules_python).

aignas (approved):

PR #1744, opened 2024-02-05
feat(bzlmod): support wheel-only additions to the hub repo

aignas:
With this PR we can reuse the same `hub-repo` and setup `download_only = True`
closures, which use only wheels but download the right platform specific wheels
independent of the host platform. That should allow users to build images with
`rules_oci` without needing to write extra code to create alias repos which
mimic the behaviour of rules_python ones.

This is stacked on #1743.

The snippet of the resultant `pylint` hub repo alias that is the main goal is:
```starlark
alias(
    name = "pkg",
    actual = select(
        {
            "//:is_python_3.9": "@pip_39_pylint//:pkg",
            "//:is_python_3.9_linux_x86_64": "@pip_39_manylinux_2_17_x86_64_39_pylint//:pkg",
            "//:is_python_3.10": "@pip_310_pylint//:pkg",
            "//:is_linux_x86_64": "@pip_39_manylinux_2_17_x86_64_39_pylint//:pkg",
            "//conditions:default": "@pip_39_pylint//:pkg",
        },
    ),
)
```

Work towards #1606

PR #1750, opened 2024-02-12
internal(config_settings): make config_setting creation reusable

aignas:
The PR #1743 explored the idea of creating extra config settings for
each target platform that our toolchain is targetting, however that has
a drawback of not being usable in `bzlmod` if someone built Python for
a platform that we don't provide a toolchain for and tried to use the
`pip.parse` machinery with that by providing the
`python_interpreter_target`. That is a niche usecase, but `rules_python`
is a core ruleset that should only provide abstractions/helpers that
work in all cases or make it possible to extend things.

This explores a way to decouple the definition of the available
`config_settings` values and how they are constructed by adding an extra
`is_python_config_setting` macro, that could be used to declare the
config settings from within the `pip.parse` hub repo. This makes the
work in #1744 to support whl-only hub repos more self-contained.

Supersedes #1743.


python/config_settings/BUILD.bazel
```diff
@@ -10,7 +10,29 @@ filegroup(
...
+    # then, use the empty string to indicate an unknown version. This
+    # also prevents version-unaware targets from inadvertently matching
+    # a select condition when they shouldn't.
+    build_setting_default = "",
+    values = [""] + VERSION_FLAG_VALUES.keys(),
+    visibility = ["//visibility:public"],
 )
+
+[
+    is_python_config_setting(
```
aignas, line 25: TODO: add a test that checks that passing `constraint_values` to this function generates a list of `config_settings` that can be used in a single select statement with the others.

python/config_settings/BUILD.bazel
```diff
@@ -10,7 +10,29 @@ filegroup(
...
+        name = "is_python_{}".format(version),
+        flag_values = {":python_version": version},
+        match_extra = [
+            # Use the internal labels created by this macro in order to handle matching
+            # 3.8 config value if using the 3.8 version from MINOR_MAPPING. If we used the
+            # public labels we would have a circular dependency loop.
+            ("_is_python_{}" if VERSION_FLAG_VALUES[x] else "is_python_{}").format(x)
```
aignas, line 32: note: This is slightly smelly, but I am not sure if we can do anything better here. I do like having the definition of the `python_version` flag here better and using a more explicit way to define the public config setting values is also an improvement over the previous structure.

If we don't want to have this comment here, we can keep this in an internal function.

python/config_settings/config_settings.bzl
```diff
@@ -16,104 +16,109 @@
...
+        # Ensure that is_python_3.9.8 is matched if python_version is set
+        # to 3.9 if MINOR_MAPPING points to 3.9.8
         default_micro_version = MINOR_MAPPING[minor_version]
+        ret[micro_version] = [minor_version] if default_micro_version == micro_version else []
+
+    return ret
+
+VERSION_FLAG_VALUES = _flag_values(TOOL_VERSIONS.keys())
+
+def is_python_config_setting(name, flag_values, match_extra, **kwargs):
```
aignas, line 63: todo: move to a `//python/private` and re-export here if this design is suitable.

aignas: Note, that it is not possible because we would get a circular import chain. Leaving this as is.

aignas: Moved all of the code to private and re-exported.

python/config_settings/BUILD.bazel
rickeylev, line 25: Please keep the helper function with the name arg instead of using an inline list comp in the build file. This just makes it more accessible to tooling and keeps all the logic in one place.

python/private/config_settings.bzl
```diff
@@ -0,0 +1,134 @@
...
+    return (int(major), int(minor), int(micro))
+
+def _flag_values(python_versions):
+    """Construct a map of python_version to a list of toolchain values.
+
+    This mapping maps the concept of a config setting to a list of compatible toolchain versions.
+    For using this in the code, the VERSION_FLAG_VALUES should be used instead.
+
+    Args:
+        python_versions: A list of all versions.
```
rickeylev, line 36:
```suggestion
        python_versions: list of strings; all X.Y.Z python versions
```

python/private/config_settings.bzl
```diff
@@ -0,0 +1,134 @@
...
+def is_python_config_setting(name, *, python_version, match_extra = None, **kwargs):
+    """Create a config setting for matching 'python_version' configuration flag.
+
+    This function is mainly intended for internal use within the `whl_library` and `pip_parse`
+    machinery.
+
+    Args:
+        name: name for the target that will be created to be used in select statements.
+        python_version: The python_version to be passed in the `flag_values` in the `config_setting`.
+        match_extra: The labels that should be used for matching the extra versions instead of creating
```
rickeylev, line 74: Lets work "match_any" into the name of this arg to make it more clear the values here operate in a "or match any of ..." capacity instead of a "and must also match all of ..." capacity.

python/private/config_settings.bzl
```diff
@@ -0,0 +1,134 @@
...
+            them on the fly. This will be passed to `config_setting_group.match_any`.
+        **kwargs: extra kwargs passed to the `config_setting`
+    """
+    visibility = kwargs.pop("visibility", [])
+
+    flag_values = {
+        _PYTHON_VERSION: python_version,
+    }
+    if python_version not in name:
+        fail("The name must have the python version in it")
```
rickeylev, line 84:
```suggestion
        fail("The name '{}' must have the python version {} in it".format(name, python_version))
```

python/private/config_settings.bzl
```diff
@@ -0,0 +1,134 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""This module is used to construct the config settings in the BUILD file in this same package.
+"""
+
+load("@bazel_skylib//lib:selects.bzl", "selects")
+load("//python:versions.bzl", "MINOR_MAPPING", "TOOL_VERSIONS")
+
+_PYTHON_VERSION = str(Label("//python/config_settings:python_version"))
```
rickeylev, line 21: Just a small renaming for clarity. Seeing "python version" over and over again and I got lost with `_PYTHON_VERSION: python_version` -- like yeah why are we creating a tautology dict oh right thats the label to the flag on the LHS.

```suggestion
_PYTHON_VERSION_FLAG = str(Label("//python/config_settings:python_version"))
```

python/private/config_settings.bzl
```diff
@@ -0,0 +1,134 @@
...
+        name: name for the target that will be created to be used in select statements.
+        python_version: The python_version to be passed in the `flag_values` in the `config_setting`.
+        match_extra: The labels that should be used for matching the extra versions instead of creating
+            them on the fly. This will be passed to `config_setting_group.match_any`.
+        **kwargs: extra kwargs passed to the `config_setting`
+    """
+    visibility = kwargs.pop("visibility", [])
+
+    flag_values = {
+        _PYTHON_VERSION: python_version,
```
rickeylev, line 81:
```suggestion
        _PYTHON_VERSION_FLAG: python_version,
```

python/private/config_settings.bzl
```diff
@@ -0,0 +1,134 @@
...
+    """Construct a map of python_version to a list of toolchain values.
+
+    This mapping maps the concept of a config setting to a list of compatible toolchain versions.
+    For using this in the code, the VERSION_FLAG_VALUES should be used instead.
+
+    Args:
+        python_versions: A list of all versions.
+
+    Returns:
+        A map with config settings as keys and values as extra flag values to be included in
```
rickeylev, line 39: I think I know what you meant here, but the doc is a bit confusing. A `config_setting()` object can be a key in other contexts, but the keys of the dict this returns are plain strings like "3.8", not the label of a config setting.

I think you mean something like: the keys/values of the dict are fed into config_setting.flag_values ?

python/private/config_settings.bzl
```diff
@@ -0,0 +1,134 @@
...
+        _PYTHON_VERSION: python_version,
+    }
+    if python_version not in name:
+        fail("The name must have the python version in it")
+
+    match_extra = match_extra or {
+        "_{}".format(name).replace(python_version, x): {_PYTHON_VERSION: x}
+        for x in VERSION_FLAG_VALUES[python_version]
+    }
+    if not match_extra:
```
rickeylev, line 90: I don't see how this block could ever trigger? VERSION_FLAG values will always be non-empty, so it'll always generate a non-empty match_extra dict.

rickeylev (commented):
Sorry, ran out of time, didn't get all the way through.

python/private/config_settings.bzl
aignas: Thanks, the `match_extra` concept was a little difficult to understand, I think, changed the semantics to have `match_any` and it made things clearer.

python/private/config_settings.bzl
aignas: Changed the code semantics to use `match_any`.

python/private/config_settings.bzl
aignas: Let me know if the new thing is better.

python/config_settings/BUILD.bazel
```diff
@@ -10,7 +9,4 @@ filegroup(
     visibility = ["//python:__pkg__"],
 )
 
-construct_config_settings(
-    name = "construct_config_settings",
-    python_versions = TOOL_VERSIONS.keys(),
-)
```
aignas, line 16: Should I keep the args here?

python/private/config_settings.bzl
```diff
@@ -0,0 +1,143 @@
...
+
+"""This module is used to construct the config settings in the BUILD file in this same package.
+"""
+
+load("@bazel_skylib//lib:selects.bzl", "selects")
+load("//python:versions.bzl", "MINOR_MAPPING", "TOOL_VERSIONS")
+
+_PYTHON_VERSION_FLAG = str(Label("//python/config_settings:python_version"))
+
+def _ver_key(s):
```
aignas, line 23: I would like to leave this parsing of the version refactoring for a later PR.

tests/config_settings/construct_config_settings_tests.bzl
```diff
@@ -124,8 +158,55 @@ def _test_latest_micro_version_matching_impl(env, target):
 
 _tests.append(_test_latest_micro_version_matching)
 
-def construct_config_settings_test_suite(name):
+def construct_config_settings_test_suite(name):  # buildifier: disable=function-docstring
+    # We have CI runners running on a great deal of the platforms from the list below,
+    # hence use all of them within tests.
+    # NOTE @aignas 2024-02-12: contrary to what we
```
aignas, line 164:
```suggestion
```

aignas (commented):
Self review done, @rickeylev, I think this can be reviewed for a second time.

python/private/config_settings.bzl
```diff
@@ -71,23 +76,26 @@ def is_python_config_setting(name, *, python_version, match_extra = None, **kwar
...
 
-    match_extra = match_extra or {
-        "_{}".format(name).replace(python_version, x): {_PYTHON_VERSION: x}
-        for x in VERSION_FLAG_VALUES[python_version]
-    }
-    if not match_extra:
+    if python_version not in VERSION_FLAG_VALUES:
+        fail("The 'python_version' must be known to 'rules_python', choose from the values: {}".format(VERSION_FLAG_VALUES.keys()))
+
+    python_versions = VERSION_FLAG_VALUES[python_version]
```
rickeylev, line 97: In trying to understand the logic, I tried to rewrite it in a way that more directly represented what the doc said. The doc basically says, given python_version=X, and:
* Given match_any=None, then match any of the `3.X.*` states
* Given match_any=<some list>, then match of the `[3.X] + match_any` states (note no `.*` in this case). (Ah, actually my understanding on this part might be wrong...)

So, the two things we have to keep track of are (1) what do we have to create (create_config_settings)? and (2) what do we pass to config_setting_group.match_any (match_any)?

So, what about this:

```
if match_any == None:
  create_config_settings = {<build from VERSION_FLAG_VALUES[pythonn_version]}
  match_any = list(create_config_settings.keys())
else:
  create_config_settings = {_PYTHON_VERSION_FLAG: python_version}
  match_any = ...er, wait...

if len(match_any) == 1:
  assert create_config_settings is empty?? # ...er, wait ...
  native.config_setting(flag_values = match_any?)
else:
  for ... in create_config_settings: ...

  config_setting_group(match_any = match_any)
```

Er -- wait, what happened in the else-branch that assigned match_any? It looks like if match_any is used, then the python_version setting isn't part of the condition? I think that's intentional from what I vaguely recall of the use case motivating this PR. That wasn't clear from the doc and implementation, though. But also a bit weird? It means you end up passing python_version="3.8", but that variable isn't actually used besides (1) checking the name arg and (2) that the key exists in the VERSION_FLAG_VALUES dict, but no subsequent logic appears to rely on that?

And so, if we strip all that logic away, doesn't passing match_any effectively just make this a wrapper around calling `config_setting_group(match_any=match_any)` ?

Am I missing something?

python/config_settings/BUILD.bazel
aignas: We can make changes to the api later if a need arises.

python/config_settings/config_settings.bzl
```diff
@@ -15,105 +15,38 @@
...
-
-    allowed_flag_values = []
-    for micro_version in python_versions:
-        minor, _, _ = micro_version.rpartition(".")
-        minor_to_micro_versions.setdefault(minor, []).append(micro_version)
-        allowed_flag_values.append(micro_version)
-
-    allowed_flag_values.extend(list(minor_to_micro_versions))
+VERSION_FLAG_VALUES = _VERSION_FLAG_VALUES
+is_python_config_setting = _is_python_config_setting
```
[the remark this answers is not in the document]
aignas: Correct, added coments for that.

python/config_settings/config_settings.bzl
```diff
@@ -15,105 +15,38 @@
...
-    minor_to_micro_versions = {}
-
-    allowed_flag_values = []
-    for micro_version in python_versions:
-        minor, _, _ = micro_version.rpartition(".")
-        minor_to_micro_versions.setdefault(minor, []).append(micro_version)
-        allowed_flag_values.append(micro_version)
-
-    allowed_flag_values.extend(list(minor_to_micro_versions))
+VERSION_FLAG_VALUES = _VERSION_FLAG_VALUES
```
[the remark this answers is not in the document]
aignas: Made private

python/private/config_settings.bzl
aignas: Yeah, I think there was an issue with the concepts being used here. I reworked the code and I think it makes it sligtly clearer.

Thanks for raising this concern.

rickeylev (approved):

PR #1756, opened 2024-02-14
internal: support repo prefixes and config settings in alias rendering

aignas:
This does not change any logic/features in the bzlmod or legacy code, but
just changes the interfaces and how the parameters are passed. The final
result should be the same.

Summary:
- introduce a `whl_alias` struct to make code more object oriented.
- make the interface of the function as small as possible.
- unify the bzlmod and legacy code paths and simplify tests.
- allow to specify arbitrary repos and config settings
  when generating aliases, which is useful in multi-platform hub
  generation.


tests/pip_hub_repository/render_pkg_aliases/render_pkg_aliases_test.bzl
```diff
@@ -136,24 +135,25 @@ alias(
     name = "dist_info",
     actual = select(
         {
-            "@@rules_python//python/config_settings:is_python_3.2": "@pypi_32_bar_baz//:dist_info",
+            "@@//python/config_settings:is_python_3.2": "@pypi_32_bar_baz//:dist_info",
```
rickeylev, line 138: This `@rules_python//...` -> `@@//...` part of the diff is just incidental, right? Because the repo the tests are running in are the same as the target they're referring to. I'm just double checking that they'd be the proper "@rules_python//..." strings when run from a user repo.

python/private/bzlmod/pip_repository.bzl
```diff
@@ -27,10 +27,11 @@ exports_files(["requirements.bzl"])
 def _pip_repository_impl(rctx):
     bzl_packages = rctx.attr.whl_map.keys()
     aliases = render_pkg_aliases(
-        repo_name = rctx.attr.repo_name,
-        rules_python = rctx.attr._template.workspace_name,
+        aliases = {
+            key: [whl_alias(**v) for v in json.decode(values)]
```
rickeylev, line 31: Out of scope for this PR, but I wonder if passing a file might work better than passing it as strings in regular attributes. The pip hub basically needs to be given **all** the possible backing repos and targets, right? Which could be rather large. Putting that in a file seems like it'd scale better than the string attributes.

Oh hm...and that sort of sounds like the sort of intermediate file Phil's thing has been circling around...

python/private/bzlmod/pip.bzl
```diff
@@ -205,10 +207,14 @@ def _create_whl_repos(module_ctx, pip_attr, whl_map, whl_overrides):
...
-        if whl_name not in whl_map[hub_name]:
-            whl_map[hub_name][whl_name] = {}
-
-        whl_map[hub_name][whl_name][_major_minor_version(pip_attr.python_version)] = pip_name + "_"
+        major_minor = _major_minor_version(pip_attr.python_version)
+        whl_map[hub_name].setdefault(whl_name, []).append(
+            whl_alias(
+                repo = repo_name,
+                version = major_minor,
+                config_setting = Label("//python/config_settings:is_python_" + major_minor),
```
rickeylev, line 215: I'm guessing the Label() call is to canonicalize within the @rules_python context because the actual usage is in the `@pip` context? Just drop a comment to mention that. This still trips me up and I helped someone today with the same thing.

```suggestion
                # Call Label() to canonicalize because its used in a different context
                config_setting = Label("//python/config_settings:is_python_" + major_minor),
```

rickeylev (approved):

tests/pip_hub_repository/render_pkg_aliases/render_pkg_aliases_test.bzl
aignas: correct, this is incidental. Because we are doing `str(Label(...))` it should work (as is tested by the `bzlmod` example.

python/private/bzlmod/pip_repository.bzl
aignas: Interesting, I like the file idea, we could have a `label-keyed` dict here. Once I merge this, I can experiment with this idea.

tests/pip_hub_repository/render_pkg_aliases/render_pkg_aliases_test.bzl
```diff
@@ -15,14 +15,23 @@
 """render_pkg_aliases tests"""
 
 load("@rules_testing//lib:test_suite.bzl", "test_suite")
-load("//python/private:render_pkg_aliases.bzl", "render_pkg_aliases")  # buildifier: disable=bzl-visibility
+load("//python/private:render_pkg_aliases.bzl", "render_pkg_aliases", "whl_alias")  # buildifier: disable=bzl-visibility
+
+def _normalize_labels(want):
```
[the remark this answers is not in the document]
aignas: `@@` is not showing up in the actual values, but the hardcoded expected values have `@@` and this is transforming the hardcoded expected values to only have `@` under workspace builds. I added a docstring attempting to explain this, hopefully it is more clear right now.

PR #1759, opened 2024-02-14
fix(bcr): correct the gazelle module prefixing config

aignas:


kormide:
Please close this. Bad fix that I suggested. Real fix will be made in Publish to BCR.

PR #1764, opened 2024-02-16
internal(render_pkg_aliases): allow using default config setting

aignas:
This allows the code outside to decide what default is.
This is needed when we have the same version but different constraint
values as the logic where we use the 'default_version' breaks.

Noticed that otherwise the wheel hub repo default part may be non-deterministic in #1744.


aignas:
I'll close this for now.

PR #1770, opened 2024-02-18
feat(toolchains): Add armv7 linux platform to available toolchain platforms

UebelAndre:


UebelAndre:
I'm not sure what the intended behavior is here. I've compiled a python toolchain for armv7 which works totally fine in my environment but `rules_python` gets mad at me for describing a platform that it doesn't currently know about. I would love to be able to inform the rules but I don't really know how to go about that. I get that this platform is not currently defined since it's not otherwise available like the rest of the toolchains but this restriction gates me from developing on new platforms. I figured I'd try to open this PR and see what maintainers say so I know if I need to live with a "forever" patch or if there's a worthy contribution to be made.

aignas:
@UebelAndre , could you please paste the error you get without this PR? Maybe we are only fixing the symptom if we add the linux arm constant to the platforms, that said, I am fine to have the real fix in a followup PR if the the change may become too invasive.

UebelAndre:
> Can you explain this part a bit more? I was going to merge this PR, but this comment makes it sound like, while CI is passing, you are seeing errors elsewhere, something that CI isn't covering?

I was referring to the error here: https://github.com/bazelbuild/rules_python/pull/1770#issuecomment-1967079750

UebelAndre:
> Nice! So that must mean you're manually defining py_runtime(), toolchain() etc to wire it into the rest of things? Does this mean you'd be interested in something like `python.local_toolchain(path="/path/to/my/runtime")` in MODULE.bazel?

No, I structured my artifact to be similar to https://github.com/indygreg/python-build-standalone, so I'm not really manually defining anything, just trying to use the existing plumbing like what I described on https://github.com/bazelbuild/rules_python/pull/1770#issuecomment-1967079750

aignas:
It looks like the `platform` is not used for anything but checking if `windows` is in the platform string, which suspects me that we could potentially remove the keyed attribute. I understand that it would be then basically unkeyed free-form string, but I don't think if having something like this is sustainable in the long run either. Maybe having an `os`, `cpu`, `libc` triplet could be a better choice?

aignas:
I am inclined to just merge this as is as this improves the compatibility. Any objections @rickeylev?

UebelAndre:
> I am inclined to just merge this as is as this improves the compatibility. Any objections @rickeylev?

Does this mean no objections? 😅

aignas:
Could you please rebase? I think the changelog would be weirdly formatted if we merge as is.

aignas (approved):

UebelAndre:
> Could you please rebase? I think the changelog would be weirdly formatted if we merge as is.

@aignas done!

aignas (approved):

PR #1766, opened 2024-02-16
fix: remove 'site-packages' from install dir prefix

mattem:
Removes the `site-packages` path segment in unpacked external dependency installs.

This additional path segment causes issues when dealing with very large dependency trees, so much so that the value of `PYTHONPATH` exceeds the OS limit of what can be set. 

This manifests due to the Python binary launcher adding workspace roots automatically to sys path, in addition to the `site-packages` segmented path, ie the `PYTHONPATH` var will end up with both (from printing the `PYTHONPATH` in `examples/pip_parse` in rules_python.

Before:
```
/private/var/tmp/_bazel_matt/3cf6080de036ce65f48725921520350e/execroot/_main/bazel-out/darwin_arm64-fastbuild/bin/main.runfiles/rules_python~override~pip~pypi_39_certifi/site-packages
/private/var/tmp/_bazel_matt/3cf6080de036ce65f48725921520350e/execroot/_main/bazel-out/darwin_arm64-fastbuild/bin/main.runfiles/rules_python~override~pip~pypi_39_chardet/site-packages
/private/var/tmp/_bazel_matt/3cf6080de036ce65f48725921520350e/execroot/_main/bazel-out/darwin_arm64-fastbuild/bin/main.runfiles/rules_python~override~pip~pypi_39_idna/site-packages
/private/var/tmp/_bazel_matt/3cf6080de036ce65f48725921520350e/execroot/_main/bazel-out/darwin_arm64-fastbuild/bin/main.runfiles/rules_python~override~pip~pypi_39_urllib3/site-packages
/private/var/tmp/_bazel_matt/3cf6080de036ce65f48725921520350e/execroot/_main/bazel-out/darwin_arm64-fastbuild/bin/main.runfiles/rules_python~override~pip~pypi_39_requests/site-packages
/private/var/tmp/_bazel_matt/3cf6080de036ce65f48725921520350e/execroot/_main/bazel-out/darwin_arm64-fastbuild/bin/main.runfiles/rules_python~override~pip~pypi_39_alabaster/site-packages
/private/var/tmp/_bazel_matt/3cf6080de036ce65f48725921520350e/execroot/_main/bazel-out/darwin_arm64-fastbuild/bin/main.runfiles/rules_python~override~pip~pypi_39_babel/site-packages
[...]

PR #1776, opened 2024-02-23
feat(py_wheel): `deps` can provide requirement specifiers for `Requires-Dist`

pcj:
This feature adds two new providers (`PyRequirementInfo` and `PyRequirementsInfo`) to `py_wheel.bzl`.

Downstream users can use these Providers to supply additional requirement specifiers to the generated wheel metadata.

The use case is that we'd like wheels to contain files limited to transitive dependencies *not including* those which are available from pypi directly.  If we use `py_package` out of the box, we get "fat wheels" that contain too many files (all of `protobuf`, for example).  Rather, we'd like to be able to have our own "thinner" implementation of `py_package` that filters protobuf out, but also include `Requires-Dist: protobuf>=4.25.2` in the wheel metadata.

This is similar but more limited in scope than #1182.


pcj:
Example `py_project` rule that demonstrates usage:

```bazel

load("@rules_python//python:packaging.bzl", "PyRequrirementInfo", "PyRequrirementsInfo")

PYPI_LOCK_PREFIX = "../pypi/_lock/"

def _py_package_impl(ctx):
    inputs = depset(
        transitive = [dep[DefaultInfo].data_runfiles.files for dep in ctx.attr.deps] +
                     [dep[DefaultInfo].default_runfiles.files for dep in ctx.attr.deps],
    )

    files = []
    requirements = []

    for input_file in inputs.to_list():
        filename = input_file.short_path
        if filename.startswith(PYPI_LOCK_PREFIX):
            filename = filename[len(PYPI_LOCK_PREFIX):]
            name, version = filename.split("@")
            if name and version:
                requirements.append(PyRequrirementInfo(
                    name = name,
                    version = version,
                    specifier = "%s>=%s" % (name, version),
                ))
        else:
            files.append(input_file)

    return [DefaultInfo(
        files = depset(direct = files),
    ), PyRequrirementsInfo(
        label = ctx.label,
        requirements = requirements,
    )]

py_package = rule(
    doc = """\
A rule to select all files in transitive dependencies of deps which
belong to given set of Python packages.

This rule is intended to be used as data dependency to py_wheel rule.
""",
    implementation = _py_package_impl,
    attrs = {
        "deps": attr.label_list(
            doc = "",
        ),
    },
)
```

aignas:
I like the idea but was wondering if the py_project rule should be part of the rules_python, at least as an example if we are not sure that we want to increase the API surface we are maintaining. What do you think, @pcj?

pcj:
I'm happy to include it, perhaps under a directory like `examples/pycross_project/` since this particular implementation is specific to `rules_pycross`.

cc @jvolkman

jvolkman:
> I'm happy to include it, perhaps under a directory like `examples/pycross_project/` since this particular implementation is specific to `rules_pycross`?
> 
> cc @jvolkman

Not sure it makes sense to have this example in rules_python as written, with dependencies on `_lock` (a sort-of-implementation-detail, although it's described in a [comment](https://github.com/jvolkman/rules_pycross/blob/main/pycross/private/package_repo.bzl#L8-L9)) and `pypi` (a user-specified repo name).

Could the example be retooled to rely only on stuff in rules_python?

python/private/py_wheel.bzl
```diff
@@ -29,6 +29,23 @@ PyWheelInfo = provider(
     },
 )
 
+PyRequirementInfo = provider(
+    doc = "Information about an individual python requirement",
+    fields = {
+        "name": "String: the name of the pypi dependency (e.g. 'protobuf')",
+        "specifier": "String: a formatted string, to be used verbatim for 'Requires-Dist' metadata specification (e.g 'protobuf>=4.25.2', see https://pip.pypa.io/en/stable/reference/requirement-specifiers/)",
+        "version": "String: the concrete version of the pypi requirement (e.g. '4.25.2')",
```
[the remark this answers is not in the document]
pcj: Yes it is redundant.  I left `version` in because I thought *someone* might want to know the concrete version rather than having to parse the specifier, but since I don't have a specific use for it myself, should probably be removed for the sake of simplicity (could be added later)

python/private/py_wheel.bzl
```diff
@@ -29,6 +29,23 @@ PyWheelInfo = provider(
...
+    },
```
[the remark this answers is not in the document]
pcj: Probably?  I'm somewhat of an imposter when it comes to python, but after doing more reading it would make sense to include them (`List[String]`?)

pcj:
Added `examples/custom_py_project/`

.bazelci/presubmit.yml
```diff
@@ -316,17 +316,45 @@ tasks:
     working_directory: examples/multi_python_versions
     platform: windows
 
+  integration_test_custom_py_project_ubuntu_min_workspace:
```
aignas, line 319: This will deplete the number of runners we have available for running builds. I remember @rickeylev did some work to optimize this so maybe relying on the bazel integration test framework would be better here?

CHANGELOG.md
```diff
@@ -21,6 +21,9 @@ A brief description of the categories of changes:
 
 [0.XX.0]: https://github.com/bazelbuild/rules_python/releases/tag/0.XX.0
 
+* (py_wheel) Wheel deps can supply requirement specifiers to the generated wheel
```
aignas, line 24: nit: please rebase and put the bullet point into an appropriate section.

examples/custom_py_project/README.md
```diff
@@ -0,0 +1,23 @@
...
+- The `py_package` rule from `//bazel:py_package.bzl` rule processes each file in the default output, ultimately excluding anything that has `/site-packages/` in the path.
+  - the `/METADATA` file is used as a representative file from which we parse the
+    dependency name and version, creating a corresponding `PyRequirementInfo`.
+  - the `PyRequirementInfo` structs are bundled into `PyRequirementsInfo` and returned from the rule.
+  - the `PyRequirementsInfo` are picked up by the `py_wheel` rule implementation
+    and added to the generated wheel METADATA.
+  - the golden test depends on the generated wheel, whose METADATA has been
+    extracted and grepped for `Requires-Dist:`.
+
+> NOTE: this example was originally based off `examples/pip_parse_vendored`.
```
aignas, line 23: Any new example code should be compatible with `bzlmod`, IMHO so that it is easier to to maintain the examples in the future. Are there reasons to not do this in this case?

examples/custom_py_project/bazel/py_package.bzl
```diff
@@ -0,0 +1,58 @@
+"""custom implementation of py_package that filters out files that match */site-packages/*
+
+"""
+
+load("@rules_python//python:packaging.bzl", "PyRequrirementInfo", "PyRequrirementsInfo")
+
+SITE_PACKAGES = "/site-packages/"
```
aignas, line 7: This may break if we merge #1766, I wonder if there is a better way to do this.

examples/custom_py_project/README.md
pcj: I just copied over the files from `examples/pip_parse_resolved`, but perhaps a different example would be better as a base.  Alternatively, I can just remove the WORKSPACE bits.

Issue #792, opened 2022-08-16
using all_requirements causes OSError: [Errno 7] Argument list too long


anuragkanungo:
# 🐞 bug report

### Affected Rule

The issue is caused by the rule:  `all_requirements`

### Is this a regression?

Yes, the previous version in which this bug was not present was: 0.9.0
The bug exists in : 0.10.2 and 0.11.0

### Description

When using `all_requirements` in a rule like
```
py_binary(
    name = "mybinary",
    srcs = ["main.py"],
    main = "main.py",
    deps = all_requirements,
)
```

It cause the error

```
Traceback (most recent call last):
  File "/home/user/.cache/bazel/_bazel_user/bde0c2a7e98d3dd9f99d720839ba72c4/execroot/org/bazel-out/k8-fastbuild/bin/third_party/ipython/ipython", line 392, in <module>
    Main()
  File "/home/user/.cache/bazel/_bazel_user/bde0c2a7e98d3dd9f99d720839ba72c4/execroot/org/bazel-out/k8-fastbuild/bin/third_party/ipython/ipython", line 382, in Main
    os.execv(args[0], args)
OSError: [Errno 7] Argument list too long: '/home/user/.cache/bazel/_bazel_user/bde0c2a7e98d3dd9f99d720839ba72c4/execroot/org/bazel-out/k8-fastbuild/bin/third_party/ipython/ipython.runfiles/python_x86_64-unknown-linux-gnu/bin/python3'
```

## 🌍  Your Environment

**Operating System:** Ubuntu 20.04
**Output of `bazel version`:** 5.2.0

hrfuller:
This is an issue with the upstream native bazel rules not rules_python. The problem is that the PYTHONPATH set by the python_stub_template is longer than the allowed system limits. You can try to set `--noexperimental_python_import_all_repositories` to avoid adding PYTHONPATH entries for all external repositories, but you may encounter issues with rules_k8s and others that rely on implicit repository imports. Another possible path forward would be to use https://github.com/aspect-build/rules_py which uses a site-packages based import system for python binaries IIRC.

anuragkanungo:
@hrfuller, the issue happens doesn't exist with `rules_python` version 0.9.0 but it does after that, in the same list of packages. I believe something did change in the refactor that happened after 0.9.0.

el-gee:
[...]

Issue #1174, opened 2023-04-19
Packages shadowed by other modules are no longer importable under bzlmod


aschleck:
# 🐞 bug report

### Affected Rule

 The issue is caused by the rule: `py_binary`/`py_library` under bzlmod

### Is this a regression?

 Yes, this works under the WORKSPACE model. It is only bzlmod that is broken.

### Description

When two Bazel workspaces (with one depending on the other) both define packages with the same name (`common`, `core`, `helpers`, `private`, `util`, ...) the outer one will shadow the inner one. Without bzlmod this has always been solvable by using absolute imports with the workspace name as the prefix. However, under bzlmod, these absolute imports no longer work: when a module is used as an outer repository the prefix is `_main` but when a module is used as a dependency repository it is `<module name>`. And even if that was fine, there are further problems caused by `local_path_override` because it suffixes `~override` onto the folders in the runfiles directory.

From my perspective the best solution would be to make the generated `py_binary` wrapper script read the generated `_repo_mapping` file and set up the path such that absolute imports work the same way they've always worked. One way to make that happen would be to have the wrapper script create a temporary directory, create symlinks with the correct names pointing to the real folders, and then add the temporary directory to the PYTHONPATH. I could imagine there is probably some cleaner way I can't think of.

## 🔬 Minimal Reproduction

I put an example here: https://github.com/aschleck/bzlmod-shadow-python-problem (with another writeup in `README.md`.)

This repository illustrates three situations:
* [shadowed_with_workspace](https://github.com/aschleck/bzlmod-shadow-python-problem/tree/master/shadowed_with_workspace) shows a workspace depending on another workspace where both define a package named `core`. In this case, the `outer` workspace imports `dep` absolutely but `dep`'s import of itself fails due to being shadowed.
* [with_workspace](https://github.com/aschleck/bzlmod-shadow-python-problem/tree/master/with_workspace) illustrates the obvious fix: `dep` can always import itself with its workspace name as a prefix (like `import dep.core.<whatever>`. Another solution would be to use relative imports, but those get unwieldy and annoying in large repos.
[...]

PR #1780, opened 2024-02-27
fix: Stop using PYTHONPATH to set up sys.path

philsc:
This patch is an experiment for setting up `sys.path` inside
the final Python process instead of letting the runtime do it
via PYTHONPATH.

The motivation here is to address these 2 issues:
* #382 
* #792 

Try it with:
```
$ bazel run //tools:wheelmaker -- --help
```
Look at the generated files.

### DIFF
--- a/python/private/common/entrypoint.tmpl.py
+++ b/python/private/common/entrypoint.tmpl.py
@@ -1,0 +1,21 @@
+import runpy
+import sys
+from pathlib import Path
+
+from python.runfiles import runfiles
+
+RUNFILES = runfiles.Create()
+RUNFILES_DIR = Path(RUNFILES.EnvVars()["RUNFILES_DIR"])
+MAIN = RUNFILES_DIR / "%MAIN_REPO%/%MAIN_SHORT_PATH%"
+
+IMPORTS = %IMPORTS%
+
+# Work around sys.path[0] escaping the sandbox by deleting it.
+# See https://github.com/bazelbuild/rules_python/issues/382 for more info.
+# TODO(phil): How do we distinguish between safe-path Python and non-safe-path
+# Python? I.e. how do we know if sys.path[0] should be deleted or not?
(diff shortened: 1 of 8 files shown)


python/private/common/entrypoint.tmpl.py
```diff
@@ -0,0 +1,21 @@
...
+
+RUNFILES = runfiles.Create()
+RUNFILES_DIR = Path(RUNFILES.EnvVars()["RUNFILES_DIR"])
+MAIN = RUNFILES_DIR / "%MAIN_REPO%/%MAIN_SHORT_PATH%"
+
+IMPORTS = %IMPORTS%
+
+# Work around sys.path[0] escaping the sandbox by deleting it.
+# See https://github.com/bazelbuild/rules_python/issues/382 for more info.
+# TODO(phil): How do we distinguish between safe-path Python and non-safe-path
```
aignas, line 15: For inspiration, see: https://github.com/bazelbuild/rules_python/blob/627830e0756e6f7c09f2a95561963121dff9c166/python/private/py_console_script_gen.py#L50

aignas:
Do you think it would help #1174 as well?

python/private/common/entrypoint.tmpl.py
philsc: Super cool. Thank you!

philsc:
> Do you think it would help #1174 as well?

Not as-is, no. But I think it _could_. In the `python/private/common/entrypoint.tmpl.py` file we could add an import hook that tries to resolve imports with the repository name.

kwlzn:
hmm, I noticed this was closed without merging. any reason why? we are hitting this in our repo and would love to contribute to a proper fix. @philsc @rickeylev

aignas:
The solution using the script bootstrap and making it the default is in #2521. See our official docs config settings page on how to enable it in the meantime.

PR #1787, opened 2024-03-01
feat(gazelle): Add "python_default_visibility" directive

### LINKED ISSUE #1682: gazelle: Visibilities are incorrect for projects with multiple `# gazelle:python_root` directives

### LINKED ISSUE #1783: Add `python_visibility` gazelle directive
# 🚀 feature request

### Relevant Rules

+ gazelle

### Description

**Summary:** Add a `py(thon)_visibility` gazelle directive that acts like the `go_visibility` directive. This directive would append additional labels to the `visibility` attribute on generated targets.

**Example:**

```starlark
# gazelle:python_visibility //tests:__pkg__
# gazelle:python_visibility //bar:baz

py_library(
    name = "directive_python_visibility",
    srcs = ["foo.py"],
    visibility = [
        "//:__subpackages__",
        "//bar:baz",
        "//tests:__pkg__",
    ],
)
```

I have a branch that does this already. I'm mostly making this Issue for discussion before I submit the PR.

Branch: https://github.com/dougthor42/rules_python/tree/gazelle-python-visibility

Manually adding other vis labels is the only other option I've found. That's not really an option when adding gazelle to an existing, large project.

dougthor42:
Fixes #1783. Provides a way to fix #1682.

Add the `python_default_visibility` directive. This directive sets the
`visibility` attribute on all generated `py_*` rules. It accepts a
comma-separated list of labels to add as visibility targets, similar
to how the base `default_visibility` directive works.

Setting this directive multiple times will override previous values.

Two special values are also accepted: `NONE` and `DEFAULT`. See
./gazelle/README.md#directive-python_default_visibility for details.

The directive also accepts a special string `"$python_root"` that gets
replaced with the `python_root` value, if set. If not set, `"$python_root"`
is replaced with the empty string.

### DIFF
--- a/gazelle/README.md
+++ b/gazelle/README.md
@@ -245,6 +245,7 @@ py_libary(
 Instructs gazelle to use these visibility labels on all _python_ targets
 (typically `py_*`, but can be modified via the `map_kind` directive). The arg
 to this directive is a a comma-separated list (without spaces) of labels.
+
 For example:
 
 ```starlark
@@ -283,6 +284,38 @@ py_library(
     ...,
 )
 ```
+
+Two special values are also accepted as an arg to the directive:
+
++   `NONE`: This removes all default visibility. Labels added by the
+    `python_visibility` directive are still included.
++   `DEFAULT`: This resets the default visibility.
+
+For example:
+
+```starlark
+# gazelle:python_default_visibility NONE
+
+# results in no `visibility` attribute being set:
+py_library(
+    name = "...",
+    srcs = [...],
+)
+```
+
+```starlark
+# gazelle:python_default_visibility //foo:bar
+# gazelle:python_default_visibility DEFAULT
+
+# results in:
+py_library(
+    ...,
+    visibility = ["//:__subpackages__"],
+    ...,
+)
+```
+
+These special values can be useful for descendant `BAZEL.build` packages.
 
 
 #### Directive: `python_visibility`:
--- a/gazelle/README.md
+++ b/gazelle/README.md
@@ -269,11 +269,10 @@ You can also inject the `python_root` value by using the exact string
 `$python_root`:
 
 ```starlark
-# an absurd example
 # gazelle:python_default_visibility //$python_root:__pkg__,//foo/$python_root/tests:__subpackages__,//$python_root/$python_root/foo:bar
 
-# assuming the "# gazelle:python_root" directive is set in ./py/src/BUILD.bazel
-# results in:
+# Assuming the "# gazelle:python_root" directive is set in ./py/src/BUILD.bazel,
+# the results will be:
 py_library(
     ...,
     visibility = [
@@ -285,7 +284,7 @@ py_library(
 )
 ```
 
-Two special values are also accepted as an arg to the directive:
+Two special values are also accepted as an argument to the directive:
 
 +   `NONE`: This removes all default visibility. Labels added by the
     `python_visibility` directive are still included.
@@ -296,7 +295,6 @@ For example:
 ```starlark
 # gazelle:python_default_visibility NONE
 
-# results in no `visibility` attribute being set:
 py_library(
     name = "...",
     srcs = [...],
@@ -307,7 +305,6 @@ py_library(
 # gazelle:python_default_visibility //foo:bar
 # gazelle:python_default_visibility DEFAULT
 
-# results in:
 py_library(
     ...,
     visibility = ["//:__subpackages__"],
@@ -315,7 +312,7 @@ py_library(
 )
 ```
 
-These special values can be useful for descendant `BAZEL.build` packages.
+These special values can be useful for sub-packages.
 
 
 #### Directive: `python_visibility`:
--- a/gazelle/README.md
+++ b/gazelle/README.md
@@ -266,10 +266,11 @@ py_library(
 ```
 
 You can also inject the `python_root` value by using the exact string
-`$python_root`:
+`$python_root`. All instances of this string will be replaced by the `python_root`
+value.
 
 ```starlark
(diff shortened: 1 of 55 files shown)


dougthor42:
I think this is ready for review now.

Sorry for the big PR! Most of it is just test cases though.

gazelle/python/configure.go
```diff
@@ -163,6 +167,20 @@ func (py *Configurer) Configure(c *config.Config, rel string, f *rule.File) {
 			config.SetBinaryNamingConvention(strings.TrimSpace(d.Value))
 		case pythonconfig.TestNamingConvention:
 			config.SetTestNamingConvention(strings.TrimSpace(d.Value))
+		case pythonconfig.DefaultVisibilty:
+			switch directiveArg := strings.TrimSpace(d.Value); directiveArg {
+			case "NONE":
+				config.SetDefaultVisibility([]string{})
+			case "DEFAULT":
```
dougthor42, line 174: I also thought of `"RESET"` instead.

gazelle/python/configure.go
```diff
@@ -99,6 +100,8 @@ func (py *Configurer) Configure(c *config.Config, rel string, f *rule.File) {
 	}
 
 	gazelleManifestFilename := "gazelle_python.yaml"
+	// TODO: figure out how to keep this in sync with pythonconfig.go New *Config
+	defaultVisibilityFmtString := "//%s:__subpackages__"
```
f0rmiga, line 104: Make this an exported `const` in the `pythonconfig.go`.

gazelle/python/configure.go
f0rmiga: I think `DEFAULT` is more self-descriptive.

gazelle/README.md
```diff
@@ -238,6 +240,84 @@ py_libary(
...
+    ...,
+)
+```
+
+You can also inject the `python_root` value by using the exact string
+`$python_root`:
+
+```starlark
+# an absurd example
+# gazelle:python_default_visibility //$python_root:__pkg__,//foo/$python_root/tests:__subpackages__,//$python_root/$python_root/foo:bar
```
f0rmiga, line 273: Thinking more, `//$python_root/$python_root/foo:bar` will probably confuse users. Let's keep this more straightforward and remove it?

gazelle/python/testdata/directive_python_default_visibility/test6_default_label/subpkg/BUILD.out
```diff
@@ -0,0 +1,10 @@
+load("@rules_python//python:defs.bzl", "py_library")
+
+# Reset the default visibility to the default for all child packages.
+# gazelle:python_default_visibility DEFAULT
```
f0rmiga, line 4: What if we set `# gazelle:python_visibility` together with `# gazelle:python_default_visibility`?

f0rmiga: How about a test case that uses `# gazelle:python_visibility` on a sub-package?

gazelle/README.md
```diff
@@ -238,6 +240,84 @@ py_libary(
...
+
+# results in:
+py_library(
+    ...,
+    visibility = ["//:__subpackages__"],
+    ...,
+)
+```
+
+These special values can be useful for descendant `BAZEL.build` packages.
```
f0rmiga, line 318:
```suggestion
These special values can be useful for sub-packages.
```

f0rmiga (changes requested):

gazelle/python/configure.go
dougthor42: Thanks, done.

aignas (commented):
LGTM, but would be great to have a `CHANGELOG.md` entry announcing the feature. I think pointing users to `gazelle` docs would be sufficient enough.

It's really nice to have extensive test coverage!

@f0rmiga, what do you think about readiness of this?

f0rmiga (approved):

f0rmiga:
Thanks for addressing my comments! Thanks @aignas for the review. I'll let you merge this when the changelog looks good to you.

aignas (approved):

PR #1788, opened 2024-03-03
feat: Download platform agnostic wheels using rctx.download

michaelboulton:
If we can determine that there's a platform agnostic wheel (dep-ver-py3-none-any.whl) then download it using rctx.download, allowing it to be cached.

This partially solves https://github.com/bazelbuild/rules_python/issues/1357 but it still doesn't handle dependencies that have platform specific binary wheels like https://pypi.org/project/psycopg2-binary/ .

(I don't know the ins and outs of pip package resolution, but it seems like there is no easy way to get pip just to print "what wheel should be downloaded for this platform" without actually downloading it, which is why this PR can only handle platform agnostic wheels)

### DIFF
--- a/python/pip_install/pip_repository.bzl
+++ b/python/pip_install/pip_repository.bzl
@@ -416,6 +416,56 @@ def _pip_repository_impl(rctx):
 
     return
 
+def _try_finding_platform_agnostic_wheel(rctx, requirement):
+    """Tries to find a platform agnostic wheel for the given requirement, otherwise returns None"""
+    split_req = [i for i in requirement.split(" ") if i != ""]
+    requirement_name, requirement_version = split_req[0].split("==")
+    if "[" in requirement_name:
+        # Remove 'extras' specification
+        requirement_name = requirement_name.split("[")[0]
+
+    # Get all possible expected hashes
+    requirement_hashes = [s.removeprefix("--hash=") for s in split_req[1:]]
+    split_hashes = [s.split(":") for s in requirement_hashes]
+    by_hash_type = {}
+    for alg, digest in split_hashes:
+        if alg in by_hash_type:
+            by_hash_type[alg].append(digest)
+        else:
+            by_hash_type[alg] = [digest]
+
+    rctx.download(
+        url = "https://pypi.org/pypi/{}/{}/json".format(requirement_name, requirement_version),
+        output = "package_spec.json",
+    )
+    package_contents = json.decode(rctx.read("package_spec.json"))
+    if not rctx.delete("package_spec.json"):
+        fail("failed to delete the package_spec.json file")
+
+    # Filter yanked uploads
+    downloads_for_tag = [d for d in package_contents["urls"] if not d["yanked"]]
+    platform_agnostic_wheels = [d for d in downloads_for_tag if d["filename"].endswith("-py3-none-any.whl")]
+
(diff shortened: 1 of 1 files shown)


python/pip_install/pip_repository.bzl
```diff
@@ -416,6 +416,56 @@ def _pip_repository_impl(rctx):
...
+    split_hashes = [s.split(":") for s in requirement_hashes]
+    by_hash_type = {}
+    for alg, digest in split_hashes:
+        if alg in by_hash_type:
+            by_hash_type[alg].append(digest)
+        else:
+            by_hash_type[alg] = [digest]
+
+    rctx.download(
+        url = "https://pypi.org/pypi/{}/{}/json".format(requirement_name, requirement_version),
```
jvolkman, line 438: There are many people that use private or alternative indexes; hard-coding `pypi.org` wouldn't work in these scenarios.

Also, I'm not sure that the JSON api is widely supported outside of pypi. [Artifactory](https://jfrog.com/artifactory/) is a common repository and it only supports the simple (HTML) API as far as I can tell.

python/pip_install/pip_repository.bzl
```diff
@@ -766,18 +816,28 @@ def _whl_library_impl(rctx):
...
-        rctx,
-        op = "whl_library.ResolveRequirement({}, {})".format(rctx.attr.name, rctx.attr.requirement),
-        arguments = args,
-        environment = environment,
-        quiet = rctx.attr.quiet,
-        timeout = rctx.attr.timeout,
-    )
+    target_platforms = rctx.attr.experimental_target_platforms
+    platform_agnostic_wheel = _try_finding_platform_agnostic_wheel(rctx, rctx.attr.requirement)
+    if platform_agnostic_wheel and not target_platforms:
```
jvolkman, line 821: If my understanding is correct, this logic would prefer a pure-python wheel over a native wheel which is not actually what `pip` would do. Pip would prefer a native wheel if one exists for the target platform, and this does occur in a handful of packages that ship native extensions as a performance optimization for certain platforms and a pure-python wheel for the rest ([SQLAlchemy](https://pypi.org/project/SQLAlchemy/2.0.27/#files) does this, for example).

jvolkman (commented):
I'm pretty sure @aignas had a PR somewhere that used Bazel's downloader in some situations, but I think it was using the pypi simple API, not the JSON API. Not sure what happened to that?

aignas:
Yeah it is #1625, Right now I am hoping to get #1744 done first and then come back to this topic so that we can use the Simple API for all of those things.

python/pip_install/pip_repository.bzl
michaelboulton: I was using the simple API  and switched to using the JSON API because it was far simpler to just parse the output for one specific version :shrug:  I wasn't really aware it was widely unsupported

michaelboulton:
> Yeah it is #1625, Right now I am hoping to get #1744 done first and then come back to this topic so that we can use the Simple API for all of those things.

I had a look at something like this but doing all the platform detection appropriately seemed like a big chunk of work (as I can see in those PRs) which was overkill for my simple use case. I'm glad someone is looking at it though, and as your work is far more fleshed out I'll close this PR.

(If anyone does want to use this patch, it worked for me just to speed up clean builds, but YMMV)

PR #1792, opened 2024-03-10
feat(bzlmod): introduce pypi_index for using bazel's downloader

aignas:
This is a variant of #1625 and was inspired by #1788. In #1625, we
attempt to parse the simple API HTML files in the same `pip.parse`
extension and it brings the follownig challenges:

* The `pip.parse` cannot be easily use in `isolated` mode and it may
  be difficult to implement the isolation if bazelbuild/bazel#20186
  moves forward.
* Splitting the `pypi_index` out of the `pip.parse` allows us to accept
  the location of the parsed simple API artifacts encoded as a bazel
  label.
* Separation of the logic allows us to very easily implement usage of
  the downloader for cross-platform wheels.
* The `whl` `METADATA` might not be exposed through older versions of
  Artifactory, so having the complexity hidden in this single extension
  allows us to not increase the complexity and scope of `pip.parse` too
  much.
* The repository structure can be reused for `pypi_install` extension
  from #1728.

TODO:
- [ ] Add unit tests for functions in `pypi_index.bzl` bzlmod extension if
  the design looks good.
- [ ] Changelog.

Work towards #1357, #1643.


aignas:
Out of scope of this PR:
- Further usage of the downloaded artifacts to implement something
  similar to #1625 or #1744. This needs #1750 and #1764.
- Making the lock file the same on all platforms - We would need
  to fully parse the requirements file.
- Support for different dependency versions in the `pip.parse` hub repos
  based on each platform - we would need to be able to interpret
  platform markers in some way, but `pypi_index` should be good already.
- Implementing the parsing of METADATA to detect dependency cycles.
- Support for `requirements` files that are not created via
  `pip-compile`.
- Support for other lock formats, though that would be reasonably
  trivial to add.

Open questions:
- Support for VCS dependencies in requirements files - We should
  probably handle them as `overrides` in the `pypi_index` extension and
  treat them in `pip.parse` just as an `sdist`, but I am not sure it
  would work without any issues.

aignas:
This POC has served well, but I'll close it to focus on more targeted PRs that implement ideas outlined in [this design doc](https://docs.google.com/document/d/1vAupaen4tIZxeqdpkYleuUUjrVuptdps_mz4Y_V4EIo/edit#heading=h.ga06bz2494ba). Since the doc uses this PR as reference material I'll close this and start fresh.

Issue #1578, opened 2023-11-25
Support for keyring auth with bzlmod pip parse


brett-patterson-ent:
# 🚀 feature request

### Relevant Rules

In `MODULE.bazel`:
```bazel
pip = use_extension("@rules_python//python/extensions:pip.bzl", "pip")
pip.parse(
  ...
)
```

### Description

All of our PyPI access is through a private Artifact Registry repo that requires authentication. We typically do this authentication through the use of [`keyring`](https://cloud.google.com/artifact-registry/docs/python/authentication#keyring), which requires the keyring packages to be installed for the Python interpreter that is doing the pip installation.

By default, a hermetic installation of Python won't have these packages installed. While it's a bit of a chicken/egg problem, I've made some progress by taking inspiration from rules_python's [internal dependencies](https://github.com/bazelbuild/rules_python/blob/main/python/pip_install/repositories.bzl) and fetching the necessary wheels using `http_archive`. I was able to get this working with `compile_pip_requirements` by using the `extra_deps` argument so that the `dependency_resolver.py` script runs with the keyring libraries available.

However, I've been unable to get the same approach working with `pip.parse` so that `wheel_installer.py` runs with these libraries available. Is there a standard way to do this? Or a way to take a similar approach of bundling additional dependencies into the tool's PYTHONPATH? I found that there was a [private attribute](https://github.com/bazelbuild/rules_python/blob/da10ac49efee1b02cbfa3b22a39e68bf3fe5bbe2/python/pip_install/pip_repository.bzl#L921) specifying this.

### Describe the solution you'd like

 
A way to pipe down additional dependencies to include in the `_python_path_entries` attribute would work, I think. However, given that the actual `pip` installation happens in a subprocess I'm not 100% sure of that solution. I'm open to any different approaches that would be better.

### Describe alternatives you've considered

It looks like `pip.parse` does let you pass in an `environment` argument that (I think) ultimately gets set for the pip subprocess inside of `wheel_installer.py`. I experimented with setting `PYTHONPATH` there but was unsuccessful.

I also did get a setup working with `WORKSPACE` rather than bzlmod where the `pip_parse` call uses the system interpreter, but I'd like to use bzlmod and not involve the system Python if at all possible.

dougthor42:
I'm running into the same issue. Can you elaborate on:

> I was able to get this working with `compile_pip_requirements` by using the `extra_deps` argument so that the `dependency_resolver.py` script runs with the keyring libraries available.

[...]

PR #1794, opened 2024-03-11
build(internal_deps): Update 'maintained by' comments to be correct

dougthor42:
This fixes a minor error in the "maintained by" comments in
`./MODULE.bazel` and `./python/pip_install/repositories.bzl`. The
comments now agree with what `DEVELOPING.md` says.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -17,7 +17,7 @@ internal_deps.install()
 use_repo(
     internal_deps,
     "rules_python_internal",
-    # START: maintained by 'bazel run //tools/private:update_pip_deps'
+    # START: maintained by 'bazel run //tools/private/update_deps:update_pip_deps'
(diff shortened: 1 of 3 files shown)


dougthor42:
Note: I'm not sure if `fix(internal_deps)` is the best type and scope for the PR. The PR has no impact on end users but does have a very minor impact on developers.

I found this while working on #1578.

aignas (approved):

aignas:
@coeuvre, do you know what the build failure is about? It seems to fail with the RBE setup, but this PR hasn't changed anything, nor have we changed `rules_python` config regarding RBE in recent days.

coeuvre:
@aignas Sorry, I recently made some changes to Bazel CI and accidentally broke RBE config for Bazel 6.4.0. It is fixed now.

PR #1795, opened 2024-03-11
fix(pip): Add `keyring` dev dependency, allowing additional pypi indexes to use it for auth

### LINKED ISSUE #1578: Support for keyring auth with bzlmod pip parse

dougthor42:
WIP.

Fixes #1578. Probably.


aignas (commented):
For my curiosity. `keyring` sounds like something interacting with the OS and I wonder how this works with the bazel sandbox.

Does using `.netrc` not work for your usecase? I know that it would be less secure, but just curious about the parts that you need to get working:
1. Is it `pip-compile`?
2. Or is it `pip.parse` `bzlmod` extension?

Right now the PR would affect both.

dougthor42:
> `keyring` sounds like something interacting with the OS

Yes, it typically does. There _may_ be an option/implementation that doesn't touch the OS, but I'm not aware of how to use it if such a thing exists.

As for the sandbox, sadly I have no idea. I'm still quite new to bazel. What I do know is that adding these deps appeared to work so far :rofl:

> Does using `.netrc` not work?

We don't auth using a username and password, so I don't think that'll work. We're using the the [Google Application Default Credentials "ADC"](https://cloud.google.com/docs/authentication/provide-credentials-adc) to auth.

Note that this PR does not include support for Google ADC. I'm still playing around with how I want to add support for such. Maybe a directive...? but that's a topic for another time.

> parts that you need to get working?

`pip-compile` to start, but `pip.parse` will also need to know about the extra index url. Right now the priority is getting the requirements lockfile generation working.

Background: I'm trying to add `bazel`, `rules_python` and `gazelle` to an existing monorepo, and I'd love to to use `bzlmod` so that we don't have to migrate later.

The first step is to get a python requirements lockfile that can be fed into `pip.parse`. I've opted to use `bazel run //:requirements.update` for such as we already have a `requirements.in`-like file. To get that to work, I followed the instructions described in #1578. The net result was:

```starlark
# MODULE.bazel
http_archive = use_repo_rule("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")

_GENERIC_WHEEL = "..." # same as ./python/pip_install/repositories.bzl#L100

http_archive(
    name = "pypi__keyring",
    ...
)

# repeat http_archive for all deps required by (a) keyring and (b) keyrings.google-artifactregistry-auth
# keyrings.google-artifactregistry-auth
# google-auth
# requests
# charset-normalizer
# idna
# urllib3
# certifi
# pyasn1
# pyasn1-modules
# rsa
# cachetools
```

```starlark
# BUILD.bazel
compile_pip_requirements(
    name = "requirements",
    src = "requirements.txt",

    extra_args = [
        # Pull from our package registry
        "--extra-index-url",
        "https://us-west2-python.pkg.dev/[REDACTED]/simple",
          # Just verbose enough so that we can see pip_tools is actually doing something
        "-v",
    ],
    # Until https://github.com/bazelbuild/rules_python/pull/1795 is merged, we have
    # to do manually add the keyring and keyrings.google-artifactregistry-auth
    # dependencies along with __all__ transitive deps (well, all that are not already
    # part of rules_python)
    extra_deps = [
        "@pypi__keyring//:lib",
        "@pypi__keyrings_google_artifactregistry_auth//:lib",
        "@pypi__google_auth//:lib",
        "@pypi__requests//:lib",
        "@pypi__charset_normalizer//:lib",
        "@pypi__idna//:lib",
        "@pypi__urllib3//:lib",
        "@pypi__certifi//:lib",
        "@pypi__pyasn1//:lib",
        "@pypi__pyasn1_modules//:lib",
        "@pypi__rsa//:lib",
        "@pypi__cachetools//:lib",
    ],
)
```

The 2nd step will be to make sure that `pip.parse` correctly installs all dependencies, inlcuding those from our private index.

Then I can move on to the rest of the gazelle stuff (`gazelle_python_manifest.update`, `bazel run //:gazelle`).

dougthor42:
Thinking on it a little more, a better solution might just be to allow users to define what extra internal deps should be installed:

```starlark
pip.parse(
    hub_name = "pypi",
    python_version = "3.10",
    additional_internal_python_deps = [    # new!
        "keyring",
        "keyrings.google-artifactregistry-auth==1.0.0",
    ],
)
use_repo(pip, "pypi")
```

The `additional_internal_python_deps` arg would accept a list of pip requirement specifier strings and would install those into the internal dep tree early on in the process. That way the `rules_python` project doesn't have to explicitly cater to all types of pypi indexes and their keyring requirements - we could make the user do that instead.

Thoughts? If that makes more sense I'll work on such a thing.

aignas:
I am thinking that in the long term we should use bazel downloader to download artifacts from PyPI and this means that we should go towards a solution where the bazel's credential helper is used. This would mean that we can extract the files downloaded by the downloader (which may hook into the credential helper) with fewer dependencies.

This would work for `pip.parse`, but it would not for the `pip-compile`. For the `pip-compile` case I think adding extra dependencies that call to the system's credential helper would be ideal and using [`keyring`](https://pypi.org/project/keyring/) seems like the best way here.

In the mean time, I think asking for doing `pip install` before downloading the wheels might not work out, doing this many times might cause a lot of overhead. I am thinking that it may be better to instead have a custom repository rule, that extends the `host_python` to include the extra packages you want and you could pass the interpreter via the `interpreter_target`.

Thinking about it more, I am wondering if it is possible to do it easily and my guess is that it isn't. However, getting all of the dependencies needed to work with keyring effectively may require some fiddling. Maybe having a repo rule that downloads a list of dependencies and extracts it in a single `site-packages` could be an option here? That way the user could say:
```
pip.parse(
    build_deps = [
        "@my_repo//path:BUILD.bazel",
    ]
)
```

I'll post it here for posterity and I'll have to think about it more. The easiest solution would be to
> Perform locking outside bazel
but that may not be the solution we want here.

dougthor42:
The issue I see with supporting `keyring` is that there are many different artifact registries and many different keyring providers for them.

+ GCP's Artifact Registry, already mentioned in this PR, uses a special keyring provider `google-artifactregistry-auth`
+ AWS CodeArtifact looks like it uses URLs that contain auth tokens, so the base `keyring` is probably sufficient, if needed at all.
+ Azure Artifacts uses another special keyring provider [`artifacts-keyring`](https://github.com/microsoft/artifacts-keyring)
+ on-prem / self-hosted pypi indexes use a variety of things, one example is [`pypiserver`](https://pypi.org/project/pypiserver/) uses [`passlib`](https://pypi.org/project/passlib/).

It's a bit of a dependency explosion.

----

> I think asking for doing `pip install` before downloading the wheels might not work out, doing this many times might cause a lot of overhead.

Maybe I'm not familiar enough with bazel yet, but aren't both of these statements true?

1.  The `pip install` for the internal rules_python/gazelle deps only happens once at the start of any `bazel (build|run|test)` command
2.  Bazel caches things so that (1) doesn't happen _every_ time `bazel (build|run|test)` is executed

Won't those two items mean that we won't be calling pip install many times?

----

> bazel downloader / bazel credential helper

I'm not familiar with those yet but I did take a quick glace at the [docs](https://bazel.build/reference/command-line-reference#flag--credential_helper) (severely lacking, IMO haha) and the [design propsal](https://github.com/bazelbuild/proposals/blob/main/designs/2022-06-07-bazel-credential-helpers.md).

One thing we'll want to consider is that the creds used for, say, RBE, won't necessarily be the same as the ones used for a private python package index. And users might have multiple private indexes all with different creds and auth methods.

aignas:
Because pip install is executed once for each dependency (essentially with --no-deps option) in parallel, we would need to somehow create the execution environment with the keyring packages repeatedly. The current inclusion of the existing deps is done in this way to use bazel downloader and not need to redownload the keyring multiple times.

Remember that all of this is happening before build action can be executed and we are bootstraping the project here, so we are with limited options here.

I'd be interested into how the credentials helper could be configured to do this task instead of adding all of the keyring dependencies. As you say, adding deps for each keyring provider is not feasible.

However, maybe there is a way to bootstrap everything in a way that does not too much tech debt and keeps the configurability.

dougthor42:
That sounds good to me.

For now, I have a couple options that get me un-stuck and moving forward:

1. Clone the repo and use [`git_override`](https://bazel.build/versions/6.2.0/rules/lib/globals#git_override) to point to the clone and this branch.
2. Use [`single_version_override`](https://bazel.build/versions/6.2.0/rules/lib/globals#single_version_override) to apply a set of patches to the `rules_python` from the BCR.

Neither of those are ideal because of the issues you mentioned, but it'll let me focus on getting things _working_. Then I can start looking into the bazel downloader and credential helper.

Sounds like we want to close this PR? Or perhaps just let it linger in DRAFT for a while?

aignas:
@dougthor42 , would love to hear if https://github.com/bazelbuild/rules_python/pull/1827 helps with `pip.parse` if you can use the bazel downloader there? I think a general setup could be:
```
# MODULE.bazel contents
pip.parse(
    hub_name = "keyring",
    requirements_lock = "keyring_lock.txt",
)

pip.parse(
    hub_name = "pip",
    requirements_lock = "main_lock.txt",
    experimental_index_url = "my-private-thing",
)

# BUILD.bazel
compile_pip_requirements(
    name = "keyring_deps",
)

compile_pip_requirements(
    name = "private_deps",
    deps (or data) = ["@keyring//keyring"],
)

# .bazelrc
Whatever is needed to use the credentials helper.   
```

dougthor42:
Oh that looks neat. I should be able to give it a go on Monday or Tuesday. Maybe earlier if schedules work out in my favor.

Questions:

1. Would you like me to report progress/errors here, in that PR, or somewhere else?
2. Do we need a manifest file for the `keyring` hub?

aignas:
1. I am interested in your progress and feel free to document issues here. If you see that the `experimental_index_url` does not work like you would expect, then #1827 would be a good place to leave feedback.
2. I was hoping that you can pass `keyring` and related deps via `extra_deps` like you mentioned in https://github.com/bazelbuild/rules_python/pull/1795#issuecomment-1990112798. I am not sure if you have to have a manifest, but I think you could, if you don't want to setup the keyring dependencies manually with `http_archive`.

dougthor42:
## TL;DR:

I was able to run the `keyring_deps.update` and `requirements.update` targets and my `requirements_lock.txt` didn't change :tada:.

However, actually attempting to `bazel build` something resulted in a failure to auth to our private index. I need to investigate this more, perhaps related to the credentials helper.


## The Config

This is a representation of the config I'm using. It's basically the same as what you described, but I had to also include `use_repo(pip, "keyring")`.

<details>

```
# .bazelrc
common --enable_bzlmod

# ~/dthor/dev/rules_python is a checkout of aignas/feat/pip-simpleapi @ 5b6a444
build \
    --override_module=rules_python=/usr/local/google/home/dthor/dev/rules_python \
    --override_module=rules_python_gazelle_plugin=/usr/local/google/home/dthor/dev/rules_python/gazelle
```

```starlark
# MODULE.bazel
...
pip.parse(    # new
    hub_name = "keyring",
    python_version = "3.11",
    requirements_lock = "keyring_lock.txt",
)
use_repo(pip, "keyring")
pip.parse(    # existing
    hub_name = "pypi",
    python_version = "3.11",
    experimental_index_url = "https://[REDACTED]/simple",
)
use_repo(pip, "pypi")
```

```starlark
# BUILD.bazel
...
compile_pip_requirements(    # New
    name = "keyring_deps",
    src = "keyring_deps.txt",
    requirements_txt = "keyring_lock.txt",
)
compile_pip_requirements(    # Existing
    name = "requirements",
    src = "requirements.txt",
    extra_args = [
        # Pull from our package registry
        "--extra-index-url",
        "https://[REDACTED]/simple",
        # Just verbose enough so that we can see pip_tools is actually doing something
        "-v",
    ],
    extra_deps = [    # added
        "@keyring//keyring",
        "@keyring//keyrings_google_artifactregistry_auth",
    ],
)
...
```

```
# keyring_deps.txt
keyring==23.5.0 ; python_full_version >= "3.10.4" and python_version < "3.12"
keyrings-google-artifactregistry-auth==1.0.0 ; python_full_version >= "3.10.4" and python_version < "3.12"
```
</details>


## Running

I had a little hiccup but got it resolved.

<details>

`bazel run //:keyring_deps.update` resulted in a lockfile that looks correct to me:

```
keyring_lock.txt
#
# This file is autogenerated by pip-compile with Python 3.11
# by the following command:
#
#    bazel run //:keyring_deps.update
#
cachetools==5.3.3 \
    --hash=sha256:0abad1021d3f8325b2fc1d2e9c8b9c9d57b04c3932657a72465447332c24d945 \
    --hash=sha256:ba29e2dfa0b8b556606f097407ed1aa62080ee108ab0dc5ec9d6a723a007d105
    # via google-auth
certifi==2024.2.2 \
    --hash=sha256:0569859f95fc761b18b45ef421b1290a0f65f147e92a1e5eb3e635f9a5e4e66f \
    --hash=sha256:dc383c07b76109f368f6106eee2b593b04a011ea4d55f652c6ca24a754d1cdd1
    # via requests
cffi==1.16.0 \
    --hash=sha256:0c9ef6ff37e974b73c25eecc13952c55bceed9112be2d9d938ded8e856138bcc \
...
```

`bazel run //:requirements.update` resulted in a request for auth to our package index:

```
$ bazel run //:requirements.update
INFO: Analyzed target //:requirements.update (96 packages loaded, 7432 targets configured).
INFO: Found 1 target...
Target //:requirements.update up-to-date:
  bazel-bin/requirements.update
INFO: Elapsed time: 1.193s, Critical Path: 0.52s
INFO: 5 processes: 5 internal.
INFO: Build completed successfully, 5 total actions
INFO: Running command line: bazel-bin/requirements.update _main/requirements.txt _main/requirements_lock.txt //:requirements.update '--resolver=backtracking' --allow-unsafe --generate-hashes --extra-index-url https://[REDACTED]/simple -v
Updating requirements_lock.txt
Using indexes:
  https://pypi.org/simple
  https://[REDACTED]/simple
  Looking in indexes: https://pypi.org/simple, https://[REDACTED]/simple

                          ROUND 1                           
User for [REDACTED]: 
```

This is not expected, as the Google artifact registry uses the Google application default credentials for auth.

It looks like `keyring` (and thus `pip_tools`) doesn't recognize that the "Google Auth" provider isn't present. If if did, the 4x verbose logs (`-vvvv` in `compile_pip_requirements`) look like:

```
...
Found index url https://[REDACTED]/simple/
  Keyring provider requested: auto
  Keyring provider set: import
  Getting credentials from keyring for https://[REDACTED]/simple/
  Loading KWallet
  Loading SecretService
  Loading Windows
  Loading chainer
  Loading libsecret
  Loading macOS
  Loading Google Auth
  Checking None for explicit credentials as part of auth process...
  Checking Cloud SDK credentials as part of auth process...
  Making request: POST https://oauth2.googleapis.com/token
...
```

Your code hangs after logging `Loading macOS`.

**Turns out this is because I didn't include the google auth as an extra_dep in `compile_pip_requirements`:**

```diff
     requirements_txt = "requirements_lock.txt",
-    extra_deps = ["@keyring//keyring"],
+    extra_deps = [
+        "@keyring//keyring",
+        "@keyring//keyrings_google_artifactregistry_auth",
+    ],
 )
```

**This diff is already applied to the Config section above.**

</details>


## Trying a Build

`bazel build //...` was a no-go :disappointed:.

<details>

On first pass, I got a request for user creds and an error related to the user prompt:

```
===== stdout start =====                                                                                                                                                                                                                      
Looking in indexes: https://pypi.org/simple, https://[REDACTED]/simple                                                                                                                                              
User for [REDACTED]:                                                                                                                                                                                                             ===== stdout end =====
===== stderr start =====
…
  File "/usr/local/google/home/dthor/.cache/bazel/_bazel_dthor/dbe74c4144b5c9a438d84a119652bef9/external/rules_python~~internal_deps~pypi__pip/pip/_internal/utils/misc.py", line 208, in ask_input
    return input(message)                                  
           ^^^^^^^^^^^^^^                                  
EOFError: EOF when reading a line 
```

When I got the same issue with my hacky branch, it was resolved by adding `pypi__keyrings_google_artifactregistry_auth` to `rules_python/MODULE.bazel`.

I then realized that I needed to add the `experimental_index_url` arg to `pip.parse`... :facepalm: That resulted in auth errors :cry: but it was using the bazel downloader, yay!

```
$ bazel build //src/labrad:all
WARNING: Download from https://[REDACTED]/simple/alabaster/ failed: class com.google.devtools.build.lib.bazel.repository.downloader.UnrecoverableHttpException GET returned 401 Unauthorized
WARNING: Download from https://[REDACTED]/simple/ansi2html/ failed: class com.google.devtools.build.lib.bazel.repository.downloader.UnrecoverableHttpException GET returned 401 Unauthorized
WARNING: Download from https://[REDACTED]/simple/accessible_pygments/ failed: class com.google.devtools.build.lib.bazel.repository.downloader.UnrecoverableHttpException GET returned 401 Unauthorized
WARNING: Download from https://[REDACTED]/simple/altair/ failed: class com.google.devtools.build.lib.bazel.repository.downloader.UnrecoverableHttpException GET returned 401 Unauthorized
WARNING: Download from https://[REDACTED]/simple/absl_py/ failed: class com.google.devtools.build.lib.bazel.repository.downloader.UnrecoverableHttpException GET returned 401 Unauthorized
WARNING: Download from https://[REDACTED]/simple/alembic/ failed: class com.google.devtools.build.lib.bazel.repository.downloader.UnrecoverableHttpException GET returned 401 Unauthorized
        File "/usr/local/google/home/dthor/.cache/bazel/_bazel_dthor/dbe74c4144b5c9a438d84a119652bef9/external/rules_python~/python/private/bzlmod/pip.bzl", line 446, column 30, in _pip_impl
                _create_whl_repos(module_ctx, pip_attr, hub_whl_map, whl_overrides, simpleapi_cache)
        File "/usr/local/google/home/dthor/.cache/bazel/_bazel_dthor/dbe74c4144b5c9a438d84a119652bef9/external/rules_python~/python/private/bzlmod/pip.bzl", line 186, column 40, in _create_whl_repos
                index_urls = simpleapi_download(
        File "/usr/local/google/home/dthor/.cache/bazel/_bazel_dthor/dbe74c4144b5c9a438d84a119652bef9/external/rules_python~/python/private/pypi_index.bzl", line 90, column 51, in simpleapi_download
                contents[pkg] = download.packages.contents()
        File "/usr/local/google/home/dthor/.cache/bazel/_bazel_dthor/dbe74c4144b5c9a438d84a119652bef9/external/rules_python~/python/private/pypi_index.bzl", line 146, column 26, in lambda
                download.wait() if download_kwargs.get("block") == False else download,
Error in wait: java.io.IOException: Error downloading [https://[REDACTED]/simple/absl_py/] to /usr/local/google/home/dthor/.cache/bazel/_bazel_dthor/dbe74c4144b5c9a438d84a119652bef9/modextwd/rules_python~~pip/https___[REDACTED]_simple_absl_py.html: GET returned 401 Unauthorized
WARNING: Download from https://[REDACTED]/simple/aenum/ failed: class com.google.devtools.build.lib.bazel.repository.downloader.UnrecoverableHttpException GET returned 401 Unauthorized
WARNING: Download from https://[REDACTED]/simple/annotated_types/ failed: class com.google.devtools.build.lib.bazel.repository.downloader.UnrecoverableHttpException GET returned 401 Unauthorized
ERROR: Analysis of target '//src/labrad:auth' failed; build aborted: error evaluating module extension pip in @@rules_python~//python/extensions:pip.bzl
INFO: Elapsed time: 1.434s, Critical Path: 0.00s
INFO: 1 process: 1 internal.
FAILED: 
    Fetching module extension pip in @@rules_python~//python/extensions:pip.bzl; starting
```

</details>

dougthor42:
Now that #1827 is basically ready and it's been tested with a private index (https://github.com/bazelbuild/rules_python/pull/1827#issuecomment-2032689270 and https://github.com/bazelbuild/rules_python/pull/1827#issuecomment-2038756206), this PR isn't needed anymore. Closing.

PR #1797, opened 2024-03-12
docs: Update notice about rctx.getenv on Bazel 7

vonschultz:
Bazel 7.1 has been released, and includes support for `repository_ctx.getenv`. Therefore, update the comments and docs to reflect the current status of the feature with released Bazel versions.


vonschultz:
Update of https://github.com/bazelbuild/rules_python/pull/1673/.

PR #1801, opened 2024-03-12
feat: Add support for python-wheel data directory

### LINKED ISSUE #1777: Add support for python-wheel data directory
# 🚀 feature request

### Relevant Rules

`//python:packaging.bzl - py_wheel`

### Description

The `py_wheel` rule should support the `.data` directory as specified in [PEP 427, The .data directory](https://peps.python.org/pep-0427/#the-data-directory).   This is _different_ than including data files in the source tree.  

It seems to me that changes are needed all the way down in `wheelmaker`, with arguments and logic that mirrors the support of the current flag 'extra_distinfo_file'.  This would be exposed in `py_wheel` as perhaps `data_files`, and a similar ability to rename the files.

I haven't been able to think of any alternatives, but would be open to hear about them from others.

---

Finally, let me take a moment to say thank you for a wonderful project.  It is really easy to use and adopt, especially with bazelmod.  

If the solution I've described is on the right track, I can take the first pass at implementing and testing it.

BradHolmes:
Fixes #1777

* Adds `data_files` attribute to `py_wheel` rule. 
* Minimal validation of the data-files target directories per [specification](https://packaging.python.org/en/latest/specifications/binary-distribution-format/#installing-a-wheel-distribution-1-0-py32-none-any-whl)
* Added two tests.  
* Added example

### DIFF
--- a/examples/wheel/BUILD.bazel
+++ b/examples/wheel/BUILD.bazel
@@ -312,6 +312,20 @@ py_wheel(
     deps = [":example_pkg"],
 )
 
+# Package just a specific py_libraries, without their dependencies
+py_wheel(
+    name = "minimal_data_files",
+    testonly = True,  # Set this to verify the generated .dist target doesn't break things
+
+    # Re-using some files already checked into the repo.
+    data_files = {
+        "//examples/wheel:NOTICE": "scripts/NOTICE",
+        "README.md": "data/target/path/README.md",
+    },
+    distribution = "minimal_data_files",
+    version = "0.0.1",
+)
+
 py_test(
     name = "wheel_test",
     srcs = ["wheel_test.py"],
@@ -321,6 +335,7 @@ py_test(
         ":custom_package_root_multi_prefix_reverse_order",
         ":customized",
         ":filename_escaping",
+        ":minimal_data_files",
         ":minimal_with_py_library",
         ":minimal_with_py_library_with_stamp",
         ":minimal_with_py_package",
--- a/python/private/py_wheel.bzl
+++ b/python/private/py_wheel.bzl
@@ -120,6 +120,7 @@ See [`py_wheel_dist`](#py_wheel_dist) for more info.
 
 _feature_flags = {}
 
+ALLOWED_DATA_FILE_PREFIX = ("purelib", "platlib", "headers", "scripts", "data")
 _requirement_attrs = {
     "extra_requires": attr.string_list_dict(
         doc = ("A mapping of [extras](https://peps.python.org/pep-0508/#extras) options to lists of requirements (similar to `requires`). This attribute " +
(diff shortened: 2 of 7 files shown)


tools/wheelmaker.py
```diff
@@ -452,25 +463,25 @@ def parse_args() -> argparse.Namespace:
     return parser.parse_args(sys.argv[1:])
 
 
+def _parse_file_pairs(content: List[str]) -> List[List[str]]:
+    """
+    Parse ; delimited lists of files into a 2D list.
+    """
+    return [i.split(";") for i in content or []]
```
aignas, line 470: Am I correct to assume that we will always split `i` into 2 elements? Maybe we should pass the following:
```suggestion
    return [i.split(";", 1) for i in content or []]
```

aignas (commented):
It would be also great to add an example in `examples/wheel` directory.

I think the implementation is really nice, thank you a lot for submitting this.

examples/wheel/BUILD.bazel
```diff
@@ -321,6 +335,7 @@ py_test(
         ":custom_package_root_multi_prefix_reverse_order",
         ":customized",
         ":filename_escaping",
+        ":minimal_data_files",
```
aignas, line 338: I would expect an addition of a test under `wheel_test.py` which checks the directory structure, similarly to how the existing tests do. Right now we are just building the test wheel but we don't have any test verifying that the data files end up in the right locations.

aignas (commented):
I can also review, stamp and merge this :)

examples/wheel/BUILD.bazel
BradHolmes: Oh, I hadn't followed up on what that py-test rule was doing.  Thanks, and added.

python/private/py_wheel.bzl
```diff
@@ -473,6 +479,28 @@ def _py_wheel_impl(ctx):
...
 
+    for target, filename in ctx.attr.data_files.items():
+        target_files = target.files.to_list()
+        if len(target_files) != 1:
+            fail(
+                "Multi-file target listed in data_files %s",
+                filename,
+            )
+
+        if not filename.startswith(ALLOWED_DATA_FILE_PREFIX):
```
aignas, line 490: We should probably check that / follows straight after the allowed prefix value.

Maybe something like `filename.partition("/")[0] not in ALLOWED_DATA_FILE_PREFIX)` would do the trick?

aignas (commented):
One more comment, but looksgood otherwise

python/private/py_wheel.bzl
BradHolmes: Fixed and added another unit test

aignas (approved):

PR #1803, opened 2024-03-13
 tweak(whl_library): Extend PyWheelInfo provider

mgosk:
Adds `version` and `distribution` fields to `PyWheelInfo` provider.

Bigger projects (like mine) have rules to generate one file with versions of all subsequent components. I would like to expose needed fields in convenient and "bazelish" way.


mgosk:
Looks like version var is not fully resolved at this moment. Closing PR

PR #1806, opened 2024-03-15
fix(internal): add WORKSPACE.bzlmod to ensure dependency separation

aignas:
Before this PR, WORKSPACE file would be also used when using
'rules_python' under bzlmod and when developing locally it would yield
to inconsistent behaviour between the examples and the main workspace.

This PR ensures that the `dev_pip` is used via `bzlmod`. Whilst there
may be a little bit of duplication from now on, it ensures that
non-bzlmod code paths do not leak into the bzlmod dev setup.


rickeylev (approved):

rickeylev:
Oh wow, how did we not notice this sooner? This would explain a few oddities we've seen. Thanks!

PR #1819, opened 2024-03-22
feat(gazelle): Add "python_test_file_pattern" directive

### LINKED ISSUE #1816: Gazelle considers both `test_*.py` and `*_test.py` files to be `py_test` targets.
# 🐞 bug report

### Affected Rule

+ gazelle

### Is this a regression?

Nope! Just unexpected behavior.

### Description

A typical python project will use only **one of** `*_test.py` or `test_*.py` to name test files.

However, gazelle will make `py_test` targets for both types:

https://github.com/bazelbuild/rules_python/blob/e86252ffd6d1a1bf32ae99933acc5ab49b78ec1e/gazelle/python/generate.go#L116

https://github.com/bazelbuild/rules_python/blob/e86252ffd6d1a1bf32ae99933acc5ab49b78ec1e/gazelle/python/generate.go#L198

This can lead to unexpected behavior:

#### Example 1: unit test utilities

Assume that a project uses `*_test.py` for unit/integration tests. Using `pytest`, the config would be:

```toml
[tool.pytest.ini_options]
python_files = ["*_test.py"]
```

Also assume that there needs to be some utilities related to testing. A typical solution would be to make a `test_utils.py` file because that does not get collected by the pytest test runner.

`gazelle` will incorrectly (at least according to the user) make:

```starlark
py_test(
    name = "test_utils",
    srcs = ["test_utils.py"],
)
```

The user expects a `py_library` instead.

#### Example 2: Non-code tests, such as electrical or physical testing

Code that deals with electrical or physical testing (voltage, current, vibration, pressure, etc) is typically called `test_foo` or `foo_test`. For example, code that determines the breakdown voltage of a part might be called `test_breakdown_voltage.py`. Code might also need to create collections or suites of non-code tests, and would live in `test_suites.py`.

Again, according to the user, gazelle will incorrectly make `py_test` targets for such instead of `py_library` targets.

## 🔬 Minimal Reproduction

1. Make `test_foo.py`
2. Make `foo_test.py`
3. Run `gazelle`
4. See that both were generated as `py_test` targets.

## 🔥 Exception or Error

N/A

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
$ lsb_release -a
No LSB modules are available.
Distributor ID: Debian
Description:    Debian GNU/Linux rodete
Release:        n/a
Codename:       rodete
  </code>
</pre>

<pre>
  <code>
$ bazel version
Build label: 7.1.0
Build time: Mon Mar 11 17:55:31 2024 (1710179731)
Build timestamp: 1710179731
Build timestamp as int: 1710179731
  </code>
</pre>

**Rules_python version:**

<pre>
  <code>
0.31.0 @ bdb2aa2e5c0ca435918037aa3568c0b0c4c9c1ab
  </code>
</pre>

**Anything else relevant?**

Don't think so.

dougthor42:
Add the `python_test_file_pattern` directive. This directive allows
users to configure what python files get mapped to the `py_test`
rule.

The default behavior is unchanged: both `test_*` and `*_test.py`
files generate `py_test` targets if the directive is unset.

The directive supports multiple glob patterns, separated by a comma.

Note: The original code used, effectively, `test_*` for one of the
patterns. This code uses `test_*.py` instead. These are equivalent
because of the `.py` extension check prior to pattern matching.

Fixes #1816.


gazelle/python/generate.go
```diff
@@ -54,6 +55,24 @@ func GetActualKindName(kind string, args language.GenerateArgs) string {
...
+}
+
+func matchesAnyGlob(s string, globs []glob.Glob) bool {
+	for _, g := range globs {
+		if g.Match(s) {
+			return true
+		}
+	}
+	return false
+}
```
dougthor42, line 74: Is there a better way to do this in go?

In python, these would be replaced by something like:

```python
compiled_globs = [MustCompile(s) for s in string.split(",")]
...
else if any(glob.match(f) for glob in compiled_globs):
    pyTestFilenames.Add(f)
```

gazelle/python/testdata/directive_python_test_file_pattern/README.md
```diff
@@ -0,0 +1,18 @@
+# Directive: `python_test_file_pattern`
```
dougthor42, line 1: Naming. I also considered these other options:

+ `python_test_file_patterns`  (plural)
+ `python_test_file_glob`
+ `python_test_file_globs`

LMK if you'd prefer one of those or an entirely new one.

gazelle/python/generate.go
```diff
@@ -54,6 +55,24 @@ func GetActualKindName(kind string, args language.GenerateArgs) string {
 	return kind
 }
 
+func makeCompiledGlobs(globs []string) []glob.Glob {
+	compiledGlobs := []glob.Glob{}
+	for _, value := range globs {
+		compiledGlob := glob.MustCompile(value)
```
aignas, line 61: I might prefer a more helpful error message if the glob fails. Could we add an integration test? Maybe it is good enough already? Right now this code would panic

gazelle/python/generate.go
aignas: This is good enough. Slightly verbose, but easy to understand.

gazelle/python/testdata/directive_python_test_file_pattern/README.md
aignas: The current choice is good.

dougthor42 (commented):
I also added a test that subpackages can override the value.

gazelle/python/configure.go
```diff
@@ -181,6 +182,9 @@ func (py *Configurer) Configure(c *config.Config, rel string, f *rule.File) {
 			}
 		case pythonconfig.Visibility:
 			config.AppendVisibility(strings.TrimSpace(d.Value))
+		case pythonconfig.TestFilePattern:
+			globStrings := strings.Split(strings.TrimSpace(d.Value), ",")
+			config.SetTestFilePattern(globStrings)
```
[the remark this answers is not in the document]
dougthor42: > The empty pattern yielding no test targets is not super intuitive, IMHO.

Agreed.

That said, I would argue that the empty pattern resetting to defaults is _also_ not super intuitive. Some readers will see that and think "no test targets" while others will think "resets to default".

I like explicit things 😁. There's no ambiguity and it's easier to grok when returning to something N years down the road.

If we wanted to support yielding no test targets[^1] it should be something like `gazelle:python_test_file_pattern NONE` or have a different directive altogether (`gazelle:python_no_tests` or something).

[^1]: I can't think of a reason why someone would want that, but that doesn't mean there aren't reasons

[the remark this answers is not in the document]
wingsofovnia: Yielding no test targets can (and imho should) be achieved by ignoring files with `gazelle:python_ignore_files`. That said, setting pattern to NONE should not be allowed. `gazelle:python_test_file_pattern ` (blank) should also fail.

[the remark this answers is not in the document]
dougthor42: > yielding no test targets can (and imho should) be achieved by ignoring files with `gazelle:python_ignore_files`

I disagree. There are definitely cases where you want to yield no test targets but do want to yield lib and binary targets.

```
project/
+ BUILD.bazel
+ MODULE.bazel
+ src/myfoo/
    + BUILD.bazel
    + foo.py
    + foo_test.py
    + subdir/
        + BUILD.bazel
        + run_test.py      # py_binary
        + example_test.py  # py_library
```

Using `gazelle:python_ignore_files` would mean that _no targets_ are generated in `subdir` above, when really we'd want only py_library and py_binary targets made.

There's also the case where test files are outside of the package, as described by [PyPA Tutorial](https://packaging.python.org/en/latest/tutorials/packaging-projects/#creating-the-package-files). In that case, you can be sure that all targets in `src/mypackage` below are `py_library` and `py_binary`.

```
project/
+ BUILD.bazel
+ MODULE.bazel
+ src/mypackage/               # guaranteed to have no py_test targets
    + BUILD.bazel
    + run_electrical_test.py   # py_binary
    + electrical_test.py       # py_library
+ tests/                       # any py_* targets
    + BUILD.bazel
    + foo_test.py              # py_test
    + bar_test.py              # py_test
    + test_utils.py            # py_library
    + run_all_test.py          # py_binary
```

> also, it's a bit confusing to set pattern to `NONE`, some might thing there is `NONE` file

A valid point. Perhaps using something that is unlikely to be a filename would be better? Maybe one of these or similar?

```starlark
# gazelle:python_test_file_pattern %NONE%
# gazelle:python_test_file_pattern !!NO_FILES
# gazelle:python_test_file_pattern ~!~NO_PATTERN~!~
```

One issue is that linux filenames can be basically anything haha so it's not feasible to pick a string that _can't_ be a filename.

If we do go with a more "special" string, I'd probably want to update the `python_default_visibility` directive special strings to match.

> `gazelle:python_test_file_pattern` (blank) should also fail.

+1 to that.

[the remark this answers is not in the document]
wingsofovnia: Agreed with arguments around `gazelle:python_ignore_files`. 

How about...not having any magic names at all? I mean, by the time someone wants to reset the `gazelle:python_test_file_pattern`, he/she already learned what default values, and if resetting patterns is needed, it can be done with `gazelle:python_test_file_pattern *_test,test_*`.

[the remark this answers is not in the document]
dougthor42: I'm good with no magic names at all and a "reset" by explicitly setting the value to defaults. `gazelle:python_test_file_pattern *_test.py,test_*.py`

That said, I'd still like to mention the two (minor) issues with no magic values:

1. [can ignore] There's no way to defer to `rules_python`'s defaults. For example, if `rules_python` changes the default from `*_test.py,test_*.py` to something else, users would have to update their configs manually _if they wanted to always use the `rules_python` default_.
    + TBH this is probably a good thing because it results in less unexpected behavior.
2. We're back to the `# gazelle:python_test_file_pattern` (no value) for preventing _any_ files from mapping to `py_test`.
    + As mentioned previously, this is not very intuitive.

For (2), perhaps just making sure that's documented and tested is sufficient? Unless we implement _another_ directive `gazelle:python_no_test_files` or something (which I'm against - it seems to only complicate things).

> have dedicated no-args directive for resetting to default ... it only makes sense if applied conventionally everywhere in the plugin

+1. I'm slightly unsure about having directives to reset to default (as you said, people can just set the default manually), but if we add them they definitely should be applied everywhere.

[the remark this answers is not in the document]
wingsofovnia: As far as I can see, the plugin currently has no switches for enabling/disabling _parts_ of the generation process. Setting _pattern_ to none is basically an abuse of the directive to disable tests and I'd rather have an explicit `python_no_test_files` for that so that there is clear indication no tests are being processed for a) humans, and b) plugin itself, so that it can potentially deviate its processing depending on that fact (e.g. print debug/warning message etc).

This should be, however, out of context for this PR. I'd say, we should just disallow no-arg `python_test_file_pattern`, provide no magic values and wrap-up the PR.

Have I already mentioned this is great work and nice improvement 🙂? Thank you! Looking forward to it. I have exactly the place where I will use it already.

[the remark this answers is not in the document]
dougthor42: > Setting _pattern_ to none sounds to me like an abuse of the directive

I can see that and it's a good point.

> I'd rather have an explicit `python_no_test_files` so that there is a clear indication no tests are being processed ... we should just disallow no-arg `python_test_file_pattern`, provide no magic values and wrap-up the PR.

SGTM. I'd like to continue the discussion, so I've opened #1826.

> this is great work and nice improvement

Thanks! I've been really enjoying being able to add more features.

gazelle/README.md
```diff
@@ -438,7 +504,7 @@ for more information on extending Gazelle.
 
 If you add new Go dependencies to the plugin source code, you need to "tidy" the go.mod file.
 After changing that file, run `go mod tidy` or `bazel run @go_sdk//:bin/go -- mod tidy`
-to update the go.mod and go.sum files. Then run `bazel run //:update_go_deps` to have gazelle
+to update the go.mod and go.sum files. Then run `bazel run //:gazelle_update_repos` to have gazelle
```
dougthor42, line 507: I wasn't able to run `//:update_go_deps` so I changed it to what worked for me. Please LMK if it's incorrect.

Though note that I _didn't actually update all go deps_.

gazelle/go.mod
```diff
@@ -9,6 +9,7 @@ require (
 	github.com/bmatcuk/doublestar v1.3.4
 	github.com/emirpasic/gods v1.18.1
 	github.com/ghodss/yaml v1.0.0
+	github.com/gobwas/glob v0.2.3
```
dougthor42, line 12: This seems to be the most popular `glob` package, though I don't see any github activity in 6 years. Other options were:

+ https://pkg.go.dev/github.com/ryanuber/go-glob
+ https://pkg.go.dev/v.io/v23/glob

dougthor42 (commented):
I've added the test case for "no value causes error" and updated the docs for this directive. I think we're ready for re-review.

gazelle/python/configure.go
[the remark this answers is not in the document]
aignas: All this SGTM, thanks for the discussion. I've been enjoying the read! :)

gazelle/go.mod
wingsofovnia: How about `doublestar.Match`? It is used by `bazel-gazelle` and is already included in this `go.mod:9` too. 

```go
import (
	"fmt"

	"github.com/bmatcuk/doublestar/v4"
)

func matchesAnyGlob(s string, globs []string) bool {
	for _, g := range globs {
		if ok, _ := doublestar.Match(g, s); ok {
			return true
		}
	}
	return false
}

func main() {
	defaultGlob := []string{"*_test.py", "test_*.py"}
	fmt.Println(matchesAnyGlob("test_my_target.py", defaultGlob))
	fmt.Println(matchesAnyGlob("my_target_test.py", defaultGlob))

	customGlob := []string{"*_pytest.py", "pytest_*.py"}
	fmt.Println(matchesAnyGlob("pytest_upload.py", customGlob))
	fmt.Println(matchesAnyGlob("download_pytest.py", customGlob))
}
```

dougthor42 (commented):
removed `gobwas/glob` and migrated/upgraded to doublestar/v4.

wingsofovnia (approved):
Thank you for the great work!

aignas (approved):

gazelle/python/generate.go
```diff
@@ -54,6 +54,19 @@ func GetActualKindName(kind string, args language.GenerateArgs) string {
 	return kind
 }
 
+func matchesAnyGlob(s string, globs []string) bool {
+	for _, g := range globs {
+		ok, err := doublestar.Match(g, s)
```
aignas, line 59: I am slightly concerned that we are compiling the glob here. Would it be possible to compile it elsewhere? When parsing the config would be another option.

aignas (changes requested):

gazelle/python/generate.go
aignas: Given that this is used in larger repos I am worried that we perform a linear number of glob compilations as opposed to doing it once when loading the config.

gazelle/pythonconfig/pythonconfig.go
```diff
@@ -425,3 +435,19 @@ func (c *Config) SetDefaultVisibility(visibility []string) {
 func (c *Config) DefaultVisibilty() []string {
 	return c.defaultVisibility
 }
+
+// SetTestFilePattern sets the file patterns that should be mapped to 'py_test' rules.
+// Will exit 1 if the pattern is invalid.
+func (c *Config) SetTestFilePattern(patterns []string) {
+	for _, p := range patterns {
+		if !doublestar.ValidatePattern(p) {
+			log.Fatalf("ERROR: Failed to compile glob '%v'. Error: syntax error in pattern\n", p)
```
dougthor42, line 444: `globstar` errors with [`path.ErrBadPAttern`](https://pkg.go.dev/path/filepath#pkg-variables) but also states "[users should _not_ count on the returned error, `doublestar.ErrBadPattern`, being equal to `path.ErrBadPattern`](https://pkg.go.dev/github.com/bmatcuk/doublestar/v4#Match)" so I've opted to inject the string manually.

dougthor42: Actually... I could just do:

```go
// We only care if the pattern is valid. Use `Match` instead of `ValidatePattern` (which
// just returns `bool`) so that we always use the error message provided by `doublestar`.
if _, err := doublestar.Match(p, ""); err != nil {
    log.Fatalf("ERROR: Failed to compile glob '%v'. Error: %v\n", p, err)
}
...
```

Any preference between
(a) using `ValidatePattern` and manually injecting the error string as I've already done or
(b) abusing `Match`?

gazelle/pythonconfig/pythonconfig.go
```diff
@@ -425,3 +435,19 @@ func (c *Config) SetDefaultVisibility(visibility []string) {
 func (c *Config) DefaultVisibilty() []string {
 	return c.defaultVisibility
 }
+
+// SetTestFilePattern sets the file patterns that should be mapped to 'py_test' rules.
+// Will exit 1 if the pattern is invalid.
+func (c *Config) SetTestFilePattern(patterns []string) {
+	for _, p := range patterns {
+		if !doublestar.ValidatePattern(p) {
```
dougthor42, line 443: Would this check be better in `configure.go` `case pythonconfig.TestFilePattern:`?

gazelle/python/generate.go
wingsofovnia: Looking at the match's [code](https://github.com/bmatcuk/doublestar/blob/180028ba525d3116cc917ac3179e79428acfd89c/match.go#L74) there is nothing to "compile" in glob really, like we typically do with regex. This repo already does `double start.Match` in the loop [here](https://github.com/bazelbuild/rules_python/blob/b9f39bf09760ff75aedd64ce99c9573c1a97068f/gazelle/python/generate.go#L188) and so does bazel-gazelle [here](https://github.com/bazelbuild/bazel-gazelle/blob/1b331b6d9cb2d981580a8d3e2dfc551185790737/walk/config.go#L145) and it hasn't been an issue so far.

aignas:
Ah, thanks for checking. Then I am happy to merge everything! :)

gazelle/python/configure.go
```diff
@@ -181,6 +182,13 @@ func (py *Configurer) Configure(c *config.Config, rel string, f *rule.File) {
 			}
 		case pythonconfig.Visibility:
 			config.AppendVisibility(strings.TrimSpace(d.Value))
+		case pythonconfig.TestFilePattern:
+			value := strings.TrimSpace(d.Value)
+			if value == "" {
+				log.Fatal("ERROR: Directive 'python_test_file_pattern' requires a value.")
```
wingsofovnia, line 188: nit: there is no need to double severity here as `Fatal` will be already in the log. also, go's error messages are conventionally lowercased since they can be nested (e.g. `something bad happened: directive 'python...`)
```suggestion
				log.Fatal("directive 'python_test_file_pattern' requires a value.")
```

gazelle/pythonconfig/pythonconfig.go
wingsofovnia: I'd say it would be slightly more aligned with existing code to do to in a `case` as there are already some `invalid value` error, parsing etc and these are simple setters.

gazelle/pythonconfig/pythonconfig.go
wingsofovnia: I guess the word `abusing` in your comment says for itself :) Looking at the `Match`'s code, `err.ErrBadPattern` brings no more than just an error flag (no error message, no extra context etc.) so it is effectively does the same as having true/false. I'd keep it as is. Except maybe a nit regarding go error message [style](https://go.dev/wiki/CodeReviewComments#error-strings) (see the comment in `configure.go`).

```suggestion
			log.Fatalf("invalid glob pattern '%s'", p)
```

gazelle/python/configure.go
dougthor42: TIL, thanks! Updated.

As an FYI, the `ERROR:` and `\n` were included to maintain consistency with the other `log.Fatal` messages:

```shell
$ rg 'log.Fatal.?\("'
manifest/generate/generate.go
90:             log.Fatalf("ERROR: %v\n", err)
109:            log.Fatalf("ERROR: %v\n", err)

pythonconfig/pythonconfig.go
444:                    log.Fatalf("ERROR: Failed to compile glob '%v'. Error: syntax error in pattern\n", p)

python/lifecycle.go
41:                     log.Fatalf("failed to write parser zip: %v", err)
47:                     log.Fatalf("cannot write %q: %v", helperPath, err)

python/generate.go
238:                    log.Fatalf("ERROR: %v\n", err)
319:                    log.Fatalf("ERROR: %v\n", err)
353:                    log.Fatalf("ERROR: %v\n", err)
384:                    log.Fatalf("ERROR: %v\n", err)
504:            log.Fatalf("ERROR: %v\n", err)

python/configure.go
188:                            log.Fatal("ERROR: Directive 'python_test_file_pattern' requires a value.")
```

Prior to this PR, 7 of 9 calls included both `ERROR:` and `\n`.

gazelle/pythonconfig/pythonconfig.go
dougthor42: 👍  moved.

gazelle/pythonconfig/pythonconfig.go
dougthor42: 👍  kept, aside from style nit

aignas (approved):

PR #1822, opened 2024-03-23
fix(whl_library): only add group machinery when it is needed

### LINKED ISSUE #1760: cycle support which replaces `:pkg` with `:_pkg`breaks aspects
# 🐞 bug report

### Affected Rule

pip_parse/pip_repository

### Is this a regression?

Yes this is a regression.

### Description

The change in this PR: https://github.com/bazelbuild/rules_python/pull/1166/files 

specifically unconditionally introducing an impl label `_pkg` https://github.com/bazelbuild/rules_python/pull/1166/files#diff-ca70482c01526a5439dab7ac2cb14e637684e50daca723d8e6b4e58ca45cbc67R20

that is aliased back to `pkg` breaks aspects. In an aspect, you will get the target of the alias, which will be `_pkg`. So, with this change, we need to realias back to the public alias.

This causes breakages for anyone using aspects.

## 🔬 Minimal Reproduction

You can clearly see the generated build files now make the real `:_pkg` target private.

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
macos and linux
  </code>
</pre>

<pre>
  <code>
Bazelisk version: v1.16.0
Build label: 6.3.2
Build target: bazel-out/darwin_arm64-opt/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Tue Aug 8 15:51:44 2023 (1691509904)
Build timestamp: 1691509904
Build timestamp as int: 1691509904
  </code>
</pre>

**Rules_python version:**

<pre>
  <code>
0.31.0
  </code>
</pre>

aignas:
This removes the `alias` additions when the package groups are not used
fixing alias that attempt to traverse the entire py_library dependency
tree.

I noticed that for some code that I want to write for `whl_library`
having it explicit whether we are using groups or not in the `pip.parse`
and or `whl_library` makes the code easier to understand and write.

Fixes #1760

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,9 @@ A brief description of the categories of changes:
 * (gazelle) In `project` or `package` generation modes, do not generate `py_test`
   rules when there are no test files and do not set `main = "__test__.py"` when
   that file doesn't exist.
+* (whl_library) The group redirection is only added when the package is part of
+  the group potentially fixing aspects that want to traverse a `py_library` graph.
+  Fixes [#1760](https://github.com/bazelbuild/rules_python/issues/1760).
 
 ### Added
 
--- a/python/pip_install/private/generate_whl_library_build_bazel.bzl
+++ b/python/pip_install/private/generate_whl_library_build_bazel.bzl
@@ -63,14 +63,14 @@ filegroup(
 )
 
 filegroup(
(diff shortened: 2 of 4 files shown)


arrdem (approved):
lgtm

rickeylev (approved):

Issue #1791, opened 2024-03-07
Ability to define your own requirements lock file or python version for third party pip dependencies


finn-ball:
# 🐞 bug report

### Affected Rule
```
use_extension("@rules_python//python/extensions:pip.bzl", "pip")
```

### Is this a regression?

Between old `WORKSPACE` rules and `bzlmod`, potentially yes.

### Description
A third party dependency called foo uses rules_python and bzlmod to bring in pip dependencies:
```
pip = use_extension("@rules_python//python/extensions:pip.bzl", "pip")
pip.parse(
    hub_name = "foo_pip_deps",
    python_version = "3.11",
    requirements_lock = "@foo//:requirements_lock.txt",
)
use_repo(pip, "foo_pip_deps")
```
This dependency is implicitly used as part of a ruleset this third party defines.

If I try to redefine this to use a different python_version and my own requirements_lock file in my bar project:
```
pip = use_extension("@rules_python//python/extensions:pip.bzl", "pip")
pip.parse(
    hub_name = "foo_pip_deps",
    python_version = "3.9",
    requirements_lock = "@bar//:requirements_lock.txt",
)
use_repo(pip, "foo_pip_deps")
```
Then I get the error:
```
Duplicate cross-module pip hub named `foo_pip_deps`: pip hub names must be unique across modules. First defined by module `bar`, second attempted by module `foo`
```

In order to define your own `requirements_lock` or `python_version` in this scenario, you'd have to patch the third party dependency. This shouldn't be the case.

## 🔬 Minimal Reproduction

Depend on any third party rule which uses pip and try to update your own MODULE file following the above.

## 🔥 Exception or Error

<pre><code>

</code></pre>

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
linux
  </code>
</pre>

**Output of `bazel version`:**

<pre>
  <code>
7.0.2
  </code>
</pre>

**Rules_python version:**

<pre>
  <code>
0.31.0
  </code>
</pre>

**Anything else relevant?**

hofbi:
Please keep this open. The discussion is taking place in the PR providing the solution: https://github.com/bazelbuild/rules_python/pull/1829

PR #1829, opened 2024-03-31
feat: add ability to override hub name

finn-ball:
Allows the user to override pip dependencies:

https://github.com/bazelbuild/rules_python/issues/1791

This primarily helps solve the issue whereby if your python project `A` depends on python project `B` and python project `C` such that `A -> B, A -> C`, we can now create pip files which resolve for all three repositories. Currently the dependencies in either `B` or `C` will override pip dependencies in `A`.


python/private/bzlmod/pip.bzl
```diff
@@ -421,6 +427,12 @@ A dict of labels to wheel names that is typically generated by the whl_modificat
 The labels are JSON config files describing the modifications.
 """,
         ),
+    "override": attr.bool(
```
aignas, line 430: I personally think this is OK since this works similarly to how archive overrides work, but I would like @rickeylev and @groodt to also weigh in here.

This is an obvious case where we may want to have an escape hatch and because it is a root module, we can do this.

However, I am wondering about what happens if a module like this gets published.

hofbi:
I think I ran exactly into this issue when I want to consume [rules_ros](https://github.com/mvukov/rules_ros) which defines [this hub](https://github.com/mvukov/rules_ros/blob/049418ec41397080ec2fdff1e76eb7e95d9a0d57/MODULE.bazel#L65). But I want to override the pip deps with the onces we define to have a single version used for all dependencies.

aignas:
Thanks for the message. Could you explain, what is the reason for using a particular version of the python dependencies for rules_ros? Is it security, other type of compliance, or just need to fix bad behaviour in rules_ros?

The bzlmod APIs may still need a little bit of refinement before releasing 1.0 and I would like to understand if the feature you'd like is just a workaround a different design flaw that you are experiencing. In order to ensure that we get the right primitives in place I would like to analyse each request separately.

hofbi:
The reason for overriding the versions of the Python dependencies is that we want to have the same version for all targets. If we don't do that, we could end up in a target that uses the same python package at 2 different versions

hofbi:
@aignas Is there any information that we can provide to help you clarifying this?

hofbi:
So that we can follow https://opensource.google/documentation/reference/thirdparty/oneversion

finn-ball:
It would match what WORKSPACE files do in that if you defined the python libraries before the third party dependency, you can override your thirdparty dependency's python libraries.

aignas:
Coming back to this after thinking about how the `override` API is going to
work for the `python` bzlmod extension (see PRs attached to #2081 if anyone
wants to follow along).

I am going to write some thoughts here to explore the problem a little bit more
so that I am sure that I am understanding the implications of the change
correctly. The alternative semantics of the APIs that we can implement here
are either merging overrides or complete override and ignoring hub repo defs
from non-root modules.

Merging the overrides with the initial definition. Since the initial definition
will be done in a non-root module, this makes the implementation difficult -
`module_ctx.modules[0]` will always be the root module and then what definition
takes precedence may be difficult to debug. We could definitely do this, but we
would probably need to have a separate tag class here.

Supporting only overriding - the root module user will have to specify all of
the parameters them selves. Since the root module comes first in the processing
order, we'll just have to silence the error that would be thrown. What we need
to do is to store somewhere that the `hub_repo` has been defined by the root
module and continue.

I am not fully sure if the current implementation does not have a bug around
this, so it would be great to have added unit tests, similar to how #2204 has
added them. I think that way we can correctly identify edge cases in the
overrides and ensure that we are handling correctly them. If I am reading the
code correctly, the current implementation might work only if the
`pip.parse(override = True)` is done with a `python_version` that is not the
same as the one from the non-root module.

What is more, I don't think that storing a single `pip_attr` and a list of
`pip_attr.python_version` is sound - the separate invocations in general may
have different attribute values and we should store something like:
```starlark
    python_attrs = {python_attr.python_version: pip_attr}
```

To sum up, I think I am happy to merge an implementation that has:
* Unit tests similar to how #2204 does them - maybe extract a function that
  returns a `pip_hub_map` and then unit test that. We should definitely add a
  test ensuring that overriding for multiple python versions works.
* Ignores non-root module definitions of hub repos with the same name. The root
  module will be responsible for all of the overrides.
* Has extra docs on overrides explaining the API semantics.

aignas (changes requested):
Needs further changes

hofbi:
@finn-ball Do you plan to proceed here at some point? I think that would be a useful feature to have.

finn-ball:
Sure, I can take a look again at this again next week!

aignas:
With #3837, I believe this is no longer needed, but if you disagree, please reopen the PR.

PR #1831, opened 2024-04-03
feat: Allow additional tools for sphinx_docs

castler:
Some Sphinx plugins require that other (not necessarily Python) tools are available.

One example is the plugin https://github.com/basejumpa/sphinxcontrib-umlet, which requires that UMLet is somehow within the sandbox.

* Adds `tools` arg to `sphinx_docs` to allow passing in arbitrary tools
  that are made available at runtime
* Performs location expansion on `extra_opts`, which allows passing the
  location of the tools onto sphinx.

### DIFF
--- a/sphinxdocs/private/sphinx.bzl
+++ b/sphinxdocs/private/sphinx.bzl
@@ -55,6 +55,7 @@ def sphinx_docs(
         formats,
         strip_prefix = "",
         extra_opts = [],
+        tools = [],
         **kwargs):
     """Generate docs using Sphinx.
 
@@ -88,6 +89,11 @@ def sphinx_docs(
             source files. e.g., given `//docs:foo.md`, stripping `docs/`
             makes Sphinx see `foo.md` in its generated source directory.
         extra_opts: (list[str]) Additional options to pass onto Sphinx building.
+            On each provided option, a location expansion is performed.
+            See `ctx.expand_location()`.
+        tools: (list[label]) Additional tools that are used by Sphinx and its plugins.
+            This just makes the tools available during Sphinx execution. To locate
+            them, use `extra_opts` and `$(location)`.
         **kwargs: (dict) Common attributes to pass onto rules.
     """
     add_tag(kwargs, "@rules_python//sphinxdocs:sphinx_docs")
@@ -102,6 +108,7 @@ def sphinx_docs(
         formats = formats,
         strip_prefix = strip_prefix,
         extra_opts = extra_opts,
+        tools = tools,
         **kwargs
     )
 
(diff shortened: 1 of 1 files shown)


aignas:
Since sphinx is enabled in the next release Ithink this does not need an extrachngalog entry, but the docstring for the tools could include a short example with a particular plugin. Why the opts need expansion is still a little unclear to me. Is it that some plugins may expect the tools to be found in the PATH and that case would not be covered. To be honest, I am not super familiar with sphinx and these questions could be just my ignorance.

castler:
> . Why the opts need expansion is still a little unclear to me. Is it that some plugins may expect the tools to be found in the PATH and that case would not be covered.

Sure, something like this?

```
sphinx_docs(
    ...
    name = ...,
    srcs = ...,
    extra_opts = [
        "-Dsome_binary=$(execpath //:some_binary_target)",
    ],
    tools = [
        "//:some_binary_target",
    ],
)
```

sphinxdocs/private/sphinx.bzl
```diff
@@ -238,7 +245,8 @@ def _run_sphinx(ctx, format, source_path, inputs, output_prefix):
     args.add("-j", "auto")  # Build in parallel, if possible
     args.add("-E")  # Don't try to use cache files. Bazel can't make use of them.
     args.add("-a")  # Write all files; don't try to detect "changed" files
-    args.add_all(ctx.attr.extra_opts)
+    for opt in ctx.attr.extra_opts:
+        args.add(ctx.expand_location(opt))
```
rickeylev, line 249: Please update the doc to note that extra_opts performs location expansion

sphinxdocs/private/sphinx.bzl
```diff
@@ -88,6 +89,7 @@ def sphinx_docs(
             source files. e.g., given `//docs:foo.md`, stripping `docs/`
             makes Sphinx see `foo.md` in its generated source directory.
         extra_opts: (list[str]) Additional options to pass onto Sphinx building.
+        tools: (list[label]) Additional tools that are used by Sphinx and its plugins.
```
rickeylev, line 92:
```suggestion
        tools: (list[label]) Additional tools that are used by Sphinx and its plugins.
            This just makes the tools available during Sphinx execution. To locate
            them, use `extra_opts` and `$(location)` expansion
```

rickeylev (commented):
Mostly lgtm, just need the extra_opts doc updated (I couldn't suggest an edit since it wasn't a modified line)

castler:
> Mostly lgtm, just need the extra_opts doc updated (I couldn't suggest an edit since it wasn't a modified line)

Thanks, applied requested changes.

rickeylev (approved):

PR #1833, opened 2024-04-05
fix: Upgrade protobuf from 21.7 to 23.1/24.4

comius:
The older protobuf version (21.7) is still using legacy struct providers. If used rules_python will fail after the default of --incompatible_disallow_struct_provider_syntax is changed.

Only 23.1 is on BCR (that's why using this version in MODULE.bazel).
Unpatched 23.1 fails with `no such attribute 'exec_tools' in 'genrule' rule`.
Protobuf v24.4 and 25.3. fails on Windows.
That's why using 26.1

Issue: https://github.com/bazelbuild/bazel/issues/19467

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -402,11 +402,12 @@ tasks:
     working_directory: examples/py_proto_library
     platform: macos
   integration_test_py_proto_library_windows_workspace:
-    <<: *reusable_build_test_all
     <<: *common_workspace_flags
     name: "examples/py_proto_library: Windows, workspace"
     working_directory: examples/py_proto_library
     platform: windows
+    build_targets: ["//..."] # Building everything triggers problems in protobuf
+    test_targets: ["//..."]
 
   integration_test_pip_repository_annotations_ubuntu_workspace:
     <<: *reusable_build_test_all
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -10,7 +10,7 @@ bazel_dep(name = "platforms", version = "0.0.4")
 
 # Those are loaded only when using py_proto_library
 bazel_dep(name = "rules_proto", version = "5.3.0-21.7")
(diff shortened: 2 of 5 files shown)


comius:
cc @haberman, @mkruskal-google, on Windows platform protoc fails to build. I tried versions 24.4, 25.3, 26.1. 

Bazel version 7.1.1 without bzlmod

(version 23.1 with bzlmod works, but only because of a patch in https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/protobuf/23.1/patches/0007-bazel-Get-rid-of-exec_tools.-13401.patch 

failures are:
```
(10:32:00) ERROR: C:/b/y5tfdgmr/external/com_google_protobuf/src/google/protobuf/compiler/BUILD.bazel:97:11: Compiling src/google/protobuf/compiler/command_line_interface.cc [for tool] failed: (Exit 2): cl.exe failed: error executing CppCompile command (from target @@com_google_protobuf//src/google/protobuf/compiler:command_line_interface)

external/com_google_protobuf/src/google/protobuf/compiler/command_line_interface.cc(12): fatal error C1083: Cannot open include file: 'google/protobuf/compiler/command_line_interface.h': No such file or directory

(10:32:00) ERROR: C:/b/y5tfdgmr/external/com_google_protobuf/src/google/protobuf/compiler/csharp/BUILD.bazel:22:11: Compiling src/google/protobuf/compiler/csharp/csharp_source_generator_base.cc [for tool] failed: (Exit 2): cl.exe failed: error executing CppCompile command (from target @@com_google_protobuf//src/google/protobuf/compiler/csharp:csharp)

external/com_google_protobuf/src/google/protobuf/compiler/csharp/csharp_source_generator_base.cc(8): fatal error C1083: Cannot open include file: 'google/protobuf/compiler/csharp/csharp_source_generator_base.h': No such file or directory
```

PR #1835, opened 2024-04-06
fix(gazelle): Consistent substitution pattern for python_default_visibility directive

dougthor42:
Make the substitution pattern for `python_default_visibility` consistent
with the existing `python_*_naming_convention` pattern.

In #1787 I added the `python_default_visibility` directive and used a
substitution pattern `$python_root`. However, I missed that the existing
`python_*_naming_convention` directives include a trailing `$`.

This PR is just:
```
rg -l -F "\$python_root" | xargs sed -i 's/\$python_root/$python_root$/g'
```


dougthor42:
Notes:
1. We should get this in before 0.32.0 release so that there doesn't have to be any deprecation timeline.
2. I figured this didn't need a changelog entry because it's just part of adding the `python_default_visibility` directive.

aignas (approved):

PR #1852, opened 2024-04-14
deps: load cc symbols from @rules_cc

martis42:
Adds dependency on rules_cc 0.0.9 for both bzlmod and workspace

This elevates rules_cc from a dev dependency to a full dependency of rules_python.

Work towards incompatible_stop_exporting_language_modules, see https://github.com/bazelbuild/bazel/issues/19455

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,6 +43,9 @@ user.bazelrc
 *.swp
 *.swo
 
+# CLion
+.clwb
+
 # Python cache
 **/__pycache__/
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,6 +73,7 @@ A brief description of the categories of changes:
   the whl and sdist files will be written to the lock file. Controlling whether
   the downloading of metadata is done in parallel can be done using
   `parallel_download` attribute.
+* (deps): `rules_python` depends now on `rules_cc` 0.0.9
 
 [0.XX.0]: https://github.com/bazelbuild/rules_python/releases/tag/0.XX.0
 [python_default_visibility]: gazelle/README.md#directive-python_default_visibility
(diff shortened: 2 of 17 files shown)


martis42:
I am a bit lost about the broken CI. I suspect that somehow in the workspace setup a too old rules_cc  version is being used. Given `python/repositories.bzl` used `maybe` to fetch dependencies, this seems possible. But I am lost why only 2 workspace tests red, although there are many more using the wokspace setup.

aignas:
Note, that the failing builds are using bazel `6.4.0`. If I remember correctly, they are the only two workflows doing that. Could this be the reason?

martis42:
@aignas 
Some experimentation showed the problem is independent from the Bazel version.
I can't explain what happens exactly, but making the build green requires registering `rules_cc` before [this line](https://github.com/bazelbuild/rules_python/blob/main/WORKSPACE#L36) in the WORKSPACE file.
To achieve it I added `rules_cc` to the internal deps, which are loaded before this line. This causes `rules_cc` to be loaded in 2 places (internal and external deps), but given the same is already happening for `bazel_skylib` I assume this workaround is fine.

rickeylev:
Thanks for this! It's been on my cleanup list for a long time.

Since this introduces a new dependency (it's not conceptually new, but will invoke the various remote-repo subsystems, so counts as new), please mention it in the changelog for completeness. I think rules_cc is largely just thin wrappers around the builtin-bazel rules still?

Otherwise LGTM

martis42:
Yes, rules_cc are still mostly thin wrappers https://github.com/bazelbuild/rules_cc/blob/main/cc/defs.bzl#L59

rickeylev (approved):

PR #1855, opened 2024-04-17
refactor: use bazel-skylib for creating hub repo aliases

aignas:
With this PR the code becomes more maintainable and easier to inspect.
Since bazel-skylib is already a dependency of rules_python, this is
a backwards compatible change.

Skipping the CHANGELOG notes because it should not be an externally
visible change.

Work towards #735.


python/private/render_pkg_aliases.bzl
```diff
@@ -56,27 +57,27 @@ def _render_whl_library_alias(
...
             no_match_error = None
 
-    if default:
-        selects["//conditions:default"] = default
-
     return render.alias(
         name = name,
         actual = render.select(
-            selects,
+            {tuple(sorted(v, key = lambda x: ("is_python" not in x, x))): k for k, v in sorted(selects.items())},
```
philsc, line 70: Adding a comment here would be super helpful. It takes a bit of time to understand what this line done.

python/private/render_pkg_aliases.bzl
```diff
@@ -56,27 +57,27 @@ def _render_whl_library_alias(
...
             no_match_error = no_match_error,
+            key_repr = lambda x: repr(x[0]) if len(x) == 1 else render.tuple(x),
```
philsc, line 72: I'm assuming you're doing this to support the `selects.with_or` construct from skylib?

EDIT: Okay, pretty clear from looking at the rest of the changes. Might be worth mentioning as a comment here maybe?

philsc (approved):
Looks great!

rickeylev (approved):

Issue #1663, opened 2024-01-02
Promoting `experimental_requirement_cycles` from experimental


Philip-Murzynowski:
# 🚀 feature request

### Relevant Rules

I have been making use of `experimental_requirement_cycles` and greatly benefiting from it. I wanted to ask if there is a release plan or anything I could look into to help incorporate it into a release.

### Description

 

The experimental feature is introduced [here](https://github.com/bazelbuild/rules_python/blob/4c2d7d9d6608795522322a9b2294b5053e41b979/CHANGELOG.md?plain=1#L31) in the CHANGELOG.

### Describe the solution you'd like

 As this is my first activity in this repository, please let me know if you would like more information.

arrdem:
As the primary author of that feature I'm 50/50 on promoting it to stable. It's a fine solution for making true dependency cycles work (eg. `a <-> b` in all cases of extras requirements). I'm less convinced that it's a good solution for making requirements using extras work since we have to convert what should be optional dependencies into strong dependencies.

Consider `a[]` requiring `b` with `b` requiring `a`, and `a[c]` requiring `c` with `c` requiring `a`. `py_requirement("a")` is now ambiguous. Does the user intend `a[c]`? `a[]`? Do we intend the `requirements.txt` locks to function as virtualenvs from which everything* is installed or as a dependency solution set from which only the needed requirements should be installed?

Personally I prefer the latter. I think having to merge all the Airflow optional dependencies into a single requirement group isn't a great solution because it means that any artifact which requires Airflow takes a dependency on the entire group of requirements. It would be better if say the mysql extra wasn't always implied because `py_requirement("airflow")` now means the entire group with all extras.

There's plenty of room to critique setuptools/pip here for extras really behaving as just sneaky transitive deps not proper package configuration https://discuss.python.org/t/provide-information-on-installed-extras-in-package-metadata/15385.

There's the related problem of needing to merge packaging cycles when there are multiple, again see the newer Airflow packaging. Is that really the right behavior? Or could we do something clever to break these cycles into separate components which would behave more reasonably as optional extras?

aignas:
Some thoughts about a better solution here: https://github.com/bazelbuild/rules_python/pull/1728#discussion_r1660617136

Not sure yet how to integrate the current `experimental_index_url` with the code snippet in there, but it would be an alternative that would not require the users to specify the `experimental_requirement_cycles` at all and it would be my preferred solution.

gholms:
[...]

PR #1856, opened 2024-04-17
feat(pip_parse): support referencing dependencies to packages via hub

aignas:
With this change we can in theory have multi-platform libraries in the
dependency cycle and use the pip hub repo for the dependencies. With
this we can also make the contents of `whl_library` not depend on what
platform the actual dependencies are. This allows us to support the
following topologies:

* A platform-specific wheel depends on cross-platform wheel.
* A cross-platform wheel depends on cross-platform wheel.
* A whl_library can have `select` dependencies based on the interpreter
  version, e.g. pull in a `tomli` dependency only when the Python
  interpreter is less than 3.11.

Relates to #1663.
Work towards #735.


arrdem:
Hmmm gave this a scan and don't have any immediate complaints. While any-any wheels are absolutely my majority use case, I'm a bit concerned about how to model their participation in cycles with platform-dependent artifacts but you've already got that called out as TBD so seems fine.

philsc (approved):
Looks reasonable. I am admittedly not at all familiar with the existing group_library code, which most of the PR seems concerned with.

rickeylev (approved):

PR #1863, opened 2024-04-23
feat(gazelle): Add "include_dep" Python comment annotation

### LINKED ISSUE #1862: Add a new python comment annotation that adds targets to `deps`
# 🚀 feature request

### Relevant Rules

+ gazelle

### Description

Some Python modules may have "hidden" imports. One prime example of this is [`pyvisa`](https://pyvisa.readthedocs.io/en/latest/) when using the [`pyvisa-py` backend](https://pyvisa.readthedocs.io/projects/pyvisa-py/en/latest/index.html). A basic Python file that uses such looks like:

```python
# foo.py
import pyvisa

if __name__ == "__main__":
    rm = pyvisa.ResourceManager('@py')
    print(rm.list_resources())  # ('USB0::0x1AB1::0x0588::DS1K00005888::INSTR')
```

Note that `import pyvisa_py` is not present in the file.

When Gazelle runs on this file, it generates:

```starlark
py_binary(
    name = "foo",
    srcs = ["foo.py"],
    deps = ["@pypi//pyvisa"],
)
```

However, this fails to run with `ModuleNotFoundError: No module named 'pyvisa_py'`.

Add a new Python annotation `# gazelle:include_dep` that allows users to add arbitrary targets to generated `deps`:

```python
# foo.py
import pyvisa

# gazelle:include_dep @pypi//pyvisa_py

if __name__ == "__main__":
    rm = pyvisa.ResourceManager('@py')
    print(rm.list_resources())  # ('USB0::0x1AB1::0x0588::DS1K00005888::INSTR')
```

```starlark
py_binary(
    name = "foo",
    srcs = ["foo.py"],
    deps = [
        "@pypi//pyvisa",
        "@pypi//pyvisa_py",
    ],
)
```

A corollary would be to also add a `# gazelle:include_data` annotation.

The current workaround is to import the module and then just not use it:

```python
# foo.py
import pyvisa
import pyvisa_py

del pyvisa_py  # `del` it to satisfy linters.

if __name__ == "__main__":
    rm = pyvisa.ResourceManager('@py')
    print(rm.list_resources())  # ('USB0::0x1AB1::0x0588::DS1K00005888::INSTR')
```

However, this doesn't allow users to add dependencies on Python targets that happen to be (a) outside of the `python_root` and (b) auto-run during some event[^1].

[^1]: One prime example is pytest's `conftest.py`. Gazelle already has some code that automatically adds this, but Gazelle won't add it if `conftest.py` is outside of of the `python_root` dir. Perhaps I should make a separate Issue for such...

dougthor42:
Add a new Python comment annotation for Gazelle: `include_dep`. This
annotation accepts a comma-separated string of values. Values _should_
be targets names, but no validation is done.

The annotation can be added multiple times, and all values are combined
and de-duplicated.

For `python_generation_mode = "package"`, the `include_dep` annotations
found across all files included in the generated target.

The `parser.annotations` struct is updated to include a new `includeDep`
field, and `parser.parse` is updated to return the `annotations` struct. All
target builders then add the resolved dependencies.

Fixes #1862.

Example:

```python
# gazelle:include_dep //foo:bar,:hello_world,//:abc
# gazelle:include_dep //:def,//foo:bar
```

will cause gazelle to generate:

```starlark
deps = [
    ":hello_world",
    "//:abc",
    "//:def",
    "//foo:bar",
]
```


dougthor42:
Still TODO:

+ [ ] Udpate CHANGELOG
+ [ ] Update documentation

gazelle/python/parser.go
```diff
@@ -164,9 +165,32 @@ func (p *python3Parser) parse(pyFilenames *treeset.Set) (*treeset.Set, map[strin
...
+func removeDupesFromStringTreeSetSlice(array []string) []string {
+	s := treeset.NewWith(godsutils.StringComparator)
+	for _, v := range array {
+		s.Add(v)
+	}
+	dedupe := make([]string, s.Size())
+	for i, v := range s.Values() {
+		dedupe[i] = fmt.Sprint(v)
+	}
+	return dedupe
```
dougthor42, line 193: Coming from python, this function feels icky to me. I'd really like to be able to just do

```python
allAnnotations.includeDep = list(set(allAnnotations.includeDep))
```

above but I couldn't figure out how to do so. The issue is related to types: the `treeset.Values()` is type `[]interface{}`, `treeset.NewWith`'s `values` arg is `interface{}`, but `includeDep` is `[]string`. From what I could find, there's no easy way to "cast" `[]string` to `[]interface{}` and back.

If you have any suggestions on how to clean things up, I'd be happy to hear them. Otherwise no reply is needed.

gazelle/python/parser.go
```diff
@@ -164,9 +165,32 @@ func (p *python3Parser) parse(pyFilenames *treeset.Set) (*treeset.Set, map[strin
...
+
+		// Collect all annotations from each file into a single annotations struct.
+		for k, v := range annotations.ignore {
+			allAnnotations.ignore[k] = v
+		}
+		allAnnotations.includeDep = append(allAnnotations.includeDep, annotations.includeDep...)
 	}
 
-	return modules, mainModules, nil
+	allAnnotations.includeDep = removeDupesFromStringTreeSetSlice(allAnnotations.includeDep)
```
dougthor42, line 176: Do we even want to remove duplicates here? We could just rely on the target builders to handle that.

gazelle/python/parser.go
```diff
@@ -164,9 +165,32 @@ func (p *python3Parser) parse(pyFilenames *treeset.Set) (*treeset.Set, map[strin
 				mainModules[res.FileName].Add(m)
 			}
 		}
+
+		// Collect all annotations from each file into a single annotations struct.
+		for k, v := range annotations.ignore {
+			allAnnotations.ignore[k] = v
+		}
```
dougthor42, line 172: `annotations.ignore` is not used in `generate.go` so we could leave this out if wanted. Thoughts? Leaving it out would mean that the returned `allAnnotations` is not representative of everything that was parsed.

gazelle/python/parser.go
```diff
@@ -247,12 +272,15 @@ type annotation struct {
 type annotations struct {
 	// The parsed modules to be ignored by Gazelle.
 	ignore map[string]struct{}
+	// Labels that Gazelle should include as deps of the generated target.
+	includeDep []string
```
dougthor42, line 276: Should this be plural? Same elsewhere.

I originally went singular to be consistent with `annotations.ignore`.

dougthor42:
I think this is ready for review.

Note that this might be a bit of a footgun. Users can add non-python targets to `deps` and then things will fail during `bazel build` with `<target> does not have mandatory providers: 'PyInfo' or 'CcInfo' or 'PyInfo'.` The docs I've added try to address this.

If that's deemed too big of risk, I'd be happy to discuss alternative solutions.

aignas:
Don't have time to look into this right now, but after reading your last comment, I am wondering if the right way to avoid the footgun would be to use the import path as an argument to the gazelle comment, but that is almost as a noop import statement using just python? Which puts us back to square 1.

I'll think about this a little more.

gazelle/python/parser.go
f0rmiga: I'd make it plural because it is a list. I don't see a need to make `ignore` plural, though.

gazelle/python/parser.go
f0rmiga: I've been meaning to pick up https://github.com/f0rmiga/datanalgo and add the missing data structures using generics. I can try to resurrect that, which would be ideal to use with the Gazelle extension.

gazelle/README.md
```diff
@@ -425,6 +425,108 @@ py_library(
...
+```
+
+The annotations are:
+
+| **Annotation**                                                | **Default value** |
+|---------------------------------------------------------------|-------------------|
+| [`# gazelle:ignore imports`](#annotation-ignore)              | N/A               |
+| Tell Gazelle to ignore import statements. `imports` is a comma-separated list of imports to ignore. | |
+| [`# gazelle:include_dep targets`](#annotation-include_dep)    | N/A               |
+| Tell Gazelle to include  import statements. `targets` is a comma-separated list of target names to include as dependencies. | |
```
f0rmiga, line 456:
```suggestion
| Tells Gazelle to include a set of dependencies, even if they are not imported in a Python module. `targets` is a comma-separated list of target names to include as dependencies. | |
```

gazelle/README.md
```diff
@@ -425,6 +425,108 @@ py_library(
...
+def bar():  # gazelle:annotation_name value
+    pass
+```
+
+The annotations are:
+
+| **Annotation**                                                | **Default value** |
+|---------------------------------------------------------------|-------------------|
+| [`# gazelle:ignore imports`](#annotation-ignore)              | N/A               |
+| Tell Gazelle to ignore import statements. `imports` is a comma-separated list of imports to ignore. | |
```
f0rmiga, line 454:
```suggestion
| Tells Gazelle to ignore import statements. `imports` is a comma-separated list of imports to ignore. | |
```

gazelle/README.md
```diff
@@ -425,6 +425,108 @@ py_library(
...
+This annotation accepts a comma-separated string of values. Values _must_
+be Python targets, but _no validation is done_. If a value is not a Python
+target, building will result in an error saying:
+
+```
+<target> does not have mandatory providers: 'PyInfo' or 'CcInfo' or 'PyInfo'.
+```
+
+Adding non-Python targets to the generated target is a feature request being
+tracked in [Issue #1865](https://github.com/bazelbuild/rules_python/issues/1865).
```
f0rmiga, line 501: This is all fine, we can't save the world.

gazelle/python/testdata/annotation_include_dep/test.yaml
```diff
@@ -0,0 +1,17 @@
+# Copyright 2023 The Bazel Authors. All rights reserved.
```
f0rmiga, line 1: Could you please update all the new license headers you are adding?

```suggestion
# Copyright 2024 The Bazel Authors. All rights reserved.
```

f0rmiga:
> Don't have time to look into this right now, but after reading your last comment, I am wondering if the right way to avoid the footgun would be to use the import path as an argument to the gazelle comment, but that is almost as a noop import statement using just python? Which puts us back to square 1.
> 
> I'll think about this a little more.

We are overthinking here. If the user gets to this point, the entire implementation will look wrong.

gazelle/python/parser.go
dougthor42: Updated 👍

gazelle/python/testdata/annotation_include_dep/test.yaml
dougthor42: Done!

dougthor42 (commented):
>> I'll think about this a little more.
>
> We are overthinking here. If the user gets to this point, the entire implementation will look wrong (the user code).

👍  we won't try to prevent the user from adding a non-python dependency.

f0rmiga (approved):

f0rmiga:
Thank you, @dougthor42!

PR #1864, opened 2024-04-23
docs: document how to configure MODULE.bazel for different use cases

rickeylev:
This largely comes from being regularly tagged on BCR PRs about how a module should be configured. This info was sort of already in the getting started docs, but that isn't an obvious place to go and bloated the what should be simplified steps to getting started.


aignas (approved):
Thanks!

PR #1875, opened 2024-05-04
fix(whl_library): fix the dependency generation for multi-python depenency closures

aignas:
We start using the recently introduced `is_python_config_setting` to make
it possible to have a working select statement when multiple python
version selection needs to happen in a `whl_library`.

This adds further fixes so that the correct dependencies are pulled in when the
`python_version` string flag is unset thus making this implementation suitable
for `bzlmod` use case where we would use a single `whl_library` instance for
multiple python versions within the hub.

Work towards #735.


examples/bzlmod/MODULE.bazel
```diff
@@ -122,12 +122,14 @@ pip.parse(
...
-        "cp39_linux_x86_64",
-        "cp39_linux_*",
         "cp39_*",
+        "cp310_*",
+        "cp311_*",
+        "cp312_*",
     ],
     hub_name = "pip",
     python_version = "3.9",
+    quiet = False,
```
rickeylev, line 132: Just checking: debug code?

rickeylev (approved):

PR #1892, opened 2024-05-12
refactor: Mark internal and `python` extension as reproducible

fmeum:
This avoids unnecessary lockfile entries for users of rules_python.


fmeum:
@aignas Could you review this?

MODULE.bazel
```diff
@@ -5,7 +5,7 @@ module(
 )
 
 bazel_dep(name = "bazel_features", version = "1.9.1")
-bazel_dep(name = "bazel_skylib", version = "1.3.0")
+bazel_dep(name = "bazel_skylib", version = "1.6.1")
```
aignas, line 8: Could you please add a note in the `CHANGELOG` that the `bazel_skylib` version has been bumped? I think this detail might be great to see at a glance when we cut the next release.

aignas (commented):
LGTM in general, would love a line in `CHANGELOG` about the bumped dependency.

It does mean that the `rules_python_gazelle_plugin` will always have a warning because it depends on `1.3`, so maybe we should bump the `gazelle/MODULE.bazel` as well?

fmeum:
I also marked the `python` extension as reproducible.

aignas (approved):

Issue #1315, opened 2023-07-14
gazelle: ERROR: failed to parse: read |0: file already closed


siennathesane:
# 🐞 bug report

### Affected Rule

I believe it is this one:

```bzl
load("@rules_python_gazelle_plugin//:def.bzl", "GAZELLE_PYTHON_RUNTIME_DEPS")
gazelle(
    name = "gazelle",
    data = GAZELLE_PYTHON_RUNTIME_DEPS,
    gazelle = "@rules_python_gazelle_plugin//python:gazelle_binary",
)
```

### Is this a regression?

I don't know, I'm a new user

### Description

Gazelle fails to generate build configurations

## 🔬 Minimal Reproduction

I don't know how to reproduce this, sorry

## 🔥 Exception or Error

```sh
$ bazel run //:gazelle                                
INFO: Invocation ID: 2bd54e55-b45f-4f5e-871f-df99e91d5daa
INFO: Analyzed target //:gazelle (37 packages loaded, 713 targets configured).
INFO: Found 1 target...
Target //:gazelle up-to-date:
  bazel-bin/gazelle-runner.bash
  bazel-bin/gazelle
INFO: Elapsed time: 0.364s, Critical Path: 0.03s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
INFO: Running command line: bazel-bin/gazelle
Traceback (most recent call last):
  File "/private/var/tmp/_bazel_sienna/16dcdf3abbda1ef8268213b48b0911a1/execroot/veridian/bazel-out/darwin_arm64-fastbuild/bin/gazelle.runfiles/rules_python_gazelle_plugin/python/parse.py", line 106, in <module>
    exit(main(sys.stdin, sys.stdout))
  File "/private/var/tmp/_bazel_sienna/16dcdf3abbda1ef8268213b48b0911a1/execroot/veridian/bazel-out/darwin_arm64-fastbuild/bin/gazelle.runfiles/rules_python_gazelle_plugin/python/parse.py", line 91, in main
    outputs.append(parse(repo_root, rel_package_path, filenames[0]))
  File "/private/var/tmp/_bazel_sienna/16dcdf3abbda1ef8268213b48b0911a1/execroot/veridian/bazel-out/darwin_arm64-fastbuild/bin/gazelle.runfiles/rules_python_gazelle_plugin/python/parse.py", line 66, in parse
    content = file.read()
[...]

PR #1895, opened 2024-05-13
feat(gazelle): pure golang helper

### LINKED ISSUE #1599: Gazelle failing when executed from within a virtual environment
# 🐞 bug report

### Affected Rule

 The issue is caused by the Gazelle Python plugin.

### Is this a regression?

 Yes, the previous version in which this bug was not present was: 0.26.0

### Description

 Gazelle (with the Python plugin enabled) crashes when executed from within a virtual environment.

## 🔬 Minimal Reproduction

A reproduction can be found at https://github.com/psalaberria002/rules_python_gazelle_repro.

```
$ bazel run //:venv
$ source .venv/bin/activate
$ bazel run //:gazelle
```

## 🔥 Exception or Error

<pre><code>
$ bazel run gazelle                                                          
INFO: Analyzed target //:gazelle (60 packages loaded, 9983 targets configured).
INFO: Found 1 target...
Target //:gazelle up-to-date:
  bazel-bin/gazelle-runner.bash
  bazel-bin/gazelle
INFO: Elapsed time: 2.561s, Critical Path: 2.34s
INFO: 7 processes: 6 internal, 1 linux-sandbox.
INFO: Build completed successfully, 7 total actions
INFO: Running command line: bazel-bin/gazelle
Could not find platform independent libraries <prefix>
Could not find platform dependent libraries <exec_prefix>
Python path configuration:
  PYTHONHOME = (not set)
  PYTHONPATH = (not set)
  program name = 'python3'
  isolated = 0
  environment = 1
  user site = 0
  safe_path = 0
  import site = 1
  is in build tree = 0
  stdlib dir = '/install/lib/python3.11'
  sys._base_executable = ''
  sys.base_prefix = '/install'
  sys.base_exec_prefix = '/install'
  sys.platlibdir = 'lib'
  sys.executable = ''
  sys.prefix = '/install'
  sys.exec_prefix = '/install'
  sys.path = [
    '/install/lib/python311.zip',
    '/install/lib/python3.11',
    '/install/lib/python3.11/lib-dynload',
  ]
Fatal Python error: init_fs_encoding: failed to get the Python codec of the filesystem encoding
Python runtime state: core initialized
ModuleNotFoundError: No module named 'encodings'

Current thread 0x00007faea97d9740 (most recent call first):
  <no Python frame>
gazelle: failed to wait for std_modules: exit status 1
</code></pre>

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
Operating System: Pop!_OS 22.04 LTS               
Kernel: Linux 6.4.6-76060406-generic
Architecture: x86-64
  </code>
</pre>

<pre>
  <code>
Bazelisk version: v1.12.0
Build label: 7.0.0rc6
Build time: Mon Dec 4 22:34:51 2023 (1701729291)
Build timestamp: 1701729291
Build timestamp as int: 1701729291
  </code>
</pre>

**Rules_python version:**

<pre>
  <code>
0.27.1
  </code>
</pre>

**Anything else relevant?**
It started to break in 1.27.0, and 1.27.1 doesn't solve it.

Running Gazelle from outside the venv works as expected.

I have also tried creating the virtual environment with `python -m venv .venv` to rule out any possible issues
with our rules_pyvenv, but it also fails with the same error.

### LINKED ISSUE #1825: gazelle extension breaks cc toolchain resolution during cross compiling on darwin arm64
# 🐞 bug report

### Affected Rule

 gazelle extension

### Is this a regression?

 No, previous versions also has this issue

### Description

After enabling python extension in gazelle 
https://github.com/xytan0056/cgo_test/blob/30a62b2654561f0d08e98dbf14c2e6a7c8d55e74/BUILD.bazel#L22
```
gazelle_binary(
    name = "gazelle_bin",
    languages = [
...
        "@com_github_bazelbuild_rules_python_gazelle//python",
    ]
)
```
We cannot cross compile gazelle on "darwin arm64" to "darwin amd64" with `--incompatible_enable_cc_toolchain_resolution`

## 🔬 Minimal Reproduction

https://github.com/xytan0056/cgo_test/tree/30a62b2654561f0d08e98dbf14c2e6a7c8d55e74
```bash
bazel build --announce_rc  //:gazelle_bin --platforms=@io_bazel_rules_go//go/toolchain:darwin_amd64 --incompatible_enable_cc_toolchain_resolution
```
fails
## 🔥 Exception or Error
<pre><code>

        File "/virtual_builtins_bzl/common/cc/cc_toolchain_alias.bzl", line 26, column 48, in _impl
        File "/virtual_builtins_bzl/common/cc/cc_helper.bzl", line 219, column 17, in _find_cpp_toolchain
Error in fail: Unable to find a CC toolchain using toolchain resolution. Target: @@bazel_tools//tools/cpp:current_cc_toolchain, Platform: @@io_bazel_rules_go//go/toolchain:darwin_amd64, Exec platform: @@local_config_platform//:host
</code></pre>

The error is gone when either
1. python extension is removed or 
2. CC toolchain resolution is disabled via `--noincompatible_enable_cc_toolchain_resolution`

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
uname -a
Darwin tanx-D5VYHJD2X7 23.4.0 Darwin Kernel Version 23.4.0: Wed Feb 21 21:44:43 PST 2024; root:xnu-10063.101.15~2/RELEASE_ARM64_T6000 arm64
  </code>
</pre>

<pre>
  <code>
Bazelisk version: development
Build label: 7.1.1
Build time: Thu Mar 21 18:08:59 2024 (1711044539)
Build timestamp: 1711044539
Build timestamp as int: 1711044539
  </code>
</pre>

**Rules_python version:**

<pre>
  <code>
v0.0.0-20240321125912-c5c03b2477dd
  </code>
</pre>

**Anything else relevant?**

hunshcn:
Remove gazelle plugin's python deps and make it hermetic. No more relying on the system interpreter.

Use TreeSitter to parse Python code and use https://github.com/pypi/stdlib-list to determine whether a module is in std lib.

Fixes #1825
Fixes #1599
Related #1315


hunshcn:
TODO: add change log and remove 3.11.txt. Continue tomorrow

gazelle/BUILD.bazel
hunshcn, line 1: Can I make `distribution`'s visibility to be public?
Although this is strange, I don't want to import rules_python just for visibility.

hunshcn:
now //gazelle have no python dep. So I try to remove rules_python. But it will case test error. https://buildkite.com/bazel/rules-python-python/builds/7651#018f760c-1662-459b-a5b0-2c6b72446b62
Although it's not a big impact, it puzzles me. Can someone answer it for me?

hunshcn:
All tests passed. docs the failure of docs can be ignored. It is now ready for review.

aignas (commented):
This change is very exciting, so thank you very much for working on this.

Could you change the PR description to remove the words "pain in the ass", please? This will become a commit message and I would love to keep the git commit messages more formal.

gazelle/python/BUILD.bazel
```diff
@@ -42,35 +41,18 @@ go_library(
...
-    ],
-    imports = ["."],
-)
-
-filegroup(
-    name = "helper.zip",
-    srcs = [":helper"],
-    output_group = "python_zip_file",
+genrule(
+    name = "stdlib_list",
```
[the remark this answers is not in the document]
hunshcn: to change the name.

I cannot use "@python_stdlib_list_3_11//file" as embedsrcs. Matched no files regardless of using `file` or `3.11.txt`

gazelle/def.bzl
```diff
@@ -12,8 +12,14 @@
...
-]
+def non_module_deps_impl(_):
+    http_file(
+        name = "python_stdlib_list",
+        sha256 = "3c1dbf991b17178d6ed3772f4fa8f64302feaf9c3385fef328a0c7ab736a79b1",
+        url = "https://raw.githubusercontent.com/pypi/stdlib-list/8cbc2067a4a0f9eee57fb541e4cd7727724b7db4/stdlib_list/lists/3.11.txt",
+        downloaded_file_path = "3.11.txt",
+    )
+
+non_module_deps = module_extension(implementation = non_module_deps_impl)
```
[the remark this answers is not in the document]
hunshcn: done

gazelle/python/BUILD.bazel
```diff
@@ -21,11 +20,11 @@ go_library(
     # generated files, but helper.zip is generated by a build rule.
     #
     # You will get a benign error like when running gazelle locally:
-    # > 8 gazelle: .../rules_python/gazelle/python/lifecycle.go:26:3: pattern helper.zip: matched no files
+    # > 8 gazelle: .../rules_python/gazelle/python/std_modules.go:24:3: pattern 3.11.txt: matched no files
     #
     # See following for more info:
     # https://github.com/bazelbuild/bazel-gazelle/issues/1513
-    embedsrcs = [":helper.zip"],  # keep
+    embedsrcs = ["stdlib_list.txt"],  # keep
```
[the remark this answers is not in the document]
hunshcn: This is not a simple matter. Because there may be multiple python version in monorepo, this may require a new directive. I suggest keeping the status quo in this pr and using stdlist of 3.11.

gazelle/python/BUILD.bazel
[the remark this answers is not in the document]
aignas: Maybe using `bazel-skylib`'s `copy_file` rule would be a good solution here?

gazelle/python/BUILD.bazel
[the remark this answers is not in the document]
aignas: SGTM, could you please create an issue in the backlog to add this feature? I agree that having a directive or solving this issue in some way would be useful.

gazelle/WORKSPACE
```diff
@@ -34,16 +34,11 @@ local_repository(
     path = "..",
 )
 
-load("@rules_python//python:repositories.bzl", "py_repositories", "python_register_toolchains")
+load("@rules_python//python:repositories.bzl", "py_repositories")
```
aignas, line 37: Should we remove the reference to `rules_python` altogether and the initialization of `py_repositories` from `rules_python`? We should just include the `bazel-skylib` ourselfs if that is our dependency.

gazelle/python/BUILD.bazel
```diff
@@ -21,11 +20,11 @@ go_library(
     # generated files, but helper.zip is generated by a build rule.
```
aignas, line 20:
```suggestion
    # generated files, but stdlib_list.txt is generated by a build rule.
```

gazelle/python/file_parser.go
```diff
@@ -0,0 +1,201 @@
...
+}
+
+func NewFileParser() *FileParser {
+	return &FileParser{}
+}
+
+func ParseCode(code []byte) (*sitter.Node, error) {
+	parser := sitter.NewParser()
+	parser.SetLanguage(python.GetLanguage())
+
```
aignas, line 61: I guess it might be not super important as going to tree-sitter will be a performance win, but I wonder if the parser can be reused for multiple files? In case we have many python files that we need to parse, that would avoid us from creating unnecessary objects.

gazelle/python/file_parser_test.go
```diff
@@ -0,0 +1,259 @@
...
+
+func TestParseImportStatements(t *testing.T) {
+	t.Parallel()
+	type unit struct {
+		name     string
+		code     string
+		filepath string
+		result   []module
+	}
+	units := []unit{
```
aignas, line 32:
```suggestion
	units := []struct{
		name     string
		code     string
		filepath string
		result   []module
	}
```

Trying to follow the UBER style guide for go would condense this line a little: https://github.com/uber-go/guide/blob/master/style.md#test-tables

aignas (commented):
LGTM in general, a few minor additions.

gazelle/python/file_parser.go
aignas: I guess let's just create a new parser for each new file, it's unimportant to optimize for this right now.

aignas:
@dougthor42, @linzhp, could you test this PR in your repos? Does this speed things up? Does it slow things down? Since you two have bigger repos, it would be nice to use them as tests to ensure there are no hidden edge cases.

gazelle/WORKSPACE
hunshcn: I think so. I have tried but failed. #issuecomment-2109503018 .And I don't known why

linzhp:
Very exciting to see this! I will test it in Uber later this week if that's ok.

dougthor42:
Oh this looks really great! Let's see how it does...

/cc @ssmall, another Googler who will be using this.


## Summary

**This branch causes Gazelle to take ~35% of the time, a savings of 40s for me. Hot damn that's awesome!**

+ "Before this PR" commit: 55f31a306c8d69b46e50a1b2618c93081d4e11e8
+ "After this PR" commit: f6a190add150c12eeb862566de15df6f2f630444

| Before this PR | After this PR | Delta | % |
| :--: | :--: | :--: | :--: |
| 60.8 s | 21.2 s | 40.6 s | ~35% |


## Background info:

+ 4731 python files. `find src -type f -name "*.py" | wc -l`
+ 1618 directories. `find src -type d -not -name "__pycache__" | wc -l`
+ 499 `BUILD.bazel` files previously generated by Gazelle. `find src -type f -name "BUILD.bazel" | wc -l`
+ ~220 defined Python requirements in `requirements.in`
+ bazel version: 7.1.1
+ 24 core AMD EPYC vCPU
+ 96 GB RAM
+ SSD
+ Using experimental bazel downloader for `pip.parse`? yes.

```
$ lsb_release -a
No LSB modules are available.
Distributor ID: Debian
Description:    Debian GNU/Linux rodete
Release:        n/a
Codename:       rodete
```


## Test:

1. Run `hyperfine --warmup 1 --runs 5 'bazel run //:gazelle'`
    + only 5 runs instead of the default 10 because the old method takes >1min and I'm impatient.
    + warmup allows bazel downloader to cache my project's python dependencies
2. Make sure that the `git status` of the test project is clean.


### Before this PR:

Commit 55f31a306c8d69b46e50a1b2618c93081d4e11e8

```console
$ hyperfine --warmup 1 --runs 5 'bazel run //:gazelle'
Benchmark 1: bazel run //:gazelle
  Time (mean ± σ):     60.756 s ±  0.867 s    [User: 53.651 s, System: 7.891 s]
  Range (min … max):   60.018 s … 61.873 s    5 runs
```


### After this PR:

Commit f6a190add150c12eeb862566de15df6f2f630444

```console
$ hyperfine --warmup 1 --runs 5 'bazel run //:gazelle'                                                                                                                                                                                                                                                                        
Benchmark 1: bazel run //:gazelle
  Time (mean ± σ):     21.190 s ±  0.174 s    [User: 27.083 s, System: 2.679 s]
  Range (min … max):   20.964 s … 21.392 s    5 runs
```

gazelle/python/BUILD.bazel
```diff
@@ -42,35 +42,17 @@ go_library(
...
-    imports = ["."],
-)
-
-filegroup(
-    name = "helper.zip",
-    srcs = [":helper"],
-    output_group = "python_zip_file",
+copy_file(
+    name = "stdlib_list",
+    src = "@python_stdlib_list_3_11//file",
```
aignas, line 53: After thinking a little bit, I think we should have a select statement and include other `python_stdlib_lists` for other python versions. Before this PR the user could do `bazel build --@rules_python//python/config_settings:python_version=3.12 //:gazelle` to get a gazelle binary that includes the `3.12` interpreter in the helper.zip (note, the system interpreter was needed to launch the `helper.zip`, but it would then use the specific python version to run the interpretation of the files.

So something like the following could work:
```suggestion
    src = select(
        {
            "@rules_python//python/config_settings:is_python_3.8": "@python_stdlib_list_3_8//file",
            "@rules_python//python/config_settings:is_python_3.9": "@python_stdlib_list_3_9//file",
            "@rules_python//python/config_settings:is_python_3.10": "@python_stdlib_list_3_10//file",
            "@rules_python//python/config_settings:is_python_3.11": "@python_stdlib_list_3_11//file",
            "@rules_python//python/config_settings:is_python_3.12": "@python_stdlib_list_3_12//file",
            # This is the same behaviour as previously
            "//conditions:default": "@python_stdlib_list_3_11//file",
        },
    ),
```

I still think that the ticket #1901 makes sense, because then the users would not need to compile the gazelle binary for each python version, but I do think that we should not regress the current behaviour.

hunshcn:
Because #1887 introduced an error (see https://github.com/bazelbuild/rules_python/pull/1905 for details). I dropped #1887 in the current branch, otherwise it may affect the test of @linzhp  (if it is a multi-language mixed repo). 

dougthor42‘s test was not affected, because I hadn't rebase main at that time. ...

It has no effect on the current pr, just a description.

gazelle/python/python_test.go
```diff
@@ -42,9 +41,10 @@ const (
 	gazelleBinaryName = "gazelle_binary"
 )
 
-var gazellePath = mustFindGazelle()
+var gazellePath string
```
aignas, line 44: Why is this necessary? Since this is a global variable, it may be better to set it in `TestMain` or as it was done previously?

gazelle/def.bzl
```diff
@@ -15,5 +15,13 @@
 """This module contains the Gazelle runtime dependencies for the Python extension.
 """
 
+load("@bazel_skylib//lib:modules.bzl", "modules")
+load(":deps.bzl", "python_stdlib_list_deps")
+
 GAZELLE_PYTHON_RUNTIME_DEPS = [
 ]
+
+non_module_deps = modules.as_extension(
```
aignas, line 24: Could you please move this into `gazelle/python/private:extensions.bzl`? please? The `modules.as_extension` is something that is not our external API within the gazelle plugin, so it should not be a part of `def.bzl`.

gazelle/python/python_test.go
hunshcn: `var gazellePath = mustFindGazelle()` will make `go test .` failed. Although this will not affect bazel test.

I simply modified it, and no longer use global variables. There is no need for global variables here.

linzhp:
This works in Uber

aignas (approved):
LGTM, thank you so much for this contribution!

PR #1904, opened 2024-05-16
feat: add whl_filegroup

### LINKED ISSUE #1311: Target aliases don't work with annotations in pip_parse()
### Affected Rule

pip_parse

### Is this a regression?

Didn't try this in other versions

### Description

if you pass both arguments `incompatible_generate_aliases` and `annotations` to pip_parse(), specified annotations will be ignored.

## 🔬 Minimal Reproduction

<pre>
  <code>
NUMPY_ANNOTATIONS = {
    "numpy": package_annotation(
        additive_build_content = """\
filegroup(
    name = "includes",
    srcs = glob(["site-packages/numpy/core/include/**/*.h"]),
)
cc_library(
    name = "numpy_headers",
    hdrs = [":includes"],
    strip_include_prefix="site-packages/numpy/core/include/",
)
""",
    ),
}

pip_parse(
    name = "pypi",
    annotations = NUMPY_ANNOTATIONS,
    incompatible_generate_aliases = True,
    requirements = "//:requirements_lock.txt",
)
  </code>
</pre>

## 🔥 Exception or Error

When you try to use the annotated `cc_library`, it will just fail, e.g.
`no such target '@pypi//numpy:numpy_headers': target 'numpy_headers' not declared in package 'numpy'`

Without `incompatible_generate_aliases` it works as intended.

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
gLinux
  </code>
</pre>

<pre>
  <code>
Build label: 6.1.0
Build time: Mon Mar 6 17:09:47 2023 (1678122587)
Build timestamp: 1678122587
Build timestamp as int: 1678122587
  </code>
</pre>

**Rules_python version:**

<pre>
  <code>
0.23.1
  </code>
</pre>

lalten:
This adds a `whl_filegroup` rule which can take a wheel as input and extract it, making the
output available to consumers. A pattern is allowed to better control what files
are extracted.

This is handy to, e.g. get the numpy headers from the numpy wheel so they can be
used in a cc_library. e.g., 

Closes #1903 
Fixes #1311


python/pip.bzl
```diff
@@ -262,6 +262,38 @@ def multi_pip_parse(name, default_version, python_versions, python_interpreter_t
...
     )
 
+def _whl_filegroup_impl(ctx):
+    out_dir = ctx.actions.declare_directory(ctx.attr.name)
+    ctx.actions.run_shell(
+        outputs = [out_dir],
+        inputs = [ctx.file.whl],
+        tools = [ctx.executable._get_wheel_records_tool],
+        arguments = [ctx.attr.pattern],
+        command = '{tool} {wheel} "$1" | xargs unzip -q -d {out_dir} {wheel}'.format(
```
aignas, line 272: Right now this is not hermetic as it depends on `xargs` and `unzip`. I think we could in theory do the `unzipping` entirely within python?

python/pip.bzl
```diff
@@ -262,6 +262,38 @@ def multi_pip_parse(name, default_version, python_versions, python_interpreter_t
...
+    )]
+
+whl_filegroup = rule(
+    _whl_filegroup_impl,
+    doc = "Extract files matching a regex pattern from a wheel file.",
+    attrs = {
+        "pattern": attr.string(default = "", doc = "Only files that match this pattern will be extracted."),
+        "whl": attr.label(mandatory = True, allow_single_file = True, doc = "The wheel to extract files from."),
+        "_get_wheel_records_tool": attr.label(
+            default = Label("//python/pip_install/tools/wheel_installer:get_wheel_records"),
```
aignas, line 290: I would love this rule to be defined in `//python/private/whl_filegroup.bzl` with the `get_wheel_records` tool being made more independent from the `pip_install` code. That code is mostly about working in the `repository_rule` context and thus may be making certain assumptions that this code should not, IMHO.

Then re-exporting the symbol in `pip.bzl` would be one option, another one could be to just have a `//python/whl_filegroup.bzl` that re-exports the symbol.

python/pip_install/tools/wheel_installer/wheel.py
```diff
@@ -583,6 +583,17 @@ def version(self) -> str:
...
+    @property
+    def record(self) -> dict[str, tuple[str, int]]:
+        with installer.sources.WheelFile.open(self.path) as wheel_source:
+            record_content = wheel_source.read_dist_info("RECORD")
+        return {
+            file: (filehash, int(filelen))
+            for line in record_content.splitlines()
+            for file, filehash, filelen in [line.split(",")]
+            if filehash  # Skip RECORD itself, which has no hash or length
+        }
```
aignas, line 595: Just having a tool that uses the installer directly instead of reusing this class would be preferable due to reasons mentioned elsewhere.

tests/integration/whl_filegroup/BUILD.bazel
```diff
@@ -0,0 +1,21 @@
+load("@rules_cc//cc:defs.bzl", "cc_library", "cc_test")
+load("@rules_python//python:pip.bzl", "whl_filegroup")
+
+whl_filegroup(
+    name = "numpy_includes",
+    pattern = "numpy/core/include/numpy",
+    whl = "@pypi//numpy:whl",
+)
```
aignas, line 8: Right now this test won't be executed as part of CI and I would love to ensure that this does not break.

We could create a `whl` in the build context, similarly to how the `examples/wheel` is doing and then use the `whl_filegroup` to do the extracting. We could still include some headers in the `whl` file and have a `cc_library` and `cc_test` as a verification and document that this is mimicking a usecase where `numpy` headers need to be used.

The main recommendation here would be to avoid testing in a separate workspace. :)

It could be also nice to add some documentation in `docs/sphinx/`.

aignas (commented):
Thanks for the PR, looking promising :)

python/pip.bzl
lalten: yep, absolutely. done.

python/pip_install/tools/wheel_installer/wheel.py
lalten: moved this to the tool itself. Don't actually need the installer dep as well, I just roughly replicated https://github.com/pypa/installer/blob/501bdc2ae800a27e362186fe72928f8b5c206e00/src/installer/sources.py#L179-L193

tests/integration/whl_filegroup/BUILD.bazel
lalten: Yea, CI also fails due to stumbling over these files and not seeing @pypi// somehow. Need to look into this, not sure what's different from the other integration tests?

lalten: Moved the test next to the implementation into the same workspace

python/pip.bzl
lalten: moved everything to python/private/whl_filegroup, with a re-export from pip.bzl.

tests/integration/whl_filegroup/BUILD.bazel
aignas: Thanks!

FYI, we usually keep the bzl tests in the tests directory, that way we can have more clear distinction about what is production and non production code.

At least that is true for all new code.

rickeylev: Yes, please move all the test related parts under //tests somewhere, probably a tests/whl_filegroup directory.

Part of the "production vs non production" code distinction is to avoid adding dependencies on test-only code in production code paths. e.g. a BUILD file that loads() something only for testing incurs having to download those repos, parse those packages, parse those loaded bzl files, etc.

rickeylev:
Aside from the pending comments, this looks pretty good. Thanks!

rickeylev (changes requested):

tests/integration/whl_filegroup/BUILD.bazel
lalten: done

python/private/whl_filegroup/whl_filegroup.bzl
```diff
@@ -0,0 +1,56 @@
...
+            ctx.file.whl.path,
+            out_dir.path,
+            ctx.attr.pattern,
+        ],
+        executable = ctx.executable._extract_wheel_files_tool,
+        mnemonic = "ExtractWheelFiles",
+    )
+    return [DefaultInfo(
+        files = depset([out_dir]),
+        runfiles = ctx.runfiles(files = [out_dir]),
```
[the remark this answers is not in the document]
lalten: This is an interesting one, thanks for bringing this up. You're right that we shouldn't have build-time headers in the runfiles. But I can totally imagine use cases where the user would want to extract some data files that need to be available at runtime.
I added a `runfiles` attr that controls this.

python/private/whl_filegroup/whl_filegroup.bzl
```diff
@@ -0,0 +1,56 @@
+"""Implementation of whl_filegroup rule."""
+
+def _whl_filegroup_impl(ctx):
+    out_dir = ctx.actions.declare_directory(ctx.attr.name)
+    ctx.actions.run(
+        outputs = [out_dir],
+        inputs = [ctx.file.whl],
+        tools = [ctx.executable._extract_wheel_files_tool],
```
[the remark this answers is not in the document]
lalten: done

python/private/whl_filegroup/BUILD.bazel
```diff
@@ -0,0 +1,85 @@
...
+bzl_library(
+    name = "whl_filegroup_bzl",
+    srcs = ["whl_filegroup.bzl"],
+    visibility = ["//:__subpackages__"],
+)
+
+py_binary(
+    name = "extract_wheel_files",
+    srcs = ["extract_wheel_files.py"],
+    visibility = ["//:__subpackages__"],
```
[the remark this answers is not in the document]
lalten: done

python/private/whl_filegroup/whl_filegroup.bzl
```diff
@@ -0,0 +1,56 @@
...
+        arguments = [
+            ctx.file.whl.path,
+            out_dir.path,
+            ctx.attr.pattern,
+        ],
+        executable = ctx.executable._extract_wheel_files_tool,
+        mnemonic = "ExtractWheelFiles",
```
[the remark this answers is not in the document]
lalten: done

python/private/whl_filegroup/whl_filegroup.bzl
```diff
@@ -0,0 +1,56 @@
+"""Implementation of whl_filegroup rule."""
+
+def _whl_filegroup_impl(ctx):
+    out_dir = ctx.actions.declare_directory(ctx.attr.name)
+    ctx.actions.run(
```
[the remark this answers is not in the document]
lalten: done

rickeylev (approved):

PR #1905, opened 2024-05-16
fix(gazelle): generate empty py_library only when need

hunshcn:
#1887 incorrectly generated an empty py_library for all dirs, although it will eventually be removed because it is empty, but it will easily conflict with other existing rules.

Fix this error, and only generate an empty py_library for bazel packages with the same name py_library rule.


hunshcn:
The first commit is the badcase of #1887 error. The second commit was fixed.

@aignas Need to merge as soon as possible.

aignas (approved):

PR #1909, opened 2024-05-20
docs: implement Starlark domain plugin for Sphinx

rickeylev:
This implements a Sphinx plugin to support Starlark as a first-class domain in Sphinx.
A Starlark domain allows writing object descriptions directly using Sphinx's object
markup language, which allows better integration with cross references, understanding
and distinguishing types, rendering information about things, and referencing types
from other projects.

Note that this doesn't affect the docs today because the proto_to_markdown tool is
still generating regular markdown; updating that to generate Sphinx domain markdown
will be done separately.

Summary of features:
* Types and arguments can be documented using Python syntax, e.g. `list[str]` This includes
  unions, generics, and custom types. Each component of the type expression is linked and
  cross referenced appropriately. Each object can have its approprate pieces documented and
  defined (e.g. a rule can have attributes, attributes can have their defaults etc
  documented, etc)
* An index of Starlark objects is generated automatically. This makes it easy, for example,
  to find the rules that have a particular attribute.
* The following objects can be documented: functions, methods, rules, repository rules,
  providers, aspects, bzlmod extensions, tag classes, targets, flags, and
  attributes/fields of the aforementioned objects.
* Arbitary docs can cross reference to Starlark types. e.g., a manually written "Getting
  Started" doc can write `{bzl:obj}PyInfo.some_field` and it will automatically link to
  the appropriate API docs.
* Resolution of cross-references is much smarter and customizable. Instead of relying
  Markdown's link resolution rules, the Sphinx's crossreference resolution hooks are
  used. This allows more concise references (e.g., just a rule's base name), distinguishing
  a particular object type (e.g. a function vs rule), or referring to an absolute object.

### DIFF
--- a/sphinxdocs/private/sphinx.bzl
+++ b/sphinxdocs/private/sphinx.bzl
@@ -278,7 +278,6 @@ def _run_sphinx(ctx, format, source_path, inputs, output_prefix):
         mnemonic = "SphinxBuildDocs",
         progress_message = "Sphinx building {} for %{{label}}".format(format),
         env = env,
-        execution_requirements = {"no-sandbox": ""},
     )
     return output_dir
 
--- a/sphinxdocs/private/sphinx.bzl
+++ b/sphinxdocs/private/sphinx.bzl
@@ -15,6 +15,7 @@
 """Implementation of sphinx rules."""
 
 load("@bazel_skylib//lib:paths.bzl", "paths")
+load("@bazel_skylib//rules:common_settings.bzl", "BuildSettingInfo")
 load("//python:py_binary.bzl", "py_binary")
 load("//python/private:util.bzl", "add_tag", "copy_propagating_kwargs")  # buildifier: disable=bzl-visibility
 
@@ -119,6 +120,7 @@ def sphinx_docs(
         output_group = "html",
         **common_kwargs
     )
+
     py_binary(
         name = name + ".serve",
         srcs = [_SPHINX_SERVE_MAIN_SRC],
@@ -187,6 +189,7 @@ _sphinx_docs = rule(
         ),
         "_extra_defines_flag": attr.label(default = "//sphinxdocs:extra_defines"),
         "_extra_env_flag": attr.label(default = "//sphinxdocs:extra_env"),
+        "_quiet_flag": attr.label(default = "//sphinxdocs:quiet"),
     },
 )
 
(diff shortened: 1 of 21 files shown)


sphinxdocs/private/sphinx.bzl
```diff
@@ -273,6 +278,7 @@ def _run_sphinx(ctx, format, source_path, inputs, output_prefix):
         mnemonic = "SphinxBuildDocs",
         progress_message = "Sphinx building {} for %{{label}}".format(format),
         env = env,
+        execution_requirements = {"no-sandbox": ""},
```
aignas, line 281: Why no-sandbox?

sphinxdocs/src/sphinx_stardoc/stardoc.py
```diff
@@ -0,0 +1,1565 @@
...
+    elif ":BUILD" in label:
+        label = label[: label.find(":BUILD")]
+    else:
+        raise InvalidValueError(
+            f"Malformed label: Label must end with .bzl or :BUILD*, got {label}"
+        )
+
+    # Make a //foo:bar.bzl convert to foo.bar, not .foo.bar
+    if label.startswith("//"):
+        label = label.lstrip("/")
```
aignas, line 79: Technically you can have dots in the repo name, so this may break or have degenerate cases.

aignas (approved):
The stardocs.py extension being a single file make the thing a little unwieldy imho. Having it across multiple could improve the readability, but it also has a lot of similar logic, so I am in favour of continuing with a single big file for now.

sphinxdocs/private/sphinx.bzl
rickeylev: Oops, debugging stuff left over. Removed.

sphinxdocs/src/sphinx_stardoc/stardoc.py
rickeylev: Ahhh, good point. Directories and files could have dots in them, too.

I've added a TODO for this for now. I think I'll change the internal representation to use `@repo//file%symbol` notation instead. The dotted notation grew out of originally only having within-same-file references, but it didn't evolve well once targets became cross-referencable, and I found it more natural to write the `@repo//foo:bar%baz` in docs rather than `repo.foo.bar.baz`.

PR #1919, opened 2024-05-23
docs: generate Starlark domain markup instead of regular markdown

rickeylev:
This switches our doc generation over to using the Starlark domain markup from the sphinx_stardoc plugin instead of using regular markdown. This allows the docs generated from code to better integrate with each other and other parts of the doc site.

Overview of changes:

* Makes the doc paths under the API directory more directly mirror their actual location. e.g. moves "defs.md" -> "python/defs.md". This is so the //tools doc  entries have a more natural location, but can also be used for our other top-level directories.
* Adds API docs for some of the well known targets we have. These aren't automatically generated, but use the Starlark domain markup, so integrate nicely with everything.
* Ensures default values are parsable as Python expressions. Stardoc returns values like "<function foo>" or 'Label(*, "//bar")' in some cases for the default value of args/attrs.
* Ensures function signatures don't crash doc rendering. Stardoc gives bad/incomplete information, so reconstructing the original signature of a function is tricky.
* Allows references flags using leading slashes and a value, e.g. `--foo=bar`. This makes it more natural to write while cross referencing to the flag.
* Implements `{any}` xref resolution. It was just totally broken before.
* Adds some additional bzl files that get documented.
* Adds some more Bazel external references.
* Fixes some missing bzl_library dependencies.
* A few minor QoL improvements to the docs dev server:
  * Print the serving directory when CTRL+C is received. This makes it easier to find the raw files that are being generated.
  * Fix an error during shutdown about an unterminated generator.
* The `sphinx_stardocs.footer` arg is removed. This was always just a hack to get extra link targets into the generated bzl docs. It's no longer needed when the bzl domain is used.
* Using `@repo//pkg:file.bzl%Name` syntax is supported in type expressions (e.g. `:type:` option or `{type}` role) by quoting the label. The quoting is necessary because, under the hood, the expressions are parsed as Python.
* Objects directives support an `:origin-key` directive. This records the label identity that Bazel sees for an object (as from the Stardoc origin_key field). The markdown generate doesn't generate this for everything yet because some things are documented twice (e.g. py_binary in defs.bzl and py_binary.bzl), which would cause a crash (type things trying to define the same id).
* Add `*` and `**` to var-args and var-kwargs in signatures.
* Allow providers to be refered to using the `type` role. This allows providers to be referenced in `:type:` directives (e.g. in a provider field).

### DIFF
--- a/docs/sphinx/BUILD.bazel
+++ b/docs/sphinx/BUILD.bazel
@@ -76,24 +76,34 @@ sphinx_inventory(
 sphinx_stardocs(
     name = "bzl_api_docs",
     docs = {
-        "api/cc/py_cc_toolchain.md": dict(
+        "api/python/cc/py_cc_toolchain.md": dict(
             dep = "//python/private:py_cc_toolchain_bzl",
             input = "//python/private:py_cc_toolchain_rule.bzl",
             public_load_path = "//python/cc:py_cc_toolchain.bzl",
         ),
-        "api/cc/py_cc_toolchain_info.md": "//python/cc:py_cc_toolchain_info_bzl",
-        "api/defs.md": "//python:defs_bzl",
-        "api/entry_points/py_console_script_binary.md": "//python/entry_points:py_console_script_binary_bzl",
-        "api/packaging.md": "//python:packaging_bzl",
-        "api/pip.md": "//python:pip_bzl",
+        "api/python/cc/py_cc_toolchain_info.md": "//python/cc:py_cc_toolchain_info_bzl",
+        "api/python/defs.md": "//python:defs_bzl",
+        "api/python/entry_points/py_console_script_binary.md": "//python/entry_points:py_console_script_binary_bzl",
+        "api/python/packaging.md": "//python:packaging_bzl",
+        "api/python/pip.md": "//python:pip_bzl",
+        "api/python/py_binary.md": "//python:py_binary_bzl",
+        "api/python/py_cc_link_params_info.md": "//python:py_cc_link_params_info_bzl",
+        "api/python/py_library.md": "//python:py_library_bzl",
+        "api/python/py_runtime.md": "//python:py_runtime_bzl",
+        "api/python/py_runtime_info.md": "//python:py_runtime_info_bzl",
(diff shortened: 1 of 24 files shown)


docs/sphinx/BUILD.bazel
```diff
@@ -112,6 +122,8 @@ sphinx_build_binary(
         requirement("sphinx_rtd_theme"),
         requirement("myst_parser"),
         requirement("readthedocs_sphinx_ext"),
+        requirement("typing_extensions"),
+        "//sphinxdocs/src/sphinx_stardoc",
```
aignas, line 126: I thought that `shpinx_stardoc` includes the `typing_extensions`? Is it really needed here as well?

sphinxdocs/src/sphinx_stardoc/stardoc.py
```diff
@@ -608,25 +620,52 @@ def make_xref(name):
...
-            sig_node += paramlist_node
-
-            if signature.return_annotation is not signature.empty:
-                sig_node += addnodes.desc_returns("", signature.return_annotation)
+            try:
+                signature = inspect.signature_from_str(params_text)
+            except SyntaxError:
+                # Stardoc doesn't provide accurate info, so the reconstructed
+                # signature might not be valid syntax. Rather than fail, just
+                # provide a plain-text description of the approximate signature.
```
aignas, line 628: Is it something that should be reported upstream?

aignas (approved):
The docs look much better, thanks!

sphinxdocs/src/sphinx_stardoc/stardoc.py
rickeylev: Yes. I added a comment here to reference https://github.com/bazelbuild/stardoc/issues/225

docs/sphinx/BUILD.bazel
rickeylev: It used to, but I removed it from the sphinx_stardoc target so it could be consumed by users outside our repo. There was some interest from others in using this sphinx and stardoc stuff, so I figured that'd give them at least some way to try it out in the meantime.

PR #1928, opened 2024-05-29
doc: fix `toolchains.md` mistake

### LINKED ISSUE #1921: Doc mistake about configuring Python toolchains
# 🐞 bug report

### Affected Rule

 
No rules affacted

### Is this a regression?

 
No

### Description

> ### Root modules
> 
> Root modules are always the top-most module. These are special in two ways:
> 
> 1. Some `rules_python` bzlmod APIs are only respected by the root module.
> 2. The root module can force module overrides and specific module dependency
>    ordering.
> 
> When configuring `rules_python` for a root module, you typically want to
> explicitly specify the Python version you want to use. This ensures that
> dependencies don't change the Python version out from under you. Remember that
> `rules_python` will set a version by default, but it will change regularly as
> it tracks a recent Python version.
> 
> NOTE: If your root module only uses Python for development of the module itself,
> you should read the dev-only library module section.
> 
> ```
> bazel_dep(name="rules_python", version=...)
> python = use_extension("@rules_python//extensions:python.bzl", "python")
> 
> python.toolchain(python_version = "3.12", is_default = True)
> ```

`python = use_extension("@rules_python//extensions:python.bzl", "python")` is used in documentation, but when I following this , python toolchain configuration is not takeing affect. Aftert changing to `python = use_extension("@rules_python//python/extensions:python.bzl", "python")`, it works well.

It is like a doc mistake. 

I am new to bazel, correct me if it is wrong.

## 🔥 Exception or Error

<pre><code>

</code></pre>

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
macOs 14.2.1
  </code>
</pre>

<pre>
  <code>
Bazelisk version: development
Build label: 7.1.2
Build time: Wed May 8 20:50:30 2024 (1715201430)
Build timestamp: 1715201430
Build timestamp as int: 171520143
  </code>
</pre>

**Rules_python version:**

<pre>
  <code>
0.32.2
  </code>
</pre>

**Anything else relevant?**

xinnjie:
Fixes #1921

### DIFF
--- a/docs/sphinx/toolchains.md
+++ b/docs/sphinx/toolchains.md
@@ -63,7 +63,7 @@ specify `dev_dependency = True` to the bzlmod APIs:
 bazel_dep(name = "rules_python", version=..., dev_dependency = True)
 
 python = use_extension(
-    "@@rules_python//python/extensions:python.bzl",
+    "@rules_python//python/extensions:python.bzl",
     "python",
     dev_dependency = True
 )
--- a/docs/sphinx/toolchains.md
+++ b/docs/sphinx/toolchains.md
@@ -38,7 +38,7 @@ you should read the dev-only library module section.
 
 ```
 bazel_dep(name="rules_python", version=...)
-python = use_extension("@rules_python//extensions:python.bzl", "python")
+python = use_extension("@rules_python//python/extensions:python.bzl", "python")
 
 python.toolchain(python_version = "3.12", is_default = True)
 ```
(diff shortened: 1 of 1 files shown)


docs/sphinx/toolchains.md
```diff
@@ -63,7 +63,7 @@ specify `dev_dependency = True` to the bzlmod APIs:
 bazel_dep(name = "rules_python", version=..., dev_dependency = True)
 
 python = use_extension(
-    "@rules_python//extensions:python.bzl",
+    "@@rules_python//python/extensions:python.bzl",
```
aignas, line 66: Let's leave the usage of regular labels
```suggestion
    "@rules_python//python/extensions:python.bzl",
```

aignas (approved):
Thanks for the PR!

Issue #1434, opened 2023-09-21
Allow customized coveragerc for Python coverage


tingilee:
### Description of the feature request:

For python coverage collection, `.coveragerc` is hardcoded in https://github.com/bazelbuild/bazel/blame/d435c6dd6e977a5c3ea1bc726557a9321948a317/tools/python/python_bootstrap_template.txt#L393-L397. To provide configurability, we'd like to introduce custom `.coveragerc` to set configurations including https://coverage.readthedocs.io/en/stable/excluding.html#advanced-exclusion. 

### Which category does this issue belong to?

Configurability, Python Rules

### What underlying problem are you trying to solve with this feature?

Allow coverage configurations. In particular, we're interested in excluding import statements being collected as executable lines in coverage reports. 

### Which operating system are you running Bazel on?

macOS, Linux

### What is the output of `bazel info release`?

release 6.3.0

tingilee:
> How does the binary get the path to the config file? 

In general, I'd like to propose that we set environment variable in rules_python to allow overrides:
```
coverage_env["PYTHON_COVERAGERC"] = ":coveragerc_file"
```
this matches how the coverage binary is defined, for example: 
```
coverage_env["PYTHON_COVERAGE"] = "bazel_python_internal_pytest_pip_deps_coverage/pypi__coverage/coverage"
```

> How do other languages plumb through coverage settings?

I have not seen coverage setting in other languages, ie. cc or Go. This seems to be pretty specific to Python or implementation of coveragepy. 

> Do coverage settings need to be per-test, or does a single build-wide config suffice? 

.coveragerc configuration should be customized for every py_test target. This would allow us to choose/customize coverage collection per test invocation

raylu:
> Do coverage settings need to be per-test, or does a single build-wide config suffice?

I would definitely settle for a single, build-wide config

ewianda:
This will be very useful. I am trying to migrate from pytest-cov and the coverage runs are almost 4X slower. From the look of it, it seems the time is spent on generating coverage report for third party code. Currently I am using the old PYTHON_COVERAGE tool to configure coverage e.g 

```
"""Coverage entry point."""
import argparse
import os
import re
import sys

from coverage.cmdline import main

if __name__ == "__main__":
    sys.argv[0] = re.sub(r"(-script\.pyw|\.exe)?$", "", sys.argv[0])
    sys.path.append(sys.path.pop(0))
    parser = argparse.ArgumentParser()
[...]

PR #1929, opened 2024-05-29
fix: don't require system Python to perform bootstrapping

rickeylev:
This is a pretty major, but surprisingly not that invasive, overhaul of how binaries
are started. It fixes several issues and lays ground work for future improvements.

In brief:

* A system Python is no longer needed to perform bootstrapping.
* Errors due to `PYTHONPATH` exceeding environment variable size limits is no
  longer an issue.
* Coverage integration is now cleaner and more direct.
* The zipapp `__main__.py` entry point generation is separate from the Bazel
  binary bootstrap generation.
* Self-executable zips now have actual bootstrap logic.

The way all of this is accomplished is using a two stage bootstrap process. The first
stage is responsible for locating the interpreter, and the second stage is responsible
for configuring the runtime environment (e.g. import paths). This allows the first
stage to be relatively simple (basically find a file in runfiles), so implementing it
in cross-platform shell is feasible. The second stage, because it's running under the
desired interpreter, can then do things like setting up import paths, and use the
`runpy` module to call the program's real main.

This also fixes the issue of long `PYTHONPATH` environment variables causing an error.
Instead of passing the import paths using an environment variable, they are embedded
into the second stage bootstrap, which can then add them to sys.path.

This also switches from running coverage as a subprocess to using its APIs directly.
This is possible because of the second stage bootstrap, which can rely on
`import coverage` occurring in the correct environment.

This new bootstrap method is disabled by default. It can be enabled by setting
`--@rules_python//python/config_settings:bootstrap_impl=two_stage`. Once the new APIs
are released, a subsequent release will make it the default. This is to allow easier
upgrades for people defining their own toolchains.

The two-stage bootstrap ignores errors during lcov report generation, which
partially addresses https://github.com/bazelbuild/rules_python/issues/1434

Fixes https://github.com/bazelbuild/rules_python/issues/691

* Also fixes some doc cross references.
* Also fixes the autodetecting toolchain and directs our alias to it

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+:::{default-domain} bzl
+:::
+
 # rules_python Changelog
 
 This is a human-friendly changelog in a keepachangelog.com style format.
@@ -31,7 +34,7 @@ A brief description of the categories of changes:
   marked as `reproducible` and will not include any lock file entries from now
   on.
 
-* (gazelle): Remove gazelle plugin's python deps and make it hermetic. 
+* (gazelle): Remove gazelle plugin's python deps and make it hermetic.
   Introduced a new Go-based helper leveraging tree-sitter for syntax analysis.
   Implemented the use of `pypi/stdlib-list` for standard library module verification.
 
@@ -80,6 +83,16 @@ A brief description of the categories of changes:
   invalid usage previously but we were not failing the build. From now on this
   is explicitly disallowed.
 * (toolchains) Added riscv64 platform definition for python toolchains.
+* (rules) A new bootstrap implementation that doesn't require a system Python
+  is available. It can be enabled by setting
+  {obj}`--@rules_python//python:config_settings:bootstrap_impl=two_phase`. It
+  will become the default in a subsequent release.
+  ([#691](https://github.com/bazelbuild/rules_python/issues/691))
+* (providers) `PyRuntimeInfo` has two new attributes:
+  {obj}`PyRuntimeInfo.stage2_bootstrap_template` and
+  {obj}`PyRuntimeInfo.zip_main_template`.
+* (toolchains) A replacement for the Bazel-builtn autodetecting toolchain is
+  available. The `//python:autodetecting_toolchain` alias now uses it.
 
 [precompile-docs]: /precompiling
 
--- a/python/config_settings/BUILD.bazel
+++ b/python/config_settings/BUILD.bazel
@@ -1,6 +1,7 @@
 load("@bazel_skylib//rules:common_settings.bzl", "string_flag")
 load(
     "//python/private:flags.bzl",
+    "BootstrapImplFlag",
     "PrecompileAddToRunfilesFlag",
     "PrecompileFlag",
     "PrecompileSourceRetentionFlag",
@@ -49,6 +50,15 @@ string_flag(
     name = "pyc_collection",
     build_setting_default = PycCollectionFlag.DISABLED,
     values = sorted(PycCollectionFlag.__members__.values()),
+    # NOTE: Only public because its an implicit dependency
+    visibility = ["//visibility:public"],
+)
+
+string_flag(
+    name = "bootstrap_impl",
+    # todo: default to deprecated_one_stage initially
+    build_setting_default = BootstrapImplFlag.SCRIPT,
+    values = sorted(BootstrapImplFlag.__members__.values()),
     # NOTE: Only public because its an implicit dependency
     visibility = ["//visibility:public"],
 )
--- a/python/config_settings/BUILD.bazel
+++ b/python/config_settings/BUILD.bazel
@@ -56,8 +56,7 @@ string_flag(
 
 string_flag(
     name = "bootstrap_impl",
-    # todo: default to deprecated_one_stage initially
-    build_setting_default = BootstrapImplFlag.SCRIPT,
+    build_setting_default = BootstrapImplFlag.SYSTEM_PYTHON,
     values = sorted(BootstrapImplFlag.__members__.values()),
     # NOTE: Only public because its an implicit dependency
     visibility = ["//visibility:public"],
--- a/python/private/BUILD.bazel
+++ b/python/private/BUILD.bazel
@@ -376,8 +376,53 @@ exports_files(
     visibility = ["//visibility:public"],
 )
 
+filegroup(
+    name = "stage1_bootstrap_template",
+    srcs = ["stage1_bootstrap_template.sh"],
+    # Not actually public. Only public because it's an implicit dependency of
+    # py_runtime.
+    visibility = ["//visibility:public"],
+)
+
+filegroup(
+    name = "stage2_bootstrap_template",
+    srcs = ["stage2_bootstrap_template.py"],
+    # Not actually public. Only public because it's an implicit dependency of
+    # py_runtime.
+    visibility = ["//visibility:public"],
+)
+
+filegroup(
+    name = "zip_main_template",
+    srcs = ["zip_main_template.py"],
+    # Not actually public. Only public because it's an implicit dependency of
+    # py_runtime.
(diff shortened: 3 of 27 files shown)


python/config_settings/BUILD.bazel
```diff
@@ -52,3 +53,12 @@ string_flag(
     # NOTE: Only public because its an implicit dependency
     visibility = ["//visibility:public"],
 )
+
+string_flag(
+    name = "bootstrap_impl",
```
aignas, line 58: I was curious if we can find a better name that does not talk about stages. In some sense a casual user only cares if they need system python or not, so maybe that should be the flags name.

python/private/BUILD.bazel
```diff
@@ -376,9 +376,51 @@ exports_files(
...
+    # Not actually public. Only public because it's an implicit dependency of
+    # py_runtime.
+    visibility = ["//visibility:public"],
+)
+
+alias(
+    name = "bootstrap_template",
+    actual = select({
+        ":is_two_stage_bootstrap_enabled": "stage1_bootstrap_template.sh",
+        "//conditions:default": "python_bootstrap_template.txt",
```
aignas, line 407: Adding a comment that would say that the windows OS is using a different bootstrap method altogether would be good.

python/private/common/providers.bzl
```diff
@@ -216,13 +258,58 @@ correctly.
...
+The following substitutions are made during template expansion:
+* `%stage2_bootstrap%`: A runfiles-relative string to the stage 2 bootstrap file.
+* `%python_binary%`: The path to the target Python interpreter. There are three
+  types of paths:
+  * An absolute path to a system interpreter (e.g. begins with `/`).
+  * A runfiles-relative path to an interpreter (e.g. `somerepo/bin/python3`)
+  * A program to search for on PATH, i.e. a word without spaces, e.g. `python3`.
+* `%workspace_name%`: The name of the workspace for the built target.
+
+:::{std:versionadded} 0.33.0
```
aignas, line 311: I really like this way of documenting things!

python/private/stage1_bootstrap_template.sh
```diff
@@ -0,0 +1,117 @@
+#!/bin/bash
+
+set -e
```
aignas, line 3: What about undefined vars?
```suggestion
set -eu
```

Although I see that it would cause a few problems with how we use `-z` right now.

python/private/stage2_bootstrap_template.py
```diff
@@ -0,0 +1,501 @@
+# This is a "stage 2" bootstrap. We can assume we've running under the desired
+# interpreter, with some of the basic interpreter options/envvars set.
+# However, more setup is required to make the app's real main file runnable.
```
aignas, line 3: I assume that most of the contents got copied from the main template? Maybe we should mark it as such?

The `CamelCase` usage in functions is something that is legacy, I presume.

aignas (approved):
LGTM, thank you!

python/private/stage2_bootstrap_template.py
```diff
@@ -0,0 +1,501 @@
...
+    )
+
+    # COVERAGE_DIR is set if coverage is enabled and instrumentation is configured
+    # for something, though it could be another program executing this one or
+    # one executed by this one (e.g. an extension module).
+    if os.environ.get("COVERAGE_DIR"):
+        cov_tool = FindCoverageEntryPoint(module_space)
+        if cov_tool is None:
+            PrintVerboseCoverage(
+                "Coverage was enabled, but python coverage tool was not configured."
```
aignas, line 461: should we print a more helpful message with a link to docs?

python/private/stage2_bootstrap_template.py
```diff
@@ -0,0 +1,501 @@
...
+            python_path_entries = Deduplicate(python_path_entries)
+    else:
+        cov_tool = None
+
+    sys.stdout.flush()
+    # TODO: Putting all the import paths at the front isn't good because it
+    # can interfere with the stdlib. But it emulates the historical behavior
+    # of using PYTHONPATH to add import paths.
+    # NOTE: The sys.path must be modified before coverage is imported/activated
+    sys.path[0:0] = python_path_entries
```
aignas, line 494: Since we now know where the python interpreter is, maybe we should have `stdlib` first and only then the rest of the packages? That way we don't shadow the `stdlib` and we may be able to avoid some surprising behaviour.

I am curious if we should keep the old behaviour with the 2 stage bootstrap or make minor modifications since we know better about the environment we are going to be executing.

python/private/stage1_bootstrap_template.sh
rickeylev: It would seem -u causes empty arrays to be treated as undefined on some platforms. Probably depending on the bash version? My local bash is OK with it, but not the mac/RBE shells. Looking around, it seems properly handling empty arrays with -u is very baroque, so I removed -u. (the empty array in question is interpreter_args, but I'd like to land another pr to allow populating that, and then this would be an annoying problem).

rickeylev:
Well, it's about my EOD, so I thought I'd post the latest laugh/cry/wtf moment from fixing CI failures. It's kind of an amusing regression that actually looks like a promising fix of another undesirable behavior that has long been a plague.

The new bootstrap doesn't suffer from the `sys.path[0]` problem where the directory of main is added to sys.path. This is even if safepath is disabled or running in a pre 3.11 version. Yay!

Why? Because the e.g. `python foo/bar.py` path passed resolves to e.g. `bazel-out/bin/.../foo`, which only contains other generated files, thus preventing _most_ imports for that sys.path entry.

But, ironically, fixing that causes another problem. Boo! Without that entry, the runfiles root is then the first sys.path entry. One of our tests performs `import __init__` (odd, but allowed), because it assumes that will come from the repo root. What else has `__init__` ? The runfiles root has an empty one from being auto-generated. So it was actually the undesirable non-safepath behavior making this work.

But, good news: because we can perform more arbitrary sys.path manipulations, we can manipulate sys.path as we need. Stated another way: we can implement `SAFEPATH=True|False` behavior in our bootstrap regardless of Python version. Yay!

python/private/stage2_bootstrap_template.py
rickeylev: Ok, I looked into this. It looks fine.

In the google runtime, it goes [runfiles, stdlib, third party deps], but without any explanation as to why. I suspect it's just historical. It's just simple prepend/append without any special logic.

I updated the code to append the path and CI is happy. The one exception is the safepath behavior -- that path is still prepended. I think that's fine and matches regular Python behavior better anyways.

python/private/stage2_bootstrap_template.py
rickeylev: I tried a few things, but I'm not sure they took. The github UI never showed it as a copy. I'm not sure what the commit history showed, but since it all gets squashed anyways, I'm not sure it would matter.

python/config_settings/BUILD.bazel
hunshcn: the changelog and pr descriptions are also needs to be modified

CHANGELOG.md
```diff
@@ -80,6 +83,16 @@ A brief description of the categories of changes:
   invalid usage previously but we were not failing the build. From now on this
   is explicitly disallowed.
 * (toolchains) Added riscv64 platform definition for python toolchains.
+* (rules) A new bootstrap implementation that doesn't require a system Python
+  is available. It can be enabled by setting
+  {obj}`--@rules_python//python:config_settings:bootstrap_impl=two_phase`. It
```
hunshcn, line 88: It looks a typo, should be 

`--@rules_python//python/config_settings:bootstrap_impl`

rickeylev: Oops, good catch, thanks @hunshcn 

Sent https://github.com/bazelbuild/rules_python/pull/1959

PR #1933, opened 2024-05-31
feat: Upgrade to protobuf 27.0 and remove py_proto_library

### LINKED ISSUE #1080: Confusing error message if rules_proto isn't loaded
# 🐞 bug report

### Affected Rule

 The issue is caused by the rule: py_proto_library

### Is this a regression?

 No

### Description

If rules_proto isn't loaded, using py_proto_library fails with a somewhat confusing error message:

```
	File "[…]/external/rules_python/python/private/proto/py_proto_library.bzl", line 20, column 38, in <toplevel>
		ProtoLangToolchainInfo = proto_common.ProtoLangToolchainInfo
Error: 'proto_common' value has no field or method 'ProtoLangToolchainInfo'
```

Maybe the error message could be improved slightly, e.g. by checking `hasattr(proto_common, 'ProtoLangToolchainInfo')`.

## 🔬 Minimal Reproduction

WORKSPACE:

```python

http_archive(
    name = "rules_python",
    sha256 = "29a801171f7ca190c543406f9894abf2d483c206e14d6acbd695623662320097",
    strip_prefix = "rules_python-0.18.1",
    url = "https://github.com/bazelbuild/rules_python/releases/download/0.18.1/rules_python-0.18.1.tar.gz",
)

py_repositories()
```

BUILD:

```python
load("@rules_python//python:proto.bzl", "py_proto_library")

py_proto_library(
    name = "test_py_pb2",
    deps = [":test_proto"],
)

proto_library(
    name = "test_proto",
    srcs = ["test.proto"],
)
```

test.proto:

```
syntax = "proto3";
message Test {}
```

## 🔥 Exception or Error

```
	File "[…]/external/rules_python/python/private/proto/py_proto_library.bzl", line 20, column 38, in <toplevel>
		ProtoLangToolchainInfo = proto_common.ProtoLangToolchainInfo
Error: 'proto_common' value has no field or method 'ProtoLangToolchainInfo'
```

## 🌍  Your Environment

Debian GNU/Linux

6.0.0

**Rules_python version:**

0.18.1

### LINKED ISSUE #1293: Support generation of pyi files with py_proto_library
# 🚀 feature request

### Relevant Rules

New feature for `py_proto_library`

### Description

`protoc` v3.20 and above supports to generate `.pyi` files from protos. Whould be great if `py_proto_library` would/could generate those as well by using the `--pyi_out` argument to `protoc`.

A new attribute for `py_proto_library` to enable/disable the generation of `.pyi` files and a way to reference them for packaging/editor support.

Switching to https://github.com/grpc/grpc (which has added support here: https://github.com/grpc/grpc/pull/32872), but we don't use grpc and would prefer to stick to primary bazel rules.

### LINKED ISSUE #1438: feat: support toolchainzation in py_proto_library
See: https://github.com/bazelbuild/rules_proto/issues/179

### LINKED ISSUE #1597: py_proto_library messes up sibling subpackage imports with generated proto source
# 🐞 bug report

### Affected Rule

 The issue is caused by the rule: py_proto_library

### Is this a regression?

 

No, but did work in com_google_protobuf's `py_proto_library` rule (which we are trying to migrate from).

### Description

 

Note: This seems somewhat related to https://github.com/bazelbuild/rules_python/issues/1551.

`proto_library` requires that the proto source be in the same package as the target. If we need to depend on an external repo's `.proto` file and create our own `proto_library` + `py_proto_library` targets, the workaround is to use a `genrule`. 

Something about the generated source triggers the "_virtual_imports", even though https://github.com/bazelbuild/rules_python/issues/1551#issuecomment-1804831583 seems to indicate this means we are definitively dealing with `strip_import_prefix`. 

Unlike with `strip_import_prefix`, in this case, prepending the `_virtual_imports/<proto_library_name>` to the path causes other normal Python libraries along the path to fail. It seems the `_virtual_imports` should not be triggered for generated sources at all (this was the case for com_google_protobuf's version) or the `imports` needs to be updated to properly handle this case.

In the repro below, the generated protobuf Python library is at: `bazel-bin/my_project/proto/_virtual_imports/foo_proto/my_project/proto/foo_pb2.py`.

`sys.path` contains `<execroot>/.../bin/my_project/proto/proto_fail.runfiles/<workspace_name>/my_project/proto/_virtual_imports/foo_proto` before `<execroot>/.../bin/my_project/proto/proto_fail.runfiles/<workspace_name>`.

But `_virtual_imports/foo_proto/` replicates the repo hierarchy to the proto target, i.e. it contains `my_project/proto`. This means that imports of other packages under `my_project` are picked from the former path entry and fail.

## 🔬 Minimal Reproduction

my_project/proto/BUILD
```
genrule(
    name = "generated_proto",
    srcs = ["@external_repo//path/to:foo.proto"],
    outs = ["foo.proto"],
    cmd = "cp $< $@",
)

proto_library(
    name = "foo_proto",
    srcs = ["foo.proto"],
)

py_proto_library(
    name = "foo_py_pb2",
    deps = [":foo_proto"],
)

py_binary(
    name = "proto_fail",
    srcs = [
        "proto_fail.py",
    ],
    deps = [
        "//my_project/api:bar,
        ":vertipro_py_pb2",
    ],
)
```

my_project/proto/proto_fail.py
```
from my_project.api import bar
from my_project.proto import foo_pb2
```

## 🔥 Exception or Error

<pre><code>

$ bazel run //my_project/proto:proto_fail

bin/my_project/proto/proto_fail.runfiles/.../my_project/proto/proto_fail.py", line 1, in <module>
    from my_project.api import bar
ModuleNotFoundError: No module named 'my_project.api'

</code></pre>

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
Debian Linux 6.5.6-1
  </code>
</pre>

<pre>
  <code>
bazel 6.4.0
  </code>
</pre>

**Rules_python version:**

<pre>
  <code>
rules_python-0.27.1
  </code>
</pre>

**Anything else relevant?**

### LINKED ISSUE #1703: [Gazelle][proto] How to automatically resolve protobuf target
# 🐞 bug report

### Affected Rule

gazelle resolve

### Is this a regression?
No

### Description

I have a python file with a proto definition imported. When I try to use gazelle to auto generate py_library target, it fails to validate dependency. I look into some code in bazel_gazelle/resolve/index.go, no proto target defined in there.

## 🔬 Minimal Reproduction

<pre>
  <code>
import a.b.c_pb2
  </code>
</pre>

and then run bazel run :py_gazelle

## 🔥 Exception or Error

<pre>
  <code>
gazelle: ERROR: failed to validate dependencies for target "//:check": "a.b.c_pb2" at line 11 from "check.py" is an invalid dependency: possible solutions:
	3. Ignore it with a comment '# gazelle:ignore a.b.c_pb2' in the Python file.

"a.b" at line 11 from "check.py" is an invalid dependency: possible solutions:
	3. Ignore it with a comment '# gazelle:ignore a.b' in the Python file.

"a" at line 11 from "check.py" is an invalid dependency: possible solutions:
	3. Ignore it with a comment '# gazelle:ignore a' in the Python file.

  </code>
</pre>

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
ubuntu 20.04
  </code>
</pre>

<pre>
  <code>
Bazelisk version: v1.16.0
Build label: 6.2.1
Build time: Fri Jun 2 16:59:58 2023 (1685725198)
Build timestamp: 1685725198
Build timestamp as int: 1685725198
  </code>
</pre>

**Rules_python version:**

<pre>
  <code>
0.25
  </code>
</pre>

**Anything else relevant?**
rules_python_gazelle_plugin version
0.28

### LINKED ISSUE #1707: Spurious "directory does not exist" warnings when building py_proto_library with external proto_library
# 🐞 bug report

### Affected Rule

The issue is caused by the rule:  `py_proto_library`

### Is this a regression?

I think some combinations of protobuf and rules_python didn't print this message in the past, but I'm not sure.

### Description

When creating a `py_proto_library` rule that depends on a well-known type in the `protobuf` repo, building it will generate spurious warnings of the form

> warning: directory does not exist

## 🔬 Minimal Reproduction

Create a trivial Bazel workspace

- MODULE.bazel:

  ```py
  bazel_dep(name = "protobuf", version = "23.1")
  bazel_dep(name = "rules_python", version = "0.28.0")
  ```

- BUILD:

  ```py
  load("@rules_python//python:proto.bzl", "py_proto_library")
  
  py_proto_library(
      name = "any_py_proto",
      deps = ["@protobuf//:any_proto"],
  )
  ```

## 🔥 Exception or Error

Output of `bazel build :any_py_proto`:

```
[...]
INFO: From ProtoCompile external/protobuf~23.1/python/google/protobuf/empty_pb2.py:
external/protobuf~23.1/.: warning: directory does not exist.
INFO: From ProtoCompile external/protobuf~23.1/python/google/protobuf/source_context_pb2.py:
external/protobuf~23.1/.: warning: directory does not exist.
INFO: From ProtoCompile external/protobuf~23.1/python/google/protobuf/any_pb2.py:
external/protobuf~23.1/.: warning: directory does not exist.
INFO: From ProtoCompile external/protobuf~23.1/python/google/protobuf/duration_pb2.py:
external/protobuf~23.1/.: warning: directory does not exist.
INFO: From ProtoCompile external/protobuf~23.1/python/google/protobuf/api_pb2.py:
external/protobuf~23.1/.: warning: directory does not exist.
INFO: From ProtoCompile external/protobuf~23.1/python/google/protobuf/field_mask_pb2.py:
external/protobuf~23.1/.: warning: directory does not exist.
INFO: From ProtoCompile external/protobuf~23.1/python/google/protobuf/struct_pb2.py:
external/protobuf~23.1/.: warning: directory does not exist.
INFO: From ProtoCompile external/protobuf~23.1/python/google/protobuf/timestamp_pb2.py:
external/protobuf~23.1/.: warning: directory does not exist.
INFO: From ProtoCompile external/protobuf~23.1/python/google/protobuf/type_pb2.py:
external/protobuf~23.1/.: warning: directory does not exist.
INFO: From ProtoCompile external/protobuf~23.1/python/google/protobuf/wrappers_pb2.py:
external/protobuf~23.1/.: warning: directory does not exist.
INFO: From ProtoCompile external/protobuf~23.1/python/google/protobuf/compiler/plugin_pb2.py:
external/protobuf~23.1/.: warning: directory does not exist.
INFO: From ProtoCompile external/protobuf~23.1/python/google/protobuf/descriptor_pb2.py:
external/protobuf~23.1/.: warning: directory does not exist.
[...]
```

## 🌍  Your Environment

macOS Sonoma

```
Build label: 7.0.1
Build time: Thu Jan 18 18:05:58 2024 (1705601158)
Build timestamp: 1705601158
Build timestamp as int: 1705601158
```

**Rules_python version:**

0.28

**Anything else relevant?**

Not sure whether this is an issue with rules_python, protobuf, or Bazel itself. https://github.com/bazelbuild/bazel/issues/7157 and https://github.com/envoyproxy/envoy/issues/28479 seem related.

### LINKED ISSUE #1924: Relax Cannot generate Python code for a .proto whose path contains '-' check
With bzlmod many repos now use their github repo name as the module name, instead of the previous com_github_whatever convention. Because of this you might end up with dashes in a name, and no matter what `repo_name` you give it in your MODULE.bazel the underlying path will be the canonical module name. Because of this if you have a repo like `protoc-gen-validate` and try to use a python proto from it, it fails when you include it as a third party dependency, but not when you build in that repo directly with this error:

https://github.com/bazelbuild/rules_python/blob/6ca2f58bedca519f5ea6431ad7e11aaac84415c0/python/private/proto/py_proto_library.bzl#L61-L64

Would it be possible for the check to only apply to the relative path in the repo, excluding the repo name itself? I don't believe the repo name should appear in the import path, so I think that would be safe?

### LINKED ISSUE #1935: Unable to get bazel 7 py_proto_library example working outside of rules_python dir
# 🐞 bug report

### Affected Rule

Bazel 7.1.2 + rules_python 0.32.2 seems to be unable to import python proto libraries

### Description
I've created a minimal repro here https://github.com/Ryang20718/bazel_rules_python_proto/tree/main

```bash
git clone https://github.com/Ryang20718/bazel_rules_python_proto.git
cd bazel_rules_python_proto
USE_BAZEL_VERSION=7.1.2 bazelisk test //foxglove:all --noenable_bzlmod --announce_rc -s --sandbox_debug
```

will result in failures. however, the same exact directory structure inside the example folder of `rules_python` will succeed

```bash
git clone https://github.com/Ryang20718/bazel_rules_python_proto.git
cd bazel_rules_python_proto/rules_python_working/examples/py_proto_library
USE_BAZEL_VERSION=7.1.2 bazelisk test //foxglove:all --noenable_bzlmod --announce_rc -s --sandbox_debug
```

I've been staring at differences for days, but I'm unsure of what could possibly be the difference?

## 🌍  Your Environment

**Operating System:**

ubuntu 20.04

7.1.2
**Rules_python version:**

0.32.2

comius:
Protobuf team is taking ownership of `py_proto_library` and the implementation was moved to protobuf repository.

Remove py_proto_library from rules_python, to prevent divergent implementations.

Make a redirect with a deprecation warning, so that this doesn't break any users.

Expected side effect of this change is also that the protobuf version is sufficiently updated that there is no more use of legacy struct providers.

Closes #1935 
Closes #1924 
Closes #1925 
Closes #1703 
Closes #1707 
Closes #1597 
Closes #1293 
Closes #1080
Fixes #1438


examples/bzlmod/py_proto_library/example.com/proto/BUILD.bazel
```diff
@@ -1,5 +1,5 @@
-load("@rules_proto//proto:defs.bzl", "proto_library")
-load("@rules_python//python:proto.bzl", "py_proto_library")
+load("@com_google_protobuf//bazel:proto_library.bzl", "proto_library")
+load("@com_google_protobuf//bazel:py_proto_library.bzl", "py_proto_library")
```
rickeylev, line 2: The python rules are in the protobuf repo directly, not in a python-specific repo? I thought I remember seeing a BCR PR where they were adding various proto_{java,python,etc} repos so e.g. java wasn't pulled in when only python was used.

internal_setup.bzl
```diff
@@ -35,9 +34,6 @@ def rules_python_internal_setup():
 
     bazel_skylib_workspace()
 
-    rules_proto_dependencies()
```
rickeylev, line 38: For workspace builds, shouldn't there be some sort of proto-setup call still? Did I miss where it is?

rickeylev (commented):
Please also add an entry to CHANGELOG.md to note that the rules_python proto rules are deprecated and will be removed in a future release.

CI has a failure when compiling protos. My guess would be the those builds don't have the additional compiler flags that were added. But, this would imply that all users also have to add those flags? That doesn't seem feasible. Can protobuf add those compile flags instead itself?

internal_setup.bzl
comius: There's a `protobuf_deps` call just after. We only need to setup protobuf, now that the rules are moving there.

examples/bzlmod/py_proto_library/example.com/proto/BUILD.bazel
comius: > The python rules are in the protobuf repo directly, not in a python-specific repo?
Yes.

> BCR PR where they were adding various proto_{java,python,etc} repos 

Hm, do you mean `rules_proto_grpc_*`? I imagine those support gRPC, but I wasn't part of that effort.

aignas:
I've referenced a lot of issues that have to do with proto library rule, feel free to disagree and alter the list if some of them are out of place.

aignas (approved):

comius:
> Please also add an entry to CHANGELOG.md to note that the rules_python proto rules are deprecated and will be removed in a future release.

Done.

> CI has a failure when compiling protos. My guess would be the those builds don't have the additional compiler flags that were added. But, this would imply that all users also have to add those flags? That doesn't seem feasible. Can protobuf add those compile flags instead itself?

I see 3 possible options:
- adding the flags
- fixing the code so that there are no problems
- using precompiled binaries

The last option is something that will be done in protobuf. It was done previously in rules_proto, but in an awkward and unmaintained way.

aignas:
@comius, when are you going to work on using the precompiled binaries? I wonder if we should merge this PR after that part is fixed.

comius:
> @comius, when are you going to work on using the precompiled binaries? I wonder if we should merge this PR after that part is fixed.

I hope we don't have to wait, because this PR also removes dependency on extra old protobuf that uses legacy struct providers. I'd like to disable that support in Bazel and that would then break rules_python.

Old py_proto_library is getting precompiled protoc from a really old version of rules_proto, which was never properly maintained.

alexeagle:
Note that due to https://github.com/protocolbuffers/protobuf/issues/17075 this caused a regression - the latest rules_python release has a `py_proto_library` that can be used with a prebuilt toolchain to avoid compiling protoc, but the one in protocolbuffers/protobuf does not. Thus users with a non-functional cc toolchain will be stuck on the last rules_python release.

rickeylev:
@comius FYI we're going to revert this PR for now.

The main reason is what @alexeagle is reporting -- that users will be stuck on a release without a way to upgrade. Breaking changes are OK, but have to allow a graceful upgrade path; see https://rules-python.readthedocs.io/en/latest/contributing.html#breaking-changes

From the issue Alex filed, it sounds like part of the problem here is sequencing of releases? That this PR to be released after something else in proto-something is available?

That said, there are several other changes in this PR I have questions about:

* It looks like building of doc-related protos was disabled in CI, with a comment saying protobuf compilation fails? Our docgen relies on protos, though.
* The py_proto_library bzlmod example was also disabled -- this is partially a test that it works under bzlmod. If its broken, then something seems very wrong.
* You said fixing the need to add the copt flags would block removing legacy struct providers? I don't see the connection.

comius:
> @comius FYI we're going to revert this PR for now.
> 
> The main reason is what @alexeagle is reporting -- that users will be stuck on a release without a way to upgrade. Breaking changes are OK, but have to allow a graceful upgrade path; see https://rules-python.readthedocs.io/en/latest/contributing.html#breaking-changes
> 
> From the issue Alex filed, it sounds like part of the problem here is sequencing of releases? That this PR to be released after something else in proto-something is available?
> 
> That said, there are several other changes in this PR I have questions about:
> 
> * It looks like building of doc-related protos was disabled in CI, with a comment saying protobuf compilation fails? Our docgen relies on protos, though.

It was disabled only on some platforms with GCC. Not on all of them.

> * The py_proto_library bzlmod example was also disabled -- this is partially a test that it works under bzlmod. If its broken, then something seems very wrong.

Same here.

> * You said fixing the need to add the copt flags would block removing legacy struct providers? I don't see the connection.

Try running rules_python build with --incompatible_disallow_struct_provider_syntax and the connection will become clear. Struct providers are used in an extra old protoc, that's comping from rules_proto.

alexeagle:
I'll try to clarify:
I've been working on the rollout of Ivo's "toolchainization" of proto tools (the `protoc` and any language-specific plugins. We have to work through a large number of hard-coded `@com_google_protobuf//:protoc` references (to a `cc_binary`) which are listed here https://github.com/bazelbuild/rules_proto/discussions/213

We announced that Python is working, but the fork of `py_proto_library` to the protobuf repo occurred at an earlier commit and lost the fix. My fix for that has just landed: https://github.com/protocolbuffers/protobuf/pull/17078

With my devrel hat on, I'd like to see that fix released by protobuf before rules_python drops the rule so this feature doesn't temporarily regress - it's hard to explain to users "yeah it was fixed, but then it was unfixed, and you're on this one release of rules_python in that range, sorry footgun..."

rickeylev:
Thanks @alexeagle, @comius 

So if I understand correctly, we just have to wait for the next protobuf release, and then can re-merge this change?

Will the protobuf fix also address having to add the copt flag and the various CI config changes that were necessary?

Is that it, or should something also be done in the meantime? (opt-in env var? bump rules_proto or protobuf version past 21 but before 27? idk).

alexeagle:
https://github.com/aspect-build/toolchains_protoc/pull/13 demonstrates the breakage caused by the PR, for users who don't have cc toolchain.

> just have to wait for the next protobuf release, and then can re-merge this change

Yes, next protobuf release fixes the `py_proto_library` rule so I that's what I would prefer. I know @comius is now blocked on some other changes downstream of this one, so I wont object to a roll-forward as soon as you like.

> having to add the copt flag and the various CI config changes

I don't have context on that, and don't care about their compiler options for building `protoc` from source since I don't do that. It seems like they break it as often as they fix it.

aignas:
Created a new PR and added all of you as reviewers: #1951.

PR #1837, opened 2024-04-08
feat(bzlmod): support cross-platform whl setups within the main hub repo

aignas:
With this change we add support for platform-specific wheel registration and
doing the selection of which wheel is used at build time.

This supports:
- Different package versions for different platforms.
- Use string_flags to configure what to fetch/select:
    - only whls, only sdist or auto mode.
    - libc version and `musl` vs `glibc` selection.
    - universal2 vs arch wheels for mac.
    - target osx version selection.

Summary of changes:
- The `uv pip install` would only warn the user of yanked packages but
  would not refuse to install them. Update our implementation to better
  match the same behaviour.
- A previous PR has added the support for passing it in the
  `requirements_by_platform` and this just add the necessary code to make
  sure that we can also do the dependency management when parsing the
  `whl` `METADATA` files.
- Only configure `dev_pip` deps for `linux` and `osx` platforms to not raise
  issues later.
- Add a function for generating a `whl_library` name from a `filename`.
- Add a macro for generating all config settings for a particular set of parameters.
- Update `render_pkg_aliases` to also use those config settings.
- Update the toolchain selection `target_settings` to use the `py_linux_libc`
  config setting. With this the user can register a `musl` linux toolchain if
  needed. We can also use similar `flag_values` to resolve #1876.
- Integrate everything in the `pip` extension and setup cross-platform
  select statements.

Work towards #1357, #260

Stacked on #1937

PR #1937, opened 2024-06-02
fix(pip.parse): make the pip extension reproducible if PyPI is not called

### LINKED ISSUE #1643: Make MODULE.bazel.lock file have the same contents no matter the platform bazel is executed on

aignas:
With this PR we can finally have fewer lock file entries in setups which
do not use the `experimental_index_url` feature yet. This is fully
backwards compatible change as it relies on `bazel` doing the right
thing and regenerating the lock file.

Fixes #1643.

### DIFF
--- a/python/private/bzlmod/pip.bzl
+++ b/python/private/bzlmod/pip.bzl
@@ -482,13 +482,20 @@ def _pip_impl(module_ctx):
 def _pip_non_reproducible(module_ctx):
     _pip_impl(module_ctx)
 
-    if bazel_features.external_deps.extension_metadata_has_reproducible:
-        # We allow for calling the PyPI index and that will go into the MODULE.bazel.lock file
-        return module_ctx.extension_metadata(reproducible = False)
-    else:
-        return None
+    # We default to calling the PyPI index and that will go into the
+    # MODULE.bazel.lock file, hence return nothing here.
+    return None
 
-def _pip_parse_ext_attrs():
+def _pip_parse_ext_attrs(**kwargs):
+    """Get the attributes for the pip extension.
+
+    Args:
+        **kwargs: A kwarg for setting defaults for the specific attributes. The
+        key is expected to be the same as the attribute key.
+
+    Returns:
+        A dict of attributes.
+    """
     attrs = dict({
         "experimental_extra_index_urls": attr.string_list(
             doc = """\
@@ -503,6 +510,7 @@ This is equivalent to `--extra-index-urls` `pip` option.
             default = [],
         ),
         "experimental_index_url": attr.string(
+            default = kwargs.get("experimental_index_url", ""),
             doc = """\
 The index URL to use for downloading wheels using bazel downloader. This value is going
 to be subject to `envsubst` substitutions if necessary.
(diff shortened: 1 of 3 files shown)


aignas:
FYI @fmeum

python/private/bzlmod/pip.bzl
```diff
@@ -462,6 +468,26 @@ def _pip_impl(module_ctx):
...
+        return module_ctx.extension_metadata(reproducible = is_extension_reproducible)
+    else:
+        return None
+
+def _pip_non_reproducible(module_ctx):
+    _pip_impl(module_ctx)
+
+    if bazel_features.external_deps.extension_metadata_has_reproducible:
+        # We allow for calling the PyPI index and that will go into the MODULE.bazel.lock file
+        return module_ctx.extension_metadata(reproducible = False)
```
fmeum, line 487: This is equivalent to not returning metadata at all, so you could remove the entire block.

fmeum, line 487: This is equivalent to not returning metadata at all, so you could remove the entire block.

fmeum, line 487: This is equivalent to not returning metadata at all, so you could remove the entire block.

python/private/bzlmod/pip.bzl
```diff
@@ -714,7 +729,54 @@ extension.
 """,
         ),
     },
-    **_extension_extra_args()
+)
+
+pip_internal = module_extension(
```
fmeum, line 734: This needs to declare os/arch dependence.

python/private/bzlmod/pip.bzl
aignas: Good point, will do.

python/private/bzlmod/pip.bzl
aignas: Not necessarily, since we are calling PyPI to get the right dependencies, the lock file will be the same on all platforms, so this can be removed, I'll change it so that it is the default behaviour.

fmeum: Nice!

groodt (approved):
LGTM

fmeum:
@aignas Would it be possible to get a release soon after this is merged? Bazel itself suffers from the per-os lockfile and having the extension marked as reproducible would greatly simplify the process of updating Bazel to build with Bazel 7.2.0.

aignas:
I don't have time today, and I would like to get #1837 into the release as well. If all goes well, the earliest I can do maybe tomorrow (or early next week) and Richard is on vacation this week. So early next week is probably a good worst case scenario.

fmeum:
No hurry, any time next week would be perfect!

PR #1946, opened 2024-06-10
fix: make the pythons_hub platform independent

### LINKED ISSUE #1643: Make MODULE.bazel.lock file have the same contents no matter the platform bazel is executed on

aignas:
With this change we can finally have platform independent bzlmod lock
files in rules_python. This cleans up code paths that were not being
used in the `pip` extension for a while and the replacement is noted in
the CHANGELOG.

Fixes #1643

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,7 +51,7 @@ A brief description of the categories of changes:
   which is an implementation detail which should not be relied on and is there
   purely for better debugging experience.
 * (bzlmod) The `pythons_hub//:interpreters.bzl` no longer has platform-specific
-  labels where left there for compatibility reasons. Move to
+  labels which where left there for compatibility reasons. Move to
   `python_{version}_host` keys if you would like to have access to a Python
   interpreter that can be used in a repository rule context.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,10 @@ A brief description of the categories of changes:
   `{pip_hub_prefix}_{wheel_name}_{py_tag}_{abi_tag}_{platform_tag}_{sha256}`,
   which is an implementation detail which should not be relied on and is there
   purely for better debugging experience.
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -50,6 +50,10 @@ A brief description of the categories of changes:
   `{pip_hub_prefix}_{wheel_name}_{py_tag}_{abi_tag}_{platform_tag}_{sha256}`,
   which is an implementation detail which should not be relied on and is there
   purely for better debugging experience.
+* (bzlmod) The `pythons_hub//:interpreters.bzl` no longer has platform-specific
+  labels where left there for compatibility reasons. Move to
```
aignas, line 54:
```suggestion
  labels which where left there for compatibility reasons. Move to
```

PR #1952, opened 2024-06-12
chore: access ctx.version_file to inform Bazel stamping info is needed

rickeylev:
This enables an optimization in Bazel et al to better track when targets actually need stamping information. Accessing the `ctx.version_file` tells Bazel the target depends on stamping information, which related systems can use to optimize processing of the build graph.


aignas (approved):
Is there anything useful for the public changelog?

rickeylev:
> Is there anything useful for the public changelog?

No, not really. This optimization is for RBE-type stuff that has a deeper integration with Bazel than is typical. I don't think it affects local builds.

PR #1953, opened 2024-06-12
fix(rules): drop the unused argument

### LINKED ISSUE #1954: py_binary zip file on non-Windows is broken
Introduced by #1929.
See https://github.com/bazelbuild/bazel/pull/22718

Created #1953 to fix this and this ticket should be closed once we have tests for this behaviour.

aignas:
It seems that there is an extra argument that is there but I am not sure about
the reason, it is most likely a leftover. As part of this change we add an
analysis test for non-windows zip building. I could reproduce the failure and
now the newly added test passes.

Fixes #1954

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,14 @@ A brief description of the categories of changes:
 ### Removed
 * Nothing yet
 
+## [0.33.1] - 2024-06-13
+
+[0.33.1]: https://github.com/bazelbuild/rules_python/releases/tag/0.33.1
+
+### Fixed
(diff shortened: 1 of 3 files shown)


aignas:
FYI @fmeum

aignas:
Once this is merged, we will need to cherry-pick and release a 0.33.1, IMHO.

aignas:
@rickeylev, feel free to take over the PR and finish it. Since I have started it, if you stamp after making any extra changes, you should be able to merge it. :)

rickeylev (approved):
Oops, thanks. Yeah, I missed that arg and there weren't tests covering it. Fix LGTM.

PR #1960, opened 2024-06-14
ci: add a custom dependabot file

aignas:
This way the requirements files in the examples and tests are not
managed, which should reduce the PR noise.


rickeylev (approved):
Oh thank you, all those dependabots were annoying.

PR #1973, opened 2024-06-15
fix(bzlmod): use same target in requirement('foo') and all_requirements

jvolkman:
We have at least one instance in our pre-bzlmod workspace like:
```
deps = list(all_requirements)
deps.remove(requirement('foo'))
```

But when using bzlmod, the generated `all_requirements` list contains targets like `@@rules_python~~pip~my_deps//sqlalchemy:sqlalchemy`, whereas `requirement('sqlalchemy')` will return `@@rules_python~~pip~my_deps//sqlalchemy:pkg`. So this operation now fails.

This change makes `all_requirements` also use the `:pkg` targets to match `requirement`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,7 @@ A brief description of the categories of changes:
 * `protobuf`/`com_google_protobuf` dependency bumped to `v24.4`
 
 ### Fixed
-* Nothing yet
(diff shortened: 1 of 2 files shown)


aignas (approved):
LGTM, thanks!

PR #1974, opened 2024-06-16
refactor(bzlmod): generate fewer targets for pip_config_settings

aignas:
Remove the `python_version` from `flag_values` and reduces the number of
targets we create by a lot because we no longer need to create all of the micro
version combinations to also match when we want to match e.g. 3.10 on a
particular platform by using a trick I've learned in #1968:
```starlark
select({
    "is_python_3.x": "my_config_setting"
    "//conditions:default": "is_python_3.x"
})
```

Then further optimization was done on how we create the aliases and config
settings and optimizing the usage of `pip_whl` flag finally reduced the
internal number targets. The number config settings targets in
`//tests/private/pip_config_settings/...` where we have multiple python
versions has changed:

* `:is_*` - from **995** to **996** (the extra addition being is_python_default).
* `:_.*is_*` - from **28272** to **2480** (just python version) and then to **496** (final).

Work towards #260

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ A brief description of the categories of changes:
   reduce the total number of targets in the hub repo.
 
 ### Fixed
+* Nothing yet
 
 ### Removed
 * Nothing yet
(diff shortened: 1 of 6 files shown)


groodt (approved):
LGTM

PR #1976, opened 2024-06-17
feat(gazelle): Add directives for label format & normalisation 

### LINKED ISSUE #1939: Gazelle: Sanitize distribution names in modules mapping / manifest, not at runtime
# 🚀 feature request

### Relevant Rules

* `modules_mapping`
* `gazelle_python_manifest`
* `gazelle`

### Description

Gazelle plugin uses `gazelle_python.yaml` (`gazelle:python_manifest_file_name`) to derive `deps` for py targets. One might expect that a dependency label will be `${manifest.pip_repository}//${manifest.modules_mapping.some.import}`, however, Gazelle does extra sanitisation right before using the mapping value [here](https://github.com/bazelbuild/rules_python/blob/ae1e1a014d88a9862c232daafcb6b1025b2928ae/gazelle/pythonconfig/pythonconfig.go#L266).

Although it is not a big deal when used with `rules_python`, it makes harder to reuse Gazelle plugin with other python rules, more specifically [rules_pycross](https://github.com/jvolkman/rules_pycross), which has a different target naming convention, which is aligned with [Python normalised names](https://packaging.python.org/en/latest/specifications/name-normalization/). In [this ticket](https://github.com/jvolkman/rules_pycross/issues/100), there is an approach (or hack really) described for patching `gazelle_python.yaml` but further sanitisation that happens inside the plugin itself makes it impossible to use.

I do understand that compatibility with other rules most likely isn't of highest priority, here are some other reasons:
* It would be slightly less confusing when target name derivation (normalisation / alignment) happens in once place, rather than in `modules_mapping` and/or `gazelle_python_manifest` and/or plugin itself.
* Given that there are ongoing efforts to #1360, it might be a good improvement going forward.

* Do not `SanitizeDistribution` in `pythonconfig.Config#FindThirdPartyDependency`, take the value as is.
* Do `SanitizeDistribution` in `modules_mapping`, or `gazelle_python_manifest`. I'd say `modules_mapping` is best but please advice.

wingsofovnia:
Adds new directives to alter default Gazelle label format to third-party dependencies useful for re-using Gazelle plugin with other rules, including `rules_pycross`.

Fixes #1939


.bazelrc
```diff
@@ -4,8 +4,8 @@
...
...example.com/another_proto,examples/bzlmod/py_proto_library/example.com/proto,examples/bzlmod/runfiles,examples/bzlmod/tests,examples/bzlmod/tests/other_module,examples/bzlmod/whl_mods,examples/bzlmod_build_file_generation,examples/bzlmod_build_file_generation/other_module/other_module/pkg,examples/bzlmod_build_file_generation/runfiles,examples/multi_python_versions/libs/my_lib,examples/multi_python_versions/requirements,examples/multi_python_versions/tests,examples/pip_parse,examples/pip_parse_vendored,examples/pip_repository_annotations,examples/py_proto_library,examples/py_proto_library/example.com/another_proto,examples/py_proto_library/example.com/proto,gazelle,gazelle/manifest,gazelle/manifest/generate,gazelle/manifest/hasher,gazelle/manifest/test,gazelle/modules_mapping,gazelle/python,gazelle/python/private,gazelle/pythonconfig,tests/integration/compile_pip_requirements,tests/integration/compile_pip_requirements_test_from_external_repo,tests/integration/ignore_root_user_error,tests/integration/ignore_root_user_error/submodule,tests/integration/pip_parse,tests/integration/pip_parse/empty,tests/integration/pip_repository_entry_points,tests/integration/py_cc_toolchain_registered
```
wingsofovnia, line 8: Side effect of `pre-commit run --all-files`. Should I leave it?

gazelle/pythonconfig/pythonconfig.go
```diff
@@ -443,3 +460,46 @@ func (c *Config) SetTestFilePattern(patterns []string) {
...
+// FormatThirdPartyDependency returns a label to a third-party dependency performing all formating and normalization.
+func (c *Config) FormatThirdPartyDependency(repositoryName string, distributionName string) label.Label {
+	conventionalDistributionName := strings.ReplaceAll(c.labelConvention, distributionNameLabelConventionSubstitution, distributionName)
+
+	var normConventionalDistributionName string
+	switch norm := c.LabelNormalization(); norm {
+	case SnakeCaseLabelNormalizationType:
+		normConventionalDistributionName = strings.ToLower(conventionalDistributionName)
+		normConventionalDistributionName = strings.ReplaceAll(normConventionalDistributionName, "-", "_")
+		normConventionalDistributionName = strings.ReplaceAll(normConventionalDistributionName, ".", "_")
```
aignas, line 493: we also need to replace any occurrences of `__` to `_`. This is done in our starlark code, so it is essentially the same as `pep503` but with `_`, so doing the same regex as below would be ideal.

aignas (commented):
LGTM, only the `snake_case` to be modified to be more like pep503 left.

gazelle/pythonconfig/pythonconfig.go
wingsofovnia: This piece was moved exactly from `SanitizeDistribution` to make sure nothing breaks existing behaviour. Could you please point to the Starlark code you had in mind? 

I found [this](https://github.com/bazelbuild/rules_python/blob/49d180fc773da54df63386c64790dbf8b7616735/python/private/py_wheel.bzl#L229-L264), and from what my internal 🧠 interpreter tells me it escapes char by char, so `distribution----with--dashes` becomes `distribution____with__dashes`, not `distribution_with_dashes` if I replace `"[-_.]+"` with `_` as in `pep503`.

aignas: The rules_python//python/private:normalize_name.bzl is the code that I had in mind.

wingsofovnia: Thank you. I updated the code and actually found extra corner case where `normalize_name.bz` behaves differently than `SnakeCaseLabelNormalizationType` and the old code: prefix/trailing dashes. This has been fixed, the mapping updated, more tests added.

aignas (approved):
Thank you!

PR #1982, opened 2024-06-17
refactor(internal): allow setting linking mode for internal cc_details APIs

rickeylev:
This upstreams a Google patch to control how py_binary links C++ dependencies. This just
extends the internal helper functions a bit to allow specifying some of the otherwise
hard coded values and passing in additional args to the cc_details struct.


aignas (approved):
LGTM.

oprypin (approved):

PR #1986, opened 2024-06-18
fix: enable auto exec groups using attribute

rickeylev:
This allows the precompile action to resolve its tools separately from other toolchains,
allowing toolchain types with otherwise incompatible `exec_compatible_with` definitions
to be used together.


aignas (approved):
LGTM, Does it fix #1967? If yes, then it is a really nice alternative to #1968, which should be closed if this PR is merged.

rickeylev:
It does not. It was explained to me that, when AEGs are enabled, all toolchains are still resolved, it's just that each performs its exec platform selection separately instead of trying to find a common exec platform to satisfy all of them.

PR #1987, opened 2024-06-18
refactor!: Remove entrypoint

groodt:
Removes the entrypoint macro. This has been replaced with `py_console_script_binary`.

I was getting familiar again with the code-base today and remembered that we were on a path to deprecate `entrypoint` in favor of `py_console_script_binary`. So I did a very quick clean-up. 

It has been mentioned for a while that it would be removed, so I *think* removal is valid now from a compatibility standpoint: https://github.com/bazelbuild/rules_python/pull/1987/files#diff-fed3a6d7d568aea2237b65544571ef9519b5f249ddbb23d521e6e55161c10452L33

Either way, we probably don't want to carry this into the 1.0


aignas (commented):
Thanks for the cleanup. Could you please add a note in CHANGELOG.md about breaking changes.

aignas (approved):

PR #1991, opened 2024-06-18
chore: use `**` globs instead of explcit BUILD.bazel reference in distribution filegroups

rickeylev:
This is to make Google patching and imports of rules_python easier. Within Google, `BUILD` is the file name instead of `BUILD.bazel`. It's also easy for the distribution filegroups to accidentally miss files, so to prevent that, just glob everything.


oprypin (approved):

groodt (approved):
LGTM

PR #1998, opened 2024-06-20
fix: Add `configure=True` for internal_config_repo

meteorcloudy:
Mark internal_config_repo as `configure = True` so that it won't be vendored by Bazel vendor mode.

https://bazel.build/rules/lib/globals/bzl#parameters_7

### DIFF
--- a/python/private/internal_config_repo.bzl
+++ b/python/private/internal_config_repo.bzl
@@ -92,6 +92,7 @@ def _internal_config_repo_impl(rctx):
 
 internal_config_repo = repository_rule(
     implementation = _internal_config_repo_impl,
+    configure = True,
     environ = [_ENABLE_PYSTAR_ENVVAR_NAME],
 )
 


meteorcloudy:
Not sure if we should also do it for `rules_python~~python~pythons_hub`?

rickeylev (approved):

rickeylev:
Oh interesting, I haven't seen this vendor mode before.

Is there any guidance on when configure=True should be set or not? The gist I get is, if the repo inspects the local system, then it should be configure=True.

But, it seems like environment variables are part of "inspecting the local system" ? Does that mean any repo rule that consults environment variables shouldn't be configure=True? For this internal config repo, I think it makes sense to not vendor it; all it does is propagate an environment variable for later consumption by the rules.

A more complicated case that comes to mind are the pip repo rules. These represent remote dependencies, so vendoring seems like a good idea. However, iirc, there's several environment variables they might use.

meteorcloudy:
> Oh interesting, I haven't seen this vendor mode before.

Yes, this is a new feature we just implemented at HEAD and will be [backported](https://github.com/bazelbuild/bazel/pull/22794) to 7.3. See the documentation here: https://bazel.build/external/vendor

> Is there any guidance on when configure=True should be set or not? The gist I get is, if the repo inspects the local system, then it should be configure=True.

The `configure` attribute was introduced a long time ago at https://github.com/bazelbuild/bazel/commit/b7e8f8402a48994d9a6410e4cd567932ea54c8c0. I always felt it has some duplicated meaning with the `local` attribute. 

But they do behave differently:
- Repos with `local = True` are always re-fetched if the corresponding RepositoryDelegatorFunction is triggered in Skyframe (e.g after a server restart).
- Repos with `configure = True` behaves mostly like a normal repo rule, which is "cached" with the marker file. But it works with `bazel sync --confgure` or `bazel fetch --confgure` to only re-fetch configure repos.

I think they both mean to mark the repo rule as not hermetic, but `local = True` is more strict.

With the introduction of vendor mode, we made the decision to not vendor repos with `local = True` and `configure = True`.

I think in general, if a repo may produce different content in different environment and it doesn't access network (so no need to vendor for offline build), it should be marked as `configure = True` or `local = True`.

PR #1999, opened 2024-06-21
fix(multiplatform): Add i386 Linux support

nicbadiu:
Enable i386 Linux platform detection support.

Without this change `rules_python` complains about an unrecognized platform when run on a 32-bit platform, e.g. where `os.uname().machine` returns `i686`. This effectively makes `rules_python` unusable on 32-bit platforms.

### DIFF
--- a/python/versions.bzl
+++ b/python/versions.bzl
@@ -532,6 +532,17 @@ PLATFORMS = {
         os_name = LINUX_NAME,
         arch = "armv7",
     ),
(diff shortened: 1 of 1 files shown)


rickeylev:
What's the error you're seeing?

I'm curious -- how are you even getting a 32 bit interpreter for linux...? You must be using a pretty old version or building it yourself?

In any case, it should be fine to add an additional platform entry here.

rickeylev (approved):

PR #2000, opened 2024-06-21
internal: repos to create a toolchain from a locally installed Python

rickeylev:
This adds the primitives for defining a toolchain based on a locally installed Python.
Doing this consists of two parts:

* A repo rule to define a Python runtime pointing to a local Python installation.
* A repo rule to define toolchains for those runtimes.

The runtime repos create platform runtimes, i.e, it sets py_runtime.interpreter_path.
This means the runtime isn't included in the runfiles.

Note that these repo rules are largely implementation details, and are definitely not
stable API-wise. Creating public APIs to use them through WORKSPACE or bzlmod will
be done in a separate change (there's a few design and behavior questions to discuss).

This is definitely experimental quality. In particular, the code that tries
to figure out the C headers/libraries is very finicky. I couldn't find solid docs about
how to do this, and there's a lot of undocumented settings, so what's there is what
I was able to piece together from my laptop's behavior.

Misc other changes:
* Also fixes a bug if a pyenv-backed interpreter path is used for precompiling:
  pyenv uses `$0` to determine what to re-exec. The `:current_interpreter_executable`
  target used its own name, which pyenv didn't understand.
* The repo logger now also accepts a string. This should help prevent accidentally
  passing a string causing an error. It's also just a bit more convenient when
  doing development.
* Repo loggers will automatically include their rule name and repo name. This
  makes following logging output easier.
* Makes `repo_utils.execute()` report progress.
* Adds `repo_utils.getenv`, `repo_utils.watch`, and `repo_utils.watch_tree`:
  backwards compatibility functions for their `rctx` equivalents.
* Adds `repo_utils.which_unchecked`: calls `which`, but allows for failure.
* Adds `repo_utils.get_platforms_os_name()`: Returns the name used in `@platforms` for
  the OS reported by `rctx`.
* Makes several repo util functions call `watch()` or `getenv()`, if available. This
  makes repository rules better respect environmental changes.
* Adds more detail to the definition of an in-build vs platform runtime
* Adds a README for the integration tests directory. Setting up and using one is a bit
  more involved than other tests, so some docs help.
* Allows integration tests to specify bazel versions to use.


python/private/platform_runtime_repo.bzl
```diff
@@ -0,0 +1,196 @@
...
+        ],
+        quiet = True,
+    )
+    rctx.delete("getinfo.py")
+    info = json.decode(output)
+
+    # The cc_library.includes values have to be non-absolute paths, otherwise
+    # the toolchain will give an error. Work around this error by making them
+    # appear as part of this repo.
+    rctx.symlink(info["include"], "include")
```
aignas, line 117: Yeah the symlink creation is good, but note that on Windows it will do a copy if the symlinks are not enabled. Maybe worth putting into a comment.

That said, this only works on non-windows, I guess.

python/private/platform_runtime_repo.bzl
```diff
@@ -0,0 +1,196 @@
...
+
+config_vars = [
+    "LDLIBRARY", "LIBDIR", "INSTSONAME", "LIBDEST", "PY3LIBRARY",
+    "SHLIB_SUFFIX",
+]
+data.update(zip(config_vars, sysconfig.get_config_vars(*config_vars)))
+print(json.dumps(data))
+"""
+
+def _platform_runtime_repo_impl(rctx):
```
aignas, line 96: nit, it would be great to gate this to only work on specific OSes, e.g. check at the top if we are running on Windows and bail if not.

python/private/platform_runtime_repo.bzl
rickeylev: A copy would be fine. It's copying the whole directory, right?

python/private/platform_runtime_repo.bzl
rickeylev: Aside from the symlink on the include dir, is there something else in here that won't work on windows?

Honestly, I have no idea how correct/accurate/portable the code for find the include dir etc is. I couldn't find much docs about this, so reverse engineers what I could find on SO and what I saw on my linux laptop and desktop. No idea how things look on a mac or windows.

aignas: We could probably use the CI what breaks. My thinking was to be on the safe side and enable only the platforms that are known to work. That way we can segment the problem into smaller chunks.

python/private/platform_runtime_repo.bzl
aignas: I think so.

rickeylev:
Alright, I mentioned this in chat, but I'll post a more full description of some of the problems a local auto configuring toolchain encounters.

Ideally, we want something like this:

```
# The user inputs something like this:
local_runtime(
  name = "linux_runtime",
  path = "/usr/bin/python3" # Turns out to be python 3.10
)
local_runtime(
  name = "windows_runtime",
  path = "C:\Program Files\whatever\python3.exe" # turns out to be python 3.11
)

# And it generates something like this:

toolchain(
  name = "linux_toolchain",
  toolchain = "@linux_runtime//:runtime",
  target_compatible_with = [linux],
  target_settings = [is_python_3.10],
)
toolchain(
  name = "windows_toolchain",
  toolchain = "@windows_runtime//:runtime"
  target_compatible_with = [windows],
  target_settings = [is_python_3.11],
)
```

Unfortunately, I don't see a way to do that without triggers the evaluation of the `@linux_runtime` and `@windows_runtime` repos. This is because of the constraint settings:

In order to know the TCW values, we have to...IDK. I think manually associate it with the local_runtime() call? But, that also seems wrong -- /usr/bin/python3 is valid for both linux and mac. So I guess transform rctx.os to a constraint sting (but we want to ignore rctx.os on linux if its a windows path)? Maybe the host constraint is the other option?

For the target settings, we have to run python in order to determine the Python value. But a linux host can't run the windows path and vice versa. So now...idk. Omit target_settings? Consider the whole toolchain incompatible?

But, lets say its the happy path -- we can run the given path, get the version, figure out the os, etc. This is all computed _within_ the runtime repo rule. If a different repo rule generates the toolchain() calls, that info needs to be passed in -- which means in order to access that info, it has to trigger evaluation of the runtime repo. This is true even if we do something like `toolchain(target_compatible_with=["@linux_runtime//:os_alias"])` (where os is an alias to the appropriate `@platforms` target) -- even though its a target reference, toolchain evaluation will have to resolve that target, and thus trigger evaluation of the repo.

The only paths I see to deal with all this are:
1. Just accept evaluation will occur. It should be relatively cheap? If a runtime can't be setup (e.g. linux trying to run a windows path), silently fail (create like an empty no-op runtime structure or something).
2. Require (or optionally allow; requiring seems to defeat the point of an auto-configuring thing) passing in platform and Python version info. If we want to get advanced, we can write the platform and version into a separate repo if we know it at repo-evaluation time (this allows resolving the constraints, while avoiding evaluating the runtime repo itself)
3. Make the `local_toolchain(...)` api have platform-specific arg stuff. e.g. `local_runtime(windows_path="C:\...", linux_path="...", ...)`; though I'm not sure how to encode python version. It seems like a full and complete set of args would be like a list of `(str path, list platform_constraints, list target_settings)` tuples or something.

aignas:
I like the approach 1. out of your listed ones. It should be cheap to just have:
```starlark
def _linux_impl(rctx):
    if not _is_linux(rctx):
        rctx.file("BUILD.bazel", """\
toolchain(
    name = "linux_toolchain",
    toolchain = "@linux_runtime//:runtime",
    target_compatible_with = ["@platforms//:incompatible"],
    target_settings = [],  # maybe we have to add an incompatible target setting here.
)"""
        return

    ...
```

Because the repository rule is not doing any network IO and the failure is
really fast, then I think we should be good here.

Talking about python versions, do you want this toolchain to be used when
`is_python_3.x` is true? What about the default case when the `python_version`
is unset? If we have `hermetic toolchains` also in the mix, which one is
preferred?

I think that if we could affect the default value of
`//python/config_settings:python_version` based on the result of this linux
toolchain or the `python` extension evaluation, it would be super nice, because
that could simplify a lot of code that we have that handles the case where the
`python_version` string flag is unset, but we have to default to something.

rickeylev:
> do you want this toolchain to be used when is_python_3.x is true?

Yes. We end up having to run python no matter what, so we have the version. Integrating it with the version-aware parts just requires adding the target_settings value. So it should be easy.

>  If we have hermetic toolchains also in the mix, which one is preferred?

By default, the local ones, because they're opt in. It's technically up to the user, though, since they ultimately can futz with the toolchain registration order.

rickeylev:
OK, ready for review.

I've been sitting on this awhile, so I pulled back the scope to just the repository rules that can set up the runtime and toolchains. It works, but you can see that there's a decent amount of configuration needed in MODULE.bazel. Integrating it into the python bzlmod extension can clean that up, but it got surprisingly messy. I'll post in chat about some of the design/behavior things we need to decide on. Using it in WORKSPACE is probably similar, but I haven't tried it with workspace builds yet.

python/private/local_runtime_repo_setup.bzl
```diff
@@ -0,0 +1,140 @@
...
+        flag_values = {
+            _PYTHON_VERSION_FLAG: major_minor,
+        },
+    )
+    native.config_setting(
+        name = "_is_major_minor_micro",
+        flag_values = {
+            _PYTHON_VERSION_FLAG: major_minor_micro,
+        },
+    )
```
aignas, line 132: As discussed over a call, it could be nice to add another flag to allow the user to disable the toolchain with a line in the `.bazelrc`:

```
build --@rules_python//python/config_settings:local_toolchain=disable
```

This could be done in a followup PR to keep this more scoped.

python/private/local_runtime_repo_setup.bzl
```diff
@@ -0,0 +1,140 @@
...
+
+    native.alias(
+        name = "os",
+        # Call Label() to force the string to evaluate in the context of
+        # rules_python, not the calling BUILD-file code. This is because
+        # the value is an `@platforms//foo` string, which @rules_python has
+        # visibility to, but the calling repo may not.
+        actual = Label(os),
+        visibility = ["//visibility:public"],
```
aignas, line 118: I think we might want to do something similar in the `pip_config_settings` macro, however, since the `@platforms` dep is a special one, it might be not that urgent, as the canonical label will be always `@platforms//foo` rather than `@platforms~~//foo` or something similar.

python/private/repo_utils.bzl
```diff
@@ -262,12 +341,50 @@ def _outputs_to_str(result):
             lines.append("<{} empty>".format(name))
     return "\n".join(lines)
 
+def _get_platforms_os_name(rctx):
```
aignas, line 344: Note to myself, maybe I should have a followup to use this in `select_requirement` function. Might be useful to deduplicate.

python/private/repo_utils.bzl
```diff
@@ -262,12 +341,50 @@ def _outputs_to_str(result):
...
+def _watch(rctx, *args, **kwargs):
+    """Calls rctx.watch, if available."""
+    if hasattr(rctx, "watch"):
+        rctx.watch(*args, **kwargs)
+
+# TODO: Remove after Bazel 6 support dropped
+def _watch_tree(rctx, *args, **kwargs):
+    """Calls rctx.watch_tree, if available."""
+    if hasattr(rctx, "watch_tree"):
+        rctx.watch_tree(*args, **kwargs)
```
aignas, line 374: Thanks, I'll use it in #2059.

python/private/repo_utils.bzl
```diff
@@ -262,12 +341,50 @@ def _outputs_to_str(result):
...
+    getenv = _getenv,
     is_repo_debug_enabled = _is_repo_debug_enabled,
-    debug_print = _debug_print,
-    which_checked = _which_checked,
     logger = _logger,
+    watch = _watch,
+    watch_tree = _watch_tree,
+    which_checked = _which_checked,
+    which_unchecked = _which_unchecked,
 )
```
aignas, line 390: It feels like this is getting big, but it's good to leave this as is for now. All of the things are low level `repository_ctx` related things, so it's fine.

tests/integration/local_toolchains/.bazelrc
```diff
@@ -0,0 +1,5 @@
+common --action_env=RULES_PYTHON_BZLMOD_DEBUG=1
```
aignas, line 1: Oh, using this in tests is really useful. Should we have this in CI for the `examples` jobs as well?

aignas (approved):
I really like how clean your `repository_rule` code is. Really nice structure and I like how self-contained the code is. Thank you!

tests/integration/local_toolchains/.bazelrc
aignas: I think having a `Troubleshooting` section might be good. We could point to Troubleshooting sub-sections within Toolchains, PyPI integration, etc.

python/private/repo_utils.bzl
aignas: I like how go has the `ctx` and passes it around everywhere. It is kind of a sign that IO is being done here or there. That makes pure starlark functions really obvious.

So I like the current design a lot. :)

aignas:
BTW, I realized that the current hermetic toolchain understands if it is Linux GLIBC. Is it possible to detect if the interpreter is built for MUSL vs GLIBC?

rickeylev:
After a bit of searching, yes, i think so. The info looks to be in the platforms module: https://stackoverflow.com/questions/72272168/how-to-determine-which-libc-implementation-the-host-system-uses

aignas:
Regarding `glibc`, if you look at the hermetic toolchain, we are specifying the flag value to be set to `glibc` (I haven't got to adding `musl` toolchains yet), but what it essentially is doing is including the right interpreter when building docker images, etc.

I know that this is meant for local toolchains, so is low priority IMHO, but it seems that having the `flag_values` in the `target_settings` include the `glibc` might mean that we can correctly handle precedence of toolchains if we are targeting the `musl` libc.

rickeylev:
Ok, I was a bit wrong about how easy detecting musl is.

`platform` has a `libc_ver()` function, which can return the libc implementation, _but_ it doesn't know how to indicate something is musl. It'll just return empty string. Which maybe that'll have to suffice.

The other option is to use `packaging`, but that isn't part of the stdlib, and the way it works is a decent amount of code. A small ELF format parser to find some path to something, then runs it to get some output. There's a PR for python itself, https://github.com/python/cpython/pull/103784, however, it seems to have stalled out.

Since it seems like the answer is to somehow parse something out of the executable's ELF information, I tried some various system utils to do the same, but couldn't figure out how.

Looking through sysconfig, there do appear to be several mentions of musl in some variables, e.g. as part of command line argument. So that could be a weak signal.

aignas:
Thanks for digging.

Regarding use of packaging, it is used in all of the whl metadata parsing (and env marker eval), so using it should be OK.

We could vendor it if needed, just like what pip is doing with it's dependencies, or download it like we do for whl_library. I would probably be +1 on vendoring it since it is such a critical piece of code in the bootstrap.

PR #2001, opened 2024-06-22
chore: remove dead code from whl_target_platforms

aignas:
Remove dead code only used in tests.

Work towards #260


groodt (approved):
LGTM

PR #2002, opened 2024-06-22
fix(bzlmod): pass only non-default values to whl_library

aignas:
Before this change `isolated`, `quiet` and `timeout` would get dropped
if they were specified to some non-default value. This fixes it.

Work towards #260.


groodt (approved):
LGTM

PR #2005, opened 2024-06-22
docs: Removes dead link after docs were migrated to sphinx

groodt:
I believe the intended content is now here: https://rules-python.readthedocs.io/en/latest/#migrating-from-the-bundled-rules

Im not sure there is much value linking to that page, but I don't mind. Just let me know.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,13 +13,9 @@ Documentation for rules_python is at <https://rules-python.readthedocs.io> and i
 
 Examples live in the [examples](examples) directory.
 
-Currently, the core rules build into the Bazel binary, and the symbols in this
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -13,13 +13,9 @@ Documentation for rules_python is at <https://rules-python.readthedocs.io> and i
 
 Examples live in the [examples](examples) directory.
 
-Currently, the core rules build into the Bazel binary, and the symbols in this
-repository are simple aliases. However, we are migrating the rules to Starlark and removing them from the Bazel binary. Therefore, the future-proof way to depend on Python rules is via this repository. See[`Migrating from the Bundled Rules`](#Migrating-from-the-bundled-rules) below.
-
 The core rules are stable. Their implementation in Bazel is subject to Bazel's
```
aignas, line 16:
```suggestion
The core rules are stable. Their implementation is subject to Bazel's
```

Maybe this would be better?

aignas (approved):

PR #2011, opened 2024-06-24
refactor: call a function to define internal pypi flags instead of listcomp

rickeylev:
This is updated so tooling can more automatically process the files. In particular, it helps tools, like buildozer, process the files, which makes it easier to import the code into Google. This is because there is a named target that buildozer can be told to process, whereas, with a list comprehension, it's an arbitrary chunk of code that has to be patched, without an identifiable label.

### DIFF
--- a/python/config_settings/BUILD.bazel
+++ b/python/config_settings/BUILD.bazel
@@ -10,10 +10,10 @@ load(
 )
 load(
     "//python/private/pypi:flags.bzl",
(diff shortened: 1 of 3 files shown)


groodt (approved):
LGTM

aignas (approved):
Thanks! It caught me again. I would love to have a buildifier flag disallowing list comps in build files.

PR #2018, opened 2024-06-27
feat: add runtime_env toolchain suite to replace "autodetecting" toolchain

rickeylev:
This adds a more comprehensive replacement for the "autodetecting" toolchain. Specifically, it defines all our toolchain types so that they take precedence when specified. This prevents the hermetic toolchains (registered by default) from accidentally being used when undesired.

To keep the behavior backwards compatible, an alias is added for the autodetecting toolchain with a deprecation notice.

The name `runtime_env` was chosen instead of "autodetecting" so that it's more clear these
toolchains are not "automatic" or "detecting" anything -- they're just taking a value
from the runtime environment and using it.


aignas (approved):

python/private/BUILD.bazel
```diff
@@ -365,6 +367,13 @@ config_setting(
     },
 )
 
+# This should only be set by analysis tests to expose additional metadata to
+# aid testing, so a setting instead of a flag.
+bool_setting(
+    name = "visible_for_testing",
+    build_setting_default = False,
```
aignas, line 374:
```suggestion
    build_setting_default = False,
    # This is only because it is an implicit dependency by the toolchains.
    visibility = ["//visibility:public"],
```

python/BUILD.bazel
```diff
@@ -320,11 +319,16 @@ toolchain_type(
...
+alias(
+    name = "autodetecting_toolchain",
+    actual = "//python/runtime_env_toolchains:runtime_env_toolchain",
+    deprecation = "Use //python/runtime_env_toolchains:all instead",
+)
 
 alias(
     name = "autodetecting_toolchain_nonstrict",
-    actual = ":autodetecting_toolchain",
+    actual = "//python/runtime_env_toolchains:all",
```
aignas, line 330: Yeah, I tried this in #1968, but it did not work as it seems that `alias` cannot point to these special targets.

rickeylev: oops, fixed.

PR #2022, opened 2024-07-01
(chore): Remove the extra indirection and `bzlmod` folder

groodt:
The `bzlmod` folder seems like unnecessary indirection now. It was perhaps useful as we developed bzlmod support, but we no longer need it and it makes navigating the repository more difficult.


python/extensions/pip.bzl
```diff
@@ -14,6 +14,6 @@
 
 "pip module extension for use with bzlmod"
 
-load("//python/private/bzlmod:pip.bzl", _pip = "pip")
+load("//python/private/pypi:pip.bzl", _pip = "pip")
```
aignas, line 17: nit: maybe we should rename the `pypi:bzlmod.bzl` to `pypi:extension.bzl` and use that instead of `pypi:pip.bzl`? It does seem like we could benefit from this cleanup.

python/extensions/python.bzl
```diff
@@ -14,6 +14,6 @@
 
 "Python toolchain module extensions for use with bzlmod"
 
-load("//python/private/bzlmod:python.bzl", _python = "python")
+load("//python/private:python.bzl", _python = "python")
```
aignas, line 17: I am wondering if we should have a separate `python/private/toolchain` or `python/private/interpreter` directory, which could have all of the `python` extension related things. The `pythons_hub.bzl` could be in that folder as well.

aignas (commented):
Thanks for initiating the cleanup. Basically have 2 questions:
* Minor question about not having `pypi/bzlmod.bzl` file.
* A bigger question about creating a directory for python toolchain related things.

python/extensions/pip.bzl
groodt: SGTM. Will rename.

python/extensions/python.bzl
groodt: SGTM. Probably an independent PR? It's a bigger change to discuss. Because we could also leave the python stuff where it is and move everything else out, which then makes it less urgent to change the Python toolchain or language rules.

aignas (approved):

PR #2024, opened 2024-07-01
fix: make Python repositories `chmod`/`id/`touch` hermetic

### LINKED ISSUE #2016: Hermetic `chmod`/`touch`/`id` in `python_repository`
# 🚀 feature request

### Relevant Rules

`python_repository`

### Description

We [currently](https://github.com/bazelbuild/rules_python/blob/f56fe42e858e9cf381a7598131b49a130550d8ab/python/repositories.bzl#L196-L214) call out the to system `chmod`, `touch` and `id` binaries.

We could do the equivalent functionality with Python.

The Python interpreter is downloaded in `python_repository`. If we provide a script as a private attribute to the repository rule, we can run them with the downloaded interpreter to perform the readonly permissions change.

None. Python has the functionality to perform the `chmod`/`touch`/`id`.

mattyclarkson:
Removes the need for system binaries to make the Python interpreter library directories/files read-only.

Fixes #2016


python/repositories.bzl
```diff
@@ -104,6 +106,24 @@ def is_standalone_interpreter(rctx, python_interpreter_path):
...
+
+    if platform == host_platform:
+        return python
+
+    label = "@@{suffix}~{repo}_{host_platform}//:{python}".format(
+        suffix = rctx.attr.name.rsplit("~python", 1)[0],
+        repo = "python_3_11",  # Aligns with what is defined in `MODULE.bazel`
+        host_platform = host_platform,
+        python = python,
+    )
```
mattyclarkson, line 123: I'm doing a terrible job of getting the correct logic here. Attempting to figure it out.

Any tips welcome.

mattyclarkson: This would be _best_ solved with a label to the host Python interpreter but that would lead to a circular edge when downloading the host interpreter. This may require some more invasive refactoring.

python/repositories.bzl
```diff
@@ -104,6 +106,16 @@ def is_standalone_interpreter(rctx, python_interpreter_path):
...
 
+def _get_host_python_interpreter(rctx):
+    (os_name, arch) = get_host_os_arch(rctx)
+    host_platform = get_host_platform(os_name, arch)
+    platform = rctx.attr.platform
+
+    if platform == host_platform:
+        return "python.exe" if ("windows" in platform) else "bin/python3"
+
+    return Label("@python_host//:python")
```
aignas, line 117: I think here we should return `None` and not do anything.

python/private/touch.py
```diff
@@ -0,0 +1,43 @@
+#! /usr/bin/env python3
+
+# Copyright 2023 The Bazel Authors. All rights reserved.
```
aignas, line 3: nit: the date is wrong here

python/repositories.bzl
```diff
@@ -193,15 +205,17 @@ def _python_repository_impl(rctx):
         if "windows" not in platform:
             lib_dir = "lib" if "windows" not in platform else "Lib"
 
+            python = _get_host_python_interpreter(rctx)
```
aignas, line 208: I think we should use the `python_bin` var that is defined below instead of using the `host_python`. And we should do the `read-only` fix only if the `host` platform the same as the `platform` of the toolchain.

Could we recursively change this for `windows` as well? Right now we are calling this only if it is not Windows, but do the scripts work on Windows?

mattyclarkson: [os.chmod](https://docs.python.org/3/library/os.html#os.chmod) says:

> Although Windows supports [chmod()](https://docs.python.org/3/library/os.html#os.chmod), you can only set the file’s read-only flag with it (via the `stat.S_IWRITE` and `stat.S_IREAD` constants or a corresponding integer value). All other bits are ignored.

So, yes, it would remove write access on Windows.

mattyclarkson: The `chmod.py` _does_ set the folders and files to read-only on Windows, but does not prevent new files being created in the directories. So the same behaviour as Unix is not present. We would have to look into Windows permissions to restrict that.

The `getuid` could return the Window Security ID for the user but that doesn't really help much. I've removed the `touch` check by checking `platform`.

We now run the scripts on Windows, but they don't really do much and actually fail to prevent `.pyc` files being written. Which is no different to before this patch.

mattyclarkson: One can prevent writing files into directories on Windows with security permissions. That could be done by calling out to `icacls.exe` (in leiu of the `win32security` dependency). That will likely take significant extra testing as I'm not a Windows expert.

mattyclarkson: @aignas would you like me to continue figuring out truely read-only directories on Windows or does this PR move the needle enough to be considered for merge? I'd be happy to raise follow up issues/PRs, if necessary.

aignas:
@rickeylev, what do you think about this? How do we move with this forward? Should we just start ignoring the `pyc` files in the repo and not bother with making it `read-only`?. The making of the external repos read-only is something that we don't do in the `whl_library` (maybe we should) and I am wondering what is the overall strategy here...

I am thinking that having the `chmod`, `id` and `touch` be hermetic might be ignoring the overall picture of what we are trying to do here.

rickeylev:
> what is the overall strategy here

So, correctness wise, there are two things we're constrained by:
1. The input files (both content and the list of them) have to be stable.
2. The downloaded python binary may not be runnable by the host that downloaded it, such as in a cross-build with RBE situation.

(1) can be accomplished by any of:
* a. Making everything read only.
* b. Generating all the necessary files at repo install time.
* c. Excluding files that might be created at runtime

I think we should do (a) regardless -- its just defense in depth. For the case of Windows (or other failures) we should just be OK with leaving things writable. It's the best we can do, and users will just bypass it entirely with ignore_root_error=True anyways (they are given the choice between "guaranteed pedantic failure" and "probably works, but flaky"; they'll always choose the latter).

I think (b) is better than (c), but it's out of scope for this PR and we already have (c) in place.

(2) throws a wrench into what this PR is doing -- using the downloaded runtime itself to run commands to make things read only. So I'm not really keen on that overall. I don't mind having an alternative implementation of chmod for OP's case where coreutils aren't available. Ideally, we'd either have a repo-phase reference to a host-compatible Python, or a repo-phase reference to a host-compatible chmod tool.

So, more concretely, I ask for the following changes:

* Use the system chmod if available, otherwise use the downloaded-python's chmod. If neither is available, oh well; continue on.
* Remove the fail() code paths. Just chmod and trust that things are OK.

mattyclarkson:
I've updated to use `uutils/coreutils` and removed the fail paths.

python/private/python_repositories.bzl
```diff
@@ -106,6 +106,71 @@ def is_standalone_interpreter(rctx, python_interpreter_path, *, logger = None):
         logger = logger,
     ).return_code == 0
 
+def _chmod(rctx, platform = None):
```
aignas, line 109: Speculation, that needs to be measured: The way this is implemented it would fetch and extract the chmod binary for each python version. It would use the bazel repository cache, but having the chmod in a separate repository might be better from the performance point of view?

In general, using the chmod that is hermetic is an interesting idea, but I wonder if we should use the system chmod if it exists and if it doesn't, download the hermetic chmod. That would also satisfy the ask to support your usecase where system chmod is not available. All of this could be done in a separate host_chmod repository, we could even switch the behaviour to default to the hermetic chmod if an env var is set to a particular value.

mattyclarkson:
I'm keen to work through getting this landed. I can rework the branch to current `main`.

Could I get some hints on how to implement this so that it has the best chance to get accepted, please?

aignas:
I think there are multiple options here:

# option 1 - use this chmod

In this case I think it would be best to create a separate repository rule maybe something called `host_coreutils` and then it will be very clear that we are trying to download the host-specific coreutils version.

We register this somewhere in `internal_deps.bzl` and then use it from the hermetic toolchain. It should be "used" internally as a dep in `MODULE.bazel`

# option 2 - trust our glob and remove chmod

Remove the chmoding. This would solve the following:
* Delete the bazel cache requires a sudo, because the dirs are read-only
* Difference in what we do on Windows vs Linux today
* Need to chmod would be no-longer there

Drawback:
* Theoretical polution with pyc files if we don't use our python helpers to execute Python scripts in the repository context. Should we document that users should use `-B`?

# option 3 - do nothing

Requires on system `chmod`, which is undesirable as you have suggested by creating the PR.

# proposal

I would actually propose option 2, @rickeylev, @mattyclarkson, what do you think?

cc: @arrdem

mattyclarkson:
Option 2 is fine with me. @rickeylev?

rickeylev:
Option 2 SGTM

arrdem:
Option 1 seems bad, option 3 I don't follow the downsides of. I'm leaning 2 by default but I don't feel like I have an informed opinion here.

aignas:
So I'd propose us to go with the option 2 moving forward.

Issue #1260, opened 2023-06-08
bazel coverage fails when bazel test successful


arunkant:
# 🐞 bug report

### Affected Rule

The issue is caused by the rule: py_test

### Is this a regression?
Don't know

### Description
We are trying to run django tests through bazel. `bazel test ...` command is working fine but `bazel coverage ...` fails with error 
```
ModuleNotFoundError: No module named 'settings'
```

## 🔥 Exception or Error

<pre><code>
Traceback (most recent call last):
  File "/tmp/arunkant/bazel/_bazel_arunkant/e454b3605424c5208f266bd2853bae79/execroot/qureai/bazel-out/k8-fastbuild/bin/packages/python/image_manager/image_manager_test.runfiles/qureai/packages/python/image_manager/tests/main.py", line 13, in <module>
    execute_from_command_line(
  File "/tmp/arunkant/bazel/_bazel_arunkant/e454b3605424c5208f266bd2853bae79/execroot/qureai/bazel-out/k8-fastbuild/bin/packages/python/image_manager/image_manager_test.runfiles/python_deps_django/site-packages/django/core/management/__init__.py", line 419, in execute_from_command_line
    utility.execute()
  File "/tmp/arunkant/bazel/_bazel_arunkant/e454b3605424c5208f266bd2853bae79/execroot/qureai/bazel-out/k8-fastbuild/bin/packages/python/image_manager/image_manager_test.runfiles/python_deps_django/site-packages/django/core/management/__init__.py", line 413, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/tmp/arunkant/bazel/_bazel_arunkant/e454b3605424c5208f266bd2853bae79/execroot/qureai/bazel-out/k8-fastbuild/bin/packages/python/image_manager/image_manager_test.runfiles/python_deps_django/site-packages/django/core/management/commands/test.py", line 23, in run_from_argv
    super().run_from_argv(argv)
  File "/tmp/arunkant/bazel/_bazel_arunkant/e454b3605424c5208f266bd2853bae79/execroot/qureai/bazel-out/k8-fastbuild/bin/packages/python/image_manager/image_manager_test.runfiles/python_deps_django/site-packages/django/core/management/base.py", line 346, in run_from_argv
    parser = self.create_parser(argv[0], argv[1])
[...]

PR #2025, opened 2024-07-01
[fix] Populating imports even when python_root is current directory

linzhp:
`bazel coverage` cannot find modules in the current directory unless the current directory is added to `imports` attribute of py targets (#1260). For regular modules, this is fine because it's generally a good practice to import with fully qualified path from python root. However, this becomes a problem for modules at python root to import each other. For example, when python root is at `src`, and `src/baz_test.py` wants to import `src/baz.py`, we will need `imports = ["."]` for the `py_test` target for `bazel coverage` to work.

### DIFF
--- a/gazelle/README.md
+++ b/gazelle/README.md
@@ -240,6 +240,13 @@ py_libary(
     imports = ["../.."],  # Gazelle adds this
     ...
 )
+
+# in ./src/BUILD.bazel
+py_test(
+    srcs = ["baz_test.py"],
+    imports = ["."],  # This is not needed for "bazel test", but needed for "bazel coverage", when src/baz_test.py imports src/baz.py with "import baz".
+    ...
+)
 ```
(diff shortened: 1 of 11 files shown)


aignas:
LGTM in general, could you please add a note in `CHANGELOG.md` so that the fix gets announced when we release a new version please?

linzhp:
Actually, it seems [--@rules_python//python/config_settings:bootstrap_impl=script](https://rules-python.readthedocs.io/en/latest/api/python/config_settings/index.html#bootstrap_impl) fixed  #1260, which makes this PR unnecessary. Let me do some more testing

linzhp:
[--@rules_python//python/config_settings:bootstrap_impl=script](https://rules-python.readthedocs.io/en/latest/api/python/config_settings/index.html#bootstrap_impl) works

PR #2035, opened 2024-07-04
feat: use hermetic interpreter in shebang

oxidase:
The PR modifies a default shebang resolution:
1. User-defined 
2. Hermetic interpreter *new behaviour*
3. DEFAULT_STUB_SHEBANG

The change allows to drop requirement on a system Python in the case of a hermetic python runtime.

### DIFF
--- a/python/private/common/py_runtime_rule.bzl
+++ b/python/private/common/py_runtime_rule.bzl
@@ -96,6 +96,8 @@ def _py_runtime_impl(ctx):
             interpreter_version_info["minor"],
         )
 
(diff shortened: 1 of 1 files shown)


rickeylev:
Using the runfiles path as the shebang path isn't reliable enough because it only works if the PWD is the runfiles root.

If you want to remove the system python dependency, then you'll have better luck using [--bootstrap_impl=script](https://rules-python.readthedocs.io/en/latest/api/python/config_settings/index.html#bootstrap_impl)

Issue #1972, opened 2024-06-14
gazelle: generate py_pytest_main


alexeagle:
Related to pytest support #240 

The outcome of https://github.com/bazelbuild/rules_python/pull/723 is that pytest-specifics landed in other repos. https://github.com/caseyduquettesc/rules_python_pytest and https://github.com/aspect-build/rules_py/blob/main/docs/rules.md#py_pytest_main are a couple examples of code generators that create the entrypoint file that shims between `py_test#entry_point` and what `pytest` expects to run.

>[!WARNING]
>Note, without the shim, `py_test(srcs=["some.pytest.test.py"])` silently passes without running any tests!

The `something` added in this test case https://github.com/bazelbuild/rules_python/blob/main/gazelle/python/testdata/generated_test_entrypoint/BUILD.out#L3 alludes to the real problem, which is that gazelle knows to use that `__test__.py` as the entrypoint, but doesn't know how to generate the `something`.

It should generate a `py_pytest_main`, allowing users to control via `map_kind` which ruleset they'd like to use to provide that rule.

linzhp:
I think there are two ways to fix this issue:

1. we can create separate Gazelle extension to generate py_pytest_main with the name `__test__`. The extension will have to run before the Python extension, so the latter is aware of the newly generated `__test__` target and create single py_test target with the `main` pointing to the it. This is probably least intrusive. The new extension can live in aspect_rules_py together with py_pytest_main.
2. generating py_pytest_main in the existing Python extension like #2040 attempted to do. However, we need to be careful here. The existing extension will generate a py_test target per test file if no `__test__.py` or `:__test__` target is found. If py_pytest_main is generated, those per-file py_test targets should be removed. We may need to introduce a new Gazelle directive in case people want to generate per-file targets instead.

> allowing users to control via map_kind which ruleset they'd like to use to provide that rule.

I am not sure how useful this is. If people decide to use a different rule set to provide `py_pytest_main`, they may need different attributes, unless we can make a convention that `py_pytest_main` should be a macro without any attribute besides `name = "__test__"`.

cc @jbedard

alexeagle:
Thanks @linzhp I think that's a good idea for us to put the extension next to the rule that it produces. In `aspect configure` we pre-compile a gazelle binary, so this wouldn't increase complexity for our users either.

[...]

PR #2040, opened 2024-07-05
feat(gazelle): generate py_pytest_main __test__ target

jbedard:
Generate the `py_pytest_main` target alongside tests. See https://github.com/bazelbuild/rules_python/issues/1972


gazelle/python/generate.go
```diff
@@ -476,6 +477,31 @@ func (py *Python) GenerateRules(args language.GenerateArgs) language.GenerateRes
 		result.Imports = append(result.Imports, pyTest.PrivateAttr(config.GazelleImportsKey))
 	}
 
+	if len(pyTestTargets) > 0 {
```
linzhp, line 480: This will make Gazelle generate a pytest main target and multiple py_test targets at the same time on the first run, and then generate another py_test target with the `main` pointing to the `__test__` target in the second run. People will have to run Gazelle twice to reach a stable state. At this state, every test file will be consumed by two `py_test` targets, so `bazel test //some:all` will execute all tests twice.

gazelle/python/kinds.go
```diff
@@ -97,4 +101,10 @@ var pyLoads = []rule.LoadInfo{
 			pyTestKind,
 		},
 	},
+	{
+		Name: "@aspect_rules_py//py:defs.bzl",
```
aignas, line 105: Similarly to how we discussed the gazelle plugin being agnostic to the rule sets, it would be good to have a solution for pytest that is not specific to any specific python ruleset.

Having rules_py specific logic in a rules_python gazelle plugin sounds like our abstractions within the bazel python ecosystem are not working well enough.

We had a similar discussion about rules_pycross and rules_python targets having different naming conventions and decided to make the plugin configurable, but in this case this is overstepping that boundary and starts tailoring the gazelle plugin to very specific implementation logic.

Are there different ways to achieve the same thing?

alexeagle: We could put the plugin logic in rules_py itself, and then teach users how to compile a gazelle binary that has the plugins listed in the right order. I don't think we want to grow a Go dev-dependency if we can help it though.

f0rmiga: We can make this file a template and allow users to pass extra stuff.

aignas: I am concerned that adding this special one off for `py_pytest_main` we are not making orthogonal rulesets that are easy to compose and will require continuous maintenance to tweak things. Since there are other test frameworks (like `behave`, `robot`, etc), catering for test frameworks in the gazelle plugin means that we are willing to accept contributions for all, hence I think that solving this differently would be better.

I liked the solution that @linzhp proposed in #2040. It does not add extra test framework specific concepts to the gazelle plugin and it would work with `pytest_main` if the users are willing to define a macro and use it, so whilst that may require users to write more code, it does mean that the `rules_python` project stays better scoped.

We have a maintainers meeting every Monday evening US time and we could discuss it there next week if one of you joined?

alexeagle: A modestly sophisticated user can discover the pattern suggested in #2044 to copy-paste a macro into their codebase and configure a `# map_kind` directive. I think that's good enough for the use cases that gazelle is trying to address.

We might do something more novice-friendly specifically in Aspect CLI.

aignas:
Closing in favour of #2044.

PR #2045, opened 2024-07-08
tests: add integration test for build_python_zip

rickeylev:
This is a more comprehensive regression test for verifying `--build_python_zip` is
actually working (https://github.com/bazelbuild/rules_python/issues/1840)

This also creates a small framework to make it easier to write integration tests that
need to customize the environment bazel runs in and check the output of bazel itself.
I figure this will be helpful for writing simple verification tests for repository/bzlmod
phase logic (i.e. set the debug env vars and grep the output). While we should avoid heavy
usage of these bazel-in-bazel tests, a bit of grepping logs would go a long way for covering
edge cases that examples don't cover.


rickeylev:
There we go. CI is happy now. PTAL

aignas (approved):
Nice, I like the `print(__file__)` :D

Issue #2043, opened 2024-07-06
(Linux) The stage-1 bootstrap script does not propagate certain signals to the stage-2 bootstrap script. 


chowder:
# 🐞 bug report

### Affected Rule

The issue is caused by the rule: `py_binary` / `py_test` 

### Is this a regression?

Yes. 

### Description

(This bug report is specific to Linux) 

In the old 1-stage bootstrap script, the script eventually runs the underlying program with `os.execv`, which replaces the process entirely, so that's ok - any signals get propagated directly to our program. 

```python
def _RunExecv(python_program, main_filename, args, env):
  # type: (str, str, list[str], dict[str, str]) -> ...
  """Executes the given Python file using the various environment settings."""
  os.environ.update(env)
  PrintVerbose("RunExecv: environ:", os.environ)
  argv = [python_program, main_filename] + args
  PrintVerbose("RunExecv: argv:", python_program, argv)
  os.execv(python_program, argv)
```

https://github.com/bazelbuild/rules_python/blob/main/python/private/python_bootstrap_template.txt#L379-L386

In the new 2-stage boostrap script(s), the first stage invokes the second stage directly:

```bash
env \
  "${interpreter_env[@]}" \
  "$python_exe" \
  "${interpreter_args[@]}" \
  "$stage2_bootstrap" \
  "$@"

exit $?
```
https://github.com/bazelbuild/rules_python/blob/main/python/private/stage1_bootstrap_template.sh#L111-L116

I'm not sure if it's platform depedent - but at least on Debian and CentOS, bash will start a child process, and signals like SIGINT don't get handled nor propagated. 

I think this is a simple fix, changing the stage-2 invocation to:

```bash
exec env \
  "${interpreter_env[@]}" \
  "$python_exe" \
  "${interpreter_args[@]}" \
  "$stage2_bootstrap" \
  "$@"
```

I'm happy to raise a PR for the above if it is agreed to be OK. 

## 🔬 Minimal Reproduction

I don't have one at hand, but I can provide one if needed. 

## 🌍  Your Environment

**Operating System:**

Linux AMD64 - CentOS 7, and Debian 11

**Output of `bazel version`:**

Bazel 7.2.1 

**Rules_python version:**

0.34.0

rickeylev:
Oh yeah, I vaguely recall this being fixed in the original bootstrap. Yeah, I see a bunch of code for this in the original.

[...]

PR #2047, opened 2024-07-08
fix: use `exec` to invoke the stage-2 bootstrap for non-zip case

chowder:
When the two-stage bootstrap is used, the parent shell process runs python as a child 
process, which changes how signals are propagated. Specifically, if a signal is sent
_directly_ to the parent (e.g. `kill $parent`), the child process (python) won't receive
it and it will appear to be ignored. This is because the parent process is busy waiting
for the child process.

To fix, invoke the python process using `exec` instead. Because the process is entirely
replaced, signals are sent directly to the replacement. This can't be used for zip files,
though, because they rely on a catching the exit signal to perform cleanup of the extracted
files.

Fixes https://github.com/bazelbuild/rules_python/issues/2043

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,9 @@ A brief description of the categories of changes:
 * Nothing yet
 
 ### Fixed
+* (rules) Signals are properly received when using {obj}`--bootstrap_impl=script`
+  (for non-zip builds).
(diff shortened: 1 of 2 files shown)


rickeylev (approved):
LGTM.

https://github.com/bazelbuild/rules_python/pull/2045 adds the basics for testing this, so we can put a test in after that lands.

Issue #1890, opened 2024-05-10
[FR] simplify gazelle_python.yaml


hunshcn:
# 🚀 feature request

### Relevant Rules

`gazelle_python.yaml`

### Description

 A clear and concise description of the problem or missing capability...
`gazelle_python.yaml` will list all libs on packages
```
    matplotlib.backend_tools: matplotlib
    matplotlib.backends: matplotlib
    matplotlib.backends.backend_agg: matplotlib
    matplotlib.backends.backend_cairo: matplotlib
    matplotlib.backends.backend_gtk3: matplotlib
    matplotlib.backends.backend_gtk3agg: matplotlib
    matplotlib.backends.backend_gtk3cairo: matplotlib
    matplotlib.backends.backend_gtk4: matplotlib
    matplotlib.backends.backend_gtk4agg: matplotlib
    matplotlib.backends.backend_gtk4cairo: matplotlib
    matplotlib.backends.backend_macosx: matplotlib
    matplotlib.backends.backend_mixed: matplotlib
    matplotlib.backends.backend_nbagg: matplotlib
    matplotlib.backends.backend_pdf: matplotlib
    matplotlib.backends.backend_pgf: matplotlib
...
```

I have a three lines only requirements.in
```
numpy
requests
seaborn
```
20 packages requirements.txt

but received a 110KB gazelle_python.yaml which has 2517 keys at `.manifest.modules_mapping`

It is too big.

### Describe the solution you'd like

 If you have a solution in mind, please describe it.

Can we simplify it and make `gazelle_python.yaml` have only top level package name?

In my shallow cognition, such a long imports seems meaningless.

I'd like to support it if I can.

### Describe alternatives you've considered

 Have you considered any alternative solutions or workarounds?
no

aignas:
I am not sure if the complexity is worth it as the package imports may overlap at the top level. It could be possible to improve it, but I am not sure how realistic it is.

Let's imagine a case where you write code
```
import foo.bar
```
and the `foo.bar` import is only provided by `foo-bar-plugin` python package, but what you have in the `gazelle_python.yaml` is data that tells gazelle that `foo` maps to `foo` python package, so gazelle will be happy and will add the wrong dependency instead of erroring here.

I think this is related to #1877, but might not be.

If somebody was to implement it, I think it should:
* Do the right thing in the case outlined above and I am not sure what that right thing is yet.
* Handle cases where we have platform specific imports.
[...]

PR #2048, opened 2024-07-09
feat(gazelle): simplify gazelle_python.yaml

hunshcn:
Simplify and make gazelle_python.yaml have only top level package name.

It would work well in cases to reduce merge conflicts.

Resolve https://github.com/bazelbuild/rules_python/issues/1890

This is a compatible change and seems to have no side effects, so no options are provided.


gazelle/pythonconfig/pythonconfig.go
```diff
@@ -494,9 +502,9 @@ func (c *Config) FormatThirdPartyDependency(repositoryName string, distributionN
 		normConventionalDistributionName = strings.Trim(normConventionalDistributionName, "_")
 	case Pep503LabelNormalizationType:
 		// See https://packaging.python.org/en/latest/specifications/name-normalization/#name-format
-		normConventionalDistributionName = strings.ToLower(conventionalDistributionName) // ... "should be lowercased"
+		normConventionalDistributionName = strings.ToLower(conventionalDistributionName)                                        // ... "should be lowercased"
 		normConventionalDistributionName = regexp.MustCompile(`[-_.]+`).ReplaceAllString(normConventionalDistributionName, "-") // ... "all runs of the characters ., -, or _ replaced with a single -"
-		normConventionalDistributionName = strings.Trim(normConventionalDistributionName, "-") // ... "must start and end with a letter or number"
+		normConventionalDistributionName = strings.Trim(normConventionalDistributionName, "-")                                  // ... "must start and end with a letter or number"
```
aignas, line 507: nit, the addition of the whitespace is weird.

aignas (commented):
@linzhp, @dougthor42, you have bigger projects using `gazelle` so it would be great to hear if this has any negative impacts. @alexeagle, since you have clients using `gazelle` maybe you have time to test this change out with bigger projects to let us know how this would work in cases where there is a lot of traffic and changes to what internal files packages have would cause merge conflicts?

gazelle/manifest/testdata/gazelle_python.yaml
```diff
@@ -1,13 +1,5 @@
 manifest:
   modules_mapping:
     arrow: arrow
```
aignas, line 3: Could you please add a test where we have a different distribution providing a namespace package? Something like following would be good to have. We should check that importing arrow.plugin correctly adds arrow_plugin to the deps.
```suggestion
    arrow: arrow
    arrow.plugin: arrow_plugin
```

hunshcn: I believe this case will be handled correctly because it is covered in example/build_file_generation.

But it does make sense to add the test. I will add it at night.

hunshcn: added

dougthor42:
Just FYI this is still on my radar. I expect to have some time in the next 48hrs.

aignas:
Yeah, I'll wait to give enough time for people to look into this.

dougthor42:
**Note:** I had to cherry-pick cca1d4efe49733217e7df30a84a172ad3d6e4ae1 (#2058) to get this to work, but that's expected.

`gazelle_python.yaml` is now just **2%** :exclamation: of the original size: It went from 28737 lines down to 549 lines. The "1 insertion" was simply the file hash changing.

```
 gazelle_python.yaml | 28190 +----------------------------------------------------------------------------------
 1 file changed, 1 insertion(+), 28189 deletions(-)
```


We have 505 BUILD.bazel files. I tried randomly deleting some of them and having gazelle remake them. No change[^1] :+1: 

```console
$ find src/ -type f -name BUILD.bazel -print0 | xargs -0 -I {} bash -c '[ $[ $RANDOM % 100 ] == 0 ] && echo deleting {} && rm {}'
list of 6 files
$ bazel run //:gazelle
$ git status
On branch u/dthor/bazel-simplify-gazelle-check
nothing to commit, working tree clean
```

I did the "randomly delete" again a couple more times. All good!

I also ran some `bazel test`, but I'd expect this to have zero impact on them. They ran just fine.

Overall I think this is :100: . Very nice! :tada: :partying_face: 

[^1]: Aside from expected changes that gazelle doesn't deal with, like `data` attribute or when my random delete removed a hand-made BUILD file.

linzhp:
Shorten gazelle_python.yaml greatly. I didn't run into issues with some quick test

gazelle/python/testdata/with_third_party_requirements_from_imports_legacy/gazelle_python.yaml
```diff
@@ -0,0 +1,1678 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+manifest:
+  modules_mapping:
+    cachetools: cachetools
+    cachetools.__init__: cachetools
+    cachetools.func: cachetools
+    cachetools.keys: cachetools
+    certifi: certifi
```
aignas, line 21: This looks like copy paste from somewhere else. The codebase no longer generates these long files. Could you please fix this?

aignas (approved):
Minor nit, but otherwise looks good. I'll wait for a few more days for someone form Aspect to chime in.

gazelle/python/testdata/with_third_party_requirements_from_imports_legacy/gazelle_python.yaml
hunshcn: According to https://github.com/bazelbuild/rules_python/blob/36ed9e029c275256b520b1b5d0d3329664441ae0/gazelle/python/testdata/with_third_party_requirements_from_imports/README.md. The gazelle_python.yaml is generated from https://github.com/sramirezmartin/gazelle-toy-example/blob/main/subdir1/requirements_lock.txt.

Do you think I should add this file to git?

aignas: Just cleaning up the `gazelle_python.yaml` would be sufficient here. This is test data. Having it generated may be too much upkeep.

hunshcn: test `with_third_party_requirements_from_imports_legacy` has been removed. It was meant to be a compatibility test for the original format. But the guys' manual testing has ensured its usability.

alexeagle:
I'm sorry, I have no time the rest of this month - maybe @jbedard has an idea since he worked on python gazelle recently

gazelle/python/testdata/with_third_party_requirements_from_imports_legacy/gazelle_python.yaml
aignas: Removing sounds good.

PR #2056, opened 2024-07-11
refactor: don't load repo-phase objects from build-phase

rickeylev:
As a general practice, the repo-phase and build-phase shouldn't load code from one
another because they can't use each other's objects. It can also result in confusing
behavior because the "starlark environment" is slightly different between the two phases.

Additionally, Google's version of Bazel essentially disables repo-phase objects, so
loading e.g. http_archive results in errors. This makes it more difficult to import
rules_python into Google, as we have to maintain patches to cut out the code (and thus
we spend more time trying to import the code than working on it).

### DIFF
--- a/python/private/pypi/deps.bzl
+++ b/python/private/pypi/deps.bzl
@@ -123,10 +123,7 @@ py_library(
 """
 
 # Collate all the repository names so they can be easily consumed
-all_requirements = [name for (name, _, _) in _RULE_DEPS]
-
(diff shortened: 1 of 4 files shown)


groodt (approved):
LGTM

At canva we also ban the usage of these macros in our repo :) I'd be happy to see them gone entirely, but it doesn't seem viable at this stage because there is too much entrenched usage.

PR #2063, opened 2024-07-13
fix: redirect entrypoints to py_console_script_binary

groodt:
`entrypoints` (originally from setuptools) AKA `console_scripts` (as defined in pyproject.toml) AKA `py_console_scripts` (as defined in rules_python) are a [packaging standard](https://packaging.python.org/en/latest/specifications/entry-points/). 

It therefore makes sense to package this code alongside the other packaging rules code and not the python language rules code. The packaging rules code currently sits inside `pypi`. Looking at the code, we can see that it is packaging related because it extracts the contents of `.dist-info` directories, which originate from built distributions (wheels).


python/private/pypi/BUILD.bazel
```diff
@@ -13,6 +13,8 @@
 # limitations under the License.
 
 load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
+load("@rules_python//python:defs.bzl", "py_library")
```
aignas, line 16: Please drop rules_python in the load statements in internal code.

python/entry_points/py_console_script_binary.bzl
```diff
@@ -13,12 +13,14 @@
...
-```
+Please use `py_console_script_binary` from `//python:pip_bzl` instead.
 """
 
-load("//python/private:py_console_script_binary.bzl", _py_console_script_binary = "py_console_script_binary")
+load("//python:pip.bzl", _py_console_script_binary = "py_console_script_binary")
+
+def _compatibility_shim(**kwargs):
+    # buildifier: disable=print
+    print("Please use `py_console_script_binary` from `//python:pip_bzl` instead.")
```
aignas, line 23: I saw people using deprecation attribute in some cases instead of a print (this is in py_proto_library case). That may be less spammy, as this will cause one print invocation per usage of the rule, I think?

Maybe it would be useful to print a buildozer command for replacing the loads?

python/entry_points/py_console_script_binary.bzl
```diff
@@ -13,12 +13,14 @@
 # limitations under the License.
 
 """
-Creates an executable (a non-test binary) for console_script entry points.
-
-```{include} /_includes/py_console_script_binary.md
-```
+Please use `py_console_script_binary` from `//python:pip_bzl` instead.
```
aignas, line 16: Nit, I think this should be pip.bzl instead of pip_bzl.

Maybe it would be great to write the full load statement to help users?

aignas (commented):
In general, not against moving this under private/pypi, but I am not sure if having a single pip.bzl file where we are loading everything from is great. Having a separate file per macro/function allows for much better isolation and cache hits.

What is more, this is a build rule and quite a few things in the pip.bzl are repo rules, except for the compile requirements.

python/entry_points/py_console_script_binary.bzl
groodt: SGTM. Didn't know about it. Will FIXUP.

python/entry_points/py_console_script_binary.bzl
groodt: SGTM. Will FIXUP.

python/entry_points/py_console_script_binary.bzl
rickeylev: For macros, I think there's a doc string syntax that stardoc parses out into a "deprecation" doc text. I think it's either `"""DEPRECATED: blabla """` or `"""[DEPRECATED]: blalba"""` ? Not entirely sure.

I'm pretty sure our docgen respects it when stardoc emits it.

rickeylev:
Yeah, I'll echo much of what Ignas said:

* Don't have kitchen sink bzl files. Over time, it incurs more unused loads.
* The above is especially true for the repo/loading/analysis phases. We want to avoid the repo-rule phase and loading-phases from causing loads of the other's code.

I'm -1 on moving the public symbol simply because it seems like needless churn. I guess if you _really_ want to move it, OK. But I think we should keep the old location around for longer than typical because people are still upgrading from workspace->bzlmod, and if they don't have a stable load() path for the entry point replacements, it'll be pretty annoying to go through upgrading.

I'm fine with moving the implementation under python/private/pypi.

aignas:
@groodt, do you plan (have time) to clean this up before 1.0?

aignas:
This is essentially done. Entry point code got removed.

PR #2059, opened 2024-07-12
feat(pypi): support env markers in requirements files

aignas:
Before this change the `all_requirements` and related constants will have
packages that need to be installed only on specific platforms and will mean
that users relying on those constants (e.g. `gazelle`) will need to do extra
work to exclude platform-specific packages. The package managers that that
support outputting such files now include `uv` and `pdm`. This might be also
useful in cases where we attempt to handle non-requirements lock files.

Note, that the best way to handle this would be to move all of the requirements
parsing code to Python, but that may cause regressions as it is a much bigger
change. This is only changing the code so that we are doing extra processing
for the requirement lines that have env markers. The lines that have no markers
will not see any change in the code execution paths and the python interpreter
will not be downloaded.

We also use the `*_ctx.watch` API where available to correctly re-evaluate the
markers if the `packaging` Python sources for this change.

Extra changes that are included in this PR:
- Extend the `repo_utils` to have a method for `arch` getting from the `ctx`.
- Change the `local_runtime_repo` to perform the validation not relying on the
  implementation detail of the `get_platforms_os_name`.
- Add `$(UV)` make variable for the `uv:current_toolchain` so that we can
  generate the requirements for `sphinx` using `uv`.
- Swap the requirement generation using `genrule` and `uv` for `sphinx` and co
  so that we can test the `requirement` marker code. Note, the `requirement`
  markers are not working well with the `requirement_cycles`.

Fixes #1105.
Fixes #1868.
Work towards #260, #1975.
Related #1663.

Issue #1975, opened 2024-06-17
Support `uv` as part of `rules_python`


aignas:
Just creating this as a placeholder for discussions and documenting some thoughts/findings. `rules_uv` is doing its job well and replicating the functionality in `rules_python` would be duplicate effort. However, `rules_uv` does not support `WORKSPACE` installations (at least the [releases] don't advertise such support) and `rules_python` needs to support `WORKSPACE` installations.

[releases]: https://github.com/theoremlp/rules_uv/releases

As a result I see the following options:
1. Do not support `uv`. This has a drawback of longer compilation times with the included `pip-compile` tooling. `uv` is just a binary and could in theory be later reused for common operations with whl packages.
2. Support `uv` only on `bzlmod` by including `rules_uv` as a dependency. This would require us to expose `rules_uv` via `@python_versions` or something similar so that we don't have loads that load `rules_uv` within the main `//python/*.bzl` files to avoid breaking `WORKSPACE` users. However this would not allow us to use it for our own example testing because they should produce the same output with and without bzlmod. Since we cannot dog-food `rules_uv`, it makes it a hard proposition.
3. Write the `WORKSPACE` `deps.bzl` implementation for `rules_uv` so that we can depend on it.
4. Support `uv` by copying/reimplementing the `uv` support as a separate rule/macro whilst not relying on `rules_uv`. The drawback is that we would have to reimplement parts of it, but the benefit would be that the extra tests that we would need to have would be next to the implementation. We could also provide Windows support in this way as `rules_uv` does not support Windows yet due to relying on `bash` scripting for `pip_compile`.

groodt:
Oh nice! I've got a POC of a basic `uv` integration of option 4. The logic is not complicated and it's much simpler if we only need to support blzmod. We can discuss at maintainers meeting, but thats the option I had planned anyway.

aignas:
FYI @mark-thm, feel free to write down any thoughts that you may have :)

mark-thm:
I don't have any particularly strong feelings here outside of liking `uv` as the toolchain to use to convert requirements input files to requrements txt files.

We'd be happy to take contributions that add WORKSPACE and/or Windows support. The latter will depend on rules_multitool getting Windows support (very close to landing). The bash scripts are trivial, but do let us customize the uv arguments and return to Bazel an executable, I'm not familiar with how to do that without a bash/batch script but eager to learn.

[...]

PR #2074, opened 2024-07-18
fix(uv): fix UV_BIN usage with current_toolchain

aignas:
Before this PR the `uv` toolchain could not be used in a `genrule` it
seems because the `//python/uv:toolchain` does not have the necessary
providers for using the make variables and the re-exporting of the make
variables from the `toolchain` in the `current_toolchain` rule did not
seem to work.

This PR removes the provider construction in the `toolchain` and does
that only in the `current_toolchain`. This better mirrors how the Python
toolchains are setup (grepping `PYTHON3` is sufficient to get examples).
This also splits out work done in #2059 to decrease its scope, so that
this can be discussed separately.

Work towards #1975


python/uv/private/current_toolchain.bzl
```diff
@@ -30,7 +30,9 @@ def _current_toolchain_impl(ctx):
     # Bazel requires executable rules to create the executable themselves,
     # so we create a symlink in this rule so that it appears this rule created its executable.
     original_uv_executable = toolchain_info.uv_toolchain_info.uv[DefaultInfo].files_to_run.executable
-    symlink_uv_executable = ctx.actions.declare_file("uv_symlink_{}".format(original_uv_executable.basename))
+
+    # Use `uv` as the name of the binary to make the help message well formatted
```
groodt, line 34: nit: The only reason I went with a weird name, was because I was worried about Windows. I might be wrong, but I thought Windows requires(ed) that executables have the `.exe` suffix?

groodt (approved):
LGTM

aignas:
Regarding the name of the binary, it seems that the CI is happy, so I'll roll
with it, having stable names can make debugging easier.

I think the `.exe` extension requirement is for output files from `_binary` rules but in this case it is not that, so let's see if it breaks anywhere in the future.

Issue #1868, opened 2024-04-25
`pip.parse` crashes with `StopIteration` if platform restriction prevents package from install


sin-ack:
# 🐞 bug report

### Affected Rule

The issue is caused by the `pip.parse` extension.

### Is this a regression?

No.

### Description

If the given requirement has a platform restriction like `platform_system == "Windows"` and `pip` refuses to download the requirements, then Bazel will crash.

## 🔬 Minimal Reproduction

https://github.com/sin-ack/rules_python-repro

## 🔥 Exception or Error

<pre><code>

INFO: Repository rules_python~~pip~pip_311_colorama instantiated at:
  <builtin>: in <toplevel>
Repository rule whl_library defined at:
  /home/agni/.cache/bazel/_bazel_agni/ba6071a1d8c8e24a5f08c6b275a4cff8/external/rules_python~/python/pip_install/pip_repository.bzl:932:30: in <toplevel>
ERROR: An error occurred during the fetch of repository 'rules_python~~pip~pip_311_colorama':
   Traceback (most recent call last):
	File "/home/agni/.cache/bazel/_bazel_agni/ba6071a1d8c8e24a5f08c6b275a4cff8/external/rules_python~/python/pip_install/pip_repository.bzl", line 763, column 31, in _whl_library_impl
		repo_utils.execute_checked(
	File "/home/agni/.cache/bazel/_bazel_agni/ba6071a1d8c8e24a5f08c6b275a4cff8/external/rules_python~/python/private/repo_utils.bzl", line 145, column 29, in _execute_checked
		return _execute_internal(fail_on_error = True, *args, **kwargs)
	File "/home/agni/.cache/bazel/_bazel_agni/ba6071a1d8c8e24a5f08c6b275a4cff8/external/rules_python~/python/private/repo_utils.bzl", line 86, column 13, in _execute_internal
		fail((
Error in fail: repo.execute: whl_library.ResolveRequirement(rules_python~~pip~pip_311_colorama, colorama==0.4.6 ; python_version >= "3.11" and python_version < "4.0" and platform_system == "Windows"     --hash=sha256:08695f5cb7ed6e0531a20572697297273c47b8cae5a63ffc6d6ed5c201be6e44     --hash=sha256:4f1d9991f5acc0ca119f9d443620b77f9d6b33703e51011c16baf57afb285fc6): end: failure:
[...]

PR #2075, opened 2024-07-18
refactor(internal): move the os/arch detection to repo_utils

aignas:
This also changes the local_runtime_repo to explicitly check for
supported platforms instead of relying on a `None` value returned by the
helper method. This makes the behaviour exactly the same to the
behaviour before this PR and we can potentially drop the need for the
validation in the future if our local_runtime detection is more robust.

This also makes the platform detectino in `pypi_repo_utils` not depend
on `uname` and only use the `repository_ctx`. Apparently the
`module_ctx.watch` throws an error if one attempts to watch files on the
system (this is left for `repository_rule` it seems and one can only do
`module_ctx.watch` on files within the current workspace. This was
surprising, but could have been worked around by just unifying code.

This splits out things from #2059 and makes the code more succinct.

Work towards #260, #1105, #1868.

### DIFF
--- a/python/private/local_runtime_repo.bzl
+++ b/python/private/local_runtime_repo.bzl
@@ -46,7 +46,7 @@ def _local_runtime_repo_impl(rctx):
     on_failure = rctx.attr.on_failure
 
     platforms_os_name = repo_utils.get_platforms_os_name(rctx)
-    if not platforms_os_name:
+    if platforms_os_name not in ["linux", "osx", "windows"]:
         if on_failure == "fail":
             fail("Unrecognized host platform '{}': cannot determine OS constraint".format(
                 rctx.os.name,
--- a/python/private/local_runtime_repo.bzl
+++ b/python/private/local_runtime_repo.bzl
@@ -45,22 +45,6 @@ def _local_runtime_repo_impl(rctx):
     logger = repo_utils.logger(rctx)
     on_failure = rctx.attr.on_failure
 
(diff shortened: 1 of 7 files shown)


python/private/local_runtime_repo.bzl
```diff
@@ -46,7 +46,7 @@ def _local_runtime_repo_impl(rctx):
     on_failure = rctx.attr.on_failure
 
     platforms_os_name = repo_utils.get_platforms_os_name(rctx)
-    if not platforms_os_name:
+    if platforms_os_name not in ["linux", "osx", "windows"]:
```
rickeylev, line 49: You can just remove this block; as mentioned in the originating PR, I just had it return None because I knew there were cases it didn't handle, but didn't know what they were. Now that it handles all the cases, we can just assume whatever value is returned is correct.

python/private/repo_utils.bzl
```diff
@@ -341,25 +341,60 @@ def _outputs_to_str(result):
...
+        return "freebsd"
+    if os.startswith("openbsd"):
+        return "openbsd"
+    if os.startswith("linux"):
+        return "linux"
+    if os.startswith("windows"):
+        return "windows"
+    return os
+
+def _get_platforms_arch_name(rctx):
```
rickeylev, line 371: How about `_get_platforms_cpu_name` instead? Since it's mapping the value to the `@platforms//cpu:<name>`, where the term is "cpu", not "arch".

rickeylev (approved):

Issue #2060, opened 2024-07-12
Add an option to opt-out of PYTHONSAFEPATH


allsey87:
# 🚀 feature request

### Relevant Rules
`py_binary`

### Description
`py_binary` sets `PYTHONSAFEPATH=1` which while being a sensible default does cause problems. Most notably, in `rules_foreign_cc` where `py_binary` is used to wrap up the Meson build system (written in Python). While this is not an issue for the build system itself, it is an issue when Meson runs compilers or build scripts that are written in Python. The issue here is that Meson is starting processes which inherit Meson's environment variables including `PYTHONSAFEPATH`. This often breaks scripts which rely on importing modules from the directory in which they are contained.

### Describe the solution you'd like
I would like an option on `py_binary` that allows me to opt out of `PYTHONSAFEPATH`.
```starlark
py_binary(
   name = "mypybinary",
   safe_path = False, # defaults to True
)
```

### Describe alternatives you've considered
- Using `--action_env=PYTHONSAFEPATH=`: no effect
- Specifying the `env` argument on `py_binary` as follows: no effect

```starlark
py_binary(
   name = "mypybinary",
   env = {
      "PYTHONSAFEPATH": ""
   }
)
```

rickeylev:
> setting PYTHONSAFEPATH in env has no effect

Yeah, we should respect an outer env setting at least.

Using `env` to set this does raise a bit of a design question. Normally that attribute is only respected when `bazel run` or `bazel test` is used with the target. That behavior has always struck me as a bit surprising. But, that behavior is also what most rules do, I think? So it's not surprising insofar as its the existing convention.

Anyways, at the least, we can update the logic so it doesn't always force the value.

allsey87:
> > setting PYTHONSAFEPATH in env has no effect
> 
> Yeah, we should respect an outer env setting at least.

I would lean towards a dedicated argument which also clearly documents that this environment variable is set by default, but allowing it to be unset via the `env` argument would also be sufficient.

> Using `env` to set this does raise a bit of a design question. Normally that attribute is only respected when `bazel run` or `bazel test` is used with the target.

Do you mean `env` taking effect during `bazel build` on a `py_binary`?

allsey87:
Any thoughts @aignas? If we can agree on disabling via `env` or a dedicated attribute, I can put a PR together.

sbc100:
[...]

PR #2076, opened 2024-07-18
feat: allow overriding PYTHONSAFEPATH env var

rickeylev:
By default, PYTHONSAFEPATH is enabled to help prevent imports being found where they
shouldn't be. However, this behavior can't be disabled, which makes it harder to use this
when explicitly desired.

To fix, the bootstrap respects the caller environment's PYTHONSAFEPATH environment variable, if set.

Fixes https://github.com/bazelbuild/rules_python/issues/2060


aignas (approved):

allsey87:
I will check if this solves #2060, but I have a hunch that just setting `PYTHONSAFEPATH` as empty (which is required for the opt-out to work) will still be enough to trigger this behavior in Python.

rickeylev:
One of the tests does exactly that and checks the interpreter flag to verify that safe path is disabled, so it should work :). Feel free to re-open the issue with a repro if you find otherwise.

Actually, I'm going to re-open the issue for now. This only fixes it for --bootstrap_impl=script, not windows, and, come to think of it, certain types of zips might also still have the bug.

rickeylev:
Doh, this introduced a bug that disable safe path by default unless it was opted in to. That'll be fixed shortly in #2073.

PR #2077, opened 2024-07-18
docs: document some environment variables

rickeylev:
This is so they're more discoverable for both developers and users.


groodt (approved):
LGTM

PR #2069, opened 2024-07-16
refactor(pypi): split out more utils and start passing 'abi_os_arch' around

aignas:
This is extra preparation needed for #2059.

Summary:
- Create `pypi_repo_utils` for more ergonomic handling of Python in repo context.
- Split the resolution of requirements files to platforms into a separate function
  to make the testing easier. This also allows more validation that was realized
  that there is a need for in the WIP feature PR.
- Make the code more robust about the assumption of the target platform label.

Work towards #260, #1105, #1868.

Issue #1708, opened 2024-01-19
`pip.parse(python_version=)` should not be mandatory


mering:
# 🐞 bug report

### Affected Rule

Bzlmod `pip.parse()`

### Is this a regression?

Yes, was not a problem in WORKSPACE.

### Description

The arg documentation says the following:
https://github.com/bazelbuild/rules_python/blob/711186f144af06b431bd416b2d742874de3a2dea/python/private/bzlmod/pip.bzl#L391-L398

It specifically describes `If not specified, then the default Python version (as set by the root module or rules_python) will be used`. So this attribute should be optional but currently is mandatory.

## 🔬 Minimal Reproduction

`pip.parse()` without `python_version`.

## 🔥 Exception or Error

```
ERROR: in tag at <root>/MODULE.bazel:42:10, mandatory attribute python_version isn't being specified
```

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
Linux
  </code>
</pre>

**Output of `bazel version`:**

<pre>
  <code>
2024/01/19 15:28:42 Warning: used fallback version "6.3.2"
Bazelisk version: v1.19.0
INFO: Running bazel wrapper (see //tools/bazel for details), bazel version 6.1.2 will be used instead of system-wide bazel installation.
Build label: 6.1.2
Build target: bazel-out/k8-opt/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Tue Apr 18 15:29:54 2023 (1681831794)
Build timestamp: 1681831794
Build timestamp as int: 1681831794
  </code>
</pre>

**Rules_python version:**

<pre>
  <code>
0.27.1
  </code>
</pre>

**Anything else relevant?**

aignas:
Could you explain a little more how having a requirement on this mandatory argument make it more difficult to use `rules_python`? Are there cases where you cannot do this? Usually `requirements.txt` files are python version and platform dependent, so I'd be curious to hear out more about your usecase.

mering:
> Could you explain a little more how having a requirement on this mandatory argument make it more difficult to use `rules_python`? Are there cases where you cannot do this? Usually `requirements.txt` files are python version and platform dependent, so I'd be curious to hear out more about your usecase.

[...]

PR #2078, opened 2024-07-19
fix(pypi): fix the whl selection algorithm after #2069

aignas:
It seems that a few things broke in recent commits:
- We are not using the `MODULE.bazel.lock` file and it seems that it is easy to
  miss when the components in the PyPI extension stop integrating well
  together. This happened during the switch to `{abi}_{os}_{plat}` target
  platform passing within the code.
- The logger code stopped working in the extension after the recent additions
  to add the `rule_name`.
- `repo_utils.getenv` was always getting `PATH` env var on bazel `6.x`.

This PR fixes both cases and updates docs to serve as a better reminder. By
fixing the `select_whls` code and we can just rely on target platform triples
(ABI, OS, CPU). This gets one step closer to maybe supporting optional
`python_version` which would address #1708. Whilst at it we are also adding
different status messages for building the wheel from `sdist` vs just
extracting or downloading the wheel.

Tests:
- Added more unit tests and brought them in line with the rest of the code.
- Checked manually for differences between the `MODULE.bazel.lock` files in our
  `rules_python` extension before #2069 and after this PR and there are no
  differences except in the `experimental_target_platforms` attribute in
  `whl_library`. Before this PR you would see that we do not select any wheels
  for e.g. `MarkupSafe` and we are always building from `sdist`.

Work towards #260.

### DIFF
--- a/python/private/repo_utils.bzl
+++ b/python/private/repo_utils.bzl
@@ -58,7 +58,7 @@ def _logger(ctx, name = None):
     else:
         verbosity_level = "WARN"
 
-    env_var_verbosity = ctx.os.environ.get(REPO_VERBOSITY_ENV_VAR)
+    env_var_verbosity = _getenv(ctx, REPO_VERBOSITY_ENV_VAR)
     verbosity_level = env_var_verbosity or verbosity_level
 
     verbosity = {
@@ -284,12 +284,9 @@ def _which_describe_failure(binary_name, path):
         path = path,
     )
 
-def _getenv(rctx, name, default = None):
-    # Bazel 7+ API
-    if hasattr(rctx, "getenv"):
-        return rctx.getenv(name, default)
-    else:
-        return rctx.os.environ.get("PATH", default)
+def _getenv(ctx, name, default = None):
+    # Bazel 7+ API has ctx.getenv
+    return getattr(ctx, "getenv", ctx.os.environ.get)(name, default)
 
 def _args_to_str(arguments):
     return " ".join([_arg_repr(a) for a in arguments])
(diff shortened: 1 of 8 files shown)


python/private/repo_utils.bzl
```diff
@@ -66,18 +67,25 @@ def _logger(rctx):
         "TRACE": 3,
     }.get(verbosity_level, 0)
 
+    if hasattr(ctx, "attr"):
+        # This is `repository_ctx`.
+        name = name or "{}(@@{})".format(getattr(ctx.attr, "_rule_name", "?"), ctx.name)
+    elif not name:
+        fail("The name has to be specified when using the logger with `module_ctx`")
+
+    name = getattr(ctx, "name", "extension")
```
rickeylev, line 76: This assignment is clobbering the logic above. I think you meant for this to be in the elif branch?

rickeylev:
>  Before this PR you would see that we do not select any wheels for e.g. MarkupSafe and we are always building from sdist.

Would this show up as the "Fetching repo bla bla ..." step taking a long time? When I was working on the docs yesterday, there were a couple pypi deps (markup safe sounds familiar) that seemed to be taking a long time to become available. I chalked it up slow downloads or because I was modifying the pypi blz code, but maybe it was building it from source in the repo phase.

rickeylev (changes requested):

aignas:
Correct, it was building `MarkupSafe` from source, that is why it was taking longer than usual, I'll add a better log message to indicate that, because we can distinguish between that when we pass `url` to `whl_library`.

python/private/repo_utils.bzl
aignas: Oops, this was leftover code, now removed.

rickeylev (approved):

PR #2082, opened 2024-07-21
refactor: use logger in repo_utils.execute_* functions

rickeylev:
This is to unify how it handles printing log messages.

This also updates repo rules using repo_utils to create loggers and set the `_rule_name` attribute for the logger to use.

* Also removes defunct repo_utils.debug_print
* Also makes some functions private that aren't used elsewhere


rickeylev:
Piping `logger` through helper functions is kind of annoying. It makes me want to wrap `rctx` entirely (so that we can do `rctx.execute_checked()`, without having to pass any of the `logger` args around). Or we do something like only allow loggers to be configured through attributes; that way, any helper that needs a logger can simply create one and be confident it's getting the right settings.

python/private/repo_utils.bzl
```diff
@@ -92,6 +82,7 @@ def _logger(ctx, name = None):
         debug = lambda message_cb: _log(2, "DEBUG", message_cb),
         info = lambda message_cb: _log(1, "INFO", message_cb),
         warn = lambda message_cb: _log(0, "WARNING", message_cb),
+        fail = lambda message_cb: _log(-1, "FAIL", message_cb, fail),
```
aignas, line 85: The only annoying thing about the fail method is that one has to do a return in the code afterwards to satisfy buildifier.

python/private/toolchains_repo.bzl
```diff
@@ -236,14 +237,15 @@ actions.""",
...
+    logger = repo_utils.logger(rctx)
     rctx.file("BUILD.bazel", """\
 # Generated by python/private/toolchains_repo.bzl
 
 exports_files(["python"], visibility = ["//visibility:public"])
 """)
 
-    (os_name, arch) = get_host_os_arch(rctx)
-    host_platform = get_host_platform(os_name, arch)
+    (os_name, arch) = _get_host_os_arch(rctx, logger)
```
aignas, line 247: If I remember correctly this code was written before rctx and mctx had the architecture. I think we could look into removing the dependency on uname here.

And just use the repo utils equivalents.

aignas (approved):
LGTM.

I like the suggestion to wrap the context and have extra methods, like log.warn or execute_unchecked. I would say that we can add a note in our issue list and consider it as a followup to this PR as this PR already improves things.

We could also unify the API by having watch method for bazel 6, so our code would become more modern.

python/private/repo_utils.bzl
rickeylev: oh? I didn't see that buildifier error pop up.

aignas:
Maybe it is just a warning... I'll paste a concrete example if I see it.

python/private/toolchains_repo.bzl
rickeylev: Ahh, that would explain a bit. I always wondered why we were calling uname when rctx seemed to have the equivalent.

PR #2083, opened 2024-07-22
refactor: move python/repositories.bzl implementation under python/private

rickeylev:
This is mostly for consistency with other code, but in doing so, it looks like there
are several symbols that have public visibility that _probably_ weren't meant to be
public. Cleaning that up is left for another PR in the future.


python/repositories.bzl
```diff
@@ -13,721 +13,26 @@
...
+)
+
+py_repositories = _py_repositories
+python_register_multi_toolchains = _python_register_multi_toolchains
+python_register_toolchains = _python_register_toolchains
+
+# These symbols are of questionable public visibility. They were probably
+# not intended to be actually public.
+STANDALONE_INTERPRETER_FILENAME = _STANDALONE_INTERPRETER_FILENAME
+http_archive = _http_archive
```
aignas, line 36: Pretty sure we can remove this, but I'm fine with being incremental here.

aignas (approved):

PR #2089, opened 2024-07-25
fix: Fix up search path of bootstrapped Python toolchain dylib

nicholasjng:
Previously, `//tests/cc/current_py_cc_libs::python_libs_linking_test` failed on macOS because the bootstrapped toolchain's dylib had an incorrect LC_ID_DYLIB field set, pointing to a local directory on the Python standalone build host machine.

To fix, add a small conditional to the Python repository rule patching the LC_ID_DYLIB field of the bootstrapped Python's dylib with its fully qualified file system path. Patching is carried out with macOS's own `install_name_tool`, which is part of the standard macOS dynamic linking toolchain.

Qualifies the mentioned test as executable on Mac, now only Windows linker errors are left to fix.

### DIFF
--- a/python/private/python_repositories.bzl
+++ b/python/private/python_repositories.bzl
@@ -191,6 +191,22 @@ def _python_repository_impl(rctx):
     elif rctx.attr.distutils_content:
         rctx.file(distutils_path, rctx.attr.distutils_content)
 
+    if "darwin" in platform and "osx" == repo_utils.get_platforms_os_name(rctx):
+        # Fix up the Python distribution's LC_ID_DYLIB field.
+        # It points to a build directory local to the GitHub Actions
+        # host machine used in the Python standalone build, which causes
+        # dyld lookup errors. To fix, set the full path to the dylib as
+        # it appears in the Bazel workspace as its LC_ID_DYLIB using
+        # the `install_name_tool` bundled with macOS.
+        dylib = "lib/libpython{}.dylib".format(python_short_version)
+        full_dylib_path = rctx.path(dylib)
+        repo_utils.execute_checked(
+            rctx,
+            op = "python_repository.FixUpDyldIdPath",
+            arguments = [repo_utils.which_checked(rctx, "install_name_tool"), "-id", full_dylib_path, dylib],
+            logger = logger,
+        )
+
     # Make the Python installation read-only. This is to prevent issues due to
     # pycs being generated at runtime:
     # * The pycs are not deterministic (they contain timestamps)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,10 @@ A brief description of the categories of changes:
   It would work well in cases to reduce merge conflicts.
 
 ### Fixed
+* (rules) Fixes build targets linking against `@rules_python//python/cc:current_py_cc_libs`
+  in host platform builds on macOS, by editing the `LC_ID_DYLIB` field of the hermetic interpreter's
+  `libpython3.x.dylib` using `install_name_tool`, setting it to its absolute path under Bazel's
+  execroot.
 * (rules) Signals are properly received when using {obj}`--bootstrap_impl=script`
   (for non-zip builds).
   ([#2043](https://github.com/bazelbuild/rules_python/issues/2043))
--- a/tests/cc/current_py_cc_libs/BUILD.bazel
+++ b/tests/cc/current_py_cc_libs/BUILD.bazel
@@ -20,12 +20,14 @@ current_py_cc_libs_test_suite(name = "current_py_cc_libs_tests")
 cc_test(
     name = "python_libs_linking_test",
     srcs = ["python_libs_linking_test.cc"],
-    # Mac and Windows fail with linking errors, but its not clear why; someone
-    # with more C + Mac/Windows experience will have to figure it out.
+    # Windows fails with linking errors, but its not clear why; someone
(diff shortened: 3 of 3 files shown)


python/private/python_repositories.bzl
```diff
@@ -191,6 +191,22 @@ def _python_repository_impl(rctx):
...
+        # host machine used in the Python standalone build, which causes
+        # dyld lookup errors. To fix, set the full path to the dylib as
+        # it appears in the Bazel workspace as its LC_ID_DYLIB using
+        # the `install_name_tool` bundled with macOS.
+        dylib = "lib/libpython{}.dylib".format(python_short_version)
+        full_dylib_path = rctx.path(dylib)
+        repo_utils.execute_checked(
+            rctx,
+            op = "python_repository.FixUpDyldIdPath",
+            arguments = [repo_utils.which_checked(rctx, "install_name_tool"), "-id", full_dylib_path, dylib],
```
aignas, line 206: This fix is only possible if the repository platform and host platform are the same.

Is it possible to solve this in a different way such that we can have correct behaviour when cross-building (e.g. building on a linux host platform with a configured toolchain to build against a mac platform?

nicholasjng: I'm not sure, since the dylib format is very Apple-specific. For example, according to this thread: https://forums.developer.apple.com/forums/thread/736719, at the very bottom, `install_name_tool` is mentioned as the canonical way of fixing up rpaths (and also IDs) after the build.

Another question is what to do when using another interpreter's lib - maybe the solution is rather setting `@rpath/lib/libpython3.11.dylib`, so that explicitly setting the RPATH to an interpreter (like e.g. `python@3.11` from something like Homebrew) would work.

python/private/python_repositories.bzl
```diff
@@ -191,6 +191,22 @@ def _python_repository_impl(rctx):
     elif rctx.attr.distutils_content:
         rctx.file(distutils_path, rctx.attr.distutils_content)
 
+    if "darwin" in platform:
```
aignas, line 194: In order to not attempt to use `install_name_tool` on platforms that don't have it (e.g. Debian), we ought to change this to do:
```suggestion
    if "darwin" in platform and "osx" == repo_utils.get_platforms_os_name(rctx):
```

This way this code path is going to only execute on OSX hosts when targeting the OSX toolchains.

For now this is the best we can do. This might also fix the Debian builds being broken.

aignas (changes requested):

python/private/python_repositories.bzl
nicholasjng: Thanks, done in https://github.com/bazelbuild/rules_python/commit/078f566a44a6c10424332fee7b279e6fc86dbaa5.

aignas (commented):
LGTM, could you please add a line in `CHANGELOG` to say that this behaviour was fixed?

@rickeylev, could you PTAL and see if something is missing from your point of view? IMHO this is good to merge as is, because there was a test and things started working better than they used to before the change.

nicholasjng:
I added a note to the changelog in the latest revision. Thanks for the speedy review!

nicholasjng:
Any more issues on your end @aignas @rickeylev ?

aignas (approved):
Thanks!

keith:
Doesn't setting this to an absolute path mean you get a non-hermetic build if you link this dylib?

keith:
https://github.com/bazelbuild/rules_python/pull/2256

brentleyjones:
This change breaks multi-platform builds. The contents of `libpython.dylib` differ depending on which host machine runs the build, since `if "darwin" in platform and "osx" == repo_utils.get_platforms_os_name(rctx)` checks the host machine not the execution platform. This change, if it has to happen conditionally, needs to happen in a build target instead.

```
$ diffoscope /tmp/local.libpython3.11.dylib /tmp/remote.libpython3.11.dylib 
--- /tmp/local.libpython3.11.dylib
+++ /tmp/remote.libpython3.11.dylib
├── arm64
│ ├── otool -arch arm64 -h {}
│ │ @@ -1,3 +1,3 @@
│ │  Mach header
│ │        magic  cputype cpusubtype  caps    filetype ncmds sizeofcmds      flags
│ │ - 0xfeedfacf 16777228          0  0x00           6    32       4544 0x00100085
│ │ + 0xfeedfacf 16777228          0  0x00           6    32       4552 0x00100085
│ ├── otool -arch arm64 -L {}
│ │ @@ -1,8 +1,8 @@
│ │ -   @rpath/libpython3.11.dylib (compatibility version 3.11.0, current version 3.11.0)
│ │ +   /install/lib/libpython3.11.dylib (compatibility version 3.11.0, current version 3.11.0)
│ │     /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1336.61.1)
│ │     /usr/lib/libncurses.5.4.dylib (compatibility version 5.4.0, current version 5.4.0)
│ │     /usr/lib/libpanel.5.4.dylib (compatibility version 5.4.0, current version 5.4.0)
│ │     /System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation (compatibility version 150.0.0, current version 2202.0.0)
│ │     /usr/lib/libobjc.A.dylib (compatibility version 1.0.0, current version 228.0.0)
│ │     /System/Library/Frameworks/SystemConfiguration.framework/Versions/A/SystemConfiguration (compatibility version 1.0.0, current version 1296.60.3)
```

PR #2095, opened 2024-07-26
tests: make toolchain tests run in the same build instead of bazel-in-bazel integration tests

rickeylev:
This changes the //tests/toolchains tests to run in the same Bazel build instance instead of
being bazel-in-bazel integration tests. This is done by using a transition to set
the python version to match the desired toolchain.

This makes running the tests **much** cheaper -- there were 37 sub-bazel processes being
started (one for each python version), which was very expensive. Debugging is also much easier because they aren't part of a bazel-in-bazel invocation.


rickeylev:
Ok, CI Should be happy now.

@aignas Do we happen to have existing code to map the `TOOL_VERSIONS/PLATFORMS` platform info stuff to `target_compatible_with` values?

Over in tests/toolchains/defs.bzl, I'm generating a target for each version. However, not all versions support all platforms, e.g., 3.8.10 doesn't support windows. So I need to generate e.g.

```
py_reconfig_test(
  name = "python_3.8.10_test",
    <is x86_64-apple-darwin>: [],
    <is x86_64-unknown-linux-gnu>: [],
    ...etc, one entry for each TOOL_VERSIONS sha256 key...
    "//condiitons:default": ":incompatible",
  }),
  ...
)
```

That config setting generation looks awfully familiar, as do the platform 2-tuple strings (e.g. "aarch64-apple-darwin", "x86_64-unknown-linux-gnu").

groodt (approved):
LGTM

tests/toolchains/defs.bzl
```diff
@@ -12,192 +12,40 @@
...
-            template = ctx.file._workspace_tmpl,
-            output = workspace,
-            substitutions = {"%python_version%": ctx.attr.python_version},
+    Args:
+        name: Only present to satisfy tooling.
+    """
+    for platform_key, platform_info in PLATFORMS.items():
+        native.config_setting(
+            name = "_is_{}".format(platform_key),
+            constraint_values = platform_info.compatible_with,
```
aignas, line 29: nit: Platforms also have `flag_values` (e.g. [here](https://github.com/bazelbuild/rules_python/blob/main/python/versions.bzl#L589)). And it would be good to just add them here as well.

For now it does not change how the tests perform, but once we have `musl` toolchains, it may affect things. It is also probably fine to add it then.

aignas (commented):
LGTM, thanks for simplifying!

tests/toolchains/defs.bzl
aignas: Created #2106

MODULE.bazel
```diff
@@ -76,6 +76,15 @@ bazel_dep(name = "rules_testing", version = "0.6.0", dev_dependency = True)
 bazel_dep(name = "rules_go", version = "0.41.0", dev_dependency = True, repo_name = "io_bazel_rules_go")
 bazel_dep(name = "gazelle", version = "0.33.0", dev_dependency = True, repo_name = "bazel_gazelle")
 
+dev_python = use_extension(
+    "//python/extensions:python.bzl",
+    "python",
+    dev_dependency = True,
+)
+dev_python.rules_python_private_testing(
+    register_all_versions = True,
```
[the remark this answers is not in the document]
rickeylev: Yeah, I had a similar thought. I'll post in slack.

Issue #2096, opened 2024-07-27
gazelle_python.yaml will always be generated whether integrity matches or not.


hunshcn:
# 🐞 bug report

### Affected Rule

`gazelle_python_manifest`

### Is this a regression?

Yes

### Description
After I upgraded rules_python for a project, I found that the ci of the project failed frequently. I noticed that I was restricted by frequent requests for pypi(mirror). failed with `gazelle_python_manifest.genrule`. (Because I test with `bazel test //...`

But this is not expected, because there is integrity in my gazelle_python.yaml.

https://github.com/bazelbuild/rules_python/pull/1666 make a change that a new genrule with diff_test.

this genrule and update target are not manual, so it will be built by running `bazel test //...`. 

This obviously goes against the purpose of integrity (we don't want to regenerate gazelle_python.yaml for the same requirements).

## 🔥 Exception or Error

<pre><code>Any repo</code></pre>

## 🌍  Your Environment

**Operating System:**

<pre><code>linux/macos</code></pre>

**Output of `bazel version`:**

<pre><code>7.2.1</code></pre>

**Rules_python version:**

<pre><code>0.34.0</code></pre>

**Anything else relevant?**

PR #2097, opened 2024-07-27
fix(gazelle): make gazelle_python_manifest.genrule manual

hunshcn:
Fix https://github.com/bazelbuild/rules_python/issues/2096


aignas (approved):
Thanks!

PR #2099, opened 2024-07-29
fix(gazelle): empty list

hunshcn:
This is a code error (from typo).


hunshcn:
Reported by @Whoaa512 .

I didn't modify changelog and test because although I believe this is a mistake, I didn't think of how to trigger this problem.

Did you trigger this error in some scenes? @Whoaa512

Whoaa512:
No, I just noticed because I was referencing at rules_python while building a new language plugin.

hunshcn:
Should I modify changelog? There should be no user-oriented modification.

aignas:
I think no need to change the changelog, FYI @dougthor42.

dougthor42 (approved):
Agreed, no update to changelog needed.

PR #2106, opened 2024-08-01
test(toolchain): use flag_values in test platform defs

aignas:
This PR uses the `flag_values` from the platform definitions
for the toolchains so that in the future we can distinguish
between `musl` and `glibc` toolchains in our tests.

For now the change is no-op.

As part of this change we are also registering the coverage
tools so that we can run `bazel coverage` with no errors.

See comment on #2095.

### DIFF
--- a/python/private/python.bzl
+++ b/python/private/python.bzl
@@ -267,7 +267,7 @@ def _create_toolchain_attr_structs(mod):
                 break
         if register_all:
             arg_structs.extend([
-                _create_toolchain_attrs_struct(tag = struct(python_version = v, configure_coverage_tool = True, is_default = False))
+                _create_toolchain_attrs_struct(python_version = v)
                 for v in TOOL_VERSIONS.keys()
                 if v not in seen_versions
             ])
--- a/python/private/python.bzl
+++ b/python/private/python.bzl
@@ -267,7 +267,7 @@ def _create_toolchain_attr_structs(mod):
                 break
         if register_all:
             arg_structs.extend([
-                _create_toolchain_attrs_struct(python_version = v)
+                _create_toolchain_attrs_struct(tag = struct(python_version = v, configure_coverage_tool = True, is_default = False))
                 for v in TOOL_VERSIONS.keys()
                 if v not in seen_versions
             ])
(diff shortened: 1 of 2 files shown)


python/private/python.bzl
```diff
@@ -267,7 +267,7 @@ def _create_toolchain_attr_structs(mod):
                 break
         if register_all:
             arg_structs.extend([
-                _create_toolchain_attrs_struct(python_version = v)
+                _create_toolchain_attrs_struct(tag = struct(python_version = v, configure_coverage_tool = True, is_default = False))
```
rickeylev, line 270: Did you verify this fixes the bazel@head failure? I don't see how it would.

This would cause additional version-aware toolchains to be registered, but those shouldn't match anyways. That it only fails with Bazel at head is also a bit strange -- this would imply something about toolchain resolution order or matching changed?

In any case, I'd prefer to just pass the python version, and we just modify the non-tag code path to use the values we want (i.e don't create a fake tag struct and pass it). Its only used for this "register everything for testing purposes" case.

rickeylev (approved):
Can you also update the description to better describe the change? A PR/issue number is nice, but I prefer commits to be more self-descriptive.

python/private/python.bzl
aignas: Ah, I guess my initial guess was wrong, I think it would be best to revert this part of the PR.
```suggestion
                _create_toolchain_attrs_struct(python_version = v)
```

aignas: OK, I did some more testing and it seems that the `bazel coverage` command works, but it is not doing if the `coverage_tool` is registered. Maybe there should be a test for this at some point.

However, if the `coverage_tool` is registered, then the usage of `python.runfiles` import in the tests breaks because the `coverage.py` package has a file named `python.py` and the importlib tells me that I cannot import from `python.runfiles` because `python` is not a module.

I think we have a ticket for the coverage issue.

Thinking about this a little more, I wonder if the fact that the toolchains are registered with `target_settings` that have `flag_values` and the test is running with a transition that does not have the `flag_values` changes anything. I think it should not, because the default build setting sets the `linux_libc` to `glibc` and it should work regardless.

aignas:
```suggestion
                _create_toolchain_attrs_struct(python_version = v)
```

PR #2107, opened 2024-08-02
docs: docgen underlying rules for macro-wrapped rules

rickeylev:
Rules that have a wrapper macro don't currently show all their attributes because the
macro hides the rule from stardoc. This also generates docs for the underlying rules and
has the macro reference them so that users can find all the attributes the rule accepts.


sphinxdocs/inventories/bazel_inventory.txt
```diff
@@ -0,0 +1,26 @@
...
+dict bzl:type 1 rules/lib/dict -
+label bzl:doc 1 concepts/labels -
+attr.bool bzl:type 1 rules/lib/toplevel/attr#bool -
+attr.int bzl:type 1 rules/lib/toplevel/attr#int -
+attr.label bzl:type 1 rules/lib/toplevel/attr#label -
+attr.label_list bzl:type 1 rules/lib/toplevel/attr#label_list -
+attr.string bzl:type 1 rules/lib/toplevel/attr#string -
+attr.string_list bzl:type 1 rules/lib/toplevel/attr#string_list -
+list bzl:type 1 rules/lib/list -
+python bzl:doc 1 reference/be/python -
```
aignas, line 21: nit: is this list supposed to be complete? Should we include other symbols here, like cc_library? Those could be useful in documenting the usage of headers.

aignas (approved):

python/py_binary.bzl
```diff
@@ -23,10 +23,21 @@ load("//python/private/common:py_binary_macro_bazel.bzl", _starlark_py_binary =
 _py_binary_impl = _starlark_py_binary if config.enable_pystar else native.py_binary
 
 def py_binary(**attrs):
-    """See the Bazel core [py_binary](https://docs.bazel.build/versions/master/be/python.html#py_binary) documentation.
+    """Creates an executable Python program.
+
+    This is the public macro wrapping the underlying rule. Args are forwarded
+    on as-is unless otherwise specified. See
+    {bzl:obj}`py_binary <//python/private/common:py_binary_rule_bazel.bzl%py_binary>`
```
aignas, line 30: The link in the rendered docs just says 'py_binary', it would be better to be more verbose here and have something similar to the following for this and other macroed rules.

```suggestion
    the underlying {bzl:obj}`py_binary <//python/private/common:py_binary_rule_bazel.bzl%py_binary>` rule
```

sphinxdocs/inventories/bazel_inventory.txt
rickeylev: Oops, accidentally added these files. Removed.

But yes, this is supposed to be listing of all the various things on the Bazel docs website that are crossreferable

python/py_binary.bzl
rickeylev: Done

PR #2111, opened 2024-08-06
fix(bzlmod): silence duplicate version registrations

### LINKED ISSUE #1818: [bzlmod] Silence `Ignoring toolchain` warning for duplicate toolchains with the same configuration
With our project using bzlmod, rules_python, and grpc, with this setup in the MODULE.bazel:

```bzl
python.toolchain(
    is_default = True,
    python_version = "3.11",
)
use_repo(python, "python_3_11")
```

When building you see this warning:

```
DEBUG: .../python.bzl:45:10: WARNING: Ignoring toolchain 'python_3_11' from module 'grpc': Toolchain 'python_3_11' from module 'foo' already registered Python version 3.11 and has precedence
```

In this case it seems that grpc also registers a 3.11 toolchain when using it from bzlmod https://github.com/bazelbuild/bazel-central-registry/blob/5958c06f2695493509c6b8ad64c2774b2e25925c/modules/grpc/1.56.3.bcr.1/patches/python.patch#L22-L26

Which produces this through this code https://github.com/bazelbuild/rules_python/blob/c5c03b2477dd1ce0c06c9dc60bf816995f222bcf/python/private/bzlmod/python.bzl#L144

In this case it doesn't seem like it's a problem, or that I can do anything about it? So I think it would be nice if it didn't produce a warning. Looking at the condition I am wondering if it is meant to be an `and` instead, since `first.module.is_root == True` in this case, but I don't understand the intent enough.

## 🌍  Your Environment

**Operating System:** macOS
**Output of `bazel version`:** 7.1.1
**Rules_python version:** HEAD

aignas:
Before this PR we would warn when non-root modules register interpreter
versions and the owners of root modules would have no way to silence the
warning except for patching rules_python.

This PR reduces the default verbosity of the warning to INFO to avoid
spamming users by default.

Fixes #1818.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,10 @@ A brief description of the categories of changes:
   [#2091](https://github.com/bazelbuild/rules_python/issues/2090).
 * (gazelle) Make `gazelle_python_manifest.update` manual to avoid unnecessary
   network behavior. 
+* (bzlmod): The conflicting toolchains during `python` extension will no longer
+  cause warnings by default. In order to see the warnings for diagnostic purposes
+  set the env var `RULES_PYTHON_REPO_DEBUG_VERBOSITY` to one of `INFO`, `DEBUG` or `TRACE`.
+  Fixes [#1818](https://github.com/bazelbuild/rules_python/issues/1818).
 
 ### Added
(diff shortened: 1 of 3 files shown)


rickeylev (approved):

rockpunk:
Thank you for this! These warnings were super annoying

PR #2112, opened 2024-08-08
fix: correctly handle absolute URLs in parse_simpleapi_html.bzl

WSUFan:
This PR addresses a typo and improves the handling of absolute URLs in the
`parse_simpleapi_html.bzl` file and corrects a minor issue in the
`simpleapi_download.bzl` file.

Summary:
1. parse_simpleapi_html.bzl:
   - Introduced a new private function `_get_root_directory(url)` to extract
     the root directory from a given URL.
   - Enhanced `_absolute_url` function to correctly handle absolute URLs by
     utilizing the `_get_root_directory` function. This ensures that URLs
     starting with a "/" are correctly resolved to their full path, avoiding
     potential incorrect concatenation.

2. simpleapi_download.bzl: Corrected the handling of the `real_url` variable in
   the `_read_simpleapi` function, ensuring that the correct URL is passed to
   `_read_index_result` when using non-blocking downloads.

### DIFF
--- a/python/private/pypi/parse_simpleapi_html.bzl
+++ b/python/private/pypi/parse_simpleapi_html.bzl
@@ -96,7 +96,25 @@ def parse_simpleapi_html(*, url, content):
         whls = whls,
     )
 
+def _get_root_directory(url):
+    scheme_end = url.find("://")
+    if scheme_end == -1:
+        fail("Invalid URL format")
+
+    scheme = url[:scheme_end]
+    host_end = url.find("/", scheme_end + 3)
+    if host_end == -1:
+        host_end = len(url)
+    host = url[scheme_end + 3:host_end]
+
+    return "{}://{}".format(scheme, host)
+
 def _absolute_url(index_url, candidate):
+    if candidate.startswith("/"):
+        # absolute url
+        root_directory = _get_root_directory(index_url)
+        return "{}{}".format(root_directory, candidate)
+
     if not candidate.startswith(".."):
         return candidate
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ A brief description of the categories of changes:
     * `3.12 -> 3.12.4`
 
 ### Fixed
+* (rules) correctly handle absolute URLs in parse_simpleapi_html.bzl.
 * (rules) Fixes build targets linking against `@rules_python//python/cc:current_py_cc_libs`
   in host platform builds on macOS, by editing the `LC_ID_DYLIB` field of the hermetic interpreter's
   `libpython3.x.dylib` using `install_name_tool`, setting it to its absolute path under Bazel's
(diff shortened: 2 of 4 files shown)


python/private/pypi/parse_simpleapi_html.bzl
```diff
@@ -96,7 +96,25 @@ def parse_simpleapi_html(*, url, content):
...
+
+    scheme = url[:scheme_end]
+    host_end = url.find("/", scheme_end + 3)
+    if host_end == -1:
+        host_end = len(url)
+    host = url[scheme_end + 3:host_end]
+
+    return "{}://{}".format(scheme, host)
+
 def _absolute_url(index_url, candidate):
```
aignas, line 112: I think it may be better to pass the `index_url_root` to the `_absolute_url` instead of getting the root directory all the time within this function. Since this may be called many times, I am worried that we will be doing many string manipulations for nothing.

It would be also good to have a test so that this does not regress, but I  am not sure if it is easy to do with the current structuring of the code.

WSUFan: Yes, the current implementation isn't the fastest. But do we really care about this here? While these string manipulations do require some CPU time, it's not significant given how fast modern CPUs are. Adding another argument to this function also makes the code feel less elegant.

For the test, I will add if possible. 

Thanks for the review.

aignas:
Could you please add a test in https://github.com/bazelbuild/rules_python/blob/3b183bfedd3dd6a5bb42c77531079e347165441d/tests/pypi/parse_simpleapi_html/parse_simpleapi_html_tests.bzl so that this does not regress?

WSUFan:
> Could you please add a test in https://github.com/bazelbuild/rules_python/blob/3b183bfedd3dd6a5bb42c77531079e347165441d/tests/pypi/parse_simpleapi_html/parse_simpleapi_html_tests.bzl so that this does not regress?
> 
> And please add a note in the Changelog.md file.

Yeah. Sure. I will do it. Thanks!

WSUFan:
> Could you please add a test in https://github.com/bazelbuild/rules_python/blob/3b183bfedd3dd6a5bb42c77531079e347165441d/tests/pypi/parse_simpleapi_html/parse_simpleapi_html_tests.bzl so that this does not regress?
> 
> And please add a note in the Changelog.md file.

Hi @aignas, I have added tests and changed the Changelog.md. Please take a look when you get a chance. Let me know if you have other concerns. Thanks!

aignas (approved):
LGTM, thank you.

The current structure is alright, we can optimize the code if we see any problems later. At the end of the day, when using the `MODULE.bazel.lock` file, this code will be executed only when the PyPI dependencies change and the time spent querying the PyPI server itself is greater than the time spent parsing it.

Thank you very much for the contribution!

PR #2115, opened 2024-08-09
fix: use runfiles strategy to get runfiles root

mattem:
When attempting to calculate the current repository, the runfiles helper used a strategy of walking up the current files directory four times, however this fails to correctly resolve the runfiles root when the runfiles helper is consumed via the PyPi package and linked into a virtual environment (the root will be determined to be the venv `lib` directory).

This change makes it so the runfiles root is fetched from the helper strategy, which is either created with the `$RUNFILES_ROOT` environment variable (directory based), or calculates the root via the runfiles manifest (manifest based).

With these changes, the runfiles helper correctly identifies the runfiles root when linked into a Python virtual environment.

### DIFF
--- a/python/runfiles/runfiles.py
+++ b/python/runfiles/runfiles.py
@@ -95,6 +95,9 @@ class _DirectoryBased:
             raise TypeError()
         self._runfiles_root = path
 
+    def _GetRunfilesDir(self) -> str:
+        return self._runfiles_root
+
     def RlocationChecked(self, path: str) -> str:
         # Use posixpath instead of os.path, because Bazel only creates a runfiles
         # tree on Unix platforms, so `Create()` will only create a directory-based
@@ -118,7 +121,7 @@ class Runfiles:
 
     def __init__(self, strategy: Union[_ManifestBased, _DirectoryBased]) -> None:
         self._strategy = strategy
-        self._python_runfiles_root = _FindPythonRunfilesRoot()
+        self._python_runfiles_root = self._strategy._GetRunfilesDir()
         self._repo_mapping = _ParseRepoMapping(
             strategy.RlocationChecked("_repo_mapping")
         )
@@ -320,19 +323,6 @@ class Runfiles:
 
 # Support legacy imports by defining a private symbol.
 _Runfiles = Runfiles
-
-
-def _FindPythonRunfilesRoot() -> str:
-    """Finds the root of the Python runfiles tree."""
-    root = __file__
-    # Walk up our own runfiles path to the root of the runfiles tree from which
-    # the current file is being run. This path coincides with what the Bazel
-    # Python stub sets up as sys.path[0]. Since that entry can be changed at
-    # runtime, we rederive it here.
-    for _ in range("rules_python/python/runfiles/runfiles.py".count("/") + 1):
-        root = os.path.dirname(root)
-    return root
-
 
 def _ParseRepoMapping(repo_mapping_path: Optional[str]) -> Dict[Tuple[str, str], str]:
     """Parses the repository mapping manifest."""
--- a/tests/runfiles/runfiles_test.py
+++ b/tests/runfiles/runfiles_test.py
@@ -527,7 +527,7 @@ class RunfilesTest(unittest.TestCase):
             expected = ""
         else:
             expected = "rules_python"
-        r = runfiles.Create({"RUNFILES_DIR": "whatever"})
+        r = runfiles.Create({"RUNFILES_DIR": os.environ.get("RUNFILES_DIR")})
         assert r is not None  # mypy doesn't understand the unittest api.
         self.assertEqual(r.CurrentRepository(), expected)
 


alexeagle:
@fmeum FYI as you're the RUNFILES resolution expert :)

python/runfiles/runfiles.py
```diff
@@ -321,19 +324,6 @@ def Create(env: Optional[Dict[str, str]] = None) -> Optional["Runfiles"]:
 # Support legacy imports by defining a private symbol.
 _Runfiles = Runfiles
 
-
-def _FindPythonRunfilesRoot() -> str:
-    """Finds the root of the Python runfiles tree."""
```
fmeum, line 326: I remember that I introduced this function instead of using `RUNFILES_DIR` since I thought that the Python runfiles tree may be different from the actual runfiles directory (the one with non-Python data dependencies) when using `--build_python_zip`. Do you know whether that's true? If this isn't true, I at least don't understand why the runfiles library would have to deal with `RUNFILES_MANFIEST_FILE` at all as the Python runfiles are always arranged in a directory.

rickeylev: Yeah, that sounds vaguely familiar.

It should be easy to test these cases now. Over in tests/base_rules/BUILD.bazel, there's some existing tests for setting these zip flags.

fmeum: Indeed, some of the failures seem to show this.

@mattem Could you share an example of the venv layout? We can try to find logic that works in both cases.

mattem: @fmeum The venv created by rules_py when running a binary or test is rooted at the root of the targets runfiles tree, and takes the standard Python venv layout. For example, a target named `foo` in the `//py/tests` package depending on the PyPi version of the runfiles helper will have the following package in the `site-packages` directory.

```
bazel-bin/py/tests/external-deps/foo.runfiles/.foo.venv/lib/python3.9/site-packages/runfiles
```

The `runfiles` last segment of the path is the Python package name.

There is also the case where the user plants the venv directory in an arbitrary place in the source tree for IDE support, at that point `.foo.venv` could technically be anywhere, but likely at the source tree root or in the same package as the target. Users then likely "run" the binary via the IDE for debugging etc.

fmeum: What would the path under the venv of some Python package from an external repository look like?

mattem: As above,

```
bazel-bin/py/tests/external-deps/foo.runfiles/.foo.venv/lib/python3.9/site-packages/runfiles
╰─────────────────────┬─────────────────────╯╰───┬───╯╰─────────────┬─────────────╯╰───┬───╯
            bazel runfiles root              venv root    Python packages root   Python package
```

The symlink is from the bazel runfiles tree into the `site-packages` folder.

fmeum: I see. This structure doesn't provide enough information to extract the current repository name out of it. I don't see how to easily salvage this situation. Could whatever is creating these venvs write out a file with a mapping of python packages to Bazel repo names?

mattem: rules_py can write any file into the venv during creation, as long as it can source that data from somewhere.

fmeum: For each Python package coming from a Bazel module (not an external pure-Python dep), it would need to write that package name and the corresponding canonical repo name (`Label.workspace_name`) to a file. 

Then we could look for that file here, find the package name in the path of the current source file and obtain the repo name from it. Feel free to mention me on the PR (draft) that sets up the file if you need help.

mattem: How are we able to tell that a dependency comes from a bazel module vs an external Python dependency? They will have that `Label.workspace_name` set?

fmeum: Good point, I guess your external deps can come from everywhere? In that case you could also just store the mapping between package and `Label.workspace_name` for every package. That would match the logic here if it weren't broken due to the different runfiles structure.

mattem: Yes, dependencies can come from both external bazel dependencies or fetched from PyPi (etc) by rules_python. In fact in rules_py, there is very little difference between 1p and 3p dependencies. 

> That would match the logic here if it weren't broken due to the different runfiles structure.

To be clear, the runfiles helper only fails when it is consumed via the PyPi dependency, it works as expected when used via `@bazel_tools//tools/python/runfiles`, as I guess it doesn't attempt to fetch the repo name? Secondly, the "normal" runfiles structure is there, it's just up a few more directory segments.

fmeum: > To be clear, the runfiles helper only fails when it is consumed via the PyPi dependency, it works as expected when used via `@bazel_tools//tools/python/runfiles`, as I guess it doesn't attempt to fetch the repo name?
The runfiles helper in `@bazel_tools//tools/python/runfile` is deprecated and not fully functional with Bzlmod.

> Secondly, the "normal" runfiles structure is there, it's just up a few more directory segments.
The subtle part is that this failure isn't about not finding the runfiles strategy, it's about identifying the point of reference for lookups in that structure. For example, when you request `rules_python/foo/bar` via `Rlocation`, you need to know which Bazel repo is asking for that to look up the actual repo (and thus its canonical name) seen under that name by the caller. 

If we had a separate file that maps Python package names to Bazel repo names, we could get the Python package name from the path and thus the Bazel repo name relative to which runfiles paths would be interpreted.

thesayyn: There is another solution to this problem maybe?. here in this line, https://github.com/bazelbuild/rules_python/blob/18f76f9cc4c3c6c4470f7a881e6ea4e8b80d7bab/python/runfiles/runfiles.py#L352 if we perform a readlink we will just get the value that runfiles library expects, however that also creates a problem where it follows back out of sandbox.

thesayyn: for the interim, i came up with this patch to rules_py as a workaround; https://github.com/aspect-build/rules_py/pull/519 

that said i am curious how this affect python zips, i have created a pex rule in rules_py which did not need a special handling for runfiles.

PR #2117, opened 2024-08-11
docs: move Sphinx object inventory for Bazel under sphinxdocs

rickeylev:
This moves the inventory file for Bazel objects under the sphinxdocs directory. Inventory
files are the manifest of things that can be cross referenced from some external source.

They're moved under sphinxdocs since these are about Bazel in general and aren't specific
to rules_python.


aignas (approved):
LGTM.

PR #2120, opened 2024-08-13
feat(gazelle): Update resolve.go to provide more human-friendly error output

alex-torok:
Update the Resolver error messages in resolve.go to be easier to read.

While enabling gazelle in an existing repo, I found it somewhat difficult to skim the error messages. This PR:
1. Adds a bit of whitespace to spread different pieces of information out
2. Provides an additional remediation action ambiguous import resolution
3. Moves the filename and line number where the error is to the front of the line.
4. Provides more concrete "gazelle:resolve py" directive examples for each errored dep resolution.

Additionally, update a testcase to showcase the ambiguous import resolution error case. See the `test.yaml` file diff for an example of the old vs new error output.


aignas:
LGTM, could you please rebase and add a changelog.md item?

alex-torok:
Rebased & changelog entry added.

aignas (approved):
Thank you!

PR #2127, opened 2024-08-18
fix(bzlmod): duplicate repos when using experimental_index_url with different requirment files

ewianda:
It is not very obvious to me what is causing duplication in repo_name but this change fixes the error.


python/private/pypi/whl_library.bzl
```diff
@@ -228,7 +228,7 @@ def _whl_library_impl(rctx):
         else:
             # It is an sdist and we need to tell PyPI to use a file in this directory
             # and not use any indexes.
-            extra_pip_args.extend(["--no-index", "--find-links", "."])
+            extra_pip_args.extend(["--find-links", "."])
```
ewianda, line 231: This is a different PR, will rebase once that is merged.

PR #2134, opened 2024-08-20
fix: formatting directive bugs in test

tokongs:
I'm getting a build error in my builds because nogo complains about this bug. I guess I could configure my nogo to no check this code, but I think it should be fixed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,9 @@ A brief description of the categories of changes:
 ### Changed
 * Nothing yet
 
+### Fixed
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -27,6 +27,9 @@ A brief description of the categories of changes:
 ### Changed
 * Nothing yet
 
+### Fixed
+* (gazelle): Fix ncorrect use of `t.Fatal`/`t.Fatalf` in tests.
```
aignas, line 31:
```suggestion
* (gazelle): Fix incorrect use of `t.Fatal`/`t.Fatalf` in tests.
```

aignas (approved):

PR #2135, opened 2024-08-20
fix(bzlmod): keep the lockfile platform independent when resolving python

### LINKED ISSUE #1105: pip_parse does not respect requirement markers

aignas:
Before this PR the lockfile would become platform dependent when the
`requirements` file would have env markers. This was not caught because
we do not have MODULE.bazel.lock checked into the `rules_python`
repository because the CI is running against many versions and the lock
file is different, therefore we would not be able to run with
`bazel build --lockfile_mode=error`.

With this change we use the label to `BUILD.bazel` which is living next
to the `python` symlink and since the `BUILD.bazel` is the same on all
platforms, the lockfile will remain the same.

Summary
* refactor(uv): create a reusable macro for using uv for locking reqs.
* test(bzlmod): enable testing the MODULE.bazel.lock breakage across platforms.
* test(bzlmod): use a universal requirements file for 3.9.
  This breaks the CI, because the python interpreter file hash is added to the lock file.
* fix(bzlmod): keep the lockfile platform independent when resolving python

Fixes #1105 and #1868 for real this time.
Implements an additional helper for #1975.


aignas:
@fmeum do you have a better idea of how we could effectively do
`mctx.path(python_interpreter_label)` without writing the file resulting file
hash to the lockfile? In this case we do not care about the python interpreter
and the code would work equally well even if it was `python3` (taken from the
PATH).

@rickeylev, do you have ideas how we could test this? Should we have one of the
examples using the lockfile and `--lockfile_mode=error`? I am not fully sure
how that plays with testing against the `latest_green`.

fmeum:
I don't see a better way at the moment. We should probably add a `watch` parameter to `mctx.path` though.

Cc @Wyverald

Wyverald:
> We should probably add a `watch` parameter to `mctx.path` though.

Yeah, that seems fair.

CHANGELOG.md
```diff
@@ -33,6 +33,11 @@ A brief description of the categories of changes:
 ### Removed
 * Nothing yet
 
+### Fixed
+* (bzlmod) get the path to the host python interpreter without calling
+  `mctx.path` on Labels that have differing contents on different OSes.
+* (bzlmod) correctly watch sources when using `pypi_repo_utils`.
```
rickeylev, line 39: Can you rephrase this so it isn't using internal jargon? "pypi_repo_utils" is an internal thing not relevant to users.

python/private/pypi/pypi_repo_utils.bzl
```diff
@@ -110,7 +123,7 @@ def _execute_checked(mrctx, *, srcs, **kwargs):
         # This will ensure that we will re-evaluate the bzlmod extension or
         # refetch the repository_rule when the srcs change. This should work on
         # Bazel versions without `mrctx.watch` as well.
-        repo_utils.watch(mrctx.path(src))
+        repo_utils.watch(mrctx, mrctx.path(src))
```
rickeylev, line 126: Ahh, hah. I see how this got past CI. We pass along *args and do a conditional check for backwards compatiblity, and do nothing if it doesn't look supported.

It probably makes sense to check type(mrctx) in the watch function to help catch this in the future; can go into another PR, though

python/uv/private/pin.bzl
```diff
@@ -0,0 +1,184 @@
...
+# limitations under the License.
+
+"""A simple macro to pin the requirements.
+"""
+
+load("@bazel_skylib//rules:write_file.bzl", "write_file")
+load("//python:py_binary.bzl", "py_binary")
+load("//python/private:bzlmod_enabled.bzl", "BZLMOD_ENABLED")  # buildifier: disable=bzl-visibility
+
+_REQUIREMENTS_TARGET_COMPATIBLE_WITH = select({
```
rickeylev, line 22: Does uv only provide linux/mac binaries...?

python/uv/private/pin.bzl
```diff
@@ -0,0 +1,184 @@
...
+def _transition_python_version_impl(_, attr):
+    return {"//python/config_settings:python_version": str(attr.python_version)}
+
+_transition_python_version = transition(
+    implementation = _transition_python_version_impl,
+    inputs = [],
+    outputs = ["//python/config_settings:python_version"],
+)
+
+def _impl(ctx):
```
rickeylev, line 140: Can you use the code from //python/config_settings:transition.bzl instead? You should be able to just load py_binary from there and set python_version.

docs/BUILD.bazel
```diff
@@ -140,71 +139,12 @@ sphinx_build_binary(
...
-    "@platforms//os:windows": [],
-    "//conditions:default": ["@platforms//:incompatible"],
-}) if BZLMOD_ENABLED else ["@platforms//:incompatible"]
-
-# Run bazel run //docs:requirements.update
-genrule(
-    name = "requirements",
+# Run bazel run //docs:requirements_update
+pin(
+    name = "requirements_update",
```
rickeylev, line 144:
```suggestion
    name = "requirements.update",
```

I'd like to retain the `:requirements` and `requirements.update` names as part of the API. It's already established by pip_compile(), deviating from that just seems needless, and retaining it makes switching easier.

I realize this doesn't define a `:requirements` target, which is fine for now, but we can at least keep the `:requirements.update` name.

rickeylev:
re: testing: I think --lockfile_mode=error in the example is OK. The other option is an integration test that...I guess just sets --lockfile_mode? If there's something in particular to look for in the output, the python runner can be used to read the lock file and look for that.

I think we'll just have to see how it plays out with the downstream bazel testing. If it causes too many bogus failures, then we can factor out a smaller test under a separate CI job that sets `skip_in_bazel_downstream_pipeline`.

rickeylev (approved):
Functionally, looks fine, so approved. Just comments about some doc phrasing and code reuse.

docs/BUILD.bazel
```diff
@@ -140,71 +139,12 @@ sphinx_build_binary(
...
-    "@platforms//os:macos": [],
-    "@platforms//os:windows": [],
-    "//conditions:default": ["@platforms//:incompatible"],
-}) if BZLMOD_ENABLED else ["@platforms//:incompatible"]
-
-# Run bazel run //docs:requirements.update
-genrule(
-    name = "requirements",
+# Run bazel run //docs:requirements_update
+pin(
```
groodt, line 143: We're still keeping the api private right?

I'm not sold on the `pin` name. I think you were the one who actually convinced me the word should be `lock`. But happy to bikeshed names another time.

aignas: I'll change it to `lock`, `pin` was just a placeholder, thanks for reminding!

python/uv/private/pin.bzl
aignas: No, `windows` binaries are also provided, just the `genrule` implementation is in shell.

python/private/pypi/pypi_repo_utils.bzl
```diff
@@ -51,7 +51,20 @@ def _resolve_python_interpreter(mrctx, *, python_interpreter = None, python_inte
...
-        python_interpreter = mrctx.path(python_interpreter_target)
+        # The following line would make the MODULE.bazel.lock platform
+        # independent, because the lock file will then contain a hash of the
+        # file so that the lock file can be recalculated, hence the best way is
+        # to add this directory to PATH.
+        #
+        # hence we add the root BUILD.bazel file and get the directory of that
+        # and construct the path differently. At the end of the day we don't
+        # want the hash of the interpreter to end up in the lock file.
+        if hasattr(python_interpreter_target, "same_package_label"):
```
aignas, line 62: FYI @UebelAndre, this may help in https://github.com/bazelbuild/rules_rust/issues/3297

Issue #2118, opened 2024-08-11
bzlmod example fails to build on pip install error when building whls from sdists


rockpunk:
# 🐞 bug report

### Affected Rule

examples/bzlmod: `//:all_data_requirements`

### Is this a regression?

Yes, the previous version in which this bug was not present was: commit e682cd0e970e11d66931b9b4b70a7f50989e7789

### Description

Seems like PR #2091 breaks the bzlmod example on mac arm64. 

```
❯ ( cd examples/bzlmod && bazel build //:all_data_requirements )
INFO: Repository rules_python~~pip~pip_39_lazy_object_proxy_sdist_78247b6d instantiated at:
  <builtin>: in <toplevel>
Repository rule whl_library defined at:
  /private/var/tmp/_bazel_slayland/96ddcf3f8cdebe38a185655c2901b33f/external/rules_python~/python/private/pypi/whl_library.bzl:451:30: in <toplevel>
   
  ~~~~  snip ~~~~~~

===== stderr end =====
ERROR: Analysis of target '//:all_data_requirements' failed; build aborted: Analysis failed
INFO: Elapsed time: 2.092s, Critical Path: 0.01s
INFO: 1 process: 1 internal.
ERROR: Build did NOT complete successfully
```

## 🔬 Minimal Reproduction

NOTE: commit 6f9082fda020a9970d624e54ad3d964d64957f21 was isolated via `git bisect`

<details><summary>git bisect output</summary>

```
❯ git bisect good
6f9082fda020a9970d624e54ad3d964d64957f21 is the first bad commit
commit 6f9082fda020a9970d624e54ad3d964d64957f21
Author: Chris Chua <chris.sirhc@gmail.com>
Date:   Fri Jul 26 11:55:03 2024 +0800

    fix: use downloaded archive in sdist (#2091)

    Before this PR the extra arguments added by the `experimental_index_url`
    code paths were not used and the `sdist` was being redownloaded every
    time
    we would build from `sdist`.

    This PR fixes the code to not ignore the extra args added in

    https://github.com/bazelbuild/rules_python/pull/2091/files#diff-c007ed21502bf8ea19b98b3f1b402e7071615f8520e4291b00a71bca2cd451e8R231

    Fixes #2090

    ---------

    Co-authored-by: aignas <240938+aignas@users.noreply.github.com>

 CHANGELOG.md                        | 3 +++
 python/private/pypi/whl_library.bzl | 2 +-
[...]

PR #2142, opened 2024-08-22
feat: allow overriding rules in `whl_library` repository rule

aignas:
WIP - this is to discuss an interface for allowing users to better customize
the build process. Once this is defined, we can then implement something for
allowing users to build from source in the build context. That would require
the following missing pieces that should be implemented in other PRs:
- [ ] Using `http_archive` or implement a new `sdist_library` repository rule
  for handling `sdist`.
- [ ] A rudimentary `py_sdist_library` that allows invoking a [PEP518] build
  backend in the build phase.

Related to the discussion in #2118.

[pep518]: https://peps.python.org/pep-0518/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,13 @@ A brief description of the categories of changes:
 * (gazelle): Fix incorrect use of `t.Fatal`/`t.Fatalf` in tests.
 
 ### Added
-* Nothing yet
+* (pypi): Add macro wrappers for all of the publicly exposed targets in the
+  `whl_library` repository_rule. This allows users to override rules used to
+  extract whl targets from the `.whl` distributions. This allows `WORKSPACE`
+  and `bzlmod` users to override the load statements. The `WORKSPACE` can pass
+  an extra argument named `override_loads` to the `pip_install` macro from the
+  hub repository, whereas `bzlmod` users have a new attribute `load_symbols` in
+  the `pip.override` tag class. TODO: link to the ticket and add documentation.
 
 ### Removed
 * Nothing yet
--- a/examples/bzlmod/MODULE.bazel
+++ b/examples/bzlmod/MODULE.bazel
@@ -195,6 +195,12 @@ pip.parse(
 # The patches have to be in the unified-diff format.
 pip.override(
     file = "requests-2.25.1-py2.py3-none-any.whl",
+    # One can also override the loads for better support of using alternative implementations
+    # For allowed values, inspect the error message when typing any unsupported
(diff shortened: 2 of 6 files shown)


rickeylev (commented):
I don't think this is going to work because of bzlmod visibility. rules_python only has visibility to the repos in its module.bazel, so if another repo is rendered into the generated code, it won't be visible.

----

I was, perhaps naively, thinking to have, quite literally, a single macro call in the generated code:
```
load("@rules_python//whatever.bzl", "define_pypi_package")

define_pypi_package(
  name = "foo",
  ...
)
```

Though admittedly, I hadn't thought through what all the args would be and if such a call wouldn't be overly complicated.

In any case, the idea behind doing it as one macro was to allow additional targets to be defined. The cases I had in mind were to define a cc_library or entry point.

A second reason for doing it this way to avoid having lots of symbols to override. Though, I suppose on advantage of naming the specific symbols to load is you can reduce how much is loaded.

Perhaps splitting the difference? Have a single macro per "type" of pypi package that gets defined. e.g., a wheel based package needs to load the `define_pypi_package_from_wheel(...)` symbol; an sdist-based package needs to load the `define_pypi_from_sdist(...)` symbol.

I'm feeling pretty -1 on the granularity in this PR, though, e.g. filegroup, py_library, etc. Those really seem like implementation details of "define a pypi package"

aignas:
I tried implementing it with bigger granularity, but then the question is what gets passed as `whl` and `py_library` deps because those could be something like
```
deps = ["//foo:pkg"] + select({
    "@os:linux": ["//bar:pkg"],
    "default": [],
})
```

And because of that reason we cannot just pass `foo` and `bar` as deps. Maybe we should see what happens if we do pass `foo` and `bar` and omit the implementation details of the selects to the rule, but that may be a step back in terms of abstractions.

rickeylev:
I think I need to go look at the generated BUILD files to get a better sense of things...

Maybe step one should be refactoring so that the generated build file is simpler? See local_runtime_repo for what I mean: https://github.com/bazelbuild/rules_python/blob/main/python/private/local_runtime_repo.bzl#L32

The generated BUILD file is super simple and just calls a macro with the minimal args. You can just look at the `define_local_runtime_toolchain_impl` macro to see what ends up in that build file and don't have to dig out the generated build file.

-----

I don't see why we can't pass a select()? The `define_pypi_package` function would be a macro. Under the hood it calls the rule and passes the deps object.

aignas:
#2347 has replaced this when it comes to better abstractions. Whereas the overriding of the targets is a separate question.

Issue #2150, opened 2024-08-23
(bzlmod) Index responses containing relative paths for wheels causes bazel downloader to fail.


WillMorrisonEnspi:
# 🐞 bug report

### Affected Rule

The issue is caused by: [`_create_wheel_repos`](https://github.com/bazelbuild/rules_python/blob/b679a794c01c97a82deb2857c374474e32d4c263/python/private/pypi/extension.bzl#L102) passing index-relative paths to [`whl_library`](https://github.com/bazelbuild/rules_python/blob/b679a794c01c97a82deb2857c374474e32d4c263/python/private/pypi/extension.bzl#L310) instead of full URLs.

### Is this a regression?

Probably not.

### Description

The current implementation of `whl_library` passes URLs parsed from `simpleapi_download` verbatim to `rctx.download`. However, some indexes (like AWS CodeArtifact) return relative URLs. This causes the bazel downloader to choke when `experimental_index_url` is set to an index that returns relative URLs.

<details>
    <summary><b>Expand to show GET response for private index vs pypi</b></summary>

`curl --basic -u "aws:$auth_token" 'https://<redacted>.d.codeartifact.eu-central-1.amazonaws.com/pypi/packages-prod/simple/mypy-extensions/'`
```html
<!DOCTYPE html>
<html>
  <head>
    <title>Links for mypy-extensions</title>
  </head>
  <body>
    <h1>Links for mypy-extensions</h1>
      <a href="0.1.0/mypy_extensions-0.1.0.tar.gz#sha256=6942c0429e505a0cdcd0468fe897960fbd8113f84cb9bde1540120bc8b68eca9"  data-gpg-sig="false" >mypy_extensions-0.1.0.tar.gz</a>
      <br/>
      <a href="0.2.0/mypy_extensions-0.2.0.tar.gz#sha256=b8f40b3bca27843f6ca911d63550799aa2586804d40da3ab087bab8130da3ebf"  data-gpg-sig="false" >mypy_extensions-0.2.0.tar.gz</a>
      <br/>
      <a href="0.3.0/mypy_extensions-0.3.0.tar.gz#sha256=c4553f2475996c912658d3c2f398766cd47a65e0c2c6bb27ac8dea94a21640f4"  data-gpg-sig="false" >mypy_extensions-0.3.0.tar.gz</a>
      <br/>
[...]

PR #2153, opened 2024-08-24
fix: Handle relative paths properly in _absolute_url

WillMorrison:
This updates the simpleapi parser to handle indexes where wheel and sdist may be an index_url relative path. It also organises the conditionals with fewer negations so they're easier to read

Fixes: #2150


aignas (approved):
Thanks!

PR #2155, opened 2024-08-24
chore: remove proto library tests from CI

aignas:
Cleanup after proto libraries moved out of rules_python.


groodt (approved):
LGTM

rickeylev:
It looks like we still have a full implementation in rules_python, though? I figured python/proto.bzl would load from the protobuf repo at some time to transition

aignas:
Ah, I forgot we reverted the py_proto_library removal. Then let's wait until that change gets merged again.

PR #2160, opened 2024-08-27
fix(gazelle): Correctly resolve deps that have top-level module overlap with a gazelle_python.yaml dep module

alex-torok:
In https://github.com/bazelbuild/rules_python/pull/2048, `FindThirdPartyDependency` was updated to walk up the module import path to try to find a match in gazelle_python.yaml. This is unnecessary, as the main resolve loop will accomplish the same task.

Additionally, the change broke the ability to configure a more specific resolve override or resolve more specific indexed libraries.

For a real-world example of where this is a problem, `pytype` has a `third_party` module at its top-level. In a repo that also has a `third_party` directory, we can no longer resolve our indexed libraries in `third_party`. When the resolve loop tried to resolve `third_party.foo.bar`, `FindThirdPartyDependency` will immediately match `third_party` and not give the resolve loop a chance to look in the rule index for `third_party.foo.bar`.

The same issue appears for providing overrides that are more specific (see the updated testcase).

This PR reverts the change to FindThirdPartyDependency and updates the testcases to ensure that we can still resolve specific indexed packages, explicit resolve overrides, and third party modules even when there is an overlap in the top-level module name.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@ A brief description of the categories of changes:
 * (toolchain) Omit third-party python packages from coverage reports from
   stage2 bootstrap template.
 * (bzlmod) Properly handle relative path URLs in parse_simpleapi_html.bzl
+* (gazelle) Correctly resolve deps that have top-level module overlap with a gazelle_python.yaml dep module
 
 ### Added
 * Nothing yet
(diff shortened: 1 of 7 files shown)


alex-torok:
FYI @aignas since you reviewed https://github.com/bazelbuild/rules_python/pull/2048.

aignas (approved):
Thanks

PR #2165, opened 2024-08-29
fix: correctly check the arg count in precompiler.py

oprypin:
This is supposed to check that all argument counts are equal, but due to an incorrect inversion of the conditional, it was checking that *any* pair of argument counts is equal


rickeylev (approved):

rickeylev:
Thanks! I forgot that comparison chaining implicitly uses `and` between the pieces, but a negative comparison doesn't "expand" (I forget the logic term, commute? transitive?) like that, it needs to use `or` to be logically equivalent.

PR #2166, opened 2024-08-29
feat(rules): add PyExecutableInfo

rickeylev:
The PyExecutableInfo provider exposes executable-specific information that isn't easily accessible outside the target. The main purpose of this provider is to facilitate packaging a binary or deriving a new binary based upon the original.

Within rules_python, this will be used to pull the zip-building logic out of executables and into separate rules. Within Google, this will be used for a similar "package a binary" tool.

Along the way:
* Add runfiles references to sphinx inventory

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,7 +46,9 @@ A brief description of the categories of changes:
 * (gazelle) Correctly resolve deps that have top-level module overlap with a gazelle_python.yaml dep module
 
 ### Added
-* Nothing yet
+* (rules) Executables provide {obj}`PyExecutableInfo`, which contains
+  executable-specific information useful for packaging an executable or
+  or deriving a new one from the original.
 
 ### Removed
 * Nothing yet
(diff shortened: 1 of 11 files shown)


aignas (approved):

aignas:
Just brainstorming here.

This might be also beneficial for rules_oci users who would like to create a tarball with a different layout of the files.

I think in a similar fashion we could also list sources for each third party package so that users can also do whatever they want with that. Having some sort of depset map could be useful. That way people would not need to rely on site-packages string being in the path and we could also have a venv building rule.

rickeylev:
Yes, packaging a py_binary into a tar is the sort of use case I have in mind. That's not much different that a zip file, really. 

rules_oci + reorganizing site-packages gives me another use case idea: packaging multiple binaries into an image and reducing duplication. So don't include the python runtime in `runfiles_without_exe` (it's already available through PyRuntimeInfo), and have a way to identify which pip.parse hub things came from (to create a site-packages per hub).

PR #2167, opened 2024-08-30
refactor(flags): return FeatureFlagInfo in --python_version flag

rickeylev:
Make the `--python_version` flag also return the `FeatureFlagInfo` provider.

There are two reasons to also return the FeatureFlagInfo provider:

First, it allows the flag implementation to change the value and have that value respected
by config_setting() later. This allows, for example, the rule to use custom logic (and
information from things it depends on) to determine the effective flag value.

Secondly, it makes the flag compatible with the Google fork of this rule, which is 
implemented using FeatureFlagInfo, to help eventually converge them.

Along the way, add config_common to the Sphinx Bazel inventory.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,8 @@ A brief description of the categories of changes:
 
 ### Changed
 * (gazelle): Update error messages when unable to resolve a dependency to be more human-friendly.
+* (flags) The {obj}`--python_version` flag now also returns
+  {obj}`config_common.FeatureFlagInfo`.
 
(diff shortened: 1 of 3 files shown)


aignas (approved):
Mayeb we should note that we now return a FeatureFlag from the python version in the changelog. Rules_pycross had this implemented on top of rules_python, so maybe now it is possible to reduce the duplication?

Edit: unqueued the PR in case you would like to add the changelog item.

rickeylev:
Added to changelog.

PR #2168, opened 2024-08-30
refactor: move bootstrap tests to their own directory

rickeylev:
Move the bootstrap tests to their own directory.

This is a bit for code cleanliness (it was a bit awkward having a bunch of non-analysis
tests in the middle of the analysis tests), but also to make importing the base_tests
directory into Google easier.


rickeylev:
Well, that's a bit disappointing. I tried getting the commit history such that git recorded the BUILD.bazel change as a file move. And if you look at the individual commits, it is. But then the diff UI shows it as adding a new file without any relationship. Ah well. I wonder how it'll show up in the final commit history after squashing on submit.

aignas (approved):

PR #2171, opened 2024-08-31
fix(py_wheel): Avoid reliance on bash in `py_wheel` macro.

UebelAndre:
While trying to modernize an older repo I tried building wheels on windows and ran into a failure:
```
Windows Subsystem for Linux has no installed distributions.

Use 'wsl.exe --list --online' to list available distributions
and 'wsl.exe --install <Distro>' to install.

Distributions can also be installed by visiting the Microsoft Store:
https://aka.ms/wlstore
Errorcode: Bash/Service/CreateInstance/GetDefaultDistro/WSL_E_DEFAULT_DISTRO_NOT_FOUND
```

This appears to be caused by the `py_wheel_dist` rule which gets caught by `//...`. This target should be considered a side-effect/optional target of `py_wheel`. To fix:
1. Mark the target as `manual`, so it's only built when explicitly requested.
2. Implement copying to the directory with a Python program instead of shell, so bash
   isn't required.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,7 +46,8 @@ A brief description of the categories of changes:
 * (gazelle) Correctly resolve deps that have top-level module overlap with a gazelle_python.yaml dep module
 
 ### Added
-* Nothing yet
+* (py_wheel) Removed use of bash to avoid failures on Windows machines which do not
+  have it installed.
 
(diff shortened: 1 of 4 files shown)


rickeylev (approved):
Looks good! Thanks!

PR #2180, opened 2024-09-04
refactor(precompiler): give optimize/invalidation_mode flags default values

oprypin:
This allows one to actually run this script with *just* these flags: src, src_name, pyc

Extra: refactor how the enum member is parsed

### DIFF
--- a/tools/precompiler/precompiler.py
+++ b/tools/precompiler/precompiler.py
@@ -24,8 +24,8 @@ import sys
 
 def _create_parser() -> "argparse.Namespace":
     parser = argparse.ArgumentParser(fromfile_prefix_chars="@")
-    parser.add_argument("--invalidation_mode")
-    parser.add_argument("--optimize", type=int)
+    parser.add_argument("--invalidation_mode", default="TIMESTAMP")
+    parser.add_argument("--optimize", type=int, default=-1)
     parser.add_argument("--python_version")
 
     parser.add_argument("--src", action="append", dest="srcs")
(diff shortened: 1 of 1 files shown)


tools/precompiler/precompiler.py
```diff
@@ -24,8 +24,8 @@
 
 def _create_parser() -> "argparse.Namespace":
     parser = argparse.ArgumentParser(fromfile_prefix_chars="@")
-    parser.add_argument("--invalidation_mode")
-    parser.add_argument("--optimize", type=int)
+    parser.add_argument("--invalidation_mode", default="TIMESTAMP")
```
rickeylev, line 27: We can't default to timestamp because it would cause non-deterministic behavior / cache thrashing. The timestamp of the source file will reflect when it was e.g. installed or cloned, not necessarily when it was last modified.

I think it's fine to default it to checked_hash, though. That's captures the intent of timestamp.

tools/precompiler/precompiler.py
```diff
@@ -40,15 +40,15 @@ def _create_parser() -> "argparse.Namespace":
...
+        invalidation_mode = py_compile.PycInvalidationMode[
+            options.invalidation_mode.upper()
+        ]
+    except KeyError as e:
         raise ValueError(
             f"Unknown PycInvalidationMode: {options.invalidation_mode}"
         ) from e
 
-    if not (len(options.srcs) == len(options.src_names) == len(options.pycs)):
+    if len(options.srcs) != len(options.src_names) != len(options.pycs):
```
rickeylev, line 51: Looks like the fix got undone here?

rickeylev (changes requested):

tools/precompiler/precompiler.py
oprypin: Ah yes, this was the "default default" of Python itself, but for our purposes CHECKED_HASH is much better.

```suggestion
    parser.add_argument("--invalidation_mode", default="CHECKED_HASH")
```

tools/precompiler/precompiler.py
oprypin: Oops, will fix.

```suggestion
    if not (len(options.srcs) == len(options.src_names) == len(options.pycs)):
```

rickeylev (approved):

PR #2189, opened 2024-09-05
chore: cleanup exposed python_repository symbols and add docs

aignas:
The bazel inventory for this sphinxdocs is in #2188.

Removed the things that are definitely not used and the rest of the symbols
can be left until we become bzlmod only - they do not require much maintenance.

### DIFF
--- a/python/repositories.bzl
+++ b/python/repositories.bzl
@@ -29,7 +29,8 @@ py_repositories = _py_repositories
 python_register_multi_toolchains = _python_register_multi_toolchains
 python_register_toolchains = _python_register_toolchains
 
-# Useful for documentation.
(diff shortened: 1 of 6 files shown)


python/repositories.bzl
```diff
@@ -30,9 +29,10 @@ py_repositories = _py_repositories
 python_register_multi_toolchains = _python_register_multi_toolchains
 python_register_toolchains = _python_register_toolchains
 
+# Useful for documentation.
```
rickeylev, line 32: i.e. not intended to be public? Lets put a more explicit comment saying its only here for documentation, not public consumption, then.

rickeylev (approved):
just the doc comment, otherwise lgtm

PR #2200, opened 2024-09-07
fix: Fix incorrectly generated `Required-Dist` when specifying requirements with markers in extra_requires in py_wheel rule

sfc-gh-wzhao:
Currently, if extra requirements are provided via a file, it is allowed to contain markers, but if it is passed via extra_requires field, it will run into error since the extra in this case is blindly added in the following line in `py_wheel.bzl`
```
metadata_contents.append(
    "Requires-Dist: %s; extra == '%s'" % (requirement, option),
)
```

Thus, this PR adds a post-process for this case, to make it possible to pass something like
```
    extra_requires = {"example": [
        "pyyaml>=6.0.0,!=6.0.1",
        'toml; (python_version == "3.11" or python_version == "3.12") and python_version != "3.8"',
        'wheel; python_version == "3.11" or python_version == "3.12" ',
    ]},
```
to `py_wheel`


aignas (approved):
Thanks!

Issue #2081, opened 2024-07-20
Allow root module to customize toolchains


rickeylev:
We've had several requests where users want to modify something about the toolchain definitions that are generated for the hermetic runtimes. Typically these are small tweaks. Allowing other modules to tweak toolchains is a no-go, but allowing the root module to change things is reasonable.

The "simple" option for customizing toolchains is to add args to `python_register_toolchains` and `python.toolchain`.

The two main issues I see with adding args is:

1. bzlmod APIs tend to be "written in stone". They happen very early in the build process, so we have limited options for evolving the API. In this case, this is somewhat tempered by only the root module being allowed to use it, but it still makes me uneasy.
2. Toolchains tend to have unstable APIs. Most parts of them are just implementation details for the rules. For example, adding a _single_ arg to customize the bootstrap template made sense last year, but now there are multiple template files. Similarly, the stub_shebang attribute is slated for removal with the introduce of `--bootstrap_impl=script`.

Adding args to `python.toolchain` is, I think, a bad idea. Having something else, e.g. `python.root_config` might make a bit more sense. That said, args on `python.toolchain()` make it easy to associate something to a particular toolchain, while `python.root_config` would need some arg to specify which toolchain it wanted to restrict the setting to.

This issue is to collect the use cases and try and figure out some options.

Things people want to change:

* Overriding bootstrap template
  * Example: https://github.com/bazelbuild/rules_python/pull/2032
    * Goal: They want to implement a separate bootstrap
    * The PR proposes adding args to e.g. `python_register_toolchains()`.
    * Issues: (1) Would need to be exposed via bzmod apis. (2) There are 4 bootstrap files: python_bootstrap_template (legacy system python bootstrap), stage1 script bootstrap, stage2 bootstrap, and zip main bootstrap. Plus the launcher exe thing for windows.
    * Proposal: Indirect the references to the templates using flags.

* Customizing stub_shebang: https://bazelbuild.slack.com/archives/CA306CEV6/p1721464127008299
  * Goal: They want to set `-S PYTHONNOUSERSITE=1` for all their Python invocations
  * Proposal 1: Have a flag override for stub_shebang
  * Proposal 2: Add `interpreter_args` and make `env` populate values into the bootstrap.
  * Proposal 3: Have flags for "default interpreter args" and "default interpreter env"

* The register_coverage_tool arg
  * Goal: Having coverage included as part of the toolchain is unnecessary if a user test provides the coverage library itself

[...]

PR #2201, opened 2024-09-07
feat(toolchain): add patch_strip attr for python_repository

aignas:
This value has been hardcoded for a long time, so let's add the `patch_strip`
attribute to the `WORKSPACE` setups now so that we add less tech debt in the
`bzlmod` world later and need to migrate `bzlmod` users from the hard coded
value to something that can be configured.

This should be backwards compatible because of the default `int` value of `1`
for the `patch_strip` attribute.

Summary:
- feat: add `patch_strip` to `python_repository`.
- feat: add the default value of patch_strip to get_release_info.
- refactor: handle patch_strip key in `TOOL_VERSIONS`.

Work towards #2081.

### DIFF
--- a/python/private/python_repositories.bzl
+++ b/python/private/python_repositories.bzl
@@ -521,7 +521,7 @@ For more information see the official bazel docs
 Same as the --strip argument of Unix patch.
 
 :::{note}
-In the future the default value will be set to `0`, to mimic the wellk known
+In the future the default value will be set to `0`, to mimic the well known
 function defaults (e.g. `single_version_override` for `MODULE.bazel` files.
 :::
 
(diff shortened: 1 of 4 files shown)


python/private/python_repositories.bzl
```diff
@@ -515,6 +516,21 @@ For more information see the official bazel docs
         "netrc": attr.string(
             doc = ".netrc file to use for authentication; mirrors the eponymous attribute from http_archive",
         ),
+        "patch_strip": attr.int(
+            doc = """
+Same as the --strip argument of Unix patch.
+
+:::{note}
+In the future the default value will be set to `0`, to mimic the wellk known
```
rickeylev, line 524:
```suggestion
In the future the default value will be set to `0`, to mimic the well known
```

rickeylev (approved):

PR #2204, opened 2024-09-08
test(bzlmod): add python.toolchain unit tests

aignas:
With this PR we get rudimentary unit tests for the `python` bzlmod extension
which allows us to unit test the override behaviour that will become more
complex soon.

Summary:
- refactor: inline the python_register_toolchains
- refactor: use toolchain_info to call python_register_toolchains
- refactor: move the registration out of the main loop
- refactor: split the parsing of the modules to a separate function
- test(bzlmod): add python.toolchain module parsing tests

Work towards #2081


tests/python/python_tests.bzl
```diff
@@ -0,0 +1,253 @@
...
+    )
+    rules_python_toolchain = struct(
+        name = "python_3_11",
+        python_version = "3.11",
+        register_coverage_tool = False,
+    )
+    env.expect.that_collection(py.toolchains).contains_exactly([
+        some_module_toolchain,
+        rules_python_toolchain,
+        my_module_toolchain,
```
rickeylev, line 203: shouldn't the order be `[my, some, rules_python]` ?

aignas:
```suggestion
    env.expect.that_collection(py.toolchains).contains_exactly([
        some_module_toolchain,
        rules_python_toolchain,
        my_module_toolchain, # this was the only toolchain, default to that
```

rickeylev (approved):

PR #2214, opened 2024-09-11
refactor(whl_library): split out a whl_archive library

aignas:
Before this PR users would need to download unused dependencies to just extract
the wheels. The minimum list of deps are - `packaging` and `installer` for our
`whl_installer` tool to properly work. This also means that we could not
dogfood our `whl_installer` tool in the `pip-tools` implementation.

With this PR we start using `whl_archive` which just patches, extracts the
wheel and parses `METADATA` where possible and this is the first step towards
separating how `sdists` are treated.

Summary:
- refactor: split out whl_archive_impl function for just whls
- refactor: make the implementation of the 'whl_archive' more minimal
- use `whl_archive` in internal deps
- use a `py_library` in `pip_compile` rule.
- remove unused attrs from 'whl_archive'

### DIFF
--- a/python/private/pypi/deps.bzl
+++ b/python/private/pypi/deps.bzl
@@ -104,7 +104,7 @@ package(default_visibility = ["//visibility:public"])
 load("@rules_python//python:defs.bzl", "py_library")
 
 py_library(
-    name = "lib",
+    name = "pkg",
     srcs = glob(["**/*.py"]),
     data = glob(["**/*"], exclude=[
         # These entries include those put into user-installed dependencies by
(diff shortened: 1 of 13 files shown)


python/private/pypi/deps.bzl
```diff
@@ -130,11 +131,32 @@ def pypi_deps():
...
+                build_file_content = _GENERIC_WHEEL,
+            )
+        else:
+            maybe(
+                whl_archive,
+                name,
+                urls = [url],
+                sha256 = sha256,
+                requirement = name.split("_")[-1],
+                repo = "pypi_",
```
aignas, line 150: I thought it would be nice to dogfood the library, but we need to have a reference to the default (or some other) python interpreter in order to do this.

In `bzlmod` it is possible to do this, but in `WORKSPACE` it is not, so we may need to leave the `deps.bzl` as is.

PR #2217, opened 2024-09-12
chore: remove mandatory default values for python version flag

aignas:
This is a different take on #2205.

Summary:
- Remove version constraints from the `//python/config_settings:python_version`.
- Remove `is_python_config_setting` macro and use a different implementation to
  retain the same functionality.

This in general will improve the situation on how the toolchains are registered
and hopefully is a step towards resolving issues for #2081.


python/private/python_version_flag.bzl
```diff
@@ -0,0 +1,48 @@
...
+
+def _python_version_flag_impl(ctx):
+    value = ctx.build_setting_value
+    if ctx.attr.values and value not in ctx.attr.values:
+        fail((
+            "Invalid --python_version value: {actual}\nAllowed values {allowed}"
+        ).format(
+            actual = value,
+            allowed = ", ".join(sorted(ctx.attr.values)),
+        ))
```
rickeylev, line 28: Lets just drop the values restriction on the flag entirely. I'm just not convinced all the extra wiring is worth it, and it just artificially constrains the values, requiring extra work from users to get a value into the list.

Since the flag is no longer coupled to the known values, move the flag definition to be directly in the BUILD file, like the other flags.

Worst case, you set a value that doesn't map to a toolchain, and you end up getting the default. Surprising, but not that bad. We can also build in guard elsewhere that use information from the analysis phase instead of loading phase. The two ideas that come to mind are:
* In py_binary or equiv, make the //python:toolchain_type (fake) optional. If toolchain lookup fails, then we compare the flag to the list of known values. If the flag isn't in the list, then we print an error about the missing toolchain and mention the flag not being in the list of known values as a hint.
* In the python_version flag definition, make `values` a label and point it to e.g. `@pythons_hub//:known_versions` (if bzlmod is true). Then fail at analysis time in the flag logic.
* Make everything depend on an internal "check the flag value" target that does the above.

The second option in particular would make the bzlmod/workspace situation much easier -- we can just put an `X if bzlmod else None` clause in the attr.label value itself.

python/private/python_repositories.bzl
```diff
@@ -54,6 +55,15 @@ def py_repositories():
         internal_config_repo,
         name = "rules_python_internal",
     )
+    maybe(
+        hub_repo,
+        name = "pythons_hub",
```
rickeylev, line 60: Hrm. How do we get the list of values into pythons_hub for workspace builds?
Or is the idea: we just want the repo to _exist_ so that load() can work?

python/private/config_settings.bzl
```diff
@@ -161,74 +118,65 @@ def is_python_config_setting(name, *, python_version, reuse_conditions = None, *
...
+        name: A dummy value
+        values: {type}`list[str]` The list of version values.
+    """
+    for name in [
+        "python_version",
+    ] + [
+        "is_python_version_unset" if not version else ("is_python_" + version)
+        for version in [""] + values
+    ]:
+        native.alias(
```
rickeylev, line 134: These aliases seem gratuitous. How about just loading a list from pythons_hub instead?

However, see main comment for another idea that might just make all the config_setting generation much simpler.

rickeylev (commented):
I had an inkling of an idea that might make the config_setting generation we do much easier: use FeatureFlagInfo.

Basically, have a separate flag read the full-version flag. It returns the major.minor part. Because it returns the value using FeatureFlagInfo, it can then affect config_setting resolution. We can then do away with all the config setting groups.

So something like:

```
def _python_major_minor_impl(ctx):
  full_version = ctx.attr._python_version_flag
  major_minor = <parse full into X.Y>
  return [FeatureFlagInfo(value=major_minor)]

python_major_minor = rule(
  impl = _python_major_minor,
  build_setting_value = config.string(flag=False)
  attrs = {"_python_version_flag": attr.label("//python/config_settings:python_version")}
)

python_major_minor(name = "_python_major_minor")

[
config_setting(
  name = "is_python_3.{}",
  flag_values = {":_python_major_minor": "3.{}"}
)
for v in (5, 6, 7, 8, 9, 10, ...)
]
```

aignas:
I have pivoted the change and just removed all of the `is_python_config_setting` with the suggested `rule`. This has been a really nice thing to do and it solved a lot of complexity I needed to go over when I was implementing the `pypi` integration.

python/private/config_settings.bzl
```diff
@@ -178,40 +39,70 @@ def construct_config_settings(name = None):  # buildifier: disable=function-docs
         # also prevents version-unaware targets from inadvertently matching
         # a select condition when they shouldn't.
         build_setting_default = "",
-        values = [""] + VERSION_FLAG_VALUES.keys(),
+        visibility = ["//visibility:public"],
+    )
+
+    _python_major_minor(
+        name = "python_version_major_minor",
```
aignas, line 46: name is up for bike shedding.

rickeylev: I think the name is fine -- just make it more internal sounding -- prefix it with "_". It's not a flag, and shouldn't be directly set, so lets just hide it as best we can so it isn't an attractive nuisance. 

(or move it under private somewhere, but we have other internal names here, so i think an underscore is sufficient).

python/private/config_settings.bzl
```diff
@@ -18,149 +18,10 @@
...
-        name = "_{}_group".format(name),
-        match_any = match_any,
-        visibility = ["//visibility:private"],
-    )
-    native.alias(
-        name = name,
-        actual = "_{}_group".format(name),
-        visibility = kwargs.get("visibility", []),
-    )
+_PYTHON_MINOR_VERSION_FLAG = str(Label("//python/config_settings:python_version_major_minor"))
```
rickeylev, line 24: Suggest to name the symbol after the underlying name

```suggestion
_PYTHON_MAJOR_MINOR_VERSION_FLAG = str(Label("//python/config_settings:python_version_major_minor"))
```

rickeylev (approved):
Other than the underscore prefix and CI errors, LGTM

Issue #2216, opened 2024-09-11
Add in-build toolchain registration to python.toolchain


gholms:
# 🚀 feature request

### Relevant Rules

* `//python/extensions:python.bzl`:  `python.toolchain`
* `//python:repositories.bzl`:  `python_register_toolchains`

### Description

While the underlying rules for python toolchains/runtimes support python stacks which are actual Bazel-built targets passed in by label, `python_register_toolchains`, and by extension, the `python.toolchain` tag class, does not offer a way to do this.  At present it supports only downloading hermetic toolchains from GitHub by version number.  While one can technically do all the behind-the-scenes work that it performs using one's own module/workspace code, what exactly they do and what expectations they have for the repositories one points them at are not documented.  Especially as we make more use of additional types of toolchains (cc toolchains, exec tools toolchains), discovering and writing this much code is an increasing burden.  We should provide better support for people who supply their own python stacks, either built from source or precompiled.

### Describe the solution you'd like

Digging through `python_register_toolchains` and a bunch of other code, it looks like each platform expects three toolchains, a coverage tool, a config setting, and the seven other targets which `toolchain_aliases` repositories name.  For each case like this, we can either define a contract for what targets the user-supplied repository+package needs to provide (perhaps using some helpful macros or tag classes from us), then ask them to tell us where to fetch it from and how to patch it; or prompt for each of these things individually.  Either way, we could do this either by extending the existing `python_register_toolchains` and `py.toolchain`, or by defining a new pair of these things dedicated to this use case.

For python stacks built from source, we can allow people to supply just one repository+package to use everywhere and let Bazel do all the platform management for us.  I don't know how best to build a good UX around this for python stacks supplied as binaries without forcing a lot of repetition onto people who want to support lots of platforms.  But to an extent that work is also self-imposed, and people could use macros or a platform-based, `toolchain_aliases`-like repository to manage that complexity.

### Describe alternatives you've considered

Instead of supporting the in-build use case ourselves, we could document what people need to do in order to do it themselves and still be able to integrate with the python ecosystem correctly.  That makes it far less likely that people will overlook things like they would with source code dives, but it does effectively force people to reimplement much of our finicky loading phase code.  Most of the smaller bits comprising it aren't exposed publicly, but perhaps opening some of it up would help support that scenario a bit.

aignas:
[...]

PR #2221, opened 2024-09-15
refactor: move hermetic Python runtime setup into macro

rickeylev:
This stems from conversation in #2216. Moving the logic into a loading-phase macro makes it
easier to understand how a repo with a python-build-standalone extracted archive has
its toolchain defined. This also makes it easier to compare with the local runtime repo
setup to try and find commonality in the future.

Along the way:
* Remove some Bazel 5 compatibility code: coverage_tool is now always passed to py_runtime.
  (Bazel 5 support was dropped earlier this year and we stopped testing with it; it's
  likely already broken with Bazel 5).
* Add `semver.to_dict`, which makes it easier to pass into format() functions.


python/private/hermetic_runtime_repo_setup.bzl
```diff
@@ -0,0 +1,161 @@
...
+            format.
+        python_bin: {type}`str` The path to the Python binary within the
+            repositoroy.
+        coverage_tool: {type}`str` optional target to the coverage tool to
+            use.
+    """
+    _ = name  # @unused
+    python_version_info = python_version.split(".")
+    major = python_version_info[0]
+    minor = python_version_info[1]
```
aignas, line 53: Nit, using the `semver` function from #2218 would be nice here.

python/private/hermetic_runtime_repo_setup.bzl
```diff
@@ -0,0 +1,161 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Setup a python-build-standalone based toolchain."""
+
+load("@rules_cc//cc:defs.bzl", "cc_import", "cc_library")
+load("@rules_python//python:py_runtime.bzl", "py_runtime")
+load("@rules_python//python:py_runtime_pair.bzl", "py_runtime_pair")
+load("@rules_python//python/cc:py_cc_toolchain.bzl", "py_cc_toolchain")
+load("@rules_python//python/private:py_exec_tools_toolchain.bzl", "py_exec_tools_toolchain")
```
aignas, line 20: I think the loads here could be simplified:
```suggestion
load("//python:py_runtime.bzl", "py_runtime")
load("//python:py_runtime_pair.bzl", "py_runtime_pair")
load("//python/cc:py_cc_toolchain.bzl", "py_cc_toolchain")
load("//python/private:py_exec_tools_toolchain.bzl", "py_exec_tools_toolchain")
```

python/private/hermetic_runtime_repo_setup.bzl
```diff
@@ -0,0 +1,161 @@
...
+    # Used to only download coverage toolchain when the coverage is collected by
+    # bazel.
+    native.config_setting(
+        name = "coverage_enabled",
+        values = {"collect_code_coverage": "true"},
+        visibility = ["//visibility:private"],
+    )
+    py_runtime_kwargs = {}
+    if coverage_tool:
+        # Earlier versions of Bazel don't support this attribute.
```
aignas, line 127: I think it is supported from `6.0` onwards that is the lowest supported version, so we can drop the `if` here.

python/private/hermetic_runtime_repo_setup.bzl
```diff
@@ -0,0 +1,161 @@
...
+        files = [":files"],
+        interpreter = python_bin,
+        interpreter_version_info = {
+            "major": str(major),
+            "micro": str(python_version_info[2]),
+            "minor": str(minor),
+        },
+        python_version = "PY3",
+        implementation_name = "cpython",
+        pyc_tag = "cpython-{}{}".format(major, minor),
```
aignas, line 141: I am so used to this being something like `cp{major}{minor}` in the wheels. It's interesting that this is different, should we have a small comment here linking to some docs?

python/private/hermetic_runtime_repo_setup.bzl
```diff
@@ -0,0 +1,161 @@
...
+    py_cc_toolchain(
+        name = "py_cc_toolchain",
+        headers = ":python_headers",
+        libs = ":libpython",
+        python_version = python_version,
+    )
+
+    py_exec_tools_toolchain(
+        name = "py_exec_tools_toolchain",
+        precompiler = "@rules_python//tools/precompiler:precompiler",
```
aignas, line 160: nit: not using `@rules_python`
```suggestion
        precompiler = Label("//tools/precompiler:precompiler"),
```

aignas (approved):
Only minor things, so LGTM!

python/private/hermetic_runtime_repo_setup.bzl
rickeylev: Done

python/private/hermetic_runtime_repo_setup.bzl
rickeylev: Done

python/private/hermetic_runtime_repo_setup.bzl
rickeylev: Linked to the pep

python/private/hermetic_runtime_repo_setup.bzl
rickeylev: Done

PR #2222, opened 2024-09-15
feat(bzlmod): add python.override APIs

aignas:
Before this PR the users could not override how the hermetic toolchain
is downloaded when in `bzlmod`. However, the same APIs would be
available to users using `WORKSPACE`. With this we also allow root
modules to restrict which toolchain versions the non-root modules, which
may be helpful when optimizing the CI runtimes so that we don't waste
time downloading multiple `micro` versions of the same `3.X` python
version, which most of the times have identical behavior.

Whilst at it, tweak the `semver` implementation to allow for testing of
absence of values in the original input.

Work towards #2081 and this should be one of the last items that are
blocking #1361 from the API point of view.

Replaces #2151.


docs/toolchains.md
```diff
@@ -169,6 +169,20 @@ Variables](https://bazel.build/reference/be/make-variables). See the
...
+
+One can perform various overrides for the registered toolchains from the root
+module. For example, the following use cases would be supported using the
+existing attributes:
+
+* Limiting the available toolchains for the entire `bzlmod` transitive graph
+  via {attr}`python.override.available_python_versions`.
+* Setting particular `X.Y.Z` python versions when modules request `X.Y` version
+  via {attr}`python.override.minor_mapping`.
+* Adding custom {attr}`python.single_version_platform_override.coverage_tool`.
```
rickeylev, line 182:
```suggestion
* Per-version control of the coverage tool used using
  {attr}`python.single_version_platform_override.coverage_tool`.
```

docs/toolchains.md
```diff
@@ -169,6 +169,20 @@ Variables](https://bazel.build/reference/be/make-variables). See the
...
+* Adding new python versions via {bzl:obj}`python.single_version_override` or
```
rickeylev, line 183:
```suggestion
* Adding additional Python versions via {bzl:obj}`python.single_version_override` or
```

docs/toolchains.md
```diff
@@ -169,6 +169,20 @@ Variables](https://bazel.build/reference/be/make-variables). See the
...
 
+### Overriding toolchain defaults and adding more versions
+
+One can perform various overrides for the registered toolchains from the root
+module. For example, the following use cases would be supported using the
+existing attributes:
+
+* Limiting the available toolchains for the entire `bzlmod` transitive graph
+  via {attr}`python.override.available_python_versions`.
+* Setting particular `X.Y.Z` python versions when modules request `X.Y` version
```
rickeylev, line 180:
```suggestion
* Setting particular `X.Y.Z` Python versions when modules request `X.Y` version
```

examples/bzlmod/MODULE.bazel
```diff
@@ -37,6 +37,54 @@ python.toolchain(
     python_version = "3.10",
 )
 
+# One can override the actual toolchain versions that are available, which can be useful
+# to when optimizing what gets downloaded and when.
```
rickeylev, line 41:
```suggestion
# One can override the actual toolchain versions that are available, which can be useful
# when optimizing what gets downloaded and when.
```

examples/bzlmod/MODULE.bazel
```diff
@@ -37,6 +37,54 @@ python.toolchain(
...
+python.override(
+    available_python_versions = [
+        "3.10.9",
+        "3.9.19",
+        # The following is used by the `other_module` and we need to include it here
+        # as well.
+        "3.11.8",
+    ],
+    # Also override the `minor_mapping` so that when the modules specify a particular
+    # `3.X` version, we decide what gets used.
```
rickeylev, line 51:
```suggestion
    # Also override the `minor_mapping` so that the root module,
    # instead of rules_python's defaults, controls what full version
    # is used when `3.x` is requested.
```

python/extensions/python.bzl
```diff
@@ -12,7 +12,34 @@
...
+```starlark
+python = use_extension("@rules_python//python/extensions:python.bzl", "python")
+python.toolchain(
+    is_default = True,
+    python_version = "3.11",
+)
+use_repo(python, "python_3_11")
+```
+
+For more in-depth documentation see the {rule}`python.toolchain`.
```
rickeylev, line 30: python.toolchain is a tagclass, so `{rule}` won't xref to it. obj should work, though, or `{tag-class}` I think.

```suggestion
For more in-depth documentation see the {obj}`python.toolchain`.
```

python/private/python_repository.bzl
```diff
@@ -302,27 +301,11 @@ python_repository = repository_rule(
...
-The path to the single file, or the executable if the target is executable,
-determines the entry point for the python coverage tool.  The target and its
-runfiles will be added to the runfiles when coverage is enabled.
-
-The entry point for the tool must be loadable by a Python interpreter (e.g. a
-`.py` or `.pyc` file).  It must accept the command line arguments
-of coverage.py (https://coverage.readthedocs.io), at least including
-the `run` and `lcov` subcommands.
-
-The target is accepted as a string by the python_repository and evaluated within
```
rickeylev, line 321: I think we need to keep this part? It's basically trying to say: if you're using bzlmod, you need to worry about bzlmod use_repo visibility with the value you put here.

rickeylev (commented):
Reviewed everything but the python.bzl part and its tests. Just some doc fixes thus far.

python/private/python.bzl
```diff
@@ -284,25 +305,234 @@ def _fail_multiple_default_toolchains(first, second):
...
+
+    available_versions = default["tool_versions"]
+    kwargs = default.setdefault("kwargs", {})
+
+    if tag.sha256 or tag.urls:
+        if not (tag.sha256 and tag.urls):
+            _fail("Both `sha256` and `urls` overrides need to be provided together")
+            return
+
+        for platform in tag.sha256 or []:
```
rickeylev, line 328:
```suggestion
        for platform in (tag.sha256 or []):
```

python/private/python.bzl
```diff
@@ -329,79 +559,299 @@ def _get_bazel_version_specific_kwargs():
...
     },
+)
+
+_override = tag_class(
+    doc = """Tag class used to override defaults and behaviour of the module extension.
+
+:::{versionadded} 0.36.0
+:::
+""",
+    attrs = dict(
```
rickeylev, line 661: nit: use `{}` literal with `|`. It's just a bit less indentation and flows a bit nicer.
```
attrs = {
  "key": ...
} | OTHER_DICT
```

python/extensions/python.bzl
```diff
@@ -12,7 +12,34 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"Python toolchain module extensions for use with bzlmod"
+"""Python toolchain module extensions for use with bzlmod.
+
+## Basic usage
```
rickeylev, line 17: I think the `##` headers are messing up the rendered docs -- in the rtd preview doesn't show "python" as a heading anymore.

rickeylev (approved):

python/private/python_repository.bzl
aignas: We are right now duplicating the docs in both places, so I think maybe we can just point to the docs in the runtime? Which one is the source of truth?

PR #2230, opened 2024-09-17
fix: Prefix bootstrap file to prevent pytest reentrance

mortenmj:
Pytest will, by default, discover tests in files named `test_*.py`. The existing behaviour of the bootstrap setup is to create a file named e.g. `test_foo_stage2_bootstrap.py` for a target named `test_foo`. This causes pytest to reenter, leading to a crash.

### DIFF
--- a/python/private/common/py_executable_bazel.bzl
+++ b/python/private/common/py_executable_bazel.bzl
@@ -330,7 +330,7 @@ def _create_stage2_bootstrap(
         imports,
         runtime_details):
     output = ctx.actions.declare_file(
-        "{}_stage2_bootstrap.py".format(output_prefix),
+        "_{}_stage2_bootstrap.py".format(output_prefix),
         sibling = output_sibling,
     )
     runtime = runtime_details.effective_runtime


python/private/common/py_executable_bazel.bzl
```diff
@@ -330,7 +330,7 @@ def _create_stage2_bootstrap(
         imports,
         runtime_details):
     output = ctx.actions.declare_file(
-        "{}_stage2_bootstrap.py".format(output_prefix),
+        "_{}_stage2_bootstrap.py".format(output_prefix),
```
rickeylev, line 333:
```suggestion
        # Prepend with underscore to prevent pytest from trying to
        # process the bootstrap for files starting with `test_`
        "_{}_stage2_bootstrap.py".format(output_prefix),
```

rickeylev (approved):
Mostly LGTM, just some small doc changes needed. I suggested an edit for one. Can you please add some text to CHANGELOG.md, too? e.g. under the fixed header, add something like:

```
* (rules) Prevent pytest from trying run the generated stage2
  bootstrap .py file when using {obj}`--bootstrap_impl=script`
```

PR #2236, opened 2024-09-20
fix(platforms): include flag_values in config_settings

aignas:
This function that allows us to generate config settings for
supported platforms did not get the flag_values addition.


rickeylev (approved):
Remind me -- what is gen_python_config_settings() used for? The only usage I see is in the toolchain_aliases.

aignas:
I don't think it is used (apart from example/wheels tests), but the symbol is public. I would love to move the symbol to a non-public location, but I am not sure if that ship has not sailed yet.

PR #2237, opened 2024-09-20
fix(ci): use --enable_workspace for bazel-in-bazel tests

aignas:
The CI on `main` started failing and is affecting all PRs.

Most likely a fixup to #2184.


groodt (approved):
LGTM

PR #2238, opened 2024-09-20
fix(whl_filegroup): Make RECORD from wheel available

TimotheusBachinger:
Why this change is being made, briefly.
 * We need to also get the `RECORD` file from a wheel in order to package everything
 * It seems a bit weird to me that the `RECORD` file is *not part* of `whl_filegroup` even no pattern filtering was applied (or I am misunderstanding something)
 * Further: neither `hash` nor `filelen` have been used in the code, so why are only files considered with those attributes?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,7 @@ A brief description of the categories of changes:
 * Nothing yet
 
 ### Fixed
-* Nothing yet
+* (whl_filegroup): Provide per default also the `RECORD` file
 
 ### Added
 * Nothing yet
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -66,7 +66,7 @@ A brief description of the categories of changes:
   bootstrap .py file when using {obj}`--bootstrap_impl=script`
 * (toolchain) The {bzl:obj}`gen_python_config_settings` has been fixed to include
   the flag_values from the platform definitions.
-
+* (whl_filegroup): Provide per default also the `RECORD` file
```
aignas, line 69: Could you please rebase and update the `CHANGELOG.md` to point to things added in an unreleased version?

aignas (commented):
LGTM otherwise, thanks for the contribution!

CHANGELOG.md
TimotheusBachinger: hope it is now where you want it :-)

aignas (approved):

Issue #2212, opened 2024-09-10
per-binary opt-in of pyc doesn't function correctly


rickeylev:
# 🐞 bug report

### Affected Rule

py_binary/py_test

### Description

The intent of the `py_binary.pyc_collection` attribute is to allow per-target opt-in for using pyc files from dependencies _without specifying command line build flags_. This allows for a couple use cases:
* Build tools can have precompiling enabled (separate from the requested target)
* Gradual enabling of precompiling can be done (on a per-target basis) without having to figure out where and how a user is setting build flags (which isn't always possible, e.g. during the dev edit-run cycle).

Right now, what happens is, when `py_binary.pyc_collection=include_pyc`, pyc files from transitive targets are collected, but only if they opted into precompiling with their target-level precompile attribute. When the vast majority of targets _don't_ set that, the net effect is no pyc files are collected. This happens because pyc file generation is conditional on the precompile settings, e.g. if its not enabled, no action generates the file.

To fix, the basic thing that needs to be done is _always_ generate a pyc file and always include it PyInfo.pyc_files. Then, when the binary looks at the transitive pyc files, it'll find them all.

A caveat is: don't add the pyc files to runfiles unless the target opted into precompiling. Doing so would cause all downstream targets to have the pyc, as opposed to only the binaries that want it. It's OK to put the pyc in default outputs (the default outputs of deps aren't included).

rickeylev:
Argh. Ok, this turned into a bit of tangled mess. I think I'm going to need to back out a few changes and, within, Google, we'll just have to carry a couple patches.

The problem is that we want the binary-level `pyc_collection` attribute to have control over whether pyc files end up in the binary. However, the --precompile flag ends up having precedence because it causes libraries to put pyc files into runfiles. This means a binary can't opt-out if --precompile=enabled is the default. So now we have the awkward situation of where a binary can't opt-out with --precompile=enabled is the default *and* a binary can't opt-in because it requires setting --precompile=enabled (thus preventing opt out).

The _general_ behavior we want is:

* `--precompile=enabled, attr.pyc_collection=disabled` => Don't include pyc files
* `--precompile=enabled, attr.pyc_collection=include_pyc` => Include pyc files
* `--precompile=disabled, attr.pyc_collection=disabled` => Don't include pyc files
* `--precompile=disabled, attr.pyc_collection=include_pyc` => Include pyc files

[...]

Issue #2178, opened 2024-09-04
pipy dependency generated py_library ignores .pyc files in installed package.


qingyouzhao:
# 🐞 bug report

### Affected Rule

https://github.com/bazelbuild/rules_python/blob/main/docs/pypi-dependencies.md
Specifically `py_library` with `name = "pkg"` generated by 
```
pip = use_extension("@rules_python//python/extensions:pip.bzl", "pip")
pip.parse(
    hub_name = "my_deps",
    python_version = "3.11",
    requirements_lock = "//:requirements_lock_3_11.txt",
)
use_repo(pip, "my_deps")
```

### Is this a regression?

I don't know.

### Description

I couldn't import a module from pip package that ships python modules with .pyi and .pyc files.
For example, I have a example_package that ships with the following file structure
```
/example_package
  __init__.py
  foo.pyc
  foo.pyi
```
According to https://peps.python.org/pep-3147/#flow-chart, 
```
import foo from example_package
```
Should be successful. But the `py_library` generated specifically excludes `"**/*.pyc"` from `data`. If I manually remove the `"**/*.pyc"` exclusion, the import statement works.
```
py_library(
    name = "pkg",
    srcs = glob(
        ["site-packages/**/*.py"],
        exclude=[],
        # Empty sources are allowed to support wheels that don't have any
        # pure-Python code, e.g. pymssql, which is written in Cython.
        allow_empty = True,
    ),
    data = [] + glob(
        ["site-packages/**/*"],
        exclude=["**/* *", "**/*.py", "**/*.pyc", "**/*.pyc.*", "**/*.dist-info/RECORD"],
    ),
    # This makes this directory a top-level in the python import
    # search path for anything that depends on this.
    imports = ["site-packages"],
    deps = [],
    tags = ["pypi_name=example_package", "pypi_version=3.21.0"],
    visibility = ["//visibility:public"],
)
```

How
## 🔬 Minimal Reproduction

1. Creating a pip package following https://packaging.python.org/en/latest/ but replace the foo.py with compiled foo.pyc file.
2. Add pip dependency to a bazel workspace following guides in https://github.com/bazelbuild/rules_python/blob/main/docs/pypi-dependencies.md
3. In a py_library or py_binary, `import foo from example_package` should work but errors out with module not found error.

## 🔥 Exception or Error

```
ImportError: cannot import name 'foo' from 'example_package'
```

## 🌍  Your Environment

**Operating System:**

```
Linux
```

[...]

PR #2243, opened 2024-09-22
fix(precompiling)!: make binary-level precompile opt-in/opt-opt work

rickeylev:
This makes binary-level opt-in/opt-out of precompiling work as intended. Previously,
when `pyc_collection=include_pyc` was set on a binary, only transitive libraries that
had explicitly enabled precompiling were being included (which was moot anyways --
libraries put their files in runfiles, so no matter what, their files were included).

The intent was that, when a binary set `pyc_collection=include_pyc`, then precompiled
files would be used for all its transitive dependencies (unless they had, at the
target-level, disabled precompiling). Conversely, if `pyc_collection=disabled` was set,
the precompiled files would not be used (unless a target had, at the target level,
enabled precompiling).

To make it work as desired, the basic fix is to make it so that libraries have a place to
put the implicit pyc files (the ones automatically generated), and have the binaries
include those when requested. The net effect is a library has 4 sets of files it produces:
* required py files: py source files that should always go into the binary's runfiles
* required pyc files: precompiled pyc files that should always go into the binary's
  runfiles (e.g., when a library sets `precompile=enabled` directly).
* implicit pyc files: precompiled pyc files for a library that are always generated, but
  it's up to the binary if they go into the runfiles
* implicit pyc source files: the source py file for an implicit pyc file. When a binary
  *doesn't* include the implicit pyc file, it must include the source py file (otherwise
  none of the library's code ends up included).

Similarly, in order to allow a binary to decide what files are used, libraries must
stop putting the py/pyc files into runfiles themselves. While this is potentially
a breaking change, I found that, within Google, there was no reliance on this behavior,
so should be safe enough. That said, I added `--add_srcs_to_runfiles` to restore
the previous behavior to aid in transitioning.

**BREAKING CHANGES**
1. `py_library` no longer puts its srcs into runfiles directly.
2. Removed `--precompile_add_to_runfiles`
3. Removed `--pyc_collection`
4. `precompile=if_generated_source` removed
5. `precompile_source_retention=omit_if_generated_source` removed

Though 2 through 5 are technically breaking changes, I don't think precompiling
was very usable anyways, so usages of those flags/values is rare.

Fixes https://github.com/bazelbuild/rules_python/issues/2212


rickeylev:
Ok, finally ready for review.

aignas:
Before I have a look at this, there is a bug report in #2178. If I try to include the `pyc` files from these whls that include them, will this feature work? I think the correct behaviour would be to have explicit `pyc` files in `py_library` that get included even if the `precompile` flag is disabled.

CHANGELOG.md
```diff
@@ -47,6 +55,9 @@ A brief description of the categories of changes:
 * (rules) `compile_pip_requirements` passes `env` to the `X.update` target (and
   not only to the `X_test` target, a bug introduced in
   [#1067](https://github.com/bazelbuild/rules_python/pull/1067)).
+* (precompiling) The {obj}`pyc_collection` attribute now correctly
+  enables (or disables) using pyc files from targets transitively
+>>>>>>> 607a9c3c (wip: fix pyc_collection bug)
```
aignas, line 60: merge/rebase leftovers

rickeylev:
If a py_library includes pyc files in srcs, then...I'm not sure. It won't be an action conflict, because the pyc is a source file, not a generated file. Both the generated pyc and source pyc will try to go to the same location in runfiles, but I don't recall offhand what'll happen. Runfiles is a bit more lenient about duplicate files. I think it'll pick the first file, where first is according to whatever the depset order happens to result in.

> I think the correct behaviour would be to have explicit pyc files in py_library that get included even if the precompile flag is disabled.

Assuming the pyc files are for the matching python version, then yes, I agree.

I'm not entirely sure if we can safely include pyc files, though. If the wheel's pycs are using timestamp checking (the default), then...well I don't see how a wheel could possibly include timestamp-based pyc files reliablity. From our CI, we saw frequent issues due to runtime-generated pycs files, and ultimately had to ignore them. However, I never fully understood why -- presumably everything is in runfiles, with per-file symlinks, _and_ sandboxed, so it's not clear how the runtime was "escaping" and creating files in the backing repo directory along side the original source files (maybe it was specific to the stdlib?). But all those issues went away when we made it ignore pyc source files. :shrug:

CHANGELOG.md
rickeylev: Fixed

python/private/flags.bzl
```diff
@@ -21,6 +21,40 @@ unnecessary files when all that are needed are flag definitions.
 load("@bazel_skylib//rules:common_settings.bzl", "BuildSettingInfo")
 load("//python/private:enum.bzl", "enum")
 
+def _FlagEnum_flag_values(self):
+    return sorted(self.__members__.values())
+
+def FlagEnum(**kwargs):
```
aignas, line 27: question: Should `FlagEnum` be also used in some of the `pypi` integration flags?

python/private/proto/py_proto_library.bzl
```diff
@@ -127,16 +134,19 @@ def _py_proto_aspect_impl(target, ctx):
...
     ]
 
 _py_proto_aspect = aspect(
     implementation = _py_proto_aspect_impl,
-    attrs = {} if _incompatible_toolchains_enabled() else {
-        "_aspect_proto_toolchain": attr.label(
-            default = ":python_toolchain",
-        ),
-    },
+    attrs = _py_common.API_ATTRS | (
```
aignas, line 143: Nice to also use the common API in the `py_proto_aspect`!

aignas (approved):
LGTM

python/private/flags.bzl
rickeylev: Sure. I just created this because writing that same `sorted(e.members.values())` was looking pretty redundant. This PR was big enough and I didn't want to go update all the stuff in flags.bzl to use it.

PR #2246, opened 2024-09-24
Feat configure coverage rc env variable

ewianda:
PR Instructions/requirements
  section at the bottom.
* Body text describes:


examples/bzlmod/BUILD.bazel
```diff
@@ -11,6 +11,8 @@ load("@python_3_9//:defs.bzl", py_test_with_transition = "py_test")
 load("@python_versions//3.10:defs.bzl", compile_pip_requirements_3_10 = "compile_pip_requirements")
 load("@rules_python//python:defs.bzl", "py_binary", "py_library", "py_test")
 
+exports_files([".coveragerc"])
```
ewianda, line 14: I was expecting that this is not necessary, but it keeps failing without this

aignas: It is not needed only if the default visibility is public within the BUILD.bazel file.

examples/bzlmod/MODULE.bazel
```diff
@@ -36,6 +36,10 @@ python.toolchain(
     configure_coverage_tool = True,
     python_version = "3.10",
 )
+python.converage(
+    name = "coverage",
```
aignas, line 40: What does this `coverage` name mean?

aignas (commented):
This is certainly interesting. It seems that there are `//tests/python/python_tests.bzl` failing due to the mock `module_ctx` not being updated.

Shall we mark this as Ready for review and ask others to also have a look? Before doing that could you please update the PR description?

ewianda:
> This is certainly interesting. It seems that there are `//tests/python/python_tests.bzl` failing due to the mock `module_ctx` not being updated.
> 
> Shall we mark this as Ready for review and ask others to also have a look? Before doing that could you please update the PR description?

I am going to work on the description and get the tests to pass, then we  can open it up for review.

examples/bzlmod/MODULE.bazel
ewianda: It the name of the repository, so maybe calling it `repo_name` or maybe we don't need a name at all.

python/private/py_test_toolchain.bzl
```diff
@@ -0,0 +1,77 @@
...
+
+py_test_toolchain = rule(
+    implementation = _py_test_toolchain_impl,
+    attrs = {
+        "coverage_rc": attr.label(
+            allow_single_file = True,
+        ),
+    },
+)
+_TOOLCHAIN_TEMPLATE = """
```
aignas, line 42: Please create a macro that is used in the template - right now there is very little logic in here, but as time goes by and we make it more and more, having the code in a macro helps with maintainability and testability.

I would expect this to maybe have a test runner (e.g. `pytest`) in the long term, where the user could specify a `pytest` dependency for that.

python/private/python.bzl
```diff
@@ -850,6 +857,19 @@ The coverage tool to be used for a particular Python interpreter. This can overr
         ),
     },
 )
+_converage = tag_class(
```
aignas, line 860: For now this toolchain is completely separate from the `python` toolchain and it would be impossible to extend it in the future to be able to use dependencies from PyPI, because `python` extension is used inside the `pip` extension.

For this reason, I would like us to consider adding a separate `bzlmod` extension, especially, since this maybe could be always run with `dev_dependency` as it is about the test env.

In general this could have an API:
```
python_test = use_extension("@rules_python//python/extensions:python_test.bzl", "python_test")

python_test.configure(
    testrunner = "my_label_for_test_tool",  # could be `pytest-bazel` PyPI dep
    coverage_tool = "my_label_for_coverage_tool",  # should default to the rules_python provided one
    coveragerc = ".coveragerc",
)
```

The users who want a completely custom experience could define their own `pytest_toolchain` and register it instead of using the one defined by us.

aignas (commented):
Added some extra thoughts about the API design

examples/bzlmod/tests/BUILD.bazel
```diff
@@ -48,6 +48,12 @@ py_test(
     deps = ["//libs/my_lib"],
 )
 
+py_test(
```
ewianda, line 51: I observed that the COVERAGE_RC env is not available for `py_test_3_9` et.al. 
Is this expected or is there an issue with the way the module is being registered

python/private/py_test_toolchain.bzl
```diff
@@ -0,0 +1,108 @@
...
+        target_compatible_with = [],
+        exec_compatible_with = [],
+        toolchain = "{}_toolchain".format(name),
+        toolchain_type = toolchain_type,
+    )
+
+def _toolchains_repo_impl(repository_ctx):
+    kwargs = dict(
+        name = repository_ctx.name,
+        coverage_rc = str(repository_ctx.attr.coverage_rc),
```
ewianda, line 75: I was expecting that calling `render.str` will convert the `Label` to a string. I see that `render.str` is calling `repr`

aignas: Yup, `str` is the right thing to do. `render` utilities is for generating `BUILD.bazel` files and they should not be used in macros.

python/private/py_test_toolchain.bzl
```diff
@@ -0,0 +1,108 @@
...
+            for k, v in kwargs.items()
+        ])),
+    )
+    repository_ctx.file("BUILD.bazel", build_content)
+
+py_test_toolchain_repo = repository_rule(
+    _toolchains_repo_impl,
+    doc = "Generates a toolchain hub repository",
+    attrs = {
+        "toolchain_type": attr.string(doc = "Toolchain type", mandatory = True),
```
aignas, line 91: I think we should use a `attr.label` here.

python/private/python_test.bzl
```diff
@@ -0,0 +1,69 @@
...
+
+    Configure the test toolchain for setting coverage resource file.
+
+    """
+    for mod in module_ctx.modules:
+        for tag in mod.tags.configure:
+            register_py_test_toolchain(
+                coverage_rc = tag.coveragerc,
+                register_toolchains = False,
+            )
```
aignas, line 31: We should mark the `python_test` extension as reproducible (see the `python.bzl`) end of the impl.

examples/bzlmod/tests/BUILD.bazel
aignas: You can debug this with `--toolchain_resolution_debug=<name of the toolchain>`. I don't think this is expected.

ewianda:
The test under Bazel 6.4.0 test is failing. My guess is because this native java `py_test`, If that is the case is the a mechanism to run the test conditionally

examples/bzlmod/tests/BUILD.bazel
ewianda: the issue was that the rule transitions was not forwarding the env variables.

aignas:
Ah, this is because bazel 6.4 is not using the starlark implementation of the rules, so it is expected to not work for now.

examples/bzlmod/tests/BUILD.bazel
```diff
@@ -48,6 +48,24 @@ py_test(
...
```
aignas, line 51: You could use `bazel_features` in here and set `target_compatible_with` for the `py_test` targets based on the bazel version in that way.

ewianda: Thanks for the tip

aignas:
Idea for why the Windows build is failing - it is using a very different bootstrap mechanism and it could be that the coveragerc is not wired properly. I am not 100% sure how we should do this. @rickeylev, any thoughts here?

ewianda:
@rickeylev @aignas I have tried to figure out the issue with windows with no success. I noticed that 

`C:\Users\ELVISW~1\AppData\Local\Temp\Bazel.runfiles__dysvgw6\runfiles\_main\.coveragerc`  is **missing**

`C:\Users\ElvisWianda\Documents\rules_python\examples\bzlmod\bazel-bin\tests\coverage_rc_is_set_test.exe.runfiles\_main/.coveragerc`
**exists**

ewianda:
ping @aignas @rickeylev

aignas (commented):
I like the general direction, but my main question is about how we could implement the additions of the toolchain so that we could have different flavours of `py_test`.

python/private/py_test_rule.bzl
```diff
@@ -52,4 +53,7 @@ py_test = create_executable_rule(
     implementation = _py_test_impl,
     attrs = dicts.add(AGNOSTIC_TEST_ATTRS, _BAZEL_PY_TEST_ATTRS),
     test = True,
+    exec_groups = {
+        "test": exec_group(toolchains = [config_common.toolchain_type(PY_TEST_TOOLCHAIN_TYPE, mandatory = False)]),
+    },
```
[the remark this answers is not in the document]
ewianda: Based on the C++ Testing Toolchain approach, it is better to have this configured at the toolchain level rather than the rule level. If I understand this correctly, users will be required to specify `extend_foo` on every `py_test` rule

python/private/py_executable.bzl
```diff
@@ -326,6 +329,12 @@ def _get_runtime_details(ctx, semantics):
                 direct.append(effective_runtime.coverage_tool)
             if effective_runtime.coverage_files:
                 transitive.append(effective_runtime.coverage_files)
+        if is_test:
+            py_test_toolchain = ctx.exec_groups["test"].toolchains[PY_TEST_TOOLCHAIN_TYPE]
```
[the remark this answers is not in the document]
ewianda: I think we should use the ability to use a callback macro at the level of the toolchain. I didn't use that approach here since I was not adding an executable. Will change the implementation.

python/private/py_test_toolchain.bzl
```diff
@@ -0,0 +1,160 @@
...
+
+def _py_test_toolchain_impl(ctx):
+    return [
+        platform_common.ToolchainInfo(
+            py_test_info = PyTestProviderInfo(
+                get_runner = struct(
+                    func = _get_runner,
+                    args = {
+                        "coverage_rc": ctx.attr.coverage_rc,
+                    },
```
aignas, line 78: nit: I think we could use a lambda here so that the call site can just do `get_runner(my_extra_args)`.

python/private/python_test.bzl
```diff
@@ -0,0 +1,55 @@
...
+
+    Configure the test toolchain for setting coverage resource file.
+
+    """
+    for mod in module_ctx.modules:
+        for tag in mod.tags.configure:
+            register_py_test_toolchain(
+                name = "py_test_toolchain",
+                coverage_rc = tag.coveragerc,
+                register_toolchains = False,
```
aignas, line 31: I see that the `register_toolchains` arg here has been borrowed from the `uv` toolchain. I am personally not a fan for the `func` name to say `register_py_test_toolchain` and then there being an arg `register_toolchains = False`. So what are we doing here? Maybe just not registering the toolchains and calling the function `py_test_toolchain_repositories` would be more acurate?

python/private/python_test.bzl
```diff
@@ -0,0 +1,55 @@
...
+
+def _python_test_impl(module_ctx):
+    """Implementation of the `coverage` extension.
+
+    Configure the test toolchain for setting coverage resource file.
+
+    """
+    for mod in module_ctx.modules:
+        for tag in mod.tags.configure:
+            register_py_test_toolchain(
```
aignas, line 28: Should we allow non-root modules to do the registrations? I think we could just ignore any non-root modules.

aignas (commented):
In general I think I like the approach here and would like to hear what @rickeylev thinks about such a way to do dependency injection.

I can see how I could add a `@pypi//pytest_bazel` target in the `python_test.configure` call and in this way tell bazel that we have to use that package for running all of the tests. I would probably then need to somehow add the extra invocation of `python -m pytest_bazel` somewhere in the `py_executable.bzl` or related in order to benefit from the extra dependency in there.

Remaining bits before merging:
- [ ] Have more than just me believing in fruitfulness of this approach.
- [ ] Add this to the `//docs:BUILD.bazel` and add an EXPERIMENTAL note in the symbols.
- [ ] Profit?

ewianda:
> In general I think I like the approach here and would like to hear what @rickeylev thinks about such a way to do dependency injection.
> 
> I can see how I could add a `@pypi//pytest_bazel` target in the `python_test.configure` call and in this way tell bazel that we have to use that package for running all of the tests. I would probably then need to somehow add the extra invocation of `python -m pytest_bazel` somewhere in the `py_executable.bzl` or related in order to benefit from the extra dependency in there.
> 
> Remaining bits before merging:
> 
> * [ ]  Have more than just me believing in fruitfulness of this approach.
> * [ ]  Add this to the `//docs:BUILD.bazel` and add an EXPERIMENTAL note in the symbols.
> * [ ]  Profit?

@rickeylev any chance to take a look, I work on your suggestions @aignas

ewianda:
@aignas is this still something to pursue? I want to upgrade rules_python, and find myself needing to carry forward a patch for this.

aignas:
I'm torn about this. On one hand I don't think we have the coverage story sorted out properly, but I think the proposed PR may be a little bit hard to maintain. There was no consensus among the maintainers about maintaing this sort of API long term.

There was some desire to make this a user space pnoblem - like what pytest-bazel is doing, but there is a need for more integration. Maybe there can be a way in between.

ewianda:
> I'm torn about this. On one hand I don't think we have the coverage story sorted out properly, but I think the proposed PR may be a little bit hard to maintain. There was no consensus among the maintainers about maintaing this sort of API long term.
> 
> There was some desire to make this a user space pnoblem - like what pytest-bazel is doing, but there is a need for more integration. Maybe there can be a way in between.

Fair enough, I had proposed a low maintenance approach (open for debate) of just providing the coveragerc using an environemnt variable. This is how I am currently solving this as illustrated [here ](https://github.com/ewianda/rules_python/pull/4/files). This should not be far out, given that we already use the `PYTHON_COVERAGE` env variable to override the coverage tool. Thoughts

This approach was discused in https://github.com/bazel-contrib/rules_python/issues/1434

and implemented in https://github.com/bazelbuild/bazel/pull/19656/

PR #2251, opened 2024-09-25
refactor: add builders to make creating PyInfo/depset/runfiles easier

rickeylev:
This add some builders "classes" to make constructing PyInfo, depsets, and runfiles
a bit cleaner. Previously, building such objects required creating multiple local
variables for each arg/attribute of the object (e.g. list_of_files and list_of_depsets
for building a depset). Keeping track of all the variables was verbose and tedious.

To make it easier, encapsulate that bookkeeping in some objects.

This APIs will, in an upcoming PR, be exposed to users to make it easier to build and
merge PyInfo objects.

DO NOT MERGE:
 * Add tests
 * Depends on 2249


python/private/py_info.bzl
```diff
@@ -113,3 +116,124 @@ This field is currently unused in Bazel and may go away in the future.
...
+def _PyInfoBuilder_build(self):
+    if config.enable_pystar:
+        kwargs = dict(
+            direct_pyc_files = self.direct_pyc_files.build(),
+            transitive_pyc_files = self.transitive_pyc_files.build(),
+        )
+    else:
+        kwargs = {}
+
+    # todo: for bazel 6, make this return BuiltinPyInfo
```
aignas, line 222: in 2 months this will be an outdated comment. So it is fine if we never get to this. Is it blocking the merging of the PR? will it break users on bazel 6?

aignas (approved):

python/private/py_info.bzl
rickeylev: Oh oops, outdated comment. The _EffectivePyInfo trick addressed this. Removed.

Issue #1647, opened 2022-05-22
py_common API


UebelAndre:
### Description of the feature request:

Similar to [cc_common](https://bazel.build/rules/lib/cc_common) is there a `py_common` available?

### What underlying problem are you trying to solve with this feature?

I would like to have some rules which generate additional actions but would otherwise be interchangeable with a [py_library](https://bazel.build/reference/be/python#py_library). I've done similar things before using the `cc_common` API but cannot find one for python.

### Which operating system are you running Bazel on?

Any

### What is the output of `bazel info release`?

release 5.1.1

UebelAndre:
@rickeylev do you know if this would be something that's easy to do? I still find myself wanting to make custom rules that build python binaries. Having an interface for that would be fantastic

UebelAndre:
@rickeylev would it be acceptable to have an experimental flag that gates the use of some `py_common` struct that would contain a single function, `py_executable_base_impl`, for now?

https://github.com/bazelbuild/rules_python/blob/084b877c98b580839ceab2b071b02fc6768f3de6/python/private/common/py_executable.bzl#L140-L156

This would cover my primary use for `py_common` which is simply writing Bazel rules that equate to `py_binary` or `py_test`.

The only changes to `py_executable_base_impl` (or whatever would be a part of the interface) is that `ctx.attrs` and `ctx.files` are pulled in as a separate parameters so other rules have the chance to address missing fields without actually needing to add attributes to the rule. This approach would have some splash damage where access directly from `ctx` would need to be updated but the changes should be otherwise pretty straight forward.

UebelAndre:
@aignas do you have any thoughts here?

arrdem:
This is largely a dupe of https://github.com/bazelbuild/rules_python/issues/1860; interested in whatever the outcome here is.

UebelAndre:
I would be willing to help drive some MVP but I'm concerned about potentially wasting time without prior design sign-off from the maintainers.

aignas:
Playaing devil's advocate here - why would we have to support this API when `bazel` is working on [subrule](https://bazel.build/rules/lib/builtins/Subrule) support? Is it not possible to use the `py_binary` and friends together with the `subrule`?

Aside from that, I am personally not super familiar with that part of the code, so cannot comment on implications or the maintenance burden in the long run. The request in general looks reasonable, but I would love to not expose anything that is not needed and the `semantics` arg should not be exposed, IMHO.

[...]

PR #2252, opened 2024-09-25
feat: add public API for analysis-phase logic

rickeylev:
This adds a public API for rules (i.e. analysis-phase code) to use code from rules_python.
The main motivation for this is so that users can propagate PyInfo without having to know
all the fields of PyInfo and implement the merging logic. With upcoming PRs adding additional
fields to PyInfo, this becomes much more important.

The way the API is exposed is through a target. There are three reasons for this:
1. It avoids loading phase costs when the implementation of the API functions change.
   Within Google, this makes changes to rules_python much cheaper and easier to submit
   and revert. This also allows us to worry less about the loading-phase impact of
   our code.
2. Because a target can have dependencies, it allows us to hide some details
   from users. For example, if we want a flag to affect behavior, we can add it to the
   API target's attributes; users don't have to add it to their rule's attributes
3. By having the API take the user's `ctx` as an argument, it allows us to capture it
   and use it as part of future API calls (this isn't used now, but gives us
   flexibility in the future).


python/private/api/api.bzl
```diff
@@ -0,0 +1,55 @@
...
+    Args:
+        ctx: {type}`ctx` current rule ctx
+
+    Returns:
+        {type}`PyCommonApi`
+    """
+
+    # A generic provider is used to decouple the API implementations from
+    # the loading phase of the rules using an implementation.
+    return ctx.attr._py_common_api[ApiImplInfo].impl
```
aignas, line 45: I wonder if this could be used by in the `aspect_rules_py` where they are reimplementing the entire `py_*` rules. Is this abstraction powerful enough for `rules_python` to be completely overridden?

aignas (approved):
LGTM

python/private/api/api.bzl
rickeylev: Sort of. The main limitation is some operations still require the user ctx to have particular settings, e.g. in order to run the precompiler action, it needs the py_precompile exec group defined on the user rule. So the main missing piece is some way that allows users to define rules, where both us and users have a loose coupling, to reduce the amount of churn and breaking changes.

PR #2253, opened 2024-09-26
fix(bzlmod): set the default_version to the python_version flag

aignas:
With this change we set the default value of `--python_version` when
the `python.toolchain` is used in `bzlmod` and we generate the
appropriate config settings based on the registered toolchains and
given overrides by the root module.

This means that we expect the `--python_version` to be always set to
a non-empty value under `bzlmod` and we can cleanup code which was
handling `//conditions:default` case. This also means that we can
in theory drop the requirement for `python_version` in `pip.parse`
and just setup dependencies for all packages that we find in the
`requirements.txt` file and move on. This is left as future work
by myself or anyone willing to contribute. We can also start reusing
the same `whl_library` instance for multi-platform packages because
the python version flag is always set - this will simplify the layout
and makes the feature non-experimental anymore under bzlmod.

Summary:
* Add `@pythons_hub` to the `WORKSPACE` with dummy data to make
  pythons_hub work.
* Add `MINOR_MAPPING` and `PYTHON_VERSIONS` to pythons_hub to
  generate the config settings.
* Remove handling of the default version in `pypi` code under bzlmod.

Work towards #2081, #260, #1708


internal_setup.bzl
```diff
@@ -22,13 +22,25 @@ load("@rules_bazel_integration_test//bazel_integration_test:deps.bzl", "bazel_in
...
 load("//python/private/pypi:deps.bzl", "pypi_deps")  # buildifier: disable=bzl-visibility
 
 def rules_python_internal_setup():
     """Setup for rules_python tests and tools."""
 
     internal_config_repo(name = "rules_python_internal")
+    if not BZLMOD_ENABLED:
+        hub_repo(
+            name = "pythons_hub",
+            minor_mapping = {},
```
aignas, line 37: If we wanted, we could do extra plumbing within the `pythons_hub` and pass the correct values for `minor_mapping` and `python_versions` from the `//python:versions.bzl`, but given that WORKSPACE is on the way out and this is temporary, I thought it would be best to leave it like this.

tests/pypi/config_settings/config_settings_tests.bzl
```diff
@@ -528,7 +529,7 @@ def config_settings_test_suite(name):  # buildifier: disable=function-docstring
 
     config_settings(
         name = "dummy",
-        python_versions = ["3.8", "3.9", "3.10"],
+        python_versions = ["3.7", "3.8", "3.9", "3.10"],
```
aignas, line 532:
```suggestion
        # NOTE: We don't have any `3.7` in `//python:versions.bzl` and 
        # this value has been selected on purpose to test what happens if the value
        # is unknown to rules_python - i.e. we should still be able to work.
        python_versions = ["3.7", "3.8", "3.9", "3.10"],
```

python/private/pypi/config_settings.bzl
```diff
@@ -109,9 +109,14 @@ def config_settings(
 
     for python_version in [""] + python_versions:
         is_python = "is_python_{}".format(python_version or "version_unset")
-        native.alias(
+
+        # Do not rely on aliases in the `rules_python` so that we can just blindly
```
rickeylev, line 113: nit: I think this doc is trying to say:

> The aliases defined in `@rules_python//python/config_settings` may not have config settings for the versions we need, so define our own config settings instead.

?

python/private/pypi/config_settings.bzl
```diff
@@ -109,9 +109,14 @@ def config_settings(
...
+        # add all of the found wheels on the internet to the select statement.
+        native.config_setting(
             name = is_python,
-            actual = Label("//python/config_settings:" + is_python),
+            flag_values = {
+                Label("//python/config_settings:_python_version_major_minor"): python_version,
```
rickeylev, line 118: Hm, maybe we _should_ expose python_version_major_minor as a public setting. Otherwise, users don't have a good way to perform e.g. `select({"3.10": ..., "3.11": ...})`, which is the most common type of selection on python version.

Don't have to address in this PR though.

rickeylev (approved):

rickeylev:
CI's error is about running it under Bazel 6 -- I think it's trying to say pythons_hub isn't defined. Though, the (Mac, pystar, workspace) job is also failing and that is using Bazel 7. Looks like something to do with workspace builds.

rickeylev:
Ok, think I found the problem (wrong load in a test file). Pushed a fix. CI is acting _weird_ right now, though. Maybe it'll be in a better mood when you log on; I suggest pushing a no-op change and having it start over.

aignas:
> > This also means that we can in theory drop the requirement for python_version in pip.parse
> 
> Oh? How? Aren't we still stuck with the case where the same python version interpreter has to go find and download something? sdists?
> 
> I think I see what you mean for a pure-wheel build. Well, maybe. When requirements.in is resolved by e.g. compile_pip_requirements to a particular python-specific requirements.txt, the python version has, effectively, been decided. But -- I don't recall requirements.txt storing the python version? So how do we know that `foo == 1.2.3 # hash blabla` means el.g. `foo-1.2.3-cpython-313.tar.gz` (as opposed to `foo-1.2.3-cpython-310.tar.gz`) ?

The idea is that we can just always define `whl_library` instances for all of the dists we find in a requirements file and we don't need to know the version of the python interpreter that we are getting. Sometimes the wheels could be missing though, but in the issue linked to the PR (#1708), the same `requirements.txt` is used for each `python_version`, so sometimes `requirements.txt` can be multi-version compatible.

This is only true for `wheels` though. For `sdists` we would need to iterate through registered python toolchain repository minor versions and add a `whl_library` instance per python version for each `sdist` we find. This is until we can build `sdist` to `whl` in a build context.

Issue #2239, opened 2024-09-20
`extra_pip_args` do not seem to be passed to `whl_library`


swarren12:
# 🐞 bug report

### Affected Rule

I guess the affected rule is `whl_library`, but from a "dumb end user" point of view, it's `pip.parse`.

### Is this a regression?

Yes.

### Description

**Expected behaviour**: specifying either `experimental_index_url` or `extra_pip_args = [ "--index-url=..." ]` allows fetching from a custom index.
**Actual behaviour**: some parts of the build appear to resolve against the custom URL, but later parts fail.

I think this is related to https://github.com/bazelbuild/rules_python/issues/2152, but at least for me using the current `HEAD` of `rules_python` does not seem to fix the issue.

I'm currently using these rules in an air-gapped CI environment, so it does not have access to `pypi.org`. On v0.34.0, the following snippet is working fine:

```python
pip.parse(
    hub_name = "lib_python",
    experimental_index_url = "https://internal.repo/repository/pypi_proxy/simple/",
    extra_pip_args = [
        "--index-url=https://internal.repo/repository/pypi_proxy/simple/",
        "--trusted-host", "internal.repo",
    ],
    python_version = "3.11",
    requirements_lock = "//:requirements_lock.txt",
)
```

However, after upgrading to `0.35.0`, this breaks with the following error:

```
...
ERROR: .../scipy/BUILD.bazel:10:6 ...//scipy:pkg depends on ...scipy//:pkg in repository ..._scipy which failed to fetch. no such package '..._scipy//': 
rules_python:whl_library(..._scipy) FAIL: repo.execute: whl_library.ResolveRequirement(..._scipy, scipy==1.14.0): end: failure:
  command: .../external/rules_python~~python~python_3_11_host/python -m python.private.pypi.whl_installer.wheel_installer --requirement scipy==1.14.0 --isolated --extra_pip_args "{\"arg\":[]}" --pip_data_exclude "{\"arg\":[]}" --environment "{\"arg\":{}}"
...
WARNING: Retrying (Retry(total=0, connect=None, read=None, redirect=None, status=None)) after connection broken by 'NewConnectionError('<pip._vendor.urllib3.connection.HTTPSConnection object at 0x7f09e0f38cd0>: Failed to establish a new connection: [Errno 111] Connection refused')': /simple/scipy/
ERROR: Could not find a version that satisfies the requirement scipy==1.14.0 (from versions: none)
[...]

PR #2254, opened 2024-09-26
fix: pass `extra_pip_args` to `parse_requirements`

swarren12:
Bug fix for #2239.

The `extra_pip_args` variable was not being correctly passed to `parse_requirements()`, causing further calls to ignore anything that was set here. 

This was a problem if, for example, the `--trusted-host` argument was needed to talk to the index set by `experimental_index_url`.


swarren12:
Fixed by https://github.com/bazelbuild/rules_python/pull/2258 so will close this one.

PR #2257, opened 2024-09-26
chore: don't add migration tag if Starlark implementation is enabled

rickeylev:
This avoids the tag being added when it doesn't need to be, which can look confusing to users without context about what it means.

Work towards #1361


groodt (approved):
LGTM

PR #2265, opened 2024-10-01
feat(toolchain): add python 3.13.0, updated other versions and bumped coverage to 7.6.1

vfdev-5:
Description:
- Added python 3.13.0 toolchain
- Updated other versions in the toolchain
- Bumped coverage version to 7.6.1
- Updated CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,6 +53,16 @@ A brief description of the categories of changes:
   interpreter.
 * (api) Added {obj}`merge_py_infos()` so user rules can merge and propagate
   `PyInfo` without losing information.
+* (toolchains) New Python versions available: 3.13.0 using the [20241008] release.
+* (toolchains): Bump default toolchain versions to:
+    * `3.8 -> 3.8.20`
+    * `3.9 -> 3.9.20`
+    * `3.10 -> 3.10.15`
+    * `3.11 -> 3.11.10`
+    * `3.12 -> 3.12.7`
+[20241008]: https://github.com/indygreg/python-build-standalone/releases/tag/20241008
+* (coverage) Add support for python 3.13 and bump `coverage.py` to 7.6.1.
+
 
 ### Removed
 * Nothing yet
--- a/python/versions.bzl
+++ b/python/versions.bzl
@@ -130,6 +130,17 @@ TOOL_VERSIONS = {
         },
         "strip_prefix": "python",
     },
+    "3.8.20": {
+        "url": "20241002/cpython-{python_version}+20241002-{platform}-{build}.tar.gz",
+        "sha256": {
+            "aarch64-apple-darwin": "2ddfc04bdb3e240f30fb782fa1deec6323799d0e857e0b63fa299218658fd3d4",
+            "aarch64-unknown-linux-gnu": "9d8798f9e79e0fc0f36fcb95bfa28a1023407d51a8ea5944b4da711f1f75f1ed",
+            "x86_64-apple-darwin": "68d060cd373255d2ca5b8b3441363d5aa7cc45b0c11bbccf52b1717c2b5aa8bb",
+            "x86_64-pc-windows-msvc": "41b6709fec9c56419b7de1940d1f87fa62045aff81734480672dcb807eedc47e",
+            "x86_64-unknown-linux-gnu": "285e141c36f88b2e9357654c5f77d1f8fb29cc25132698fe35bb30d787f38e87",
+        },
+        "strip_prefix": "python",
+    },
     "3.9.10": {
         "url": "20220227/cpython-{python_version}+20220227-{platform}-{build}.tar.gz",
         "sha256": {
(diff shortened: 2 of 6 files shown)


python/versions.bzl
```diff
@@ -497,6 +497,19 @@ TOOL_VERSIONS = {
         },
         "strip_prefix": "python",
     },
+    "3.13.0": {
```
aignas, line 500: Please update the rest of the toolchains to use the same indygreg release. If the same version is already present in the `TOOL_VERSIONS`, then we should overwrite the old values with the new values.

aignas (changes requested):

python/versions.bzl
vfdev-5: @aignas you mean to update for all other python versions like 3.12.X etc ?
I wonder whether there is an automatic way to get sha256 hashes for the versions? For 3.13 I did that manually, downloading tar.gz and running `sha256sum`...

aignas:
If you run the target defined in https://github.com/bazelbuild/rules_python/blob/413690fd09b068ab43ec9c1d37d20d8017747f2c/python/private/BUILD.bazel#L509 then rules_python will download all of the toolchains and print the checksums. The only thing you need to have is the list of empty platform identifiers, like:
```starlark
    "3.13.0": {
        "url": "20240909/cpython-{python_version}rc2+20240909-{platform}-{build}.tar.gz",
        "sha256": {
            "aarch64-apple-darwin": "",
            "aarch64-unknown-linux-gnu": "",
            "ppc64le-unknown-linux-gnu": "",
            "s390x-unknown-linux-gnu": "",
            "x86_64-apple-darwin": "",
            "x86_64-pc-windows-msvc": "",
            "x86_64-unknown-linux-gnu": "",
        },
        "strip_prefix": "python",
    },
```

Sorry for being late here.

I think you also need to change any reference to `MINOR_MAPPING`, which is using the values in the [folowing way](https://github.com/bazelbuild/rules_python/blob/413690fd09b068ab43ec9c1d37d20d8017747f2c/WORKSPACE#L49). I think it is the only place. Adding a new Python version is always slightly annoying.

CHANGELOG.md
```diff
@@ -38,6 +38,14 @@ A brief description of the categories of changes:
 ### Added
 * (py_wheel) Now supports `compress = (True|False)` to allow disabling
   compression to speed up development.
+* (toolchains) New Python versions available: 3.13.0rc2 using the [20240909] release.
+* (toolchains): Bump default toolchain versions to:
+    * `3.8 -> 3.8.20`
+    * `3.9 -> 3.9.20`
+    * `3.10 -> 3.10.15`
+    * `3.11 -> 3.11.10`
+    * `3.12 -> 3.12.6`
```
aignas, line 47: I think this part of the change is lost somewhere in the `git` dungeons. Could you please check if you have pushed it?

vfdev-5:
@aignas thanks for the review! I agree there was something weird in the last commit in addition to a merge conflict with CHANGELOG.md. I fixed it and now it should be good.

python/versions.bzl
```diff
@@ -497,15 +547,42 @@ TOOL_VERSIONS = {
...
+            "ppc64le-unknown-linux-gnu": "0fb6ac723bc534d2e92e3d454b494d6f9641d5e47cb92308b4ae792d085527ec",
+            "s390x-unknown-linux-gnu": "a61b37280c9ff311a88c7c267a1bf3a5acc7c918d5d2bb68907ce963f8b8d04d",
+            "x86_64-apple-darwin": "60eaf2c9a63926f5edfe7a0ecf4b2fbb37de6764f0c0100be245d098e3995da7",
+            "x86_64-pc-windows-msvc": "1c74dd0499ef7d993d8942ebe6692c90c50d09719e0335f8e210b91277f8e7c0",
+            "x86_64-unknown-linux-gnu": "04ffccbbb1beedd0bc80347c061fe527833001681ee543c3ba85dac5134ed500",
+        },
+        "strip_prefix": "python",
+    },
+    "3.13.0": {
+        "url": "20241002/cpython-{python_version}rc3+20241002-{platform}-{build}.tar.gz",
```
aignas, line 564: Since the official release of 3.13.0 has been cut, I think we can wait for indigreg release to crop up once again before merging this. That will save some back-and-forth.

aignas:
It seems that they were really fast: https://github.com/indygreg/python-build-standalone/releases/tag/20241008 contains the new builds and even the free-threading ones. I think adding the freethreading values in the `TOOL_VERSIONS` should be done in #2129.

aignas:
Could you also please add new binaries for `coverage.py`? This is needed since it is a new major version of python. I think just upgrading to the latest version (https://pypi.org/project/coverage/#files) should be sufficient. The docs on how to do it are in `DEVELOPING.md` file.

Feel free to leave notes there about how to add a new minor python version.

vfdev-5:
> Could you also please add new binaries for `coverage.py`? This is needed since it is a new major version of python. I think just upgrading to the latest version (https://pypi.org/project/coverage/#files) should be sufficient. The docs on how to do it are in `DEVELOPING.md` file.
> 
> Feel free to leave notes there about how to add a new minor python version.

I'm not very sure how this works, I updated the PR, please review

aignas (approved):
LGTM, thanks!

PR #2274, opened 2024-10-05
chore: support removal of builtin providers

comius:
The builtin providers PyInfo, PyRuntimeInfo, and PyCcLinkParamsProvider are being removed,
which means Bazel throws an error while compiling bzl files if there is a reference to a
top-level symbol that doesn't exist anymore. For backwards compatibility, rules_python
consumes/produces these providers, so the symbols are used in various places.

To fix, use `native.legacy_globals` and Bazel version detection to conditionally emit
the symbols into `@rules_python_internal`. If they aren't present, they are reported
as None.

This mimics equivalent functionality in bazel_features; bazel_features isn't used because
it would require users to update their WORKSPACE to initialize some dependencies before
rules_python can perform its initialization.

Removal of the builtin symbols is controlled by `--incompatible_autoload_externally`
(which is in Bazel 8 and has been cherry-picked into earlier version). If the flag is
enabled with "@rules_python" or "-@rules_python" the providers are removed from Bazel.


aignas (commented):
I am somewhat opposed to having a breaking change to our `WORKSPACE` users, but looking at the code I don't see a quick workaround for that.

I am wondering if there is a different way to achieve this? I looked at the repos that we actually [need] from `bazel_features` and it seems that under `WORKSPACE` there is no way to initialize them inside the `py_repositories` call. So there is definitely no way but to have this extra call done after `py_repositories()`. I think having `bazel-features` in the long term could be inevitable and could make the code cleaner, but the fact that all of the `WORKSPACE` users would need to do these additions is unfortunate.

That said, everyone should migrate to `bzlmod` anyway, so this could be a good push to do this. Any other opinions? We at least need a decent note in the `CHANGELOG.md` explaining the migration steps.

[need]: https://github.com/bazel-contrib/bazel_features/blob/main/private/repos.bzl

rickeylev:
Is there a reason it can't go in internal_config_repo? I didn't fully trace what bazel_features is doing, but it looked like it was just generating code like:

```
globals = struct(
  PyInfo = getattr(getattr(native, "legacy_globals", None), "PyInfo", None)
```

That looks fairly simple to drop into python/private/internal_config_repo.bzl

rickeylev:
I updated the rules_python_internal repo to do the globals detection and updated reexports.bzl.

I think its working? I think bazel@head still isn't quite ready, so wasn't able to fully test with autoloading enabled (got some undefined symbol errors for stuff in bazel_tools, but those were also present with the PR as it was before I made my modifications). I didn't have a chance to lookup how the autoloading flag works, either, so maybe I set it wrong.

I'm not sure I'll have time tonight to undo the py_repositories_deps() changes.

python/private/py_repositories.bzl
```diff
@@ -59,4 +59,10 @@ def py_repositories():
         sha256 = "2037875b9a4456dce4a79d112a8ae885bbc4aad968e6587dca6e64f3a0900cdf",
         strip_prefix = "rules_cc-0.0.9",
     )
+    http_archive(
```
aignas, line 62: nit: I don't think we need this anymore if we are using the `rules_python_internal` repo.

rickeylev: Done

rickeylev:
Ok, finished cleanup etc. I think this is good to go

rickeylev (approved):

PR #2278, opened 2024-10-08
refactor!(bzlmod): introduce pypi.install extension

### LINKED ISSUE #2268: `experimental_index_url` in dependency tree breaks other pip.parse includes

```
```

aignas:
With this PR we are introducing a new extension for managing PyPI dependencies
and it seems that we would fix #2268 in doing so. The code has been sitting
around for long enough so that we have ironed out most of the bigger bugs
already and the maintainers have agreed for long enough that pip.parse is
not the best named, so I have chosen pypi.install.

This should not actively break the code and will make some of the experimental
parameters noop. This could yield to weird failures in cases where people are
building docker containers, so I am thinking that we should potentially add
a helpful error message prompting the users to migrate. However I am not
sure how that works out when multiple module versions are at play in the same
module graph.

Work towards #260


aignas:
I don't know if I am super happy about the breaking change here, but I do think fixing the bug here takes the priority. Let me know what you think. I think we have discussed enough times and agree on the new better name here.

docs/pypi-dependencies.md
```diff
@@ -44,6 +45,78 @@ hermetic host python interpreter on this platform. Linux and OSX users should se
...
+got all of the available distributions, it will select the right ones depending
+on the `sha256` values in your `requirements_lock.txt` file. The compatible
+distribution URLs will be then written to the `MODULE.bazel.lock` file. Currently
+users wishing to use the lock file with `rules_python` with this feature have
+to set an environment variable `RULES_PYTHON_OS_ARCH_LOCK_FILE=0` which will
+become default in the next release.
+
+Fetching the distribution information from the PyPI allows `rules_python` to
+know which `whl` should be used on which target platform and it will determine
+that by parsing the `whl` filename based on [PEP600], [PEP656] standards. This
```
rickeylev, line 105:
```suggestion
that by parsing the `whl` filename based on [PEP600] and [PEP656] standards. This
```

docs/pypi-dependencies.md
```diff
@@ -44,6 +45,78 @@ hermetic host python interpreter on this platform. Linux and OSX users should se
...
+allows the user to configure the behaviour by using the following publicly
```
rickeylev, line 106: Something is off about what this sentence is introducing and trying to say. I think it's trying to say that these flags can be used to control which wheels are used by a binary?

python/private/pypi/extension.bzl
```diff
@@ -461,7 +464,7 @@ def _pip_impl(module_ctx):
     is_extension_reproducible = True
 
     for mod in module_ctx.modules:
-        for pip_attr in mod.tags.parse:
+        for pip_attr in (getattr(mod.tags, "parse", None) or getattr(mod.tags, "install")):
```
rickeylev, line 467:
```suggestion
        for pip_attr in (getattr(mod.tags, "parse", None) or mod.tags.install):
```

rickeylev (commented):
re: new extension and backwards compatibility:

The main thing I have in mind is how we transition from pip.parse to pypi.install. Some specific questions:

1. If pypi.install is a drop-in replacement, why does it need to be a separate extension?
2. Why can't we transition pip.parse to the new semantics under-the-hood? I know the `pip.parse` name isn't as nice as `pypi.install`, but that's just aesthetics. 
3. Is there something pip.parse can do that pypi.install can't do, or vice versa?

aignas:
Taking a step back a little as I am not sure that this should necessarily be merged like this.

1. To fix the bug, we could instead only allow root modules to enable `experimental_index_url` for now.
2. The missing pieces that the new code does not support yet:
    1. Pull dependencies when hashes are not specified. I've seen many people in the wild using `pip freeze` to generate the requirements files and I think we should just get all of the wheels that we can find if the hashes are not present in the requirements files. This can be done as a separate PR.
    2. Direct URL dependency handling - right now we assume that we always should contact the SimpleAPI. If the URL is in the requirements file, we should just use it.

docs/pypi-dependencies.md
rickeylev: stet on this comment; i see now this was just moving content.

aignas:
See #2268 for reasons for closing this.

PR #2285, opened 2024-10-10
chore: move files out of private/common

rickeylev:
The "common" subdir was just carried over from the original file layout in Bazel. It's not
necessary in rules_python. Remove it to simplify, since various newly added code is
going directly into private, not into private/common.


mickvangelderen:
Will the rules source link be updated at https://bazel.build/reference/be/python#rules? It points to https://github.com/bazelbuild/rules_python/blob/main/python/private/common/py_binary_rule_bazel.bzl.

rickeylev:
Thanks for pointing that out! I haven't look at the Bazel hosted docs for the py rules in a awhile and didn't know they were pointing there. I'll file an issue against Bazel.

Issue #1675, opened 2024-01-08
Enabling bzlmod causes workspace toolchains to no longer be registered


rickeylev:
# 🐞 bug report

### Affected Rule

python_register_toolchains in WORKSPACE files

### Is this a regression?

Unclear

### Description

Enabling bzlmod, but still using a workspace file (i.e. an empty module file), causes the python toolchains to no longer be registered.

## 🔬 Minimal Reproduction

Modify tests/cc/py_cc_toolchain_registered in the following ways:
 * Comment out everything in MODULE.bazel
 * Change the toolchain being looked up to `@rules_python//python:toolchain_type`

Run `bazel test ... --enable_bzlmod`

**Actual behavior:**

Toolchain debug output says:

> Selected @@bazel_tools//tools/python:_autodetecting_py_runtime_pair

This means its using python from the environment's path (i.e. some system python) instead of the hermetic runtimes that were configured in WORKSPACE.

**Expected behavior** 

Toolchain debug output should print

> Selected @@python_3_11_x86_64-unknown-linux-gnu//:python_runtimes

(the name of the repo set in WORKSPACE).

**Analysis**

The cause of this is `python/repositories.bzl` detects if bzlmod is enabled, and, if so, disables toolchain registration. This was done because `native.register_toolchains` isn't normally available when bzlmod is enabled. However, when the code path is through a WORKSPACE file, it appears that symbol _is_ available.

The immediate fix is to move the "if bzlmod enabled, don't do X" logic out of python/repositories.bzl itself and into the bzlmod extension. The extension logic calls that same `python_register_toolchains` function. The "X" logic here is two parts:
1. Whether to call native.register_toolchains
2. Whether to call `toolchains_repo()` (which creates the repo with all the `toolchain()` calls)

Now that we have integration tests again, it's feasible to construct a test for this.

The thing to double check is toolchain precedence. I can't remember offhand what the precedence of workspace-registered toolchains are when bzlmod is enabled. The main thing we're looking for is that the workspace-registered toolchains are after the module.bazel ones -- this is necessary so that the version-aware rules pick up their correct toolchain. The workspace `python_register_toolchains()` call will create version-unaware toolchains, which will match everything

**Workarounds**

The best workaround is to update MODULE.bazel to register the toolchain.

```
# MODULE.bazel
bazel_dep(name = "rules_python")

python = use_extension("@rules_python//python/extensions/python.bzl", "python")
python.toolchain(python_version="3.10") # Or whatever version

[...]

PR #2289, opened 2024-10-11
fix(bzlmod): let workspace-invoked python_register_toolchains to register toolchains

rickeylev:
While migrating to bzlmod, users may have a hybrid build where WORKSPACE contains
python_register_toolchain() calls. Because these calls originate from WORKSPACE files,
the `native.register_toolchain` APIs are still available. At the same time, we still
detect that bzlmod is enabled, and thus disable calling those functions. The net
effect is, users aren't able to register toolchains in such a hybrid setup. At the
same time, because the code path is shared, we can't have the bzlmod toolchain code
try to call them, as it'll fail.

To accomodate both cases, have the bzlmod toolchain code pass a special arg so that
`python_register_toolchains` knows to skip parts that don't apply to the bzlmod toolchain
invocation.

This was all unwound by some users and pointed out in a Slack thread. A few people are
manually carrying an equivalent patch for a working hybrid mode.

Fixes https://github.com/bazelbuild/rules_python/issues/1675

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,9 @@ A brief description of the categories of changes:
 * (rules) `compile_pip_requirements` passes `env` to the `X.update` target (and
   not only to the `X_test` target, a bug introduced in
   [#1067](https://github.com/bazelbuild/rules_python/pull/1067)).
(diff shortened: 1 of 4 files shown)


aignas (approved):
Nice!

PR #2293, opened 2024-10-11
chore: remove references to @rules_cc//cc:defs.bzl

comius:
Referring to @rules_cc//cc:defs.bzl, refers to @protobuf//bazel:cc_proto_library.bzl, which fetches protobuf repository.
Referring directly to what's needed limits the fetches just to rules_cc.

Fix reference to bzl libs in rules_cc that are needed for docs generation.

This requires rules_cc 0.0.13 or higher.

Work towards https://github.com/bazelbuild/rules_python/issues/2387, https://github.com/bazelbuild/rules_python/issues/2378


meteorcloudy:
I don't see any reason the RBE jobs have to build with Bazel 6.4.0 and ubuntu1604, please upgrade to Bazel 7.3.2 and ubuntu2204

rickeylev:
We still support Bazel 6, so we should have an RBE for that. Do you mean for the Bazel 7 RBE job? It looks like its using 7.1 (which is a bit surprising; I thought our CI was setup to use 7.x, i.e. latest 7)

meteorcloudy:
I'm not sure which one is the Bazel 7 RBE job, but the for those RBE jobs, you should upgrade to `rbe_ubuntu2004`.

If Bazel 6 is still supported, then we need to fix rules_cc, sent https://github.com/bazelbuild/rules_cc/pull/253

aignas:
FYI, once bazel 8 is out, we will start removing support for bazel 6, but if we can maintain compatibility in an easy way that wouldn't hurt. As for bazel version, we are defining it in `.bazelversion` in the `rules_python` repo, maybe we should remove that file to instead test with the latest available stable version?

Gray-null (approved):

rickeylev:
* I synced this to head
* https://github.com/bazelbuild/rules_python/pull/2316 is out to update the RBE ubuntu version.
* Also set .bazelversion to 7.x (i.e. 7.3.2)

CI still show some failures, though. Some look related to compiling protobuf?

rickeylev:
Once https://github.com/bazelbuild/rules_python/pull/2379 is merged, I think this PR will be able to finally proceed. That PR drops Bazel 6 support and updates to rules_cc 0.0.14, which I think are prereqs for moving past rules_cc 0.0.9

rickeylev:
Updated to head, CI looking green. Looks like 2379 helped with this quite a bit.

rickeylev (approved):

PR #2295, opened 2024-10-11
fix(sphinxdocs),deps: allow using sphinx_stardoc with bzlmod; add stardoc 0.6.2 as dependency

rickeylev:
Using the sphinx_stardoc rule doesn't work with bzlmod because it refers to the stardoc repo, which was a dev dependency. This means, even if a user's root module has visibility to stardoc, rules_python doesn't.

To fix, make stardoc a non-dev dependency. With bzlmod semantics, stardoc won't actually be loaded unless a user depends on it by using sphinxdocs.

I ran into this while trying to port rules_testing over to sphinxdocs. The pigweed project
recently started using sphinxdocs and didn't run into this problem, but I'm pretty
sure that's because they're using workspace still, which doesn't have the visibility
constraints that bzlmod does.


aignas:
FYI, the alternative is to define the `sphinx_build_binary` and pass the dependencies yourself like I did it here: https://github.com/aignas/pytest-bazel/blob/main/docs/BUILD.bazel#L58

This is not ideal, so actually maintaining the `sphinx` deps is not bad. Should we have a sphinx docs toolchain where we register the default value?

aignas (approved):
Should we add a note in the docs? I think this may force users onto a newer version of stardoc in their own setup.

rickeylev:
Good point. Done. It'd be nice to be able to avoid it as a non-dev dependency, but I couldn't figure out any clean way to do so.

PR #2314, opened 2024-10-18
refactor(bzlmod): refactor pip_parse to make it more testable

aignas:
This PR introduces a new `parse_modules` function in the `pypi/extension.bzl`
code to mimic the structure of the `python` extension and to make it easier to
write unit tests against the extension itself.

With this we can:
* Write unit tests to see what `whl_library` repos would be created.
* Mock out the SimpleAPI interaction by pre-populating the `simpleapi_cache`.

This is just making the seam and the tests, etc will be written in followup PRs.

### DIFF
--- a/python/private/pypi/extension.bzl
+++ b/python/private/pypi/extension.bzl
@@ -367,15 +367,18 @@ You cannot use both the additive_build_content and additive_build_content_file a
     for module in module_ctx.modules:
         for attr in module.tags.override:
             if not module.is_root:
-                fail("overrides are only supported in root modules")
+                _fail("overrides are only supported in root modules")
+                return None
 
             if not attr.file.endswith(".whl"):
-                fail("Only whl overrides are supported at this time")
+                _fail("Only whl overrides are supported at this time")
+                return None
 
             whl_name = normalize_name(parse_whl_name(attr.file).distribution)
 
             if attr.file in _overriden_whl_set:
-                fail("Duplicate module overrides for '{}'".format(attr.file))
+                _fail("Duplicate module overrides for '{}'".format(attr.file))
+                return None
             _overriden_whl_set[attr.file] = None
 
             for patch in attr.patches:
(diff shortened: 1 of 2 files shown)


python/private/pypi/extension.bzl
```diff
@@ -86,7 +62,35 @@ You cannot use both the additive_build_content and additive_build_content_file a
...
+        group_map,
+        simpleapi_cache,
+        exposed_packages,
+        whl_libraries,
+        available_interpreters = INTERPRETER_LABELS,
+        simpleapi_download = simpleapi_download):
+    """create all of the whl repositories
+
+    Args:
+        module_ctx: TODO
```
rickeylev, line 80: Please add docs

IMHO just a simple type description would suffice

tests/pypi/extension/parse_modules_subject.bzl
```diff
@@ -0,0 +1,107 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+""
+
+load("@rules_testing//lib:truth.bzl", "subjects")
+
+def parse_modules_subject(info, *, meta):
```
rickeylev, line 19: this code is already written, so a bit moot now: there is subjects.struct that would probably eliminate a lot of the boilerplate.

rickeylev (approved):

python/private/pypi/extension.bzl
aignas: Did that and instead of modifying values in places I decided to return the values and sort them properly because the lock file would get constantly changed.

tests/pypi/extension/parse_modules_subject.bzl
aignas: Removed this code and used `subjects.struct` instead. Was a drop-in replacement.

PR #2315, opened 2024-10-19
docs: give user-friendly permalinks in changelog

rickeylev:
The changelog header text doesn't turn into nice linkable ids. e.g. the "0.37.0" header
is simply "id1" (and similar for all the other headers).

To fix, use block attributes (made possible by the MyST attrs_block extension) to give them
ids manually. The ids start with "v" and use dashes because a dotted number isn't
accepted as a valid id.

It was a bit tedious to update all the headers with ids, so I only did it for the last few
releases.


groodt (approved):
LGTM

PR #2317, opened 2024-10-20
chore: set 7.x in bazelversion

rickeylev:
This just upgrades what Bazel version we run by default during development.


groodt (approved):
LGTM

aignas (approved):

PR #2325, opened 2024-10-22
feat(bzlmod): cross-platform builds without experimental_index_url

### LINKED ISSUE #2268: `experimental_index_url` in dependency tree breaks other pip.parse includes

aignas:
With this change we finally generate the same lock file within the
legacy code `pip.parse` code path and it allows to slowly transition to
using the new code path as much as possible without user doing anything.

This moves the selection of the host-compatible lock file from the
extension evaluation to the build phase - note, we will generate extra
repositories here which might not work on the host platform, however, if
the users are consuming the `whl_library` repos through the hub repo
only, then everything should work. A known issue is that it may break
`bazel query` and in these usecases it is advisable to use `cquery`
until we have `sdist` cross-building from source fully working.

Summary:
- feat: reuse the `render_pkg_aliases` for when filename is not known
  but platform is known
- feat: support generating the extra config settings required
- feat: `get_whl_flag_versions` now generates extra args for the rules
- feat: make lock file generation the same irrespective of the host
  platform
- test: add an extra test with multiple requirements files
- feat: support cross-platform builds using `download_only = True` in
  legacy setups

Note, that users depending on the naming of the whl libraries will need
to start using `extra_hub_aliases` attribute instead to keep their
setups not relying on this implementation detail.

Fixes #2268
Work towards #260


aignas:
@keith, could you please check to see if this is fixing the issues you were seeing previously? The CI is not fully green, but the examples and unit tests are passing, so I think this might be ready for a spin.

tests/pypi/extension/extension_tests.bzl
```diff
@@ -150,14 +150,100 @@ def _test_simple(env):
...
+        ],
+    }})
+    pypi.whl_libraries().contains_exactly({
+        "pypi_315_simple": {
+            "dep_template": "@pypi//{name}:{target}",
+            "python_interpreter_target": "unit_test_interpreter_target",
+            "repo": "pypi_315",
+            "requirement": "simple==0.0.1 --hash=sha256:deadbeef",
+        },
+    })
```
aignas, line 173: I am happy that my selection of the mock `ctx.os` values was already validating the bug in question.

python/private/pypi/extension.bzl
```diff
@@ -319,35 +318,32 @@ def _create_whl_repos(
...
+            is_exposed = is_exposed or requirement.is_exposed
+            if get_index_urls:
+                logger.warn(lambda: "falling back to pip for installing the right file for {}".format(requirement.requirement_line))
+
+            whl_library_args["requirement"] = requirement.requirement_line
+            if requirement.extra_pip_args:
+                whl_library_args["extra_pip_args"] = requirement.extra_pip_args
+
+            # We sort so that the lock-file remains the same no matter the order of how the
+            # args are manipulated in the code going before.
```
aignas, line 331:
```suggestion
```

keith:
it's possible it's just fighting getting the lockfile updated, but with this i see:

```
% bz mod graph --lockfile_mode=update
ERROR: module extension "pip" from "@@rules_python~//python/extensions:pip.bzl" does not generate repository "pip_deps_311_torch", yet it is imported as "pip_deps_311_torch" in the usage at .../MODULE.bazel:237:20 (did you mean 'pip_deps_311_munch'?). Type 'bazel help mod' for syntax and help.
```

CHANGELOG.md
```diff
@@ -31,7 +31,11 @@ A brief description of the categories of changes:
 
 {#v0-0-0-fixed}
 ### Fixed
-- Nothing yet
+- The extension evaluation has been adjusted to always generate the same lock
+  file irrespective if `experimental_index_url` is set by any module or not.
+  Fixes [#2268](https://github.com/bazelbuild/rules_python/issues/2268). A known
+  issue is that it may break `bazel query` and in these usecases it is advisable
```
rickeylev, line 37:
```suggestion
  issue is that it may break `bazel query` and in these use cases it is advisable
```

CHANGELOG.md
```diff
@@ -31,7 +31,11 @@ A brief description of the categories of changes:
...
+  to use `cquery` until we have `sdist` cross-building from source fully working.
```
rickeylev, line 38: download_only=True is also an option, right? Assuming one is OK with not having sdists

python/private/pypi/extension.bzl
```diff
@@ -319,35 +318,30 @@ def _create_whl_repos(
...
-            # packages they want to download, in that case there will be always
-            # a requirement here, so we will not be in this code branch.
-            continue
-        elif get_index_urls:
-            logger.warn(lambda: "falling back to pip for installing the right file for {}".format(requirement.requirement_line))
-
-        whl_library_args["requirement"] = requirement.requirement_line
-        if requirement.extra_pip_args:
-            whl_library_args["extra_pip_args"] = requirement.extra_pip_args
+        for i, requirement in enumerate(requirements):
```
rickeylev, line 321: This is the key bit, right? Before, it would filter the requirements to something matching the host platform. After, it's now generating a repo for each one, and the logic is approximately the same for the index-url case (I see it generates a whl_alias() struct with repo/version/platform)

rickeylev (approved):

aignas:
> it's possible it's just fighting getting the lockfile updated, but with this i see:
> 
> ```
> % bz mod graph --lockfile_mode=update
> ERROR: module extension "pip" from "@@rules_python~//python/extensions:pip.bzl" does not generate repository "pip_deps_311_torch", yet it is imported as "pip_deps_311_torch" in the usage at .../MODULE.bazel:237:20 (did you mean 'pip_deps_311_munch'?). Type 'bazel help mod' for syntax and help.
> ```
> 
> looks like i hit that with or without experimental_index_url

@keith, I should have explained better in the PR message/changelog, that with this change you will have multiple `torch` variants registered and they will be `pip_deps_311_torch__0`, `pip_deps_311_torch__1` and similar. The spoke repository naming is an implementation and should not be relied on too much. What are you using from the spoke repository? Some headers? In this case the best option would be to use `@pip_deps//torch` and `whl_filegroup` or similar to do that.

python/private/pypi/extension.bzl
aignas: Correct.

CHANGELOG.md
aignas: It could be, but I haven't tested with regular requirements files, let me add a unit test to see how that would get wired through.

keith:
> @keith, I should have explained better in the PR message/changelog, that with this change you will have multiple `torch` variants registered and they will be `pip_deps_311_torch__{'_'.join(target_platforms)}`. The spoke repository naming is an implementation and should not be relied on too much. What are you using from the spoke repository? Some headers? In this case the best option would be to use `@pip_deps//torch` and `whl_filegroup` or similar to do that, what do you think?

we are using whl_mods and additive_build_content_file to add more than just filegroups, we also add cc_library targets. so does the same recommendation apply for how to access them?

aignas:
Yes. Does the `whl_filegroup` work for you? https://rules-python.readthedocs.io/en/latest/api/rules_python/python/pip.html#whl_filegroup

You can extract files from the `whl` and then define the `cc_library` targets outside spoke repository.

aignas:
I think I need advice on the following points:
* What should be the scheme of the generated `whl_library` repositories? Right now I am appending the target platform identifiers in order to make this deterministic and somewhat human readable. Are there better options?

I`ve been thinking about @keith usecase a little more and I think I would like to have a solution that:
1. Does not require fetching/extracting the libraries whenever the annotations change.
2. Have as little as possible in the whl_library `BUILD.bazel`.
3. Be compatible with `cc_*` rules so that wheels providing shared libraries or header files can be used.
4. Ideally the usage should come naturally through the hub repo so that users don't need to depend on implementation details of the naming of the repository rules - having spent enough time in this space I think that we can only really promise the contents of the hub repo as a stable API.

rickeylev:
> * What should be the scheme of the generated `whl_library` repositories? Right now I am appending the target platform identifiers in order to make this deterministic and somewhat human readable. Are there better options?

Not really. The basic issue is the "natural" repo name (e.g. "foo") can't be used because there are multiple foo's, but the "correct" repo name for "foo" depends on what's in it, which is a function of version, platform-specificness, glibc, c-api tags, etc. Basically how a wheel file name is formatted. But that is a very unstable name. The "obvious" name, e.g. "foo_linux_x86_64" is somewhat context dependent (maybe you have foo just for linux, while I have 2 foos (one for musl, one for glibc, or w/e)).

(Recall we have the same issue with the name of the runtime repos, too -- they only carry major.minor and os/arch, for example)

For practical purposes, I think using the platform identifiers is a happy medium. Enough to distinguish, avoid conflicts, and relatively stable, but not overly verbose and unstable.

> 1. Does not require fetching/extracting the libraries whenever the annotations change.

Hmm. The annotations affect the generated BUILD file, which is part of the repository phase. In order to avoid that we would need to...somehow generate the same BUILD file content to leave that repository untouched...but have the build-phase see something different. Maybe if the annotations logic is off-loaded to somewhere else, e.g. another repo? e.g. given
```
pip.parse(whl_mods=...)
```

What's generated is e.g.

```
# @foo_linux_x86_64/BUILD
load("@foo_linux_x86_64_whl_mods//:whl_mods.bzl", "whl_mods")
whl_library(whl_mods = whl_mods)
```

> 3. Be compatible with `cc_*` rules so that wheels providing shared libraries or header files can be used.

The two basic solutions I come up with here are:
1. Always generate a cc_library target
2. Have a way for users to apply modification to the spoke BUILD file and have a way for the hub to refer to those targets. I can think of a variety of ways to do this, but they all rely on the hub knowing what aliases it needs to create.

> 4. Ideally the usage should come naturally through the hub repo so that users don't need to depend on implementation details of the naming of the repository rules - having spent enough time in this space I think that we can only really promise the contents of the hub repo as a stable API.

Yes, I agree.

I think the main issue is having a way for spokes to tell the hub what targets are available so that the hub can create an alias to them. The two solutions that come to mind are:
1. If users are providing patches to affect the spoke, having an additional arg/setting to inform the hub to create an alias sounds reasonable.
3. For the particular case of e.g. cc_library, having every spoke generate a cc_library (or an empty target by the same name) sounds reasonable.

aignas:
Since this is a potentially breaking change, I am wondering if we should create a feature flag to enable this behaviour as opt-in, then we can enable by default and then in 1.0 it would be the only way to do it?

rickeylev:
> That is only needed if cc_library is unhappy with an empty file.

cc_library is OK with empty srcs

aignas:
OK, quick status update - once #2369 is merged, I will update this PR to be
backwards compatible (i.e. users will have to opt into the new behaviour on
`0.38.0`) and then merge this PR. Then we can cut a release. Once that is done,
we can have our `1.0` I think.

aignas:
@keith, could you please check this PR again? I think it should work for your usecase now.

keith:
Yes sorry I'm OOO but I will as soon as I'm back next week

aignas:
Since the behaviour is opt-in, I'll merge the PR now and we can make further fixes later, but I am still looking forward to feedback next week.

keith:
I think I could probably make whl_filegroup work, but it definitely breaks some use cases we're using today. For example today in my `additive_build_content_file` I do this:

```bzl
_LIBS = [
    lib
    for lib in glob(
        [
            "site-packages/torch/lib/*.so*",
            "site-packages/torch.libs/*.so*",
            "site-packages/torch/lib/*.dylib",
        ],
        allow_empty = True,
        exclude = [
            "site-packages/torch/lib/*libgomp*",  # Handled below
            # Intel only, seem to be unnecessary for our use case
            "**/libtorchbind_test*",
            "**/libbackend_with_compiler*",
            "**/libjitbackend_test*",
        ],
    )
]
```

And then I expand multiple targets based on the `_LIBS` list below using list comprehensions. I don't think this has a clear 1:1 since if I had these as a filegroup it would not be the same

keith:
seems like one option is this:

```
    extra_hub_aliases = {
        "torch": [
            "stock-torch",
            "stock-torch-headers",
            "stock-torch-api",
            "torchlib",
        ],
        "torchvision": ["torchvisionpy"],
    },
```

which pretty much gets me back to what i had before, is that one not recommended?

Issue #2326, opened 2024-10-22
bug: --incompatible_python_disallow_native_rules broken with 0.37.0


rickeylev:
# 🐞 bug report

### Affected Rule

py_binary, py_library, py_test

### Is this a regression?

Yes. Earlier versions didn't break if --incompatible_python_disallow_native_rules was specified

### Description

When specified, all core rules fail to build with the error

```
Error in fail: @//tests/base_rules/py_binary:test_py_runtime_info_provided_subject not allowed to use native.py_binary
```

or similar

## 🔬 Minimal Reproduction

Any build with --incompatible_python_disallow_native_rules set

rickeylev:
This is fixed at head and we'll release a patch release today/tomorrow. In the meantime, applying this patch will early return from the validation logic and work around the issue: https://github.com/bazelbuild/rules_python/commit/1fc0495b87b28b243e9aa3d89fe645f5f28071ba.diff

Issue #1645, opened 2023-12-21
Turn on incompatible_python_disallow_native_rules for end-to-end tests


alexeagle:
I'd like to enable that option in a repo downstream of rules_python, but it seems like there's a violation shipped with the project:

```
ERROR: /private/var/tmp/_bazel_alexeagle/3eecd93732ac7c6f4f360a54a468da50/external/rules_python~0.27.1~python~python_3_9_aarch64-apple-darwin/BUILD.bazel:63:11: in py_runtime rule @@rules_python~0.27.1~python~python_3_9_aarch64-apple-darwin//:py3_runtime: 
Traceback (most recent call last):
        File "/virtual_builtins_bzl/common/python/py_runtime_rule.bzl", line 24, column 25, in _py_runtime_impl
        File "/virtual_builtins_bzl/common/python/common.bzl", line 489, column 28, in check_native_allowed
Error in Label: invalid label in Label(): invalid package name '__EXTERNAL_REPOS__/rules_python~0.27.1~python~python_3_9_aarch64-apple-darwin/': package names may not end with '/' (perhaps you meant ":"?)
```

rickeylev:
This is, unfortunately, a bug in Bazel that I missed before release. It's fixed at head and planned for the 7.1 release in Feb.

This issue is the pkg portion of the label is empty. If the label was //something:py3_runtime, then it'd be OK. Hm, actually, maybe we could just create a sub-dir and stick an alias in there and things would be happy enough.

https://github.com/bazelbuild/bazel/issues/17773#issuecomment-1852313089

PR #2327, opened 2024-10-22
fix(rules): remove rules_python --incompatible_python_disallow_native_rules checking

rickeylev:
When --incompatible_python_disallow_native_rules is enabled, all the core rules fail with
an error that rules_python should be used. This is incorrect, since the rules_python rules
are being used. What's happening is https://github.com/bazelbuild/rules_python/pull/2257
removed the magic migration tag when pystar is enabled, but the code to check the tag
was present wasn't removed. This went unnoticed because our CI doesn't set the migration
flag.

To fix, remove the validation logic entirely. If we're in the rules_python implementation,
then there is not need to perform this validation. It was just something copy/pasted from
the original code from Bazel itself.

Also update the bazelrc to always set --incompatible_python_disallow_native_rules.

Fixes https://github.com/bazelbuild/rules_python/issues/2326
Fixes https://github.com/bazelbuild/rules_python/issues/1645


aignas:
Should we also close #1645 since we would be testing this?

aignas (approved):

aignas:
This also needs a changelog item.

rickeylev:
> also fix 1645

Hrm. Bazel 6 doesn't have the flag, so setting it in the bazelrc files is problematic. I think what I'll have to do is set it as part of the CI config. I'll give that a try, but I might drop this from the PR if it ends up being a big pain.

> needs changelog entry

Done

rickeylev:
Figure out a way to get th bazelrcs updated. I changed CI to set `--config=bazel7.x` for Bazel 7 builds, and the bazelrcs have `common:bazel7.x` lines with the necessary flags.

PR #2329, opened 2024-10-22
docs: tell how to create a release branch for patch releases

rickeylev:
This adds some doc for how to create a branch when a patch release needs to be based on the
original release and not from head.


DEVELOPING.md
```diff
@@ -38,9 +40,31 @@ To find if there were any features added or incompatible changes made, review
...
+release tag and the patch changes cherry-picked into it.
+
+In this example, release `0.37.0` is being patched to create release `0.37.1`.
+The fix being included is commit `deadbeef`.
+
+1. `git checkout -b release/0.37 0.37.0`
+1. `git push upstream release/0.37`
+1. `git cherry-pick deadbeef`
+1. Fix merge conflicts, if any. If `MODULE.bazel.lock` conflicts occur, then
+   run `tools/private/update_bzlmod_lockfiles.sh`
```
aignas, line 56:
```suggestion
   run `pre-commit run update-bzlmod-lockfiles -a`
```

DEVELOPING.md
```diff
@@ -38,9 +40,31 @@ To find if there were any features added or incompatible changes made, review
...
+1. `git cherry-pick --continue` (if applicable)
+1. `git push upstream`
```
aignas, line 58: Should we create PRs for patch releases? when the patch release happens from head I think the current process works, but I found that it is sometimes good to do a PR against the `release/0.x` branch with all of the cherry-picks. The main cherry-pick conflict will be the changelog if the process is done later.

rickeylev: Yeah, I'd kinda prefer PRs. How does one do that? I couldn't figure it out.

rickeylev: I think another bonus of PRs is they show on the releases page in the auto-generated commit log.

aignas:
```
git cherry-pick -x <commit>
```

Then note the conflicts in the message.
Then the PR can be either a single commit or all of the commits in one with a summary as the PR description.

rickeylev: The -x docs just say it appends an extra line to the commit message, which is good info, so I've added that.

But how do we create a PR that says "add commit X in branch main to branch release/0.37"? I was fiddling with the "open pull request" settings and don't see how to do that.

aignas: Yeah, not sure we can automate that easily, but,
1. Each commit has a PR number in the commit message title when we merge them to `main`.
2. When you create a PR using `gh pr create` CLI command, you will have all commits as a bullet list.
3. I suspect these PRs would be cross referenced in GH UI.

rickeylev: What is `gh pr create` doing that's special? I don't use the gh cli because i'm always encountering issues with it (it seems to constantly have issues knowing where to push to or pull from).

I suspect that what you're doing is creating a fork in your personal repo, cherry-picking and committing to that, then opening a PR to merge those into `rules_python/<version`.

But I also know you often create branches directly in rules_python and not use a personal repo, which is what these instructions describe doing -- this is the part I don't see how to create a pull request for the cherry picks to merge into the release branch.

When you've sent PRs to merge into the release branches in the past, what flow did you use?

aignas: At some point I used to create branches in `bazelbuild/rules_python` but I stopped doing that, unless it is a release branch. If it is PRs to the release branch, I would create a branch on my personal fork. Ideally, I think we should avoid creating branches here unless we want them to be visible for future generations. Thulio had previously created a few branches on whl building in build actions that we could make use of and I wouldn't have known about them if they weren't here.

I think maybe `gh pr create` for me is just creating a description that is sometimes useful. If you configure the default upstream with `gh` then it should work better, but it is annoying that it is constantly asking you if you want to push to `upstream` or your personal fork. I am happy to leave out the `gh` CLI from our release flow.

aignas (approved):

DEVELOPING.md
rickeylev: > If it is PRs to the release branch, I would create a branch on my personal fork. 

Ahhh ok. I was hoping I was just missing some GitHub trick to create PRs of specific commits without going through a personal fork. Ah well.

PR #2331, opened 2024-10-22
docs: tell how to setup readthedocs integration with sphinxdocs

rickeylev:
This comes from setting up readthedocs integration with rules_testing.

It'd be nice if so much copy/paste wasn't necessary, but I'm not sure how to best
reduce that, so at the least, document what one has to do.


aignas (approved):
I might check what I've done for `pytest-bazel` later on and submit a PR against these docs.

I think having it here for now is better than not having it, hence approving. Thanks!

Issue #2310, opened 2024-10-17
[Bazel CI] Error toolchain_config_utils.bzl is invalid because 'cc/toolchains' is not a package.



sgowroji:
CI: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/4205#019298ab-9bbe-4ef2-b7ec-e35a9a8fd69f

 Platform: MacOS

 Logs: 
```
ERROR: Error computing the main repository mapping: at /private/var/tmp/_bazel_buildkite/9f47f6beb72f15412af12490e86ed0ba/external/bazel_tools/tools/cpp/cc_configure.bzl:18:6: Label '@@rules_cc//cc/toolchains:toolchain_config_utils.bzl' is invalid because 'cc/toolchains' is not a package; perhaps you meant to put the colon here: '@@rules_cc//cc:toolchains/toolchain_config_utils.bzl'?
``` 
 Culprit: https://github.com/bazelbuild/bazel/commit/1c4e78a9ceab9794140324d64833c628e382a3da

 Steps:
  ``` 
 git clone https://github.com/bazelbuild/rules_python
git reset 0e600586aea596a488902ce90938b36ce7e9f6d5  --hard
export USE_BAZEL_VERSION=1d022f403d26505d6edda73a50e08da53109dd3b
bazel build --noenable_bzlmode --enable_workspace  -- ... 
 ``` 
 CC Greenteam @meteorcloudy

PR #2332, opened 2024-10-22
tests: use rules_cc 0.0.13 for ignore_root_user_error test

rickeylev:
This is so that the test works with Bazel@head

Fixes https://github.com/bazelbuild/rules_python/issues/2310

### DIFF
--- a/tests/integration/ignore_root_user_error/WORKSPACE
+++ b/tests/integration/ignore_root_user_error/WORKSPACE
@@ -1,3 +1,12 @@
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+http_archive(
+    name = "rules_cc",
+    sha256 = "d9bdd3ec66b6871456ec9c965809f43a0901e692d754885e89293807762d3d80",
+    strip_prefix = "rules_cc-0.0.13",
+    urls = ["https://github.com/bazelbuild/rules_cc/releases/download/0.0.13/rules_cc-0.0.13.tar.gz"],
+)
+
 local_repository(
     name = "rules_python",
     path = "../../..",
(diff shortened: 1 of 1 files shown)


aignas (approved):

aignas:
Although, looking at the CI it seems that it is still unhappy.

rickeylev:
Heh, yeah, that required a few more fixes. All the integration tests needed rules_cc added, which then needed protobuf added. Then some allow_empty=True fixes in the local toolchain test.

tests/integration/ignore_root_user_error/WORKSPACE
```diff
@@ -1,3 +1,19 @@
...
+    strip_prefix = "protobuf-27.0",
+    url = "https://github.com/protocolbuffers/protobuf/releases/download/v27.0/protobuf-27.0.tar.gz",
+)
+
+http_archive(
+    name = "rules_cc",
+    sha256 = "d9bdd3ec66b6871456ec9c965809f43a0901e692d754885e89293807762d3d80",
+    strip_prefix = "rules_cc-0.0.13",
+    urls = ["https://github.com/bazelbuild/rules_cc/releases/download/0.0.13/rules_cc-0.0.13.tar.gz"],
+)
```
aignas, line 15: @rickeylev, I think this means that the `py_repositories` function adding these deps needs to be updated. This looks like the wrong thing to do if we want to keep supporting `bazel@HEAD` with a released `rules_python` version?

rickeylev: Good point. Something felt off updating so many workspace files for tests. I've started https://github.com/bazelbuild/rules_python/pull/2335 to update py_repositories.

PR #2333, opened 2024-10-23
chore: bump twine dependencies and cleanup for 1.0

aignas:
This change updates the `requirements.txt` for all of the platforms for
the `twine` component that is used in wheel publishing rules. Before that
the deps would be maintained by `dependabot.txt` and it seems that it would
constantly exclude the `requirements.txt` from the updated files, this means
that the `requirements.txt` (used on Linux) has gone out of sync and it would
be better to use a different mechanism to keep them in sync.

Hence the proposal is to:
- Use `uv` to sync the requirements.
- Add `requirements_linux.txt` to explicitly highlight that the file is for
  Linux.
- Keep `requirements.txt` but note that it is deprecated.
- Add `requirements_universal.txt` in case people want to use a single
  requirements file.
- Bump the dependencies to the latest versions and bring them back in sync.

Whilst at it I had to:
- support extra parameters in the locking format and allow setting visibility
- chore: add `rules_multirun` as a `dev_dependency`
- refactor: `uv pip compile` requirements for tools/publish
- chore: bump uv to 0.4.25


rickeylev:
> feat: support extra parameters in the locking format and allow setting visibility

This is just for the uv backed one, right? Which is still private. My point being: it shouldn't be in the changelog, since it's just a private thing we use for ourselves.

CI has some failures, I think its just missing WORKSPACE equivalent updates.

CONTRIBUTING.md
```diff
@@ -65,6 +65,15 @@ and setup. Subsequent runs will be faster, but there are many tests, and some of
 them are slow. If you're working on a particular area of code, you can run just
 the tests in those directories instead, which can speed up your edit-run cycle.
 
+## Updating tool dependencies
+
+It's suggested to routinely update the tool versions within our repo - some of the
+tools are using requirement files compiled by `uv` and others us other means. In order
```
rickeylev, line 71: typo: "use other"

rickeylev (approved):

CONTRIBUTING.md
aignas:
```suggestion
tools are using requirement files compiled by `uv` and others use other means. In order
```

PR #2334, opened 2024-10-23
fix: allow spaces in whl_librarys

### LINKED ISSUE #617: Heads up: `setuptools` contain files with whitespaces in their names
Hello, I would like to thank the authors, maintainers and all contributors for making such a cool project 😄 

The reason why I was brought here is that recently we received a issue in the setuptools repository (https://github.com/pypa/setuptools/issues/3072) that seems related to the fact that `bazel` does not support whitespaces in file names, and this might generate problems for people using `rules_python`.

As noted [in of the issue's comment](https://github.com/pypa/setuptools/issues/3072#issuecomment-1028474077), even before the recent changes, `setuptools` already had files containing whitespaces.

My main objective here is just to give you guys a heads up, so you are aware about potential problems that might arise. (I understand that it might be a very complicated problem to solve and that it will take a long time...)

For the time being `setuptools`' lead developer have kindly provided a patch with a workaround, but it is difficult to guarantee that in the future other files containing whitespace in their names will not be added (specially via vendored dependencies).

mark-thm:
Fixes #617.

Modern `setuptools` versions contain files critical to setuptools functionality loaded proactively on module load that contain spaces. Bazel 7.4.0+ now supports files with spaces in their names.


CHANGELOG.md
```diff
@@ -31,7 +31,7 @@ A brief description of the categories of changes:
 
 {#v0-0-0-fixed}
 ### Fixed
-- Nothing yet
+- #617: allow spaces in filenames included in `whl_library`s
```
rickeylev, line 34:
```suggestion
* (pypi) Allow spaces in filenames included in `whl_library`s
  ([616](https://github.com/bazelbuild/rules_python/issues/617))
```

python/private/glob_excludes.bzl
```diff
@@ -0,0 +1,34 @@
...
+def _version_dependent_exclusions():
+    """Returns glob exclusions that are sensitive to Bazel version.
+
+    Bazel 7.4.0+ added support for files with spaces. Prior versions of Bazel
+    do not support files with spaces.
+
+    Returns:
+        a list of glob exclusion patterns
+    """
+    major, minor, _ = native.bazel_version.split(".")
```
rickeylev, line 26: native.bazel_version is available during the repo phase, but not build phase (annoying, I know).

You should be able to use `@bazel_features` to do the version detection, though.

rickeylev (changes requested):

python/private/glob_excludes.bzl
mark-thm: Yeah, realized that after I pushed this: https://github.com/bazel-contrib/bazel_features/pull/79

aignas:
FYI, it would be better to wait for #2347 so that testing here is easier.

python/private/py_repositories.bzl
```diff
@@ -65,4 +65,10 @@ def py_repositories():
         strip_prefix = "protobuf-27.0",
         url = "https://github.com/protocolbuffers/protobuf/releases/download/v27.0/protobuf-27.0.tar.gz",
     )
+    http_archive(
+        name = "bazel_features",
+        sha256 = "c2596994cf63513bd44180411a4ac3ae95d32bf59148fcb6087a4642b3ffef11",
+        strip_prefix = "bazel_features-1.20.0",
+        url = "https://github.com/bazel-contrib/bazel_features/releases/download/v1.20.0/bazel_features-v1.20.0.tar.gz",
+    )
```
mark-thm, line 73: I need to also run
```
load("@bazel_features//:deps.bzl", "bazel_features_deps")
bazel_features_deps()
```

but it's no longer possible to do an inline load call, what's the right way to accomplish this?

aignas: #2274 also faced the same issue and we decided to backport the detection to our internal configuration repository rule. Would it be possible to do something similar in here too?

Otherwise, see what was proposed in there - adding an extra function `py_repositories_deps` that each user would have to call.

Another solution would be to only support the spaces in `bzlmod`, but that is not ideal either.

python/glob_excludes.bzl
```diff
@@ -0,0 +1,19 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"Utilities for glob exclusions."
+
+load("//python/private:glob_excludes.bzl", _glob_excludes = "glob_excludes")
+
+glob_excludes = _glob_excludes
```
aignas, line 19: I don't think we need to expose this to public?

aignas (commented):
Please consider rebasing`

rickeylev:
Here's an easy way to detect 7.4: `hasattr(native, "legacy_globals")`

rickeylev:
To clarify: put that in `python/private/utils.bzl`, near the bottom you'll see various `IS_BAZEL_BLABLA` constants

python/private/py_repositories.bzl
rickeylev: I re-implemented this to avoid the need for bazel_features.

Given that we're stuck with workspace for another year, we probably want to introduce a `py_repositories_deps()` call for workspace, but we can punt that to a separate PR.

rickeylev (commented):
I pushed some changes to address review comments and avoid the issue with workspace+bazel_features

python/private/pypi/deps.bzl
```diff
@@ -115,7 +116,7 @@ py_library(
         "**/*.dist-info/RECORD",
         "BUILD",
         "WORKSPACE",
-    ]),
+    ] + glob_excludes.version_dependent_exclusions()),
```
aignas, line 119: What about the `whl_library_targets` macro? Should we include the excludes in there too?

mark-thm: Yes, we should

rickeylev (approved):

python/private/pypi/deps.bzl
mark-thm: done

python/private/BUILD.bazel
```diff
@@ -179,6 +179,11 @@ bzl_library(
     srcs = ["full_version.bzl"],
 )
 
+bzl_library(
+    name = "glob_excludes_bzl",
+    srcs = ["glob_excludes.bzl"],
+)
```
mark-thm, line 185: @rickeylev does this need to express a dependency on `:util.bzl`?

rickeylev: Yes, that would be most correct. Fixed.

aignas (commented):
LGTM

Issue #1689, opened 2024-01-12
Make it easier to load wheels from the local file system


georgevreilly-stripe:
# 🚀 feature request

### Relevant Rules

Possibly in `whl_library`, but more likely in `requirements.txt`

### Description

I had created a patched wheel that I wished to test with Bazel and rules_python *before* publishing it to our internal Artifactory package index. There was no easy way to accomplish this.

I eventually worked out that I could test my patched wheel by using [pip2pi](https://pypi.org/project/pip2pi/) to create a local package index on my filesystem, then adding `--index-url=/tmp/pypi/simple` to the beginning of my `requirements.txt`.

See [this writeup](https://github.com/georgevreilly/torch21?tab=readme-ov-file#using-the-patched-wheel-with-rules_python) for full details.

### Describe the solution you'd like

 If you have a solution in mind, please describe it.

I think the easiest thing for users is if they can specify a file path in `requirements.txt`; e.g.,

```
sympy==1.12
./tmp/torch-2.1.0+stripe.5-cp38-cp38-manylinux1_x86_64.whl
triton==2.1.0
```

Pip allows this, after all.

### Describe alternatives you've considered

 Have you considered any alternative solutions or workarounds?

PR #2345, opened 2024-10-25
type(feat): Allow parsing of wheel file paths in requirements.in 

stridge-cruxml:
I have a `.whl` file inside my bazel repo. I would like to add this to `requirements.in` like

```
path/to/my.whl
```

Adding `extra_data` to `pip_compile` and passing this through is sufficient.
eg
```
compile_pip_requirements(
  ...
  extra_data = ["//path/to:my.whl"],
)
```

An issue I came across is pip tools puts the absolute path to the wheel in `requirements.txt` which isn't great.
So I change the path back to a relative one in `dependency_resolver.py`

### DIFF
--- a/python/private/pypi/dependency_resolver/dependency_resolver.py
+++ b/python/private/pypi/dependency_resolver/dependency_resolver.py
@@ -20,6 +20,7 @@ import shutil
 import sys
 from pathlib import Path
 from typing import Optional, Tuple
+import multiprocessing
 
 import click
 import piptools.writer as piptools_writer
@@ -97,6 +98,7 @@ def main(
     extra_args: Tuple[str, ...],
 ) -> None:
     bazel_runfiles = runfiles.Create()
+    runfiles_dir = Path(bazel_runfiles._python_runfiles_root) / "_main"
 
     requirements_file = _select_golden_requirements_file(
         requirements_txt=requirements_txt,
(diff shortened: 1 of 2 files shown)


python/private/pypi/requirements.bzl.tmpl.workspace
```diff
@@ -61,12 +61,23 @@ def install_deps(**whl_library_kwargs):
...
-            name = name,
-            requirement = requirement,
-            group_name = group_name,
-            group_deps = group_deps,
-            annotation = _get_annotation(requirement),
-            **whl_config
-        )
+        if "@ file://" in requirement:
+          whl_file = whl_files[name]
+          whl_library(
```
aignas, line 66: nit: the indentation needs to be fixed here.

aignas (commented):
Yeah I think this is as good as we can get here.

If you would like to also add the wiring for `bzlmod` (e.g. wire `pip.override` to the `whl_library` `whl_name` attribute), then I'd be happy to merge this approach. After all we are planning to release `1.0` at some point soon and `bzlmod` is a first class citizen.

python/private/pypi/requirements.bzl.tmpl.workspace
stridge-cruxml: Done.

stridge-cruxml:
I fixed the indentation.

I am not going to dedicate more time to this for a while due to other priorities. Existing code I have all uses WORKSPACE still. If there is someone else who would be willing to implement the bzlmod side of things they are most welcome.

aignas:
#1689 is the ticket that this PR would fix.

PR #2347, opened 2024-10-25
refactor(pypi): use a macro to define whl_library targets

aignas:
Summary:
- refactor: Start implementing whl_library_targets
- refactor: start using whl_library_targets macro
- refactor: generate config settings in the new macro
- refactor: copy_files in the new macro
- refactor: move entry_point generation to the macro
- refactor: move the py_library and whl generation to the new macro

This makes the code more maintainable by reducing the amount of tests
that are comparing BUILD.bazel outputs.


python/private/pypi/whl_library_targets.bzl
```diff
@@ -0,0 +1,339 @@
...
+        name,
+        *,
+        whl_name,
+        dep_template,
+        data_exclude = [],
+        srcs_exclude = [],
+        tags = [],
+        filegroups = {
+            DIST_INFO_LABEL: ["site-packages/*.dist-info/**"],
+            DATA_LABEL: ["data/**"],
```
aignas, line 42: FYI, I would like to include extra folders here, what me and @arrdem discussed in #1730. It would be valuable to have those files so that the `hub` repo can provide more `filegroup` targets out of the box. This may make things easier where the `annotations` API is heavily used by our users.

cc: @keith, not sure if this help in defining the `cc_library` targets, but regardless, this may be a good step forward.

python/private/pypi/whl_library_targets.bzl
```diff
@@ -0,0 +1,339 @@
...
+                deps_by_platform = dependencies_by_platform,
+                tmpl = dep_template.format(name = "{}", target = WHEEL_FILE_PUBLIC_LABEL),
+                select = getattr(native, "select", select),
+            ),
+            visibility = impl_vis,
+        )
+
+    if hasattr(rules, "py_library"):
+        _data_exclude = [
+            "**/* *",
```
rickeylev, line 221: I seem to recall a recent PR that included names with spaces? Is that being lost here?

python/private/pypi/whl_library_targets.bzl
```diff
@@ -0,0 +1,339 @@
...
+                exclude = _data_exclude,
+            ),
+            # This makes this directory a top-level in the python import
+            # search path for anything that depends on this.
+            imports = ["site-packages"],
+            deps = _deps(
+                deps = dependencies,
+                deps_by_platform = dependencies_by_platform,
+                tmpl = dep_template.format(name = "{}", target = PY_LIBRARY_PUBLIC_LABEL),
+                select = getattr(native, "select", select),
```
rickeylev, line 254: This is a weird getattr() ? When does native.select _not_ exist??

python/private/pypi/generate_whl_library_build_bazel.bzl
```diff
@@ -14,406 +14,64 @@
...
-          from generated deps lists so as to avoid direct cycles. These dependencies will be provided
-          at runtime by the group rules which wrap this library and its fellows together.
+        **kwargs: Extra args serialized to be passed to the
+            {obj}`whl_library_targets`.
 
     Returns:
         A complete BUILD file as a string
     """
 
+    kwargs["name"] = "unused"
```
rickeylev, line 47: Instead of setting name=unused, how about setting it to the name of the package being processed?

python/private/pypi/generate_whl_library_build_bazel.bzl
```diff
@@ -14,406 +14,64 @@
...
-                whl_file_label = whl_file_label,
-                tags = repr(tags),
-                data_label = DATA_LABEL,
-                dist_info_label = DIST_INFO_LABEL,
-                entry_point_prefix = WHEEL_ENTRY_POINT_PREFIX,
-                srcs_exclude = repr(srcs_exclude),
-                data = repr(data),
-                impl_vis = repr(impl_vis),
-            ),
+            """load("@rules_python//python/private/pypi:whl_library_targets.bzl", "whl_library_targets")""",
```
rickeylev, line 60: Since we've basically gutted this file, how about moving this to be a template string, similar to how we've switched to doing elsewhere?

```
_BUILD_TEMPLATE = """
load("@rules_python//python/private/pypi:whl_library_targets.bzl", "whl_library_targets")

whl_library_targets(
{kwargs}
)
"""

_BUILD_TEMPLATE.format(render.kwargs(kwargs))

```

(render.kwargs just popped into my head, a `"\n".join([...])` also SGTM)

python/private/pypi/whl_library_targets.bzl
```diff
@@ -0,0 +1,339 @@
...
+    "PY_LIBRARY_IMPL_LABEL",
+    "PY_LIBRARY_PUBLIC_LABEL",
+    "WHEEL_ENTRY_POINT_PREFIX",
+    "WHEEL_FILE_IMPL_LABEL",
+    "WHEEL_FILE_PUBLIC_LABEL",
+)
+
+def whl_library_targets(
+        name,
+        *,
```
rickeylev, line 34:
```suggestion
        *,
        name,
```

python/private/pypi/whl_library_targets.bzl
```diff
@@ -0,0 +1,339 @@
...
+    "DATA_LABEL",
+    "DIST_INFO_LABEL",
+    "PY_LIBRARY_IMPL_LABEL",
+    "PY_LIBRARY_PUBLIC_LABEL",
+    "WHEEL_ENTRY_POINT_PREFIX",
+    "WHEEL_FILE_IMPL_LABEL",
+    "WHEEL_FILE_PUBLIC_LABEL",
+)
+
+def whl_library_targets(
```
rickeylev, line 32: Oof, what a signature. It is what it is, though. Really kind of helps make clear how complex generating these BUILD files for pypi stuff is and has become, though.

rickeylev (approved):

python/private/pypi/whl_library_targets.bzl
aignas: This is still not merged - #2334.

python/private/pypi/whl_library_targets.bzl
aignas:
```suggestion
                # Actually, `select` is not part of `native`, but in order to support
                # bazel 6.4 in unit tests, I have to somehow pass the `select` 
                # implementation in the unit tests and I chose this to be routed through
                # the `native` struct. So, tests` will be successful in `getattr` and
                # the real code will use the fallback provided here.
                select = getattr(native, "select", select),
```

python/private/pypi/generate_whl_library_build_bazel.bzl
aignas: Decided to use the whl basename instead as the name, because it allows us to be more flexible in the future - like setup target_compatible_with based on the filename.

PR #2351, opened 2024-10-27
fix(bzlmod): generate config_setting values for all python toolchains (#2350)

aignas:
fix(bzlmod): generate config_setting values for all python toolchains (#2350)

#2253 broke how the config settings are generated and only generated the
config setting values for the python version values that we would have
the registered toolchains for. This PR restores the previous behaviour.
However, if the root module uses `python.override` to remove the allowed
toolchains, then `config_settings` will be also affected.

(cherry picked from commit 9340a8199cac0912305dad427a5cdc72ee4fc7ff)

Conflicts:
- CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,10 @@ A brief description of the categories of changes:
 * Particular sub-systems are identified using parentheses, e.g. `(bzlmod)` or
   `(docs)`.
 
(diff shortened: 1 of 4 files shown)


aignas:
The cherry-pick with the PR method is a little clumsy, I will just push to the release branch instead. Closing this.

PR #2357, opened 2024-10-28
feat: add features.version value

rickeylev:
This adds a `features.version` value that is populated by git when `git archive` is run. This means it will be expanded when our release action runs to the tag name. Otherwise, it isn't expanded. When it isn't expanded, the value is set to the empty string.

### DIFF
--- a/python/features.bzl
+++ b/python/features.bzl
@@ -13,6 +13,9 @@
 # limitations under the License.
 """Allows detecting of rules_python features that aren't easily detected."""
 
+_VERSION_PRIVATE = "$Format:%(describe:tags=true)$"
+
 features = struct(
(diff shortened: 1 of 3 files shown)


python/features.bzl
```diff
@@ -13,6 +13,9 @@
 # limitations under the License.
 """Allows detecting of rules_python features that aren't easily detected."""
 
+_VERSION_PRIVATE = "$Format:%(describe:tags=true)$"
```
aignas, line 16: Could you please point to `git archive` manual in here. It is very much unclear how this gets populated in reality. Pointing to `.gitattributes` in the comment would be also good.

rickeylev: Done

aignas (approved):

PR #2365, opened 2024-11-01
fix(pypi): use local version specifiers for patched whl output

aignas:
Before this change the installation of the patched wheels using `uv` or
similar would break. This change fixes that by using local version
specifier, which is better than using a build tag when installing the
wheels.

Before the change:
```console
$ cd examples/bzlmod
$ bazel build @pip//requests:whl
$ uv pip install <path to requests wheel in the example>
error: The wheel filename "requests-2.25.1-patched-py2.py3-none-any.whl" has an invalid build tag: must start with a digit
```

After:
```
$ uv pip install <path to requests wheel in the example>
Resolved 5 packages in 288ms
Prepared 5 packages in 152ms
Installed 5 packages in 13ms
 + certifi==2024.8.30
 + chardet==4.0.0
 + idna==2.10
 + requests==2.25.1+patched (from file:///home/aignas/src/github/aignas/rules_python/examples/bzlmod/bazel-bzlmod/external/rules_python~~pip~pip_39_requests_py2_none_any_c210084e/requests-2.25.1+patched-py2.py3-none-any.whl)
 + urllib3==1.26.20
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,8 @@ A brief description of the categories of changes:
   `bzlmod` users would have observed.
 * (pypi) (Bazel 7.4+) Allow spaces in filenames included in `whl_library`s
   ([617](https://github.com/bazelbuild/rules_python/issues/617)).
+* (pypi) The patched wheel filenames from now on are using local version specifiers
+  which fixes usage of the said wheels using standard package managers.
 
 {#v0-0-0-added}
 ### Added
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -47,6 +47,8 @@ A brief description of the categories of changes:
   ([617](https://github.com/bazelbuild/rules_python/issues/617)).
 * (pypi) When {attr}`pip.parse.experimental_index_url` is set, we need to still
   pass the `extra_pip_args` value when building an `sdist`.
+* (pypi) The patched wheel filenames from now on are using local version specifiers
```
groodt, line 50:
```suggestion
* (pypi) The patched wheel filenames from now on are using local version identifiers
```

Alternative suggestion for full CHANGELOG entry:
`(pypi) Patched wheel filenames use local version identifiers to enable installation with standards-compliant package managers`

groodt (approved):
LGTM

PR #2366, opened 2024-11-01
change(pypi): use the python package version instead of sha in repo names

aignas:
Initially the repo name generate was very conservative and we had sha256
as the endings of the repo names. With this change we have more human
readable repository names. In order to still comply with the bazel label
rules, we are normalizing the versions to only have `_` and `.`
characters.

This is an alternative fix to the error message seen in
https://github.com/bazelbuild/rules_python/pull/1837#issuecomment-2143270349


aignas:
Having dots in the filenames is preblematic and the current scheme is good enough.

PR #2368, opened 2024-11-02
fix(pypi): correctly pass `extra_pip_args` when building sdists

aignas:
Before this change the `extra_pip_args` would not be passed down the
chain if `experimental_index_url` is set. This adds a test and fixes the
bug.

Work towards #260


groodt (approved):
LGTM

PR #2372, opened 2024-11-05
feat(toolchain): support freethreaded toolchains

aignas:
Before this PR freethreaded toolchains were not possible to be used,
this adds the minimum plumbing to get the things working. Coverage
support is also added.

Whilst at it:
- Add plumbing to print checksums only for a particular python version.
- Bump the remaining toolchain versions that used to use the 20241008 release
- Pass around the loaded platform list so that we are only defining toolchains for the platforms that we have loaded the hermetic toolchain for.

Tested:
```
$ bazel run --//python/config_settings:python_version=3.13.0 --//python/config_settings:py_freethreaded="yes" //python/private:current_interpreter_executable
...
Python 3.13.0 experimental free-threading build (main, Oct 16 2024, 03:26:14) [Clang 18.1.8 ] on linux
>>>
```

Closes #2129.
Work towards #2386.


vfdev-5:
Seems like this PR is working to build XLA with Python free-threading using the following command:
```bash
bazel build \
    --@rules_python//python/config_settings:py_freethreaded="yes" \
    --define=use_enabled_free_threading=true \
    --test_output=all --spawn_strategy=sandboxed \
    --override_repository=rules_python=/rules_python \
    //xla/...
```
It downloads however two versions of python:
```
ls /root/.cache/bazel/_bazel_root/83d0ab0b1a11dcb5adba61b76e88291b/external/python                                                                                                         
...
python_x86_64-unknown-linux-gnu-freethreaded/                                                                            
python_x86_64-unknown-linux-gnu/
...
```
I assume that free-threaded version is used for building python extensions...

CHANGELOG.md
```diff
@@ -60,9 +68,13 @@ A brief description of the categories of changes:
   and one extra file `requirements_universal.txt` if you prefer a single file.
   The `requirements.txt` file may be removed in the future.
 * The rules_python version is now reported in `//python/features.bzl#features.version`
+<<<<<<< HEAD
```
vfdev-5, line 71: this should be fixed

aignas:
@vfdev-5, I am not fully sure about why you have the toolchain in your cache. Could you please check it with `bazel query` or use `--toolchain_resolution_debug=python` to check?

vfdev-5:
In my tests I deleted completely the cache and run xla build from the very beginning. In the console I saw it downloading python for platform `x86_64-unknown-linux-gnu` and after that `x86_64-unknown-linux-gnu-freethreaded`.
I'll retry with your suggestions to figure out

python/config_settings/BUILD.bazel
```diff
@@ -92,6 +92,24 @@ string_flag(
     visibility = ["//visibility:public"],
 )
 
+string_flag(
+    name = "py_freethreaded",
```
aignas, line 96: This is similar to `py_linux_libc`, but we could have other names as well. Suggestions?

python/config_settings/BUILD.bazel
```diff
@@ -92,6 +92,24 @@ string_flag(
...
+    build_setting_default = "no",
+    values = [
+        "yes",
+        "no",
+    ],
```
aignas, line 101: TODO: use an enum here

python/private/hermetic_runtime_repo_setup.bzl
```diff
@@ -45,11 +45,12 @@ def define_hermetic_runtime_toolchain_impl(
         python_version: {type}`str` The Python version, in `major.minor.micro`
             format.
         python_bin: {type}`str` The path to the Python binary within the
-            repositoroy.
+            repository.
         coverage_tool: {type}`str` optional target to the coverage tool to
             use.
     """
     _ = name  # @unused
+    is_freethreaded = Label("//python/config_settings:is_py_freethreaded")
```
aignas, line 53: nit: this can be a constant at the top of the file.

python/private/hermetic_runtime_repo_setup.bzl
```diff
@@ -65,21 +66,27 @@ def define_hermetic_runtime_toolchain_impl(
             # Platform-agnostic filegroup can't match on all patterns.
             allow_empty = True,
             exclude = [
-                # Unused shared libraries. `python` executable and the `:libpython` target
```
aignas, line 68: TODO: remove the resorting here.

python/private/hermetic_runtime_repo_setup.bzl
```diff
@@ -96,14 +103,65 @@ def define_hermetic_runtime_toolchain_impl(
...
+        ],
+        visibility = ["//visibility:private"],
+    )
+    native.config_setting(
+        name = "is_freethreaded_windows",
+        flag_values = {
+            Label("//python/config_settings:py_freethreaded"): "yes",
+        },
+        constraint_values = [
+            "@platforms//os:windows",
```
aignas, line 145: Discuss: should this go to somewhere outside the macro definition? Maybe we could have `//python/private/config_settings:BUILD.bazel` where such things can go? The downside is that the visibility will not be `private` anymore.

python/private/pythons_hub.bzl
```diff
@@ -149,6 +152,9 @@ This rule also writes out the various toolchains for the different Python versio
             doc = "Default Python version for the build in `X.Y` or `X.Y.Z` format.",
             mandatory = True,
         ),
+        "loaded_platforms": attr.string_list_dict(
+            doc = "The mapping from platform to version",
```
aignas, line 156:
```suggestion
            doc = "The list of loaded platforms keyed by the toolchain full python version",
```

python/private/toolchains_repo.bzl
```diff
@@ -78,6 +81,7 @@ py_toolchain_suite(
             python_version = python_version,
         )
         for platform, meta in PLATFORMS.items()
+        if platform in loaded_platforms
```
aignas, line 84: TODO: maybe `platforms` should only contain the keys that are needed here. We should accept a parameter instead of using the global variable `PLATFORMS`.

python/private/toolchains_repo.bzl
```diff
@@ -390,6 +396,9 @@ def _get_host_platform(os_name, arch):
     """
     host_platform = None
     for platform, meta in PLATFORMS.items():
+        if "freethreaded" in platform:
```
aignas, line 399: `"freethreaded"` should be a constant in the `//python:versions.bzl`

python/versions.bzl
```diff
@@ -586,121 +613,135 @@ MINOR_MAPPING = {
 }
 
 PLATFORMS = {
```
aignas, line 615: This should be returned by a function instead of doing list comprehensions.

python/versions.bzl
```diff
@@ -548,28 +548,55 @@ TOOL_VERSIONS = {
...
+            }.items()
+            for suffix, build in {
+                "": platform + "-{build}.tar.gz",
+                "-freethreaded": "".join([
+                    platform,
+                    "-shared" if "windows" in platform else "",
+                    "-freethreaded+" + opt,
+                    "-full.tar.zst",
+                ]),
+            }.items()
```
aignas, line 583: The logic of which build to choose when we have a freethreaded variant should be inside a function.

aignas (commented):
Done with the self-review.

groodt (approved):
LGTM

CHANGELOG.md
```diff
@@ -80,6 +88,7 @@ A brief description of the categories of changes:
   and use the `pip` backend to do the downloading. This was only available for
   users setting {bzl:obj}`pip.parse.experimental_index_url`, but now users have
   more options whilst we continue to work on stabilizing the experimental feature.
+* (toolchain) The support for freethreaded Python toolchains is now available.
```
rickeylev, line 91: small typo
```suggestion
* (toolchain) Support for freethreaded Python toolchains is now available. 
```

Should this also mention how to activate it, or point to docs on how to do so?

python/config_settings/BUILD.bazel
rickeylev: I think py_freethreaded SGTM

python/private/hermetic_runtime_repo_setup.bzl
```diff
@@ -67,19 +69,23 @@ def define_hermetic_runtime_toolchain_impl(
             exclude = [
                 # Unused shared libraries. `python` executable and the `:libpython` target
                 # depend on `libpython{python_version}.so.1.0`.
-                "lib/libpython{major}.{minor}.so".format(**version_dict),
+                "lib/libpython{major}.{minor}*.so".format(**version_dict),
```
rickeylev, line 72: Just checking/curious: I'm guessing the `libX.Y*` pattern works because the extracted standalone interpreters only contain appropriately relevant .so files?

python/private/hermetic_runtime_repo_setup.bzl
```diff
@@ -96,14 +102,65 @@ def define_hermetic_runtime_toolchain_impl(
...
         includes = [
             "include",
-            "include/python{major}.{minor}".format(**version_dict),
-            "include/python{major}.{minor}m".format(**version_dict),
+        ] + select({
+            _IS_FREETHREADED: [
+                "include/python{major}.{minor}t".format(**version_dict),
+                # FIXME @aignas 2024-11-05: the following looks fishy - should
+                # we have a config setting for `m` (i.e. DEBUG builds)?
+                "include/python{major}.{minor}tm".format(**version_dict),
```
rickeylev, line 110: I think the m-suffix was actually removed, but I can't find the docs that talk about it. I swear I read about this when I was looking into wheel format names and such.

But yes, "tm" is fishy.

python/private/python_register_toolchains.bzl
```diff
@@ -73,6 +73,9 @@ def python_register_toolchains(
         minor_mapping: {type}`dict[str, str]` contains a mapping from `X.Y` to `X.Y.Z`
             version.
         **kwargs: passed to each {obj}`python_repository` call.
+
+    Returns:
+        on bzlmod this returns the loaded platform labels.
```
rickeylev, line 78:
```suggestion
        On bzlmod this returns the loaded platform labels. Otherwise None.
```

rickeylev (approved):
So if I remember the toolchain generation logic, then all one has to do is set the flag, and it'll trigger downloading and usage of the free-threaded interpreter?

Otherwise just some small comments, LGTM

python/private/hermetic_runtime_repo_setup.bzl
rickeylev: I think its fine to locally define some config settings like this. These are fairly specific and not particularly complicated.

python/private/hermetic_runtime_repo_setup.bzl
aignas:
```suggestion
```

python/private/hermetic_runtime_repo_setup.bzl
aignas: This is for exclusion, that is why `libX.Y*.so` is just as good as `libX.Y.so`.

PR #2377, opened 2024-11-06
fix(pypi): handle multiple versions of the same package when parsing requirements files

### LINKED ISSUE #2337: pip.parse doesn't fetch anything with conflicting markers
If I have a `requirements.in` like this:

```
--extra-index-url https://download.pytorch.org/whl/cpu
torch==2.4.1; platform_machine != "x86_64"
torch==2.4.1+cpu; platform_machine == "x86_64"
```

Which I use to generate a universal requirements.txt with uv using: `uv pip compile --universal requirements.in -o requirements_lock.txt --emit-index-url --no-strip-markers`, I end up with multiple entries for torch in that lockfile:

```
SNIP
torch==2.4.1+cpu ; platform_machine == 'x86_64'
    # via -r requirements.in
torch==2.4.1 ; platform_machine != 'x86_64'
    # via -r requirements.in
SNIP
```

When I attempt to pull those in with:

```bzl
pip.parse(
    hub_name = "pip",
    python_version = "3.11",
)
use_repo(pip, "pip")
```

and use them with:

```bzl

alias(
    name = "foo",
    actual = requirement("torch"),
)
```

It fails with:

```
ERROR: no such package '@@rules_python~~pip~pip//torch': BUILD file not found in directory 'torch' of external repository @@rules_python~~pip~pip. Add a BUILD file to a directory to mark it as a package.
ERROR: /tmp/pyrepro/BUILD.bazel:3:6: no such package '@@rules_python~~pip~pip//torch': BUILD file not found in directory 'torch' of external repository @@rules_python~~pip~pip. Add a BUILD file to a directory to mark it as a package. and referenced by '//:foo'
```

If I remove the non-matching entry from the lockfile it works, so I assume it's the nature of having 2 conflicting requirements even though they are disambiguated by the marker.

To repro in this project: [repro.zip](https://github.com/user-attachments/files/17499528/repro.zip)

Run `bazel fetch ...`

ewianda:
This change makes it possible to handle local versions of packages,
which is extremely useful with PyTorch. With this change, it is
possible to have different local versions of the same package in the
`requirements.txt` file translated to valid `whl_library` repositories.

Fixes #2337


examples/bzlmod/requirements_lock_3_9.txt
```diff
@@ -1,7 +1,7 @@
 # This file was autogenerated by uv via the following command:
-#    bazel run //examples:bzlmod_requirements_3_9.update
+#    uv pip compile requirements.in --universal --generate-hashes --no-strip-extras --emit-index-url --index-strategy unsafe-first-match -p 3.9 -o requirements_lock_3_9.txt
```
ewianda, line 2: Not sure of a better to demonstrate/test that this feature works

python/private/pypi/parse_requirements.bzl
```diff
@@ -92,7 +92,7 @@ def parse_requirements(
         # are returned as just the base package name. e.g., `foo[bar]` results
         # in an entry like `("foo", "foo[bar] == 1.0 ...")`.
         requirements_dict = {
-            normalize_name(entry[0]): entry
+            entry[1]: entry
```
aignas, line 95: `parse_requirements` and `parse_requirements_txt` have unit tests in `//tests/pypi/` It would be better to add a test there instead of an example.

python/private/pypi/parse_requirements.bzl
```diff
@@ -32,6 +32,19 @@ load(":index_sources.bzl", "index_sources")
...
+def _extract_version(entry):
+    """Extract the version part from the requirement string.
+
+
+    entry: The requirement string.
+    """
+    version_start = entry.find("==")
+    if version_start != -1:
+        # Extract everything after '==' until the next space or end of the string
+        version = entry[version_start + 2:].split(" ")[0]
```
aignas, line 44: nit: I prefer using `partition` in the code because it just creates three return values.
```suggestion
        version, _, _ = entry[version_start + 2:].partition(" ")
```

python/private/pypi/parse_requirements.bzl
```diff
@@ -32,6 +32,19 @@ load(":index_sources.bzl", "index_sources")
 load(":parse_requirements_txt.bzl", "parse_requirements_txt")
 load(":whl_target_platforms.bzl", "select_whls")
 
+def _extract_version(entry):
+    """Extract the version part from the requirement string.
+
+
+    entry: The requirement string.
```
aignas, line 39:
```suggestion
    Args:
        entry: {type}`str` The requirement string.
```

python/private/pypi/parse_requirements.bzl
```diff
@@ -92,7 +105,7 @@ def parse_requirements(
         # are returned as just the base package name. e.g., `foo[bar]` results
         # in an entry like `("foo", "foo[bar] == 1.0 ...")`.
         requirements_dict = {
-            normalize_name(entry[0]): entry
+            (entry[0], _extract_version(entry[1])): entry
```
aignas, line 108: You are dropping the `normalize_name` here, could we keep it here please?

aignas (commented):
Thanks for the contribution!

Does this fix the #2337 issue? Looking at the fix here this may be the case, but I am not sure.

aignas:
LGTM in general, could you please add a CHANGELOG.md note as well that this PR fixes an open issue? Something like:
> * (pypi): Correctly handle multiple versions of the same package in the requirements files which is useful when including different PyTorch builds (e.g. <pytorch+cpu> vs <pytorch CUDA>) for different target platforms. Fixes [#2337](https://github.com/bazelbuild/rules_python/issues/2337).

aignas (approved):

aignas (changes requested):
Ah, could you please move the changelog entry to the unreleased block? It might have messed up during a merge/rebase.

PR #2389, opened 2024-11-11
chore: change the parse_all_requirements_files default value

aignas:
With this PR we are changing the defaults in the upcoming `0.39` version, and
if all goes well, the release after that will be `1.0`.

Work towards #1361


groodt (approved):
LGTM

PR #2390, opened 2024-11-11
feat: add attrs/fields for the runtime's ABI flags

rickeylev:
This adds attributes and fields for storing the runtimes ABI flags value, i.e. `sys.abiflags`.

For freethreaded interpreters, the abi flags contain `t`, which is used creating e.g.
virtualenvs.

The attribute can be directly specified, or if it's not, will be computed based on the
`--py_freethreaded` flag.

### DIFF
--- a/python/private/py_runtime_rule.bzl
+++ b/python/private/py_runtime_rule.bzl
@@ -17,6 +17,7 @@ load("@bazel_skylib//lib:dicts.bzl", "dicts")
 load("@bazel_skylib//lib:paths.bzl", "paths")
 load("@bazel_skylib//rules:common_settings.bzl", "BuildSettingInfo")
 load(":attributes.bzl", "NATIVE_RULES_ALLOWLIST_ATTRS")
+load(":flags.bzl", "FreeThreadedFlag")
 load(":py_internal.bzl", "py_internal")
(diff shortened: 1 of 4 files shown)


aignas (approved):

python/private/py_runtime_rule.bzl
```diff
@@ -101,6 +102,13 @@ def _py_runtime_impl(ctx):
             interpreter_version_info["minor"],
         )
 
+    abi_flags = ctx.attr.abi_flags
```
aignas, line 105: What about doing
```
diff --git a/python/private/hermetic_runtime_repo_setup.bzl b/python/private/hermetic_runtime_repo_setup.bzl
index 3f7bb5d7..207d021b 100644
--- a/python/private/hermetic_runtime_repo_setup.bzl
+++ b/python/private/hermetic_runtime_repo_setup.bzl
@@ -198,6 +198,10 @@ def define_hermetic_runtime_toolchain_impl(
             _IS_FREETHREADED: "cpython-{major}{minor}t".format(**version_dict),
             "//conditions:default": "cpython-{major}{minor}".format(**version_dict),
         }),
+        abi_info = select({
+            _IS_FREETHREADED: "t",
+            "//conditions:default": "",
+        }),
     )

     py_runtime_pair(
```

Instead of depending on the `_py_freethreaded_flag` implementation?

Issue #2378, opened 2024-11-06
Support Bazel 8


rickeylev:
I was trying to run our test with Bazel 8 and have been encountering a variety of problems.

The overall impression I'm getting is that the the core code is working, but various dev dependencies aren't.

Debugging this has been rather tough. Usually this is just a matter of updating some dependencies, but I've yet to find a combination of dependencies that works with both Bazel 7 and Bazel 8.

Issues I've seen thus far:

* [x] WORKSPACE.bzlmod is non-empty. It needs to become empty as WORKSPACE files become no-ops
* [ ] gazelle has `native.sh_binary` references: https://github.com/bazel-contrib/bazel-gazelle/issues/1959
* [x] We need to update rules_bazel_integration_test version to fix some native references.
* [ ] rules_cc shenanigans -- it seems that between 0.0.9 and 0.0.14 there have been various changes to how to reference bzl files. This ends up breaking doc building and the bzl_library deps. What we'll probably want to do is stop using cc:defs.bzl and start using the more specific loads instead.
* [ ] Some issue with stardoc. Upgrading to 0.7.1 seems to fix it, _but_ the release notes for stardoc say 0.7 _requires_ Bazel 7. This will probably be OK; stardoc is a dev dependency, and only otherwise used via sphinxdocs, which is use-at-your-own-risk.

Various notes:
rules_cc centric:
* **rules_cc 0.0.9 can't be used with bazel 8: it lacks some fixes**
* **rules_cc 0.0.10 can't be used with bazel 6: it has a dep on stardoc 7, which requires (in its MODULE definition) bazel 7+**
* rules_cc 0.0.11, 0.0.12, and 0.0.13 can't be used with Bazel 6: it tries to use `extension_metadata(reproducible=True)`
* rules_cc 0.0.14 might be usable with Bazel 6 (mod graph passes) and relnotes say they restored bazel 6 compatibility
* rules_cc 0.0.14 with protobuf 24.4 doesn't work -- protobuf loads `@rules_cc//:defs.bzl#cc_proto_library`, which was removed in rules_cc 0.0.14`. This was fixed in protobuf 27.0.
* So basically, we must use rules_cc 0.0.14 and protobuf 27.0 or higher together.

stardoc centric:
* **Bazel 6 with bzlmod and stardoc 0.7 is a problem: Bazel will fail during module phase because stardoc requires (in its MODULE) Bazel 7.**
[...]

PR #2395, opened 2024-11-12
tests: use $(rootpaths) to get executable files paths for Bazel 8 compatibility

rickeylev:
In Bazel 8, the singular `$(rootpath)` expansions require that the target expands to a
single file. The py rules have an unfortunate legacy behavior where their default outputs
are the executable and the main py file, thus causing an error.

To fix, use the plural `$(rootpaths)`, then post-process the space-separated string to get
just the executable portion of it.

Along the way...
* Add tests/integration/py_cc_toolchain_registered/bazel-* symlink to Bazel ignore.
  This avoids an infinite symlink expansions error and performance issues when those
  symlinks exist.

Work towards https://github.com/bazelbuild/rules_python/issues/2378


aignas (approved):
The `MODULE.bazel.lock` file change looks weird - I don't see why it would have to change - is it the `.bazelignore` or are we missing some CI checks...

PR #2397, opened 2024-11-12
chore: remove Thulio from BCR app configs

aignas:
Just tidy up the config. I've put my name in there, but I am happy to make the
BCR app use someone else's name as well. @oprypin, @rickeylev, what do you
think?

### DIFF
--- a/.bcr/config.yml
+++ b/.bcr/config.yml
@@ -12,8 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
----
 fixedReleaser:
-  login: aignas
-  email: 240938+aignas@users.noreply.github.com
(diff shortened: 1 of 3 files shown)


.bcr/config.yml
```diff
@@ -12,7 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+---
 fixedReleaser:
-  login: f0rmiga
```
rickeylev, line 16: I think leave this part as-is for now? I think the BCR app has to be installed for/by the user and authorized by the bazel org for that user and repo. Unless you've set that up already (seems like you'd have mentioned it before hand), lets leave this part be.

rickeylev (approved):

PR #2402, opened 2024-11-13
feat(musl): add musl toolchain

### LINKED ISSUE #1211: Downloaded python toolchain has shared library dependencies
# 🐞 bug report

### Affected Rule

n/a

### Is this a regression?

no

### Description

The python toolchains downloaded by this rule depend on an (outdated) version of the glibc.
This does not work for a number of environments.
Maybe a statically linked python binary or different versions of the toolchains could be provided instead.
I'd be happy to help work on a more hermetic (as in: not relying on system-wide libraries) python toolchain.

## 🔬 Minimal Reproduction

Register any of the builtin python toolchain for download, find thy `python3` binary in the cache and run:

```
ldd /home/malte/.cache/bazel/_bazel_malte/e58bd5cf140e0accfbab91a2f501a0a3/external/python3_10_x86_64-unknown-linux-gnu/bin/python3
        linux-vdso.so.1 (0x00007ffd062db000)
        /home/malte/.cache/bazel/_bazel_malte/e58bd5cf140e0accfbab91a2f501a0a3/external/python3_10_x86_64-unknown-linux-gnu/bin/../lib/libpython3.10.so.1.0 (0x00007fcef8400000)
        libcrypt.so.1 => /lib/x86_64-linux-gnu/libcrypt.so.1 (0x00007fcef996a000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007fcef9965000)
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007fcef9960000)
        libutil.so.1 => /lib/x86_64-linux-gnu/libutil.so.1 (0x00007fcef995b000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007fcef9872000)
        librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007fcef986d000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fcef8000000)
        /lib64/ld-linux-x86-64.so.2 (0x00007fcef99c5000)
```

As you can see, the library has versioned glibc dependencies. This is problematic in environments with newer glibcs (arch linux, fedora) or environments that use different libcs (alpine) or don't always provide a "default" glibc for tools (nix / nixos).

## 🔥 Exception or Error

When using Fedora as a host system, the build fails like this:

```
❯ bazel run @python3_10//:python3
INFO: Invocation ID: 8c2857a4-fd52-4757-9e09-9a8fa88e2440
INFO: Build option --//bazel/settings:tpm_simulator has changed, discarding analysis cache.
INFO: Analyzed target @python3_10//:python3 (1 packages loaded, 17 targets configured).
INFO: Found 1 target...
Target @python3_10_x86_64-unknown-linux-gnu//:bin/python3 up-to-date (nothing to build)
INFO: Elapsed time: 1.148s, Critical Path: 0.12s
INFO: 1 process: 1 internal.
INFO: Running command line: /home/builder/.cache/bazel/_bazel_builder/eab0d61a99b6696edb3d2aff87b585e8/external/python3_10_x86_64-unknown-linux-gnu/bin/python3
/home/builder/.cache/bazel/_bazel_builder/eab0d61a99b6696edb3d2aff87b585e8/external/python3_10_x86_64-unknown-linux-gnu/bin/python3: error while loading shared libraries: libcrypt.so.1: cannot open shared object file: No such file or directory
```

The OS provides `/usr/lib64/libcrypt.so.2`, but not `libcrypt.so.1`.

## 🌍  Your Environment

**Operating System(s):**

- Fedora 37, Fedora 38
- NixOS

```
Build label: 6.1.2
Build time: Tue Apr 18 15:29:54 2023 (1681831794)
Build timestamp: 1681831794
Build timestamp as int: 1681831794
```

**Rules_python version:**

```
0.21.0
```

**Anything else relevant?**

This issue exists: #716, but the bug was not actually fixed here.
Forcing downstream users to install old glibc versions system wide seems like it will become harder over time.

aignas:
With this and the previously landed #2406 users can use this in `sdist`
or `whl`-only setups alike. The PR itself is very simple and just adds
`musl` toolchains.

Whilst at it also move the `WhlLibc` flag out of the `pypi` namespace
since it is also used for `toolchain` matching now.

Fixes #1211


python/private/flags.bzl
```diff
@@ -132,3 +132,14 @@ FreeThreadedFlag = enum(
     # Do not use freethreaded python toolchain and wheels.
     NO = "no",
 )
+
+# Determines which libc flavor is preferred when selecting the toolchain and
+# linux whl distributions.
+#
+# buildifier: disable=name-conventions
+LibcFlag = enum(
```
rickeylev, line 140: Use flag_enum instead (slightly cleaner code in the flag def)

rickeylev (approved):

PR #2404, opened 2024-11-13
chore: change bcr presubmit to drop bazel 6, add last_rc (bazel 8)

rickeylev:
This removes Bazel 6 and adds last_rc (Bazel 8) in our BCR presubmit config. Right now, BCR releases fail because they try to test with Bazel 6, which doesn't work.


rickeylev:
bcr pr: https://github.com/bazelbuild/bazel-central-registry/pull/3180

There's a weird failure with bazel 8 on windows with a couple of the tests. So, maybe we should omit last_rc for now, otherwise future BCR releases will get hung up on that.

aignas (approved):

Issue #2386, opened 2024-11-08
Support freethreaded toolchains and wheels


aignas:
Scope:
* [x] Adding URLs that reference indygreg freethreaded toolchains. (#2372)
* [x] Running coverage using freethreaded toolchains. (#2372)
* [x] Depending on whls when freethreaded toolchains are used. (#2460)
* [x] Using a freethreaded toolchain interpreter in `pip.parse` rules to build from sdist. (#2406).

Better support for sdist building should be done separately.

Issue #2276, opened 2024-10-07
Support Windows on ARM64


skeptic-monkey:
# 🚀 feature request

### Relevant Rules

Mostly all the rules included in bazelbuild/rules_python repository (hermetic python toolchain, pip_parse, ...)

### Description

There is currently no support to build natively for Windows / ARM64 platform which is not convenient.
The Snapdragon X being used in popular laptops (surface pro for example), it now becomes quite important to be able to build for this platform natively.

### Describe the solution you'd like

The ideal would be to download an arm64 version of python and use arm64 pip requirements (and per architecture requirement seem to be already supported).

### Describe alternatives you've considered

A workaround would already to support building using an hermetic x86_64 python with x86_64 pip dependencies while using an arm64 host, relying on the x96_64 -> arm64 Windows emulation.

This doesn't work either, because pip_parse discards the x86_64 dependencies because it doesn't match the host cpu.

aignas:
I don't think I see Windows ARM64 interpreter builds for indygreg even in the [latest] release, hence I am not sure if that is something that `rules_python` has to provide.

Given that users can build their own Python interpreter and register it as a toolchain and use the pip parse rules to pull the dependencies (pending the work as part of #260, still experimental but in #2278 becomes one step closer to stabilization), I am inclined to close this as won't do.

Once indygreg provides builds for Windows ARM64 interpreter, we can add them as part of regular maintenance tasks.

[latest]: https://github.com/indygreg/python-build-standalone/releases/tag/20241008

skeptic-monkey:
It seems that indygreg might support it soon.
see https://github.com/indygreg/python-build-standalone/issues/386

aignas:
So since the interpreter bits have landed, it would be great if someone with time and such machine could step up to contribute to iron out the remaining issues.

skeptic-monkey:
We do have windows arm64 machines. I'll give it a try.

aignas:
FYI, the `bzlmod` support for `experimental_index_url` feature will be easier to add once #3058 is merged.

mbland:
While working on bazel-contrib/rules_scala#1761, I used the following patch to use `rules_python` 1.5.3 successfully on Windows ARM64:

```diff
diff --git a/examples/crossbuild/MODULE.bazel b/examples/crossbuild/MODULE.bazel
index cedb0731..668bf4b0 100644
--- a/examples/crossbuild/MODULE.bazel
[...]

PR #2406, opened 2024-11-14
refactor!(toolchain): remove uname dep in the repository_rule stage

### LINKED ISSUE #2145: getting error indicating that uname is not in path on WSL
Running Linux in WSL,
set up all Bazel dependencies
running bazel build command

hitting error 

```
ERROR: An error occurred during the fetch of repository 'python3_12':
        File "/home/williamohara/.cache/bazel/_bazel_williamohara/de29fde09bd221a16d787c59318532b3/external/rules_python/python/private/toolchains_repo.bzl", line 117, column 39, in _toolchain_aliases_impl
                (os_name, arch) = get_host_os_arch(rctx)
        File "/home/williamohara/.cache/bazel/_bazel_williamohara/de29fde09bd221a16d787c59318532b3/external/rules_python/python/private/toolchains_repo.bzl", line 419, column 50, in get_host_os_arch
                arguments = [repo_utils.which_checked(rctx, "uname"), "-m"],
        File "/home/williamohara/.cache/bazel/_bazel_williamohara/de29fde09bd221a16d787c59318532b3/external/rules_python/python/private/repo_utils.bzl", line 163, column 13, in _which_checked
                fail((
Error in fail: Unable to find the binary 'uname' on PATH.
```
it seems to be generated by this point in the toolchains_repo.bzl file
https://github.com/bazelbuild/rules_python/blob/5eff339f75ea663151b9df7e68c18c417ea91d62/python/private/toolchains_repo.bzl#L430

I have uname in my path - is there some dependency I am missing?

aignas:
Before this PR we would shell out to `uname` on UNIX systems to get the `arch`
of the toolchain - on Windows we would not need to do it because there used to
be only a single Windows platform. With this change we can correctly support
the resolution of the python interpreter on various platforms and I have also
added an env variable to customize the selection, so that users can use `musl`
or a `freethreaded` interpreter if they wish.

As part of this change, I have restricted visibility of the config settings
used in the toolchain alias repo so that we are creating fewer targets. This is
a very good time to do this before `1.0.0`.

Fixes #2145
Work towards #2276
Work towards #2386
Work towards #1211 to unblock #2402
Work towards #1361


python/private/toolchains_repo.bzl
```diff
@@ -388,57 +367,53 @@ multi_toolchain_aliases = repository_rule(
...
+        )
+        preference = repo_utils.getenv(rctx, env_var)
+        if preference == None:
+            logger.info("Consider using '{}' to select from one of the platforms: {}".format(
+                env_var,
+                candidates,
+            ))
+        elif preference not in candidates:
+            logger.fail("Please choose a preferred interpreter out of the following platforms: {}".format(candidates))
+            return None
```
rickeylev, line 407: Should we have a convention for what to do after logger.fail()? It took me a minute to remember that fail() will _actually_ fail.
I'm thinking either:
* `return logger.fail()`. This looks odd, so helps reminds that fail doesn't return, while keeping buildifier lint happy
* `return None  # To satisfy lint: unreachable because logger.fail raises error`

python/private/toolchain_aliases.bzl
```diff
@@ -0,0 +1,74 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Create toolchain alias targets."""
+
+load("@rules_python//python:versions.bzl", "PLATFORMS")
+
+def toolchain_aliases(*, name, platforms, visibility = None, native = native):
+    """Cretae toolchain aliases for the python toolchains.
```
rickeylev, line 20:
```suggestion
    """Create toolchain aliases for the python toolchains.
```

rickeylev (approved):

python/private/toolchains_repo.bzl
aignas: I think that `return logger.fail` is probably the best option here, will change and add a comment to the `logger.fail` documentation.

PR #2407, opened 2024-11-14
chore!(pip.parse): remove parse_all_requirements_files attribute

### LINKED ISSUE #2268: `experimental_index_url` in dependency tree breaks other pip.parse includes

aignas:
Remove the deprecated symbol and use the default `pip` extension in
`rules_python` to pull `twine` as part of the dependencies.

Work towards #1361
Fixes #2268 for all the users by default


groodt (approved):
LGTM

PR #2413, opened 2024-11-15
feat(gazelle): Update to latest version of go-tree-sitter

### LINKED ISSUE #2396: Gazelle seems to not like PEP 695 – Type Parameter Syntax
TL;DR: I _think_ we need to bump go-tree-sitter for python 3.12 support.

# 🐞 bug report

### Affected Rule

Gazelle

### Is this a regression?

I could argue it either way, lol. I'm going to go with "probably" because it's related to #1895.

### Description

[PEP 695](https://peps.python.org/pep-0695/) added a Type Parameter syntax that looks like:

```python
# Before
from typing import TypeVar

_T = TypeVar("_T")

def func(a: _T, b: _T) -> _T:
    ...

# After
def func[T](a: T, b: T) -> T:
    ...
```

If a python file uses this new `def func[T](...)` syntax, **and** the file has `if __name__ == "__main__"`, then Gazelle will **incorrectly generate a `py_library` target** rather than `py_binary`.

## 🔬 Minimal Reproduction

1. Make a python file:

```python
# foo.py
def func[T](a: T, b: T) -> T:
    ...

if __name__ == "__main__":
    print("hi")
```

2. Run gazelle `bazel run //:gazelle`

### Expected Result

```starlark
py_binary(
    name = "foo",
    srcs = ["foo.py"],
)
```

### Actual Result

```starlark
py_library(
    name = "foo",
    srcs = ["foo.py"],
)
```

## 🔥 Exception or Error

None

## 🌍  Your Environment

**Operating System:**

gLinux

```
$ bazel version
Bazelisk version: v1.20.0
Build label: 7.3.2
Build time: Tue Oct 1 17:46:05 2024 (1727804765)
Build timestamp: 1727804765
Build timestamp as int: 1727804765
```

**Rules_python version:**

```
$ bazel mod graph                                                                                                       
<root> (pyle@0.0.0)                    
├───aspect_bazel_lib@2.7.1            
│   └─── ...
├───bazel_skylib@1.7.1                 
│   └─── ...
├───buildifier_prebuilt@6.4.0             
│   └─── ...     
├───gazelle@0.35.0                 
│   └─── ...
├───rules_helm@0.7.0 
│   └─── ...
├───rules_python@0.37.2 
│   ├───bazel_skylib@1.7.1 (*) 
│   ├───platforms@0.0.9 (*) 
│   ├───protobuf@24.4 (*) 
│   ├───rules_proto@6.0.0-rc1 (*) 
│   ├───stardoc@0.6.2 (*) 
│   ├───bazel_features@1.11.0 
│   │   └───bazel_skylib@1.7.1 (*) 
│   └───rules_cc@0.0.9 
│       └───platforms@0.0.9 (*) 
└───rules_python_gazelle_plugin@0.37.2 
    ├───bazel_skylib@1.7.1 (*) 
    ├───gazelle@0.35.0 (*) 
    ├───rules_go@0.45.1 (*) 
    └───rules_python@0.37.2 (*)
```

**Anything else relevant?**

I haven't been able to find a human-readable changelog of `tree-sitter` or `go-tree-sitter` that explicitly says when it added support for python 3.12...

maffoo:
The latest version of go-tree-sitter includes grammer changes for python 3.12. To use it, we need to patch the build files generated by the go_repository to support building with uplevel files references. This should fix #2396.


dougthor42 (commented):
Hmm... still seems to not work on our `pyle` repo. I'll have more time to investigate early next week.

Please add a test case for python 3.12 syntax to `gazelle/python/testdata`. Specifically, we want to make sure that imports, gazelle annotations, and `if __name__` that come _after_ py3.12-only syntax are still parsed. For example:

```python
# my_baz.py
import foo
# gazelle:include_dep //:bar    

def baz[T]() -> T:  # 3.12-only syntax
    pass

# these still get parsed
import foobar
# gazelle:include_dep //:hello

if __name__ == "__main__":
    pass
```

Should generate (assuming per-file target):

```starlark
py_binary(
    name = "my_baz",
    srcs = ["my_baz.py"],
    deps = [
        "//:bar",
        "//:foo",
        "//:foobar",
        "//:hello",
    ],
)
```

If you only get 2 deps or a `py_library`, then the parsing still failed :cry:

hunshcn:
maybe we can publish go-tree-sitter with patches to bazel registry? like https://registry.bazel.build/modules/circl

maffoo:
I think I got things working now. Had to remove `go-tree-sitter` from `go.mod` because this seems to override the custom `go_respository` definition and results in patches not being applied to `go-tree-sitter` before building. I'm not sure whether we still need `go.mod` for other dependencies or if they can be handled by the `go_repository` definitions in `deps.bzl` too; experts please weight in.

I also moved the patch file into a `patches` subdirectory and added a test case exercising the py3.12 syntax that fails on master but succeeds on this branch. I tried to create a minimal failing test case based on our internal code; interestingly, I found that the function docstring is key to getting this to fail on master, no idea why that is the case...

dougthor42:
Just relaying info that @maffoo and I had at work:

This still doesn't work for `pyle`:

```
$ bazel run //:gazelle src/pyle/tools
INFO: Invocation ID: e06d2fbb-5642-4159-9abf-f1402367d298
INFO: Streaming build results to: https://btx.cloud.google.com/invocations/e06d2fbb-5642-4159-9abf-f1402367d298
WARNING: For repository 'bazel_skylib', the root module requires module version bazel_skylib@1.6.1, but got bazel_skylib@1.7.1 in the resolved dependency graph.
ERROR: no such package '@@[unknown repo 'com_github_smacker_go_tree_sitter' requested from @@rules_python_gazelle_plugin~]//python': The repository '@@[unknown repo 'com_github_smacker_go_tree_sitter' requested from @@rules_python_gazelle_plugin~]' could not be resolved: No repository visible as '@com_github_smacker_go_tree_sitter' from repository '@@rules_python_gazelle_plugin~'
ERROR: /usr/local/google/home/dthor/.cache/bazel/_bazel_dthor/dbe74c4144b5c9a438d84a119652bef9/external/rules_python_gazelle_plugin~/python/BUILD.bazel:6:11: no such package '@@[unknown repo 'com_github_smacker_go_tree_sitter' requested from @@rules_python_gazelle_plugin~]//python': The repository '@@[unknown repo 'com_github_smacker_go_tree_sitter' requested from @@rules_python_gazelle_plugin~]' could not be resolved: No repository visible as '@com_github_smacker_go_tree_sitter' from repository '@@rules_python_gazelle_plugin~' and referenced by '@@rules_python_gazelle_plugin~//python:python'
ERROR: Analysis of target '//:gazelle' failed; build aborted: Analysis failed
INFO: Elapsed time: 1.437s, Critical Path: 0.51s
INFO: 1 process: 1 internal.
```

> Here's something weird. If I run `bazel sync` in my `rules_pyle/gazelle` directory it applies patches to go-tree-sitter:
> ```
> DEBUG: /usr/local/google/home/maffoo/.cache/bazel/_bazel_maffoo/051c217dd68266c8e1eaa3206da93fe4/external/gazelle~/internal/go_repository.bzl:345:10: bazel_gazelle/go_repository: applying patches. name=com_github_smacker_go_tree_sitter, patches=[Label("//:patches/0001-go-tree-sitter-build.patch")]format_text_overflowformat_text_wrap
> ```
> but whenever I try to run some other build command I see `patches=[]`, no patches are applied, and the build fails.

> I think this is the fundamental problem. The `go_repository` definitions are only having an effect in the `WORKSPACE`. The `MODULE.bazel` uses a different mechanism with the `go_deps` extension and this doesn't know about the patches we want to apply. There's a way to specify patches for repositories brought in with `go_deps` but it only works in the root module so we can't use it (it's fine if running bazel in `rules_python/gazelle`, but when we pull this in as a dependency to `pyle` the patches get ignored again. The only way I can see to fix this is to update `bazel-gazelle` itself or add go-tree-sitter to the bazel central registry.
---

> maybe we can publish go-tree-sitter with patches to bazel registry?

It's looking like we might have to do that. We can apply patches like @maffoo has done, but then those patches won't be used when some other project uses `rules_python_gazelle_plugin` because things like [`go_deps.module_override`](https://github.com/bazel-contrib/bazel-gazelle/blob/28f43a444ecc4c0498928787ebb1e61fef08f6eb/internal/bzlmod/go_deps.bzl#L819-L837) and will only work on the [root module](https://github.com/bazel-contrib/bazel-gazelle/blob/28f43a444ecc4c0498928787ebb1e61fef08f6eb/internal/bzlmod/go_deps.bzl#L89).

@hunshcn How do you recommend we publish to BCR? Fork go-tree-sitter, add MODULE.bazel and rules, and go from there? Or is there another option?

hunshcn:
Just like https://github.com/bazelbuild/bazel-central-registry/pull/674
Maybe @fmeum can provide some information about.

dougthor42:
Ah thanks, that process looks pretty easy. I'll start it now.

maffoo:
I filed https://github.com/bazel-contrib/bazel-gazelle/issues/1984 to propose allowing default patches on libraries included by go_deps and put up https://github.com/bazel-contrib/bazel-gazelle/pull/1985 to actually implement this. We'll see what the bazel-gazelle maintainers say...

ewianda:
Will the BCR option not be limited to the bzlmod implementation? 

I have a fork of the gazelle plugin and made this upgrade using http_archive approach, which is compatible with MODULE.bazel and WORKSPACE. 

https://github.com/benchsci/rules_python_gazelle/blob/8b5c75b46bcf270bd7360c71daaa44e2c7a01fdd/MODULE.bazel#L14-L23

https://github.com/benchsci/rules_python_gazelle/commit/8b5c75b46bcf270bd7360c71daaa44e2c7a01fdd#diff-7ad5c39e688ac3ccc3847bdf065963457ec66024308f03dcf1045cafa2c38108


I can contribute the change here.

maffoo:
Superseded by #2496.

Issue #2318, opened 2024-10-20
Stage-2 bootstrap script incorrectly removes inherited PYTHONPATH entries on Python 3.11 and above


chowder:
# 🐞 bug report

### Affected Rule

The issue is caused by the rule: `py_binary` / `py_test`

### Is this a regression?

Yes. 

### Description

[The stage-1 bootstrap script runs `python` with `PYTHONSAFEPATH=1`](https://github.com/bazelbuild/rules_python/blob/main/python/private/stage1_bootstrap_template.sh#L104-L115), which means that no unsafe paths are preprended to the runtime environment of stage-2.

However, [the stage-2 script unconditionally deletes `sys.path[0]`](https://github.com/bazelbuild/rules_python/blob/main/python/private/stage2_bootstrap_template.py#L13). This is harmless most of the time, since the search path that is deleted is `/usr/lib/python311.zip` (or similar).

But if the target caller attempts to propagate `PYTHONPATH` to the application, the first entry in `PYTHONPATH` is incorrectly removed. 

## 🔬 Minimal Reproduction

`script.py`:
```python
import sys
print("\n".join(sys.path))
```

`BUILD.bazel`:
```
py_binary(
    name = "script",
    srcs = [":script.py"],
)
```

Build and run the target with `PYTHONPATH=foo:bar`, observe that `foo` is missing from `sys.path`. 

## 🌍  Your Environment

**Operating System:**

Linux AMD64 - CentOS 7, and Debian 11

**Output of `bazel version`:**

Bazel 7.2.1

**Rules_python version:**

0.36.0 (but present in 0.37.0 too)

chowder:
I suppose something is to be said about the smell that is attempting to set `PYTHONPATH` for a Bazel target. 

Where I run into this problem is with VSCode/PyCharm debuggers which run their own bootstrap code via `PYTHONPATH` propagation, before running your actual target. 

In any case, a simple fix for this is to replace [this line](https://github.com/bazelbuild/rules_python/blob/main/python/private/stage2_bootstrap_template.py#L11-L13), with:

```python
if not getattr(sys.flags, "safe_path", False):
    del sys.path[0]
```

Happy to raise a PR if agreed.

rickeylev:
> simple fix is to replace the del sys.path[0] line with guard

Yes. A PR for that would be approved.

PR #2418, opened 2024-11-17
fix: only delete first sys.path entry in the stage-2 bootstrap if PYTHONSAFEPATH is unset or unsupported

chowder:
Unnconditionally deleting the first `sys.path` entry on the stage-2 bootstrap incorrecly removes a valid search path on Python 3.11 and above, since `PYTHONSAFEPATH` is already unconditionally set in stage-1. 

It should be deleted only if it is unset or unsupported. 

Fixes https://github.com/bazelbuild/rules_python/issues/2318


python/private/stage2_bootstrap_template.py
```diff
@@ -4,13 +4,17 @@
...
-# TODO: Use sys.flags.safe_path to determine whether this removal should be
-# performed
-del sys.path[0]
+# and is a special case of #7091.
+#
+# Python 3.11 introduced an PYTHONSAFEPATH (-P) option that disables this
+# behaviour, which we set in the stage 1 bootstrap.
+# So the prepended entry needs to be removed only if the above option is either
+# unset or not supported by the interpreter.
+# TODO(#7091): Remove this hack when we drop support for Python 3.10 and below
```
chowder, line 15: In case it wasn't clear, the original comment was:

> `TODO(#7091): Remove this hack when no longer necessary.`

I've changed it to: 

> `TODO(#7091): Remove this hack when we drop support for Python 3.10 and below`

tests/integration/no_unsafe_paths/test.py
```diff
@@ -0,0 +1,44 @@
...
+        # We can use this as a marker to verify that during bootstrapping,
+        # (1) no unexpected paths were prepended, and (2) no paths were
+        # accidentally dropped.
+        #
+        major, minor, *_ = sys.version_info
+        archive = f"python{major}{minor}.zip"
+
+        # < Python 3.11 behaviour
+        if (major, minor) < (3, 11):
+            # Because of https://github.com/bazelbuild/rules_python/blob/0.39.0/python/private/stage2_bootstrap_template.py#L415-L436
```
chowder, line 35: Whilst adding tests for this change, I discovered what might be another bug? 

https://github.com/bazelbuild/rules_python/blob/0.39.0/python/private/stage2_bootstrap_template.py#L415-L436 

This snippet always executes on Python 3.10 and below, because `safe_path` hasn't been implemented yet.

But it will never execute on Python 3.11, because `PYTHONSAFEPATH` is unconditionally set in stage 1. 

Not sure if this is expected behaviour.

tests/integration/no_unsafe_paths/BUILD.bazel
```diff
@@ -0,0 +1,34 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+load("@python//3.10:defs.bzl", py_test_3_10 = "py_test")
+load("@python//3.11:defs.bzl", py_test_3_11 = "py_test")
+
+py_test_3_10(
```
rickeylev, line 18: Instead of an integration test, use `sh_py_run_test`. These are much lighter weight (don't require a bazel-within-bazel to run) and easier to run and debug.

See tests/bootstrap_impls/BUILD.bazel for examples. Do something akin to this:

```
load("//tests/support:sh_py_run_test.bzl", "py_reconfig_test")

py_reconfig_test(
  srcs=["whatever.py"],
  python_version="3.10",
  bootstrap_impl = "script",
  target_compatible_with = _SUPPORTS_BOOTSTRAP_SCRIPT
)
```

You can probably copy/paste this build file and replace the `py_test_XX` calls with `py_reconfig_test` and it should just work.

Adding them to `tests/bootstrap_impls` is probably easiest. A new directory, `tests/no_unsafe_paths` is also SGTM.

rickeylev (changes requested):

rickeylev:
Thanks for the PR. This mostly LGTM. Just need to use a non-integration test for this instead. See comment.

tests/integration/no_unsafe_paths/BUILD.bazel
chowder: These rules are neat - thanks!

tests/support/support.bzl
```diff
@@ -39,3 +41,8 @@ PRECOMPILE_SOURCE_RETENTION = str(Label("//python/config_settings:precompile_sou
 PYC_COLLECTION = str(Label("//python/config_settings:pyc_collection"))
 PYTHON_VERSION = str(Label("//python/config_settings:python_version"))
 VISIBLE_FOR_TESTING = str(Label("//python/private:visible_for_testing"))
+
+SUPPORTS_BOOTSTRAP_SCRIPT = select({
+    "@platforms//os:windows": ["@platforms//:incompatible"],
+    "//conditions:default": [],
+}) if IS_BAZEL_7_OR_HIGHER else ["@platforms//:incompatible"]
```
chowder, line 48: Is this a good place to put this?

rickeylev: LGTM

python/private/stage2_bootstrap_template.py
rickeylev, line 15:
```suggestion
# NOTE: This can be removed when Python 3.10 and below is no longer supported
```

rickeylev: Good clarification, thanks for that. I updated it a bit more.

rickeylev:
Hrm. I think "intended behavior" here is all murky.

I'm pretty sure the intent of `del sys.path[0]` was to have safe path like behavior before python 3.11. However, that extra logic you found in stage2_bootstrap negates that? And the comment for that code points back to it being there for some sort of compatibility reason with how the system-python bootstrap behaved (but I'm not sure how correct that comment is, in retrospect).

At the least, we can confidently say: for 3.11 and later, there shouldn't be any `del sys.path[0]` nor any re-adding it back. Instead, we rely on stage1 to set PYTHONSAFEPATH instead.

tests/integration/no_unsafe_paths/test.py
rickeylev: Hmm, I wish I could remember what this was about, exactly. Maybe if I went and dug through the change history...

This _might_ be about how PYTHONSAFEPATH being set by the outer environment is respected (by the bootstrap impl). While stage1 bootstrap will set it by default, it first checks to see if the outer environment has set it. If so, it'll respect that.

aignas (approved):
LGTM, I agree with the reasoning in @rickeylev's message above.

rickeylev:
Ok, I think I know what this logic is about. It's fixing up the path0 entry to point to the proper directory (runfiles instead of the filepath on the command line).

At startup, path0 looks like e.g. `<exec root>/tests/bootstrap_impls`. That directory comes from the invocation `python $runfiles_root/tests/bootstrap_impls/foo_stage2.py`, which is symlink back to `<exec root>/tests/bootstrap_impls`.

But, that directory doesn't have much in it, plus it means things outside the runfiles become importable. So the later `prepend_path_entries = ...` logic fixes that: it addes back the correct path, `<execroot>/tests/bootstrap_impls/foo.runfiles/_main/tests_bootstrap_impls`.

Hence, if safe path is disabled, we have to go manipulating sys.path. If safe path is enabled, then there is nothing to manipulate.

I did come across an edge case while looking into this: I happen to have PYTHONPATH set in my env. on python 3.11, because of that `del sys.path[0]` line, it ends up popping that off when it shouldn't.

chowder:
> I did come across an edge case while looking into this: I happen to have PYTHONPATH set in my env. on python 3.11, because of that del sys.path[0] line, it ends up popping that off when it shouldn't.

This PR should address that problem, shouldn't it? 

The burning question I have is - should `PYTHONSAFEPATH` be set in stage-1? 

Removing it will have all Python versions behave identically, and in line with historical behaviour. 

Is there something undesirable about having the script's directory (where it doesn't follow symlinks) in `sys.path[0]`?

chowder:
Pretty sure the build failures with `last_rc` Bazel (8.0.0rc3) are unrelated to these changes. 

It's because `@bazel_features` evaluates 8.0.0rc3 to be "less than" 8.0.0, so it still thinks [these](https://github.com/bazel-contrib/bazel_features/blob/main/private/globals.bzl#L23-L31) are still valid globals.

aignas:
> The burning question I have is - should `PYTHONSAFEPATH` be set in stage-1?
> 
> Removing it will have all Python versions behave identically, and in line with historical behaviour.

I think that PYTHONSAFEPATH was being set in the python template in the bazel repo since quite a while ago, so the historic behaviour might be to have PYTHONSPAFEPATH to be used if it is possible.

However I can't find a link to the old source quickly, so my memory may be failing me.

chowder:
Anything left outstanding from my end to get this across the line? :)

rickeylev (approved):

rickeylev:
LGTM. Sorry, took me awhile to get the free time to finish reviewing.

PR #2420, opened 2024-11-17
refactor: Add GAZELLE_VERBOSE env and log parser failures

dougthor42:
While investigating #2396 and why #2413 doesn't appear to be working for us, I realized that one of the things I was making heavy use of was additional parser logging that I had added. This adds some of that logging. I also throw in some documentation because I found it helpful.

Users can (attempt to) get additional parse failure information by setting the `GAZELLE_VERBOSE` environment variable to `1`. Eg:

```console
$ GAZELLE_VERBOSE=1 bazel run //:gazelle
```

Here are some example logs:

```console
$ GAZELLE_VERBOSE=1 bazel run //:gazelle
INFO: Invocation ID: a4e026d8-17df-426c-b1cc-d3980690dd53
...
INFO: Running command line: bazel-bin/gazelle
INFO: Streaming build results to: https://btx.cloud.google.com/invocations/a4e026d8-17df-426c-b1cc-d3980690dd53
gazelle: WARNING: failed to parse "hello/get_deps.py". The resulting BUILD target may be incorrect.
gazelle: Parse error at {Row:1 Column:0}:
def search_one_more_level[T]():
gazelle: The above was parsed as: (ERROR (identifier) (call function: (list (identifier)) arguments: (argument_list)))
gazelle: ERROR: failed to generate target "//hello:get_deps" of kind "py_library": a target of kind "pyle_py_binary" with the same name already exists. Use the '# gazelle:python_library_naming_convention' directive to change the naming convention.
$
$ bazel run //:gazelle
INFO: Invocation ID: 726c9fd6-f566-4c30-95ef-c4781ad155de
...
INFO: Running command line: bazel-bin/gazelle
INFO: Streaming build results to: https://btx.cloud.google.com/invocations/726c9fd6-f566-4c30-95ef-c4781ad155de
gazelle: WARNING: failed to parse "hello/get_deps.py". The resulting BUILD target may be incorrect.
gazelle: ERROR: failed to generate target "//hello:get_deps" of kind "py_library": a target of kind "pyle_py_binary" with the same name already exists. Use the '# gazelle:python_library_naming_convention' directive to change the naming convention.
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,7 +61,8 @@ Unreleased changes template.
 
 {#v0-0-0-added}
 ### Added
-* Nothing yet.
+* (gazelle): Parser failures will now be logged to the terminal. Additional
+  details can be logged by setting `GAZELLE_VERBOSE=1`.
 
 {#v0-0-0-removed}
 ### Removed
--- a/gazelle/python/file_parser.go
+++ b/gazelle/python/file_parser.go
@@ -58,7 +58,7 @@ func NewFileParser() *FileParser {
 
 // ParseCode instantiates a new tree-sitter Parser and parses the python code, returning
 // the tree-sitter RootNode.
-// It print a warning if parsing fails.
+// It prints a warning if parsing fails.
 func ParseCode(code []byte, path string) (*sitter.Node, error) {
 	parser := sitter.NewParser()
 	parser.SetLanguage(python.GetLanguage())
--- a/gazelle/python/file_parser.go
+++ b/gazelle/python/file_parser.go
@@ -17,6 +17,7 @@ package python
 import (
 	"context"
 	"fmt"
+	"log"
 	"os"
 	"path/filepath"
 	"strings"
@@ -55,7 +56,10 @@ func NewFileParser() *FileParser {
 	return &FileParser{}
 }
 
-func ParseCode(code []byte) (*sitter.Node, error) {
+// ParseCode instantiates a new tree-sitter Parser and parses the python code, returning
+// the tree-sitter RootNode.
+// It print a warning if parsing fails.
+func ParseCode(code []byte, path string) (*sitter.Node, error) {
 	parser := sitter.NewParser()
 	parser.SetLanguage(python.GetLanguage())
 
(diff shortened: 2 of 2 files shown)


gazelle/python/file_parser.go
```diff
@@ -64,9 +68,27 @@ func ParseCode(code []byte) (*sitter.Node, error) {
 		return nil, err
 	}
 
-	return tree.RootNode(), nil
+	root := tree.RootNode()
+	if root.HasError() {
+		log.Printf("WARNING: failed to parse %q. The resulting BUILD target may be incorrect.", path)
+
+		verbose, envExists := os.LookupEnv("GAZELLE_VERBOSE")
+		if envExists && verbose == "1" {
```
dougthor42, line 76: If people are happy with this idea, I will make a more generalized function that can be reused. Maybe even add verbosity levels.

I didn't see any option in the core `gazelle` that allows for different verbosity levels, so if there is one please LMK.

gazelle/python/file_parser.go
```diff
@@ -64,9 +68,27 @@ func ParseCode(code []byte) (*sitter.Node, error) {
...
+			for i := 0; i < int(root.ChildCount()); i++ {
+				child := root.Child(i)
+				if child.IsError() {
+					log.Printf("Parse error at %+v:\n%+v", child.StartPoint(), child.Content(code))
+					log.Printf("The above was parsed as: %v", child.String())
```
dougthor42, line 81: Add comments demo'ing the logged message.

gazelle/python/file_parser.go
```diff
@@ -55,7 +56,10 @@ func NewFileParser() *FileParser {
 	return &FileParser{}
 }
 
-func ParseCode(code []byte) (*sitter.Node, error) {
+// ParseCode instantiates a new tree-sitter Parser and parses the python code, returning
+// the tree-sitter RootNode.
+// It print a warning if parsing fails.
```
rickeylev, line 61: typo:prints
```suggestion
// It prints a warning if parsing fails.
```

CHANGELOG.md
```diff
@@ -61,7 +61,8 @@ Unreleased changes template.
 
 {#v0-0-0-added}
 ### Added
-* Nothing yet.
+* (gazelle): Parser failures will now be logged to the terminal. Additional
+  details can be logged by setting `GAZELLE_VERBOSE=1`.
```
rickeylev, line 65: You can add this to docs/environment-variables.md, too.

rickeylev (approved):
I changed the title to refactor, since it isn't providing a feature, more just better diagnostics for debugging failures.

gazelle/python/file_parser.go
```diff
@@ -64,9 +68,27 @@ func ParseCode(code []byte) (*sitter.Node, error) {
...
+		if envExists && verbose == "1" {
+			for i := 0; i < int(root.ChildCount()); i++ {
+				child := root.Child(i)
+				if child.IsError() {
+					log.Printf("Parse error at %+v:\n%+v", child.StartPoint(), child.Content(code))
+					log.Printf("The above was parsed as: %v", child.String())
+				}
+			}
+		}
+	}
```
aignas, line 85: nit, it is much more idiomatic to do early returns:
```suggestion
	if !root.HasError() {
		return root, nil
	}
	
	log.Printf("WARNING: failed to parse %q. The resulting BUILD target may be incorrect.", path)

	// TODO: if root.HasError, should we still return err=nil?
	verbose, envExists := os.LookupEnv("GAZELLE_VERBOSE")
	if !envExists || verbose != "1" {
		return root, nil
	}
	
	for i := 0; i < int(root.ChildCount()); i++ {
		child := root.Child(i)
		if child.IsError() {
			log.Printf("Parse error at %+v:\n%+v", child.StartPoint(), child.Content(code))
			log.Printf("The above was parsed as: %v", child.String())
		}
	}
	return root, nil
```

aignas (approved):
In general LGTM, but have questions if we should return `err=nil` if `root.HasError()`. If the Python file cannot be parsed, why should we continue?

gazelle/python/file_parser.go
dougthor42: Thanks! I'll update to that in a followup PR.

dougthor42 (commented):
> if we should return err=nil if root.HasError(). If the Python file cannot be parsed, why should we continue?

IMO we _should_ continue on parse errors. It's possible that the parse error is unrelated to anything that gazelle cares about, and in that case the resulting target definition will be correct.

PR #2422, opened 2024-11-18
fix(uv): Fix sha256 for uv binary for aarch64-apple-darwin

### LINKED ISSUE #2411: uv tool sha256sum is incorrect for aarch64-apple-darwin
# 🐞 bug report

### Affected Rule

 The issue is caused by the rule: `uv = use_extension("@rules_python//python/uv:extensions.bzl", "uv")`

### Is this a regression?

Yes, the previous version in which this bug was not present was: Yes. Bad hash introduced here: https://github.com/bazelbuild/rules_python/pull/2333/files#diff-1757fb851d4a183b61f34ea410721b38095ac28194ec1b6b233152155febb69c.

### Description

The `sha256sum` for `uv` or aarch64-apple-darwin is incorrect. It's coded as: `35786030f926e3d34d186edc0ea3989698e57755852af9ae4b39da5109abcbfa`. It should be: `bb2ff4348114ef220ca52e44d5086640c4a1a18f797a5f1ab6f8559fc37b1230`. See here for confirmation: https://github.com/astral-sh/uv/releases/download/0.4.25/uv-aarch64-apple-darwin.tar.gz.sha256.

## 🔬 Minimal Reproduction

Use `uv` on an aarch64 macbook

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
Macbook Pro M1
  </code>
</pre>

<pre>
  <code>
Bazelisk version: 1.20.0
Build label: 7.3.2
Build time: Tue Oct 1 17:48:21 2024 (1727804901)
Build timestamp: 1727804901
Build timestamp as int: 1727804901
  </code>
</pre>

**Rules_python version:**

<pre>
  <code>
0.38.0
  </code>
</pre>

**Anything else relevant?**

See here to see if this has been fixed: https://github.com/bazelbuild/rules_python/blob/main/python/uv/private/versions.bzl#L73

schultetwin:
`uv` cannot be used on aarch64-apple-darwin computers because the sha256sum is incorrect.

The correct version can be seen [here](https://github.com/astral-sh/uv/releases/download/0.4.25/uv-aarch64-apple-darwin.tar.gz.sha256).

Fixes #2411

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,6 +78,8 @@ Other changes:
 * (pypi): Correctly handle multiple versions of the same package in the requirements
   files which is useful when including different PyTorch builds (e.g. <pytorch+cpu> vs <pytorch+cu118> ) for different target platforms.
   Fixes ([2337](https://github.com/bazelbuild/rules_python/issues/2337)).
+* (uv): Correct the sha256sum for the `uv` binary for aarch64-apple-darwin.
+  Fixes ([2411](https://github.com/bazelbuild/rules_python/issues/2411)).
 
 {#v0-0-0-added}
 ### Added
--- a/examples/bzlmod/MODULE.bazel.lock
+++ b/examples/bzlmod/MODULE.bazel.lock
@@ -9623,7 +9623,7 @@
     },
     "@@rules_python~//python/uv:extensions.bzl%uv": {
       "general": {
-        "bzlTransitiveDigest": "umgu1yR4Wmqb1pJa+9hAcs9dPbeqBsPLceKiaEg3DdQ=",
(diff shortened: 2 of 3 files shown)


aignas (approved):
Thanks. Just out of curiosity, how are you using the `uv` tool today?

schultetwin:
In the repo I discovered this in, it was being used to build a universal lock file to simplify lock file management, as there were aarch64-darwin, x86_64-darwin and x86_64-linux computers being used in this repo, and I wanted something that would work for all. 

I think this would also have been do able using multiple pip_parse and multiple compile_pip_requirements, but uv worked, and was much faster than previously (where that only resolved for one platform).

Issue #2423, opened 2024-11-18
[pypi] PEP508 Remove the need for Python in env_marker evaluation in hub repos


aignas:
Related #2319 

This is to add a parser for PEP508 in starlark and to use it in the hub_repository instead of evaluating markers with Python in the module extension evaluation phase.

This will be needed in other places as well:
* Using the same starlark parser in parsing the METADATA.
* Parsing `uv.lock` and other lock files that have the marker as part of their deps.
* Allowing us to push the `env_marker` evaluation to the analysis phase together with the other config setting evaluation.
* Allowing to properly support the full spec of PEP508 in the `whl_library` without exploding the number of args we need to pass to `whl_library`.
* Allowing to create something similar to a `whl_archive` repository rule that would not depend on `Python` that we could dog-food ourselves for the repo rules that need Python at the repo phase (i.e. the current `whl_library`).

aignas:
Python dependency can only be removed once we have the #2909. I am closing this in favor of tracking the work in #2747, so that we can enable that to users once we have the API for them to configure things.

Issue #2319, opened 2024-10-21
[pypi] Support python micro version in env marker based dependency evaluation


psalaberria002:
# 🚀 feature request

### Relevant Rules

The `whl_installer` Python program needs to be updated.

### Description

whl_installer should be aware of the selected full Python version (<major>.<minor>.<patch>) when resiolving dependencies.

Today, it is only aware of the minor version.

See the [Redis  5.1.1 wheel](https://files.pythonhosted.org/packages/15/f1/feeeaaaac0f589bcbc12c02da357cf635ee383c9128b77230a1e99118885/redis-5.1.1-py3-none-any.whl). In the METADATA file it has `Requires-Dist: async-timeout>=4.0.3; python_full_version < "3.11.3"`.

If the Python runtime is set to 3.11.10, the async-timeout dependency should be excluded, but it isn't. The repo BUILD file looks like this:
```
load("@bazel_skylib//rules:copy_file.bzl", "copy_file")
load("@rules_python//python:defs.bzl", "py_library", "py_binary")

package(default_visibility = ["//visibility:public"])

filegroup(
    name = "dist_info",
    srcs = glob(["site-packages/*.dist-info/**"], allow_empty = True),
)

filegroup(
    name = "data",
    srcs = glob(["data/**"], allow_empty = True),
)

filegroup(
    name = "whl",
    srcs = ["redis-5.1.1-py3-none-any.whl"],
    data = ["@python_deps//async_timeout:whl"],
    visibility = ["//visibility:public"],
)

py_library(
    name = "pkg",
    srcs = glob(
        ["site-packages/**/*.py"],
        exclude=[],
        # Empty sources are allowed to support wheels that don't have any
        # pure-Python code, e.g. pymssql, which is written in Cython.
        allow_empty = True,
    ),
    data = [] + glob(
        ["site-packages/**/*"],
        exclude=["**/* *", "**/*.py", "**/*.pyc", "**/*.pyc.*", "**/*.dist-info/RECORD"],
    ),
    # This makes this directory a top-level in the python import
    # search path for anything that depends on this.
    imports = ["site-packages"],
    deps = ["@python_deps//async_timeout:pkg"],
    tags = ["pypi_name=redis", "pypi_version=5.1.1"],
    visibility = ["//visibility:public"],
)
```

aignas:
This is something of a known issue and the current design has this limitation described in the issue. To properly solve this I would like to have the following solution:
[...]

PR #2424, opened 2024-11-18
refactor(pypi): move config setting processing to the macro

aignas:
Before hand we would pass around `whl_alias` struct instances and it
would go through rounds of starlark functions which would end up
rendered as BUILD.bazel files in the hub repository. This means that
every time the output of the said functions would change, the
`MODULE.bazel.lock` would also possibly change.

Since we now have much better unit testing framework, we start relying
on these functions being evaluated at loading phase instead of repo rule
phase, which means that we can make the `BUILD.bazel` files not change a
lot when the underlying code changes, this should make the code more
maintainable in the long run and unblocks the code to accept extra
things, like env marker values or similar.

Summary:
- Remove the `repo` field from `whl_alias`.
- Rename `whl_alias` to `whl_config_setting`
- Move the tests around
- Simplify the `pkg_aliases` tests to use `contains` instead of exact
  equality to make things less brittle.
- Simplify the `pkg_aliases` tests to use different mocks to make
  expectations easier to understand.
- Make `whl_config_setting` hashable by using tuples instead of lists.
- Adjust how we store the `whl_config_setting` in the PyPI extension
  and optimize the passing to the `hub_repository`.

This is needed for #2319 and #2423.

To be in future PRs:
* Remove the need to pass `osx_versions`, etc to the `pkg_aliases` macro.


python/private/pypi/pkg_aliases.bzl
```diff
@@ -132,3 +145,253 @@ def pkg_aliases(
...
+         * A list of config settings that are generated by ./pip_config_settings.bzl
+         * The list of default version settings.
+    """
+    prefixes = []
+    suffixes = []
+    setting_supported_versions = {}
+
+    if filename.endswith(".whl"):
+        if (0, 0) in glibc_versions:
+            fail("Invalid version in 'glibc_versions': cannot specify (0, 0) as a value")
```
rickeylev, line 288: Same comment here about wording as above.

python/private/pypi/pkg_aliases.bzl
```diff
@@ -132,3 +145,253 @@ def pkg_aliases(
...
+        for p in prefixes
+        for suffix, versions in setting_supported_versions.items()
+    }
+
+    if suffixes or target_platforms or versioned:
+        return [
+            ":is_cp{}{}_{}".format(python_version, p, s)
+            for p in prefixes
+            for s in suffixes or [_non_versioned_platform(p) for p in target_platforms or []]
+        ], versioned
```
rickeylev, line 339: Can you unroll this a bit? The nested list comps and `or`'d expression make this hard to read and I'm wary of remembering precedence correctly.

python/private/pypi/render_pkg_aliases.bzl
```diff
@@ -62,33 +62,65 @@ which has a "null" version value and will not match version constraints.
...
+        if alias.filename:
+            config_setting = """whl_config_setting(\n{}\n)""".format(
+                render.indent("\n".join([
+                    "{} = {},".format(k, repr(v))
+                    for k, v in {
+                        "filename": alias.filename,
+                        "target_platforms": alias.target_platforms,
+                        "version": alias.version,
+                    }.items()
+                    if v
```
rickeylev, line 79: How about creating text_util.render_kwargs() ? I've had to write similar code to this before. Looks like this same logic is duplicated down below, too.

rickeylev (approved):

python/private/pypi/pkg_aliases.bzl
aignas: OK, I will clean this up.

aignas:
I need to revisit the error message wording here. Will do later.

PR #2426, opened 2024-11-18
feat(pypi):  Add extra_hub_aliases to pip_repository too

gholms:
The extra_hub_aliases feature from #2369 has bindings in the pip extension
for bzlmod users, but not pip_repository for WORKSPACE users.  But it
_can_ work for either, so this patch moves it to the list of attrs the
two share and makes it work with the latter.


CHANGELOG.md
```diff
@@ -71,6 +71,7 @@ Unreleased changes template.
 
 Other changes:
 * (python_repository) Start honoring the `strip_prefix` field for `zstd` archives.
+* (pypi) extra_hub_aliases now works with pip_parse in WORKSPACE files.
```
aignas, line 74:
```suggestion
* (pypi) {bzl:obj}`pip_parse.extra_hub_aliases` now works in WORKSPACE files.
```

aignas (approved):
LGTM

Issue #2169, opened 2024-08-31
plain invocation of python subprocess doesn't inherit sys.path for bootstrap_impl=script


rickeylev:
A side-effect of no longer propagating import paths using the `PYTHONPATH` envvar is that subprocesses don't inherit the paths. This is usually a good thing, but ends up breaking plain calls to python that assume they're going to inherit the current python's settings.

An example is pre-commit and its invocation of virtual env:

```
# add pre-commit to requirements and process through pip.parse

# BUILD.bazel
load("//python/entry_points:py_console_script_binary.bzl", "py_console_script_binary")

py_console_script_binary(
    name = "pre-commit",
    pkg = "@dev_pip//pre_commit",
    script = "pre-commit",
)

bazel run --@rules_python//python/config_settings:bootstrap_impl=script //:pre-commit
```

Eventually, it'll run: `[sys.executable, '-mvirtualenv', ...]`, its sys.path will be just the stdlib, and fail to import `virtualenv`

-----

This is sort of WAI. Part of the purpose of bootstrap_impl=script is to no longer use the envvar so that PYTHONPATH doesn't get too long and bleed into subprocesses.

I'm not sure how to work around this. I guess a legacy option to set the env var?

I'm not sure how this is supposed to work outside of bazel, either. It must assume that it's invoked in a venv or something? The surrounding code seems to indicate it's setting _up_ a venv for pre-commit itself...or something. This all seems odd -- I would have to create a venv with virtualenv in it to run pre-commit so pre-commit can create its own venv? That doesn't sound right.

aignas:
At $dayjob I had a similar usecase where I need to have python interpreter with all dependencies set up and with the way the new thing is setup, I am not sure how I could achieve that using the new bootstrap. I had a py_binary that was using sys.executable and just forward the args to the interpreter and using pythonpath env var would just work, but with the new method, I would need to also setup the sys.path myself before invoking the interpreter.

Maybe at the very least there is a way to workaround this where sys.executable could be set to something that sets up the sys.path.

rickeylev:
I was reading some Python docs (venv or site, i can't remember), and they gave me the following idea:
* A binary creates a wrapper for the interpreter. This becomes sys.executable
* In that same directory, there's a pth file
* At interpreter startup (or site init?), it reads pth files from its directory.

ewianda:
> I was reading some Python docs (venv or site, i can't remember), and they gave me the following idea:
> 
[...]

PR #2439, opened 2024-11-23
fix: correctly obtain relative path required for the venv created by `--bootstrap_impl=script`

chowder:
Computing the relative path from the venv interpreter to the underlying interpreter was
incorrectly using the actual interpreter's directory depth, not the venv interpreter's
directory depth, when computing the distance from the venv interpreter to the runfiles root.
The net effect is the correct relative path would only be computed for binaries with
the same directory depth as the actual interpreter (e.g. 2).

This went undetected in CI because the tests for this logic just happen to have the
same directory depth as the actual interpreter used.

To fix, compute the relative path to the runfiles root using the venv interpreter
directory. Also added a test in a more nested directory to test this case.

Along the way:
* Change relative path computation to compute a minimum relative path.
* Fix the internals to pass a runfiles-root relative path, not main-repo relative path,
  for the actual interpreter, as intended.

Fixes https://github.com/bazelbuild/rules_python/issues/2169

### DIFF
--- a/tests/bootstrap_impls/venv_relative_path_tests.bzl
+++ b/tests/bootstrap_impls/venv_relative_path_tests.bzl
@@ -12,110 +12,93 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"Unit tests for yaml.bzl"
+"Unit tests for relative_path computation"
 
-load("@bazel_skylib//lib:unittest.bzl", "asserts", "unittest")
+load("@rules_testing//lib:test_suite.bzl", "test_suite")
 load("//python/private:py_executable_bazel.bzl", "relative_path")  # buildifier: disable=bzl-visibility
 
-def _relative_path_test_impl(ctx):
-    env = unittest.begin(ctx)
+_tests = []
 
+def _relative_path_test(env):
     # Basic test cases
 
-    asserts.equals(
-        env,
-        "../../c/d",
+    env.expect.that_str(
         relative_path(
             from_ = "a/b",
             to = "c/d",
         ),
-    )
+    ).equals("../../c/d")
 
-    asserts.equals(
-        env,
-        "../../c/d",
+    env.expect.that_str(
         relative_path(
             from_ = "../a/b",
             to = "../c/d",
         ),
-    )
+    ).equals("../../c/d")
 
-    asserts.equals(
-        env,
-        "../../../c/d",
+    env.expect.that_str(
         relative_path(
             from_ = "../a/b",
             to = "../../c/d",
         ),
-    )
+    ).equals("../../../c/d")
 
-    asserts.equals(
-        env,
-        "../../d",
+    env.expect.that_str(
         relative_path(
             from_ = "a/b/c",
             to = "a/d",
         ),
-    )
-
-    asserts.equals(
-        env,
-        "d/e",
+    ).equals("../../d")
+    env.expect.that_str(
         relative_path(
             from_ = "a/b/c",
             to = "a/b/c/d/e",
         ),
-    )
+    ).equals("d/e")
 
     # Real examples
 
     # external py_binary uses external python runtime
-    asserts.equals(
-        env,
-        "../../../../../rules_python~~python~python_3_9_x86_64-unknown-linux-gnu/bin/python3",
+    env.expect.that_str(
         relative_path(
-            from_ = "../rules_python~/python/private/_py_console_script_gen_py.venv/bin",
-            to = "../rules_python~~python~python_3_9_x86_64-unknown-linux-gnu/bin/python3",
+            from_ = "other_repo~/python/private/_py_console_script_gen_py.venv/bin",
+            to = "rules_python~~python~python_3_9_x86_64-unknown-linux-gnu/bin/python3",
         ),
+    ).equals(
+        "../../../../../rules_python~~python~python_3_9_x86_64-unknown-linux-gnu/bin/python3",
     )
 
     # internal py_binary uses external python runtime
-    asserts.equals(
-        env,
-        "../../../../rules_python~~python~python_3_9_x86_64-unknown-linux-gnu/bin/python3",
+    env.expect.that_str(
         relative_path(
-            from_ = "test/version_default.venv/bin",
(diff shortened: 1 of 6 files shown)


ewianda:
The following rules were tested and validated:

- [x] **py_binary**  
- [x] **py_test**  
- [x] **py_console_script**  
- [x] **rules_oci Python image**

---

rickeylev (commented):
Ah, hah. I'm guessing the reason this worked in CI is the path depth of the test programs was the same as the path depth of the underlying interpreter? :joy: 

Thanks for the PR!

The relative path logic looks overly complicated, though? All it has to do is calculate how many up-escapes are needed to get from the `{prefix}.venv/bin/python3` location to the runfiles root, then append the `../`-stripped value of the actual interpreter to that.

I think the `../` prefixed part of e.g. `runtime.interpreter.short_path` (or any `.short_path`) is there in lieu of `external/`, and represents "go up one dir from the main repo" (because its referring to some external repo).

chowder:
Yeah it's kinda complicated (don't like it) - I wasn't sure what were all the edge cases here, so writing it generically was the only way I knew how to do it.

rickeylev:
> would something like this suffice

Yes, it should. I pushed a change that basically did that -- compute the relative path between two runfiles-root-relative paths instead of main-repo-relative paths. I kept the relative_path function since I like that it will compute a minimum relative-path.

I also pushed a test to reproduce the problem. It creates a test in a much more deeply nested directory than where the actual interpreter is. As well as some other cleanup.

rickeylev:
> I wasn't sure what were all the edge cases here

I think you got them all, so good job :slightly_smiling_face: . That `short_path` is a main-repo relative path, and gets `../` prepended for external things is the main edge case. The code was a bit confusing because it _said_ it was passing runfiles-root-relative paths around, but it actually wasn't (`interpreter_actual_path` was the plain `short_path` value, which then got fixed up later), and the some later code was papering over that. I went and fixed that.

tests/bootstrap_impls/venv_relative_path_tests.bzl
```diff
@@ -0,0 +1,104 @@
...
+            to = "../c/d",
+        ),
+    ).equals("../../c/d")
+
+    env.expect.that_str(
+        relative_path(
+            from_ = "../a/b",
+            to = "../../c/d",
+        ),
+    ).equals("../../../c/d")
```
aignas, line 44: Is this correct?
Lets say current dir is `/foo/bar/`
Then `../a/b` maps to `/foo/a/b`
Then `../../c/d` maps to `/foo/c/d`
Then the relative path of `/foo/c/d` with respect to `/foo/a/b` is `../../c/d`

rickeylev:
The CI failure is due to https://github.com/bazel-contrib/bazel_features/issues/82, so I'll force-merge past it.

rickeylev:
Ah, actually, I don't think I can force-merge because I'm not an admin. I'll ask about getting admin back. Otherwise, we can just switch the "upcoming bazel" CI to 8rc2 instead of 8rc3

tests/bootstrap_impls/venv_relative_path_tests.bzl
chowder: I don't think this was incorrect.

If current dir was `/foo/bar`, then `../../c/d` maps to `/c/d, not `/foo/c/d`

So from `/foo/a/b` to `/c/d` is indeed `../../../c/d`

chowder:
@rickeylev Thanks for the changes. :)

ewianda:
Apologies. Just noticed that `rules_pkg` is failing 

```
  File "/home/ewianda/.cache/bazel/_bazel_ewianda/00aeeab5df0e41ab9e87aa07d33c6a73/sandbox/linux-sandbox/43/execroot/_main/bazel-out/k8-opt-exec-ST-d57f47055a04/bin/external/rules_pkg~/pkg/private/tar/build_tar.runfiles/_main/../rules_pkg~/pkg/private/tar/build_tar.py", line 486, in <module>
    main()
  File "/home/ewianda/.cache/bazel/_bazel_ewianda/00aeeab5df0e41ab9e87aa07d33c6a73/sandbox/linux-sandbox/43/execroot/_main/bazel-out/k8-opt-exec-ST-d57f47055a04/bin/external/rules_pkg~/pkg/private/tar/build_tar.runfiles/_main/../rules_pkg~/pkg/private/tar/build_tar.py", line 477, in main
    output.add_manifest_entry(entry, file_attributes)
  File "/home/ewianda/.cache/bazel/_bazel_ewianda/00aeeab5df0e41ab9e87aa07d33c6a73/sandbox/linux-sandbox/43/execroot/_main/bazel-out/k8-opt-exec-ST-d57f47055a04/bin/external/rules_pkg~/pkg/private/tar/build_tar.runfiles/_main/../rules_pkg~/pkg/private/tar/build_tar.py", line 342, in add_manifest_entry
    self.add_file(entry.src, entry.dest, **attrs)
  File "/home/ewianda/.cache/bazel/_bazel_ewianda/00aeeab5df0e41ab9e87aa07d33c6a73/sandbox/linux-sandbox/43/execroot/_main/bazel-out/k8-opt-exec-ST-d57f47055a04/bin/external/rules_pkg~/pkg/private/tar/build_tar.runfiles/_main/../rules_pkg~/pkg/private/tar/build_tar.py", line 111, in add_file
    self.tarfile.add_file(
  File "/home/ewianda/.cache/bazel/_bazel_ewianda/00aeeab5df0e41ab9e87aa07d33c6a73/sandbox/linux-sandbox/43/execroot/_main/bazel-out/k8-opt-exec-ST-d57f47055a04/bin/external/rules_pkg~/pkg/private/tar/build_tar.runfiles/rules_pkg~/pkg/private/tar/tar_writer.py", line 251, in add_file
    with open(file_content, 'rb') as f:
         ^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: 'bazel-out/k8-fastbuild/bin/external/rules_pkg~/pkg/private/tar/_build_tar.venv/bin/python3'

```

PR #2440, opened 2024-11-25
fix: upgrade to bazel_features 1.21.0 for Bazel 8 rc3 support

rickeylev:
bazel_features has a bug where it doesn't detect Bazel 8 rc3 correctly. This is 
fixed in release 1.21.0.

From the discussion on https://github.com/bazel-contrib/bazel_features/issues/82,
it's likely this only occurs due to the "rc" notation of the Bazel release, so an
earlier bazel_features version will probably work fine once Bazel 8 is no longer rc.


rickeylev:
The CI failure looks to be because of https://github.com/bazelbuild/stardoc/issues/261

It only affects building our docs (or more precisely, building the stardoc protos), so I'll force-merge past it.

Issue #2441, opened 2024-11-25
sphinx_docs doesn't propagate tags to .run target


tpudlik:
# 🐞 bug report

### Affected Rule

The issue is caused by the rule: `sphinx_docs`

### Is this a regression?

I don't believe so.

### Description

The sphinx_docs macro generates (among others) the `<name>.run` target, but [doesn't pass the tags to it](https://github.com/bazelbuild/rules_python/blob/438b12edca345db7d06021dde97cb3bd6b1bae84/sphinxdocs/private/sphinx.bzl#L190-L193). This means that if you try to use `tags = ["manual"]`, or more generally `--build_tag_filters`, to not build the docs as part of the wildcard build in some circumstances, you will generally fail: the docs will still be built because they're a dependency of the `.run` target.

I think at least the `tags` should be forwarded to `sphinx_run`, but perhaps all `common_kwargs` should be?

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
Linux
  </code>
</pre>

**Output of `bazel version`:**

<pre>
  <code>
1ff5af18d045e8f30a2a0367470db4e8225a785c
  </code>
</pre>

**Rules_python version:**

<pre>
  <code>
0.36.0
  </code>
</pre>

But this issue is present at HEAD, too.

**Anything else relevant?**

I'm happy to send a PR fixing this, but wanted to double-check if forwarding all `common_kwargs` is correct, or if I should be more surgical here. @rickeylev what's your recommendation?

rickeylev:
Forwarding common_kwargs makes sense.

tpudlik:
Sounds good, I'll send the PR.

PR #2442, opened 2024-11-25
fix: Propagate common kwargs to sphinx_run

tpudlik:
This ensures tags like `manual` (and other attributes) are copied into the `{name}.run`
target and properly respected.

Fixes https://github.com/bazelbuild/rules_python/issues/2441


tpudlik:
I took a quick look at the failures, but they don't seem to be related to this PR.

PR #2447, opened 2024-11-27
fix(render_pkg_aliases): correctly render when we have target_platforms set

### LINKED ISSUE #2446: Duplicate entries with platform constraints in requirements.txt regression
If I have a `requirements.in` like this:

```
--extra-index-url https://download.pytorch.org/whl/cpu
torch==2.4.1; platform_machine != "x86_64"
torch==2.4.1+cpu; platform_machine == "x86_64"
```

Which I use to generate a universal requirements.txt with uv using: `uv pip compile --universal requirements.in -o requirements_lock.txt --emit-index-url --no-strip-markers`, I end up with multiple entries for torch in that lockfile:

```
SNIP
torch==2.4.1+cpu ; platform_machine == 'x86_64'
    # via -r requirements.in
torch==2.4.1 ; platform_machine != 'x86_64'
    # via -r requirements.in
SNIP
```

When I attempt to pull those in with:

```bzl
pip.parse(
    hub_name = "pip",
    python_version = "3.11",
)
use_repo(pip, "pip")
```

and use them with:

```bzl

alias(
    name = "foo",
    actual = requirement("torch"),
)
```

It fails with:

```
	File "/private/var/tmp/_bazel_ksmiley/edd497f7df052e6d5ffcf384da842af2/external/rules_python~~pip~pip/torch/BUILD.bazel", line 9, column 35, in <toplevel>
		"//_config:is_python_3.11": "pip_311_torch_linux_x86_64_osx_x86_64_windows_x86_64",
Error: dictionary expression has duplicate key: "//_config:is_python_3.11"
ERROR: /private/var/tmp/_bazel_ksmiley/edd497f7df052e6d5ffcf384da842af2/external/rules_python~~pip~pip/torch/BUILD.bazel: no such target '@@rules_python~~pip~pip//torch:pkg': target 'pkg' not declared in package 'torch' defined by /private/var/tmp/_bazel_ksmiley/edd497f7df052e6d5ffcf384da842af2/external/rules_python~~pip~pip/torch/BUILD.bazel
ERROR: /Users/ksmiley/Downloads/pyrepro/BUILD.bazel:3:6: no such target '@@rules_python~~pip~pip//torch:pkg': target 'pkg' not declared in package 'torch' defined by /private/var/tmp/_bazel_ksmiley/edd497f7df052e6d5ffcf384da842af2/external/rules_python~~pip~pip/torch/BUILD.bazel and referenced by '//:foo'
ERROR: Analysis of target '//:foo' failed; build aborted: Analysis failed
````

This setup temporarily worked after this landed https://github.com/bazelbuild/rules_python/pull/2377 but then regressed in https://github.com/bazelbuild/rules_python/commit/2abca35cf67f60d41485eaab1f53ff7a3c6e9b50

## 🔬 Minimal Reproduction

Build in this project: [pyrepro.zip](https://github.com/user-attachments/files/17926869/pyrepro.zip)

aignas:
It seems that during #2424 I broke the rendering of aliases for the
cases when the target platform is set. This means that the feature for
multiplatform whls when `experimental_index_url` has never worked even
though it was advertised. This ensures that the rendering is happening
correctly and adds extra missing tests.

Whilst at it:
- add an extra test for `pip.parse` handling of env markers that I added
  to ensure that the error is not in the module extension.
- Cleanup unused code - error message constant and the repo arg in
  `whl_config_setting`.

Fixes #2446

### DIFF
--- a/python/private/pypi/extension.bzl
+++ b/python/private/pypi/extension.bzl
@@ -68,6 +68,7 @@ def _create_whl_repos(
         pip_attr,
         whl_overrides,
         simpleapi_cache,
+        evaluate_markers = evaluate_markers,
         available_interpreters = INTERPRETER_LABELS,
         simpleapi_download = simpleapi_download):
     """create all of the whl repositories
@@ -78,6 +79,7 @@ def _create_whl_repos(
         whl_overrides: {type}`dict[str, struct]` - per-wheel overrides.
         simpleapi_cache: {type}`dict` - an opaque dictionary used for caching the results from calling
             SimpleAPI evaluating all of the tag class invocations {bzl:obj}`pip.parse`.
+        evaluate_markers: the function to use to evaluate markers.
         simpleapi_download: Used for testing overrides
         available_interpreters: {type}`dict[str, Label]` The dictionary of available
             interpreters that have been registered using the `python` bzlmod extension.
--- a/python/private/pypi/render_pkg_aliases.bzl
+++ b/python/private/pypi/render_pkg_aliases.bzl
@@ -44,33 +44,17 @@ If the value is missing, then the "default" Python version is being used,
 which has a "null" version value and will not match version constraints.
 """
 
-NO_MATCH_ERROR_MESSAGE_TEMPLATE_V2 = """\
-No matching wheel for current configuration's Python version.
-
-The current build configuration's Python version doesn't match any of the Python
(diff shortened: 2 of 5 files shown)


aignas:
@rickeylev, this is ready for review/land and the only failure is due to stardoc.

rickeylev (approved):

PR #2458, opened 2024-11-30
doc: improve alias error msg in pip hubs

aignas:
Summary:
* Add more flag values that we actually use.
* Point users to online docs.

I still think it is not good enough when somebody needs to debug, but I
am not sure what to add. Maybe we just need user documentation outlining
what each config setting means and what values it is using? Maybe we
should generate the config settings inside the `rules_python` repository
so that we have better visibility into what needs to be documented and
explained.

### DIFF
--- a/python/private/pypi/pkg_aliases.bzl
+++ b/python/private/pypi/pkg_aliases.bzl
@@ -36,22 +36,31 @@ load(":whl_target_platforms.bzl", "whl_target_platforms")
 # it. It is more of an internal consistency check.
 _VERSION_NONE = (0, 0)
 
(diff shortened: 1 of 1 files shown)


rickeylev (approved):
One trick we could do is map default to a error rule. e.g.

```
... = select(default: ":unknown_config")
unknown_config(name = "unknown_config", template = """
Unknown python. Curretn version is {python_version}
"""
)

def unknown_config_impl(ctx):
  fail(ctx.attr.template.format(
    python_version = ctx.attr._python_version_flag[BuildInfo].value
  ))

```

This would allow putting the current python version into the message directly.

PR #2461, opened 2024-12-02
test: Enable non-ABI3 libs linking test for Windows

nicholasjng:
Also explicitly include ABI3 in the name of the other Windows-only testcase, since that is the whole point of the test, and no ABI3-only libs exist on non-Windows platforms.

-----------

Follow-up of #1820.


nicholasjng:
also:

>\# someone with more C + Windows experience will have to figure it out.

couldn't be me, but I am glad it all worked out in the end. :^)

aignas (approved):

PR #2470, opened 2024-12-06
chore: bump the changelog to 1.0

### LINKED ISSUE #1361: Release version 1.0

### LINKED ISSUE #2459: Missed incrementing module compatibility_level after changing pip_parse module extension repo name tag to `hub_name`
### Affected Rule
https://registry.bazel.build/modules/grpc (probably more)

The issue appears in the pip module extension (`@rules_python//python/extensions:pip.bzl`), specifically renaming the `name` tag to `hub_name` without incrementing the rules_python module compatibility level from 1 to 2.

### Is this a regression?
Yes - rules_python didn't increment its compatibility_level when introducing a breaking change from `name` to `hub_name` in the pip extension, causing issues with transitive dependencies that use the older API.

### Description
The core issue is that rules_python made a breaking API change (changing `name` to `hub_name` in the pip extension) without incrementing the compatibility_level parameter from 1 (set in version 0.22) to 2. This causes problems in dependency chains where:

1. A root module uses rules_python 1.0.0-rc2 (which requires `hub_name`)
2. A transitive dependency (in this case, protoc-gen-validate via grpc) uses an older version with `name`
3. Because the compatibility_level wasn't incremented, Bazel doesn't recognize this as a breaking change and attempts to resolve both versions together

This manifests as an error when running `bazel mod tidy`:
```
ERROR: in tag at https://bcr.bazel.build/modules/protoc-gen-validate/1.0.4/MODULE.bazel:64:10, unknown attribute name provided. Type 'bazel help mod' for syntax and help.
```

## 🔬 Minimal Reproduction
Here's a minimal reproduction case demonstrating the compatibility level issue:

1. MODULE.bazel:
```python
bazel_dep(name = "rules_python", version = "1.0.0-rc2")
bazel_dep(name = "protobuf", version = "29.0")
bazel_dep(name = "grpc", version = "1.68.0")

pip.parse(
    hub_name = "pypi",  # New API
    python_version = "3.11",
    requirements_lock = "//python:requirements.txt",
)
use_repo(pip, "pypi")
```

The issue occurs because protoc-gen-validate (introduced as a transitive dependency by grpc) uses the older `name` parameter, but the compatibility_level wasn't incremented to prevent this version mixing.

The complete reproduction case includes additional files (BUILD.bazel, proto definitions, and Python code) which can be found in the linked BCR issue: https://github.com/bazelbuild/bazel-central-registry/issues/3291

## 🔥 Exception or Error
```
bazel mod tidy
ERROR: in tag at https://bcr.bazel.build/modules/protoc-gen-validate/1.0.4/MODULE.bazel:64:10, unknown attribute name provided. Type 'bazel help mod' for syntax and help.
```

## 🌍 Your Environment
**Operating System:**
```
Linux (amd64)
```

```
bazel 7.4.1
```

**Rules_python version:**
```
1.0.0-rc2
```

**Anything else relevant?**

Current workaround is to use `multiple_version_override`:
```python
multiple_version_override(
    module_name = "rules_python",
    versions = [
        "1.0.0-rc2",
        "0.22.0",
    ],
)
```

The proper fix would be to increment the compatibility_level parameter in rules_python. This would prevent Bazel from attempting to mix incompatible versions of the rules in a single build.

aignas:
Fixes #1361
Closes #2459 as won't do

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,27 @@ Unreleased changes template.
 
 {#v0-0-0-changed}
 ### Changed
+* Nothing changed.
+
+{#v0-0-0-fixed}
+### Fixed
+* Nothing fixed.
+
+{#v0-0-0-added}
+### Added
+* Nothing added.
+
+{#v0-0-0-removed}
+### Removed
+* Nothing removed.
+
+{#v1-0-0}
+## Unreleased
+
+[1.0.0]: https://github.com/bazelbuild/rules_python/releases/tag/1.0.0
+
+{#v1-0-0-changed}
+### Changed
 
 **Breaking**:
 * (toolchains) stop exposing config settings in python toolchain alias repos.
@@ -82,7 +103,7 @@ Other changes:
 * (deps) bazel_features 1.21.0; necessary for compatiblity with Bazel 8 rc3
 * (deps) stardoc 0.7.2 to support Bazel 8.
 
-{#v0-0-0-fixed}
+{#v1-0-0-fixed}
 ### Fixed
 * (toolchains) stop depending on `uname` to get the value of the host platform.
 * (pypi): Correctly handle multiple versions of the same package in the requirements
@@ -103,7 +124,7 @@ Other changes:
 * (rules) `py_test` and `py_binary` targets no longer incorrectly remove the 
   first `sys.path` entry when using {obj}`--bootstrap_impl=script`
 
-{#v0-0-0-added}
+{#v1-0-0-added}
 ### Added
 * (gazelle): Parser failures will now be logged to the terminal. Additional
   details can be logged by setting `RULES_PYTHON_GAZELLE_VERBOSE=1`.
@@ -121,7 +142,7 @@ Other changes:
   initialize the interpreter via venv startup hooks.
 * (runfiles) (Bazel 7.4+) Added support for spaces and newlines in runfiles paths
 
-{#v0-0-0-removed}
+{#v1-0-0-removed}
 ### Removed
 * (pypi): Remove `pypi_install_dependencies` macro that has been included in
   {bzl:obj}`py_repositories` for a long time.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,7 +67,7 @@ Unreleased changes template.
 * Nothing removed.
 
 {#v1-0-0}
-## Unreleased
+## [1.0.0] - 2024-12-05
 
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -52,6 +52,27 @@ Unreleased changes template.
...
+{#v0-0-0-added}
+### Added
+* Nothing added.
+
+{#v0-0-0-removed}
+### Removed
+* Nothing removed.
+
+{#v1-0-0}
+## Unreleased
```
rickeylev, line 70:
```suggestion
## [1.0.0] - 2024-12-05
```

rickeylev (approved):

Issue #2410, opened 2024-11-14
PEP621 Support building from `sdist` in a build action


aignas:
Support building from `sdist` as a build action instead of `repository_rule`.

Things needed:
* [ ] A way to use a different `repository` rule that would not do any building with `pip`.
* [ ] Parsing the dependencies from `pyproject.toml` so that the `deps` field is populated at the repository time.
* [ ] A `pip` toolchain or something similar for building packages. Maybe we should have a `toolchain` for each build backend in the https://peps.python.org/pep-0621/
* [ ] A way to specify build dependencies see #1325.
* [ ] How do we build Python extensions?

Initial thoughts:
* It would be good to create an `sdist_archive` rule, which would download a given `sdist` by URL and then would parse `pyproject.toml`. If the `pyproject.toml` is not found or it is dynamic and requires Python to correctly infer the dependencies, we may need to fail at least in the first iteration.
* Using Python in the repository context is OK, but adds a lot of complication in the long run - we should attempt to limit its usage as much as possible.

I don't have time to work on this myself, but writing it down so that an issue for PEP621 based sdist building exists. Feel free to add thoughts here on possible design.

aignas:
Might be related: #1103

groodt:
I think they're quite different if I read this issue.

I think this issue title is misleading, but you can confirm what you mean. 😊

1103 is for package publishers. It's for teams who want to use bazel as a build backend to build a wheel or sdist. Similar to how meson, poetry, flit are build backends.

I read this is issue as an enabler for package consumers. Basically a mechanism that is an improvement to patching. I read this issue as a way to prepare a "build environment" for difficult packages and being able to run with "--no-build-isolation" so that sdist (packages without wheels) can be more easily imported into the Bazel build.

aignas:
Ah yes, you are correct here.

groodt:
I think we should then consider removing PEP 621 from the title. pyproject.toml is only a standard for storing metadata.

aignas:
I would like to restrict us looking only at sdists that comply with PEP621. We will need to somehow get the dependencies for the sdist in the loading phase and then building of the sdist in the build phase can happen correctly. Doing this for sdists that do not comply with PEP621 might be trickier.

WDYT?

groodt:
My 2c:

I love the intention of making "sdist" handling better. Even isolating them into the naughty corner, makes things better for users, because it highlights what is going on when there are failing builds.

[...]

PR #2476, opened 2024-12-06
fix: Add libdir to library search path

shs96c:
We discovered when dealing with libraries such as `psycopg2` that the wheel would attempt to link against `libpython.a`. This fix points the linker at the correct python version being used.


python/private/pypi/attrs.bzl
```diff
@@ -15,6 +15,12 @@
 "common attributes for whl_library and pip_repository"
 
 ATTRS = {
+    "add_libdir_to_library_search_path": attr.bool(
```
aignas, line 18: NOTE: this is enabled for both, WORKSPACE and bzlmod: https://rules-python--2476.org.readthedocs.build/en/2476/api/rules_python/python/extensions/pip.html#pip.parse.add_libdir_to_library_search_path

However, bzlmod will not forward this value to underlying `whl_library` instances, so `extension.bzl` needs to be updated. I think we should only forward this value to `whl_libraries` that are following the legacy `pip` way of downloading deps or downloading an sdist.

python/private/pypi/whl_library.bzl
```diff
@@ -139,11 +139,28 @@ def _parse_optional_attrs(rctx, args, extra_pip_args = None):
...
+    # in order to point the linker at the correct python intepreter.
+    if rctx.attr.add_libdir_to_library_search_path:
+        if "LDFLAGS" in env:
+            fail("Can't set both environment LDFLAGS and add_libdir_to_library_search_path")
+        command = [pypi_repo_utils.resolve_python_interpreter(rctx), "-c", "import sys ; sys.stdout.write('{}/lib'.format(sys.exec_prefix))"]
+        result = rctx.execute(command)
+        if result.return_code != 0:
+            fail("Failed to get LDFLAGS path: command: {}, exit code: {}, stdout: {}, stderr: {}".format(command, result.return_code, result.stdout, result.stderr))
+        libdir = result.stdout
+        env["LDFLAGS"] = "-L{}".format(libdir)
```
aignas, line 158: Should we instead extend the `LDFLAGS` instead of replacing the value?

aignas (commented):
Because this is gated by a feature toggle, it seems to be low risk to include.

rickeylev:
Just want to respond to something the comment says:
```
# [calling the python interpreter in repo-phase]
# is super hacky, but working out something nice is tricky.
# This is in particular needed for psycopg2 which attempts to link libpython.a,
# in order to point the linker at the correct python intepreter.
```

It's not possible to reliably get the correct Python interpreter. This is because the repo phase doesn't know, and can't know, what the build phase is going to do (i.e. what python interpreter will be used). The basic change in this PR is somewhat an improvement, but yeah -- just wanted to point out it'll still be prone to issues. To really fix this, the sdist building needs to move into the build phase.

aignas:
I'll add some ideas about this in #2410.

shs96c:
Is there anything else you'd like me to do for this PR?

python/private/pypi/attrs.bzl
aignas: The dict in [here](https://github.com/bazelbuild/rules_python/blob/main/python/private/pypi/extension.bzl#L204) needs to be updated.

python/private/pypi/whl_library.bzl
aignas: I agree that the code is safe to use as is, but I am just wondering out loud what the correct behaviour should be. If the user needs the `LDFLAGS` for some particular package, then they cannot use this feature. That said we do have a ticket #2410, which I think we should focus instead of trying to get this perfect now.

aignas (approved):

PR #2477, opened 2024-12-06
fix: Make sure wheelmaker uses the default shell env

shs96c:
Clean macOS installs place `python` in `/usr/local/bin`, which is not searched by default when calling `actions.run()`. This caused builds to fail when they were executing python tools on an unmodified macOS host.

### DIFF
--- a/python/private/py_wheel.bzl
+++ b/python/private/py_wheel.bzl
@@ -514,6 +514,8 @@ def _py_wheel_impl(ctx):
         outputs = [outfile, name_file],
         arguments = [args],
         executable = ctx.executable._wheelmaker,
+        # The default shell env is used to better support toolchains that look
+        # up python at runtime using PATH.
         use_default_shell_env = True,
         progress_message = "Building wheel {}".format(ctx.label),
     )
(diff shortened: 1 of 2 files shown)


rickeylev:
I'm guessing you're using the "runtime env" toolchain? i.e. the one that looks up python on PATH at runtime?

I don't see how the wheel maker would fail otherwise. Normally, the underlying py_binary should have an absolute path reference to its interpreter.

This is a bit of a catch 22 because use_default_shell_env introduces OS-specific behavior, which is undesirable. However, the basic way the runtime-env toolchain works is by relying on PATH lookup.

Modifying the runtime env toolchain to check PATH, then check some other common locations seems...well, not sure if that's better or not.

python/private/py_wheel.bzl
```diff
@@ -514,6 +514,7 @@ def _py_wheel_impl(ctx):
         outputs = [outfile, name_file],
         arguments = [args],
         executable = ctx.executable._wheelmaker,
+        use_default_shell_env = True,
```
rickeylev, line 517:
```suggestion
        # The default shell env is used to better support toolchains that look
        # up python at runtime using PATH.
        use_default_shell_env = True,
```

rickeylev (approved):

PR #2478, opened 2024-12-06
fix: Correctly resolve macOS SDK paths

shs96c:
XCode has facilities for accurately telling us where SDKs are installed. This is important to use, particularly when there may be multiple SDKs or versions of XCode installed.


python/private/pypi/whl_library.bzl
```diff
@@ -55,9 +55,31 @@ def _get_xcode_location_cflags(rctx):
...
+            if "productName" in sdk and
+               sdk["productName"] == "macOS" and
+               "darwinos" not in sdk["canonicalName"]
+        ]
+
+        # Now we'll get two entries here (one for internal and another one for public)
+        # It shouldn't matter which one we pick.
+        xcode_sdk_path = potential_sdks[0]["sdkPath"]
+    else:
+        xcode_sdk_path = "{}/SDKs/MacOSX.sdk".format(xcode_root)
```
aignas, line 79: Thinking out loud, is this path absolute? I am wondering if we should have the path in the `rules_python_internal` where we then can include the path by importing it as a constant.

python/private/pypi/whl_library.bzl
```diff
@@ -158,19 +180,22 @@ def _create_repository_execution_environment(rctx, python_interpreter, logger =
...
             entries = rctx.attr._python_path_entries,
         ),
-        _CPPFLAGS: " ".join(cppflags),
     }
 
+    # Gather any available CPPFLAGS values
+    #
+    # We may want to build in an environment without a cc toolchain.
+    # In those cases, we're limited to --donwload-only, but we should respect that here.
+    if not rctx.attr.download_only:
```
aignas, line 194: Let's also exclude when the `filename` is specified and it is a `whl`.

aignas (commented):
LGTM, but please include a CHANGELOG note about the changed behaviour on the MacOS platforms.

python/private/pypi/whl_library.bzl
```diff
@@ -55,9 +55,31 @@ def _get_xcode_location_cflags(rctx):
         # This is a full xcode installation somewhere like /Applications/Xcode13.0.app/Contents/Developer
         # so we need to change the path to to the macos specific tools which are in a different relative
         # path than xcode installed command line tools.
-        xcode_root = "{}/Platforms/MacOSX.platform/Developer".format(xcode_root)
+        xcode_sdks_json = rctx.execute([
```
rickeylev, line 58: Please use `repo_utils.execute_checked` from `//python/private:repo_utils.bzl. We have wrappers that do some book keeping to aid debugging etc. (See a few lines above for example)

python/private/pypi/whl_library.bzl
```diff
@@ -158,19 +180,22 @@ def _create_repository_execution_environment(rctx, python_interpreter, logger =
...
             rctx,
             entries = rctx.attr._python_path_entries,
         ),
-        _CPPFLAGS: " ".join(cppflags),
     }
 
+    # Gather any available CPPFLAGS values
+    #
+    # We may want to build in an environment without a cc toolchain.
+    # In those cases, we're limited to --donwload-only, but we should respect that here.
```
rickeylev, line 193:
```suggestion
    # In those cases, we're limited to --download-only, but we should respect that here.
```

rickeylev (changes requested):

aignas (approved):
LGTM

rickeylev (approved):

PR #2479, opened 2024-12-06
fix: Fixing compile pip requirements when running on an RBE

shs96c:
We have observed that making this change makes builds more likely to run successfully on an RBE.

### DIFF
--- a/python/private/pypi/dependency_resolver/dependency_resolver.py
+++ b/python/private/pypi/dependency_resolver/dependency_resolver.py
@@ -27,9 +27,9 @@ from piptools.scripts.compile import cli
 
 from python.runfiles import runfiles
 
-# Replace the os.replace function with shutil.copy to work around os.replace not being able to
+# Replace the os.replace function with shutil.move to work around os.replace not being able to
 # replace or move files across filesystems.
-os.replace = shutil.copy
+os.replace = shutil.move
 
 # Next, we override the annotation_style_split and annotation_style_line functions to replace the
 # backslashes in the paths with forward slashes. This is so that we can have the same requirements
--- a/python/private/pypi/pip_compile.bzl
+++ b/python/private/pypi/pip_compile.bzl
@@ -21,6 +21,8 @@ make it possible to have multiple tools inside the `pypi` directory
 
 load("//python:defs.bzl", _py_binary = "py_binary", _py_test = "py_test")
 
+_DEFAULT_TAGS = ["no-sandbox", "no-remote-exec", "requires-network"]
+
 def pip_compile(
         name,
         srcs = None,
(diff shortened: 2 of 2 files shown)


python/private/pypi/dependency_resolver/dependency_resolver.py
```diff
@@ -27,9 +27,9 @@
 
 from python.runfiles import runfiles
 
-# Replace the os.replace function with shutil.copy to work around os.replace not being able to
+# Replace the os.replace function with shutil.move to work around os.replace not being able to
 # replace or move files across filesystems.
-os.replace = shutil.copy
+os.replace = shutil.move
```
aignas, line 32: FYI, I am looking at the PR that made the change previously (#1053) and I can't see why one wouldn't want the `shutil.move`. I also can't see looking at the source of `shutil.move` if it would be problematic, it will also try to fallback to `shutil.copy2`, so I think the `shutil.move` is a robust solution here.

python/private/pypi/pip_compile.bzl
```diff
@@ -21,6 +21,8 @@ make it possible to have multiple tools inside the `pypi` directory
 
 load("//python:defs.bzl", _py_binary = "py_binary", _py_test = "py_test")
 
+_DEFAULT_TAGS = ["no-sandbox", "no-remote-exec", "requires-network"]
```
aignas, line 24: I am not sure if we want to drop the `requires-network` tag if someone specifies their own tags, e.g. `manual`. That said, I think I see the intention - allow one to drop the `no-sandbox` and `no-remote-exec` tags.

Please consider how to implement this differently.

aignas (commented):
Thanks for the PR. Please also include a line in the CHANGELOG.md to inform users about the change in behaviour. It LGTM other than the tag handling.

python/private/pypi/pip_compile.bzl
rickeylev: I'm a bit curious, too, why one would want to drop these tags. Doesn't pip_compile more-or-less assume it's being run using `bazel run` (i.e its local, does network stuff, and intends to update local files) ?


@shs96c are you, perhaps using `bazel build` to create requirements using e.g. RBE? (I like that idea)

rickeylev: Hah, totally forgot what the title says: ... when running on RBE

Looking at pip_compile, I'm not even sure how you got that working. I don't see anything that runs the tool as an action.

shs96c: We have some other wrappers around this that we use, but the core thing we do is to call `pip_compile`

shs96c: I'm not sure of a nicer way of handling the tags, but I've added a `remoteable` parameter, and will append tag values based on that. If you'd like this handled in some other way, just LMK and I'll make the change.

shs96c: @aignas, if you've any advice on how to handle tags in the way you'd like, please let me know!

PR #2480, opened 2024-12-06
build: Update doublestar to a version that works with the latest Gazelle

shs96c:


gazelle/deps.bzl
```diff
@@ -14,16 +14,9 @@
...
-    "@bazel_gazelle//:deps.bzl",
-    _go_repository = "go_repository",
-)
+load("@bazel_gazelle//:deps.bzl", "go_repository")
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
-def go_repository(name, **kwargs):
-    if name not in native.existing_rules():
-        _go_repository(name = name, **kwargs)
-
```
aignas, line 26: Why are you removing this?

gazelle/deps.bzl
```diff
@@ -70,8 +63,8 @@ def go_deps():
     go_repository(
         name = "com_github_bmatcuk_doublestar_v4",
         importpath = "github.com/bmatcuk/doublestar/v4",
-        sum = "h1:FH9SifrbvJhnlQpztAx++wlkk70QBf0iBWDwNy7PA4I=",
-        version = "v4.6.1",
+        sum = "h1:fdDeAqgT47acgwd9bd9HxJRDmc9UAmPpc+2m0CXv75Q=",
+        version = "v4.7.1",
```
aignas, line 67: Nit: a CHANGELOG.md entry to tell that a change was done would be great!

aignas (commented):
FYI @dougthor42.

gazelle/deps.bzl
rickeylev: Any idea what the original logic was there for? It looks like the intent was something similar to `maybe(http_archive, ...)` I usually see. Is this like a way that predates using `maybe`?

In any case, yeah, these "define repos for depenencies" things use a conditional definition so that the root workspace can override them. That intent should be kept.

dougthor42:
I ran this locally and didn't see any issues :tada:. The previous comments still need to be addressed though.

gazelle/deps.bzl
shs96c: We've been carrying this patch for a while, and no-one on the team remembers why we removed this, but I agree it seems better to leave this as is. I've reverted that change.

gazelle/deps.bzl
shs96c: Done!

gazelle/deps.bzl
```diff
@@ -14,15 +14,12 @@
...
-    _go_repository = "go_repository",
-)
+load("@bazel_gazelle//:deps.bzl", _go_repository = "go_repository")
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
 def go_repository(name, **kwargs):
-    if name not in native.existing_rules():
-        _go_repository(name = name, **kwargs)
+     if name not in native.existing_rules():
+         _go_repository(name = name, **kwargs)
```
dougthor42, line 22: nit: fix whitespace (4 spaces, not 5)

dougthor42 (commented):
Just a whitespace nit but otherwise LGTM. When that's fixed I'll give the :+1:

gazelle/deps.bzl
shs96c: Good spot. Fixed!

dougthor42 (approved):
:+1:

dougthor42:
@aignas @rickeylev I'm not yet super familiar with the CI pipeline, so I'm not sure if that [buildkite failure](https://buildkite.com/bazel/rules-python-python/builds/10070#0193dee3-c96b-46b2-9331-5efb77f30ecd) is a blocker or not, or if it's just a flake.

shs96c:
I think this was just a flaky run: [could not download Bazel: failed to download bazel: failed to download bazel:](https://buildkite.com/bazel/rules-python-python/builds/10070#0193dee3-c96b-46b2-9331-5efb77f30ecd/68-73).

If someone with BuildKite permissions is around, there's a "retry" button, which will allow that one build to be retried.

Issue #2469, opened 2024-12-05
[Bazel CI] Multiple failures in integration tests 



sgowroji:
CI: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/4313#01939501-8a44-41dc-a7b2-fb29fd9561d3

 Platform: Ubuntu

 Logs: 
```
FAIL: //tests/integration:compile_pip_requirements_workspace_test_bazel_self 
FAIL //tests/integration:ignore_root_user_error_workspace_test_bazel_self
FAIL //tests/integration:pip_parse_workspace_test_bazel_self
FAIL //tests/integration:py_cc_toolchain_registered_workspace_test_bazel_self
``` 
 Culprit: 

 Steps:
  ``` 
 git clone https://github.com/bazelbuild/rules_python
git reset 60da4a39e24381486d3d3bd089d516e3e0cad732  --hard
export USE_BAZEL_VERSION=3615f64a96f98efdb96ad48ed15cb0966d9961be
bazel test ... 
 ``` 
 CC Greenteam @wyverald

rickeylev:
* rules_python (RBE)
`ERROR: While parsing option --toolchain_resolution_debug --remote_download_toplevel: Failed to build filter: value looks like another flag (--remote_download_toplevel). Either escape the value with "\-\-", or pass an explicit value to the flag.`

I don't think this is a rules_python issue? We just use the Bazel CI provided RBE configs.

* rules_python (default, windows)
  * //tests/bootstrap_impls:run_binary_zip_no_test -- `AssertionError: Cannot find .runfiles directory for c:\b\5bw2dcqy\execroot\_main\bazel-out\x64_windows-fastbuild\bin\tests\bootstrap_impls\run_binary_zip_no_test.exe.runfiles/_main/tests/bootstrap_impls/_run_binary_zip_no_test_bin
Test case failed: using no runfiles env vars`

No idea as to cause.

* rules_python (default, workspace) (mac, linux)
`rules_python/python/private/reexports.bzl:37:17: name 'PyInfo' is not defined (did you mean 'CcInfo'?)` (same with PyRuntimeInfo)

I thought I had fixed this previously? Well, I'll have to look again.

* examples/pip_repository_annotations
* examples/py_proto_library
* examples/pip_parse_vendored
* examples/build_file_generation

ERROR: error loading package 'external': Both --enable_bzlmod and --enable_workspace are disabled, but one of them must be enabled to fetch external dependencies.`

[...]

PR #2481, opened 2024-12-06
fix: normalize argv0 so runfiles root can be found on windows with bazel 9

rickeylev:
When the shell test invokes the python binary, it uses a combination of forward slashes and
backslashes. Under Bazel 9, that mixture of slashes is preserved. This later breaks a regex
that looks for the OS-specific path separator.

To fix, normalize forward slashes to the OS path separator.

Oddly, it's not Bazel that is passing the mixture of slashes (it's the shell), but behavior seems to
vary based on which version of Bazel is used.

Along the way, copy the nicer `print_verbose` function from the stage2 bootstrap into the old
bootstrap. It prints debug information in a nicer format.

### DIFF
--- a/python/private/python_bootstrap_template.txt
+++ b/python/private/python_bootstrap_template.txt
@@ -89,9 +89,28 @@ def FindPythonBinary(module_space):
   """Finds the real Python binary if it's not a normal absolute path."""
   return FindBinary(module_space, PYTHON_BINARY)
 
-def PrintVerbose(*args):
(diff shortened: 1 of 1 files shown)


aignas (approved):

aignas:
Do we need to have `CHANGELOG.md` notes for this? I think it is a very low level change that will probably not affect anyone though.

rickeylev:
I'm ok with omitting a changelog entry for it. It's for an unreleased version of bazel (which I'm not sure can be used with rules_python right now), and the bug report didn't come from a user, so I doubt anybody has actually encountered it.

PR #2483, opened 2024-12-07
chore: load specific bzl files instead of generic defs.bzl

rickeylev:
Update code and examples to load the object-specific bzl files instead of the
generic `defs.bzl`. This is mostly for code hygiene, but came out of trying to diagnose
why Bazel 9 workspace builds kept erroing with defs.bzl somehow related. Removing 
the internal usages of defs.bzl doesn't seem to fully fix it, but does seem to eliminate
some errors, make some progress, and narrow down what's going on.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -76,7 +76,7 @@ Once you've imported the rule set using either Bzlmod or WORKSPACE, you can then
 load the core rules in your `BUILD` files with the following:
 
 ```starlark
-load("@rules_python//python:defs.bzl", "py_binary")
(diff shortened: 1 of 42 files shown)


aignas (approved):
Should we make the `defs.bzl` deprecated? We could add a deprecation notice that they will be removed in the upcoming major version of `rules_python`?

rickeylev:
I'd be fine with "soft" deprecating it. It's been around for a long time and there's quite a few usages of it out there. Re-exporting the symbols is very low maintenance, so I don't want to remove it without a long period.

PR #2485, opened 2024-12-07
fix: define rules_python_internal earlier so Bazel 9 doesn't try to use PyInfo et al builtins

rickeylev:
For Bazel 9 workspace builds, if `@rules_python_internal` isn't defined early enough,
an earlier version of `@rules_python` gets defined and the logic to not use the builtin
PyInfo et al symbols doesn't occur. Since Bazel 9 doesn't have these builtins, an error
occurs.

This seems to only happen if the main module is rules_python. The example workspaces don't
see to have an issue. I'm not sure why, but it seems similar to the behavior where
autoloading is disabled for specific repos, rules_python among them.

To fix, move the `@rules_python_internal` repo definition to be earlier in the WORKSPACE
processing. With that repo defined, the conditional logic takes place, and things seem
to be happy.


aignas (approved):
LGTM. It seems that there is a stray load in `internal_dev_setup.bzl`.

PR #2487, opened 2024-12-09
chore: replace rules_proto with com_google_protobuf

rickeylev:
rules_proto is deprecated and recent versions simply forward onto com_google_protobuf.
Older versions (e.g. 6.x used by us today), however, use the rules_proto or native
(Bazel builtin) implementation. When those older versions are used with Bazel 9,
which has removed various proto things, errors occur.

To fix, switch to using com_google_protobuf directly. More recent versions of rules_proto
just forward onto com_google_protobuf anyways, so this just removes the extra dependency
and having to deal with WORKSPACE setup.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -21,6 +21,10 @@ load("//:internal_dev_deps.bzl", "rules_python_internal_deps")
 
 rules_python_internal_deps()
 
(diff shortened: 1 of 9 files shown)


aignas (approved):

aignas (commented):
We don't need to modify our MODULE.bazel?

PR #2492, opened 2024-12-10
ci: fix CI after bazel 8 release

### LINKED ISSUE #2378: Support Bazel 8

aignas:
Summary:
* Remove the old `7.x` config from all `.bazelrc`.
* Drop bazel 6 from example testing/support. The `.bazelrc`
  for enabling `WORKSPACE` cannot work across bazel 6,7,8.
* Add missing `BUILD.bazel` file for integration tests.
* Remove an integration test runner for `bazel 6.x`.
* RBE test for bazel 8 is still not working. 
* bump `rules_java` for internal WORKSPACE dependencies to
  a version that supports `8.x`.
* start running bazel-in-bazel integration tests using bazel `8.x`.
* until bazel-contrib/rules_bazel_integration_test#414 is merged,
  we need to use bazel 7 for the delete_packagese pre-commit hook,
  not sure about how to track this as the `cgrindel/bazel-lib` needs
  a new version.

Fixes #2378


aignas:
Didn't have time to finish this. Tried doing:
* Bumping minimum tested version of bazel to 7.4.1
* Remove the `7.x` config from bazelrc
* Add `--enable_workspace` to workspace only examples (actually, should we move them to internal or somewhere, so that they don't advertise WORKSPACE support anymore?).

The build is failing with `rules_java` and it seems that it might be defined somewhere early in the process for the WORKSPACE builds.

aignas:
Reverted the changes and trying a more minimal stop-gap solution.

rickeylev:
> The build is failing with rules_java ...
> ERROR: Error computing the main repository mapping: cannot load '@@rules_java//java:rules_java_deps.bzl': no such file

Ugh, that error. I've seen that a ton while trying to get Bazel 9 workspace working. From what I've been able to tell, it means rules_java needs to be upgraded.

Ah, I see we have CI coverage for bazel 8 bzlmod, but not workspace :facepalm: .

rickeylev:
This is what I have in some WIP stuff for bazel 9:

```
    http_archive(
        name = "rules_java",
        urls = [
            "https://github.com/bazelbuild/rules_java/releases/download/8.6.2/rules_java-8.6.2.tar.gz",
        ],
        sha256 = "a64ab04616e76a448c2c2d8165d836f0d2fb0906200d0b7c7376f46dd62e59cc",
    )
```

There may also be a change to the `rules_java_dependencies()` call in internal_dev_setup.bzl needed --  `load("@rules_java//java:deps.bzl", "rules_java_dependencies")` might be needed. (not entirely sure)

WORKSPACE
```diff
@@ -21,6 +21,21 @@ load("//:internal_dev_deps.bzl", "rules_python_internal_deps")
 
 rules_python_internal_deps()
 
+load("@rules_java//java:rules_java_deps.bzl", "rules_java_dependencies")
```
aignas, line 24: NOTE: I could not get this setup to work from `internal_dev_setup.bzl`.

rickeylev (approved):

rickeylev (approved):

PR #2495, opened 2024-12-11
fix: Strip leading slash for repo url

amartani:


rickeylev (approved):

rickeylev:
Thanks for the PR! CI failures look unrelated, so I've approved and queued it. We'll have the CI stuff fixed in the next day or so.

PR #2496, opened 2024-12-11
fix(gazelle): Support parsing files that use Python3.12 PEP 695 (Type Parameter Syntax) by using dougthor42's fork of go-tree-sitter

### LINKED ISSUE #2396: Gazelle seems to not like PEP 695 – Type Parameter Syntax

dougthor42:
Replaces #2413.

Fixes #2396.

This updates the `go-tree-sitter` dependency to use my fork that includes `BUILD.bazel` files. Specifically, the `BUILD.bazel` files in the fork include references to top-level code like `array.h` which the original Gazelle-generated files for `go-tree-sitter` were not able to handle. I also include the test cases that @maffoo created in #2413 and verified that they (a) fail before the fix and (b) pass after the fix.

The fork is: https://github.com/dougthor42/go-tree-sitter

The branch that includes all changes is: https://github.com/dougthor42/go-tree-sitter/tree/for-rules-python-gazelle-plugin

A couple notes:
+ I have a PR open to get `go-tree-sitter` into BCR [here](https://github.com/bazelbuild/bazel-central-registry/pull/3366). However:
    1. I'm having trouble getting tests to pass and to get things running locally to validate it
    2. Using BCR would not fix things for people who still use WORKSPACE (right?)
+ The fork is _mostly_ [autogenerated BUILD.bazel files from gazelle](https://github.com/smacker/go-tree-sitter/commit/cfa9bdf58beb30159807c1846e1c761bce1e6158) but also contains:
    + [manual updates so that build files reference the toplevel `array.h` and other files](https://github.com/smacker/go-tree-sitter/commit/63f89cd3d471e7e81b51dc7e7205b201fed70fc1)
    + [replace all `smacker` with `dougthor42` so that `go build` works](https://github.com/smacker/go-tree-sitter/commit/8a73cbdb0e9b5febc314f8b9b4241aa01cc86bd0)
    + various other more minor things.
+ I was unable to get `go mod edit -replace` to work, so I've just manually updated `go.mod` and whatnot everywhere. If someone with more go knowledge has a suggestion I'm happy to hear it.


fmeum:
If you actually do maintain a fork with proper build files, we shouldn't need a BCR entry. Gazelle does preserve build files if they exist.

aignas (approved):
Thanks for adding tests!

Whilst I would prefer it to be in BCR, I understand your desire to also cater for `WORKSPACE` users.

dougthor42:
> Whilst I would prefer it to be in BCR, I understand your desire to also cater for `WORKSPACE` users.

I mean, I wouldn't say I _desire_ to cater to WORKSPACE users, haha. If we just dropped WORKSPACE support altogether I'd be super happy but I know that wouldn't go over well with some people.

When we _do_ drop WORKSPACE support, then the BCR-based option is indeed the better solution.

PR #2504, opened 2024-12-14
fix(pypi): Fix use_hub_alias_dependencies with WORKSPACE

gholms:
The code path pip_parse follows when using a WORKSPACE file with
use_hub_alias_dependencies enabled forgets to pass requirement cycles
along to alias creation, leading to the _groups package never being
created and aliases skipping them.  Requirement cycles are just ignored
entirely.  In this patch we attempt to fix that so grouping works more
or less the same way as it does under bzlmod with that flag enabled.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,6 +67,9 @@ Unreleased changes template.
   Fixes [#2450](https://github.com/bazelbuild/rules_python/issues/2450).
 * (gazelle) Gazelle will now correctly parse Python3.12 files that use [PEP 695 Type
   Parameter Syntax][pep-695]. (#2396)
+* (pypi) Using {bzl:obj}`pip_parse.experimental_requirement_cycles` and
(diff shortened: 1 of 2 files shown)


aignas:
Thanks, tested in the examples folder by building `sphinx` with the following patch:
```
$ gd
diff --git a/examples/pip_parse/WORKSPACE b/examples/pip_parse/WORKSPACE
index bb4714d9..17b1288d 100644
--- a/examples/pip_parse/WORKSPACE
+++ b/examples/pip_parse/WORKSPACE
@@ -24,6 +24,8 @@ pip_parse(
     # environment = {"HTTPS_PROXY": "http://my.proxy.fun/"},
     name = "pypi",

+    use_hub_alias_dependencies = False,
+
     # Requirement groups allow Bazel to tolerate PyPi cycles by putting dependencies
     # which are known to form cycles into groups together.
     experimental_requirement_cycles = {
```

It seems that is working well in both cases.

aignas (approved):

PR #2512, opened 2024-12-18
docs: Fix toolchain implementation file name

nicholasjng:
The code example comment uses `toolchain_impls` (extra s), while the target references
use `toolchain_impl` (no s). This made me go back and double-check when reading the
custom toolchain example (especially the definitions in L425ff).

Remove the extra "s" in the comment so it matches what the target references are.

### DIFF
--- a/docs/toolchains.md
+++ b/docs/toolchains.md
@@ -376,7 +376,7 @@ Here, we show an example for a semi-complicated toolchain suite, one that is:
 Defining toolchains for this might look something like this:
 
 ```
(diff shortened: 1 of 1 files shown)


rickeylev:
Thanks! Nope, not pedantic at all -- it's confusing if the file names aren't consistent.

re: visual bars: I'd be +1 on that. There's a lot of content in the example, so easy to get lost in what file something is for.

rickeylev (approved):

PR #2514, opened 2024-12-19
Updated pip and packaging versions to work with free-threading packages

vfdev-5:
We had an issue to install jaxlib with bazel when running the following command (using rules_python v0.39):
```bash
          bazel test \
              --repo_env=HERMETIC_PYTHON_VERSION=3.13-ft \
              --repo_env=JAX_NUM_GENERATED_CASES=$JAX_NUM_GENERATED_CASES \
              --repo_env=JAX_ENABLE_X64=$JAX_ENABLE_X64 \
              --repo_env=JAX_SKIP_SLOW_TESTS=$JAX_SKIP_SLOW_TESTS \
              --repo_env=PYTHON_GIL=$PYTHON_GIL \
              --repo_env=TSAN_OPTIONS="halt_on_error=1" \
              --//jax:build_jaxlib=false \
              --nocache_test_results \
              --test_output=all \
              //tests:cpu_tests
```

According to @vam-google, this was due to old pip/packaging versions. We updated them and this helped to make work the whole building/testing pipeline: https://github.com/jax-ml/jax/pull/24898
So, we would like to upstream the patch: https://github.com/jax-ml/jax/pull/24898/files#diff-e3dc8d7d2bf5d057f95b86bcff7360b6c99fa1f458882fd112b58da4aceb53e4

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,6 +65,8 @@ Unreleased changes template.
 * (rules) The version-aware rules have been folded into the base rules and
   the version-aware rules are now simply aliases for the base rules. The
   `python_version` attribute is still used to specify the Python version.
+* (pypi) Updated versions of packages: `pip` to 24.3.1 and
+  `packaging` to 24.2.
 
 {#v0-0-0-deprecations}
 #### Deprecations
--- a/python/private/pypi/dependency_resolver/dependency_resolver.py
+++ b/python/private/pypi/dependency_resolver/dependency_resolver.py
@@ -16,6 +16,7 @@
 
 import atexit
 import os
+import re
 import shutil
 import sys
 from pathlib import Path
@@ -117,7 +118,6 @@ def main(
     absolute_path_prefix = resolved_requirements_file[
         : -(len(requirements_file) - len(repository_prefix))
     ]
-
     # As srcs might contain references to generated files we want to
     # use the runfiles file first. Thus, we need to compute the relative path
     # from the execution root.
@@ -162,11 +162,18 @@ def main(
     argv.append(
         f"--output-file={requirements_file_relative if UPDATE else requirements_out}"
     )
-    argv.extend(
+    src_files = [
         (src_relative if Path(src_relative).exists() else resolved_src)
         for src_relative, resolved_src in zip(srcs_relative, resolved_srcs)
-    )
+    ]
+    argv.extend(src_files)
     argv.extend(extra_args)
+
+    # Replace in the output lock file
+    # the lines like: # via -r /absolute/path/to/<requirements_file>
+    # with: # via -r <requirements_file>
+    # For Windows, we should explicitly call .as_posix() to convert \\ -> /
+    absolute_src_prefixes = [Path(src).absolute().parent.as_posix() + "/" for src in src_files]
 
     if UPDATE:
         print("Updating " + requirements_file_relative)
@@ -185,14 +192,14 @@ def main(
             # and we should copy the updated requirements back to the source tree.
             if not absolute_output_file.samefile(requirements_file_tree):
                 atexit.register(
-                    lambda: shutil.copy(
-                        absolute_output_file, requirements_file_tree
-                    )
+                    lambda: shutil.copy(absolute_output_file, requirements_file_tree)
                 )
-        cli(argv, standalone_mode = False)
+        cli(argv, standalone_mode=False)
         requirements_file_relative_path = Path(requirements_file_relative)
         content = requirements_file_relative_path.read_text()
         content = content.replace(absolute_path_prefix, "")
+        for absolute_src_prefix in absolute_src_prefixes:
+            content = content.replace(absolute_src_prefix, "")
         requirements_file_relative_path.write_text(content)
     else:
         # cli will exit(0) on success
@@ -214,6 +221,15 @@ def main(
                 golden = open(_locate(bazel_runfiles, requirements_file)).readlines()
                 out = open(requirements_out).readlines()
                 out = [line.replace(absolute_path_prefix, "") for line in out]
+
+                def replace_via_minus_r(line):
+                    if "# via -r " in line:
+                        for absolute_src_prefix in absolute_src_prefixes:
+                            line = line.replace(absolute_src_prefix, "")
+                        return line
+                    return line
+
+                out = [replace_via_minus_r(line) for line in out]
                 if golden != out:
                     import difflib
 
--- a/python/private/pypi/deps.bzl
+++ b/python/private/pypi/deps.bzl
@@ -51,8 +51,8 @@ _RULE_DEPS = [
     ),
     (
         "pypi__packaging",
-        "https://files.pythonhosted.org/packages/49/df/1fceb2f8900f8639e278b056416d49134fb8d84c5942ffaa01ad34782422/packaging-24.0-py3-none-any.whl",
-        "2ddfb553fdf02fb784c234c7ba6ccc288296ceabec964ad2eae3777778130bc5",
+        "https://files.pythonhosted.org/packages/88/ef/eb23f262cca3c0c4eb7ab1933c3b1f03d021f2c48f54763065b6f0e321be/packaging-24.2-py3-none-any.whl",
+        "09abb1bccd265c01f4a3aa3f7a7db064b36514d2cba19a2f694fe6150451a759",
     ),
     (
         "pypi__pep517",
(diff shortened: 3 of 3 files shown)


vfdev-5:
@aignas can you help with updating the tests to take new versions into account?

aignas:
Yeah, not sure about the fix, but it does seem that the suggested change is breaking the reproducibility of the compiled requirement files. I am wondering though if this is just a test setup issue.

Unfortunately, I have no time to work on this.

tests/integration/compile_pip_requirements/requirements_lock.txt
```diff
@@ -9,8 +9,6 @@
 pip==22.3.1 \
     --hash=sha256:65fd48317359f3af8e593943e6ae1506b66325085ea64b706a998c6e83eeaf38 \
     --hash=sha256:908c78e6bc29b676ede1c4d57981d490cb892eb45cd8c214ab6298125119e077
-    # via -r requirements.in
```
vfdev-5, line 12: The recent pip==24.3.1 (vs 24.0) now adds the annotation as an absolute path vs relative path previously:
```diff
- # via -r requirements.in
+ # via -r /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/5b986f10e8407a028a0eb165b28c6c27/execroot/_main/bazel-out/k8-fastbuild/bin/requirements.update.runfiles/_main/requirements.in
```

The idea is to remove annotations in the tests only to match `requirements_lock.txt` on the reproducibility process.

aignas: If the CI of ours needs to be fixed, that means that this PR will break users that depend on annotations and asking all of the users add `--no-annotate` to their setups is not something that we can do here.

vfdev-5: Do you have any ideas how to make this less BC breaking?
On the other hand, I think it would be a tiny BC breaking change in rules python 1.0 compared to others. This would also enable the use of more recent tools instead of being stuck with old versions.

aignas: FYI, this was handled in the past, but the dependency update broke.

See https://github.com/bazelbuild/rules_python/blob/main/python/private/pypi/dependency_resolver/dependency_resolver.py#L195

python/private/pypi/dependency_resolver/dependency_resolver.py
```diff
@@ -185,14 +186,23 @@ def main(
...
+        # with: # via -r requirements_extra.in
+        pattern = re.compile(r"(# via -r )(/.+/requirements.*\.in)")
+
+        def replace_abspath_with_name(match):
+            assert match is not None
+            # match[1] is "# via -r "
+            # match[2] is "/absolute/path/to/requirements*.in"
+            return match[1] + Path(match[2]).name
+
+        content = pattern.sub(replace_abspath_with_name, content)
```
vfdev-5, line 205: @aignas is this something like what you thought about in this message: https://github.com/bazelbuild/rules_python/pull/2514#discussion_r1895346149 ?

How about `UPDATE == False` case?

aignas: I was actually thinking that the `absolute_path_prefix` replacement stopped working due to some reason and that we should fix the `absolute_path_prefix` matching, but it seems that in reality it is something else.

Would it be possible to have the `absolute_path_prefix` contain the actual absolute path prefix that works here?

vfdev-5: I'm not totally sure what does `absolute_path_prefix` replacement, running the test
```
cd tests/integration/compile_pip_requirements

echo '' > requirements_lock.txt && \
    bazel run //:requirements.update && \
    git diff --exit-code
```
**on `main`**, I have the following values in the vars:
```python
resolved_requirements_file: /tmp/jax/rules_python/tests/integration/compile_pip_requirements/requirements_lock.txt
absolute_path_prefix: /tmp/jax/rules_python/tests/integration/compile_pip_requirements/ 
```
However, requirements lock content does not contain these paths (even before `content = content.replace(absolute_path_prefix, "")` line):
```
# This file is autogenerated by pip-compile with Python 3.9                                                
# by the following command:                  
#                                                                                                          
#    bazel run //:requirements.update                                                                      
#                                                                                                          
                                                                                                           
# The following packages are considered to be unsafe in a requirements file:                               
pip==22.3.1 \
    --hash=sha256:65fd48317359f3af8e593943e6ae1506b66325085ea64b706a998c6e83eeaf38 \
    --hash=sha256:908c78e6bc29b676ede1c4d57981d490cb892eb45cd8c214ab6298125119e077
    # via -r requirements.in
setuptools==65.6.3 \
    --hash=sha256:57f6f22bde4e042978bcd50176fdb381d7c21a9efa4041202288d3737a0c6a54 \
    --hash=sha256:a7620757bf984b58deaf32fc8a4577a9bbc0850cf92c20e1ce41c38c19e5fb75
    # via -r requirements_extra.in
```
So, nothing is replaced with `content = content.replace(absolute_path_prefix, "")`.

If you can share more details on what was the intention of `content = content.replace(absolute_path_prefix, "")`, I can try to figure out whether this could work with never pip.

aignas: The thing is that on Windows things work a little bit differently and it could be that this is only needed there. I can't fully remember the exact reasons for the code, but I do think that trying to use that might be useful.

vfdev-5: Let me recap the problem.

New pip==24.3.1 (vs 24.0) now adds the annotation as an absolute path vs relative path previously:
```diff
- # via -r requirements.in
+ # via -r /root/.cache/bazel/_bazel_root/a76a5e2851b1f2cc4ad9886a6a6aacc6/execroot/_main/bazel-out/k8-fastbuild/bin/requirements.update.runfiles/_main/requirements.in
```

In the test, for example, `tests/integration/compile_pip_requirements`, we use piptools compile as following:
```python
argv = ['--output-file=requirements_lock.txt', 'requirements.txt', '--resolver=backtracking', '--allow-unsafe', '--generate-hashes']
cli(argv)
```
Files `requirements.txt` and `requirements_lock.txt` are and will be located in `/root/.cache/bazel/_bazel_root/a76a5e2851b1f2cc4ad9886a6a6aacc6/execroot/_main/bazel-out/k8-fastbuild/bin/requirements.update.runfiles/_main/` folder.

The content of `requirements.txt` is:
```txt
-r requirements.in
```
After the execution, pip produces the lock file with the following content:
```txt
#
# This file is autogenerated by pip-compile with Python 3.9
# by the following command:
#
#    bazel run //:requirements.update
#

# The following packages are considered to be unsafe in a requirements file:
pip==22.3.1 \
    --hash=sha256:65fd48317359f3af8e593943e6ae1506b66325085ea64b706a998c6e83eeaf38 \
    --hash=sha256:908c78e6bc29b676ede1c4d57981d490cb892eb45cd8c214ab6298125119e077
    # via -r /root/.cache/bazel/_bazel_root/a76a5e2851b1f2cc4ad9886a6a6aacc6/execroot/_main/bazel-out/k8-fastbuild/bin/requirements.update.runfiles/_main/requirements.in
setuptools==65.6.3 \
    --hash=sha256:57f6f22bde4e042978bcd50176fdb381d7c21a9efa4041202288d3737a0c6a54 \
    --hash=sha256:a7620757bf984b58deaf32fc8a4577a9bbc0850cf92c20e1ce41c38c19e5fb75
    # via -r /root/.cache/bazel/_bazel_root/a76a5e2851b1f2cc4ad9886a6a6aacc6/execroot/_main/bazel-out/k8-fastbuild/bin/requirements.update.runfiles/_main/requirements_extra.in
```
And the test expects to see 
```
...
# via -r requirements.in
...
# via -r requirements_extra.in
```

On my linux machine, `absolute_path_prefix` is set to `/tmp/jax/rules_python/tests/integration/compile_pip_requirements/` where the test source code is located. It is computed from `resolved_requirements_file` which is again points to the original requirements file: `/tmp/jax/rules_python/tests/integration/compile_pip_requirements/requirements_lock.txt`

I'm not sure if we can efficiently use `absolute_path_prefix` without breaking something else (like test on Windows).

We can probably fetch the absolute path of the source `requirement.txt` file <-> `argv[1]` and remove it from all lines `# via -r ...`.  What do you think ?

Concerning my code in this PR, it may not always work as source file can be anything and not only `requirements*.in` as I'm regexping now...

aignas: Thanks for the recap.

I guess we should ensure that neither source absolute paths nor the output absolute paths are in the requirements file. I think a solution that does not rely on `regexp` matching the filename could be good enough.

aignas:
I think the current solution might be good enough.

aignas (approved):
Thanks for the fix.

aignas:
Could you please add a line in the Changelog file so that I could merge the PR?

CHANGELOG.md
```diff
@@ -65,6 +65,8 @@ Unreleased changes template.
 * (rules) The version-aware rules have been folded into the base rules and
   the version-aware rules are now simply aliases for the base rules. The
   `python_version` attribute is still used to specify the Python version.
+* (pypi) Updated versions of packages: `pip` to 24.3.1 and
```
vfdev-5, line 68: @aignas I drafted this change in the Changelog. Please confirm if this is OK to you or suggest how to improve it. Thanks!

vonschultz:
According to my `git bisect` this pull request has reintroduced bug #908.

PR #2519, opened 2024-12-22
refactor(pypi): A better error message when the wheel select hits no_match

### LINKED ISSUE #2466: [doc] Print a better error message when the hub repo aliases do not match config settings
Followup from #2458

The general design would be to create our own `no_match_rule` which could be instantiated as follows:
```starlark
def compatible_settings(settings):
    """Return a list of strings"""
    ret = []
    for setting in settings:
        ret += select({setting: [setting], "//conditions:default": []})

    return ret

no_match(
    name = "no_match",
    compatible_settings = compatible_settings(ALL_SETTINGS),
    # The flags that we use for selecting the wheels
    python_version = "@rules_python//python/config_settings:python_version",
    ...
)
```

The impl of the rule could be as simple as:
```starlark
def _impl(ctx):
    flags = {
        name: getattr(ctx.attr, name)[Provider].value
        for name in ALL_FLAG_NAMES
    }
    msg = _TEMPLATE.format(
        compatible_settings = render.indent("\n".join(ctx.attr.compatible_settings))
        flags = render.indent("\n".join(flags))
    )
    fail(msg)
```

And it is wired inside the `pkg_aliases` rule by doing:
```starlark
        actual = {
            "//conditions:default": "//_config:no_match"
        }
```

This would trigger the `_impl` rule implementation above and it would fail building the target.

Definition of done:
* We can see compatible config settings with the current configuration.
* We can see the flag values of the current configuration.
* If we do `bazel build //_config/...`, the build should not fail. Add a `build_test` in the `//tests/pypi/integration`.
* We have an analysis test that checks the failure and (bonus points) checks that the error message is templated with the correct compatible settings. Add an analysis test in `//tests/pypi/pkg_aliases`.

Part of #260.

aignas:
With this change we get the current values of the python configuration
values printed in addition to the message printed previously. This
should help us advise users who don't have their builds configured
correctly.

We are adding an extra `build_setting` which we can set in order to get
an error message instead of a `DEBUG` warning. This has been documented
as part of our config settings and in the `no_match_error` in the
`select` statement.

Example output now
```console
$ bazel cquery --@rules_python//python/config_settings:python_version=3.12 @dev_pip//sphinx
WARNING: Build option --//python/config_settings:current_config has changed, discarding analysis cache (this can be expensive, see https://bazel.build/advanced/performance/iteration-speed).
DEBUG: /home/aignas/src/github/aignas/rules_python/python/private/config_settings.bzl:200:14: The current configuration rules_python config flags is:
    @@//python/config_settings:pip_whl: "auto"
    @@//python/config_settings:pip_whl_glibc_version: ""
    @@//python/config_settings:pip_whl_muslc_version: ""
    @@//python/config_settings:pip_whl_osx_arch: "arch"
    @@//python/config_settings:pip_whl_osx_version: ""
    @@//python/config_settings:py_freethreaded: "no"
    @@//python/config_settings:py_linux_libc: "glibc"
    @@//python/config_settings:python_version: "3.12"

If the value is missing, then the default value is being used, see documentation:
https://rules-python.readthedocs.io/en/latest/api/rules_python/python/config_settings
ERROR: /home/aignas/.cache/bazel/_bazel_aignas/6f0de8c9128ee8d5dbf27ba6dcc48bdd/external/+pip+dev_pip/sphinx/BUILD.bazel:6:12: configurable attribute "actual" in @@+pip+dev_pip//sphinx:_no_match_error doesn't match this configuration: No matching wheel for current configuration's Python version.

The current build configuration's settings do not match any of the Python wheels available for this distribution. This distribution supports the following configuration settings:
    //_config:is_cp3.11_py3_none_any
    //_config:is_cp3.13_py3_none_any

To determine the current build configuration flag values, see the debug message above or re-run the build with:
    --@@//python/config_settings:current_config=fail

For other configuration, run:

This instance of @@+pip+dev_pip//sphinx:_no_match_error has configuration identifier 29ffcf8. To inspect its configuration, run: bazel config 29ffcf8.

ERROR: Analysis of target '@@+pip+dev_pip//sphinx:sphinx' failed; build aborted: Analysis failed
INFO: Elapsed time: 0.100s
INFO: 0 processes.
```

Fixes #2466

### DIFF
--- a/docs/api/rules_python/python/config_settings/index.md
+++ b/docs/api/rules_python/python/config_settings/index.md
@@ -248,7 +248,7 @@ Fail the build if the current build configuration does not match the
 Values:
 * `fail`: Will fail in the build action ensuring that we get the error
   message no matter the action cache.
-* ``: The default value, that will just print a warning.
+* ``: (empty string) The default value, that will just print a warning.
 
 :::{seealso}
 {obj}`pip.parse`
--- a/docs/api/rules_python/python/config_settings/index.md
+++ b/docs/api/rules_python/python/config_settings/index.md
@@ -240,3 +240,21 @@ instead.
 :::
 
 ::::
+
+::::{bzl:flag} current_config
+Fail the build if the current build configuration does not match the
+{obj}`pip.parse` defined wheels.
+
+Values:
+* `fail`: Will fail in the build action ensuring that we get the error
+  message no matter the action cache.
+* ``: The default value, that will just print a warning.
+
+:::{seealso}
+{obj}`pip.parse`
+:::
+
+:::{versionadded} 1.1.0
+:::
+
+::::
--- a/python/private/config_settings.bzl
+++ b/python/private/config_settings.bzl
@@ -17,12 +17,21 @@
 
 load("@bazel_skylib//lib:selects.bzl", "selects")
 load("@bazel_skylib//rules:common_settings.bzl", "BuildSettingInfo")
+load("//python/private:text_util.bzl", "render")
 load(":semver.bzl", "semver")
 
 _PYTHON_VERSION_FLAG = Label("//python/config_settings:python_version")
 _PYTHON_VERSION_MAJOR_MINOR_FLAG = Label("//python/config_settings:python_version_major_minor")
 
-def construct_config_settings(*, name, default_version, versions, minor_mapping):  # buildifier: disable=function-docstring
+_DEBUG_ENV_MESSAGE_TEMPLATE = """\
+The current configuration rules_python config flags is:
+    {flags}
+
+If the value is missing, then the default value is being used, see documentation:
+{docs_url}/python/config_settings
+"""
+
+def construct_config_settings(*, name, default_version, versions, minor_mapping, documented_flags):  # buildifier: disable=function-docstring
     """Create a 'python_version' config flag and construct all config settings used in rules_python.
 
     This mainly includes the targets that are used in the toolchain and pip hub
@@ -33,6 +42,8 @@ def construct_config_settings(*, name, default_version, versions, minor_mapping)
         default_version: {type}`str` the default value for the `python_version` flag.
         versions: {type}`list[str]` A list of versions to build constraint settings for.
         minor_mapping: {type}`dict[str, str]` A mapping from `X.Y` to `X.Y.Z` python versions.
+        documented_flags: {type}`list[str]` The labels of the documented settings
+            that affect build configuration.
     """
     _ = name  # @unused
     _python_version_flag(
@@ -44,6 +55,7 @@ def construct_config_settings(*, name, default_version, versions, minor_mapping)
     _python_version_major_minor_flag(
         name = _PYTHON_VERSION_MAJOR_MINOR_FLAG.name,
         build_setting_default = "",
+        python_version_flag = _PYTHON_VERSION_FLAG,
         visibility = ["//visibility:public"],
     )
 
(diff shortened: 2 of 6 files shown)


docs/api/rules_python/python/config_settings/index.md
```diff
@@ -240,3 +240,21 @@ instead.
...
 ::::
+
+::::{bzl:flag} current_config
+Fail the build if the current build configuration does not match the
+{obj}`pip.parse` defined wheels.
+
+Values:
+* `fail`: Will fail in the build action ensuring that we get the error
+  message no matter the action cache.
+* ``: The default value, that will just print a warning.
```
rickeylev, line 251:
```suggestion
* ``: (empty string) The default value, that will just print a warning.
```

python/private/config_settings.bzl
```diff
@@ -135,8 +166,45 @@ _python_version_major_minor_flag = rule(
     implementation = _python_version_major_minor_flag_impl,
     build_setting = config.string(flag = False),
     attrs = {
-        "_python_version_flag": attr.label(
-            default = _PYTHON_VERSION_FLAG,
-        ),
+        "python_version_flag": attr.label(mandatory = True),
```
rickeylev, line 169: I don't see why this was made a public attribute?

python/private/config_settings.bzl
```diff
@@ -101,6 +113,25 @@ def construct_config_settings(*, name, default_version, versions, minor_mapping)
...
+    )
+    native.config_setting(
+        name = "is_not_matching_current_config",
+        # We use the rule above instead of @platforms//:incompatible so that the
+        # printing of the current env always happens when the _current_config rule
+        # is executed.
+        #
+        # NOTE: This should in practise only happen if there is a missing compatible
+        # `whl_library` in the hub repo created by `pip.parse`.
+        flag_values = {"current_config": "will-never-match"},
```
rickeylev, line 130: Are you sure this won't print spurious warnings?

IIRC, Bazel will evaluate all the select condition keys because it ensures that only one matches. This would imply that given e.g. `select({"py38": ..., ":current_config": ...})`, that `_print_current_config_impl()` is going to run, even if it won't match, and thus would print a message.

rickeylev (changes requested):

python/private/config_settings.bzl
aignas: So the whole structure is:
```
alias(
    name = "repo_alias",
    actual = select({
        "hub_select_condition_1": "repo_1",
        "hub_select_condition_2": "repo_2"
        "//conditions:default": "no_match_alias"
    }),
)

alias(
    name = "no_match_alias",
    select({"this-custom-incompatible-that-prints": "//python:none"}
    tags = ["manual"], # do we need this?
)
```

So my thinking is that the `no_match_alias` will not be analysed unless hub select conditions don't match, because the `no_match_alias` is not the key of the select dict but the value.

Unless we do `bazel build @pypi//...` We should not build the `no_match_alias` target. Maybe we have to have a manual tag?

python/private/config_settings.bzl
aignas: I thought that passing the actual names in the macro itself would be cleaner. This rule is private anyway, so that does not change anything in practise? I will revert the change.

python/private/config_settings.bzl
rickeylev: > So the structure is default maps to no-match-alias

Ahhh, ok. I was looking for this structure but missed it. LGTM.

> should we add manual to no-match-alias?

Yes. It just avoids a false failure when `...` et al accidentally pick it up. Which would probably be when poking around, debugging, or some situation where the false failure is probably going to be annoying.

python/private/config_settings.bzl
rickeylev: Reverting SGTM.

rickeylev (approved):
LGTM after the discussed changes (add manual tag, revert python_version_flag being public)

PR #2525, opened 2024-12-24
fix(pypi): use host micro version instead of a constant

aignas:
This is a workaround for #2319, but I would like to find a better way to solve
it, so I don't think we should resolve the ticket just yet.


aignas:
Feel free to use this patch as a workaround, but the real solution will have to be different.

PR #2531, opened 2024-12-24
fix(pypi): change the parallelisation scheme for querying SimpleAPI

### LINKED ISSUE #2100: Silence download 404 errors when using bazel downloader (`experimental_index_url` and `experimental_extra_index_urls`)
# 🚀 feature request

### Relevant Rules

`pip.parse` when using `experimental_index_url` and `experimental_extra_index_urls`

### Description

Using the bazel downloader can result in a lot of warnings:

```
$ bazel test //src/pyle_xc/fab/package_tester:web_api_test
WARNING: Download from https://[REDACTED]/simple/rsa/ failed: class java.io.FileNotFoundException GET returned 404 Not Found
WARNING: Download from https://[REDACTED]/simple/uv/ failed: class java.io.FileNotFoundException GET returned 404 Not Found
WARNING: Download from https://[REDACTED]/simple/h11/ failed: class java.io.FileNotFoundException GET returned 404 Not Found
WARNING: Download from https://[REDACTED]/simple/ply/ failed: class java.io.FileNotFoundException GET returned 404 Not Found
WARNING: Download from https://[REDACTED]/simple/jq/ failed: class java.io.FileNotFoundException GET returned 404 Not Found
```

Can these be silenced in some way?

This may not be feasible/reasonable via rules_python because it's part of [Bazel core](https://github.com/bazelbuild/bazel/blob/9fbf427d3c9943f0efa659f9934e77358c64e9a7/src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/HttpDownloader.java#L117).

An attribute on `pip.parse` would be nice.

I'm investigating the [bazel `--logging` flag](https://bazel.build/reference/command-line-reference#flag--logging) and asking around.

aignas:
Instead of querying everything in parallel and yielding a lot of 404
warnings, let's query the main index first and then query the other
indexes only for the packages that were not yet found.

What is more, we can print the value of `experimental_index_url_overrides`
for the users to use.

Whilst at it, add a unit test to check the new logic.

Fixes #2100, since this is the best `rules_python` can do for now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,10 @@ Unreleased changes template.
 * Bazel 6 support is dropped and Bazel 7.4.1 is the minimum supported
   version, per our Bazel support matrix. Earlier versions are not
   tested by CI, so functionality cannot be guaranteed.
+* ({bzl:obj}`pip.parse`) From now we will make fewer calls to indexes when
+  fetching the metadata from SimpleAPI. The calls will be done in parallel to
+  each index separately, so the extension evaluation time might slow down if
+  not using {bzl:obj}`pip.parse.experimental_index_url_overrides`.
 
 {#v0-0-0-fixed}
 ### Fixed
--- a/python/private/pypi/extension.bzl
+++ b/python/private/pypi/extension.bzl
@@ -653,6 +653,11 @@ The indexes must support Simple API as described here:
 https://packaging.python.org/en/latest/specifications/simple-repository-api/
 
 This is equivalent to `--extra-index-urls` `pip` option.
+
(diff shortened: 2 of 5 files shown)


aignas:
@dougthor42, if the CI passes, could you give this a spin please?

FYI @ewianda, this may slow down the fetching, but only if you are not using `experimental_index_url_overrides`.

dougthor42:
Seems to work well for me! I didn't see any unintended side effects.

Note that we already use `experimental_index_url_overrides` for all private-index stuff.

rickeylev (approved):

PR #2536, opened 2024-12-26
fix(gazelle): Don't ignore `setup.py` files when running Gazelle

### LINKED ISSUE #2108: Gazelle incorrectly ignores files called `setup.py`.
# 🐞 bug report

### Affected Rule

+ gazelle

### Is this a regression?

No, seems like it's been present from the begining (a5a7ffbf4 / #514).

### Description

Gazelle incorrectly ignores files called `setup.py`.

I believe this was added because the _root_ `setup.py` is something that doesn't need a bazel target (I'm not actually sure if that's true...) or maybe it's because the root `setup.py` is annoying to run Gazelle on.

Anyway, the problem is that _any_ file called `setup.py` will also be ignored.

I'm fine with the _root_ `setup.py` being ignored, though I reserve the right to change my mind in the future :laughing: 

This is the issue: https://github.com/bazelbuild/rules_python/blob/fa13b0138924245c55f79922509e4688f252ff71/gazelle/pythonconfig/pythonconfig.go#L128-L130

We can very easily update the [test case](https://github.com/bazelbuild/rules_python/tree/main/gazelle/python/testdata/python_ignore_files_directive) to account for this.

## 🔬 Minimal Reproduction

Dir structure:

```
.
|-- src/
|   +-- setup.py
|
+-- MODULE.bazel
```

### Expected result:

A `//src/setup` target is made in `src/BUILD.bazel`.

### Actual result

A `//src/setup` target is **not** made in `src/BUILD.bazel`.

## 🔥 Exception or Error

N/A

## 🌍  Your Environment

**Operating System:**

gLinux

7.2.0

**Rules_python version:**

0.31.0

**Anything else relevant?**

dougthor42:
Don't ignore `setup.py` files when running Gazelle.

Fixes #2108.

I believe that `setup.py` was originally ignored because it, when found that the repo root, is part of `setuptools` config and may have caused problems with Gazelle.

I've been running our Google Quantum code with this patch for a long while now and not seen any issues. I figured it was time to upstream it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,6 +75,8 @@ Unreleased changes template.
   are now printing more details and include the currently active flag
   values. Fixes [#2466](https://github.com/bazelbuild/rules_python/issues/2466).
 * (py_proto_library) Fix import paths in Bazel 8.
++ (gazelle) Gazelle no longer ignores `setup.py` files by default. To restore
+  this behavior, apply the `# gazelle:python_ignore_files setup.py` directive.
 
 [pep-695]: https://peps.python.org/pep-0695/
 
--- a/gazelle/python/testdata/dont_ignore_setup/BUILD.in
+++ b/gazelle/python/testdata/dont_ignore_setup/BUILD.in
@@ -1,0 +1,1 @@
+# gazelle:python_generation_mode file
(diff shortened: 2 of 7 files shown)


aignas (approved):
Thanks

Issue #2537, opened 2024-12-27
Metadata to aid static analysis


rickeylev:
Static analysis tools need additional information to help understand Python targets correctly. The use case of particular note are type checkers.

See discussion https://github.com/bazelbuild/rules_python/discussions/2455_

At a minimum, these tools need the original source files. For type checkers, pyi files are also needed. Similarly, things like proto (or other code gen) would also benefit from a way to provide pyi files.

* [ ] Add PyInfo fields: `{direct, transitive} x {original_sources, pyi_files}`
* [ ] Add `py_library.{pyi_{srcs, deps}`
* [ ] Update `py_library` et al to populate, merge, propagate this information
* [ ] Update pypi generation code to set attributes appropriately.

aignas:
Just a note for future selves - sometimes `whl` metadata has extra dependencies for type checking that are not included by default and we could potentially leverage the same heuristics as used in #2425 - i.e. the types packages have special names that are/can be easy to detect.

One question though is what happens if the user hasn't specified the types deps in the `hub` repo requirements? I am somewhat concerned that this could brake `bazel query` if the type stubs are not present.

Thinking about this more, I have a feeling that automatically propagating stubs from the external typing stub packages is probably a separate ticket as its scope is too large. Initially (as part of this ticket) we should only handle the `pyi` files that are present in the `whl_library` repos.

rickeylev:
All the items in the todo are done!

If downstream users could try using them / migrate to them, and provide feedback, that would be the last missing piece.

I'm going to leave this open for a bit before closing.

aignas:
No problems have been reported, so let's close this.

Issue #296, opened 2020-04-26
Strict rules for PY3


dayfine:
Will we have strict rules that check for deps and type annotations? i.e. `py_strict_library`, `py_strict_binary`, `py_strict_test`.

abergmeier:
> check for deps

Can you elaborate? Not sure what is meant.

> type annotations

Then `py_strict*` would be a macro also instantiating a type annotation rule?

dayfine:
At google we have (actually) `pytype_strict_{binary,library}` to check for typings using  https://github.com/google/pytype, and check for strict dependencies, i.e. depend on build targets for what you import.

As I realized this involves pytype, I suppose this could be a quite some work...

thundergolfer:
People are using macros to do this sort of thing right now, and Bazel Aspects alternatively. In future I think there should be a 'paved road' for writing type-annotated Python in Bazel, though can't say yet whether there'd be rules like `py_strict_binary` in `rules_python` that execute `pytype`, `mypy`, or [`pyre`](https://github.com/facebook/pyre-check).

rickeylev:
(I'm the author of the strict deps checking rule at Google)

"Strict deps" is a term we use to mean if you have "import foo" in your code, then some direct dependency provides "foo". It helps prevent your code from breaking when its dependencies's dependencies have changed.

The strict deps checking and pytype checking are separate implementations. We just bundle them into a single macro because usually people want both (both have their own separate stand alone macros, too).

The core design of how strict deps works would probably _mostly_ work for Bazel built Python code. It has to make some assumptions with how things work, of course. We've also had to work around a variety of short comings in the Python providers that starlark currently has.

wrt naming: having macro names that reflect a particular aspect/validator being enforced also comes with a combinatorical drawback. You end up with py_strict_*, pytype_*, mypy_*, pytype_strict_*, mypy_strict_*, etc etc. While it works, and there's a certain appeal to having "what it does" right there in the rule name, it's also a headache for new comers ("Which of the 12 rules do I use?") and naming. I wonder if there's a better way (the particular idea in my head is: what if I want to make sure my code works with pytype and mypy and pyre and strict deps and etc mypy_pytype_pyre_strict_library?).

dayfine:
Ok I see that there is open source consideration here that different type checkers may be used. So strict_deps and type checking are definitely two separate considerations.

[...]

PR #2538, opened 2024-12-27
feat: add pyi attributes/fields, original source fields

rickeylev:
This adds attributes and fields of use to static analysis.

For type definition files (usually `.pyi` files), the `pyi_srcs` and `pyi_deps`
fields are added to the rules. They end up in the PyInfo fields direct_pyi_files
and transitive_pyi_files.

So that static analysis tools can retain access to a target's Python source files,
even if precompiling is enabled, `direct_original_sources` and
`transitive_original_sources` fields are added to PyInfo.

Work towards https://github.com/bazelbuild/rules_python/issues/2537, https://github.com/bazelbuild/rules_python/issues/296

### DIFF
--- a/python/private/attributes.bzl
+++ b/python/private/attributes.bzl
@@ -375,6 +375,35 @@ in the resulting output or not. Valid values are:
 * `omit_source`: Don't include the original py source.
 """,
         ),
+        "pyi_deps": attr.label_list(
+            doc = """
+Dependencies providing type definitions the library needs.
+
+These are dependencies that satisfy imports guarded by `typing.TYPE_CHECKING`.
+These are build-time only dependencies and not included as part of a runnable
+program (packaging rules may include them, however).
+
+:::{versionadded} VERSION_NEXT_FEATURE
+:::
+""",
+            providers = [
+                [PyInfo],
+                [CcInfo],
+            ] + _MaybeBuiltinPyInfo,
+        ),
+        "pyi_srcs": attr.label_list(
+            doc = """
+Type definition files for the library.
+
+These are typically `.pyi` files, but other file types for type-checker specific
+formats are allowed. These files are build-time only dependencies and not included
+as part of a runnable program (packaging rules may include them, however).
+
+:::{versionadded} VERSION_NEXT_FEATURE
+:::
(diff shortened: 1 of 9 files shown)


python/private/attributes.bzl
```diff
@@ -375,6 +375,35 @@ in the resulting output or not. Valid values are:
 * `omit_source`: Don't include the original py source.
 """,
         ),
+        "pyi_deps": attr.label_list(
+            doc = """
+Dependencies providing type definitions the library needs.
+
+These are dependencies that satisfy imports guarded by `typing.TYPE_CHECKING`.
+These are build-time only dependencies and not included as part of a runnable
+program (packaging rules may include them, however).
```
aignas, line 384: Gazelle has a feature that populates the deps or data field for the pyi packages that one may have in the requirement file. I am not sure how big of a lift it is to accommodate for that but it raises some questions.

1. What happens if pyi sources propagate through deps vs pyi_deps.
2. Do I understand correctly that the pyi_deps will not be fetched if none of the actions consumes them?

aignas (approved):
LGTM. Gazelle support can come in a later PR.

FYI @ewianda, @dougthor42.

rickeylev:
> 1. What happens if pyi sources propagate through deps vs pyi_deps.

The deps attribute will try to include things into the final output for runtime. The pyi_deps attribute won't.

Given a target with only pyi files (i.e it only has things for build time, nothing for runtime), it doesn't matter much if something is put into deps or pyi_deps.

Given a target with e.g. both py and pyi files (i.e it has things for build time and things for runtime), putting it in pyi_deps means none of it will end up in the final output. This is useful if a source file has something guarded by `typing.TYPE_CHECKING` (the type checker needs the additional dependencies, but the runtime doesn't). Putting it in the deps attribute means the py files will be included in the final output (like normal), and the pyi files won't.

> 2. Do I understand correctly that the pyi_deps will not be fetched if none of the actions consumes them?

It'll be fetched, but it won't cause any actions to actually run. The analysis phase will need to resolve the target, but (assuming there is no type checker using the pyi files) none of the files/actions will be executed.

PR #2541, opened 2024-12-29
refactor: fold per-target python version into base rules

rickeylev:
Today, specifying the Python version for a target requires using the version-aware
rules in `transition.bzl` (or the generated equivalents bound to a specific Python
version). With the rules rewritten in Bazel, that functionality can be moved into
the base rules themselves. Moving the logic into the base rules simplifies the
implementation and avoids having to re-implement subtle behaviors in the wrappers
to correctly emulate the wrapped target.

For backwards compatibility, the symbols in `transition.bzl` are left as aliases
to the underlying rules.


python/config_settings/transition.bzl
```diff
@@ -14,266 +14,15 @@
...
-
-def py_test(name, python_version, **kwargs):
-    return _py_rule(_py_test, _transition_py_test, name, python_version, **kwargs)
 
-def _strip_suffix(s, suffix):
-    if s.endswith(suffix):
-        return s[:-len(suffix)]
-    else:
-        return s
+py_binary = _py_binary
```
aignas, line 27: Very small nit, we could create a macro that populate the deprecation message, that way users dont need to read our chngelog, which they might not do.

python/config_settings/transition.bzl
```diff
@@ -14,266 +14,15 @@
...
-    if "shard_count" in kwargs:
-        common_attrs["shard_count"] = kwargs.pop("shard_count")
-    if "local" in kwargs:
-        common_attrs["local"] = kwargs.pop("local")
-
-    # Binary-specific extra attributes.
-    if "output_licenses" in kwargs:
-        common_attrs["output_licenses"] = kwargs.pop("output_licenses")
-
-    rule_impl(
```
aignas, line 242: Does this mean that we will have fewer targets? The way I read the code is that the new version does not add this extra private target.

aignas (approved):

python/config_settings/transition.bzl
rickeylev: Correct.

python/config_settings/transition.bzl
rickeylev: Good idea. Done

PR #2542, opened 2025-01-02
chore: add check for version_next markers

rickeylev:
In a couple recent PRs, I put "VERSION_NEXT_{feature,patch}" as place holders since I
wasn't sure what the next appropriate version would be for unreleased code. e.g.
specifying `1.0.1` becomes invalid if a subsequent PR would make it `1.1.0`.

To help remind us to populate these values before a release, have the release workflow
check for the marker strings.

### DIFF
--- a/.github/workflows/create_archive_and_notes.sh
+++ b/.github/workflows/create_archive_and_notes.sh
@@ -15,6 +15,14 @@
 
 set -o errexit -o nounset -o pipefail
 
+# Exclude dot directories, specifically, this file so that we don't
+# find the substring we're looking for in our own file.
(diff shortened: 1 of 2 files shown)


aignas:
I think I have already used 1.1.0 in a few places, but I agree with using such an approach.

aignas (approved):
I think we should also add a note about this convention somwhere in our guidelines.

aignas:
Removed the auto-merge in case you would like to add the note in this PR.

rickeylev:
> I think we should also add a note about this convention somwhere in our guidelines.

Good idea. I've updated contributing.md with the guidance.

Issue #2551, opened 2025-01-08
Invoking bazel with a different PATH re-fetches pip_repository


Ubehebe:
# 🐞 bug report

Forked from https://github.com/bazelbuild/rules_python/issues/2283#issuecomment-2566924541

### Affected Rule

 The issue is caused by the rule:  pip_repository

### Is this a regression?

 Yes, the previous version in which this bug was not present was: ~0.32.2

### Description

I run `bazel build ...` from a git pre-commit hook as part of my development workflow. Although it is not well-documented, git modifies the `PATH` env var when running hooks. (You can see this for yourself by putting `echo "$PATH"` in a pre-commit hook, then comparing the output of `git hook run pre-commit` versus running the hook script directly.)

Meanwhile, since https://github.com/bazelbuild/rules_python/pull/2000, the pip_repository rule has internal code that calls repository_ctx.getenv("PATH"):  https://github.com/bazelbuild/rules_python/blob/main/python/private/repo_utils.bzl#L271. repository_ctx.getenv is [documented](https://bazel.build/rules/lib/builtins/repository_ctx#getenv):

> any change to the value of the variable named by name will cause this repository to be re-fetched.

Putting the git and rules_python behaviors together causes the bug. Doing a `bazel build ...` directly and then doing it indirectly via `git hook run pre-commit` re-fetches the pip_repository. (The reverse order also works.)

## 🔬 Minimal Reproduction

https://github.com/Ubehebe/rules_python_2283_mvce

## 🔥 Exception or Error

Console messages printed when pip_repository is unexpectedly re-fetched. Example:

```
Collecting pandas==2.1.4 (from -r /var/folders/28/qqqxx4y55dz08x1bzcc9z7jm0000gn/T/tmpyann5bc8 (line 1))
```

## 🌍  Your Environment

**Operating System:**

```
$ uname -a
Darwin 24.1.0 Darwin Kernel Version 24.1.0: Thu Oct 10 21:03:11 PDT 2024; root:xnu-11215.41.3~2/RELEASE_ARM64_T6020 arm64
```

**Output of `bazel version`:**

```
Bazelisk version: development
Build label: 8.0.0
Build target: @@//src/main/java/com/google/devtools/build/lib/bazel:BazelServer
Build time: Mon Dec 09 18:02:00 2024 (1733767320)
Build timestamp: 1733767320
Build timestamp as int: 1733767320
```

**Rules_python version:**

1.0.0

[...]

PR #2552, opened 2025-01-08
fix: _which_unchecked: don't watch PATH if binary exists.

Ubehebe:
Currently, the _which_unchecked helper unconditionally watches the `PATH` env var via repository_ctx.getenv. getenv is documented https://bazel.build/rules/lib/builtins/repository_ctx#getenv:

> any change to the value of the variable named by name will cause this repository to be re-fetched.

Thus, any change to `PATH` will cause any repository rule that transitively calls _which_unchecked to be re-fetched. This includes python_repository and whl_library.

There are reasonable development workflows that modify `PATH`. In particular, when git runs a hook, it adds the value of `GIT_EXEC_PATH` to `PATH` before invoking the hook. If the hook invokes bazel (for example, a pre-commit hook running `bazel build ...`), it will cause the Python repository rules to be re-fetched.

This commit lowers the repository_ctx.getenv("PATH") call to its only use site in _which_unchecked, which happens to be a failure case (when the binary is not found). This allows the success case to not watch `PATH`, and therefore not to re-fetch the repository rule when it changes.

Fixes https://github.com/bazelbuild/rules_python/issues/2551.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -119,6 +119,9 @@ Unreleased changes template.
   change. Fixes [#2468](https://github.com/bazelbuild/rules_python/issues/2468).
 + (gazelle) Gazelle no longer ignores `setup.py` files by default. To restore
   this behavior, apply the `# gazelle:python_ignore_files setup.py` directive.
+* Don't re-fetch whl_library, python_repository, etc. repository rules
+  whenever `PATH` changes. Fixes
+  [#2551](https://github.com/bazelbuild/rules_python/issues/2551).
 
 [pep-695]: https://peps.python.org/pep-0695/
 
--- a/python/private/repo_utils.bzl
+++ b/python/private/repo_utils.bzl
@@ -256,7 +256,7 @@ def _which_checked(mrctx, binary_name):
 def _which_unchecked(mrctx, binary_name):
     """Tests to see if a binary exists.
 
-    This is also watch the `PATH` environment variable.
+    Watches the `PATH` environment variable if the binary doesn't exist.
 
     Args:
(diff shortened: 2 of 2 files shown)


aignas:
This makes sense to me, I am happy to stamp it if you could add a more descriptive commit message about before and after behaviour and a Changelog.md item since having a self contained git history with the motivation for changes is always really god for digging the code later.

aignas:
Should we wait for this PR to release 1.1.0? @rickeylev, WDYT?

CHANGELOG.md
```diff
@@ -119,6 +119,9 @@ Unreleased changes template.
   change. Fixes [#2468](https://github.com/bazelbuild/rules_python/issues/2468).
 + (gazelle) Gazelle no longer ignores `setup.py` files by default. To restore
   this behavior, apply the `# gazelle:python_ignore_files setup.py` directive.
+* Don't re-fetch whl_library, python_repository, etc. repository rules
```
Ubehebe, line 122: I'm not exactly sure what to put in the parentheses here. This change affects anything that transitively calls which_unchecked, which I think includes whl_library (the failure mode I observed), python_repository, and local_runtime_repo. Feel free to re-word.

Ubehebe:
> add a more descriptive commit message about before and after behaviour and a Changelog.md item

Done. I added the changelog entry under 1.1.0 (feel free to move).

CHANGELOG.md
rickeylev: LGTM

rickeylev (approved):

rickeylev:
I'll include it in 1.1

Issue #2554, opened 2025-01-09
Trailing slashes not removed from pip index URLs in environment variables before concatenation.


WillMorrisonEnspi:
# 🐞 bug report

### Affected Rule

The `simpleapi_download` function called by the `pip.parse` repo rule.

### Is this a regression?

No

### Description

Environment variables are [substituted](https://github.com/bazelbuild/rules_python/blob/89d850aab819eb2dea9d6340beab1ca810dbe18d/python/private/pypi/simpleapi_download.bzl#L172) *after* [stripping trailing / characters](https://github.com/bazelbuild/rules_python/blob/89d850aab819eb2dea9d6340beab1ca810dbe18d/python/private/pypi/simpleapi_download.bzl#L96), meaning that an environment variable containing a trailing slash in `experimental_index_url` will not have that trailing slash stripped. The index URL is concatenated with packages and results in paths with double slashes where there should be single ones. This in turn leads to 404 errors when attempting to fetch index URLs.

Given that the existing code tries to handle index URLs with trailing slashes already, and that the [Simple API spec](https://packaging.python.org/en/latest/specifications/simple-repository-api/) says "All URLs which respond with an HTML5 page MUST end with a /" and "the root URL MUST be a valid HTML5 page", it seems reasonable to consider this a bug within rules_python rather than just making users manually remove trailing slashes from `$PIP_EXTRA_INDEX_URL` themselves.

### `MODULE.bazel`:
```python
bazel_dep(name = "rules_python", version = "1.0.0")

python = use_extension("@rules_python//python/extensions:python.bzl", "python")
python.toolchain(
    is_default = True,
    python_version = "3.11.5",
)
pip = use_extension("@rules_python//python/extensions:pip.bzl", "pip")
pip.parse(
    hub_name = "reqs",
    envsubst = ["PIP_EXTRA_INDEX_URL"],
    experimental_index_url = "$PIP_EXTRA_INDEX_URL",
    python_version = "3.11.5",
    requirements_lock = "//:requirements.txt",
)
use_repo(pip, "reqs")
```

```sh
export PIP_EXTRA_INDEX_URL='https://<redacted>.d.codeartifact.eu-central-1.amazonaws.com/pypi/packages-prod/simple/'
bazel fetch '//...'
```

## 🔥 Exception or Error

<pre><code>
[...]

PR #2557, opened 2025-01-12
fix: Avoid creating URLs with empty path segments from index URLs in environment variables

WillMorrison:
This change updates `_read_simpleapi` such that it correctly handles the case where the index URL is specified in an environment variable and contains a trailing slash. The URL construction would have introduced an empty path segment, which is now removed.

Fixes: #2554


tests/pypi/simpleapi_download/simpleapi_download_tests.bzl
```diff
@@ -119,6 +119,121 @@ def _test_fail(env):
...
+        downloads[url[0]] = output
+        return struct(success = True)
+
+    simpleapi_download(
+        ctx = struct(
+            os = struct(environ = {"INDEX_URL": "https://example.com/main/simple/"}),
+            download = download,
+            read = lambda i: "contents of " + i,
+            path = lambda i: "path/for/" + i,
+            get_auth = lambda ctx, urls, ctx_attr: struct(),
```
aignas, line 206: I think adding an extra arg to `simpleapi_download` would be better. `ctx` in general does not have `get_auth` attribute.

CHANGELOG.md
```diff
@@ -57,6 +57,7 @@ Unreleased changes template.
 {#v0-0-0-fixed}
 ### Fixed
 * (gazelle) Providing multiple input requirements files to `gazelle_python_manifest` now works correctly.
+* (bazel downloader) Handle trailing slashes in pip index URLs in environment variables
```
aignas, line 60:
```suggestion
* (pypi) Handle trailing slashes in pip index URLs in environment variables,
  fixes [#2554](https://github.com/bazelbuild/rules_python/issues/2554).
```

aignas (commented):
Minor comments only.

tests/pypi/simpleapi_download/simpleapi_download_tests.bzl
WillMorrison: Done.

aignas (approved):

PR #2558, opened 2025-01-12
fix(pypi): include pyi files in data attribute

rickeylev:
Restore the previous behavior of pyi files being included in data. This is because
certain packages (librosa, at least) expect the pyi files to be available at runtime.


aignas (approved):
SGTM for reverting. Should we include them in both places though?

PR #2567, opened 2025-01-17
docs: update gazelle README.md

alexeagle:
There's no longer a python subprocess since switching to tree-sitter in https://github.com/bazelbuild/rules_python/pull/1895


dougthor42 (approved):
:tada:

No changelog entry needed, IMO.

PR #2565, opened 2025-01-16
docs: prefer python_version attribute for specifying python version

JeroenSchmidt:
This PR is updating the Examples and Docs relating to the recent changes `refactor: fold per-target python version into base rules` introduced in v.1.1.0.

See the brief conversation between myself and @rickeylev in this diff https://github.com/bazelbuild/rules_python/commit/611eda87e48a3943808216cf4e4de875040b09aa

PR #2569, opened 2025-01-21
chore: remove internal usage of deprecated py_binary ad py_test

aignas:
This goes together with #2565 to remove the internal usage of the deprecated
symbols. This also fixes the  compile_pip_requirements  symbol to print the
correct deprecation message.

Builds on top of 611eda8

The example message that would be printed is as follows:
```
The 'py_test' symbol in '@+python+python_3_11//:defs.bzl'
is deprecated. It is an alias to the regular rule; use it directly instead:

load("@rules_python//python:py_test.bzl", "py_test")

py_test(
    name = "versioned_py_test",
    srcs = ["dummy.py"],
    main = "dummy.py",
    python_version = "3.11.11",
)
```


aignas:
This is also deprecating the symbols in the `pythons_hub` and the other parts as was suggested in #2565. If this is not (yet) desirable, please comment on this PR. I personally think that we might be able to do this, but I am not sure if there are no problems (e.g. with using `3.10` in non-root modules).

If we want to start printing deprecation messages in `3.10` or add a flag that is controlling this behaviour via `rules_python_internal` or some other method, I am happy to amend to this PR.

Maybe the `with_deprecation` symbol should be in `//python/private:with_deprecacion`?

I ran out of time today, so that is it for today.

aignas:
This now only warns if `RULES_PYTHON_DEPRECATION_WARNINGS` is `1`, which defaults to `0`. I am fine with defaulting it to `1` if we want to default to a more noisy default.

python/private/internal_config_repo.bzl
```diff
@@ -79,6 +85,7 @@ def _internal_config_repo_impl(rctx):
 
     rctx.file("rules_python_config.bzl", _CONFIG_TEMPLATE.format(
         enable_pystar = enable_pystar,
+        enable_deprecation_warnings = enable_deprecation_warnings,
```
rickeylev, line 88: nit: Just inline the bool_from_environ() call here. The other one's aren't inlined because they have more complex logic to determine their value.

rickeylev (approved):
Just nits/style, otherwise LGTM

rickeylev:
This is great, thank you! I didn't realized I missed so many other spots, sorry about that.

re: deprecating pythons_hub symbols: Yes, deprecating those also makes sense

re: move deprecation helper elsewhere: yes, I agree

re: environment variable: Thanks for creating this. I like this idea a lot.

rickeylev:
addendum: I've approved, but not queued for merge, in case you want to move with_deprecated elsewhere before merging. I'm OK with that being done in a separate PR, too.

PR #2576, opened 2025-01-23
fix: set PYTHONNOUSERSITE 

froody:
Set PYTHONNOUSERSITE to prevent adding user directories outside bazel. I found a py_test run with `bazel test` was importing modules from ~/.local/lib/python3.10/site-packages/ which allowed it to run successfully on my machine but fail on other hosts.

Before:

USER_SITE directory in sys.path when running python code via
`bazel run` or `bazel test`

After:

sys.path only contains paths relative to the `bazel-out` directory

### DIFF
--- a/python/private/python_bootstrap_template.txt
+++ b/python/private/python_bootstrap_template.txt
@@ -528,6 +528,10 @@ def Main():
   # See: https://docs.python.org/3.11/using/cmdline.html#envvar-PYTHONSAFEPATH
   new_env['PYTHONSAFEPATH'] = '1'
 
+  # Set the PYTHONNOUSERSITE environment variable to prevent Python from adding
+  # directories uknown to bazel to sys.path. This is important for reproducibility.
+  new_env['PYTHONNOUSERSITE'] = '1'
+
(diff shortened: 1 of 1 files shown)


rickeylev:
Hi @froody, thanks for your contribution!

I think this is a good idea, so +1 overall setting this. Changes needed to merge:

* Pass `-s` to the interpreter instead of an environment variable. envvars contaminate subprocesses, which is undesirable.
* Please update the bootstrap=script impl. i.e update python/private/stage1_bootstrap_template.sh to pass `-s` too
* Please update CHANGELOG.md and add an entry under Changed.
* We'll need tests. Have a look at tests/bootstrap_impls. I think you can copy/paste run_binary_zip_no_test, modify bin.py to print out sys.flags.no_user_site, and check the value in run_binary_zip_no_test.sh

I'm pretty certain this is gonna break something about some user's workflow, though I don't know what that would be.

aignas:
I would also like to see the behaviour switched based on a config flag so that users that need to opt out from the behaviour can do so.

Defaulting to always including `-s` should be good enough.

PR #2577, opened 2025-01-24
fix(pypi): correctly translate ppc64le to bazel platforms

aignas:
Bump the `platforms` version and correctly translate the ppc64le value.

See https://github.com/bazelbuild/platforms/pull/105


rickeylev (approved):
Do we need a change to the workspace deps, too?

aignas:
Hmm, I could not find us pulling `platforms` as an explicit dependency, so we could probably leave it as is?

rickeylev:
SGTM

PR #2580, opened 2025-01-25
refactor(uv): move around uv implementation files

aignas:
This PR starts establishing a structure that will eventually become a
part of our API. This is a prerequisite for #2578 which removes the
versions.bzl file in favour of a more dynamic configuration of the
extension. We also remove the `defs.bzl` to establish a one symbol per
file convention.

Things that I wish we could change is `//python/uv:extensions.bzl` and
the fact that we have `extensions` in the load path. I think it cannot
be removed, because that may break the BCR test. On the other hand,
maybe we could remove it and do an alpha release to verify this
assumption.

Work towards #1975


rickeylev (approved):

rickeylev:
I think it should be OK for BCR because it sets dev_dependency. When BCR runs, it creates a root module that depends on the module-under-test/release. A way to check this would be to add a dev dep use_extension() of a non-existent path in rules_python, then run one of the examples.

Issue #2173, opened 2024-09-01
Cleanup proto rules before releasing 1.0


aignas:

rickeylev:
I think @alexeagle @comius were working on this? Last I remember, we were waiting for a protobuf release of something before we could switch rules_python to switch to loading them and mark them as deprecated.

alexeagle:
I have not seen a migration plan for that. I don't think it's a wise plan to remove these from rules_python at this point. In fact much of the ecosystem has already moved off of the protobuf repo, now that the toolchain flag allows it.

aignas:
I think I must be confused as to what is the ruleset that should be used for protobuf python library - is it not the py_proto_library from the protobuff repo that should be used instead of the py_proto_library in our repo? What is the maintained rule that people should be advised to use? I thought that it was definitely not py_proto_library from rules_python but not sure about the replacement.

alexeagle:
I find it confusing as well. The recent changes are anonymous and have no context or indication of what they are doing:

<img width="1275" alt="Screenshot 2024-09-02 at 3 44 50 PM" src="https://github.com/user-attachments/assets/858b58a2-5703-4d19-ac05-14f8218ca4dc">

aignas:
I guess then it means that we'll just continue having the api as is and we leave the removal for 2.0.

tpudlik:
This is the first I hear of a plan to deprecate `py_proto_library` in `rules_python` in favor of something in `protobuf`. Is there an issue or doc with more context?

aignas:
#1933 was the first time I heard of desire to do something like this.

tpudlik:
@comius could you provide some clarity on what the plan is? If even the `rules_python` maintainers seem unsure, there's no hope for regular users like myself!

aignas:
I wonder what is the functional difference between what we have and what is in https://github.com/protocolbuffers/protobuf/blob/cf688f339c80dc372a34302de6dae3d749ec8e72/bazel/py_proto_library.bzl#L132

aignas:
Related slack discussion: https://bazelbuild.slack.com/archives/CKU1D04RM/p1732581509956829

Issue #2515, opened 2024-12-19
`import` paths from external Bazel modules are broken with Bazel 8.0.0


armandomontanez:
# Summary
`imports` runfiles handling of external projects appears to have broken with the switch to `--legacy_external_runfiles=False` in Bazel 8.0.0.

# Minimal reproducer
https://github.com/armandomontanez/bazel_reproducers/tree/main/external_runfiles_broken_rules_python

Reproduces on both macOS and linux.

# Additional notes
I haven't unwound yet whether or not all `import` handling of external `py_*` rules are completely broken, I noticed this specifically with `py_proto_library` rules that use `strip_import_prefix` and live in an external repository.

I noticed this when working with nanopb where `sys.path` reported the import path as:

```
/private/var/tmp/_bazel_amontanez/e724b21efc8bc19866072fbc72ee5907/sandbox/darwin-sandbox/7/execroot/_main/bazel-out/darwin_x86_64-opt-exec-ST-e8d06b237d06/bin/external/nanopb+/protoc-gen-nanopb.runfiles/_main/external/nanopb+/_virtual_imports/nanopb_proto
```

when the files actually existed at:
```
/private/var/tmp/_bazel_amontanez/e724b21efc8bc19866072fbc72ee5907/sandbox/darwin-sandbox/7/execroot/_main/bazel-out/darwin_x86_64-opt-exec-ST-e8d06b237d06/bin/external/nanopb+/protoc-gen-nanopb.runfiles/_main/../nanopb+/_virtual_imports/nanopb_proto
```

tpudlik:
I wonder if this is an issue specifically with how `py_proto_library` constructs import paths? This logic basically [lives here](https://github.com/protocolbuffers/protobuf/blob/main/bazel/py_proto_library.bzl#L74-L79). (Note that this is now in protobuf, not in rules_python.)

I touched that code in the past, let me take a closer look.

tpudlik:
I came up with a fix for `py_proto_library` in the protobuf repo. This fixes the reproducer (but I've not yet done any testing to verify that nothing else breaks). However, to my surprise, there now appear to be copies of this implementation in both protobuf and rules_python, so I guess they both need to be fixed :/

This fix is very hacky, though. I need to think about how to address this more cleanly.

```
diff --git a/bazel/py_proto_library.bzl b/bazel/py_proto_library.bzl
index 1c4059842..46b22d1a3 100644
--- a/bazel/py_proto_library.bzl
[...]

PR #2581, opened 2025-01-25
feat(proto): shim to upstream for py_proto_library

### LINKED ISSUE #2543: Avoid downloading protobuf dependancy when not needed
Hi we recently migrated a project from workspace to bzlmod and noticed our builds now pull in more things from the internet as a result.
In particular we saw that our builds started to download protobuf (from https://github.com/protocolbuffers/protobuf/releases/download/v29.0-rc2/protobuf-29.0-rc2.zip).
This was despite none of our targets referencing (either directly nor indirectly) `py_proto_library` (or any other protobuf target from `rules_python` or any other module).

It seems that in a workspace project dependencies are fetched lazily [1] whereas in a bzlmod project, dependencies are fetched eagerly.

Reading the source of this library, I saw a comment in the MODULE.bzl https://github.com/bazelbuild/rules_python/blob/29fdcc2c292d64e7c66746d1c701ab67aa05c79b/MODULE.bazel#L12 that suggests the `protobuf` dep will be loaded only if the `py_proto_library` is used. However, that doesn't match my tests where  `protobuf` is download by all projects that depend on `rules_python` even if they do not use _any_ rules from it [2].

I am wondering

1. if there is a way to use `rules_python` without pulling `protobuf` if we do not need protobuf support?
2. If the comment linked above needs updating/removing?

#### Related 

I think https://github.com/bazelbuild/rules_python/issues/2498 is similar although that is about rules_kotlin and is more about toolchains fetched as module extensions rather than modules downloaded directly.

#### Notes

1: https://bazel.build/extending/repo#when_is_the_implementation_function_executed
2: You can test this by creating a custom registry (https://bazel.build/external/registry) and modifying the `source.json` file of all the protobuf modules to set the `url` field to something that does not exist. Building a bzlmod project that uses rules_python but does not have any `py_proto_library` targets will then fail showing that bazel does try to fetch protobuf.

aignas:
With this PR we attempt to remove our implementation of py_proto_library
and we just forward the calls to the upstream.

Hopefully this fixes #2543.
Work towards #2173.


aignas:
@tpudlik, did you have a chance to also submit the fix for #2515 to the upstream, i.e. the `protocolbuffers` org?

rickeylev:
I noticed the proto version was updated -- does this mean 29.3 has all the parts we need, so we can put this shim in now?

aignas:
I think we could do it with the old version, but the fix that @tpudlik put in is not in the upstream, so I am wondering what we should do...

aignas:
Maybe we can provide users with a small patch before it gets picked up by the protocolbuffers team. I think it would be the most we could do from our side.

tpudlik:
Yes, sorry, I've not contributed it yet! Sorry, it slipped my mind. I'll send a CL (PR) this week, but it will take a little while to make it into a released version, of course.

aignas:
Closing in favour of #2604

PR #2583, opened 2025-01-26
fix: Enable location expansion for `sh_py_run_test`

philsc:
I noticed that my `$(location //path/to:target)` wasn't getting
expanded when writing a test. This patch fixes the issue by forwarding
the already-expanded environment from the inner target to the outer
target.


aignas (approved):
LGTM

PR #2584, opened 2025-01-26
revert: Updated pip and packaging versions to work with free-threading packages (#2514)

### LINKED ISSUE #908: compile_pip_requirements no longer uses predictable paths

aignas:
This reverts commit fbf8bc10a466c498fc80c6b58c939a87a8d9e929 (#2514)

Also, update the CHANGELOG about the reverting.

Fixes #908, which is about the `pip-compile` not using the right files
for performing the locking. It seems that the `pip` upgrade regressed
this error.


rickeylev (approved):
Please update description to mention the problem and why this fixes it. Otherwise LGTM

PR #2588, opened 2025-01-28
feat(python.toolchain): support file-based default Python version

### LINKED ISSUE #2587: Support environment/file-based default Python version in python.toolchain
# 🚀 feature request

### Relevant Rules

`python.toolchain`
where `python = use_extension("@rules_python//python/extensions:python.bzl", "python")`

### Description

It should be possible to set the default Python version using a file or an environment variable.

In `python.toolchain` you need to set the `is_default` attribute. You can have several Python versions defined, and for one of them you need to set `is_default = True`. Since a `MODULE.bazel` file cannot read an environment variable or read a file, you need to set the default in `MODULE.bazel`. Before Bzlmod, you could simply load a file created by a repo rule and have access to the value of an environment variable or the contents of a `.python-version` file in the `WORKSPACE.bazel` file, and set the `default_version` in a call to `python_register_multi_toolchains` accordingly.

This is important for interoperability in projects where not everything is run through Bazel; `pyenv` for example uses the `PYENV_VERSION` environment variable, falling back on the `.python-version` file if the environment variable is not set.

A new attribute for `python.toolchain` called e.g. `default_python_version_file`, which can be the label of a file generated by a repository rule, or a file in the current repo. If the `python_version` attribute of `python.toolchain` matches the contents of the file, treat it as the default (as if `is_default = True`).

One could imagine having both an attribute for setting the default Python version with a file and another attribute for setting the default Python version  with an environment variable, which would lower the threshold for any user wanting to react to an environment variable, at the cost of a more complex implementation in rules_python. Since the same functionality can be achieved with a repository rule, I'm leaning towards telling users to write a repository rule.

One could also imagine having the default Python version be defined in `MODULE.bazel`, as the source of truth, and have something like `write_source_file` update a `.python-version` file in the repo for compatibility with other tools. But done this way, there's no way to support environment variables as a way of setting the default Python version.

vonschultz:
This change adds a new `default_version_file` attribute to `python.toolchain`. If set, the toolchain compares the file's contents to its `python_version`, and if they match, treats that toolchain as default (ignoring `is_default`). This allows Bazel to synchronize the default Python version with external tools (e.g., pyenv) that use a `.python-version` file or environment variables.

Fixes #2587.


python/private/python.bzl
```diff
@@ -104,7 +104,14 @@ def parse_modules(*, module_ctx, _fail = fail):
                 # * rules_python needs to set a soft default in case the root module doesn't,
                 #   e.g. if the root module doesn't use Python itself.
                 # * The root module is allowed to override the rules_python default.
-                is_default = toolchain_attr.is_default
+                if toolchain_attr.default_version_file == None:
+                    is_default = toolchain_attr.is_default
+                else:
+                    is_default = (
+                        module_ctx.read(toolchain_attr.default_version_file) == toolchain_version
+                    )
```
aignas, line 112:
```suggestion
                if toolchain_attr.default_version_file:
                    version_from_file = module_ctx.read(toolchain_attr.default_version_file).strip()
                    is_default = version_from_file == toolchain_version
                else:
                    is_default = toolchain_attr.is_default
```

I think the inverted if reads a little bit easier and I also think we should have `.strip()` on the contents of the file to remove whitespace.

python/private/python.bzl
```diff
@@ -651,7 +671,9 @@ can result in spurious build failures.
         ),
         "is_default": attr.bool(
             mandatory = False,
-            doc = "Whether the toolchain is the default version",
+            doc = """\
+Whether the toolchain is the default version. Ignored if `default_version_file`
+is set.""",
```
aignas, line 676:
```suggestion
Whether the toolchain is the default version. 

:::{versionchanged} VERSION_NEXT_FEATURE
This setting is ignored if {attr}`default_version_file` is set.
:::
""",
```

python/private/python.bzl
```diff
@@ -635,6 +643,18 @@ Then the python interpreter will be available as `my_python_name`.
             mandatory = False,
             doc = "Whether or not to configure the default coverage tool provided by `rules_python` for the compatible toolchains.",
         ),
+        "default_version_file": attr.label(
+            mandatory = False,
+            allow_single_file = True,
+            doc = """\
+File saying what the default Python version should be. If the contents of the
+file match the `python_version` attribute, this toolchain is the default version.
+If this attribute is set, the `is_default` attribute is ignored.
```
aignas, line 652:
```suggestion
file match the {attr}`python_version` attribute, this toolchain is the default version.
If this attribute is set, the {attr}`is_default` attribute is ignored.
```

aignas (commented):
The implementation in general looks to me and I think this is probably a fine way to support such usecase. My only ask would be to see if there is a way to add a unit test in `//tests/python/` `.bzl` file.

EDIT: sorry it took a while to dedicate attention to this and I really appreciate the write up and the explanation behind the usecase. Thank you once more for your contributions.

aignas (changes requested):
I actually wanted to write an example here and realized that the current API is not very clear to the user. What one would have to do is do something like below:
```starlark
python.toolchain(
    version = "3.9",
    default_version_file = ".python_version"
)
python.toolchain(
    version = "3.10",
)
```

which is very similar and the user may be confused as to where and how to use the API. Is the user supposed to have:

```starlark
python.toolchain(version = "3.10", default_version_file = ".python_version")
python.toolchain(version = "3.11", default_version_file = ".python_version")
python.toolchain(version = "3.12", default_version_file = ".python_version")
```
?

Should we have a separate `tag_class` that sets the defaults? I think I want to finish #2578 PR to understand if we can learn something from the builder API implemented there. The potential is to have a more DRY version:
```starlark
python.defaults(version_file = ".python_version)
python.toolchain(version = "3.9")
python.toolchain(version = "3.10")
python.toolchain(version = "3.11")
```

vonschultz:
Exactly, it would be
```starlark
python.toolchain(version = "3.10", default_version_file = ".python-version")
python.toolchain(version = "3.11", default_version_file = ".python-version")
python.toolchain(version = "3.12", default_version_file = ".python-version")
```
yes. I don't think that's ideal, though. I like the idea of a separate `tag_class` for defaults. Then we could potentially support both something like
```starlark
python.defaults(python_version_file = ".python-version")
python.toolchain(version = "3.9")
python.toolchain(version = "3.10")
python.toolchain(version = "3.11")
```
and something like
```starlark
python.defaults(python_version = "3.11")
python.toolchain(version = "3.9")
python.toolchain(version = "3.10")
python.toolchain(version = "3.11")
```
and perhaps even implement a default of looking in the root module `.python-version` file if nothing else has been specified.

aignas:
#2578 implements the design for defaults that got approved, so I think we could almost copy paste that here.

vonschultz:
I originally had the idea that supporting reading the Python version from a file would be sufficient to also solve the use case where you want to read it from an environment variable, since you could generate a file with whatever logic you need. However, I read in the documentation of `module_ctx.path` that "remote repositories and module extensions are executed during the analysis phase and thus cannot depends on a target result (the label should point to a non-generated file)." For this reason I also included a `python_version_env` attribute in my latest patch, so both the file use case and the environment variable use case can be solved directly in `rules_python`.

python/private/python.bzl
```diff
@@ -104,7 +153,14 @@ def parse_modules(*, module_ctx, _fail = fail):
                 # * rules_python needs to set a soft default in case the root module doesn't,
                 #   e.g. if the root module doesn't use Python itself.
                 # * The root module is allowed to override the rules_python default.
-                is_default = toolchain_attr.is_default
+                if default_python_version:
+                    is_default = default_python_version == toolchain_version
+                    if toolchain_attr.is_default and not is_default:
+                        fail("The 'is_default' attribute doesn't work if you set " +
```
aignas, line 159:
```suggestion
                        logger.warn("The 'is_default' attribute doesn't work if you set " +
```

Maybe having the logger warn the user is enough here? I imagine people may have a `.python-version` file that they expect bazel to not read and then suddenly `rules_python` will start failing, which would be surprising behaviour.

python/private/python.bzl
```diff
@@ -570,6 +678,49 @@ def _get_bazel_version_specific_kwargs():
...
+""",
+        ),
+        "python_version_file": attr.label(
+            mandatory = False,
+            allow_single_file = True,
+            doc = """\
+File saying what the default Python version should be. If the contents
+of the file match the {attr}`python_version` attribute of a toolchain,
+this toolchain is the default version. If this attribute is set, the
+{attr}`is_default` attribute of the toolchain is ignored.
```
aignas, line 715: Please document that we default to `.python-version` if we find it.

python/private/python.bzl
```diff
@@ -78,6 +78,55 @@ def parse_modules(*, module_ctx, _fail = fail):
...
+                # named by `name` will cause this repository to be re-fetched.
+                if "getenv" in dir(module_ctx):
+                    getenv = module_ctx.getenv
+                else:
+                    getenv = module_ctx.os.environ.get
+                default_python_version = getenv(default_python_version_env, default_python_version)
+    if not default_python_version:
+        fallback_python_version_file = module_ctx.path("@@//:.python-version")
+        if fallback_python_version_file.exists:
+            default_python_version = module_ctx.read(fallback_python_version_file).strip()
```
aignas, line 128:
```suggestion
        if fallback_python_version_file.exists:
            default_python_version = module_ctx.read(fallback_python_version_file).strip()
```

We should ensure that the watch is done for the version file if we decide to read. I remember that `module_ctx.path` no longer watches things by default, but I can't remember if `module_ctx.read` does or not. Maybe  would be safer to pass `watch="yes"` to the read?

https://bazel.build/versions/8.1.0/rules/lib/builtins/module_ctx#read

python/private/python.bzl
```diff
@@ -653,7 +804,14 @@ error to run with root access instead.
...
         "is_default": attr.bool(
             mandatory = False,
-            doc = "Whether the toolchain is the default version",
+            doc = """\
+Whether the toolchain is the default version.
+
+:::{versionchanged} VERSION_NEXT_FEATURE
+This setting is ignored if the default version is set using the `defaults`
+tag class.
+:::
```
aignas, line 813: Should we deprecate this in favour of the `defaults` tag_class?

cc: @rickeylev what do you think?

aignas (commented):
Thanks a lot for the new design. I like how clean the interfaces look and I really like that `rules_python` reads `.python-version` by default, which makes it integrate with existing python tooling better.

I am torn if we should include this in 1.3.0 or if the ship has sailed. Since we are releasing right now and there may be a few more comments here and there, I would vote to just rebase and put this under unreleased CHANGELOG section.

aignas:
> I originally had the idea that supporting reading the Python version from a file would be sufficient to also solve the use case where you want to read it from an environment variable, since you could generate a file with whatever logic you need. However, I read in the documentation of module_ctx.path that "remote repositories and module extensions are executed during the analysis phase and thus cannot depends on a target result (the label should point to a non-generated file)." For this reason I also included a python_version_env attribute in my latest patch, so both the file use case and the environment variable use case can be solved directly in rules_python.

I think you could read the env in a repo rule and write a file (but everything has to be in `repository_ctx` and no build actions). However, this solution seems to be much cleaner.

python/private/python.bzl
vonschultz: If we enter this branch, wouldn't there be surprising behavior no matter what? If the user has said `python.toolchain(python_version = "3.11", is_default = True)` and the `.python-version` file says `3.12`, suddenly `rules_python` will start using `3.12` even though the user has explicitly asked for `3.11` to be the default. It's not that some obscure logic used to choose 3.11 and now some obscure logic chooses 3.12 and that's worth a warning — the user explicitly asked for something that we're now actively refusing to do. That feels more like a `fail` than a `logger.warn`, doesn't it? In this scenario, if the user wants `rules_python` to ignore the `.python-version` file and keep using `3.11`, they would have to migrate to the `python.defaults(python_version = "3.11")` syntax and stop using `is_default`. I suppose this may be considered a breaking change, which we might need to document better in the change log and perhaps the failure message or warning. If we don't want a breaking change, we'd have to change the logic so the `is_default` overrides the `.python-version` file: in that case, I think a warning would be suitable.

If we enter this branch because the user is using both `python.defaults(python_version = "3.12")`  and `python.toolchain(python_version = "3.11", is_default = True)`, I think that should be a `fail` rather than a `logger.warn`, because the user is explicitly asking for things that aren't consistent. We could try to keep track of whether the `.python-version` file or the `python.defaults` tag class is the source of the discrepancy and do different error messages (or warning, perhaps, in the case of the `.python-version` file).

python/private/python.bzl
```diff
@@ -78,6 +78,55 @@ def parse_modules(*, module_ctx, _fail = fail):
...
+                # Bazel version 7.1.0 and later support module_ctx.getenv(name, default):
+                # When building incrementally, any change to the value of the variable
+                # named by `name` will cause this repository to be re-fetched.
+                if "getenv" in dir(module_ctx):
+                    getenv = module_ctx.getenv
+                else:
+                    getenv = module_ctx.os.environ.get
+                default_python_version = getenv(default_python_version_env, default_python_version)
+    if not default_python_version:
+        fallback_python_version_file = module_ctx.path("@@//:.python-version")
```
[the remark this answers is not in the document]
vonschultz: It didn't actually work. :see_no_evil: It does work when I switch to using `Label("@@//:.python-version")`, which resolves to the root module's file. There is a potential issue if there is a `.python-version` file in the root module, but no `BUILD` or `BUILD.bazel` file in the root module. I don't know any way to detect it and ignore it:
```
		fallback_python_version_file = module_ctx.path(Label("@@//:.python-version"))
Error in path: Unable to load package for //:.python-version: BUILD file not found in any of the following directories. Add a BUILD file to a directory to mark it as a package.
```

For testing, there's the mocked tests in `//tests/python`, which doesn't really catch what `.python-version` we resolve to because of the mocking, and I've been playing around in `examples/multi_python_versions`, but I can't have the example test both the `python.defaults` tag class and the `.python-version` fallback, unless I copy the entire example. Not sure how to proceed.

python/private/python.bzl
```diff
@@ -78,6 +78,55 @@ def parse_modules(*, module_ctx, _fail = fail):
...
+                default_python_version = _one_or_the_same(
+                    default_python_version,
+                    module_ctx.read(default_python_version_file).strip(),
+                )
+            if default_python_version_env:
+                # Bazel version 7.1.0 and later support module_ctx.getenv(name, default):
+                # When building incrementally, any change to the value of the variable
+                # named by `name` will cause this repository to be re-fetched.
+                if "getenv" in dir(module_ctx):
+                    getenv = module_ctx.getenv
```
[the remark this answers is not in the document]
vonschultz: Done.

python/private/python.bzl
aignas: Thanks, makes sense.

python/private/python.bzl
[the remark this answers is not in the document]
aignas: We can get the path to module file of the root module in the same way and then try reading .python-version in the same dir.

python/private/python.bzl
vonschultz: There were some complications with trying to read `.python-version` unbidden, so I removed that in 5bdfbe2 (see commit message).

python/private/python.bzl
vonschultz: We can skip the surprising behavior if we don't read `.python-version` unbidden, so I stop trying that in 5bdfbe2 (see commit message).

python/private/python.bzl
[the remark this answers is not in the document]
vonschultz: The easiest approach is to just have the users say `python_version_file="//:.python-version"` if they want to opt in. That's not necessarily a bad thing, as being explicit allows anyone to grep the repo for the label and find out what cares about that file (5bdfbe2).

aignas (approved):
Thank you very much!

examples/multi_python_versions/MODULE.bazel
```diff
@@ -17,10 +17,8 @@ python.defaults(
 )
 python.toolchain(
     configure_coverage_tool = True,
-    python_version = "3.8",
-)
-python.toolchain(
-    configure_coverage_tool = True,
+    # Only set when you have mulitple toolchain versions.
+    is_default = True,
```
vonschultz, line 21: Adding back `is_default = True,` in the merge commit wasn't the intention, was it? When we're using `python.defaults`, we shouldn't also use `is_default = True`.

aignas: I probably messed up the merge conflict handling.

vonschultz: Fix in #2924, which also moves other examples and docs over to `python.defaults`.

PR #2589, opened 2025-01-28
docs: using python_version attribute for specifying python version

JeroenSchmidt:
Updates examples and docs to tell to use the base rules and the python_version attribute
instead of the wrapper transition rules.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,7 @@ Unreleased changes template.
 
 {#v0-0-0-fixed}
 ### Fixed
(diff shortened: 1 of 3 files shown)


rickeylev (approved):
They didn't permit those trivial/mechanical changes to BUILD.bazel? Well, ok then.

We still appreciate your doc contributions!

JeroenSchmidt:
Hi @rickeylev 

I left out 'examples/multi_python_versions/tests/BUILD.bazel' because I thought it was updated in https://github.com/bazelbuild/rules_python/pull/2569 and due to the comment in the last PR. I can open another one for that BUILD file if it's still needed.

Issue #2489, opened 2024-12-09
--bootstrap_impl=script breaks py_binary with py_package + py_wheel


ewianda:
# 🐞 bug report

### Affected Rule
py_package + py_wheel

### Is this a regression?
 Yes, related commit https://github.com/bazelbuild/rules_python/pull/2409

### Description
Using the resulting py_binary with py_package+py_wheel fails to find the interpreter symlink. This also breaks `pkg_tar` from rules_pkg

## 🔬 Minimal Reproduction

```diff
diff --git a/examples/wheel/BUILD.bazel b/examples/wheel/BUILD.bazel
index 58a43015..f06f6ee8 100644
--- a/examples/wheel/BUILD.bazel
+++ b/examples/wheel/BUILD.bazel
@@ -17,6 +17,7 @@ load("@bazel_skylib//rules:write_file.bzl", "write_file")
 load("//examples/wheel/private:wheel_utils.bzl", "directory_writer", "make_variable_tags")
 load("//python:packaging.bzl", "py_package", "py_wheel")
 load("//python:pip.bzl", "compile_pip_requirements")
 load("//examples/wheel/private:wheel_utils.bzl", "directory_writer", "make_variable_tags")
 load("//python:packaging.bzl", "py_package", "py_wheel")
 load("//python:pip.bzl", "compile_pip_requirements")
+load("//python:py_binary.bzl", "py_binary")
 load("//python:py_library.bzl", "py_library")
 load("//python:py_test.bzl", "py_test")
 load("//python:versions.bzl", "gen_python_config_settings")
@@ -414,3 +415,29 @@ py_console_script_binary(
     pkg = "@pypiserver//pypiserver",
     script = "pypi-server",
 )
+
+py_binary(
+    name = "main_binary",
+    srcs = ["main.py"],
+    main = "main.py",
+    deps = [
+        "//examples/wheel/lib:module_with_data",
+        "//examples/wheel/lib:simple_module",
+    ],
+)
+
+py_package(
+    name = "binary_package",
+    # Only include these Python packages.
+    packages = ["examples"],
+    deps = [":main_binary"],
+)
+
+py_wheel(
+    name = "minimal_binary",
+    distribution = "requires_files",
+    version = "0.1.0",
+    deps = [
+        ":binary_package",
+    ],
+)
```

## 🔥 Exception or Error

```bash
    runpy.run_path(main_filename, run_name="__main__")
  File "<frozen runpy>", line 291, in run_path
[...]

PR #2590, opened 2025-01-29
fix: add flag to use runtime venv creation when using bootstrap=script

rickeylev:
The bootstrap=script implementation was changed to use declare_symlink() to create explicit
symlinks so its venv works. Unfortunately, this broke packaging rules, which would treat
the symlinks as regular files.

To fix, introduce a flag that stops using declare_symlink() and instead creates the venv
at runtime. Creating a venv at runtime is problematic for various reasons, but this should
work well enough until packaging rules are able to handle these raw symlinks.

The location of the venv can be somewhat controlled by setting the `RULES_PYTHON_VENVS_ROOT`
environment variable. This is to better accommodate cases where using /tmp is problematic.

Along the way, sort the environment variable docs by their name.

Fixes https://github.com/bazelbuild/rules_python/issues/2489


rickeylev:
Ah doh, i forgot to undo some of the diffs. I need to revert .bazelversion and remove the rules_pkg stuff -- that was just debug code.

python/private/stage1_bootstrap_template.sh
```diff
@@ -186,7 +235,7 @@ command=(
 #
 # However, when running a zip file, we need to clean up the workspace after the
 # process finishes so control must return here.
-if [[ "$IS_ZIPFILE" == "1" ]]; then
+if [[ "$use_exec" == "0" ]]; then
```
aignas, line 238: I think the comment should be updated to mention the above logic.

rickeylev: Done

rickeylev:
PTAL. Ci is happy, tests added, cleaned up.

docs/environment-variables.md
```diff
@@ -68,3 +68,16 @@ When `1`, debug information about coverage behavior is printed to stderr.
 
 When `1`, debug information from gazelle is printed to stderr.
 :::
+
+:::{envvar} RULES_PYTHON_VENVS_ROOT
```
aignas, line 72: Just wondering if `RULES_PYTHON_VENVS_ROOT` is slightly ambiguous. This is the bootstrap venv and not a venv that a user can benefit from. Or can they? If I run something like:
```
RULES_PYTHON_VENVS_ROOT=${PWD}/.venv bazel run //:foo_binary -- --help
```

Then I would have an interpreter/venv that I can use to run `foo_binary` later or point it to the editor? My current understanding is that it is not the case.

Maybe having `RULES_PYTHON_BOOTSTRAP_VENV_ROOT` could be useful?

Another slightly related question - why are we using plural?

python/private/stage1_bootstrap_template.sh
```diff
@@ -121,13 +124,60 @@ if [[ "$IS_ZIPFILE" == "1" ]]; then
...
-      echo >&2 "ERROR: Python to use found on PATH: $PYTHON_BINARY_ACTUAL"
+      echo >&2 "ERROR: Python to use not found on PATH: $PYTHON_BINARY_ACTUAL"
       exit 1
     fi
   fi
   # The bin/ directory may not exist if it is empty.
   mkdir -p "$(dirname $python_exe)"
   ln -s "$symlink_to" "$python_exe"
+elif [[ "$RECREATE_VENV_AT_RUNTIME" == "1" ]]; then
+  runfiles_venv="$RUNFILES_DIR/$(dirname $(dirname $PYTHON_BINARY))"
```
aignas, line 135: please move this closer to where it is actually used.

python/private/stage1_bootstrap_template.sh
```diff
@@ -121,13 +124,60 @@ if [[ "$IS_ZIPFILE" == "1" ]]; then
...
+  python_exe="$venv/bin/$(basename $PYTHON_BINARY_ACTUAL)"
+  if [[ ! -e "$python_exe" ]]; then
+    ln -s "$symlink_to" "$python_exe"
+  fi
+  if [[ ! -e "$venv/pyvenv.cfg" ]]; then
+    ln -s "$runfiles_venv/pyvenv.cfg" "$venv/pyvenv.cfg"
+  fi
+  if [[ ! -e "$venv/lib" ]]; then
+    ln -s "$runfiles_venv/lib" "$venv/lib"
+  fi
```
aignas, line 178: Just for my understanding, I am reading that these are absolute symlinks, right? That means that if I move my runfiles dir with the launcher script, the thing should still work?

aignas (commented):
LGTM in general with a few questions.

I think that #2598 should be merged first. I am not 100% sure if the two PRs are related or not - my current understanding is that we cannot have the `venv` being set by the features introduced here and build a zip at the same time, but I would like CI to verify that.

The only question is the `env` variable name. The `_VENVS_` plurality is probably the only thing I would like to decide on before we set that in stone.

rickeylev:
> merge #2598 first

SGTM.

> I am not 100% sure if the two PRs are related or not - my current understanding is that we cannot have the `venv` being set by the features introduced here and build a zip at the same time, but I would like CI to verify that.

They're not related -- the problematic case was trying to use e.g. rules_pkg to take a non-zip py_binary and "naively" put it into a zipfile. When a zip py_binary is created, it doesn't use declare_symlink (it instead recreates _just_ the interpreter symlinks).

The confusion is understandable, though -- there's just too many permuations of outputs a py_binary produces: `build_python_zip={yes,no} x bootstrap={script,system_python}`, plus how that affects the `python_zip_file` output group, and now this flag (which only affects `build_python_zip=no, bootstrap=script`).

> The only question is the `env` variable name. The `_VENVS_` plurality is probably the only thing I would like to decide on before we set that in stone.

I picked a plural name to capture that there _could_ be multiple venvs in the directory. This could occur if the binary has another binary in the data dependencies.

Ideas for a new name? I agree the name could be a bit confusing. It basically acts as the directory that any temporary stuff will be created into. Re-using it for e.g. where a zip-py_binary extracts into seems like a reasonable thing to do (another PR for that, though).

RULES_PYTHON_EXTRACT_ROOT ?

aignas:
I like the suggestion of RULES_PYTHON_EXTRACT_ROOT.

Re confusion about the zips - thanks for the eexplanation. I guess people could workaround the current issues in the tar rules by taring the zipapp and rules_python would do the right thing when decompressing the zip file and executing it. I would not call it the best or most intuitive solution, but it is one.

python/private/stage1_bootstrap_template.sh
rickeylev: Correct, you can relocate the runfiles directory, and then run the program. This is true regardless of the `--relative_venv_symlinks` flag value.

It's not because of absolute symlinks, though. It's just because a valid venv exists in both cases. The main difference is when the venv is created.

With relative_venv_symlinks=yes, the venv is created at build time, and the bin/python3 symlink contains a runfiles-root relative path (`../../python_linux_x86/python3`), so still refers to the correct underlying interpreter.

When relative_venv_symlinks=no, the venv is created at runtime, creating a venv directory and symlinks to point to the backing paths in the runfiles. A new directory is created because the runfiles tree may not be writable. The symlinks it creates happen to be absolute, but that's just because it's easier. If it wrote relative symlinks, it would have to compute the relative path between e.g. `/tmp/whatever` and `/home/user/project/bazel-bin/foo.runfiles/blabla`, which basically equates to up-references to `/`. If the runfiles dir and that override environment are used, then it might result in a more sensible relative path. I dunno what this would benefit, though. You'd have to relocate it, run it, then package it, so you can run it elsewhere. FWIW, there is a FR for this for the zip-builds (IS_ZIPFILE=1) startup (but something for another PR)

rickeylev (commented):
Bah, I don't like that relative_venv_symlinks flag name anymore. I think I'll rename it.

It's not really about whether a relative path is used in the venv. It's about whether declare_symlink() with a "raw" path is used, which requires the runfiles to have a particular layout.

`--build_venv_at_runtime=yes|no` ; `--raw_venv_symlinks=yes|no` ?

re: RULES_PYTHON_EXTRACT_ROOT -- SGTM, I'll rename it.

re: tar a zipapp: Yes, that'd work. The zipapps are a bit annoying because they _always_ extract to a tmp dir. This adds startup overhead and /tmp files can get cleaned up (which long lived processes don't like).

rickeylev:
PTAL

Renamed to RULES_PYTHON_EXTRACT_ROOT

Renamed to `--venvs_use_declare_symlink`

docs/environment-variables.md
```diff
@@ -1,28 +1,53 @@
 # Environment Variables
```
aignas, line 1: I assume this got sorted and nothing else.

python/private/stage1_bootstrap_template.sh
```diff
@@ -121,13 +124,60 @@ if [[ "$IS_ZIPFILE" == "1" ]]; then
...
   # The bin/ directory may not exist if it is empty.
   mkdir -p "$(dirname $python_exe)"
   ln -s "$symlink_to" "$python_exe"
+elif [[ "$RECREATE_VENV_AT_RUNTIME" == "1" ]]; then
+  if [[ -n "$RULES_PYTHON_EXTRACT_ROOTT" ]]; then
+    use_exec=1
+    # Use our runfiles path as a unique, reusable, location for the
+    # binary-specific venv being created.
+    venv="$RULES_PYTHON_EXTRACT_ROOTT/$(dirname $(dirname $PYTHON_BINARY))"
+    mkdir -p $RULES_PYTHON_EXTRACT_ROOTT
```
aignas, line 140: Typo:
```suggestion
  if [[ -n "$RULES_PYTHON_EXTRACT_ROOT" ]]; then
    use_exec=1
    # Use our runfiles path as a unique, reusable, location for the
    # binary-specific venv being created.
    venv="$RULES_PYTHON_EXTRACT_ROOT/$(dirname $(dirname $PYTHON_BINARY))"
    mkdir -p $RULES_PYTHON_EXTRACT_ROOT
```

tests/bootstrap_impls/run_binary_venvs_use_declare_symlink_no_test.sh
```diff
@@ -0,0 +1,56 @@
...
+    touch EXPECTATION_FAILED
+    return 1
+  fi
+}
+
+expect_match "sys.executable:.*tmp.*python3" "$actual"
+
+venvs_root=$(mkdir -d)
+
+actual=$(RULES_PYTHON_EXTRACT_ROOT=$venvs_root $bin)
```
aignas, line 52: Since there was a typo, was this test working?

docs/environment-variables.md
rickeylev: Mostly, yes. Sorry, forgot to mention that. I've updated the description to note this, too.

I expanded the RULES_PYTHON_EXTRACT_ROOT docs a bit, but the same in spririt.

tests/bootstrap_impls/run_binary_venvs_use_declare_symlink_no_test.sh
rickeylev: Yes. Fixed (it was calling mkdir instead of mktemp).

python/private/stage1_bootstrap_template.sh
rickeylev: Doh. Fixed.

rickeylev:
Ok, done changes and fixed some doc types i saw along the way. PTAL

aignas (approved):
Thanks!

PR #2591, opened 2025-01-29
tests: don't propagage inner py_binary env to outer sh_test

rickeylev:
Per conversation in #2507, don't propagate the inner py_binary env to the outer sh_test.
This is because the sh-side expects to call the Python program with a particular environment
state, and it can't do that if the inner Python program is secretly setting things
in the environment.


rickeylev:
Reverting -- I was mistaken in what was being propagated elsewhere. Code is fine as-is.

Issue #2575, opened 2025-01-23
`bazel coverage` fails when using certain Python packages


BurnzZ:
# 🐞 bug report

### Affected Rule

`py_test`

### Is this a regression?

Not that I'm aware of.

### Description

`bazel coverage` fails without any useful messages when some Python packages are being used/imported. Some examples:

- `torchvision`
- `transformers.models.distilbert.DistilBertModel`

## 🔬 Minimal Reproduction

See https://github.com/BurnzZ/bazel-python-coverage-issue for the full code example.

1. Clone the repo:

```
git clone https://github.com/BurnzZ/bazel-python-coverage-issue
cd bazel-python-coverage-issue
```

2. Open `test.py` and uncomment either of these lines:

```python
# import torchvision
# from transformers.models.distilbert import DistilBertModel
```

3. Run the following:

```
bazel coverage --combined_report=lcov :test --nocache_test_results --test_output=all
lcov --list "$(bazel info output_path)/_coverage/_coverage_report.dat"
```

4. Notice that test coverage fails without useful context or error messages.

5. Run the following and notice that the code should run successfully when tested:

```
bazel test :test --nocache_test_results --test_output=all
```

Try commenting out the imports above and run `bazel coverage` and it should work. Curious as to why it doesn't work in some package like `torchvision` or some parts of `transformers`.

## 🔥 Exception or Error

```
$ bazel coverage --combined_report=lcov :test --nocache_test_results --test_output=all

INFO: Using default value for --instrumentation_filter: "^//".
INFO: Override the above default with --instrumentation_filter
INFO: Analyzed target //:test (0 packages loaded, 0 targets configured).
FAIL: //:test (Exit 1) (see /private/var/tmp/_bazel_user/c97d0f59e3791eddf9709b879355cbf5/execroot/_main/bazel-out/darwin_arm64-fastbuild/testlogs/test/test.log)
INFO: From Testing //:test:
==================== Test output for //:test:
--
Coverage runner: Not collecting coverage for failed test.
The following commands failed with status 1
/private/var/tmp/_bazel_user/c97d0f59e3791eddf9709b879355cbf5/sandbox/darwin-sandbox/7/execroot/_main/bazel-out/darwin_arm64-fastbuild/bin/test.runfiles/_main/test
================================================================================
[...]

PR #2597, opened 2025-02-01
feat: new COVERAGEPY_IGNORE_ERRORS env var

BurnzZ:
## Background

This is a proposal to have a workaround for https://github.com/bazelbuild/rules_python/issues/2575 where an inherent issue from [coverage.py](https://coverage.readthedocs.io/en/7.6.10/) when creating the lcov output file.

## Proposed Changes

An env var flag that a user can turn on via `--test_env=COVERAGEPY_IGNORE_ERRORS=true` when running `bazel coverage`.

## Reproducible steps

We can see how this proposal addresses the issue described from https://github.com/BurnzZ/rules_python.

1. Clone the repo: https://github.com/BurnzZ/rules_python

```
git clone https://github.com/BurnzZ/rules_python.git rules_python_tmp
cd rules_python_tmp/examples
git checkout feat-new-env-var-COVERAGEPY_IGNORE_ERRORS
```

2. Clone the reproducible example

```
git clone https://github.com/BurnzZ/bazel-python-coverage-issue.git
cd bazel-python-coverage-issue
git checkout enable-torchvision
```

3. Add these lines in line 2 of `MODULE.bazel`:

```
local_path_override(
    module_name = "rules_python",
    path = "../..",
)
```

4. Generate the coverage report:

```
bazel coverage --combined_report=lcov :test --nocache_test_results --test_output=all --test_env=COVERAGEPY_IGNORE_ERRORS=true
lcov --list "$(bazel info output_path)/_coverage/_coverage_report.dat"
genhtml --branch-coverage --output genhtml "$(bazel info output_path)/_coverage/_coverage_report.dat"
```

Issue from https://github.com/bazelbuild/rules_python/issues/2575 now resolved. 🎉

### DIFF
--- a/python/private/python_bootstrap_template.txt
+++ b/python/private/python_bootstrap_template.txt
@@ -461,7 +461,8 @@ relative_files = True
     kparams['stdout'] = sys.stderr
     kparams['stderr'] = sys.stderr
 
-  # Workaround for coveragepy having weird errors.
+  # Ignore errors finding source files during coverage.py reporting
+  # See https://coverage.readthedocs.io/en/7.6.10/cmd.html#reporting
   # See: https://github.com/bazelbuild/rules_python/issues/2575
   if os.getenv("COVERAGEPY_IGNORE_ERRORS"):
     params.append("--ignore-errors")
--- a/python/private/python_bootstrap_template.txt
+++ b/python/private/python_bootstrap_template.txt
@@ -461,6 +461,11 @@ relative_files = True
     kparams['stdout'] = sys.stderr
     kparams['stderr'] = sys.stderr
 
+  # Workaround for coveragepy having weird errors.
+  # See: https://github.com/bazelbuild/rules_python/issues/2575
+  if os.getenv("COVERAGEPY_IGNORE_ERRORS"):
+    params.append("--ignore-errors")
+
   ret_code = subprocess.call(
     params,
     **kparams
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,7 +78,8 @@ Unreleased changes template.
 
 {#v0-0-0-added}
 ### Added
-* Nothing added.
+* (coverage) New COVERAGEPY_IGNORE_ERRORS environment var that adds the `--ignore-errors`
+  flag to coverage.py when producing the lcov output file.
 
 {#v0-0-0-removed}
 ### Removed


groodt:
I think it makes sense to provide a mechanism to coveragepy to ignore errors. 

I also wonder if it makes sense to provide a way for all of the coveragepy CLI arguments to be supplied?

I think the torch errors that motivate this PR can probably also be avoided by excluding that source from coverage. Ignoring all errors would avoid the torch error, but possibly also mask others. 

To be clear, I still think there are valid scenarios to ignore all, so this PR still LGTM. 

WDYT @rickeylev

python/private/python_bootstrap_template.txt
```diff
@@ -461,6 +461,11 @@ relative_files = True
     kparams['stdout'] = sys.stderr
     kparams['stderr'] = sys.stderr
 
+  # Workaround for coveragepy having weird errors.
```
groodt, line 464:
```suggestion
  # Ignore errors finding source files during coverage.py reporting
  # See https://coverage.readthedocs.io/en/7.6.10/cmd.html#reporting
```

groodt (approved):
LGTM

BurnzZ:
> I also wonder if it makes sense to provide a way for all of the coveragepy CLI arguments to be supplied?

This was initially my approach but since there are 2 runs: `coverage run` and `coverage lcov`, we need to consider that the args could be exclusive to each. On the other hand, there might be cases that they should be the same to ensure proper processing of the coverage artifacts. I haven't thought deeply about the other use cases for this yet.

> I think the torch errors that motivate this PR can probably also be avoided by excluding that source from coverage. 

From https://github.com/nedbat/coveragepy/issues/1921#issuecomment-2629154773, I think an alternative would be `coverage run --omit="$TMPDIR/*" test_foo.py`. Will need to properly test this out.

> Ignoring all errors would avoid the torch error, but possibly also mask others.

Yeah, that's indeed another risk. 

Although, I think we should be fine since `--ignore-errors` _"tells coverage.py to ignore problems encountered trying to find source files to report on. This can be useful if some files are missing, or if your Python execution is tricky enough that file names are synthesized without real source files"_ ([reference](https://coverage.readthedocs.io/en/7.6.10/cmd.html#reporting)). We still get useful warning messages (unless `--disable_warning` is added).

rickeylev:
How about enabling ignore errors by default instead?

BurnzZ:
That's one option but it could hide other potential issues _by default_.

I think it'd be best if the user can first see the error, then decide if errors are to be ignored. This _opt-in_ approach is sufficient for the time being.

timothy-tlj-jones-canva (approved):
I heartily endorse this

aignas:
Is this needed if we have #2607?

BurnzZ:
We can close this since https://github.com/bazelbuild/rules_python/pull/2607 is on the way. 

Although there are some benefits in having an escape hatch like `--ignore-errors` could provide outside of https://github.com/bazelbuild/rules_python/issues/2575. We can revisit if needed.

Issue #2596, opened 2025-01-31
`--bootstrap_impl=script` breaks python_zip_file files


philsc:
# 🐞 bug report

### Affected Rule

 The issue is caused by the rule: 
The py_binary's `python_zip_file` output doesn't work anymore when executed.

### Is this a regression?

Without `--bootstrap_impl=script`, it seems to work fine.

### Description
Setting the flag breaks running the zip files.

## 🔬 Minimal Reproduction
See https://github.com/philsc/rules_python-par-failure
```skylark
py_binary(
    name = "bin",
    srcs = ["bin.py"],
)

filegroup(
    name = "bin_zip_filegroup",
    srcs = [":bin"],
    output_group = "python_zip_file",
)

genrule(
    name = "_bin.par",
    srcs = [":bin_zip_filegroup"],
    outs = ["bin.par"],
    cmd = " && ".join([
        "echo '#!/usr/bin/env python3' > $(OUTS)",
        "cat $(SRCS) >> $(OUTS)",
    ]),
    executable = True,
)
```

## 🔥 Exception or Error

```console
$ bazel run //:bin.par
INFO: Invocation ID: b3ca5fcf-e346-441b-92e3-e15b7ce33772
INFO: Analyzed target //:bin.par (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
Target //:bin.par up-to-date:
  bazel-bin/bin.par
INFO: Elapsed time: 0.368s, Critical Path: 0.00s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
INFO: Running command line: bazel-bin/bin.par
Traceback (most recent call last):
  File "/usr/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/data/phil_cache/bazel/_bazel_philipp.schrader/7bd7572ee132ae5e69d828af72559172/execroot/_main/bazel-out/k8-fastbuild/bin/bin.par/__main__.py", line 323, in <module>
  File "/data/phil_cache/bazel/_bazel_philipp.schrader/7bd7572ee132ae5e69d828af72559172/execroot/_main/bazel-out/k8-fastbuild/bin/bin.par/__main__.py", line 312, in main
[...]

PR #2598, opened 2025-02-01
fix: make plain zipapp work with bootstrap=script

rickeylev:
The `__main__.py` template (zip_main_template.py) was using the wrong path when creating
the interpreter symlinks. It as computing it correctly, just the wrong variable was used
in the symlink() call.

To fix, pass the correct variable.

Also adds a test to check that it's runnable.

Fixes https://github.com/bazelbuild/rules_python/issues/2596

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,8 @@ Unreleased changes template.
 
 {#v0-0-0-fixed}
 ### Fixed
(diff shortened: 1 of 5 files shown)


aignas (approved):
Thanks for the test.

The only thing left is a Changelog note about the fixed issue.

rickeylev:
Changelog updated

PR #2599, opened 2025-02-03
fix(coverage): prevent issues when packages execute files in tmp dir.

BurnzZ:
## Background

(A good addition to https://github.com/bazelbuild/rules_python/pull/2597)

This is a proposal to have a workaround for https://github.com/bazelbuild/rules_python/issues/2575 where an inherent issue from [coverage.py](https://coverage.readthedocs.io/en/7.6.10/) when creating the lcov output file.

## Proposed Changes

Add an `--omit='...'` flag pointing to tempdir.

## Reproducible steps

We can see how this proposal addresses the issue described from https://github.com/BurnzZ/rules_python.

1. Clone the repo: https://github.com/BurnzZ/rules_python

```
git clone https://github.com/BurnzZ/rules_python.git rules_python_tmp
cd rules_python_tmp/examples
git checkout fix-omit-tmpdir-coveragepy
```

2. Clone the reproducible example

```
git clone https://github.com/BurnzZ/bazel-python-coverage-issue.git
cd bazel-python-coverage-issue
git checkout enable-torchvision
```

3. Add these lines in line 2 of `MODULE.bazel`:

```
local_path_override(
    module_name = "rules_python",
    path = "../..",
)
```

4. Generate the coverage report:

```
bazel coverage --combined_report=lcov :test --nocache_test_results --test_output=all
lcov --list "$(bazel info output_path)/_coverage/_coverage_report.dat"
genhtml --branch-coverage --output genhtml "$(bazel info output_path)/_coverage/_coverage_report.dat"
```

### DIFF
--- a/python/private/python_bootstrap_template.txt
+++ b/python/private/python_bootstrap_template.txt
@@ -461,6 +461,9 @@ relative_files = True
     kparams['stdout'] = sys.stderr
     kparams['stderr'] = sys.stderr
 
+  # Prevent coveragepy report issues when packages execute files in tmp dir.
+  params.append('--omit="$TMPDIR/*"')
+
   ret_code = subprocess.call(
     params,
     **kparams
--- a/python/private/python_bootstrap_template.txt
+++ b/python/private/python_bootstrap_template.txt
@@ -462,7 +462,9 @@ relative_files = True
     kparams['stderr'] = sys.stderr
 
   # Prevent coveragepy report issues when packages execute files in tmp dir.
-  params.append('--omit="$TMPDIR/*"')
+  import tempfile
+  omit_pattern = os.path.join(tempfile.gettempdir(), "*")
+  params.append(f'--omit={omit_pattern}')
 
   ret_code = subprocess.call(
     params,
--- a/python/private/python_bootstrap_template.txt
+++ b/python/private/python_bootstrap_template.txt
@@ -462,6 +462,7 @@ relative_files = True
     kparams['stderr'] = sys.stderr
 
   # Prevent coveragepy report issues when packages execute files in tmp dir.
+  # See https://github.com/nedbat/coveragepy/issues/1921#issuecomment-2629154773.
   import tempfile
   omit_pattern = os.path.join(tempfile.gettempdir(), "*")
   params.append(f'--omit={omit_pattern}')
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,6 +77,8 @@ Unreleased changes template.
   The related issue is [#908](https://github.com/bazelbuild/rules_python/issue/908).
 * (sphinxdocs) Do not crash when `tag_class` does not have a populated `doc` value.
   Fixes ([#2579](https://github.com/bazelbuild/rules_python/issues/2579)).
+* (coverage) Prevent coveragepy report issues when packages execute files in tmp dir.
+  Fixes ([#2575](https://github.com/bazelbuild/rules_python/issues/2575)).
 
 {#v0-0-0-added}
 ### Added


python/private/python_bootstrap_template.txt
```diff
@@ -461,6 +461,9 @@ relative_files = True
     kparams['stdout'] = sys.stderr
     kparams['stderr'] = sys.stderr
 
+  # Prevent coveragepy report issues when packages execute files in tmp dir.
+  params.append('--omit="$TMPDIR/*"')
```
BurnzZ, line 465: Ref: https://github.com/nedbat/coveragepy/issues/1921#issuecomment-2629154773

I can confirm that passing `--omit="$TMPDIR/*"` would have coverage.py interpolate the env var correctly. I can reproduce this outside of Bazel.

However, when having it inside of Bazel, it doesn't seem to work _(still getting the "No source for code: ..." error)_. Even if I do:

```python
tmpdir = os.getenv('TMPDIR')
if tmpdir:
    params.extend(['--omit', f'{tmpdir}/*'])
```

File pattern seems to be correct as well: https://coverage.readthedocs.io/en/7.6.10/source.html#file-patterns.

Also tried including the env vars `GTEST_TMP_DIR` and `TEST_TMPDIR` but to no avail.

I might be missing some other intermediate tmp dir that Torch utilizes when inside Bazel. 🤔

rickeylev, line 465: subprocess.call won't exapnd environment variables. So something like this:

You want `tempfile.gettempdir()`
```suggestion
  import tempfile
  omit_pattern = os.path.join(tempfile.gettempdir(), "*")
  params.append(f'--omit={omit_pattern}')
```

BurnzZ: Ahhh brilliant! Cheers @rickeylev !

BurnzZ: Fixed by @rickeylev : https://github.com/bazelbuild/rules_python/pull/2599/commits/91d6cbede1629383ee479ccbc866dd0106e35f7c

python/private/python_bootstrap_template.txt
```diff
@@ -461,6 +461,11 @@ relative_files = True
     kparams['stdout'] = sys.stderr
     kparams['stderr'] = sys.stderr
 
+  # Prevent coveragepy report issues when packages execute files in tmp dir.
```
timothy-tlj-jones-canva, line 464: For bonus points, you could link the github issue(s) that this comment is referencing

BurnzZ: Good call. Added in https://github.com/bazelbuild/rules_python/pull/2599/commits/312d9adfe59958319fd162f54e5269f7da43d79c.

aignas:
My thinking is that the `COVERAGE_RC` issue was a similar topic - we need to be able to customize coverage in some way but currently we have no way to do that. The coverage tool right now is coupled to the `python` toolchain, but maybe we should have a `python_test` toolchain that can touch/configure the following parts:
* Test runner deps
* Test runner itself
* Coverage deps
* Coverage configuration
* Report generation?

We have a proposal here for this: #2246.

Some thoughts about this change:
* We may want to be able to change coverage configuration per target, similar to the `precompile` flag.
* `rules_python` should do the right thing out of the box.

I think there are many other things I am forgetting, but wanted to mention greater design concerns here. I am not too much against this as a temporary fix for the actual problem at hand, but I would like to have a better story for testing and configuration in `rules_python` in general, so if other people have requirements as to what is needed and/or ideas about design, please share them.

groodt:
> we need to be able to customize coverage in some way but currently we have no way to do that

Yes, indeed. There are also solutions to this which might be solved in userspace instead. Similar to https://pypi.org/project/pytest-bazel/

I do like a userspace solution, but the "batteries included" coverage supporting configuration files makes sense while we are bundling `coverage` with the rules.

BurnzZ:
Noting that this proposed PR might not be needed since the PR in https://github.com/bazelbuild/rules_python/pull/2607 that adds `--source` to the coverage.py call would also fix the issue in https://github.com/bazelbuild/rules_python/issues/2575. As per https://github.com/nedbat/coveragepy/issues/1921#issuecomment-2629154773, using either `--source` or `--omit` would fix the issue.

We can close this (alongside https://github.com/bazelbuild/rules_python/issues/2575) once https://github.com/bazelbuild/rules_python/pull/2607 is merged.

PR #2600, opened 2025-02-03
refactor: expose base rule construction via builders to allow customization for testing

rickeylev:
The py_reconfig rules work by wrapping: The outer reconfig rule applies a transition, depends
on an inner py base rule, then jumps through various hoops to ensure it looks and acts
like the target it's wrapping.

This is error prone, incomplete, and annoying code to maintain. Phil recently discovered it wasn't properly propagating the output group, so he had to add that. I wasted time trying to fix a bug I _thought_ was in it, but actually was working correctly. The logic within it is a bit hacky as it tries to emulate some of the platform-specific stuff for windows. Every time
py_reconfig gains something to transition on, there's numerous places to define, propagate,
and extract the pieces necessary to do it.

To fix this, make the py_reconfig rules not wrap an inner base py rule. Instead, they use the same underlying rule args that the base rules do. This lets them act directly as the rule they're designed to test.

Customization is done by capturing all the rule args in builder objects. The py_reconfig code constructs the same builder the base rules do, then modifies it as necessary (adding attributes, wrapping the base transition function). As a bonus, this sets some ground work to allow more easily defining derivative rules without having to copy/paste arbitrary parts of how the base rules are defined.


tests/support/sh_py_run_test.bzl
```diff
@@ -18,162 +18,77 @@ without the overhead of a bazel-in-bazel integration test.
 """
 
 load("@rules_shell//shell:sh_test.bzl", "sh_test")
-load("//python:py_binary.bzl", "py_binary")
-load("//python:py_test.bzl", "py_test")
+load("//python/private:py_binary_macro.bzl", "py_binary_macro")  # buildifier: disable=bzl-visibility
+load("//python/private:py_binary_rule.bzl", "create_binary_rule_builder")  # buildifier: disable=bzl-visibility
```
aignas, line 22: We had a ticket to have such an API, no? Is #1647 about it? We could actually expose the `create_binary_rule_builder` with a statement that this API is very early preview and might change in the subsequent releases.

If we don't want to do it now, we can do it at a later time, but getting feedback from people in that ticket could be a good way to stabilise this.

aignas (approved):

tests/support/sh_py_run_test.bzl
aignas: Doh, I now read the reference in the description. The PR is good as it is now.

rickeylev: Yes, read my mind! I think the two things to do before that are: use builders for all the args (notably attrs) and create directories in `python/api/` (bzl-per-rule) to be the public access points.

aignas: SGTM

PR #2610, opened 2025-02-14
refactor: API for deriving customized versions of the base rules

rickeylev:
This implements a "builder style" API to allow arbitrary modification of rule, attr, etc
objects used when defining a rule. The net effect is users are able to use the base
definition for our rules, but define their own with the modifications they need, without
having to copy/paste portions our implementation, load private files, or patch source.

The basic way it works is a mutable object ("builder") holds the args and state that would
be used to create the immutable Bazel object. When `build()` is called, the immutable
Bazel object (e.g. `attr.string()`) is created. Builders are implemented for most objects
and their settings (rule, attrs, and supporting objects).

This design is necessary because of three Bazel behaviors:
* attr etc objects are immutable, which means we must keep our own state
* attr etc objects aren't inspectable, which means we must store the arguments for
  creating the immutable objects.
* Starlark objects are frozen after initial bzl file evaluation, which means creation
  of any mutable object must be done at the point of use.

The resulting API resembles the builder APIs common in other languages:

```
r = create_py_binary_rule_builder()
r.attrs.get("srcs").set_mandatory(True)
r.attrs.get("deps").aspects().append(my_aspect)
my_py_binary = r.build()
```

Most objects are thin wrappers for managing a kwargs dict. As such, and because they're
wrapping a foreign API, they aren't strict in enforcing their internal state and the
kwargs dict is publicly exposed as an escape hatch.

As of this PR, no public API for e.g. `create_py_binary_rule_builder()` is exposed. That'll come in a separate PR (to add public access points under python/api).


rickeylev:
The diff here is pretty big, so I'll summarize the changes:

In attributes.bzl, the `dict[name, Attribute]` dicts are changed to `dict[name, <lambda to create builder>]`. Similar for the other dicts of attributes elsewhere. This is "necessary" because of the immutability rules of bazel.

In rule_builders.bzl, all the various builder classes to wrap the inputs to rule, aspect, attr.XXX, etc etc are implemented. It's standard "object oriented" starlark code.

Most of the rest of the changes are just whitespace, some incidental moving of code, or notes.

I put "necessary" in quotes earlier because it's necessary in the sense of providing a mutable object. It's not necessary in the sense of: I think most things about a rule won't be modified. So it feels a bit wasteful.

With that in mind, I did prototype another idea: store both a built object and a callable to create a builder for it. e.g. `struct(built=<Attribute> new_builder=lambda: FooBuilder(...))`. This worked, but the API on the user side was really annoying (you had to convert the immutable to mutable, modify it, then reassign it where it came from). You can see how that lambda for the builder is still there.

I also prototyped using regular ol' dicts throughout. This largely eliminated the builder wrappers, but dicts also had to have a special `@constructor` key so e.g. attributes could call the right attr.xxx function. The UX of this was mixed. Just imagine a big nested dict of all the stuff that gets fed into rule(), transition(), attr.xxx, etc, etc. You end up doing a lot of manual "book keeping" interacting with them. Same as before, the lambdas for the "builders" are still there, they just create plain dicts instead of structs.

I also briefly explored using a visitor pattern, but no, that was really bad UX.

There's really no escaping having a function to create "builders" of some sort.

Polymorphic attributes (rule.cfg, attr.cfg take several different data types), and things like providers/aspects (lists of objects that aren't easily identified/introspected) were also complications.

Those complications eventually lead me to lean more into the builder pattern, my reasoning being: at least we control the API surface and can make it mostly ergonomic.

It leaves me pretty unhappy with the internal details of what attributes.bzl looks like, though. A bunch of `dict[name, lambda]` being merged into one another feels very meh.

In the end, I'm sort of happy with this? Not as happy as I was hoping to be. I'm tempted to shelve it and look for another way to expose a way to create a derived rule.

Maybe keep the current `def create_xx_rule(...)` functions, but expose functions to create each attribute. So if you want to overide, you do `create_xx_rule(attrs={"srcs": create_srcs(...)})`. Ironically, this means having a function for creating each attribute, which, yep, puts us back to where we started.

python/private/attributes.bzl
```diff
@@ -253,25 +242,24 @@ def _create_native_rules_allowlist_attrs():
...
         # buildifier: disable=attr-license
-        "licenses": attr.license() if hasattr(attr, "license") else attr.string_list(),
+        ##"licenses": attr.license() if hasattr(attr, "license") else attr.string_list(),
     },
-    allow_none = True,
 )
 
 IMPORTS_ATTRS = {
-    "imports": attr.string_list(
+    "imports": lambda: rule_builders.StringListAttrBuilder(
```
aignas, line 262: I like this definition of attrs. Having the values as lambdas feels somewhat intuitive.

python/private/attributes.bzl
```diff
@@ -406,35 +394,65 @@ as part of a runnable program (packaging rules may include them, however).
...
+
 # Attributes specific to Python executable-equivalent rules. Such rules may not
 # accept Python sources (e.g. some packaged-version of a py_test/py_binary), but
 # still accept Python source-agnostic settings.
-AGNOSTIC_EXECUTABLE_ATTRS = union_attrs(
+AGNOSTIC_EXECUTABLE_ATTRS = dicts.add(
     DATA_ATTRS,
     {
-        "env": attr.string_dict(
+        "env": lambda: rule_builders.StringDictAttrBuilder(
```
aignas, line 455: Nit, instead of rule_builders, should we call it pyrule_builders? Is there anything python specific in our code? If not, we could keep it outside //python, similarly to how sphinxdocs live.

rickeylev (commented):
It sounds like we both like 1a+2b (struct-api with everything mutable by default), so'll i'll clean things up to do it that way.

Thanks for looking and thinking!

python/private/attributes.bzl
aignas: I am not sure about the maintenance status of `bazel_skylib`, but it does seem that having some of these copied/published when they are mature could be handy. 

That said, we can do the refactor later and move `repo_utils` and `text_utils` and this into this new location.

rickeylev:
I expanded this into the struct-with-mutables style API. Notable changes:

* Per chat: removed Attr, Builder suffixes; split out attr_builders.bzl. Much nicer.
* `attrb, ruleb` names: I found `attr_builders.XXX` to be pretty repetitively verbose. I accidentally typed `attrb` and then liked it. The "b" standing for "builder".
* Added docs using the sphinxdocs TYPEDEF  feature: preview: https://rules-python--2610.org.readthedocs.build/en/2610/api/rules_python/python/private/rule_builders.html#Rule
* Added `Value`; it's like Optional, but without the `.present()` and won't fail if accessed.
* Replaced `Optional` with `Value` for most attributes. Only `allow_files` and `allow_single_file` actually **need** Optional-type behavior (those two args are mutually exclusive, even if their values are compatible)
* Added implementations for all other attributes
* Added builders for toolchain type and exec group.

The main thing missing now are tests, I think.

Unfortunately, I accidentally amended somewhere in my commits and had to force-push, so the diff is "reset".

rickeylev:
I started adding tests and three things became kind of annoying: polymorphic fields, Optional, and `.extra_kwargs`.

Label's allow_files is particularly annoying because its `bool | list[str]` _and_ mutually exclusive with `allow_single_file`. Interacting with this is tedious: I can't append a value unless I know it's a list. I can't set it to a list because that could overwrite existing values. Then I have to ensure allow_single_file isn't set. So one must do:

```
if not obj.allow_files.present() or not types.is_list(obj.allow_files.get()):
  obj.allow_files.set([])
  obj.allow_single_file.clear()
obj.allow_files.get().append('.txt')
```

It took me 3 tries to get that code above right.

The `default` field has a similar problem; being a mutable collection or some non-collection value. e.g. LabelList's default is `list | callable`, similar for LabelStringDict, etc.

The `extra_kwargs` escape hatch also became awkward. e.g. `foo.extra_kwargs[x] = y` might be set, but `foo.x.whatever()` won't reflect that state at all. Which is kind of ok (it's an escape hatch to do arbitrary overrides), but always bothered me. Writing tests made it more salient that its confusing.

------

So I'm changing the API and impl to:
* Mostly drop Value and Optional entirely. getters/setters are used instead (`doc()/set_doc()`). This makes it easier to "intercept" changes and keep the whole AttributeBuilder in a correct state. I also like that the conceptual overhead of Value/Optional go away.
* Change `self.extra_kwargs` to `self.kwargs`. Under the hood, `self.kwargs` acts to hold the mutable state for the object (h/t to 2578 for this idea). This neatly unifies the "escape hatch" with the regular AttributeBuilder APIs. I also generally like that it's conceptually like Python's `foo.__dict__`, which holds the local state of `foo`. These builders are, essentially, trying to be convenience wrappers around the kwarg dicts that we want to be mutable.

In combination, it means being able to do e.g.
```
obj = LabelList()
obj.add_allow_files(".py")
obj.add_default("//some:target")
obj = StringKeyedLabelDict()
obj.add_default("key", "//foo:bar")
```

And the functions can transparently handle if the previous state was a a bool, function, etc. It also means None becomes the value which means "missing", which feels more natural. e.g. `obj.allow_files()` returns None and, under the hood, that state is used to keep it omitted from the final kwargs dict generation. So overall, it feels more ergonomic.

python/private/attributes.bzl
```diff
@@ -267,11 +222,10 @@ COMMON_ATTRS = union_attrs(
         # buildifier: disable=attr-license
         "licenses": attr.license() if hasattr(attr, "license") else attr.string_list(),
     },
-    allow_none = True,
 )
 
 IMPORTS_ATTRS = {
-    "imports": attr.string_list(
+    "imports": lambda: attrb.StringList(
```
aignas, line 228: hmmm... the `key: lambda: attrb.StringList(` does look funny :D

python/private/py_binary_rule.bzl
```diff
@@ -50,7 +33,6 @@ def create_binary_rule_builder():
         executable = True,
     )
     builder.attrs.update(AGNOSTIC_BINARY_ATTRS)
-    builder.attrs.update(_COVERAGE_ATTRS)
```
aignas, line 53: Good to remove it from here, having to add these attrs is really hard as a user.

aignas (commented):
Will look at the biggest files later.

python/private/attributes.bzl
rickeylev: Yeah, I don't see a way around it, though.

Unless, attrb.StringList() returned a lambda? ehhhh

```
ATTRS = {
    "imports": attrb.StringList(...)
}
```

That `ATTRS["imports"]` is a lambda is a bit surprising. The "imperative" case then becomes a bit convoluted:

```
  srcs = attrb.StringList()
  srcs = srcs()
  srcs.set_whatever()
```

It'd be simple enough to have two structs, one that returns a lambda, and another that didn't? e.g. `attrcb.StringList()`

I feel like the explicit lambdas are better overall. They help make explicit that there's a callback.

python/private/py_runtime_rule.bzl
```diff
@@ -188,19 +188,21 @@ py_runtime(
...
 
 If not set, then it will be set based on flags.
 """,
-        ),
-        "bootstrap_template": attr.label(
-            allow_single_file = True,
-            default = DEFAULT_BOOTSTRAP_TEMPLATE,
-            doc = """
+            ),
+            "bootstrap_template": attr.label(
```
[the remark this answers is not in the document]
rickeylev: Correct

tests/support/sh_py_run_test.bzl
```diff
@@ -54,9 +55,9 @@ _RECONFIG_OUTPUTS = _RECONFIG_INPUTS + [
 _RECONFIG_INHERITED_OUTPUTS = [v for v in _RECONFIG_OUTPUTS if v in _RECONFIG_INPUTS]
 
 _RECONFIG_ATTRS = {
-    "bootstrap_impl": attr.string(),
-    "build_python_zip": attr.string(default = "auto"),
-    "extra_toolchains": attr.string_list(
+    "bootstrap_impl": attrb.String(),
+    "build_python_zip": attrb.String(default = "auto"),
+    "extra_toolchains": attrb.StringList(
```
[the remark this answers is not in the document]
rickeylev: I'd rather keep the TitleCase to better call out that these aren't the same as a "regular" attribute object. `attr.string_list()` and `attrb.string_list()` have only a one character difference, so it would be easy to miss. This would be compounded when there's a mixture of attr.xxx and attrb.xxx calls around (seems likely -- not every rule needs to be highly extendable)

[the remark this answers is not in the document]
aignas: Yeah, I am OK with having this. After all this is an API that we control and we could easily have `attrb.string` in the future if we decided to.

python/private/attributes.bzl
aignas: Explicit is better than implicit, definitely.

python/private/attr_builders.bzl
```diff
@@ -0,0 +1,1284 @@
...
+        {type}`AttrCfg`
+    """
+    state = {
+        "inputs": inputs,
+        "outputs": outputs,
+        # Value depends on "which" key
+        # For which=impl, the value is a function or arbitrary object
+        "value": True,
+        # str: target, exec, none, or impl
+        "which": "target",
```
aignas, line 87: nit: we could have an enum here? That would make the code read a little bit easier.

python/private/attr_builders.bzl
```diff
@@ -0,0 +1,1284 @@
...
+        self: implicitly added.
+        exec_group: {type}`str | None` the exec group name to use, if any.
+    """
+    self._state["which"] = "exec"
+    self._state["value"] = exec_group
+
+def _AttrCfg_set_target(self):
+    """Sets to use the target transition."""
+    self._state["which"] = "target"
+    self._state["value"] = True
```
aignas, line 211: nit: the `_state` keys should be constants in this file. Much easier to avoid typos.

python/private/rule_builders.bzl
```diff
@@ -0,0 +1,673 @@
...
+
+def _ExecGroup_build(self):
+    kwargs = dict(self.kwargs)
+    if kwargs.get("toolchains"):
+        kwargs["toolchains"] = [
+            v.build() if hasattr(v, "build") else v
+            for v in kwargs["toolchains"]
+        ]
+    if kwargs.get("exec_compatible_with"):
+        kwargs["exec_compatible_with"] = [
```
aignas, line 159: This file has similar comments to the `attr_builders` - mainly string constants?

aignas (approved):
The comments I have are not blocking. Thanks for designing this. This looks like an easy to maintain solution.

python/private/attr_builders.bzl
rickeylev: Will do.

python/private/rule_builders.bzl
rickeylev: Will do, mostly. I'm not a huge fan of creating symbolic constants for _every_ string literal because it can harm readability, but you have a good point -- many of these strings are repeated many times across many functions, sometimes in "optional" contexts (dict.get, key in dict) where typos would be easy to miss. I rewrote a few to use constants and it is still pretty readable. I'll update literals that span multiple functions.

python/private/py_runtime_rule.bzl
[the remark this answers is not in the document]
rickeylev: And just to clarify -- this isn't because I'm against py_runtime being buildable. It's just the size/scope of this PR is already big and I didn't want to expand it even more.

rickeylev (commented):
FYI: I started writing docs with some examples for a followup PR, and it looks pretty good. The main shortcoming I see now is that the implementation function for a rule is very opaque (ctx in, list of providers out). So, that'll be the next thing to figure out, but that's largely separate from "how to define a rule".

python/private/rule_builders.bzl
rickeylev: Done

python/private/attr_builders.bzl
rickeylev: Done.

python/private/attr_builders.bzl
rickeylev: Done

PR #2613, opened 2025-02-14
docs: tell how to create branches for releases

rickeylev:
It's easier to do patch releases when the branch is already created. Some of the bugs
fixes in recent releases we probably could have easily released as patch releases if
we already had the branch ready.


aignas (approved):
+1 on doing patch releases through branches.

PR #2615, opened 2025-02-15
docs: split out release steps into separate doc

rickeylev:
Move the steps for releasing into a separate doc.

The release steps are specific to releases, which only maintainers do. This frees up
space in the developing docs for more general tips, tricks, and guidance for others.

Along the way...

* Remove the text about the core rules being part of Bazel
* Put the CLA text first -- if CLAs aren't signed _before_ code is given, it can
  result is large headaches.
* Move some more internal dev steps out of contributing docs.


aignas (approved):
Thanks!

PR #2620, opened 2025-02-20
refactor: cleanup now-unreferenced proto toolchain type

alexeagle:
Follow-up to #2604, fixes a breaking change in v1.2.0-rc0

Note that this toolchain_type became unused in that PR. We leave behind an alias to make this a non-breaking change.

Verified in a downstream repo that requires the toolchain_type to register pre-built `protoc`: https://github.com/aspect-build/toolchains_protoc/pull/50/files

### DIFF
--- a/python/proto/BUILD.bazel
+++ b/python/proto/BUILD.bazel
@@ -14,10 +14,9 @@
 
 package(default_visibility = ["//visibility:public"])
 
-# Toolchain type provided by proto_lang_toolchain rule and used by py_proto_library
-# Deprecated in favour of the implementation in https://github.com/protocolbuffers/protobuf.
-# It will be removed in a future release.
+# Deprecated; use @com_google_protobuf//bazel/private:python_toolchain_type instead.
 # Alias is here to provide backward-compatibility; see #2604
+# It will be removed in a future release.
 alias(
     name = "toolchain_type",
     actual = "@com_google_protobuf//bazel/private:python_toolchain_type",
--- a/python/proto/BUILD.bazel
+++ b/python/proto/BUILD.bazel
@@ -15,4 +15,8 @@
 package(default_visibility = ["//visibility:public"])
 
 # Toolchain type provided by proto_lang_toolchain rule and used by py_proto_library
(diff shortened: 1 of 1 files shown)


python/proto/BUILD.bazel
```diff
@@ -15,4 +15,8 @@
 package(default_visibility = ["//visibility:public"])
 
 # Toolchain type provided by proto_lang_toolchain rule and used by py_proto_library
-toolchain_type(name = "toolchain_type")
+# Alias is here to provide backward-compatibility; see #2604
+alias(
+    name = "toolchain_type",
+    actual = "@com_google_protobuf//bazel/private:python_toolchain_type",
```
aignas, line 21: Could you please add a deprecation message to the alias? Just copying the one fro proto_library would be enough.

aignas (commented):
Thanks for the fix!

python/proto/BUILD.bazel
alexeagle: done

aignas:
I meant something like https://github.com/bazelbuild/rules_python/blob/f2941df7562c4183c37d6ceeae23e7d390738d58/docs/BUILD.bazel#L182

alexeagle:
Sure - @aignas to move things along faster please feel free to just click the "pencil" icon and make trivial edits like that as your own commits on my PR 😉 
I do that for our contributors pretty frequently to spare the round-trip (unless the comment is something they might learn from)

rickeylev:
I added the deprecation attr setting. It's annoying that the toolchain type is under a "private" directory. They should give a public access point for it.

Queued for merge. I'll cherry pick this into the 1.2 release.

aignas:
> Sure - @aignas to move things along faster please feel free to just click the "pencil" icon and make trivial edits like that as your own commits on my PR 😉 I do that for our contributors pretty frequently to spare the round-trip (unless the comment is something they might learn from)

I do it when I can and when I have my laptop with me. I am not doing it now not because I don't know how to or am unwilling, so no need to point it out. :)

aignas (approved):

PR #2621, opened 2025-02-21
fix: Fall back to directory based runfiles using relative paths

mering:
This makes the runfiles libraries work when they're repackaged and extracted elsewhere.

e.g. a py_binary is packaged using `pkg_tar`, then extracted somewhere else.
The `bazel-bin/{BIN,BIN.runfiles}` structure (sans the bazel-bin prefix) is preserved, and then e.g. `./BIN` is run. This is equivalent to running `bazel-bin/BIN` (i.e no
presumption about environment variables can be made), just the location differs.

This makes it work out of the box when the runfiles tree is packaged into a container image as follows:

```starlark
py_binary(
    name = "binary",
    srcs = ["main.py"],
    data = ["runfile.txt"],
)

pkg_tar(
    name = "layer",
    srcs = [":binary"],
    strip_prefix = "/",
    include_runfiles = True,
)

oci_image(
    name = "image",
    tars = [":layer"],
)
```

### DIFF
--- a/python/runfiles/runfiles.py
+++ b/python/runfiles/runfiles.py
@@ -323,6 +323,9 @@ class Runfiles:
         Otherwise, if `env` contains "RUNFILES_DIR" with non-empty value (checked in
         this priority order), this method returns a directory-based implementation.
 
+        Otherwise, if a valid runfiles folder is found relative to this file, this
+        method returns a directory-based implementation.
+
         If neither cases apply, this method returns null.
 
         Args:
@@ -338,6 +341,10 @@ class Runfiles:
 
         directory = env_map.get("RUNFILES_DIR")
         if directory:
+            return CreateDirectoryBased(directory)
+        
+        directory =  _FindPythonRunfilesRoot()
+        if os.path.exists(os.path.join(directory, "_repo_mapping")):
             return CreateDirectoryBased(directory)
 
         return None
--- a/tests/runfiles/run_binary_with_runfiles_test.sh
+++ b/tests/runfiles/run_binary_with_runfiles_test.sh
@@ -1,0 +1,39 @@
+# Copyright 2024 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# --- begin runfiles.bash initialization v3 ---
+# Copy-pasted from the Bazel Bash runfiles library v3.
+set -uo pipefail; set +e; f=bazel_tools/tools/bash/runfiles/runfiles.bash
+source "${RUNFILES_DIR:-/dev/null}/$f" 2>/dev/null || \
+  source "$(grep -sm1 "^$f " "${RUNFILES_MANIFEST_FILE:-/dev/null}" | cut -f2- -d' ')" 2>/dev/null || \
+  source "$0.runfiles/$f" 2>/dev/null || \
+  source "$(grep -sm1 "^$f " "$0.runfiles_manifest" | cut -f2- -d' ')" 2>/dev/null || \
+  source "$(grep -sm1 "^$f " "$0.exe.runfiles_manifest" | cut -f2- -d' ')" 2>/dev/null || \
+  { echo>&2 "ERROR: cannot find $f"; exit 1; }; f=; set -e
+# --- end runfiles.bash initialization v3 ---
+set +e
+
+bin=$(rlocation $BIN_RLOCATION)
+if [[ -z "$bin" ]]; then
+  echo "Unable to locate test binary: $BIN_RLOCATION"
+  exit 1
+fi
+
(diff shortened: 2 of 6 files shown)


rickeylev:
Thanks for the fix @mering . I think this looks fine, but some surrounding changes to make:

* Can you describe the problematic case and why this fixes it?
* Tests: This looks easy to test if we just have a shell script setup the environment (unset the RUNFILES env vars) and call the binary and check it worked. The tests/support/sh_py_run_test.bzl#sh_py_run_test should make this pretty easy. There are examples of using it in tests/bootstrap_impls
* CHANGELOG.md needs to be updated.

mering:
@rickeylev Thanks for the review. See my answers below:

* When used in a container image without involving Bazel at all, the environment variables are not set but runfiles are in the expected folder `BIN.runfiles` as it was packaged via `pkg_tar` with `include_runfiles = True`
* I tried doing this but for some reason the `$RUNFILES_DIR` is still set within Python even though I unset it in the shell script
* I added this to the CHANGELOG

rickeylev:
Overall LGTM. CI is failing -- that test just needs to be updated. It's assuming the only two ways to find the runfiles root is the environment variables, but we're adding a 3rd attempt to look based upon its own file path.

> * When used in a container image without involving Bazel at all, the environment variables are not set but runfiles are in the expected folder `BIN.runfiles` as it was packaged via `pkg_tar` with `include_runfiles = True`. I updated the description with an example.

Thanks for explaining. So basically, copying `{BIN,BIN.runfiles}` somewhere else, and then running e.g. `./BIN`.

Yeah, that should work -- the structure is preserved; I don't see why it wouldn't. I've updated the description to reflect this.

> * I tried doing this but for some reason the `$RUNFILES_DIR` is still set within Python even though I unset it in the shell script.

What you're probably seeing is the bootstrap will set the RUNFILES_DIR environment variable (or manifest env var, depending on what it figures out).

er but -- if the bootstrap is always setting those env vars, how is runfiles not seeing them?

rickeylev (approved):

tests/runfiles/run_binary_with_runfiles_test.sh
```diff
@@ -0,0 +1,39 @@
...
+# --- end runfiles.bash initialization v3 ---
+set +e
+
+bin=$(rlocation $BIN_RLOCATION)
+if [[ -z "$bin" ]]; then
+  echo "Unable to locate test binary: $BIN_RLOCATION"
+  exit 1
+fi
+
+# Test invocation without RUNFILES environment variables set
```
fmeum, line 33: Sorry if I'm misunderstanding the test setup, but this doesn't look like a supported execution mode for any runfiles library: The runfiles tree lives next to the `.sh` file, which invokes the Python binary launcher as a subprocess but doesn't set any environment variables.

@mering Could you share a reproducer for the original issue that motivated this PR? I am happy to debug it, but I strongly suspect that the root cause lies elsewhere and strictly following the lookup procedures of other runfiles libraries is the best way to avoid nasty surprises.

python/runfiles/runfiles.py
```diff
@@ -339,6 +342,10 @@ def Create(env: Optional[Dict[str, str]] = None) -> Optional["Runfiles"]:
         directory = env_map.get("RUNFILES_DIR")
         if directory:
             return CreateDirectoryBased(directory)
+        
+        directory =  _FindPythonRunfilesRoot()
```
fmeum, line 346: This may not find the correct runfiles tree: It's an internal helper meant to find the tree that contains the Python files, solely for the purpose of identifying the calling repo. This tree could be entirely different from the runfiles tree, e.g. on Windows when using a Python ZIP.

tests/runfiles/run_binary_with_runfiles_test.sh
rickeylev: The situation is essentially the same as calling `bazel-bin/foo` directly (no guarantees about what the environment has set/not set), which is supported. Having e.g. a sh_binary with a data dependency that calls it as a subprocess is the same, just with an extra layer.

The bootstrap logic _should_ be taking care of this, though. It jumps through lots of hoops to find the runfiles directory/manifest and sets a runfiles environment variable, which the runfiles library should see later when Create() is called.

fmeum: It's not the same situation since the py_binary lives in the runfiles tree of the sh_binary, it doesn't have its own sibling runfiles tree. Transitive runfiles trees are no longer created, which is why cooperation via environment variables is required.

rickeylev: The py_binary is a data dependency of the sh_binary, so the py runfiles are merged with the sh runfiles. The result is the py binary within the sh runfiles uses the sh runfiles tree -- this is correct, since that's where its runfiles are.

Cooperation between the two processes using env vars isn't necessary: the py bootstrap is already performing the logic to find its runfiles root. There's actually a test of this over in tests/bootstrap_impls already.

fmeum: I see, that's because of this loop: https://github.com/bazel-contrib/rules_python/blame/d713ba704e9a6442c409134f7a701c0b6e1a9fe0/python/private/stage1_bootstrap_template.sh#L77

This is non-standard logic that most runfiles libraries don't contain. It may work well, it may be non-hermetic in edge cases, I'm not entirely sure. I'll think about this some more. It does mean that a Python process indirectly invoked within the runfiles of another binary will work, but if it runs a tool that uses a runfiles library without this trick that one won't work without the env vars.

Since setting env vars ensures hermeticity across languages, I would personally always set them.

mering:
Several other fixes were merged in the meantime and I can no longer reproduce this issue.

PR #2623, opened 2025-02-22
docs: add some docs to help contributors get started

rickeylev:
A common pattern I've seen with PRs is they lack tests. I suspect part of the reason is
authors aren't sure how to write tests or where to start. So here's some basic docs
to help.


aignas (approved):
Thanks for describing the transition tests in more detail.

PR #2625, opened 2025-02-24
ci: use Python 3.9 for mypy workflow to fix ci

rickeylev:
The mypy check on CI has been failing. The problem was the combination of:
* We were using Python 3.8
* jpetrucciani/mypy-check@master updated to use mypy 1.15
* mypy 1.15 dropped support for Python 3.8

To fix, use Python 3.9.

### DIFF
--- a/.github/workflows/mypy.yaml
+++ b/.github/workflows/mypy.yaml
@@ -22,11 +22,10 @@ jobs:
       - uses: jpetrucciani/mypy-check@master
         with:
           requirements: 1.6.0
-          python_version: 3.8
+          python_version: 3.11
           path: 'python/runfiles'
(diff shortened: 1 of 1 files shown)


aignas (approved):

aignas:
One thing to note here is that we might break the type annotations for earlier python versions more easily if we are testing on 3.11 only. We may want to either use the previous version of GH action that is using an older mypy version instead of bumping to 3.11.

That said the new style type annotations have been here for a while, so maybe just using 3.11 is a good enough thing, but maybe the oldest supported mypy version would be better.

rickeylev:
Good point. Updated to 3.9.

PR #2626, opened 2025-02-24
refactor: stop using some deprecated Starlark APIs

EdSchouten:
I am currently working on an analysis tool that is capable of parsing BUILD/*.bzl files. It currently fails to process some of the Python rules, due to the rules depending on some features that are deprecated on the Bazel side. Instead of adding implementations of these deprecated features to my brand new analysis tool, I thought I'd simply patch up the Python rules instead.


rickeylev (approved):
Thanks for the PR!

TIL workspace_name is now deprecated. Do you know when `.repo_name` became available?

python/private/toolchains_repo.bzl
```diff
@@ -301,20 +297,20 @@ this repo causes an eager fetch of the toolchain for the host platform.
...
         file = "{}/defs.bzl".format(python_version)
         rctx.file(file, content = """\
 # Generated by python/private/toolchains_repo.bzl
 
-load("{rules_python}//python:pip.bzl", _compile_pip_requirements = "compile_pip_requirements")
-load("{rules_python}//python/private:deprecation.bzl", "with_deprecation")
-load("{rules_python}//python/private:text_util.bzl", "render")
-load("{rules_python}//python:py_binary.bzl", _py_binary = "py_binary")
-load("{rules_python}//python:py_test.bzl", _py_test = "py_test")
+load("@@{rules_python}//python:pip.bzl", _compile_pip_requirements = "compile_pip_requirements")
```
groodt, line 307: Does it always make sense to load from the global root module context?

aignas:
Bazel 7.0 which is the lowest version that we support does not have `repo_name`: https://bazel.build/versions/7.0.0/rules/lib/builtins/Label

EdSchouten:
Hey @aignas,

> Bazel 7.0 which is the lowest version that we support does not have `repo_name`: https://bazel.build/versions/7.0.0/rules/lib/builtins/Label, but 7.1 has: https://bazel.build/versions/7.1.0/rules/lib/builtins/Label

It's also backported to 6.5.0:

https://bazel.build/versions/6.5.0/rules/lib/Label

> I would propose to drop support for anything below 7.0 and merge this. We need a CHANGELOG item clearly stating it.

I noticed `CHANGELOG.md` already contains this sentence, so I assume no further documentation changes are needed, right?

> * Bazel 6 support is dropped and Bazel 7.4.1 is the minimum supported
  version, per our Bazel support matrix. Earlier versions are not
  tested by CI, so functionality cannot be guaranteed.

python/private/toolchains_repo.bzl
EdSchouten: You mean that there are still some other places in this file that use `@rules_python` instead of `@@{rules_python}`? Well, I thought that looked a bit funky, but wasn't sure whether that was accidental/intentional. Happy to change the code either way.

aignas:
Thanks for the digging and the clarification, it sounds good to me then!

aignas (approved):
LGTM as is.

aignas (approved):

Issue #2241, opened 2024-09-21
Third party dependencies are incorrect when using RBE because host != exec


groodt:
# Context

This is a tracking issue to recognise the lack of support for Remote Execution (RBE) when using the [support for third party dependencies from PyPI](https://rules-python.readthedocs.io/en/latest/pypi-dependencies.html) (or similar indexes/mirrors).

A reproduction of the issue has been developed in this [rules_python_rbe_exec_exploration repository](https://github.com/groodt/rules_python_rbe_exec_exploration).

The issue occurs when there are native dependencies in the transitive closure of third party dependencies, which is very common in the Python ecosystem.

The root cause of the issue is that the current implementation that imports third party dependencies as targets of the bazel build assumes that the host platform is the same as the execution platform (host == exec). However, a very common configuration of remote execution would be macos or windows host platforms executing remote builds on linux platforms (host != exec).

# Known workarounds

1. Only use pure python third party dependencies (this is not really practical)
2. Try the experimental support for [multi-platform wheel repositories using](https://rules-python.readthedocs.io/en/latest/pypi-dependencies.html#bazel-downloader-and-multi-platform-wheel-hub-repository) `experimental_index_url` 
3. Use alternative rules such as [rules_pycross](https://github.com/jvolkman/rules_pycross) that can import third party dependencies where host != exec

# Related
- TODO

aignas:
I have realized that this will be (partially?) fixed by #2949 because then the dependency tree will be correctly resolved irrespective of the host platform, but the issue about pulling the right platform specific wheels will be still there until #260 is resolved.

PR #2629, opened 2025-02-24
refactor(pypi): translate wheel METADATA parsing to starlark

### LINKED ISSUE #2423: [pypi] PEP508 Remove the need for Python in env_marker evaluation in hub repos

aignas:
This PR starts using the newly introduced (#2692) PEP508 compliant
requirement marker parser in starlark and moves the dependency
generation from the Python language (`whl_installer`) to the Starlark
in the `whl_library` repository rule.

This PR is (almost) a pure refactor where no bugs are fixed, but this is
foundational work that also adds notes on how things will be moved
to macros (i.e. analysis phase) so that we can fix a few long standing
bugs and prepare for stabilizing the `experimental_index_url` (#260).

Refactor:
* I have migrated all of the unit tests from Python to starlark for deps
  generation from METADATA `Requires-Dist` fields.
* Read the `METADATA` file itself in Starlark.

Work towards #260, #2319, #2241
Fixes #2423


aignas:
For having unit tests working we need to first resolve #2735.

aignas:
OK, for today that is it.

Current progress:
* Moved the `METADATA` file reading/parsing to starlark as well.

Remaining things that may go either to this or another PR (depending on when this gets reviewed):
* `entry_points.txt` reading/parsing in starlark.
* `namespace_pkg_dirs` generation in starlark to make it easier to move it to the build context.

Things that will be definitely out of scope for this PR is unzipping of the wheel in starlark with [repository_ctx](https://bazel.build/rules/lib/builtins/repository_ctx#extract). The reason for that is that the `unzipping` does a few things:
* Remap paths:
    * `purelib`: `/site-packages`
    * `platlib`: `/site-packages`
    * `headers`: `/include`
    * `scripts`: `/bin`
    * `data`: `/data`
* Replace the interpreter in `posix` scripts
* Set the installer metadata.

For tackling extraction in starlark we have the following strategy:
* Create a separate `repository_rule` called `whl_archive`, which is pure starlark and is reusing all of the code that will land in this PR. It would download the wheel and extract it.
* Use the new rule in our internal dependencies, once the `sphinx` generation is working, we can use it everywhere, most likely.

However I am not too fussed with leaving the installer for some time to use Python - it is Python version agnostic, so should work on any Python version that `rules_python` bundles and it just needs the interpreter and a few Python deps (which need to be setup without Python).

rickeylev (approved):

Issue #1169, opened 2023-04-13
non-root requirement seems too much for an ordinary user of `rules_python`


qzmfranklin:
# 🐞 bug report

### Affected Rule

 The issue is caused by the rule: 

During the loading phase.

### Is this a regression?

I don't think so.

### Description

Per #713 , `rules_python` cannot be used by `root`.

CI systems such as CircleCI and BuildBuddy uses `root` user by default.

Setting up non-root in those systems aren't very straightforward.

Also, it does not look like a responsibility of an ordinary user of rules_python to have to do this.

## 🔬 Minimal Reproduction

Example build event:

https://app.buildbuddy.io/invocation/55238174-54c3-459c-8d80-72722f2d00f9

## 🔥 Exception or Error

Relevant error message:

```text
Error in fail: The current user is root, please run as non-root when using the hermetic Python interpreter. See https://github.com/bazelbuild/rules_python/pull/713.
```

<pre><code>

</code></pre>

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
Linux, CentOS8, WSL2-Ubuntu22.04
  </code>
</pre>

**Output of `bazel version`:**

<pre>
  <code>
6.1.1
  </code>
</pre>

**Rules_python version:**

<pre>
  <code>
1e869d8d945f0b406da65817cf70b4fa3105a0de
  </code>
</pre>

**Anything else relevant?**

nsubiron:
You can disable the error
```py
python_register_toolchains(
    name = ...,
    python_version = ...,
    ignore_root_user_error = True,
)
```

faximan:
It is good there is an option to disable this but the problem is that you have to choose between two evils: either you get the pyc cache misses or you have to go out of your way to support non-root builds (e.g. when building in a docker container).

We have opted for the second evil, but it creates so much extra work.

I really wish there was a way to solve the cache issue without this extra requirement.

phlax:
we hit the same issues - my feeling is that this is just not the responsibility of `rule_python` to enforce this

qzmfranklin:
To be fair, it's probably the CPython to blame, as they programmed to produce the `.pyc` files on the fly.

Not sure if they are willing to make a flag to NOT generate `.pyc` files on the `stdlib` on the fly just for this use case, though.

daixiang0:
> You can disable the error
> 
> ```python
> python_register_toolchains(
>     name = ...,
>     python_version = ...,
[...]

Issue #1008, opened 2023-01-23
Optimize pyc files


f0rmiga:
As discussed in https://github.com/bazelbuild/rules_python/pull/907, ideally we would include the pre-compiled pyc files from the Python interpreters shipped with rules_python.

PR #2636, opened 2025-02-28
fix: Downgrade "running as root" error to a warning by default

Wyverald:
Currently, by default, rules_python immediately fails when Bazel is run as root. The reasoning behind this involves .pyc files being generated for hermetic toolchains when they're first used, causing cache misses; to work around this, rules_python opts to make the toolchain installation directory read-only, but running Bazel as root would circumvent this. So rules_python actively detects if the current user is root, and hard fails. This check can be disabled by the root module by setting `python.override(ignore_root_user_error=True)`. (See more context in the linked issues/PRs.)

This causes a reverberating effect across the Bazel ecosystem, as rules_python is essentially a dependency of every single Bazel project through protobuf. Effectively, any Bazel project wishing to run as root need to add the override tag above, even if they don't have anything to do with Python at all.

This PR changes the default value of the `ignore_root_user_error` to True instead. Besides, it now unconditionally tries to make the toolchain installation directory read-only, and only outputs a warning if it's detected that the current user is root.

See previous discussions at #713, #749, #907, #1008, #1169, etc.

Fixes https://github.com/bazelbuild/rules_python/issues/1169.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,7 +59,10 @@ Unreleased changes template.
 * (pypi) The `ppc64le` is now pointing to the right target in the `platforms` package.
 * (gazelle) No longer incorrectly merge `py_binary` targets during partial updates in 
   `file` generation mode. Fixed in [#2619](https://github.com/bazelbuild/rules_python/pull/2619).
-* The hard error thrown when Bazel is run as root is now downgraded to a warning by default.
+* (bzlmod) Running as root is no longer an error. `ignore_root_user_error=True`
+  is now the default. Note that running as root may still cause spurious
+  Bazel cache invalidation
+  ([#1169](https://github.com/bazelbuild/rules_python/issues/1169)).
 
 {#v0-0-0-added}
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,7 @@ Unreleased changes template.
 * (pypi) The `ppc64le` is now pointing to the right target in the `platforms` package.
 * (gazelle) No longer incorrectly merge `py_binary` targets during partial updates in 
   `file` generation mode. Fixed in [#2619](https://github.com/bazelbuild/rules_python/pull/2619).
+* The hard error thrown when Bazel is run as root is now downgraded to a warning by default.
 
 {#v0-0-0-added}
 ### Added
--- a/python/private/python_repository.bzl
+++ b/python/private/python_repository.bzl
@@ -135,7 +135,7 @@ def _python_repository_impl(rctx):
             logger = logger,
         )
 
-        fail_or_warn = print if rctx.attr.ignore_root_user_error else fail
+        fail_or_warn = logger.warn if rctx.attr.ignore_root_user_error else logger.fail
         exec_result = repo_utils.execute_unchecked(
             rctx,
             op = "python_repository.TestReadOnly",
--- a/python/private/python_repository.bzl
+++ b/python/private/python_repository.bzl
@@ -127,37 +127,36 @@ def _python_repository_impl(rctx):
     # pycs being generated at runtime:
     # * The pycs are not deterministic (they contain timestamps)
     # * Multiple processes trying to write the same pycs can result in errors.
-    if not rctx.attr.ignore_root_user_error:
-        if "windows" not in platform:
-            lib_dir = "lib" if "windows" not in platform else "Lib"
+    if "windows" not in platform:
+        repo_utils.execute_checked(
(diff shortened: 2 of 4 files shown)


python/private/python_repository.bzl
```diff
@@ -127,37 +127,36 @@ def _python_repository_impl(rctx):
...
+        )
 
-            repo_utils.execute_checked(
-                rctx,
-                op = "python_repository.MakeReadOnly",
-                arguments = [repo_utils.which_checked(rctx, "chmod"), "-R", "ugo-w", lib_dir],
-                logger = logger,
-            )
-            exec_result = repo_utils.execute_unchecked(
+        fail_or_warn = print if rctx.attr.ignore_root_user_error else fail
```
aignas, line 138: Using logger.warn and logger.fail here would be slightly better.

aignas (commented):
Could you also change the python repo utils to add -B argument when executing python scripts please?

Thanks for defaulting it to a warning. That is a really good first step out of the current situation.

Wyverald:
> Could you also change the python repo utils to add -B argument when executing python scripts please?

Do you have a pointer where that should be done?

I also feel like it's a bit outside the scope of this PR. In fact, wouldn't adding -B completely obviate this PR, since no .pyc files would ever be generated for the hermetic toolchains? 

---

I'm also having trouble fixing the tests. Specifically, this block of code dealing with overrides: https://github.com/bazelbuild/rules_python/blob/5a5560264d26d1c4bf35428da71b82f7aa502f1e/python/private/python.bzl#L430-L432
Makes it so that only truthy override values are respected. Since I changed the default of `ignore_root_user_error` to `True` in this PR, overriding it to `False` is actually a no-op due to the code above. I'm not fully sure whether I should just change line 431 to be `if hasattr(tag, key):`.

CHANGELOG.md
```diff
@@ -59,6 +59,7 @@ Unreleased changes template.
 * (pypi) The `ppc64le` is now pointing to the right target in the `platforms` package.
 * (gazelle) No longer incorrectly merge `py_binary` targets during partial updates in 
   `file` generation mode. Fixed in [#2619](https://github.com/bazelbuild/rules_python/pull/2619).
+* The hard error thrown when Bazel is run as root is now downgraded to a warning by default.
```
rickeylev, line 62: Fixes https://github.com/bazelbuild/rules_python/issues/1169.
```suggestion
* (bzlmod) Running as root is no longer an error. `ignore_root_user_error=True`
  is now the default. Note that running as root may still cause spurious
  Bazel cache invalidation
  ([#1169](https://github.com/bazelbuild/rules_python/issues/1169)).
```

python/private/python_repository.bzl
rickeylev, line 138:
```suggestion
        fail_or_warn = logger.warn if rctx.attr.ignore_root_user_error else logger.fail
```

rickeylev (commented):
re: setting -B: Yeah, I agree; outside the scope of the PR. 

I think Ignas is referring to python/private/pypi/pypi_repo_utils.bzl, where there's a helper to run a command during repo phase.

> -B would obviate this PR

It may or may not -- honestly we're not sure. Adding it could help the situation. We could never fully figure out what was going on and it was flaky behavior that was hard to replicate. At times, it almost seemed like repo-phase invocations, build actions, and runtime executions were all racing against each other trying to create pyc files in the directory the python runtime had been downloaded into.

rickeylev:
> trouble fixing tests; False gets ignored because True is given precedence

I'm ok with that. i.e. I'm OK with ignore_root_user_error=True the only behavior; never fail, just warn. I just don't see why anyone would set ignore_root_user_error=False -- that means you're trying to guard against, what? _Accidentally_ running bazel as root?

Wyverald:
> > trouble fixing tests; False gets ignored because True is given precedence
> 
> I'm ok with that. i.e. I'm OK with ignore_root_user_error=True the only behavior; never fail, just warn. I just don't see why anyone would set ignore_root_user_error=False -- that means you're trying to guard against, what? _Accidentally_ running bazel as root?

Well, yeah... I would agree with you in that `ignore_root_user_error=False` IMO never had much of a purpose. But if you can't ever override `ignore_root_user_error=False`, there's basically no point in having the knob there. I could change the attr doc to say "this is deprecated, do not use" instead. WDYT?

rickeylev:
> change the attr doc to say "this is deprecated, do not use" instead. WDYT?

Yes. SGTM.

rickeylev (approved):

PR #2641, opened 2025-03-01
fix(pypi): use python -B for repo-phase invocations

aignas:
Before this change we would just invoke the Python interpreter. This
means that in the `rules_python` directory there would be `__pycache__`
folders created in the source tree and the same `__pycache__` folders
would be created in the python interpreter repository rules if the
directories were writable.

This change ensures that we are executing `python` with `-B` in those
contexts and reduces any likelihood of us doing the wrong thing.

Work towards #1169.

### DIFF
--- a/python/private/pypi/patch_whl.bzl
+++ b/python/private/pypi/patch_whl.bzl
@@ -27,8 +27,8 @@ other patches ensures that the users have overview on exactly what has changed
 within the wheel.
 """
 
-load("//python/private:repo_utils.bzl", "repo_utils")
 load(":parse_whl_name.bzl", "parse_whl_name")
+load(":pypi_repo_utils.bzl", "pypi_repo_utils")
 
 _rules_python_root = Label("//:BUILD.bazel")
 
@@ -102,10 +102,14 @@ def patch_whl(rctx, *, python_interpreter, whl_path, patches, **kwargs):
     record_patch = rctx.path("RECORD.patch")
     whl_patched = patched_whl_name(whl_input.basename)
 
-    repo_utils.execute_checked(
+    pypi_repo_utils.execute_checked(
         rctx,
+        python = python_interpreter,
+        srcs = [
+            "//python/private/pypi:repack_whl.py",
+            "//tools:wheelmaker.py",
+        ],
         arguments = [
-            python_interpreter,
             "-m",
             "python.private.pypi.repack_whl",
             "--record-patch",
--- a/python/private/pypi/patch_whl.bzl
+++ b/python/private/pypi/patch_whl.bzl
@@ -106,8 +106,8 @@ def patch_whl(rctx, *, python_interpreter, whl_path, patches, **kwargs):
         rctx,
         python = python_interpreter,
         srcs = [
-            "//python/private/pypi:repack_whl.py",
-            "//tools:wheelmaker.py",
+            Label("//python/private/pypi:repack_whl.py"),
+            Label("//tools:wheelmaker.py"),
         ],
         arguments = [
             "-m",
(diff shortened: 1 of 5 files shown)


python/private/pypi/patch_whl.bzl
```diff
@@ -102,10 +102,14 @@ def patch_whl(rctx, *, python_interpreter, whl_path, patches, **kwargs):
     record_patch = rctx.path("RECORD.patch")
     whl_patched = patched_whl_name(whl_input.basename)
 
-    repo_utils.execute_checked(
+    pypi_repo_utils.execute_checked(
         rctx,
+        python = python_interpreter,
+        srcs = [
+            "//python/private/pypi:repack_whl.py",
```
aignas, line 109:
```suggestion
            Label("//python/private/pypi:repack_whl.py"),
```

python/private/pypi/patch_whl.bzl
```diff
@@ -102,10 +102,14 @@ def patch_whl(rctx, *, python_interpreter, whl_path, patches, **kwargs):
...
+            "//tools:wheelmaker.py",
```
aignas, line 110:
```suggestion
            Label("//tools:wheelmaker.py"),
```

Wyverald:
I wonder how this actually helps. Per my very tenuous understanding of the issue (from working on it for a day), running Python without -B can cause .pyc files to be generated next to any used .py files. This PR would make sure rules_python's repo rules wouldn't generate .pyc files for the hermetic toolchains it downloads, but how would it prevent normal usage of the hermetic toolchains to write .pyc files into the same directories? (say, would `import os` just cause an `os.pyc` file to be written next to the `os.py` file? Is there even an `os.py` file?)

aignas:
The main place where this is helping is the rules_python internal sources.

The next location where it may help is if stdlib has python files (like for the pathlib module, os may have a python file, but I can't remember without looking) in the toolchain dirs. And this is only for the repo phase execution, which would not create any sandbox to run the tools.

For sandboxed execution of the toolchain (e.g. in build actions, etc) this should leave things the same and Iam not sure if there ever was a problem there.

python/private/pypi/pypi_repo_utils.bzl
```diff
@@ -104,11 +104,27 @@ def _construct_pypath(mrctx, *, entries):
...
+        # refetch the repository_rule when the srcs change. This should work on
+        # Bazel versions without `mrctx.watch` as well.
+        repo_utils.watch(mrctx, mrctx.path(src))
+
+    environment = kwargs.pop("environment", {})
+    pythonpath = environment.get("PYTHONPATH", "")
+    if pythonpath and not types.is_string(pythonpath):
+        environment["PYTHONPATH"] = _construct_pypath(mrctx, entries = pythonpath)
+    kwargs["environment"] = environment
+    kwargs["arguments"] = [python, "-B"] + kwargs.get("arguments", [])
```
rickeylev, line 119:
```suggestion
    # -B is added to prevent the repo-phase invocation from creating timestamp
    # based pyc files, which contributes to race conditions and non-determinism
    kwargs["arguments"] = [python, "-B"] + kwargs.get("arguments", [])
```

rickeylev (approved):

rickeylev:
Yeah, this won't affect regular execution phase stuff, just repo-phase. A theory is that the repo-phase and execution-phase are interweaving and stepping on each other. Something like: while a binary is being built, and is in the midst of evaluating the files that comprise the python runtime (i.e. evaluating glob()), a repo rule is also running. That repo rule runs the python interpreter directly, which then creates and deletes several pyc related files. The glob could run before, during, or after that sequence of file creates. And thus, we basically have a background process changes the sources a glob sees (both content and files). Multiple repo rules might be running, too, which then means they race each other and might generate different timestamps in the pyc files.

That's the gist of the theory, but I'm not 100% confident in this explanation :). It's just the best explanation we have given what we see. What we've empirically observed is that having the glob() ignore all pyc-looking things has prevented issues.

During the execution phase, its less clear if theres a similar issue. From what I recall, python has some logic to call realpath() on itself to find its "home" (where the stdlib etc are). So another theory is python is "reading through" any symlinks bazel creates and ends up in the downloaded repo directory, and then goes generating pyc files there again. I would expect sandboxing to prevent this, but I don't know how "strong" the sandbox is, or how various options might "weaken" it and incidentally enable this.

PR #2644, opened 2025-03-02
fix: spill module mapping args to a file

mattem:
Calls to the modules mapping rule contains very long command line args due to the use of the full `wheels` parameter. This change adds support for spilling the args into a file as needed.

In addition, it improves the performance of the `modules_mapping` rule:

* Remove the calls `to_list` that are unnecessary on the depset.
* Remove the iteration over the depset when passing to `args`, and other calls to `.path`, and instead let args do this lazily.

### DIFF
--- a/gazelle/modules_mapping/def.bzl
+++ b/gazelle/modules_mapping/def.bzl
@@ -25,18 +25,25 @@ module name doesn't match the wheel distribution name.
 
 def _modules_mapping_impl(ctx):
     modules_mapping = ctx.actions.declare_file(ctx.attr.modules_mapping_name)
-    args = ctx.actions.args()
     all_wheels = depset(
         [whl for whl in ctx.files.wheels],
         transitive = [dep[DefaultInfo].files for dep in ctx.attr.wheels] + [dep[DefaultInfo].data_runfiles.files for dep in ctx.attr.wheels],
     )
-    args.add("--output_file", modules_mapping.path)
+
+    args = ctx.actions.args()
+
+    # Spill parameters to a file prefixed with '@'. Note, the '@' prefix is the same
+    # prefix as used in the `generator.py` in `fromfile_prefix_chars` attribute.
+    args.use_param_file(param_file_arg = "@%s")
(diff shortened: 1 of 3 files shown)


gazelle/modules_mapping/generator.py
```diff
@@ -152,6 +152,7 @@ def data_has_purelib_or_platlib(path):
     parser = argparse.ArgumentParser(
         prog="generator",
         description="Generates the modules mapping used by the Gazelle manifest.",
+        fromfile_prefix_chars="@",
```
aignas, line 155:
```suggestion
        # Automatically read parameters from a file. Note, the '@' is the same prefix
        # as set in the 'args.use_param_file' in the bazel rule.
        fromfile_prefix_chars="@",
```

gazelle/modules_mapping/def.bzl
```diff
@@ -25,18 +25,22 @@ module name doesn't match the wheel distribution name.
...
     modules_mapping = ctx.actions.declare_file(ctx.attr.modules_mapping_name)
-    args = ctx.actions.args()
     all_wheels = depset(
         [whl for whl in ctx.files.wheels],
         transitive = [dep[DefaultInfo].files for dep in ctx.attr.wheels] + [dep[DefaultInfo].data_runfiles.files for dep in ctx.attr.wheels],
     )
-    args.add("--output_file", modules_mapping.path)
+
+    args = ctx.actions.args()
+    args.use_param_file(param_file_arg = "@%s")
```
aignas, line 34:
```suggestion
    # spill parameters to a file prefixed with '@'. Note, the '@' prefix is the same
    # prefix as used in the `generator.py` in `fromfile_prefix_chars` attribute.
    args.use_param_file(param_file_arg = "@%s")
```

aignas (commented):
In general LGTM, but would love to have some explanation of how things are wired. It took me a few minutes to look at it and google official docs to understand how it is actually wired.

aignas (approved):
Thank you.

PR #2647, opened 2025-03-03
fix(uv): Updates latest releases of uv in versions.bzl

dcode:
- Also provides a Python script in the tools directory that I used to pull the latest releases and update the file
- Also provides unittests for the Python script.

I made this change so that more recent versions of `uv` could be leveraged with `rules_python` and hopefully future releases can be incorporated more easily.


aignas:
I already have #2578 that is at odds with this PR. :)

dcode:
Great. Closing this since it's obsolete.

Issue #2645, opened 2025-03-03
`sphinx_docs` creates an implicit `build_test` target


255:
`sphinx_docs` creates an implicit `build_test` target. (Note that this is not currently documented.)

https://github.com/bazelbuild/rules_python/blob/f2941df7562c4183c37d6ceeae23e7d390738d58/sphinxdocs/private/sphinx.bzl#L196-L197

This means that commands like `bazel test //...` always include `sphinx_docs` targets. For developers (like me) that repeatedly run tests, this is pretty costly, especially given the lack of incremental builds.

Could the implicit `_build_test` target be removed from `sphinx_docs`?

I would prefer if `sphinx_docs` left the policy of whether to `build_test` each target to users. Since projects don't typically have many `sphinx_docs` targets, manually adding `build_test`s as needed should not be burdensome.

255:
@rickeylev

rickeylev:
I agree. Send a PR to remove it? Please just add a replacement build_test for these:
* //docs:docs_build_test
* //sphinxdocs/tests/sphinx_docs:docs_build_test
* //sphinxdocs/tests/sphinx_stardoc:docs_build_test

Simply putting a `build_test(...)` in the same directory in the BUILD file SGTM.

> bazel test //...

Running //... is my general flow, too, so a tip I use: `--build_tests_only`. Helps avoiding building a bunch of stuff that isn't actually run.

Actually, while were in sphinx.bzl, marking the `.run` and `.serve` targets with `tags=["manual"]` is probably a good idea, too. Those are just for interactive development, so no reason they should be build by wildcard stuff. Less to build -> more resources for building/running the tests you're interested in.

> given the lack of incremental builds.

By incremental, you mean Sphinx's support for "incremental" building? I seem to recall it has a "check if files have been modified" feature that I disabled in the rules. You could try enabling that 

You're from pigweed, right? What sort of build times do you see? Something I'd like is to have sphinx implemented as a worker process, under the assumption it'd be able to leverage sphinx's "only rebuild changed files" feature somehow.

rickeylev:
Also, as an interim workaround, setting `--test_tag_filters=-@rules_python//sphinxdocs:sphinx_docs` should cause the build test to be ignored when Bazel expands out test to run.

255:
I will put together a PR with the changes you suggest, thanks!

[...]

PR #2650, opened 2025-03-06
chore: Remove *_build_test targets from sphinx_docs (#2645)

255:
Remove implicit `build_test`s from `sphinx_docs` targets. Instead, users can decide whether or not to add `build_tests` for docs. This also keeps `sphinx_docs` builds out of `bazel test //...`, which may not be desirable.

Add `build_test`s to cover in-tree `sphinx_docs` targets. Rename the existing `build_test` for `//sphinxdocs/tests/sphinx_docs:docs` to match the new targets.

Also, tag the `sphinx_docs` `*.run` and `*.serve` targets as `"manual"` so they are excluded from wildcards. These are only needed for interactive development.

### DIFF
--- a/sphinxdocs/private/sphinx.bzl
+++ b/sphinxdocs/private/sphinx.bzl
@@ -15,7 +15,6 @@
 """Implementation of sphinx rules."""
 
 load("@bazel_skylib//lib:paths.bzl", "paths")
-load("@bazel_skylib//rules:build_test.bzl", "build_test")
 load("@bazel_skylib//rules:common_settings.bzl", "BuildSettingInfo")
 load("//python:py_binary.bzl", "py_binary")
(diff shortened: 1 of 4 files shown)


sphinxdocs/private/sphinx.bzl
```diff
@@ -177,6 +176,9 @@ def sphinx_docs(
         **common_kwargs
     )
 
+    common_kwargs_with_manual_tag = dict(common_kwargs)
+    common_kwargs_with_manual_tag.setdefault("tags", []).append("manual")
```
rickeylev, line 180: The dict() operation is shallow, so the tags object in `_with_manual` will refer to the same object as `common_kwargs`, and this will end up appending to both.

```suggestion
    common_kwargs_with_manual_tag["tags"] = list(common_kwargs.get("tags") or []) + ["manual"]
```

rickeylev (approved):

PR #2655, opened 2025-03-09
WIP sketch of adding bazel downloader support for direct urls

chrisirhc:
PR Instructions/requirements
  section at the bottom.
* Body text describes:

### DIFF
--- a/python/private/pypi/index_sources.bzl
+++ b/python/private/pypi/index_sources.bzl
@@ -59,17 +59,25 @@ def index_sources(line):
     ).strip()
 
     # Extract URL if present
-    url = ""
-    filename = ""
     if "@" in head:
         requirement = requirement_line
         # Extract URL from direct URL format
-        url = requirement.split("@")[1].split("#")[0].strip()
+        url = requirement.partition("@")[2].strip().partition(" ")[0].strip()
         # Extract filename from URL
         if url:
             filename = url.rpartition("/")[2]
             if not filename:
                 filename = url.rpartition("/")[0].rpartition("/")[2]
+
+            return struct(
+                requirement = requirement,
+                requirement_line = requirement_line,
+                version = version,
+                shas = sorted(shas),
+                marker = marker,
+                url = url,
+                filename = filename,
+            )
 
     return struct(
         requirement = requirement,
@@ -77,6 +85,4 @@ def index_sources(line):
         version = version,
         shas = sorted(shas),
         marker = marker,
-        url = url,
-        filename = filename,
     )
--- a/python/private/pypi/index_sources.bzl
+++ b/python/private/pypi/index_sources.bzl
@@ -32,6 +32,8 @@ def index_sources(line):
             * `marker` - str; the marker expression, as per PEP508 spec.
             * `requirement` - str; a requirement line without the marker. This can
                 be given to `pip` to install a package.
+            * `url` - str; URL if the requirement specifies a direct URL.
+            * `filename` - str; filename if URL is present, extracted from the URL.
     """
     line = line.replace("\\", " ")
     head, _, maybe_hashes = line.partition(";")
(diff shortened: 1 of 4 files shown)


aignas:
Thank you for the contribution.

> My guess is that direct urls can return sdists via checking the filename for the extensions listed in [here](https://docs.python.org/3.10/distutils/sourcedist.html). If I can get confirmation on this, I can update this PR to add sdist support for direct urls or leave that as a separate PR.

I think this part might be good to leave for another PR.

python/private/pypi/index_sources.bzl
```diff
@@ -55,9 +56,23 @@ def index_sources(line):
         requirement,
         " ".join(["--hash=sha256:{}".format(sha) for sha in shas]),
     ).strip()
+
     if "@" in head:
         requirement = requirement_line
-        shas = []
+
+        _, _, url_and_rest = requirement.partition("@")
```
aignas, line 63:
```suggestion
```

I think this line is duplicate

python/private/pypi/parse_requirements.bzl
```diff
@@ -292,6 +292,21 @@ def _add_dists(*, requirement, index_urls, logger = None):
         index_urls: The result of simpleapi_download.
         logger: A logger for printing diagnostic info.
     """
+
+    # Handle direct URLs in requirements
+    if hasattr(requirement.srcs, "url"):
```
aignas, line 297: it would be better to just return the attribute always but leave it as empty string if it is not set.

Then the logic would read as:
```suggestion
    if requirement.srcs.url:
```

python/private/pypi/parse_requirements.bzl
```diff
@@ -292,6 +292,21 @@ def _add_dists(*, requirement, index_urls, logger = None):
...
+        _, _, filename = url.rpartition("/")
+        direct_url_dist = struct(
+            url = url,
+            filename = filename,
+            sha256 = requirement.srcs.shas[0] if requirement.srcs.shas else "",
+            yanked = False,
+        )
+
+        # TODO should be able to handle sdist by checking the filename extension
+        return [direct_url_dist], None
```
aignas, line 308: Anything that does not have `.whl` extension could be thought as an `sdist`, IMHO.

tests/pypi/index_sources/index_sources_tests.bzl
```diff
@@ -28,23 +28,35 @@ def _test_no_simple_api_sources(env):
...
         ),
-        "foo==0.0.1 @ https://someurl.org; python_version < \"2.7\"\\    --hash=sha256:deadbeef": struct(
-            requirement = "foo==0.0.1 @ https://someurl.org --hash=sha256:deadbeef",
+        "foo==0.0.1 @ https://someurl.org/package.whl --hash=sha256:deadbeef": struct(
+            requirement = "foo==0.0.1 @ https://someurl.org/package.whl --hash=sha256:deadbeef",
+            marker = "",
+            url = "https://someurl.org/package.whl",
+            shas = ["deadbeef"],
+        ),
+        "foo==0.0.1 @ https://someurl.org/package.whl; python_version < \"2.7\"\\    --hash=sha256:deadbeef": struct(
```
aignas, line 44: Thanks for adding tests, please add one for an sdist (just a single example is good enough).

aignas (changes requested):
LGTM in general, I think we can correctly infer the `sdist` vs not via the direct URLs, so I think its fine to just do it.

Could you please also add the CHANGELOG note about the added support.

tests/pypi/index_sources/index_sources_tests.bzl
chrisirhc: Added it in parse_requirements_tests since that's where the new behavior is introduced.

PR #2656, opened 2025-03-10
fix(gazelle): Include YAML 'docstart' in gazelle manifest file

dougthor42:
Update `gazelle_python.yaml` to include the YAML docstart string:

```diff
-- a/gazelle_python.yaml
+++ b/gazelle_python.yaml
@@ -3,6 +3,7 @@
 # To update this file, run:
 #   bazel run //:gazelle_python_manifest.update
 
+---
 manifest:
   modules_mapping:
     30fcd23745efe32ce681__mypyc: black
```

While _technically_ not required, it is good practice to include. And then users don't have to exclude `gazelle_python.yaml` from their linters :upside_down_face:.

/cc @joshgc

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,11 +55,13 @@ Unreleased changes template.
 * (deps) platforms 0.0.4 -> 0.0.11
 * (py_wheel) Package `py_library.pyi_srcs` (`.pyi` files) in the wheel.
 * (py_package) Package `py_library.pyi_srcs` (`.pyi` files) in `py_package`.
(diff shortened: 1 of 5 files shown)


aignas (approved):
LGTM, I think there may be some tests fail.

dougthor42:
> I think there may be some tests fail.

Yeah, caught the one I missed and then had to re-run for a flake (seems like [this one](https://buildkite.com/bazel/rules-python-python/builds/10535#01958243-a450-4bb0-b8f4-9253b6cbc91a) just crashed out instead of being an actual failure)

Issue #2640, opened 2025-02-28
`whl_library.BuildWheelFromSource` uses pip to download dependencies and does not inject credential helper information.


dougthor42:
# 🐞 bug report

### Affected Rule

`pip.parse` and the underlying code


### Is this a regression?

Not that I can tell.


### Description

When using python package index that requires authentication, `whl_library.BuildWheelFromSource` will pass that index arg to the `pip wheel` command.

This is a problem when using the Bazel downloader (`experimental_index_url` et. al.). Bazel can authenticate to the package index via a credential helper header injection and can successfully download the source tarball. However, rules_python then tries to create a wheel from that tarball and uses `pip wheel --index-url "${INDEX_NEEDING_AUTH}" ...`. Thus building the wheel fails because pip can't auth to the private index.


## 🔬 Minimal Reproduction

Hard to repro if you don't have a private index readily available, but the gist is:

1. Use only package indexes that require authentication
2. Use experimental Bazel downloader `experimental_index_url`
3. Attempt to install a package that fits both of these requirements. `pygraphviz` is a good example of such a package.
    1. Does not have a wheel available on the package index
    2. Requires an additional package (such as `setuptools`) in order to build a wheel.


## 🔥 Exception or Error

```
root@b4337bd118dd:/bazel_starter# time bazel build --nobuild --config=local --remote_cache= --bes_backend= //...
INFO: Repository rules_python~~pip~pypi_312_pygraphviz_sdist_8b0b9207 instantiated at:
  <builtin>: in <toplevel>
Repository rule whl_library defined at:
  /root/.cache/bazel/_bazel_root/dd66987607986e0bbc3aa1b8af741d50/external/rules_python~/python/private/pypi/whl_library.bzl:469:30: in <toplevel>
INFO: repository @@rules_python~~pip~pypi_312_pygraphviz_sdist_8b0b9207' used the following cache hits instead of downloading the corresponding file.
 * Hash '8b0b9207954012f3b670e53b8f8f448a28d12bdbbcf69249313bd8dbe680152f' for https://REDACTED_1/pygraphviz/pygraphviz-1.12.tar.gz
If the definition of 'repository @@rules_python~~pip~pypi_312_pygraphviz_sdist_8b0b9207' was updated, verify that the hashes were also updated.
[...]

PR #2657, opened 2025-03-11
uv lock rule instead of genrule

aignas:
- **first stab**
- **turn off default upgrade**
- **poc is done**
- **wip: shave it**
- **Add a forgotten file**
- **wip: get a runnable rule as well**
- **manual testing setup for the runable rule**
- **Allow empty glob matches**
- **move back**
- **fixup**
- **wip**
- **fixup**
- **fixup**
- **rename**
- **fixup**
- **add notes**
- **wip**
- **add a hook for templating args**
- **move to templating part of the args**
- **wip**
- **wip**
- **add diff_test**
- **poc is done**
- **rename**
- **use param file**
- **wip**
- **wip**

PR Instructions/requirements
  section at the bottom.
* Body text describes:


python/uv/private/BUILD.bazel
```diff
@@ -88,3 +98,15 @@ bzl_library(
     srcs = ["versions.bzl"],
     visibility = ["//python/uv:__subpackages__"],
 )
+
+py_binary(
+    name = "pip_compile",
+    srcs = ["pip_compile.py"],
+    data = [
+        "//python/uv:current_toolchain",
```
aignas, line 106: Here we use a `py_binary` to depend on 2 toolchains at once. The `uv:current_toolchain` should be always transitioned to the host platform, just like the python toolchain. We could in theory just have a `pip_compile.py` script that we launch like:
```
$(PYTHON3) $(rootpath pip_compile.py) $(rootpath params.file)
```

Where we add the `uv:current_toolchain` into params.

python/uv/private/lock.bzl
```diff
@@ -15,109 +15,242 @@
...
-        ],
     )
 
+    _run_lock(
+        name = locker_target,
+        lock = name,
+    )
+
+    if maybe_out:
+        diff_test(
```
aignas, line 210: The diff test is only added if the output file already exists in the source tree.

Maybe we should have a parameter called `no_test = False` to the macro?

python/uv/private/lock.bzl
```diff
@@ -15,109 +15,242 @@
...
-            'src = Path(environ["REQUIREMENTS_FILE"])',
-            'assert src.exists(), f"the {src} file does not exist"',
-            'dst = Path(environ["BUILD_WORKSPACE_DIRECTORY"]) / "{}" / "{}"'.format(pkg, out),
-            'print(f"Writing requirements contents\\n  from {src.absolute()}\\n  to {dst.absolute()}", file=stderr)',
-            "dst.write_text(src.read_text())",
-            'print("Success!", file=stderr)',
-        ],
+        template = "//python/uv/private:copy.py",
+        substitutions = {
+            'dst = ""': 'dst = "{}/{}"'.format(pkg, out),
```
aignas, line 224: This is emulating `$(rootpath out)`.

I think we could extract a macro for writing files into the source tree.

python/uv/private/lock.bzl
```diff
@@ -15,109 +15,242 @@
...
+    params = ctx.actions.declare_file(ctx.label.name + ".params.txt")
+    ctx.actions.symlink(
+        output = params,
+        target_file = run_info.args,
+    )
+
+    executable = ctx.actions.declare_file(ctx.label.name + ".exe")
+    ctx.actions.write(
+        output = executable,
+        content = "#!/usr/bin/env bash\nset -eux\nexec {} {} \"$@\"".format(
```
aignas, line 123: This kind of works only on bash and windows requires extra setting. Maybe just using the 2 toolchains and running the python file separately is better?

I think the problem is that we need a cross-platform way to copy the file contents from some location before we run `uv` itself. Hence, either we need to write `powershell` to do that or `bash`. Write now we have python and we still need `powershell` or `bash`. Maybe the best way would be to just use `bash` or `powershell` from the start?

python/uv/private/lock.bzl
```diff
@@ -15,109 +15,242 @@
...
+    args.add_all(ctx.files.srcs)
+    args.add_all(["--custom-compile-command", "bazel run //{}:{}.update".format(
+        ctx.label.package,
+        ctx.label.name,
+    )])
+    args.add_all([
+        "--no-python-downloads",
+        "--no-cache",
+    ])
+    args.add_all(ctx.attr.args)
```
aignas, line 54: I think adding `uv` path here would be a good way to get rid of the need for `runfiles` if we decided to drop the current structure.

aignas (commented):
OK, after doing self review I think I want to have:
* A script for launching `uv` from `bash` that would be compatible with UNIX.
* A script for launching `uv` from `powershell` that would be compatible with Windows (generative AI may help with initial translation here).
* Rewrite the internals a little to drop the Python usage (or at least most of it).

python/uv/private/BUILD.bazel
```diff
@@ -13,6 +13,16 @@
 # limitations under the License.
 
 load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
+load("//python/private:bzlmod_enabled.bzl", "BZLMOD_ENABLED")  # buildifier: disable=bzl-visibility
+
+exports_files(
+    srcs = [
+        "copy.py",
+        "pip_compile.py",
```
aignas, line 21:
```suggestion

```

python/uv/private/lock.bzl
```diff
@@ -15,22 +15,258 @@
...
+
+    py_runtime = ctx.toolchains[_py_toolchain].py3_runtime
+
+    cmd = ctx.actions.declare_file(ctx.label.name)
+    ctx.actions.expand_template(
+        template = ctx.files._template[0],
+        substitutions = {
+            " uv ": " {} ".format(uv.path),
+            "bazel_out": output.path,
+            "replace": existing_output[0].path,
```
rickeylev, line 55: nit: use more obvious replacement strings

python/uv/private/lock.bzl
```diff
@@ -15,22 +15,258 @@
...
+        },
+        output = cmd,
+        is_executable = True,
+    )
+
+    run_args = []
+    lock_args = ctx.actions.args()
+    lock_args.add("--no-python-downloads")
+    run_args.append("--no-python-downloads")
```
rickeylev, line 64: nit: define a local function to add to both. Helps reduce the repetition and make it more clear if someone is purposefully omitted.

Alternatively, the run.args arg can accept a list of args, e.g. `run(args=[common_args, lock_only_args])`

python/uv/private/lock.bzl
```diff
@@ -15,22 +15,258 @@
...
+    if not ctx.attr.strip_extras:
+        lock_args.add("--no-strip-extras")
+        run_args.append("--no-strip-extras")
+
+    update_command = "bazel run //{}:{}".format(
+        ctx.label.package,
+        ctx.attr.update_target,
+    )
+    lock_args.add("--custom-compile-command", update_command)
+    run_args.extend(("--custom-compile-command", "'{}'".format(update_command)))
```
rickeylev, line 79: nit: instead of quoting/escaping here, just pass the raw string and have consumer escaping things.

python/uv/private/lock.bzl
```diff
@@ -15,22 +15,258 @@
...
+
+    lock_args.add("--no-progress")
+    lock_args.add("--quiet")
+    lock_args.add("--python", py_runtime.interpreter)
```
rickeylev, line 83: Also handle a system python (when interpreter isn't set, but interpreter_path is)

python/uv/private/lock.bzl
```diff
@@ -15,22 +15,258 @@
...
+        "template": "",
+        "uv": "",
+    },
+)
+
+def _lock_impl(ctx):
+    args = ctx.attr.args
+    srcs = ctx.files.srcs
+    existing_output = ctx.files.existing_output
+    output = ctx.outputs.output
```
rickeylev, line 42: I'd prefer not having a pre-declared output.
* Havin a predeclared output makes it ambiguous about what one should use when consuming the output (":name", or ":output.txt" ?).
* Forcing people to depend on the target lets us return extra providers, which consumers can make use of without having to update users to switch off the predeclared output

python/uv/private/lock.bzl
```diff
@@ -15,22 +15,258 @@
...
+
+    lock_args.add("--no-progress")
+    lock_args.add("--quiet")
+    lock_args.add("--python", py_runtime.interpreter)
+    lock_args.add_all(args)
+    lock_args.add_all(srcs)
+
+    ctx.actions.run(
+        executable = cmd,
+        mnemonic = "RulesPythonLock",
```
rickeylev, line 89: nit: work "Uv" into the mnemonic somewhere: RulesPythonUvLock.

I'm mixed on the RulePython prefix. We use Py elsewhere, so that would be shorter and more consistent, but I don't have a strong opinion on that part.

python/uv/private/lock.bzl
```diff
@@ -15,22 +15,258 @@
...
+        inputs = srcs + ctx.files.existing_output,
+        outputs = [output],
+        arguments = [lock_args],
+        tools = [cmd],
+        progress_message = "Locking requirements using uv",
```
rickeylev, line 94: nit: include something to better identify what's being operated on/produce. i.e labrl, input, and/or output. THis is the message that shows up in the console, after all.

python/uv/private/lock.bzl
```diff
@@ -15,22 +15,258 @@
...
+        env = ctx.attr.env,
```
rickeylev, line 95: uv might invoke python, right? There's a couple environment variables that are best to set for build actions with python -- PYTHONHASHSEED? I think the precompile rules set them. ah gtg.

rickeylev (commented):
got about half way through

python/uv/private/lock.bzl
```diff
@@ -15,22 +15,258 @@
...
+    run_args.extend(("--custom-compile-command", "'{}'".format(update_command)))
+
+    lock_args.add("--no-progress")
+    lock_args.add("--quiet")
+    lock_args.add("--python", py_runtime.interpreter)
+    lock_args.add_all(args)
+    lock_args.add_all(srcs)
+
+    ctx.actions.run(
+        executable = cmd,
```
rickeylev, line 88: I think this will be cleaner if the build action calls uv directly without the shell script wrapper. The run-wrapper can generate the sh wrapper instead.

rickeylev (commented):
Also, to clarify, this is the sort of interface I imagined:

1. `bazel build //:requirements` generates a requirements file using a build action
2. `bazel run //:requirements.update` generates a requirements file using a build action and copies it into the local client to update the requirements file
3. `bazel run //:requirements.run -- <args>` more of a direct invocation; runs uv directly; this is for debugging or experimenting with settings without having to modify the BUILD file.

The reason for (2) to use the output of (1) is because that's where the magic of bazel happens. By this i mean: build actions have isolated/deterministic/hermetic capabilities, can run remotely, are more amenable to having transitions applied, and are better about not having different output per user machine.

python/uv/private/lock.bzl
```diff
@@ -15,22 +15,258 @@
...
+    ] + [
+        "--python",
+        py_runtime.interpreter.short_path,
+    ]
+
+    executable = ctx.actions.declare_file(ctx.label.name)
+    ctx.actions.expand_template(
+        template = info.template,
+        substitutions = {
+            "replace": "{}/{}".format(ctx.label.package, ctx.attr.output),
```
rickeylev, line 206: Please use template strings that are more obvious

python/uv/private/lock.bzl
```diff
@@ -15,22 +15,258 @@
...
+    implementation = _lock_run_impl,
+    doc = """\
+""",
+    attrs = {
+        "lock": attr.label(
+            doc = "The lock target that is doing locking in a build action.",
+            providers = [_LockInfo],
+        ),
+        "output": attr.string(
+            doc = "The output that we would be updated.",
```
rickeylev, line 238: This is a path relative to something, right? Please indicate what it's relative to.

python/uv/private/lock.bzl
```diff
@@ -15,22 +15,258 @@
...
+
+    return None
+
+def lock(*, name, srcs, out, args = [], env = None, **kwargs):
     """Pin the requirements based on the src files.
 
     Differences with the current {obj}`compile_pip_requirements` rule:
-    - This is implemented in shell and `uv`.
+    - This is implemented as a rule that performs locking in a build action.
+    - Additionally one can use the runnable target.
```
rickeylev, line 268: Please document the additional public targets with a `Generates public targets:` block, e.g.

```
Generates public targets:
  * `<name>` bla bla
  * `<name>.update` bla bla
```
```

File: python/uv/private/lock.bzl [hunk 19]
```diff
@@ -40,86 +276,78 @@ def lock(*, name, srcs, out, upgrade = False, universal = True, args = [], **kwa
...
-
-    native.genrule(
+    update_target = "{}.update".format(name)
+    locker_target = "{}.run".format(name)
+    maybe_out = _maybe_file(out)
+    out_new = out + ".new"
+    target_compatible_with = kwargs.pop("target_compatible_with", select({
+        "@platforms//os:windows": ["@platforms//:incompatible"],
+        "//conditions:default": [],
+    })) if BZLMOD_ENABLED else ["@platforms//:incompatible"]
```
rickeylev, line 290: This is hard to read with the select nested inside pop() and the ternary. Please write as a regular if-else.

python/uv/private/lock.bzl
```diff
@@ -40,86 +276,78 @@ def lock(*, name, srcs, out, upgrade = False, universal = True, args = [], **kwa
...
-        ],
+        template = "//python/uv/private:copy.py",
+        substitutions = {
+            'dst = ""': 'dst = "{}/{}"'.format(pkg, out),
+            'src = ""': 'src = "{}/{}"'.format(pkg, out_new),
+        },
     )
 
-    py_binary(
+    # TODO @aignas 2025-03-13: create a py_test-like target if the design of
```
rickeylev, line 340: TBH, I don't understand the point of the diff test. Dependencies change pretty regularly it just seems like a source of spantaneous churn?

A test that verifies requirements.txt came from the right requirements.in version sounds appealing (reminds you to re-run locking), but, doesn't the diff test perform locking again?

python/uv/private/lock.bzl
aignas: We cannot do this because we need to copy the source copy of the requirements file to the ctx.outputs.output if it already exists. This is so that uv would know what versions to use for the final solution.

python/uv/private/lock.bzl
aignas: Yeah, the test is something that exists today and I am not sure about the usefulness of it either. I'll include it when I can address the changes, but we can discuss later if it is really needed. It can be opt-in target, where users can use native_test or something similar if they want.

aignas:
OK, addressed the comments, PTAL.

I see that Windows is failing, because something is not compatible with the CI Windows version. I wonder if this means we should just tell users that Windows may be unsupported?

python/uv/private/lock.bzl
aignas: I don't see an easy way to do this and I don't think it matters a lot because the output is a lock file rather than the wheels themselves, so it should be fine here.

python/uv/private/lock.bzl
aignas: Ah, yes we can, we just need to use `run_shell`.

aignas:
Open questions:
* Currently providing `keyring` and other authentication things need to happen via `system` and this has not been fully tested. This might be relevant for #2663 were users need to be able to configure credential helpers for pulling packages from internal mirrors.
* Should we use `python_toolchain` or a target to the current python interpreter? Passing `//python:none` could act as a way to indicate that I need python from the system, i.e. `python`. This would mean that we could use the `//python/bin:interpreter` which could include dependencies used during locking (e.g. `keyring` and similar). Right now I am not sure how to get that wired through. Maybe we should have a few more tests that setup a mirror that needs the `keyring` dep to connect, otherwise it is hard to not break the behaviour, but I would like to merge this as is for now because the PR is getting big and it is hard to keep track of all of the things.

python/uv/private/lock.bzl
```diff
@@ -13,113 +13,424 @@
...
+    args.add_all(srcs)
+
+    args.run_shell.add("--output-file", output)
+
+    # These arguments does not change behaviour, but it reduces the output from
+    # the command, which is especially verbose in stderr.
+    args.run_shell.add("--no-progress")
+    args.run_shell.add("--quiet")
+
+    ctx.actions.run_shell(
```
rickeylev, line 139: I'm having trouble understanding why this step needs the copy step as part of its execution.

Isn't this the same behavior?

```
srcs = list(ctx.attr.srcs)
if existing_file:
  srcs.append(existing_file)
output = declare_file(name + ".out")
ctx.actions.run([uv, "--output={output}"] + srcs, inputs=srcs, output=output)
```

uv is going to overwrite whatever --output specifies, right?

python/uv/private/lock.bzl
```diff
@@ -13,113 +13,424 @@
...
+    target_compatible_with = kwargs.pop("target_compatible_with", None)
+    if target_compatible_with == None and not BZLMOD_ENABLED:
+        target_compatible_with = ["@platforms//:incompatible"]
+    elif target_compatible_with == None:
+        target_compatible_with = []
 
-    native.genrule(
+    # FIXME @aignas 2025-03-17: should we have one more target that transitions
+    # the python_version to ensure that if somebody calls `bazel build
+    # :requirements` that it is locked with the right `python_version`?
```
rickeylev, line 378: A separate target, no. An attribute with rule-level cfg transition, yes.

This also enables tricks like this:

```
lock(python_version="3.10", srcs=select(":py310": "requirements_310.txt", ...))
```

Similarly, because an `attr.output` is not used, the outputs can be varied to e.g. include the python version (or platform, etc) into the output file name.

python/uv/private/lock.bzl
```diff
@@ -13,113 +13,424 @@
...
+    This macro creates the following targets:
+    - `name`: the target that creates the requirements.txt file in a build
+      action.
+    - `name.run`: a runnable target that can be used to pass extra parameters
+      to the same command that would be run in the `name` action. This will
+      update the source copy of the requirements file.
+    - `name.update`: a target that can be run to update the source-tree version
+      of the requirements lock file. The output can be fed to the
+      {obj}`pip.parse` bzlmod extension tag class. Note, you can use
+      `native_test` to wrap this target to make a test.
```
rickeylev, line 337: nit: the docs should better distinguish between run vs update.

Run: You can customize the args via the command line, but it requires being able to run uv (and possibly python) directly on your host.

Update: You can't customize the args via command line, but you can use RBE to generate requirements (offload execution and run for different platforms)

rickeylev (approved):

python/uv/private/lock.bzl
aignas: `uv` expects the previous output to be in the location that is defined by `--output={output}`. If you pass it as a source then you will have extra log lines `via requirements-existing.txt`, which is not what you want here.

python/uv/private/lock.bzl
```diff
@@ -13,113 +13,424 @@
...
+            ),
+        },
+        output = executable,
+        is_executable = True,
+    )
+
+    return [
+        DefaultInfo(
+            executable = executable,
+            runfiles = ctx.runfiles(transitive_files = info.srcs),
```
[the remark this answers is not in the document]
aignas: I have added a transition, but I am still new as to how to ensure that this will be in `exec` configuration.

I can follow this up with a separate PR.

python/uv/private/lock.bzl
aignas: Thanks for suggestions.

aignas:
RBE is failing with an error:
```
bazel-out/k8-opt-exec-ST-150d2d5f4ddd/bin/python/private/python3: error while loading shared libraries: /b/f/w/bazel-out/k8-opt-exec-ST-150d2d5f4ddd/bin/python/private/../lib/libpython3.11.so.1.0: cannot open shared object file: No such file or directory
```

Not sure exactly why this is happening because I am passing `interpreter.files_to_run` to the action.

I think the documentation mentioning to use `.files_to_run` may be wrong.

aignas:
It seems that the `current_py_exec_toolchain` had a bug that was not letting me use it in RBE. My analysis went as follows:
1. Setup a minimal RBE on my machine like mentioned in #2241
2. Add print statements to inspect what is in the sandbox:
    i. The python toolchain with all of the files was there
    ii. The `current_py_exec_toolchain` symlink was not a symlink and instead it was a copy.
3. Given that the symlink was dereferenced, there are multiple ways to solve this:
    i. Use a dangling symlink, like the one where we are pointing outside the sandbox
       when non-hermetic toolchain is used.
    ii. Somehow symlink all of the directories so that the `lib` and other
       folders can be found.

I chose the 3.i. method because I only need one extra symlink and it does not
require exposing extra data through the `py_runtime`. E.g. I would need to
expose the contents of `bin` folder most likely and the `lib`, etc folders for
everything to work properly.

Maybe the 3.ii solution could be also beneficial for creating #2156 and setting up
the `venv` inside the `py_executable` rules, but we can look at that later.

python/uv/lock.bzl
```diff
@@ -14,6 +14,32 @@
...
+    srcs = ["pyproject.toml"],
+)
+
+native_test(
+    name = "requirements_test",
+    src = "requirements.update",
+)
+```
+
 EXPERIMENTAL: This is experimental and may be removed without notice
```
aignas, line 43:
```suggestion
EXPERIMENTAL: This is experimental and may be changed without notice.
```

python/private/py_exec_tools_info.bzl
```diff
@@ -24,15 +24,31 @@ When running it in an action, use `DefaultInfo.files_to_run` to ensure all its
...
+This does not work correctly in case of RBE, please use exec_runtime instead.
+
+Once https://github.com/bazelbuild/bazel/issues/23620 is resolved this warning
+may be removed.
+:::
+""",
+        "exec_runtime": """
+:type: PyRuntimeInfo | None
+
+The forwarded {obj}`PyRuntimeInfo` field.
```
aignas, line 51:
```suggestion
The forwarded {obj}`PyRuntimeInfo` field.

:::{versionadded} VERSION_NEXT_FEATURE
:::
```

amaranthjinn:
Would this feature fix the issue https://github.com/bazel-contrib/rules_python/issues/2640?

aignas:
> Would this feature fix the issue #2640?

It would not, because the linked issue is about building wheels rather than locking them.

rickeylev:
> python can't find its .so files when a "regular" symlink action is used ... bazel #23620

Yeah, I'm pretty sure that's what you're seeing. I haven't looked at the PR code yet, but what I recall was you can't simply `ctx.actions.symlink(<output>, <underlying interpreter>)` because Bazel RBE is prone to creating a copy instead of a symlink. It works locally because Python has a behavior where it will check if argv[0] is a symlink, and if so, realpath() to find the _actual_ location of the python interpreter (and thus all the runtime's files). I think this is to support stuff like venvs, or creating a convenience symlink to the interpreter in one place while it's actually installed in another (e.g. /usr/lib/python3 is a symlink to /usr/lib/python3.10, or whatever).

Using declare_symlink, or a wrapper script, should work, though. I'll have a look at the PR now.

aignas:
> Beh. Looking at current_interpreter_executable.bzl, I think changing L93 to use declare_symlink() should work?
> 
> I tried setting up a local RBE to test it, but couldn't get bazel and the RBE talking. I'll have to try again when I have more time

If you use `declare_symlink` then you cannot use a `target_file` and you need to use `target_path`. So that is only possible to work with `target_path` in my tests. This is documented in https://bazel.build/rules/lib/builtins/actions#symlink.

tests/uv/lock/lock_tests.bzl
```diff
@@ -0,0 +1,105 @@
...
+            "testdata/constraints.txt",
+            "testdata/constraints2.txt",
+        ],
+        build_constraints = [
+            "testdata/build_constraints.txt",
+            "testdata/build_constraints2.txt",
+        ],
+        # It seems that the CI remote executors for the RBE do not have network
+        # connectivity. Is it only our setup or is it a property of RBE?
+        tags = ["no-remote-exec"],
```
aignas, line 40: FYI this is where the RBE is disabled for our tests

rickeylev:
I got an RBE setup locally. Yeah, current_interpreter_executable (the thing that backs ExecTools.exec_interpreter) does indeed look entirely broken with RBE. Argh >.<

So...

* An executable rule has to define its own output file. It can't forward on another file. Hence e.g. current_interpreter_executable has to call e.g. declare_file/symlink
* If it uses declare_file(), it can't use symlink(), because RBE will create a copy, and Python can no longer traverse back to its _actual_ install location.
* If it uses declare_symlink, then symlink() has to write either a bin-relative path (the file that is a sibling of the runfiles directory), or a runfiles-relative path (the copy of the executable within the runfiles tree). The former allows e.g. `ctx.actions.run(executable=...)` to work; the latter allows e.g. `ctx.actions.run_shell("<runfiles path to to executable>")` to work.

The only way I can think of to make both work is, essentially, to make the output executable a wrapper. e.g. shell code that figures out how to locate the file it wants to run and execs it.

Or, maybe if py_runtime() is directly executable? i.e. sets executable=True, and returns `DefaultInfo(executable=...)`, and the whole target gets forwarded on.

The reason I'm so keen on having an exectuable=True attribute (i.e. a thing with FilesToRun provider) that is fed to ctx.actions.run() is that is _supposed_ to be the proper abstraction -- "run the interpreter, let its executable rule figure out the details". This is in comparison to e.g. having to directly use PyRuntime, then manually pass PyRuntime.files etc to ctx.actions.run.

Well, its late now, so gotta log off.

aignas:
Hmmm... since I have a question on RB slack about the `requires-network`, I can wait on this.

I kind of am getting where you are coming from - just running python should not be that hard and should not require `py_runtime`. I'll think about it as well, gotta go.

rickeylev:
I poked this some more and have some ideas, but want to explore them a bit more.

I don't want to block this PR on them, though. How about for now, revert the changes to the exec toolchain stuff. The uv rule can still get at the PyRuntime object via exec_interpreter: `exec_interpreter[ToolchainInfo].py3_runtime`.

python/private/sentinel.bzl
```diff
@@ -25,6 +25,10 @@ SentinelInfo = provider(
 
 def _sentinel_impl(ctx):
     _ = ctx  # @unused
-    return [SentinelInfo()]
+    return [
+        SentinelInfo(),
+        # Also output ToolchainInfo
```
rickeylev, line 30:
```suggestion
        # Also output ToolchainInfo to allow it to be used for no-op toolchains
```

python/uv/private/lock.bzl
```diff
@@ -12,114 +12,484 @@
...
+        outputs = [output],
+        arguments = [args.run_shell],
+        tools = [
+            uv,
+            python_files,
+        ],
+        progress_message = "Creating a requirements.txt with uv: //{}:{}".format(
+            ctx.label.package,
+            ctx.label.name,
+        ),
```
rickeylev, line 151: nit: Use %{label} instead

```suggestion
        progress_message = "Creating a requirements.txt with uv: %{label}",
```

python/uv/private/lock.bzl
```diff
@@ -12,114 +12,484 @@
...
+_python_version_transition = transition(
+    implementation = _transition_impl,
+    inputs = [_PYTHON_VERSION_FLAG],
+    outputs = [_PYTHON_VERSION_FLAG],
+)
+
+_lock = rule(
+    implementation = _lock_impl,
+    doc = """\
+""",
```
rickeylev, line 192: nit: add doc or just omit the doc attribute

python/uv/private/lock.bzl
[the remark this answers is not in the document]
rickeylev: The transition LGTM.

When a transition is applied to the rule itself, it decides what the "target" configuration is for the current target. It doesn't affect the exec config directly.

When toolchain resolution occurs, Bazel finds a toolchain that is compatible with the current target configuration. e.g. if python_version=3.12.1 is in the target configuration, then Bazel looks for a matching exec_tools toolchain with target_compatible_with=3.12. The e.g. exec_interpreter attribute will be in the exec config, but that's fine; the toolchain is claiming all its pieces are intended to produce output valid for 3.12.

HTH

tests/uv/lock/lock_tests.bzl
```diff
@@ -0,0 +1,105 @@
...
+        constraints = [
+            "testdata/constraints.txt",
+            "testdata/constraints2.txt",
+        ],
+        build_constraints = [
+            "testdata/build_constraints.txt",
+            "testdata/build_constraints2.txt",
+        ],
+        # It seems that the CI remote executors for the RBE do not have network
+        # connectivity. Is it only our setup or is it a property of RBE?
```
rickeylev, line 39: It's not intrinsic to RBE, so must be something wit our RBE.

rickeylev (approved):
Just remove the exec tools toolchain changes as I mentioned in the other comment, otherwise LGTM

PR #2666, opened 2025-03-15
feat(rules): allow deriving custom rules from core rules

rickeylev:
This exposes public functions for creating builders for py_binary, py_test, and py_library.
It also adds some docs and examples for how to use them.

I'm calling this a "volatile" API -- it's public, but the pieces that comprise
it (e.g. all the rule args, attributes, the attribute args, etc) are likely to change
in various ways, and not all modifications to them can be supported in a backward
compatible way. Hence the "volatile" term:
* hold it gently and its fine
* shake it a bit and its probably fine
* shake it moderately and something may or may not blow up
* shake it a lot and something will certainly blow up.

### DIFF
--- a/python/api/libraries.bzl
+++ b/python/api/libraries.bzl
@@ -14,7 +14,7 @@
 
 """
 {#python-apis-libraries-bzl}
-Loading-phase APIs specific to executables (binaries/tests).
+Loading-phase APIs specific to libraries.
 
 :::{versionadded} VERSION_NEXT_FEATURE
 :::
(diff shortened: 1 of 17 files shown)


python/api/libraries.bzl
```diff
@@ -0,0 +1,27 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+{#python-apis-libraries-bzl}
+Loading-phase APIs specific to executables (binaries/tests).
```
aignas, line 17:
```suggestion
Loading-phase APIs specific to libraries.
```

aignas (approved):

Issue #2539, opened 2024-12-27
Support `python -m` invocation style


rickeylev:
Python supports two ways to start programs: by file, or by module name.

Today, only file is supported. This is a FR for `-m` to also be supported. There's two main use cases for this style:
1. Using a file not in the same Bazel package as main.
2. Specifying a main that comes from dependencies, and referencing the actual main file directly is problematic.

Some example use cases:
* When using a macro to generate a test/binary.
* pytest: a common main is used and the tests to execute are passed as arguments or discovered at runtime. This basically a special case of the above.

This should be fairly easy to implement using bootstrap script. It just requires passing the name of the main module to the stage2 bootstrap instead of the path to the main module.

Docs for `-m`: https://docs.python.org/3.10/using/cmdline.html#cmdoption-m

See also: runpy.run_module: https://docs.python.org/3.10/library/runpy.html#runpy.run_module

aignas:
A note for a possible cleanup - the py_console_script_binary could diructly leverage this feature. It could be also harvested for the initial implementation to support this.

aignas:
This has made it possible to integrate `pytest-bazel` so much easier:

"""A small wrapper for pytest-bazel.

Taken from: https://pytest-bazel.readthedocs.io/latest/usage.html
"""

load("@rules_python//python:py_test.bzl", "py_test")

def py_pytest_test(*, deps = [], data = [], **kwargs):
    """Use pytest-bazel for running pytest tests

    Args:
        deps: Deps to pass to the test.
        data: The data to pass to the test rule.
        **kwargs: Extra args passed to the py_test.
    """

    py_test(
        deps = deps + [
            "@pypi//pytest_bazel",  # assuming your hub repo name is `pypi`.
        ],
        main_module = "pytest_bazel",
        data = data + [
            "//:pyproject.toml",
        ],
        **kwargs,
    )

See aignas/pytest_bazel#14

Thank you.

PR #2671, opened 2025-03-17
feat(rules): add main_module attribute to run a module name (python -m)

rickeylev:
This implements the ability to run a module name instead of a file path, aka
`python -m` style of invocation.

This allows a binary/test to specify what the main module is without having to
have a direct dependency on the entry point file.

As a side effect, the `srcs` attribute is no longer required.

Fixes https://github.com/bazelbuild/rules_python/issues/2539

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -96,6 +96,8 @@ Unreleased changes template.
   ([#1647](https://github.com/bazelbuild/rules_python/issues/1647))
 * (rules) Added {obj}`interpreter_args` attribute to `py_binary` and `py_test`,
   which allows pass arguments to the interpreter before the regular args.
+* (rules) Added {obj}`main_module` attribute to `py_binary` and `py_test`,
+  which allows specifying a module name to run (i.e. `python -m <module>`).
 
 {#v0-0-0-removed}
 ### Removed
--- a/python/private/py_executable.bzl
+++ b/python/private/py_executable.bzl
@@ -126,6 +126,24 @@ Optional; the name of the source file that is the main entry point of the
 application. This file must also be listed in `srcs`. If left unspecified,
 `name`, with `.py` appended, is used instead. If `name` does not match any
 filename in `srcs`, `main` must be specified.
+
+This is mutually exclusive with {obj}`main_module`.
+""",
+        ),
+        "main_module": lambda: attrb.String(
+            doc = """
+Module name to execute as the main program.
+
+When set, `srcs` is not required, and it is assumed the module is
+provided by a dependency.
+
+See https://docs.python.org/3/using/cmdline.html#cmdoption-m for more
+information about running modules as the main program.
+
+This is mutually exclusive with {obj}`main`.
+
+:::{versionadded} VERSION_NEXT_FEATURE
+:::
 """,
         ),
         "pyc_collection": lambda: attrb.String(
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -96,6 +96,8 @@ Unreleased changes template.
   ([#1647](https://github.com/bazelbuild/rules_python/issues/1647))
 * (rules) Added {obj}`interpreter_args` attribute to `py_binary` and `py_test`,
   which allows pass arguments to the interpreter before the regular args.
+* (rules) Added {obj}`main_module` attribute to `py_binary` and `py_test`,
+  which allows specifying a module name to run (i.e. `python -m <module>`).
```
aignas, line 100: Could you also refactor `py_console_script_binary` to use that, please?

This is what I would have asked because the current impl of `py_console_script_binary` is:
1. get the script name as input from the user.
2. Get the `main_module` value from a file (i.e. `entrypoints.txt`)
3. Generate a `py_binary` `srcs` files.
4. Create a `py_binary` target.

I think we could cut out number 3 and use the builder API to make everything a single `py_console_script_binary` rule. no?

That said, this makes the PR much larger in scope, so I am fine if it is a followup.

python/private/py_executable.bzl
```diff
@@ -126,6 +126,24 @@ Optional; the name of the source file that is the main entry point of the
 application. This file must also be listed in `srcs`. If left unspecified,
 `name`, with `.py` appended, is used instead. If `name` does not match any
 filename in `srcs`, `main` must be specified.
+
+This is mutually exclusive with {obj}`main_module`.
+""",
+        ),
+        "main_module": lambda: attrb.String(
```
aignas, line 133: nit: what about `main` becoming `main_module` if it cannot find the file in `srcs`?

That said, then we would have worse error messages, so this solution is also good.

aignas (approved):

groodt (approved):
LGTM

🎉

groodt:
Im looking forward to trying:
`main_module = "pytest.main",`

Or even better:
`main_module = "pytest_bazel.main",`

CHANGELOG.md
rickeylev: Ah, I think I see what you mean. I took a look at that step (3) more closely and the code it generates looks very much like the `runpy.run_module()` logic I've added and the surrounding bootstrap logic. py_console_script_binary is like running a script by path in order to run a script by module name, which seems silly.

Though, reading the entry points spec: https://packaging.python.org/en/latest/specifications/entry-points/#use-for-scripts

The two methods aren't interchangable. The bootstrap runs a module as name `__main__` and relies on simply importing the module to perform appropriate "run" logic. In comparison, an entry point is a `(module_name, function_name)` tuple and loads the desired module as its normal name, then invokes the specified (no-arg) function.

Folding that "import module, run func" logic into the stage2 bootstrap does seem appealing. However, going from `(entry_point_txt_path, script_name) -> (module_name, func_name)` means the stage2 bootstrap is no longer a simple template expansion (a tool must run to parse entry_points.txt and put the content into the template). Not a huge deal, but that logic is pretty deep in py_executable.bzl.

Maybe an easier route would be attr `main_module_target`? Take a file as input; the file contains the main_module string. At build time, the stage2 template expansion reads that file. We could make this attribute internal by making py_console_script_binary a derived rule, and have the base py_binary impl check for the attribute.

Slightly different idea: check `srcs` for a special provider that gives a file with the modulename/funcname content. If present, feed it into the stage2 bootstrap.

Another complication here is `py_console_script_binary` takes a `binary_rule` arg to customize what rule is used. We did that to allow the version aware rules to be used, but it is technically part of the public API now.

...but in any case, yeah, this feels out of scope for this PR.

python/private/py_executable.bzl
rickeylev: I'd rather not because it's not clear how to transform the file name into a module name.

PR #2676, opened 2025-03-17
chore: account for new GH org of standalone interpreter

alexeagle:
The repo was donated. It also trivially removes the need for a redirect on the URL that fetches artifacts.


aignas (approved):
LGTM, thanks.

PR #2684, opened 2025-03-20
chore: ignore releasing.md for version string check

rickeylev:
The RELEASING.md docs contain the VERSION_NEXT marker string in their docs, so also
have to be ignored by the release script.


groodt (approved):
LGTM

PR #2688, opened 2025-03-21
chore: replace bazelbuild with bazel-contrib

rickeylev:
This was done using `grep | xargs sed`.

BCR presubmits require that the list of repositories match where downloads come from

Along the way, also update the URL homepages to bazel-contrib and change the email
to my personal instead of work email.


groodt (approved):
LGTM

Issue #2690, opened 2025-03-21
Incorrect platform-specific dependency selected from universal requirements.txt using pip.parse


ewianda:
# 🐞 bug report

### Affected Rule

`pip.parse` 

### Is this a regression?

 Unknown — this is observed with the current behavior in the latest release.

### Description

  
I'm using a universal `requirements.txt` generated with [uv](https://github.com/astral-sh/uv)  

I want to use platform-specific dependencies like this:

#### `requirements.in`:

```ini
optimum[onnxruntime]; sys_platform == 'darwin' and sys_platform != 'linux'
optimum[onnxruntime-gpu]; sys_platform == 'linux'
```

#### `requirements.txt` (generated):

```ini
optimum[onnxruntime]==1.17.1 ; python_full_version < '3.12' and platform_python_implementation == 'CPython' and sys_platform == 'darwin'
optimum[onnxruntime-gpu]==1.17.1 ; python_full_version < '3.12' and platform_python_implementation == 'CPython' and sys_platform == 'linux'
```

#### `MODULE.bazel`:

```python
pip.parse(
    requirements_by_platform = {
        "//:requirements.txt": "linux_*,osx_*",
    },
)
```

However, when building on **macOS**, Bazel attempts to use `optimum[onnxruntime-gpu]` instead of the appropriate `optimum[onnxruntime]`.

It seems that the logic in the following file prioritizes the wrong requirement:

🔗 https://github.com/bazel-contrib/rules_python/blob/032f6aa738a673b13b605dabf55465c6fc1a56eb/python/private/pypi/parse_requirements.bzl#L114-L126

As a result, the build fails with:

## 🔥 Exception or Error

<pre><code>
rules_python~~pip~py_deps/onnxruntime_gpu/BUILD.bazel:6:12: configurable attribute "actual" in @@rules_python~~pip~py_deps//onnxruntime_gpu:_no_matching_repository doesn't match this configuration: No matching wheel for current configuration's Python version.

The current build configuration's Python version doesn't match any of the Python
wheels available for this distribution. This distribution supports the following Python
configuration settings:
    //_config:is_cp311_cp311_manylinux_2_28_x86_64
    //_config:is_cp311_cp311_manylinux_2_31_x86_64
    //_config:is_cp311_cp311_manylinux_x86_64
</code></pre>

To determine the current config, I ran:

```
bazel config <config_id>
```

## 🔬 Minimal Reproduction

Repro steps:
[...]

PR #2692, opened 2025-03-23
refactor(pypi): implement PEP508 compliant marker evaluation

### LINKED ISSUE #2423: [pypi] PEP508 Remove the need for Python in env_marker evaluation in hub repos

aignas:
This implements the PEP508 compliant marker evaluation in starlark and
removes the need for the Python interpreter when evaluating requirements
files passed to `pip.parse`. This makes the evaluation faster and allows
us to fix a few known issues (#2690).

In the future the intent is to move the `METADATA` parsing to pure
starlark so that the `RequiresDist` could be parsed in starlark at the
macro evaluation or analysis phases. This should make it possible to
more easily solve the design problem that more and more things need to
be passed to `whl_library` as args to have a robust dependency parsing:
* #2319 needs the full Python version to have correct cross-platform
  compatible `METADATA` parsing and passing it to `Python` and back makes
  it difficult/annoying to implement.
* Parsing the `METADATA` file requires the precise list of target
  platform or the list of available packages in the `requirements.txt`.
  This means that without it we cannot trim the dependency tree in the
  `whl_library`. Doing this at macro loading phase allows us to depend
  on `.bzl` files in the `hub_repository` and more effectively pass
  information.

I can remotely see that this could become useful in `py_wheel` or an building
wheels from sdists as the environment markers may be present in various source
metadata as well. What is more `uv.lock` file has the env markers as part of
the lock file information, so this might be useful there.

Work towards #2423
Work towards #260
Split from #2629


rickeylev (approved):

python/private/pypi/pep508_env.bzl
```diff
@@ -0,0 +1,109 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""This module is for implementing PEP508 environment definition.
+"""
+
+_platform_machine_values = {
+    "aarch64": "arm64",
+    "ppc": "ppc64le",
```
aignas, line 20:
```suggestion
    "ppc64le": "ppc64le",
```

This got fixed elsewehere, #2577

aignas:
Refreshed the lockfile before and after the change:
```
bazel mod deps --lockfile_mode=refresh
```

And saw no difference in entries except for the expected:
```
$ diff MODULE.bazel.lock{before,after}
161c161
<         "bzlTransitiveDigest": "HtK14jTL6GRmVkjtcDvWJ+J5LmpFcDz5iGWjGt4RORc=",
---
>         "bzlTransitiveDigest": "1YngqOV37S3bLvhBu3wQ+h/bYrkyRpmh11ur6JIqqu0=",
164d163
<           "@@+internal_deps+pypi__packaging//packaging-24.0.dist-info/RECORD": "be1aea790359b4c2c9ea83d153c1a57c407742a35b95ee36d00723509f5ed5dd",
167,168d165
<           "@@//python/private/pypi/requirements_parser/resolve_target_platforms.py": "42bf51980528302373529bcdfddb8014e485182d6bc9d2f7d3bbe1f11d8d923d",
<           "@@//python/private/pypi/whl_installer/platform.py": "763d705ceaa41c87d8a6cfe8c6bc261a20881628b96f2fb4c4fef4b9befa53fe",
```

aignas:
OK, I think this is ready to be merged. This is probably already higher fidelity/more robust than the Python impl of `platform.py` in our repo.

Issue #2363, opened 2024-10-31
[pypi] Support pulling dependencies using direct url specifiers without using an index


mering:
# 🐞 bug report

### Affected Rule

Module extension:
```
pip = use_extension("@rules_python//python/extensions:pip.bzl", "pip")
pip.parse(...)
```

### Is this a regression?

_Not sure._

### Description

The repository rule for pip dependencies doesn't use [`repository_ctx.download()`](https://bazel.build/rules/lib/builtins/repository_ctx#download). This breaks usages of `--experimental_downloader_config` for example to use a transparent dependency mirror.

## 🔬 Minimal Reproduction

https://github.com/bazelbuild/rules_python/blob/b6fc2a080d4884510dea02cc77b4e0b8fd3a0ccd/python/private/pypi/whl_library.bzl#L245-L256

https://github.com/bazelbuild/rules_python/blob/b6fc2a080d4884510dea02cc77b4e0b8fd3a0ccd/python/private/pypi/whl_installer/wheel_installer.py#L168-L190

## 🔥 Exception or Error

Stack trace prints to the code linked above.

## 🌍  Your Environment

**Operating System:**

Linux

**Output of `bazel version`:**

7.0.2

**Rules_python version:**

0.37.2

**Anything else relevant?**

aignas:
Please see the documentation here: https://rules-python.readthedocs.io/en/latest/pypi-dependencies.html#bazel-downloader-and-multi-platform-wheel-hub-repository.

This currently has a limitation that it needs hashes for all of the files in the `requirements.txt`. The stabilization of this feature is tracked by #260.

mering:
@aignas How can I use `experimental_index_url` in combination with dependencies without an index like the following:
```
https://github.com/googleapis/proto-breaking-change-detector/archive/refs/tags/v2.4.0.tar.gz
tensorflow_graphics @ git+https://github.com/tensorflow/graphics@ca63cf788978db0bde2db4ce488e102c455da2e7
```

aignas:
The support for direct references has not been added yet, I'll reopen this issue and rename it to scope it to supporting it, but I am not sure when I can work on it.

aignas:
[...]

PR #2695, opened 2025-03-24
feat(pypi/parse_requirements): get dists by version when no hash provied

### LINKED ISSUE #2023: Enable using `experimental_index_url` without having hashes in the lock file
# 🚀 feature request

### Relevant Rules

`pip.parse` with `experimental_index_url` set to a non-empty value.

### Description

Quite often the users create a `requirements.txt` file by just using `pip freeze`. This is not ideal, but it is the reality.

Supporting the setting up of the `select` statements with all available wheels would be beneficial in making the code more robust. It would also enable to more easily support a niche usecase where bazel is used in an integration test where the latest package, so in there we have no version number and no hashes at all.

Do not do it and error without hashes specified. This might not work in cases where the hashes might not be required:
- VCS URL. The sha is the hash, but we should instead just support the direct download URLs better in the `experimental_index_url` code path.
- Direct download URL. The hash is something that we don't strictly need and `bazel` could just print a warning message that the hash needs to be added.

### Description of the implementation

The `parse_requirements` function invokes `get_index_urls` which returns a dict where we have the package URLS by its sha256 hash. Then a subsequent call to `_add_dists` would get all of the files for a particular distribution based on the hash.

Changes needed:
* Change `_add_dists` so that it gets all of the files by hash only if at least one hash is provided, otherwise, get values by version.
* Add unit tests under `//tests/pypi/parse_requirements`

Yanpei-Wang:
This pull request modifies the SimpleAPI HTML parsing to add a new
field where we can get the `sha256` values by package version. This
allows us to very easily fallback to all packages of a particular
version when using `experimental_index_url` if the hashes are not
specified.

The code deciding which packages to query the SimpleAPI for has been
also modified to only omit queries for packages that are included via
direct URL references.

If we fail to get the data from the SimpleAPI, we will fallback to
`pip` and try to install it via the legacy behaviour.

Fixes #2023
Work towards #260
Work towards #1357
Work towards #2363


python/private/pypi/parse_requirements.bzl
```diff
@@ -315,26 +315,46 @@ def _add_dists(*, requirement, index_urls, logger = None):
...
+            if maybe_sdist and not maybe_sdist.yanked:
+                sdist = maybe_sdist
+                continue
+
+            if logger:
+                logger.warn(lambda: "Could not find a whl or an sdist with sha256={}".format(sha256))
+    else:
+        # If no SHA256s provided, try to find distributions by version
+        version = requirement.srcs.version
+        if version:
```
aignas, line 340:
```suggestion
    elif requirement.srcs.version:
        # If no SHA256s provided, try to find distributions by version
```

could be a simple way to define it and it avoids having the indentation.

python/private/pypi/parse_requirements.bzl
```diff
@@ -315,26 +315,46 @@ def _add_dists(*, requirement, index_urls, logger = None):
...
+            # Look for wheels matching the version
+            for whl in index_urls.whls.values():
```
aignas, line 342: Ideally we could have `index_urls.shas_by_version` which we could use instead and unify the codepaths.

tests/pypi/parse_requirements/parse_requirements_tests.bzl
```diff
@@ -563,6 +567,62 @@ def _test_different_package_version(env):
...
+                    requirement_line = "foo==0.0.4 @ https://example.org/foo-0.0.4.whl",
+                    shas = [],
+                    version = "0.0.4",
+                    url = "https://example.org/foo-0.0.4.whl",
+                ),
+                target_platforms = ["linux_x86_64"],
+                whls = [struct(
+                    url = "https://example.org/foo-0.0.4.whl",
+                    filename = "foo-0.0.4.whl",
+                    sha256 = "",
```
aignas, line 596: We should ensure that `whl_library` reports a warning if the sha256 is not specified, but the URL is. We should ask the user to add the sha256 to their manifest. We can print the sha256 value.

tests/pypi/parse_requirements/parse_requirements_tests.bzl
```diff
@@ -563,6 +567,62 @@ def _test_different_package_version(env):
 
 _tests.append(_test_different_package_version)
 
+def _test_optional_hash(env):
+    got = parse_requirements(
+        ctx = _mock_ctx(),
+        requirements_by_platform = {
+            "requirements_optional_hash": ["linux_x86_64"],
+        },
```
aignas, line 575: Could you add a test where the `get_index_urls` is set so that we are testing the newly added code?

aignas (changes requested):
Thanks for the PR. For this to be more robust and performant, it would be great to add an extra field that is the result of SimpleAPI parsing where we store sha256s by version of the package in the index.

The code is here:
https://github.com/bazel-contrib/rules_python/blob/main/python/private/pypi/parse_simpleapi_html.bzl#L96

aignas:
OK, I have updated the implementation based on my own feedback and updated the
docs. Could you please test if it works in your environment?

rickeylev (approved):

Yanpei-Wang:
> OK, I have updated the implementation based on my own feedback and updated the docs. Could you please test if it works in your environment?

Yanpei-Wang:
> OK, I have updated the implementation based on my own feedback and updated the docs. Could you please test if it works in your environment?

Yanpei-Wang:
> OK, I have updated the implementation based on my own feedback and updated the docs. Could you please test if it works in your environment?

Thank you for your help. It's my first time doing this. I mistakenly closed it but reopened it.

Yanpei-Wang:
> OK, I have updated the implementation based on my own feedback and updated the docs. Could you please test if it works in your environment?

Thanks for the updates and feedback! I’m still new to this, so I really appreciate your help. I’ve checked the updated implementation and docs, and I’ve learned a lot. I tested it in my environment, and it works great.

PR #2700, opened 2025-03-25
chore: remove old versions of Python 3.8

aignas:
Python 3.8 has reached EOL and this PR removes old toolchains and most
of the tests. Users can still use it if they register the toolchains
themselves, but `rules_python` will no longer keep testing the toolchains.

Removing the toolchains all-together will be done at a later stage which
may require us to be more clever how we handle asks to include `3.8`. Maybe
we can just fail if the user asks for a python version that does not exist,
but I am concerned that `rules_python` depending on `protobuf` may pull in
code that requests `3.8`. I'll look at this at some later time.


aignas:
Just realized that we will need `python.toolchain` ignore anything that starts with `3.8` and tell the user that we are ignoring it. TBH, not sure how to proceed here other than ignoring. Warnings may not be great either, because the end user has no way to fix them. Only the root-module invocations of versions that are too old should error out.

arrdem:
> Just realized that we will need python.toolchain ignore anything that starts with 3.8

This doesn't make sense to me. It makes sense to say that we aren't going to manage EoL'd interpreters. I'm less sold on dropping the old requirements configurations deliberately, but deliberately breaking a back revision and creating an upgrade barrier seems like not our job.

aignas:
Just to be clear what I meant about the above is that we depend on protobuf and protobuf is declaring the python 3.8 toolchains and pip.parse with python 3.8 and that can cause issues. I was thinking of a way for this to not break.

Probably the best first step is to leave a single 3.8 version there and remove the earlier ones.

rickeylev:
I don't mind removing our tests for it and basically saying "rules_python stuff may or may not work with a 3.8 runtime"

But yeah, I don't like the idea that e.g. `pip.parse(python_version=3.8)` in a module _somewhere_ will break the whole module graph, especially if nobody is using it.

I think leaving the latest 3.8 version is a decent compromise.

rickeylev (approved):

PR #2708, opened 2025-03-27
fix: support gazelle generation_mode:update_only

### LINKED ISSUE #2707: crash when using gazelle `generation_mode:update_only`
# 🐞 bug report

### Description

The next version of gazelle will include [generation_mode](https://github.com/bazel-contrib/bazel-gazelle/commit/2b9ae2de8c83bc5812ba3587a03ccddd596301da) to allow only updating existing BUILD files and never creating new ones. This means language callbacks such as `Configure` and `GenerateRules` will not be invoked for directories that do not already have a BUILD file.

Today rules_python/gazelle assumes `GenerateRules` is invoked for every single directory and crashes if it was not invoked for a parent directory.

Short term: we should not assume `GenerateRules` was invoked for every directory.

Long term: I think `python_generation_mode: project` can be replace with `generation_mode: update_only`, although there will be minor differences (the user must `touch BUILD`) and it will be a breaking change.

## 🔬 Minimal Reproduction

```
BUILD    <--- put `# gazelle:generation_mode update_only` + enable rules_python/gazelle
libs/
   a/BUILD
```

## 🔥 Exception or Error

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x8 pc=0x100e90d08]

goroutine 1 [running]:
github.com/bazelbuild/rules_python/gazelle/pythonconfig.(*Config).NewChild(...)
	external/com_github_bazelbuild_rules_python_gazelle/pythonconfig/pythonconfig.go:213
github.com/bazelbuild/rules_python/gazelle/python.(*Configurer).Configure(0x10338d9e0, 0x14010b068f0, {0x14010b206d8, 0x14}, 0x14010b3c5a0)
	external/com_github_bazelbuild_rules_python_gazelle/python/configure.go:98 +0x1c8
github.com/bazelbuild/bazel-gazelle/walk.configure({0x140005f7400?, 0x8?, 0x14?}, 0x1400198a9a8?, 0x14010b068f0, {0x14010b206d8, 0x14}, 0x14010b3c5a0)
	external/bazel_gazelle/walk/walk.go:373 +0x78
github.com/bazelbuild/bazel-gazelle/walk.visit(0x14010b068f0, {0x140005f7400, 0x8, 0x8}, 0x1400198a9a8, 0x140016fdc80, 0x1400191ad00, 0x1400198b4b0, {0x14010b206d8, 0x14}, ...)
	external/bazel_gazelle/walk/walk.go:172 +0x3ac
github.com/bazelbuild/bazel-gazelle/walk.visit(0x14010b06820, {0x140005f7400, 0x8, 0x8}, 0x1400198a9a8, 0x140016fdc80, 0x1400186aa80, 0x1400198b4b0, {0x14010b1c868, 0x4}, ...)
	external/bazel_gazelle/walk/walk.go:235 +0xe40
github.com/bazelbuild/bazel-gazelle/walk.visit(0x140013d48f0, {0x140005f7400, 0x8, 0x8}, 0x1400198a9a8, 0x140016fdc80, 0x140016f1880, 0x1400198b4b0, {0x0, 0x0}, ...)
	external/bazel_gazelle/walk/walk.go:235 +0xe40
github.com/bazelbuild/bazel-gazelle/walk.Walk(0x140013d48f0, {0x140005f7400, 0x8, 0x8}, {0x140013e2fc0, 0x1, 0x1}, 0x0, 0x1400198b4b0)
	external/bazel_gazelle/walk/walk.go:128 +0x17c
```

jbedard:
This just fixes a crash when `generation_mode: update_only` causes `GenerateRules` to not be invoked for 100% of directories.

Fix #2707

### DIFF
--- a/gazelle/pythonconfig/pythonconfig.go
+++ b/gazelle/pythonconfig/pythonconfig.go
@@ -22,8 +22,8 @@ import (
 
 	"github.com/emirpasic/gods/lists/singlylinkedlist"
 
-	"github.com/bazelbuild/bazel-gazelle/label"
 	"github.com/bazel-contrib/rules_python/gazelle/manifest"
+	"github.com/bazelbuild/bazel-gazelle/label"
 )
 
 // Directives
@@ -125,21 +125,28 @@ const (
 
 // defaultIgnoreFiles is the list of default values used in the
 // python_ignore_files option.
-var defaultIgnoreFiles = map[string]struct{}{
-}
+var defaultIgnoreFiles = map[string]struct{}{}
 
 // Configs is an extension of map[string]*Config. It provides finding methods
 // on top of the mapping.
 type Configs map[string]*Config
 
 // ParentForPackage returns the parent Config for the given Bazel package.
-func (c *Configs) ParentForPackage(pkg string) *Config {
-	dir := path.Dir(pkg)
-	if dir == "." {
-		dir = ""
+func (c Configs) ParentForPackage(pkg string) *Config {
+	for {
+		dir := path.Dir(pkg)
+		if dir == "." {
+			dir = ""
+		}
+		parent := (map[string]*Config)(c)[dir]
+		if parent != nil {
+			return parent
+		}
+		if dir == "" {
+			return nil
+		}
+		pkg = dir
 	}
-	parent := (map[string]*Config)(*c)[dir]
-	return parent
 }
 
 // Config represents a config extension for a specific Bazel package.
--- a/gazelle/pythonconfig/pythonconfig_test.go
+++ b/gazelle/pythonconfig/pythonconfig_test.go
@@ -248,3 +248,35 @@ func TestFormatThirdPartyDependency(t *testing.T) {
 		})
 	}
 }
+
+func TestConfigsMap(t *testing.T) {
+	t.Run("only root", func(t *testing.T) {
+		configs := Configs{"": New("root/dir", "")}
+
+		if configs.ParentForPackage("") == nil {
+			t.Fatal("expected non-nil for root config")
+		}
+
+		if configs.ParentForPackage("a/b/c") != configs[""] {
+			t.Fatal("expected root for subpackage")
+		}
+	})
+
+	t.Run("sparse child configs", func(t *testing.T) {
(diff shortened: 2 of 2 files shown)


gazelle/pythonconfig/pythonconfig.go
```diff
@@ -134,12 +134,20 @@ type Configs map[string]*Config
 
 // ParentForPackage returns the parent Config for the given Bazel package.
 func (c *Configs) ParentForPackage(pkg string) *Config {
```
aignas, line 136: Since `map` is a reference type, could you please update the receiver arg to be a type of `c Configs` please?

```suggestion
func (c Configs) ParentForPackage(pkg string) *Config {
```

gazelle/pythonconfig/pythonconfig.go
```diff
@@ -134,12 +134,20 @@ type Configs map[string]*Config
...
+			dir = ""
+		}
+		parent := (map[string]*Config)(*c)[dir]
+		if parent != nil {
+			return parent
+		}
+		if dir == "" {
+			return nil
+		}
+		pkg = dir
```
aignas, line 149: Could you add a unit test for this new logic? Just testing the `Configs` in isolation would be sufficient.

Right now it is a little bit hard to read and having tests would help understand what this has to do.

jbedard: I've added a few basic tests for this method.

There should probably be some additional full tests for `update_only` and mixing that with the different `python_generation_mode` modes but I'm not familiar enough with it to do in this initial PR.

gazelle/pythonconfig/pythonconfig.go
```diff
@@ -125,21 +125,28 @@ const (
 
 // defaultIgnoreFiles is the list of default values used in the
 // python_ignore_files option.
-var defaultIgnoreFiles = map[string]struct{}{
-}
+var defaultIgnoreFiles = map[string]struct{}{}
```
jbedard, line 128: The precommit seems to be doing this? 🤷

aignas (approved):
Thank you for the extra tests!

Issue #2709, opened 2025-03-29
f-strings in bootstrap template breaks some environments


armandomontanez:
# 🐞 bug report

### Affected Rule

bootstrap template (py_binary)

### Is this a regression?

Yes, introduced by https://github.com/bazel-contrib/rules_python/pull/2481

### Description

Some f-strings accidentally snuck into the bootstrap template, causing some environments to fail.

## 🔬 Minimal Reproduction

N/A

## 🔥 Exception or Error

<pre><code>
Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
  File "bazel-out/k8-opt-exec-ST-d57f47055a04/bin/external/pigweed+/pw_rpc/py/plugin_pwpb", line 99
    f"{key}={value!r}",
                     ^
SyntaxError: invalid syntax
</code></pre>

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
Linux
  </code>
</pre>

**Output of `bazel version`:**

<pre>
  <code>
8.0.1
  </code>
</pre>

**Rules_python version:**

<pre>
  <code>
1.1.0
  </code>
</pre>

**Anything else relevant?**

aignas:
Thanks for the issue, could you please explain what environment/python version this is causing issues?

The exception message does not contain that information.

armandomontanez:
It's some build bot in the Android ecosystem, it looks like the local Python version is `3.4.3`.

PR #2714, opened 2025-03-31
fix: Fix Python 3.4.x compatibilty with bootstrap (#2709)

armandomontanez:
Fixes some f-strings, trailing commas, and out-of-order argument unpacking in the bootstrap template to restore compatibility with Python 3.4.x.

### DIFF
--- a/python/private/python_bootstrap_template.txt
+++ b/python/private/python_bootstrap_template.txt
@@ -95,19 +95,17 @@ def print_verbose(*args, mapping=None, values=None):
             for key, value in sorted((mapping or {}).items()):
                 print(
                     "bootstrap:",
-                    *args,
-                    f"{key}={value!r}",
+                    *(list(args) + ["{}={}".format(key, repr(value))]),
                     file=sys.stderr,
-                    flush=True,
+                    flush=True
                 )
         elif values is not None:
             for i, v in enumerate(values):
                 print(
                     "bootstrap:",
-                    *args,
-                    f"[{i}] {v!r}",
+                    *(list(args) + ["[{}] {}".format(i, repr(v))]),
                     file=sys.stderr,
-                    flush=True,
+                    flush=True
                 )
         else:
             print("bootstrap:", *args, file=sys.stderr, flush=True)


groodt:
Is there a reason this can't be carried by a patch at the install site? Bazel has first-class support for patching rules. 

We support current Python versions  via https://devguide.python.org/versions/

3.4 is EOL and there are likely to be other things that won't work beyond the bootstrap. It's a maintenance burden too. Much easier to decline fixes that are to support old language features than it is to try wind back code to support something beyond its useful shelf life.

armandomontanez:
I was under the impression [the bootstrap script specifically was intended to work as far back as Python 2.7](https://github.com/bazel-contrib/rules_python/blob/bfad5078acee76f9743e6f1210784288571b0c0c/python/private/python_bootstrap_template.txt#L3-L7).

The project itself is configured to use Python 3.11.

groodt:
We deprecated PY2 a while ago now and probably didn't clean up that comment. Or we should have reframed it as best effort. 

Can you link to the project? Can you include your patch where the rules are installed, or indeed store a bootstrap template in your repo?

armandomontanez:
I can work around locally, I only sent out this PR as a courtesy since it seemed like an unintentional regression.

groodt:
Thanks! Yes, we may indeed end up using it. But one thing is for sure, we probably need to more clearly document supported python versions.

aignas:
Thank you for the contribution. Since we are testing in CI only against supported Python versions, these regressions are somewhat inevitable, I guess. Most of the codebase used to work with Python 2.7 but retaining that compatibility cannot be guaranteed even within the same major semver, because we are not going to have tests for it. I am happy to merge this though and I am happy to accept future contributions to help us maintain that compatibility if it breaks again in the future.

It's like tier1 and tier2 python version system. Tier1 are the versions that we have in CI, Tier2 - all the rest. Maybe we should write that in our docs.

aignas (approved):

TimotheusBachinger:
hey @armandomontanez currently I am also trying to setup this ancient python 3.4 version (yes EOL... dont ask) as a hermetic python toolchain.
I managed to patch https://github.com/astral-sh/python-build-standalone to produce a 3.4 build and I am trying to register this now with `rules_python`.
however I am running over multiple `f-string`/`import typing`/`**kwards` issues which arise from the fact that (i think) the bootstrapping is not python 3.4 compatible.

do you happen to have an example how your project looks like?

PR #2715, opened 2025-03-31
fix(docs): CHANGELOG "astral" typo

lpulley:
It appears to have been copied from the 1.1.0 "Added" section, but I'm not sure whether "fixing" old changelogs is acceptable.


aignas (approved):

lpulley:
I only fixed the one from the upcoming release; if you are interested in fixing both, the other is still there under 1.1.0. Glad this is fixed, though! Thank you

PR #2724, opened 2025-04-01
chore: remove defunct comment about py2 compatibility

rickeylev:
The comment in the bootstrap about requiring compatibility with older Python versions
is defunct and outdated.

Python 2 support was dropped years ago. While compatibility with older Python versions is
best effort for the system_python bootstrap, Python 2 doesn't need to be supported


groodt (approved):
LGTM

PR #2725, opened 2025-04-01
chore: remove semantics.bzl

rickeylev:
semantics.bzl is an artifact of how the rules avoided patching when they were part of Bazel.
With the code moved out of Bazel, such helper files aren't necessary anymore.

Work towards https://github.com/bazel-contrib/rules_python/issues/2522


groodt (approved):
LGTM

PR #2732, opened 2025-04-03
fix(pypi): correctly fallback to pip for git direct URLs

aignas:
Whilst integrating #2695 I introduced a regression and here I add a test
for that and fix it. The code that was getting the filename from the URL
was too eager and would break if there was a git ref as noted in the
test.

Before this commit and #2695 the code was not handling all of the cases
that are tested now either, so I think now we are in a good place. I am
not sure how we should handle the `git_repository` URLs. Maybe having
`http_archive` and `git_repository` usage would be nice, but I am not sure
how we can introduce it at the moment.

Work towards #2363


aignas:
The CI failure is a flake.

rickeylev (approved):

PR #2735, opened 2025-04-04
fix(toolchains): correctly order the toolchains

### LINKED ISSUE #2685: toolchain resolution may be wrong when `python_version` does not include a patch version
Whilst writing code for #2657 I noticed that the toolchain matching is acting weirdly when using `X.Y` format in the `python_version` field. I noticed that when I write `3.12`, the first toolchain gets matched instead of the entry in `MINOR_MAPPING`.

The fix would be to change the ordering in the toolchain repo to first list the versions in the `MINOR_MAPPING` before listing the rest of the toolchains.

aignas:
Since toolchain matching is done by matching the first target that
matches target settings, the `minor_mapping` config setting is special,
because e.g. all `3.11.X` toolchains match the `python_version = "3.11"`
setting.

This just reshuffles the list so that we have toolchains that are in the
`minor_mapping` before the rest.

At the same time remove the workaround from the `lock.bzl` where the bug
was initially discovered.

Fixes #2685


python/private/python.bzl
```diff
@@ -243,10 +243,25 @@ def parse_modules(*, module_ctx, _fail = fail):
...
 
+    # sort the toolchains so that the toolchain versions that are in the
+    # `minor_mapping` are coming first. This ensures that `python_version =
+    # "3.X"` transitions work as expected.
+    minor_version_toolchains = []
+    other_toolchains = []
+    minor_mapping = list(config.minor_mapping.values())
+    for t in toolchains:
+        # FIXME @aignas 2025-04-04: How can we unit test that this ordering is
+        # consistent with what would actually work?
```
rickeylev, line 254: An analysis test could set `--extra_toolchains=@whatever:all`, then do toolchain lookup and verify it got the expected toolchain. I think there's a test that does something like that for py_cc toolchain? I can't remember, exactly. Something like this:

```
def check_version_matching_test(name):
  targets = {}
  for v in versions:
    name = "match_{}".format(v)
    lookup_toolchain(name=name, expect=v, python_version=v)
    targets[name] = name
  analysis_test(
    name = name,
    targets = {
      match_310 = 
    }
  )

def check_version_match_impl(env, targets):
  for t in dir(targets):
    info = getattr(targets, t)[WhateverInfo]
    assert info.expected == info.actual
```

Where `lookup_toolchain` is just a simple rule that applies a transition, and looks up the toolchain and returns info about it.

rickeylev (approved):

python/private/python.bzl
aignas: Ah, everything turned out to be pretty nice. The tests now download the toolchain, but they do not run Python, which probably saves us a few seconds.

PR #2737, opened 2025-04-05
fix: allow warn logging to be disabled via RULES_PYTHON_REPO_DEBUG_VERBOSITY

mattem:
Allows the logging level to be set to `FAIL`, removing `WARN` logging.


aignas (approved):
Thanks!

PR #2741, opened 2025-04-05
fix: escape name segment of stamped wheel files

mattem:
~1 year ago `py_wheel` changed to enforce PEP440, however the resulting stamp vars are no longer escaped.

For example:
Before: `bazel-bin/xxxx/xxxx-_STABLE_BUILD_SCM_VERSION_-py3-none-any.whl`
After: `bazel-bin/xxxx/xxxx-{STABLE_BUILD_SCM_VERSION}-py3-none-any.whl`

This results in downstream consumers of the wheel that are publishing, or generating "manifest" stye files of release artifacts and other metadata will now incorrectly replace the `{STABLE_BUILD_SCM_VERSION}` in the artifact name and path, leading to the incorrect artifact metadata and pathing.

Original reported here, however this could be seen as breaking, as it changes the unstamped wheel file names.
https://bazelbuild.slack.com/archives/C0174UDLAF3/p1706726049830819

### DIFF
--- a/python/private/py_wheel.bzl
+++ b/python/private/py_wheel.bzl
@@ -310,7 +310,7 @@ def _py_wheel_impl(ctx):
 
     filename_segments = [
         _escape_filename_distribution_name(ctx.attr.distribution),
-        normalize_pep440(version),
+        _escape_filename_segment(normalize_pep440(version)),
         _escape_filename_segment(python_tag),
         _escape_filename_segment(abi),
         _escape_filename_segment(ctx.attr.platform),
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,6 +90,7 @@ Unreleased changes template.
   transitions transitioning on the `python_version` flag.
   Fixes [#2685](https://github.com/bazel-contrib/rules_python/issues/2685).
 * (toolchains) Run the check on the Python interpreter in isolated mode, to ensure it's not affected by userland environment variables, such as `PYTHONPATH`.
+* (py_wheel) Ensure the filename segment is escaped in when using stamping wheel names.
 
 {#v0-0-0-added}
 ### Added


examples/wheel/wheel_test.py
```diff
@@ -440,7 +440,7 @@ def test_rule_creates_directory_and_is_included_in_wheel(self):
 
     def test_rule_expands_workspace_status_keys_in_wheel_metadata(self):
         filename = self._get_path(
-            "example_minimal_library{BUILD_USER}-0.1.{BUILD_TIMESTAMP}-py3-none-any.whl"
+            "example_minimal_library{BUILD_USER}-0.1._BUILD_TIMESTAMP_-py3-none-any.whl"
```
mattem, line 443: The distribution likely needs escaping too.

python/private/py_wheel.bzl
```diff
@@ -304,13 +304,15 @@ def _input_file_to_arg(input_file):
...
+    is_stamping = is_stamping_enabled(ctx.attr)
+
     abi = _replace_make_variables(ctx.attr.abi, ctx)
     python_tag = _replace_make_variables(ctx.attr.python_tag, ctx)
     version = _replace_make_variables(ctx.attr.version, ctx)
 
     filename_segments = [
         _escape_filename_distribution_name(ctx.attr.distribution),
-        normalize_pep440(version),
+        _escape_filename_segment(normalize_pep440(version)) if is_stamping else normalize_pep440(version),
```
rickeylev, line 315: Something strange is going on, or I'm missing something.

The net code change is the same as before: `normalize(...) if ... else normalize(...)`

The test was updated, though? Which would mean it's working?? Is the test not covering something...?

rickeylev (changes requested):

PR #2742, opened 2025-04-05
feat(toolchains): create toolchains from locally installed python

rickeylev:
This adds docs and public APIs for using a locally installed python for a toolchain.

Work towards https://github.com/bazel-contrib/rules_python/issues/2070


docs/toolchains.md
```diff
@@ -327,7 +327,86 @@ After registration, your Python targets will use the toolchain's interpreter dur
...
+
+The below is an example that will use `python3` from PATH to find the
+interpreter, then introspect its installation to generate a full toolchain.
+
+```
+# File: MODULE.bazel
+
+local_runtime_repo = use_repo_rule(
+    "@rules_python//python/local_toolchains:repos.bzl",
+    "local_runtime_repo"
```
aignas, line 354: Should we ask users to set that as `dev_dependency`?
```suggestion
    "local_runtime_repo",
    dev_dependency = True,
```

Local toolchains are only valid for root modules anyway.

docs/toolchains.md
```diff
@@ -327,7 +327,86 @@ After registration, your Python targets will use the toolchain's interpreter dur
...
+it. To do this, three pieces need to be wired together:
+
+1. Specify a path or command to a Python interpreter (multiple can be defined).
+2. Create toolchains for the runtimes in (1)
+3. Register the toolchains created by (2)
+
+The below is an example that will use `python3` from PATH to find the
+interpreter, then introspect its installation to generate a full toolchain.
+
+```
```
aignas, line 349: nit:
```suggestion
```starlark
```

docs/toolchains.md
```diff
@@ -327,7 +327,86 @@ After registration, your Python targets will use the toolchain's interpreter dur
...
+
+# Step 3: Register the toolchains
+register_toolchains("@local_toolchains//:all")
+```
+
+Note that `register_toolchains` will insert the local toolchain earlier in the
+toolchain ordering, so it will take precedence over other registered toolchains.
+
+Multiple runtimes and/or toolchains can be defined, which allows for multiple
+Python versions and/or platforms to be configured in a single MODULE.bazel.
```
aignas, line 383:
```suggestion
Python versions and/or platforms to be configured in a single `MODULE.bazel`.
```

docs/toolchains.md
```diff
@@ -327,7 +327,86 @@ After registration, your Python targets will use the toolchain's interpreter dur
...
+`/usr/bin/env python3` to run a program. This makes it easy to change what
+Python is used to run a program, but also makes it easy to use a Python version
+that isn't compatible with build-time assumptions.
+
+```
+register_toolchains("@rules_python//python/runtime_env_toolchains:all")
+```
+
+Note that this toolchain has no constraints, i.e. it will match any platform,
+Python version, etc.
```
aignas, line 402: Should we add `@platforms//host` to the constraints or suggest users adding that.

aignas (commented):
Some questions about constraints.

docs/toolchains.md
rickeylev: For runtime_env toolchain, no. This is because it's explicit purpose is "use whatever python the runtime environment has". Which would generally be a headache, but is handy when you control the deployment environment (e.g. docker image), or are otherwise OK with the the Python interpreter being picked later.

For the local toolchain, yes, that makes sense. I'll prep a separate PR to do that.

rickeylev: Ah, actually it does this already, though in a different way.

During the repo phase, it tries to run Python. If it fails, it generates a `@platforms//:incompatible` config setting. Otherwise, it generates a `@platforms//os:{rctx os name}` config setting.

I dunno why I did that instead of `@platforms:host`. I think the net effect is about the same? I'm not too familiar with how platforms:host behaves.

In any case, resolving since this is effectively done.

rickeylev:
PTAL

aignas (approved):

PR #2743, opened 2025-04-05
fix: Ensure temporary .pyc & .pyo files are excluded from the interpreters repository files

mattem:
We've seen cases the temporary versions for the `.pyc` and `.pyo` files are unstable on certain interpreter toolchains. The temp files take for form of `.pyc.NNN`, so the amended glob patten will still match both the `.pyc` and `.pyc.NNN` versions of the file names.


rickeylev (approved):
Ah geeze. I swear I had added pyc.NNN ignore stuff to these, too, but I guess not. Thanks!

PR #2744, opened 2025-04-05
chore: remove unnecessary DEFAULT_BOOTSTRAP_TEMPLATE global

rickeylev:
I think the DEFAULT_BOOTSTRAP_TEMPLATE global was used by something in the original
Bazel impl, but now it's just used in one place. Remove the shared global and just
inline the single usage.


groodt (approved):
LGTM

PR #2730, opened 2025-04-02
fix(pypi): mark the extension reproducible

aignas:
This will remove the merge conflicts and improve the usability when the
`MODULE.bazel.lock` is used together with `rules_python`. This means
that the lock file will not be used to read the `URL` and `sha256`
values for the Python sources when the `experimental_index_url` is used,
but the idea is that that information will be kept in repo cache.

Fixes #2434

Created #2731 to leverage the bazel feature to write immutable facts to
the lock file once it becomes available.

PR #2745, opened 2025-04-06
feat: add support for passing extra pip args by platform

mattem:
Supports passing extra arguments to pip based on the host platform. This is useful when arguments such as `--only-binary` are platform dependant.


.bazelrc
```diff
@@ -4,8 +4,8 @@
...
...roto_library,examples/bzlmod/py_proto_library/example.com/another_proto,examples/bzlmod/py_proto_library/example.com/proto,examples/bzlmod/runfiles,examples/bzlmod/tests,examples/bzlmod/tests/other_module,examples/bzlmod/whl_mods,examples/multi_python_versions/libs/my_lib,examples/multi_python_versions/requirements,examples/multi_python_versions/tests,examples/pip_parse,examples/pip_parse_vendored,examples/pip_repository_annotations,examples/py_proto_library,examples/py_proto_library/example.com/another_proto,examples/py_proto_library/example.com/proto,gazelle,gazelle/manifest,gazelle/manifest/generate,gazelle/manifest/hasher,gazelle/manifest/test,gazelle/modules_mapping,gazelle/python,gazelle/pythonconfig,gazelle/python/private,tests/integration/compile_pip_requirements,tests/integration/compile_pip_requirements_test_from_external_repo,tests/integration/custom_commands,tests/integration/ignore_root_user_error,tests/integration/ignore_root_user_error/submodule,tests/integration/local_toolchains,tests/integration/pip_parse,tests/integration/pip_parse/empty,tests/integration/py_cc_toolchain_registered,tests/modules/other,tests/modules/other/nspkg_delta,tests/modules/other/nspkg_gamma
```
mattem, line 7: The pre-commit hook updated this 🤷

python/private/pypi/pip_repository_attrs.bzl
```diff
@@ -21,6 +21,21 @@ repositories."""
...
+Extra arguments to pass on to pip on the assoicated platform specifiers. Args must not contain spaces.
+
+The keys in the dict take the form `<os>_<cpu>`, wildcards in the form `<os>_*` are supported.
+Key values are as those returned in Bazels `repository_os` struct, with spaces removed.
+
+Supports environment variables using the syntax `$VARNAME` or
+`${VARNAME}` (expanding to empty string if unset) or
+`${VARNAME:-default}` (expanding to default if the variable is unset
+or empty in the environment), if `"VARNAME"` is listed in the
+`envsubst` attribute. See also `envsubst`.
```
aignas, line 36: Please add a note when this has been added and we can link to the attributes in our docs.

```suggestion
{attr}`envsubst` attribute. See also {attr}`envsubst`.

:::{versionadded} VERSION_NEXT_FEATURE
:::
```

python/private/pypi/pip_repository_attrs.bzl
```diff
@@ -21,6 +21,21 @@ repositories."""
 load(":attrs.bzl", COMMON_ATTRS = "ATTRS")
 
 ATTRS = {
+    "extra_pip_args_by_platform": attr.string_list_dict(
+        doc = """Similar to `extra_pip_args`.
+
+Extra arguments to pass on to pip on the assoicated platform specifiers. Args must not contain spaces.
```
aignas, line 27: Could you please update https://github.com/bazel-contrib/rules_python/blob/main/docs/pypi-dependencies.md?plain=1#L320 section on how this feature could be used?

tests/pypi/extension/extension_tests.bzl
```diff
@@ -182,6 +185,46 @@ def _test_simple(env):
...
+                        extra_pip_args_by_platform = {
+                            "macosx_*": ["--only-binary=simple"],
+                        },
+                    ),
+                ],
+            ),
+            os_name = "mac os x",
+            os_arch = "arch64",
+            read = lambda x: {
+                "darwin.txt": "simple==0.0.2 --hash=sha256:deadb00f",
```
aignas, line 208: Could you please add extra files so that we can see that there are differences on each platform?

python/private/pypi/pip_args.bzl
```diff
@@ -0,0 +1,65 @@
...
+    Args:
+        ctx: Current repository context.
+        extra_pip_args: (string keyed string list dict): Extra pip arguments.
+
+    Returns:
+        List of arguments for the current platform.
+    """
+
+    os = ctx.os.name.replace(" ", "")
+    arch = ctx.os.arch.replace(" ", "")
```
aignas, line 29:
```suggestion
    os = ctx.os.name.replace(" ", "")
    arch = ctx.os.arch.replace(" ", "")
```

This is duplicating code from `repo_utils`. We already have code for getting `os` and `arch` labels that the rest of the code understands. See `host_platform` function in `parse_requirements.bzl` (if I remember correctly).

python/private/pypi/pip_repository.bzl
```diff
@@ -71,6 +72,8 @@ exports_files(["requirements.bzl"])
 """
 
 def _pip_repository_impl(rctx):
+    extra_pip_args = resolve_extra_pip_args(rctx, rctx.attr.extra_pip_args, rctx.attr.extra_pip_args_by_platform)
```
aignas, line 75: Do I understand correctly that this will resolve the pip args depending on the host platform? What about resolving based on the target platform? I'd rather pass the `rctx.attr.extra_pip_args` and `rctx.attr.extra_pip_args_by_platform` to the `requirements_files_by_platform` and resolve the pip args by target platform.

The current API is somewhat not great at doing the overrides per platform because we need to have these dictionaries for the overrides and if we add this, then we'll have to maintain this until v2.

aignas (changes requested):
I understand that the desire is to have per platform overrides, where we only care about the host platform.

If #2730 was not merged, the PR as is could not be merged because this creates different behaviour on different host platforms.

However, I would be much more keen to work with the target platform rather than the host platform. What is more, the direction where we are going with #260 does not align well with what is being introduced here if I understand correctly.

python/private/pypi/pip_repository.bzl
mattem: Can you get the target platform in this context?

The way `requirements_files_by_platform` seems to work is to inject the flags into the requirement lines, and pass the platform flag to pip, does that work with other flags too?

The use case we have here is that we need to force binary only fetching on certain platforms, where as on others we can't do that and allow the sdist to be built. There are some other flags we pass only on Linux too, but I don't recall them, can check tomorrow.

aignas: I created #2747 to create a better API for this, but in the mean time this is how I can see things fit:

* `requirements_by_platform` parses the files and populates the initial data structure with the `extra_pip_args` that are common for all platforms. By this time we have a list of `requirements` structs that have: the requirement line that is needed, extra_pip_args for that requirement line and the target platform.
* (this is where this PR could plug in) - for each target platform listed by the user we add the extra_pip_args based on the already defined target platform.
* If we see that there is a single `requirement` struct for target platforms `foo` and `bar` and we need to add 2 different sets of pip args, then we split the struct into two - one for `foo`, one for `bar` and we add the pip args.
- then later in the code there is code that actually creates the `whl_library` instances and wires everything together through the hub repo.

aignas:
Right now the per-platform pip args are supported if they are included in the requirements files.

Since this has diverged by quite a bit, I'll close it, but the PR to improve this is welcome.

PR #2746, opened 2025-04-06
perf: lazily load gazelle manifest files

mattem:
In large repositories where Python may not be the only language, the gazelle manifest loading is done unnecessarily, and is done during the configuration walk. 

This means that even for non-python gazelle invocations (eg `bazel run gazelle -- web/`), Python manifest files are being parsed and loaded into memory.
This issue compounds if the repository uses multiple dependency closures, ie multiple `gazelle_python.yaml` files.
In our repo, we currently have ~250 Python manifests, so loading them when Gazelle is only running over other languages is time consuming.


mattem:
The CI failure here seems unrelated, it's failed to clone the repo on the agent.

dougthor42 (commented):
I don't see any issues with this. I'd like to test it on our monorepo before approving, but I predict it'll be just fine.

For your case, what sort of speedups are you seeing? And how did you profile it (so that I can try to reproduce it)?

> we currently have ~250 Python manifests,

Do you mean ~250 `gazelle_python.yaml` files? Or 250 packages listed in a single `gazelle_python.yaml`?

mattem:
> I'd like to test it on our monorepo before approving, but I predict it'll be just fine.

Sure, this has been running internally for us for ~2 months now, but we are still on the (now very diverged) original copy of the extension. Let me know how it goes.

> For your case, what sort of speedups are you seeing? And how did you profile it (so that I can try to reproduce it)?

It's slightly tricky to say, as it depends on the directory that Gazelle is running over, but we saw a reduction of about 30 seconds when running over only web targets, where no Python manifests are required.
Gazelle can be profiled by via the `-cpuprofile` and `-memprofile` flags. The CPU profile is written to the given path, and can viewed via pprof (`go tool pprof ...`).
The other part I can upstream is the removal of `filepath.Walkdir` here, https://github.com/bazel-contrib/rules_python/blob/main/gazelle/python/generate.go#L154. There's no need to re-walk the file system, Gazelle has already collected up the files on the configuration pass. This is very noticeable on the profiles, given the walk does both file system walking and stats on files. Removing this took about 70 seconds off running Gazelle over our whole repo.

> Do you mean ~250 gazelle_python.yaml files? Or 250 packages listed in a single gazelle_python.yaml?

250 individual `gazelle_python.yaml` files. 250 packages in a single file would be on the smaller side of the majority of manifest files.

CHANGELOG.md
```diff
@@ -76,6 +76,7 @@ Unreleased changes template.
 * (pypi) The PyPI extension will no longer write the lock file entries as the
   extension has been marked reproducible.
   Fixes [#2434](https://github.com/bazel-contrib/rules_python/issues/2434).
+* (gazelle) Lazily load and parse manifest files when running Gazelle. This ensures no manifest files are loaded when Gazelle is run over a set of non-python directories.
```
dougthor42, line 79: Nit: Please add a link to the issue (if any) or to this PR for easy future reference.

dougthor42 (approved):
> 250 individual `gazelle_python.yaml` files

Holy crap! I'd be interested to learn why you have all those. But anyway, I finally got around to testing things. It all looks good on my end.

Approved, ideally with the nit resolved. :tada:

PR #2748, opened 2025-04-07
fix: ensure filenames with spaces are excluded from targets

mattem:
Some dependencies contain files with spaces in the name. These should be excluded as they are generally unsupported, and when placed in a runfiles manifest file, they cause it to be malformed.

This changes omits files with spaces in the names from glob patterns.
It also changes the `.pyo.NNN` temp file inclusion added in https://github.com/bazel-contrib/rules_python/pull/2743 as it seems it was slightly misplaced, and missed form 3p dependency targets.

### DIFF
--- a/python/private/hermetic_runtime_repo_setup.bzl
+++ b/python/private/hermetic_runtime_repo_setup.bzl
@@ -76,8 +76,11 @@ def define_hermetic_runtime_toolchain_impl(
                 # tests for the standard libraries.
                 "lib/python{major}.{minor}*/**/test/**".format(**version_dict),
                 "lib/python{major}.{minor}*/**/tests/**".format(**version_dict),
-                # During pyc creation, temp files named *.pyc.NNN are created
+                # During pyc and pyo creation, temp files named *.pyc.NNN and *.pyo.NNN are created.
                 "**/__pycache__/*.pyc.*",
+                "**/__pycache__/*.pyo.*",
+                # File names with spaces should also be ignored.
+                "**/* *",
             ] + glob_excludes.version_dependent_exclusions() + extra_files_glob_exclude,
         ),
     )
--- a/python/private/pypi/whl_library_targets.bzl
+++ b/python/private/pypi/whl_library_targets.bzl
@@ -103,7 +103,14 @@ def whl_library_targets(
     for filegroup_name, glob in filegroups.items():
         native.filegroup(
             name = filegroup_name,
-            srcs = native.glob(glob, allow_empty = True),
+            srcs = native.glob(
+                glob,
+                exclude = [
+                    # File names with spaces should be excluded.
+                    "**/* *",
+                ],
+                allow_empty = True,
+            ),
             visibility = ["//visibility:public"],
         )
 
@@ -229,10 +236,13 @@ def whl_library_targets(
             "**/*.py",
             "**/*.pyc",
             "**/*.pyc.*",  # During pyc creation, temp files named *.pyc.NNNN are created
+            "**/*.pyo.*",  # During pyo creation, temp files named *.pyo.NNNN are created
             # RECORD is known to contain sha256 checksums of files which might include the checksums
             # of generated files produced when wheels are installed. The file is ignored to avoid
             # Bazel caching issues.
             "**/*.dist-info/RECORD",
+            # File names with spaces should be excluded.
+            "**/* *",
         ] + glob_excludes.version_dependent_exclusions()
         for item in data_exclude:
             if item not in _data_exclude:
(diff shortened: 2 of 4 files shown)


python/private/hermetic_runtime_repo_setup.bzl
```diff
@@ -76,8 +76,11 @@ def define_hermetic_runtime_toolchain_impl(
                 # tests for the standard libraries.
                 "lib/python{major}.{minor}*/**/test/**".format(**version_dict),
                 "lib/python{major}.{minor}*/**/tests/**".format(**version_dict),
-                # During pyc creation, temp files named *.pyc.NNN are created
+                # During pyc and pyo creation, temp files named *.pyc.NNN and *.pyo.NNN are created.
                 "**/__pycache__/*.pyc.*",
+                "**/__pycache__/*.pyo.*",
+                # File names with spaces should also be ignored.
+                "**/* *",
             ] + glob_excludes.version_dependent_exclusions() + extra_files_glob_exclude,
```
aignas, line 84: I am thinking that it would be nice to have `glob_excludes.pyc_files()` and `glob_excludes.pyo_files()` and `glob_excludes.files_with_spaces()`. Then we can ensure that the explanation for why we need to do what we need to do can be next to their definitions.

I would also love to exclude `.pyc` and `.pyc.*` is the hermetic toolchain definition, so that the exclude is the same regardless if we are `chmod`ing the dir to be read-only or not.

What do you think?

python/private/pypi/whl_library_targets.bzl
```diff
@@ -103,7 +103,14 @@ def whl_library_targets(
     for filegroup_name, glob in filegroups.items():
         native.filegroup(
             name = filegroup_name,
-            srcs = native.glob(glob, allow_empty = True),
+            srcs = native.glob(
+                glob,
+                exclude = [
+                    # File names with spaces should be excluded.
+                    "**/* *",
```
aignas, line 110: I thought that our supported `bazel` versions support files with spaces, so why do we need to exclude them?

aignas: However, it seems that someone tried it and it did not work?

https://github.com/michael-christen/toolbox/pull/184/files

mattem: Yeah we also hit the issue with setuptools in runfiles, but with the Go runfiles library. Setuptools seems to contain files with spaces, so even if bazel itself can handle then now, the runfiles libraries can't.

python/private/hermetic_runtime_repo_setup.bzl
mattem: Yeah can add those methods.

Re: pyc, I think we'd only want the temp files excluded here? I'd originally excluded then in a different PR in a different part of the code (removed in this PR in favor of here). This change is keeping the pyc excluded in a single place.

If the pyc files are stable, then generally it would be preferable to keep them, no?

aignas: Hmmm. Yeah, if they are stable it is fine and we are already setting the vars to make them stable, so SGTM.

rickeylev: If memory serves, excluding pyc was what finally got rid of the Windows jobs getting "can't delete open file" errors. My theory was two processes both went to import at a module without a pyc. Both would start the pyc process, but one would manage to finish writing and open the pyc, then the other process would try to overwrite it. But it couldn't, because the file was open.

The secondary issue is, as pycs are created, they show as additional files added to the target, thus invalidating it, which means anything downstream has to re-run. Eventually things will settle, but they'll only stay settled as long as the repo sticks around. A similar issue can happen with the timestamps: two processes might race and end up creating slightly different timestamped pycs, thus making it look like the file changed.

aignas: This is only true for the `pyc` generation happening at `repository_rule` execution time. I have added `-B` a while ago.

When the packages are used in the regular `py_binary` and `py_test` rules I expect the `pyc` files to be created in the sandbox and not the `repository_rule` output dirs, but my claim should be checked.

rickeylev: Oh good. Yeah, that should prevent that issue, then. SGTM.

python/private/hermetic_runtime_repo_setup.bzl
```diff
@@ -76,8 +76,11 @@ def define_hermetic_runtime_toolchain_impl(
                 # tests for the standard libraries.
                 "lib/python{major}.{minor}*/**/test/**".format(**version_dict),
                 "lib/python{major}.{minor}*/**/tests/**".format(**version_dict),
-                # During pyc creation, temp files named *.pyc.NNN are created
+                # During pyc and pyo creation, temp files named *.pyc.NNN and *.pyo.NNN are created.
                 "**/__pycache__/*.pyc.*",
+                "**/__pycache__/*.pyo.*",
+                # File names with spaces should also be ignored.
```
rickeylev, line 82:
```suggestion
                # Ignore files with spaces because, while Bazel supports them,
                # the runfiles manifest format doesn't yet
```

rickeylev (approved):
Mostly LGTM. The only comment I have is to replace the "files with spaces should be ignored" with text that tells why they should be ignored. I suggested an edit to that effect in one spot. I'm OK with copy/pasting that same comment, or factor out a common function with the comment there instead

PR #2761, opened 2025-04-09
fix: run python version call in isolated mode

mattem:
Similar to https://github.com/bazel-contrib/rules_python/pull/2738, runs the call to get the Python interpreter version in isolated mode via `-I`, ensuring userland Python variables do not affect this call.


aignas (approved):
Thanks!

PR #2766, opened 2025-04-10
fix: Resolve incorrect platform specific dependency

ewianda:
This change addresses a bug where `pip.parse` selects the wrong requirement entry when multiple extras are listed with platform-specific markers.

#### 🔍 Problem:
In a `requirements.txt` generated by tools like `uv` or `poetry`, it's valid to have multiple entries for the same package, each with different extras and `sys_platform` markers, for example:

```ini
optimum[onnxruntime]==1.17.1 ; sys_platform == 'darwin'
optimum[onnxruntime-gpu]==1.17.1 ; sys_platform == 'linux'
```

The current implementation in [`[parse_requirements.bzl](https://github.com/bazel-contrib/rules_python/blob/032f6aa738a673b13b605dabf55465c6fc1a56eb/python/private/pypi/parse_requirements.bzl#L114-L126)`](https://github.com/bazel-contrib/rules_python/blob/032f6aa738a673b13b605dabf55465c6fc1a56eb/python/private/pypi/parse_requirements.bzl#L114-L126) uses a sort-by-length heuristic to select the “best” requirement when there are multiple entries with the same base name. This works well in legacy `requirements.txt` files where:
```
my_dep
my_dep[foo]
my_dep[foo,bar]
```
...would indicate an intent to select the **most specific subset of extras** (i.e. the longest name).

However, this heuristic **breaks** in the presence of **platform markers**, where extras are **not subsets**, but distinct variants. In the example above, Bazel mistakenly selects `optimum[onnxruntime-gpu]` on macOS because it's a longer match, even though it is guarded by a Linux-only marker.

#### ✅ Fix:
This PR modifies the behavior to:
1. **Add the requirement marker** as part of the sorting key.
2. **Then apply the longest-match logic** to drop duplicate requirements with different extras but the same markers.

This ensures that only applicable requirements are considered during resolution, preserving correctness in multi-platform environments.

#### 🧪 Before:
On macOS, the following entry is incorrectly selected:
```
optimum[onnxruntime-gpu]==1.17.1 ; sys_platform == 'linux'
```

#### ✅ After:
Correct entry is selected:
```
optimum[onnxruntime]==1.17.1 ; sys_platform == 'darwin'
```

close https://github.com/bazel-contrib/rules_python/issues/2690

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -81,8 +81,8 @@ Unreleased changes template.
 
 {#v0-0-0-fixed}
 ### Fixed
-* (pypi) Platform specific markers are now correctly handled when using `requirments_by_platform`
-  in pip.parse Fixed in [#2766](https://github.com/bazel-contrib/rules_python/pull/2766).
+* (pypi) Platform specific extras are now correctly handled when using
+  universal lock files with environment markers. Fixes [#2690](https://github.com/bazel-contrib/rules_python/pull/2690).
 * (runfiles) ({obj}`--bootstrap_impl=script`) Follow symlinks when searching for runfiles.
 * (toolchains) Do not try to run `chmod` when downloading non-windows hermetic toolchain
   repositories on Windows. Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -81,6 +81,8 @@ Unreleased changes template.
 
 {#v0-0-0-fixed}
 ### Fixed
+* (pypi) Platform specific markers are now correctly handled when using `requirments_by_platform`
+  in pip.parse Fixed in [#2766](https://github.com/bazel-contrib/rules_python/pull/2766).
 * (runfiles) ({obj}`--bootstrap_impl=script`) Follow symlinks when searching for runfiles.
 * (toolchains) Do not try to run `chmod` when downloading non-windows hermetic toolchain
   repositories on Windows. Fixes
--- a/python/private/pypi/parse_requirements.bzl
+++ b/python/private/pypi/parse_requirements.bzl
@@ -30,6 +30,7 @@ load("//python/private:normalize_name.bzl", "normalize_name")
 load("//python/private:repo_utils.bzl", "repo_utils")
 load(":index_sources.bzl", "index_sources")
 load(":parse_requirements_txt.bzl", "parse_requirements_txt")
+load(":pep508_req.bzl", "requirement")
 load(":whl_target_platforms.bzl", "select_whls")
 
 def _extract_version(entry):
@@ -111,8 +112,9 @@ def parse_requirements(
         # The requirement lines might have duplicate names because lines for extras
         # are returned as just the base package name. e.g., `foo[bar]` results
         # in an entry like `("foo", "foo[bar] == 1.0 ...")`.
+        # Lines with different markers are not condidered duplicates.
         requirements_dict = {
-            (normalize_name(entry[0]), _extract_version(entry[1])): entry
+            (normalize_name(entry[0]), _extract_version(entry[1]), requirement(entry[1]).marker): entry
             for entry in sorted(
                 parse_result.requirements,
                 # Get the longest match and fallback to original WORKSPACE sorting,
--- a/python/private/pypi/parse_requirements.bzl
+++ b/python/private/pypi/parse_requirements.bzl
@@ -30,22 +30,8 @@ load("//python/private:normalize_name.bzl", "normalize_name")
 load("//python/private:repo_utils.bzl", "repo_utils")
 load(":index_sources.bzl", "index_sources")
 load(":parse_requirements_txt.bzl", "parse_requirements_txt")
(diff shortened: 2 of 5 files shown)


python/private/pypi/pep508_env.bzl
```diff
@@ -30,7 +30,7 @@ _platform_system_values = {
     "windows": "Windows",
 }
 _sys_platform_values = {
-    "linux": "posix",
+    "linux": "linux",
```
ewianda, line 33: This change was required for `sys_platfomr==linux` otherwise all the different platforms are added to the list

aignas: Yeah, this change is correct, thanks for fixing this. I noticed that too in one of my PRs.

python/private/pypi/parse_requirements.bzl
```diff
@@ -46,6 +46,19 @@ def _extract_version(entry):
...
+
+    Args:
+        entry: {type}`str` The requirement string.
+    """
+    marker_start = entry.find(";")
+    if marker_start != -1:
+        # Extract everything after ';' until --hashes if it exists or end of the string
+        marker, _, _ = entry[marker_start + 1:].partition(" --hashes")
+        return marker.strip()
+    return ""
```
aignas, line 60: I have a `pep508_requirement.bzl` file that is very good for parsing these sort of things. We should probably use it?

aignas (commented):
Could you please add a CHANGELOG item for future reference?

ewianda:
@aignas is this good to, merge. It is currently a blocker for us

CHANGELOG.md
```diff
@@ -81,6 +81,8 @@ Unreleased changes template.
 
 {#v0-0-0-fixed}
 ### Fixed
+* (pypi) Platform specific markers are now correctly handled when using `requirments_by_platform`
+  in pip.parse Fixed in [#2766](https://github.com/bazel-contrib/rules_python/pull/2766).
```
aignas, line 85:
```suggestion
* (pypi) Platform specific extras are now correctly handled when using
  universal lock files with environment markers. Fixes [#2690](https://github.com/bazel-contrib/rules_python/pull/2690).
```

The breakage of the `env` markers has never been released so referencing the original issue would be more correct.

python/private/pypi/parse_requirements.bzl
```diff
@@ -111,8 +112,9 @@ def parse_requirements(
         # The requirement lines might have duplicate names because lines for extras
         # are returned as just the base package name. e.g., `foo[bar]` results
         # in an entry like `("foo", "foo[bar] == 1.0 ...")`.
+        # Lines with different markers are not condidered duplicates.
         requirements_dict = {
-            (normalize_name(entry[0]), _extract_version(entry[1])): entry
+            (normalize_name(entry[0]), _extract_version(entry[1]), requirement(entry[1]).marker): entry
```
aignas, line 117: The same `requirement` function can return you the `version` and the `name` if I remember correctly. We could derive the key for grouping from that instead.

python/private/pypi/pep508_env.bzl
aignas: #2767 fixed this, could you please rebase?

aignas (commented):
Could you please also update the PR description to be more descriptive please? The [PR template](https://github.com/bazel-contrib/rules_python/blob/main/.github/PULL_REQUEST_TEMPLATE.md) has good suggestions on how to structure it (i.e. the user visible change in behaviour should be explained).

aignas (approved):

aignas:
Thank you!

ewianda:
@aignas I am getting 

```
ERROR: /home/ewianda/.cache/bazel/_bazel_ewianda/da4b4cc49e0e621570c9e24d6f1eab95/external/rules_python~/python/private/pypi/extension.bzl:231:25: Traceback (most recent call last):
        File "/home/ewianda/.cache/bazel/_bazel_ewianda/da4b4cc49e0e621570c9e24d6f1eab95/external/rules_python~/python/private/pypi/extension.bzl", line 611, column 25, in _pip_impl
                mods = parse_modules(module_ctx)
        File "/home/ewianda/.cache/bazel/_bazel_ewianda/da4b4cc49e0e621570c9e24d6f1eab95/external/rules_python~/python/private/pypi/extension.bzl", line 487, column 36, in parse_modules
                out = _create_whl_repos(
        File "/home/ewianda/.cache/bazel/_bazel_ewianda/da4b4cc49e0e621570c9e24d6f1eab95/external/rules_python~/python/private/pypi/extension.bzl", line 231, column 25, in _create_whl_repos
                fail("Attempting to creating a duplicate library {} for {}".format(
Error in fail: Attempting to creating a duplicate library py_deps_311_optimum_py3_none_any_508bc55d for optimum

```

Where should I be  looking to fix this

ewianda:
NVM, it is coming from `uv` https://github.com/astral-sh/uv/issues/8915#issuecomment-2470873623

PR #2767, opened 2025-04-10
fix(pypi): fixes to the marker evaluation and utils

aignas:
These are just bugfixes to already merged code:
* Fix nested bracket parsing in PEP508 marker parser.
* Fix the sys_platform constants, which I noticed in #2629 but they got
  also pointed out in #2766.
* Port some of python tests for requirement parsing and improve the
  implementation. Those tests will be removed in #2629.
* Move the platform related code to a separate file.
* Rename `pep508_req.bzl` to `pep508_requirement.bzl` to follow the
  convention.

All of the bug fixes have added tests.

Work towards #2423.

### DIFF
--- a/python/private/pypi/pep508_platform.bzl
+++ b/python/private/pypi/pep508_platform.bzl
@@ -1,0 +1,57 @@
+# Copyright 2025 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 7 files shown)


python/private/pypi/pep508_platform.bzl
```diff
@@ -0,0 +1,57 @@
+# Copyright 2025 The Bazel Authors. All rights reserved.
```
rickeylev, line 1: FYI, going forward, we can omit these copyright headers

rickeylev (approved):

python/private/pypi/pep508_platform.bzl
aignas: Should we do with the old ones?

PR #2768, opened 2025-04-10
fix(rules): make the srcs trully optional

### LINKED ISSUE #2765: Support src-less invocation of py_binary
With the recent addition of the #2671, I find that in most cases I can replace the `py_console_script_binary` with just a `py_binary`, e.g:

```starlark
py_console_script_binary(
    name = "foo",
    deps = [...],
    pkg = "foo",
    script = "bar",
)
```

versus

```starlark
py_binary(
    name = "foo",
    deps = [...], # assuming "foo" is in the deps
    main_module = "bar"
)
```

However I cannot do this, because the `srcs` cannot be empty here.

An alternative to this could be to use the `python_interpreter` introduced in #2507 
```starlark
interpreter_binary(
    name = "foo",
    deps = [...],
    args = ["-m", "foo"],
)
```

aignas:
With this PR we mark the srcs attribute as optional as we can
leverage the `main_module` to just run things from the deps.

This also removes a long-standing `TODO` note.

Fixes #2765

### DIFF
--- a/python/private/py_executable.bzl
+++ b/python/private/py_executable.bzl
@@ -808,7 +808,7 @@ def _create_stage1_bootstrap(
         subs["%imports%"] = ":".join(imports.to_list())
         subs["%main%"] = "{}/{}".format(ctx.workspace_name, main_py.short_path)
     else:
-        fail("'main' or 'srcs' must be specified")
+        fail("mandatory 'srcs' attribute has not been specified")
 
     ctx.actions.expand_template(
         template = template,
--- a/python/private/py_executable.bzl
+++ b/python/private/py_executable.bzl
@@ -786,7 +786,7 @@ def _create_stage1_bootstrap(
         )
         template = runtime.bootstrap_template
         subs["%shebang%"] = runtime.stub_shebang
-    else:
+    elif main_py:
         if (ctx.configuration.coverage_enabled and
             runtime and
             runtime.coverage_tool):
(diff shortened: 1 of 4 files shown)


python/private/py_executable.bzl
```diff
@@ -786,7 +786,7 @@ def _create_stage1_bootstrap(
         )
         template = runtime.bootstrap_template
         subs["%shebang%"] = runtime.stub_shebang
-    else:
+    elif main_py:
```
aignas, line 789:
```suggestion
    elif not ctx.files.srcs:
        fail("mandatory 'srcs' files have not been provided")
    else:
```
It is probably better to do it this way instead of failing at the very end.

python/private/py_executable.bzl
```diff
@@ -807,6 +807,8 @@ def _create_stage1_bootstrap(
         subs["%import_all%"] = ("True" if ctx.fragments.bazel_py.python_import_all_repositories else "False")
         subs["%imports%"] = ":".join(imports.to_list())
         subs["%main%"] = "{}/{}".format(ctx.workspace_name, main_py.short_path)
+    else:
+        fail("mandatory 'srcs' attribute has not been specified")
```
aignas, line 811:
```suggestion
```

rickeylev (approved):
I renamed the tests, otherwise LGTM

PR #2769, opened 2025-04-10
fix: Prevent absolute path creation in uv lock template

asa:
This change fixes a bug in the `lock` rule where, when the package is at the root level, the path to `requirements.txt` is constructed incorrectly with a leading double slash (`//requirements.txt`), causing it to be interpreted as an absolute path.

This change detects if the package is empty before constructing the output path.

Work towards #1975


python/uv/private/lock.bzl
```diff
@@ -327,10 +327,15 @@ def _maybe_file(path):
 def _expand_template_impl(ctx):
     pkg = ctx.label.package
     update_src = ctx.actions.declare_file(ctx.attr.update_target + ".py")
+
+    # Fix the path construction to avoid absolute paths
+    # If package is empty (root), don't add a leading slash
+    dst = ctx.attr.output if pkg == "" else "{}/{}".format(pkg, ctx.attr.output)
```
aignas, line 333:
```suggestion
    dst = "{}/{}".format(pkg, ctx.attr.output) if pkg else ctx.attr.output
```

This is a little bit shorter. I think the `lock` file rule has not been released yet, so not updating CHANGELOG is fine here.

aignas (approved):

PR #2771, opened 2025-04-13
fix(packaging): Format `METADATA` correctly if given empty `requires_file`

FrankPortman:
An empty `requires_file` used to be okay, but at some point regressed to leaving an empty line (due to the `metadata.replace(...)`) in the `METADATA` file - rendering the wheel uninstallable.

This PR initially attempted to solve that by introducing a new list that processed `METADATA` lines go into, rather than relying on repeated string replacement. But it seems like the repeated string replace actually did more than simply process one line at a time, so I reverted to a single substitution at the end.


examples/wheel/wheel_test.py
```diff
@@ -495,6 +495,30 @@ def test_requires_file_and_extra_requires_files(self):
...
+            metadata_lines = metadata.splitlines()
+
+            requires = []
+            for i, line in enumerate(metadata_lines):
+                if line.startswith("Name:"):
+                    self.assertTrue(metadata_lines[i + 1].startswith("Version:"))
+                if line.startswith("Requires-Dist:"):
+                    requires.append(line.strip())
+
+            print(requires)
```
aignas, line 519: Should this be removed?

aignas (commented):
Thanks, I am not sure I like using regex to solve this, but I see how doing that makes the change smaller and less invasive.

Do you know where the `Requires-Dist` appears from in the first place? Maybe we should error if the requirements are empty/not-specified? I see that the metadata file is read from a file, so I assume this could be templated out somewhere else in our code.

Extra things that this PR will need:
* The packaging rule should have `:::versionchanged:::` when we agree on the behaviour change.
* The CHANGELOG entry should be there mentioning the fix briefly.

examples/wheel/wheel_test.py
FrankPortman: Yep, I just copied and pasted it from the other test. Should I remove the print statement there as well?

aignas (approved):
Thank you!

PR #2772, opened 2025-04-13
feat(toolchain): Add new make vars for Python interpreter path compliant with `--no_legacy_external_runfiles`

### LINKED ISSUE #2728: make variables for runfiles location of $(PYTHON3) to support --no_legacy_external_runfiles
# 🚀 feature request

### Relevant Rules

`rules_python`

### Description

When using `--no_legacy_external_runfiles`, env vars set via `$(PYTHON3)` are not found at runtime due to the missing location in `external/...`.

Something that looks like https://github.com/bazelbuild/rules_java/pull/272 whereby `$(PYTHON3)` or a new make var, correctly stubs out runfile locations, so that the code is compatible with `--no_legacy_external_runfiles`.

Using `sys.executable` inside binary/test code to set the env var manually.

Using non hermetic Python.

FrankPortman:
Using these new make vars in `py_binary` or `py_test` will correctly find the interpreter when setting `--no_legacy_external_runfiles`.

Fixes #2728

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -122,6 +122,8 @@ Unreleased changes template.
 * (toolchains) Local Python installs can be used to create a toolchain
   equivalent to the standard toolchains. See [Local toolchains] docs for how to
   configure them.
+* (toolchains) Expose `$(PYTHON2_ROOTPATH)` and `$(PYTHON3_ROOTPATH)` which are runfiles
+  locations equivalents of `$(PYTHON2)` and `$(PYTHON3) respectively.
 
 
 {#v0-0-0-removed}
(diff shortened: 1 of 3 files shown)


aignas (commented):
The change LGTM. I think the `py2_runtime` at some point should be dropped entirely.

Things that need to be done before this is merged:
* Document the `PYTHON3_ROOTPATH` make var somewhere, [this file](https://github.com/bazel-contrib/rules_python/blob/aa0d16c1463e4e26f6ed633ae83d9785a2ea9dfa/docs/toolchains.md?plain=1#L215) seems to be a good place to do that.
* Add `:::versionchanged::: VERSION_NEXT_FEATURE` line in the docstring for the rule, similar to how it is done elsewhere.
* Add a line in the CHANGELOG about the added feature.
* A test to ensure that this works is optional since we don't have any tests using the `PYTHON3` var yet.

I might add these things if/when I have time myself, but if you want things to move quicker, please add these things.

aignas (approved):
This is perfect, thank you!

PR #2777, opened 2025-04-15
feat: declare more providers in py_* rules

layus:
TODO:

- [ ] Update `CHANGELOG.md` as applicable
- [ ] Body text describes:
  - [ ] Why this change is being made, briefly.
  - [ ] Before and after behavior, as applicable
  - [ ] References issue number, as applicable
- [ ] Add regression tests

### DIFF
--- a/python/private/py_executable.bzl
+++ b/python/private/py_executable.bzl
@@ -1854,6 +1854,8 @@ def create_base_executable_rule():
     """
     return create_executable_rule_builder().build()
 
+_MaybeBuiltinPyInfo = [BuiltinPyInfo] if BuiltinPyInfo != None else []
+
 # NOTE: Exported publicly
 def create_executable_rule_builder(implementation, **kwargs):
     """Create a rule builder for an executable Python program.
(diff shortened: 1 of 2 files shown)


rickeylev:
This mostly LGTM. The one thing to do is update the changelog.

I renamed this to fix instead of feature. They should have always been declared.

python/private/py_executable.bzl
```diff
@@ -1854,6 +1854,8 @@ def create_base_executable_rule():
     """
     return create_executable_rule_builder().build()
 
+_MaybeBuiltinPyInfo = [BuiltinPyInfo] if BuiltinPyInfo != None else []
```
aignas, line 1857: I lost track of all of bazel releases, b which bazel versions require this line? I thought with bazel 7 and above there is no BuiltinPyInfo.

rickeylev: I think it's bazel 9 where BuiltinPyXXX goes away entirely? Bazel 7 still has BuiltinPyInfo, and bazel 8 has the "auto load" flag, where it will conditionally autoload them or omit them entirely. I don't entirely remember offhand, though.

rickeylev (approved):

rickeylev:
You're welcome!

PR #2784, opened 2025-04-17
fix(ci): use ubuntu-latest for mypy action

aignas:


groodt (approved):
LGTM

Will this risk flakes if the version is pinned to latest? Or should we bump to a static version that works?

aignas:
I think the risk is low, because mypy action has the necessary pins.

PR #2785, opened 2025-04-17
exp(uv): support uv lock build action on Google RBE

aignas:
Work towards #1975


aignas:
Experiment did not work.

PR #2788, opened 2025-04-18
fix(rules): copy_propagating_kwargs() now also copies target_compatible_with

dws:
This routine already copies `compatible_with`, which is little used, but does not copy `target_compatible_with`, which is broadly used. This seems like an oversight.

I noticed this discrepancy when working on a system that assumes that any `tags` or `target_compatible_with` parameters supplied to a macro will propagate to all rules created by that macro. In rules_python, this already works for `tags`, but not for `target_compatible_with`.

It would be great to get this accepted upstream, so that I can stop patching rules_python.


rickeylev:
Nice catch. Yeah, this looks like an oversight. Thanks!

rickeylev (approved):
From a quick grep, copy_propagating is only used in a couple places, not all rules. I updated the changelog to better reflect that.

PR #2793, opened 2025-04-20
fix: use the python micro version to parse whl metadata in bzlmod

### LINKED ISSUE #2319: [pypi] Support python micro version in env marker based dependency evaluation

aignas:
Add `<micro>` version to the target platform. Instead of `cpxy_os_cpu`
the target platform string format becomes `cpxy.z_os_cpu`. This is a
temporary measure until we get a better API for defining target
platforms.

Summary:
- [x] test `select_whls` function needs to be tested to ensure that the
  whl selection is not impacted when we have the full version in the
  target platform.
- [ ] `download_only` legacy whl code path in `bzlmod` needs further
  testing.
- [x] test `whl_config_setting` handling and config setting creation.
  The config settings in the hub repo should not use the full version,
  because from the outside, the whl is compatible with all `micro`
  versions of a given `3.<minor_version>` of the Python interpreter.
  This means that the already documented config setting do not need to
  be changed.
- [x] `pep508_deps` tests for handling the `full_python_version`
  correctly.
- [x] `pep508_deps` tests for ensuring the `default_abi` is being
  handled correctly.

Fixes #2319


rickeylev (approved):

aignas:
I will cherry pick this to the 1.4 release branch to fix #2815.

PR #2807, opened 2025-04-22
fix: requires_file preserves extras that package depends on

skongum02:
When requirements are passed in through `requires_file` the extras are not preserved.  

eg if the contents of requires file is `example[extras]==1.1.1`, bazel will currently write to the METADATA file `Requires-Dist: example==1.1.1`.  This PR attempts to fix that by adding that back if there are any extras.

The expected output should be `Requires-Dist: example[extras]==1.1.1`


examples/wheel/BUILD.bazel
```diff
@@ -374,6 +386,22 @@ py_wheel(
     deps = [":example_pkg"],
 )
 
+py_wheel(
+    name = "requires_dist_depends_on_extras",
+    distribution = "requires_dist_depends_on_extras",
+    requires = [
```
skongum02, line 392: This currently passes without any code changes, but just adding this for completeness

skongum02:
Hello @rickeylev @aignas any chance you all can take a look at this PR?

aignas (commented):
The change LGTM. Could you please add a line to the `Fixed` section in CHANGELOG?

aignas (approved):
Thanks

PR #2811, opened 2025-04-22
test(pypi): add a test case for simpleapi html parsing with %

aignas:
In addition to #2801 I wanted to ensure that we are getting the correct
filename when downloading wheels. It seems that the `%` in the wheel
filename might get through wheels that get referenced via direct URL in
the requirements.txt files.

### DIFF
--- a/tests/pypi/parse_simpleapi_html/parse_simpleapi_html_tests.bzl
+++ b/tests/pypi/parse_simpleapi_html/parse_simpleapi_html_tests.bzl
@@ -303,6 +303,24 @@ def _test_whls(env):
                 yanked = False,
             ),
         ),
+        (
+            struct(
+                attrs = [
+                    'href="/whl/cpu/torch-2.6.0%2Bcpu-cp39-cp39-manylinux_2_28_aarch64.whl#sha256=deadbeef"',
+                ],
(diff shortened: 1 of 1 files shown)


tests/pypi/parse_simpleapi_html/parse_simpleapi_html_tests.bzl
```diff
@@ -303,6 +303,24 @@ def _test_whls(env):
...
+                filename = "torch-2.6.0+cpu-cp39-cp39-manylinux_2_28_aarch64.whl",
+                url = "https://example.org/",
+            ),
+            struct(
+                filename = "torch-2.6.0+cpu-cp39-cp39-manylinux_2_28_aarch64.whl",
+                metadata_sha256 = "",
+                metadata_url = "",
+                sha256 = "deadbeef",
+                version = "2.6.0+cpu",
+                url = "https://example.org/whl/cpu/torch-2.6.0%2Bcpu-cp39-cp39-manylinux_2_28_aarch64.whl",
```
rickeylev, line 320:
```suggestion
                # A URL with % could occur if directly written in requirements.
                url = "https://example.org/whl/cpu/torch-2.6.0%2Bcpu-cp39-cp39-manylinux_2_28_aarch64.whl",
```

rickeylev (approved):

PR #2814, opened 2025-04-23
fix: try multiple times to get win32 version to handle flakes

rickeylev:
The Google tensorflow/jax devinfra team reported that Windows 2022 with Python 3.12.8
has a tendency to be flaky when calling the platform.win32 APIs. I'm very certain I
saw similar behavior in the past myself.

To fix, just call the APIs a couple times; it seems to fix itself.

cc @vam-google


vam-google (approved):
LGTM

aignas (approved):

PR #2816, opened 2025-04-23
docs: document some of our project styles/conventions

rickeylev:
Spurred by the discussion to converge on using `.` to separate generated targets, I
wrote down some of the conventions we've adopted.


CONTRIBUTING.md
```diff
@@ -173,6 +173,47 @@ The `legacy_foo` arg was removed
...
+## Style and idioms
+
+For the most part, we just accept whatever the code formatters do, so there
+isn't much style to enforce.
+
+Some miscellanous style, idioms, and conventions we have are:
+
+### Markdown/Sphinx Style
+
+* Use colons (e.g. `:::`) for prose sections of text.
```
dougthor42, line 185: Do you want to include a note on the prose type? Eg `{topic}`, `{note}`, `{warning}`, etc.

CONTRIBUTING.md
```diff
@@ -173,6 +173,47 @@ The `legacy_foo` arg was removed
...
 ```
 
+## Style and idioms
+
+For the most part, we just accept whatever the code formatters do, so there
+isn't much style to enforce.
+
+Some miscellanous style, idioms, and conventions we have are:
+
+### Markdown/Sphinx Style
```
dougthor42, line 183: Do you want to specify on `*` over `+` for unordered lists? How about indentation?

dougthor42 (approved):
Couple questions but nothing blocking. LGTM.

CONTRIBUTING.md
rickeylev: Yeah, I'd be ok with that. I was looking at markdown formatters last week, actually. All the ones I found use `-` for lists (which I don't like but w/e)

dougthor42: I prefer `+` and 4-spaces myself, but don't care enough to fight.

I have no reason for `+` but my reason for 4 spaces is that it matches with ordered listed up to 99:

```markdown
1.  foo
    + bar
    + bar
2.  baz with a long list item that ends with a
    line continuation
99. foobar
```

CONTRIBUTING.md
rickeylev: Done

CONTRIBUTING.md
```diff
@@ -173,6 +173,47 @@ The `legacy_foo` arg was removed
...
+
+* Use colons for prose sections of text, e.g. `:::{note}`, not backticks.
+* Use backticks for code blocks.
+* Max line length: 100.
```
aignas, line 187: Having `.editorconfig` in the root of the repo to ensure that this is easy to follow would be great. `nvim` for example has native support for it, so do many other editors.

CONTRIBUTING.md
```diff
@@ -173,6 +173,47 @@ The `legacy_foo` arg was removed
...
+  definition.
+* The different build phases and interface/implementation pieces of rules should
+  be in different files. Stated another way: loading code for thing shouldn't
+  incur loading code that isn't relevant.
+  * Providers should be in their own files. This allows implementing a custom
+    rule that implements the provider without loading an existing
+    implementation.
+  * The repository phase shouldn't `load()` code that is only relevant to the
+    loading-phase or subsequent phases. Vice-versa, loading-phase code should
+    not load any repository-phase code.
```
aignas, line 204: What is the rationale here, e.g. we have `normalize_name` function that is used in both, repository and loading phases.

Or am I misunderstanding the statement here?

CONTRIBUTING.md
```diff
@@ -173,6 +173,47 @@ The `legacy_foo` arg was removed
...
+  * Generally, one rule per file. The goal is that defining an e.g. library
+    shouldn't incur loading all the code for binaries, tests, packaging, etc.
+* The public access point for anything should be a separate file that re-exports
+  the specific symbols that are public. This ensures our public API is well
+  defined and prevents accidentally exposing a package-private symbol as a
+  public symbol.
```
aignas, line 210: The documentation about the symbol should be put in there - i.e. the docstrings of the file next to the implementation do not do anything and we should add proper user facing documentation next to the exports.

CONTRIBUTING.md
```diff
@@ -173,6 +173,47 @@ The `legacy_foo` arg was removed
...
+* Repository rules should have name ending in `_repo`. This helps distinguish
+  them from regular rules.
```
aignas, line 212: hmmm. What do we do with existing `whl_library`, etc? I'm `+0` for the change though, might be good to add a cleanup.

However, I would like to have a list of permited names. For example I wanted to later create a `whl_archive` repo rule (similar to `http_archive`) that would be a starlark impl of extracting a `whl` and making it usable. I guess it would have to become `whl_archive_repo`?

CONTRIBUTING.md
rickeylev: I'm going to omit the list thing for now. I tried a few autoformatters for markdown, and they're all very opinionated and not very configurable. e.g. they rewrite all the lists. I'd rather we just pick an auto formatter and use it.

I'm going to keep the line length part, though. There's a few places in our docs where the lines are way way over 100 chars and it drives me nuts.

CONTRIBUTING.md
rickeylev: Oh neat. I added a basic editorconfig

I don't see a setting for line length, though?

CONTRIBUTING.md
rickeylev: I'll rephrase it to clarify.

The basic idea is:
* The repo phase shouldn't load code that calls rule(), module_extension(), provider(), etc. The objects those define aren't usable/inspectable/valid by/to the repo phase. Vice versa: the loading phase shouldn't call code that calls repository_rule, module_extension, etc
* Similarly, utility code like repo_utils.bzl shouldn't be loaded by the loading-phase because 99% of it is only relevant to the repository phase.

"generic" code (e.g. normalize_name) is fine for any phase to load.

CONTRIBUTING.md
rickeylev: Rephrased to incorporate that

CONTRIBUTING.md
rickeylev: We can just leave what's out of compliance as it is until someone really wants to scratch the itch.

I'll give http_archive a _slight_ pass because its "http" portion makes it pretty clear it's going to be doing something external.

Is `whl_archive` a repository rule that extracts a wheel, or is it a regular rule that accepts/produces a `.whl` archive as input/output?

I'd love to rename whl_library. It's `_library` suffix has always given me cognitive dissonance because every other `xxx_library` I see is a regular rule.  I would take whl_archive over whl_library 100%.

CONTRIBUTING.md
rickeylev: done

CONTRIBUTING.md
aignas: SGTM, thanks

CONTRIBUTING.md
aignas: It has `max_line_length`, no? https://devhints.io/editorconfig.

But I've seen some discussions about people wanting to drop it, but I think we are using it.

CONTRIBUTING.md
aignas: OK, SGTM. Having the `repo` makes sense though. The API of `rules_python` should be intuitive and having this explained somewhere in the docs would be good. Maybe such conventions belong not only in the style (which is for contributors) but in the API reference index, i.e. we should tell our users how the API is roughly organized.

.editorconfig
```diff
@@ -0,0 +1,13 @@
+# Unix-style newlines with a newline ending every file
+[*]
+end_of_line = lf
+insert_final_newline = true
```
aignas, line 4:
```suggestion
insert_final_newline = true
max_line_length = 100
```

.editorconfig
```diff
@@ -0,0 +1,13 @@
...
+
+# Set default charset
+[*]
+charset = utf-8
+
+# 4 space indentation
+[*.{py,bzl}]
```
aignas, line 11:
```suggestion
[*.{py,bzl,bazel}]
```

aignas (approved):

PR #2817, opened 2025-04-24
refactor: Add log_std(out|err) bools to repo_utils that execute a subprocess

dougthor42:
While making a local patch to work around #2640, I found that I had a need for running a subprocess (`gcloud auth print-access-token`) via `repo_utils.execute_checked_stdout`. However, doing so would log that access token when debug logging was enabled via `RULES_PYTHON_REPO_DEBUG=1`. This is a security concern for us, so I hacked in an option to allow a particular `execute_(un)checked(_stdout)` call to disable logging stdout, stderr, or both.

I figure this might be useful to others so I thought I'd upstream it.

`execute_(un)checked(_stdout)` now support `log_stdout` and `log_stderr` bools that default to `True` (which is the same behavior as before this PR.

When the subprocess writes to stdout and `log_stdout = False`, the logged message will show:

```
===== stdout start =====
<log_stdout = False; skipping>
===== stdout end =====
```

If the subprocess does not write to stdout, the debug log shows the same as before:

```
<stdout empty>
```

The above also applies for stderr, with text adjusted accordingly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,7 +77,9 @@ END_UNRELEASED_TEMPLATE
 
 {#v0-0-0-added}
 ### Added
-* Nothing added.
+* Repo utilities `execute_unchecked`, `execute_checked`, and `execute_checked_stdout` now
+  support `log_stdout` and `log_stderr` keyword arg booleans. When these are `True`
+  (the default), the subprocess's stdout/stderr will be logged.
 
 {#v0-0-0-removed}
 ### Removed
--- a/python/private/repo_utils.bzl
+++ b/python/private/repo_utils.bzl
@@ -98,6 +98,8 @@ def _execute_internal(
         arguments,
         environment = {},
         logger = None,
+        log_stdout = True,
+        log_stderr = True,
         **kwargs):
     """Execute a subprocess with debugging instrumentation.
 
@@ -116,6 +118,10 @@ def _execute_internal(
         logger: optional `Logger` to use for logging execution details. Must be
             specified when using module_ctx. If not specified, a default will
             be created.
+        log_stdout: If True (the default), write stdout to the logged message. Setting
+            to False can be useful for large stdout messages or for secrets.
+        log_stderr: If True (the default), write stderr to the logged message. Setting
+            to False can be useful for large stderr messages or for secrets.
         **kwargs: additional kwargs to pass onto rctx.execute
 
     Returns:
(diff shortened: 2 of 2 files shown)


python/private/repo_utils.bzl
```diff
@@ -98,6 +98,8 @@ def _execute_internal(
         arguments,
         environment = {},
         logger = None,
+        log_stdout = True,
+        log_stderr = True,
```
dougthor42, line 102: Alternative API option (but naming is hard):

```python
log_std = "both"  # one of ["both", "stdout", "stderr"]
```

(I'd prefer an enum but this is starlark not python :melting_face:)

LMK which API people would prefer.

python/private/repo_utils.bzl
```diff
@@ -331,11 +348,11 @@ def _env_to_str(environment):
...
     return kwargs.get("timeout", "<default timeout>")
 
-def _outputs_to_str(result):
+def _outputs_to_str(result, log_stdout = True, log_stderr = True):
     lines = []
     items = [
-        ("stdout", result.stdout),
-        ("stderr", result.stderr),
+        ("stdout", result.stdout if log_stdout else "<log_stdout = False; skipping>"),
+        ("stderr", result.stderr if log_stderr else "<log_stderr = False; skipping>"),
```
dougthor42, line 355: Any tips on how this should be tested?

CHANGELOG.md
```diff
@@ -77,7 +77,9 @@ END_UNRELEASED_TEMPLATE
 
 {#v0-0-0-added}
 ### Added
-* Nothing added.
+* Repo utilities `execute_unchecked`, `execute_checked`, and `execute_checked_stdout` now
```
rickeylev, line 80: The repo_utils apis are internal, so no need to document this in the user-facing changelog.

python/private/repo_utils.bzl
rickeylev: I don't think we have any unit tests of these.

python/private/repo_utils.bzl
rickeylev: I think the log_xxx API is fine. These are internal helpers, so easy to change if we want.

rickeylev (approved):

CHANGELOG.md
dougthor42: Ack :+1:, I'll keep that in mind for next time.

PR #2819, opened 2025-04-24
fix(pypi): call python --version before marker eval

### LINKED ISSUE #2818: Error: key "python_version" not found in dictionary
# 🐞 bug report

### Affected Rule

The issue is caused by the rule: multi_pip_parse

### Is this a regression?

Yes, the previous version in which this bug was not present was: 1.3.0.

`git bisect` says first problematic commit is bfad5078acee76f9743e6f1210784288571b0c0c.

### Description

`Error: key "python_version" not found in dictionary` from `marker_expr` in python/private/pypi/pep508_evaluate.bzl

## 🔬 Minimal Reproduction

When you have `rules_python` checked out:
``` shell
cd examples/multi_python_versions/
echo 'websockets ; python_version > "3.9"' >| requirements/requirements.in 
for n in 9 10 11; do
    bazelisk run --@rules_python//python/config_settings:python_version=3.$n \
        //requirements:requirements_3_$n.update
done
bazelisk query --enable_workspace --noenable_bzlmod '@pypi//...'
```

It is worth noting that `bazelisk query --noenable_workspace --enable_bzlmod '@pypi//...'` does work, e.g. it works with Bzlmod, but is broken without.

## 🔥 Exception or Error

<pre><code>
	File "[...]/external/rules_python+/python/private/pypi/pip_repository.bzl", line 74, column 50, in _pip_repository_impl
	File "[...]/external/rules_python+/python/private/pypi/parse_requirements.bzl", line 159, column 51, in parse_requirements
		env_marker_target_platforms = evaluate_markers(reqs_with_env_markers)
	File "[...]/external/rules_python+/python/private/pypi/evaluate_markers.bzl", line 35, column 24, in evaluate_markers
		if evaluate(req.marker, env = env(platform_from_str(platform, None))):
	File "[...]/external/rules_python+/python/private/pypi/pep508_evaluate.bzl", line 146, column 27, in evaluate
		tokens = ast.parse(env = env, tokens = tokens, strict = strict)
	File "[...]/external/rules_python+/python/private/pypi/pep508_evaluate.bzl", line 231, column 40, in lambda
		parse = lambda **kwargs: _parse(self, **kwargs),
	File "[...]/external/rules_python+/python/private/pypi/pep508_evaluate.bzl", line 258, column 27, in _parse
		expr = marker_expr(env = env, strict = strict, *tokens[:3])
	File "[...]/external/rules_python+/python/private/pypi/pep508_evaluate.bzl", line 322, column 19, in marker_expr
		left = env[left]
Error: key "python_version" not found in dictionary
</code></pre>

## 🌍  Your Environment

**Operating System:**

Ubuntu 24.04

<pre>
  <code>
Bazelisk version: v1.20.0
Build label: 8.2.1
Build time: Thu Apr 17 18:22:55 2025 (1744914175)
Build timestamp: 1744914175
Build timestamp as int: 1744914175
  </code>
</pre>

**Rules_python version:**

`main`, first problematic commit is bfad5078acee76f9743e6f1210784288571b0c0c.

**Anything else relevant?**

aignas:
`bzlmod` has the full python version information statically and we don't
need to call Python to get its version, but for `WORKSPACE` that is not
the case and we have to call it before evaluating the markers in
universal requirements files.

This also fixes transitions in the `compile_pip_requirements` macro where
the `.update` target would not transition correctly based on the
`python_version` parameter.

Fixes #2818


examples/multi_python_versions/requirements/requirements_lock_3_10.txt
```diff
@@ -1,10 +1,10 @@
 #
-# This file is autogenerated by pip-compile with Python 3.10
+# This file is autogenerated by pip-compile with Python 3.9
```
vonschultz, line 2: I think this file should be generated with Python 3.10.

examples/multi_python_versions/requirements/requirements_lock_3_11.txt
```diff
@@ -1,10 +1,10 @@
 #
-# This file is autogenerated by pip-compile with Python 3.11
+# This file is autogenerated by pip-compile with Python 3.9
```
vonschultz, line 2: I think this file should be generated with Python 3.11.

examples/multi_python_versions/requirements/requirements_lock_3_10.txt
aignas: Done, it seems that the transition had a bug, so this PR will fix 2 bugs at once :D

examples/multi_python_versions/requirements/requirements_lock_3_11.txt
aignas: Done

rickeylev (approved):

PR #2822, opened 2025-04-24
fix(gazelle): Skip indexing py_binary rules if a corresponding py_library rule contains the same srcs

yushan26:
This PR ensures that py_binary rules are not indexed into Gazelle's IndexMap when there is a corresponding py_library rule with the same srcs. When both py_library and py_binary targets share the same file in their `srcs`, Gazelle previously indexed both under the same import path. This led to ambiguity and resolution errors, as Gazelle found multiple rules for the same language (py).

To resolve this, the PR updates Gazelle to skip indexing py_binary rules, allowing py_library to be the only import being indexed.

Testing
An integration test was added where both a py_library and a py_binary rule include the same script.py in their srcs. The test verifies that Gazelle correctly resolves the import to the py_library target.

### DIFF
--- a/gazelle/python/generate.go
+++ b/gazelle/python/generate.go
@@ -123,6 +123,7 @@ func (py *Python) GenerateRules(args language.GenerateArgs) language.GenerateRes
 			pyFileNames.Add(f)
 			if !hasPyBinaryEntryPointFile && f == pyBinaryEntrypointFilename {
 				hasPyBinaryEntryPointFile = true
+				pyLibraryFilenames.Add(f)
 			} else if !hasPyTestEntryPointFile && f == pyTestEntrypointFilename {
 				hasPyTestEntryPointFile = true
 			} else if f == conftestFilename {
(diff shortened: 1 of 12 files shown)


arrdem:
I don't think that just not indexing `py_binary` sources is a workable solution. A pattern we've seen before is that users will want to consider sources part of a `py_binary` and explicitly not want to manage what's perceived to be a duplicate `py_library` target, instead expecting other use sites to depend on that fileset via the `py_binary` target.

This has obvious downsides in that the entrypoint script gets materialized among other things, but it's an existing and working use-case.

I think the behavior in this case should simply be to _prefer_ the library target in case both are an option, not to just not index the binary.

linzhp:
>  A pattern we've seen before is that users will want to consider sources part of a py_binary and explicitly not want to manage what's perceived to be a duplicate

Are you talking about the scenarios when `py_binary` is generated but not `py_library`?

yushan26:
Gotcha, I updated the logic to skip indexing the py_binary rule only if there is a corresponding py_library rule that also has the same`srcs` populated. Let me know if this makes sense.

linzhp:
@dougthor42 can you also take a look at this PR?

One thing I am not sure is whether Gazelle can generate `py_binary` without generating `py_library` like @arrdem said.

linzhp:
Should we make Gazelle always generate `py_library` and stop indexing any `py_binary`? This is more consistent with other languages.

yushan26:
> Should we make Gazelle always generate `py_library` and stop indexing any `py_binary`? This is more consistent with other languages.

Yea that's probably more consistent, although this may be unexpected behavior to the existing gazelle extension.

PR #2827, opened 2025-04-25
feat: implement less/greater operators for string for env marker evaluation

rickeylev:
Right now, if two strings are compared, it results in an error.

Per spec, strings are suppose to "use the python behavior". Starlark is going
to use Java semantics underneath, but it should behave close enough for the (almost
exclusively) ASCII input that will be used.

### DIFF
--- a/python/private/pypi/pep508_evaluate.bzl
+++ b/python/private/pypi/pep508_evaluate.bzl
@@ -344,6 +344,14 @@ def _env_expr(left, op, right):
         return left in right
     elif op == "not in":
         return left not in right
+    elif op == "<":
+        return left < right
+    elif op == "<=":
+        return left <= right
+    elif op == ">":
+        return left > right
+    elif op == ">=":
+        return left >= right
     else:
         return fail("TODO: op unsupported: '{}'".format(op))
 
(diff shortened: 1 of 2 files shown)


groodt (approved):
LGTM

The implementation in packaging is here for reference https://github.com/pypa/packaging/blob/main/src/packaging/markers.py

aignas (approved):

python/private/pypi/pep508_evaluate.bzl
```diff
@@ -344,6 +344,14 @@ def _env_expr(left, op, right):
...
     elif op == "not in":
         return left not in right
+    elif op == "<":
+        return left < right
+    elif op == "<=":
+        return left <= right
+    elif op == ">":
+        return left > right
+    elif op == ">=":
+        return left >= right
```
aignas, line 354: Actually, could you please add tests to `//tests/pypi/pep508:evaluate_tests.bzl` to document what this behaviour is?

I know it will be very tedious, but having tests that cover all branches is a good form of semi-documentation.

rickeylev: Done.

(Was fairly easy once you pointed me to the test file, thanks!)

PR #2829, opened 2025-04-26
feat: allow specifying arbitrary constraints for local toolchains

rickeylev:
This adds the ability for local toolchains to have arbitrary constraints set on them.

This allows accomplishing two goals:
1. Makes it easier to enable/disable them on the command line, instead of having them
   entirely override an existing config and having to comment/uncomment the
   MODULE.bazel file sections.
2. Allows configuring them so that the repository is never initialized, which avoids
   the repository from being initialized during toolchain resolution, even if it will
   never match because of (1).


aignas (approved):
LGTM, but please add `CHANGELOG.md` entry to point to the new docs.

rickeylev:
changelog updated

PR #2832, opened 2025-04-27
refactor: add rule to do analysis time evaluation of environment markers

rickeylev:
wip/prototype to help bootstrap the impl of an analysis-time flag that evaluates
the pep508 dep specs

Creating a PR to make collab easier (maintainers can directly edit)

TODO:
* Remove the todo markers after discussion


python/private/pypi/dependency_specifier_flag.bzl
```diff
@@ -0,0 +1,84 @@
+load("//python/private:toolchain_types.bzl", "TARGET_TOOLCHAIN_TYPE")
+load(":pep508_evaluate.bzl", "evaluate")
+
+def _impl(ctx):
+    env = {}
+
+    runtime = ctx.toolchains[TARGET_TOOLCHAIN_TYPE].py3_runtime
```
aignas, line 7: I was wondering if we should get it from the toolchain info or the `//python/config_settings:python_version` flag. Probably better from toolchain as it will be more representative, but then we need to fetch the toolchain.

python/private/pypi/dependency_specifier_flag.bzl
```diff
@@ -0,0 +1,84 @@
...
+    if runtime.interpreter_version_info:
```
aignas, line 8: Ah, you are falling back to the flags, LGTM.

python/private/pypi/dependency_specifier_flag.bzl
```diff
@@ -0,0 +1,84 @@
...
+        env["python_full_version"] = full_version
+        env["implementation_version"] = full_version
+
+    # We assume cpython if the toolchain doesn't specify because it's most
+    # likely to be true.
+    env["implementation_name"] = runtime.implementation_name or "cpython"
+
+    env["os_name"] = struct()
+    env["platform_machine"] = struct()
+    env["platform_python_implementation"] = struct()
```
aignas, line 26: This could be `CPython` if `cpython` is implementation name. We could probably have that as a flag/or attach it to the `runtime`.

python/private/pypi/dependency_specifier_flag.bzl
```diff
@@ -0,0 +1,84 @@
...
+        env["python_version"] = get_flag(ctx.attr._python_version)
+        full_version = get_flag(ctx.attr._python_full_version)
+        env["python_full_version"] = full_version
+        env["implementation_version"] = full_version
+
+    # We assume cpython if the toolchain doesn't specify because it's most
+    # likely to be true.
+    env["implementation_name"] = runtime.implementation_name or "cpython"
+
+    env["os_name"] = struct()
```
aignas, line 24: We can have a select statement in that sets a string based on `@platform//os:` or the user can override it. Have the default based on some logic could help.

aignas (commented):
Shall I push some of my thoughts as code or leave this at a comment level in the PR?

python/private/pypi/dependency_specifier_flag.bzl
```diff
@@ -0,0 +1,97 @@
...
+    env["platform_python_implementation"] = runtime.implementation_name
+
+    # todo: add flag to carry this
+    env["platform_release"] = struct()
+
+    # todo: map from os constraint
+    env["platform_system"] = struct()
+
+    # todo: add flag to carry this
+    env["platform_version"] = struct()
```
aignas, line 44: A note to myself. Right now `pypi/config_settings` is generating something similar for `osx_version` and maybe this flag could be used instead of creating a new separate flag for the `osx_version`.

rickeylev:
> Shall I push some of my thoughts as code or leave this at a comment level in the PR?

Yes, go ahead and push changes.

python/private/pypi/dependency_specifier_flag.bzl
rickeylev: Done.

(aside: this value is kinda weird, see https://github.com/bazel-contrib/rules_python/issues/2826#issuecomment-2833578913 )

python/private/pypi/dependency_specifier_flag.bzl
rickeylev: I dug through peps a bit. From what i can tell, sys.implementation.name was created to replace platform.python_implementation. It turns out the `platform` uses regex parsing of the `sys.version` string to determine its values :astonished: .

The docs say there are 4 possible values, 2 of which are essentially defunct (IronPython, Jython). The other two are CPython and PyPy.

I just added an if condition to transform cpython to CPython. If we know the sys.implementation.name value that pypy uses, we could special case that, too.

rickeylev (commented):
mapped in the various things. added 2 flags.

rickeylev:
As I've poked around the various pep508.bzl files, I see there's a lot of little utilities, tests, edge cases, etc. I'm just plowing ahead and copy/pasting things. Feel free to point out specific places where something should be wired into something else

python/private/pypi/dependency_specifier_flag.bzl
```diff
@@ -0,0 +1,141 @@
...
+        }),
+    )
+
+pypa_dependency_specification = rule(
+    implementation = _impl,
+    attrs = {
+        "expression": attt.string(),
+        "os_name": attr.string(),
+        "sys_platform": attr.string(),
+        "platform_machine": attr.string(),
```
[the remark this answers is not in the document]
rickeylev: Yeah, that'd be nice. Unfortunately, I don't see a direct way to get that flag from Starlark. The closet I could find was CcToolchain.cpu. I'll ask around just in case I missed something. At the least, we can simply map all the platform:cpu values to what we know of today

rickeylev:
Ok, basic flag with tests is working.

Based on the convo in the other issue, it's pretty clear I didn't understand the jargon and renaming things is in order.

rickeylev:
renamed to env_marker_setting, since it's all about evaluate env markers.

Saw this comment you added:

> NOTE: Platform release for Android will be Android version:

I think there are android-specific flags in the android rules that could tell that info. However, we wouldn't want to depend on the android rules. If we want to cleanly support this, then a toolchain would probably be best. The way it could work is using toolchains in a plugin model type of way. Depending on the android rules runs their MODULE.bazel, which registers e.g. `(toolchain_type=rules_python_platform_release, target_compatible_with=[android], target_settings=[@android//:version=X] value=X)`

rickeylev:
How is the `extra` env marker supposed to be evaluated? I'm having trouble finding an intelligible specification or example. All I've really found is `extra == "pdf"` as an example in pypa core metadata spec. So its a plain string? But there can be many extras? I don't get it.

rickeylev:
Well, this started as a prototype, but it's actually functional now! PTAL

We could probably skip straight to analysis time dependency selection when the rest of the rest of the starlark pep 508 functionality is ready. I was trying to figure out how to re-activate those code paths for testing, but it wasn't immediately obvious.

aignas:
> How is the extra env marker supposed to be evaluated? I'm having trouble finding an intelligible specification or example. All I've really found is extra == "pdf" as an example in pypa core metadata spec. So its a plain string? But there can be many extras? I don't get it.

Yeah, `extra` is just a string and it never is a version, so `!=` and `==` in theory should be enough. However, expressions with `extra` should never passed to this `marker_setting`.

aignas:
> Well, this started as a prototype, but it's actually functional now! PTAL
> 
> We could probably skip straight to analysis time dependency selection when the rest of the rest of the starlark pep 508 functionality is ready. I was trying to figure out how to re-activate those code paths for testing, but it wasn't immediately obvious.

Regarding this, I think you are right, we should go straight to `analysis` time. My thinking is that we need to have a list of `deps` that the hub repo has. Instead of `target_platforms` just have a list of `whl` names. Once we have `Requires-Dist` list, process `extra` conditions in the expressions and combine the expressions so that we have one expression per package. filter out the requires dist so that it only contains the dependencies from `deps`, so that `bazel query` still works. Generate extra targets using what is in this PR.

python/private/pypi/env_marker_setting.bzl
```diff
@@ -0,0 +1,265 @@
...
+    "@platforms//os:linux": "linux",
+    "@platforms//os:osx": "darwin",
+    "@platforms//os:windows": "win32",
+    "@platforms//os:wasi": "wasi",
+    # NOTE: The below values are approximations. The sys.platform() docs
+    # don't have documented values for these OSes. Per docs, the
+    # sys.platform() value reflects the OS at the time Python was *built*
+    # instead of the runtime (target) OS value.
+    "@platforms//os:freebsd": "freebsd",
+    "@platforms//os:openbsd": "openbsd",
```
aignas, line 44: If I am reading it correctly, this value might be `freebsd8` and I am wondering if we should do some trickery to combine this value with the `platform_release`?

python/private/pypi/env_marker_setting.bzl
```diff
@@ -0,0 +1,265 @@
...
+    "@platforms//cpu:wasm64": "wasm64",
+    "@platforms//cpu:x86_32": "x86_32",
+    "@platforms//cpu:x86_64": "x86_64",
+    # The value is empty string if it cannot be determined:
+    # https://docs.python.org/3/library/platform.html#platform.machine
+    "//conditions:default": "",
+}
+
+# todo: copied from pep508_env.bzl
+_platform_system_select_map = {
```
aignas, line 88: It would be great to support users overriding the value with their own supplied flag.

python/private/pypi/env_marker_setting.bzl
```diff
@@ -0,0 +1,265 @@
...
+        full_version = _get_flag(ctx.attr._python_full_version)
+        env["python_full_version"] = full_version
+        env["implementation_version"] = full_version
+
+    # We assume cpython if the toolchain doesn't specify because it's most
+    # likely to be true.
+    env["implementation_name"] = runtime.implementation_name or "cpython"
+    env["os_name"] = ctx.attr.os_name
+    env["sys_platform"] = ctx.attr.sys_platform
+    env["platform_machine"] = ctx.attr.platform_machine
```
aignas, line 150: Having label flags for these that fallback to auto-detected values would be great.

python/private/pypi/env_marker_setting.bzl
```diff
@@ -0,0 +1,265 @@
...
+    # `platform.python_implementation()`, however, PEP 421 introduced
+    # `sys.implementation.name` to replace it. There's now essentially just two
+    # possible values it might have: CPython or PyPy. Rather than add a field to
+    # the toolchain, we just special case the value from
+    # `sys.implementation.name`
+    platform_python_impl = runtime.implementation_name
+    if platform_python_impl == "cpython":
+        platform_python_impl = "CPython"
+    elif platform_python_impl == "pypy":
+        platform_python_impl = "PyPy"
```
aignas, line 162:
```suggestion
    elif platform_python_impl == "pypy":
        platform_python_impl = "PyPy"
    elif platform_python_impl == "rustpython":
        platform_python_impl = "RustPython"
```

Although that is not strictly correct: https://github.com/RustPython/RustPython/blob/cd1c9be0e1a3b59c5d49c9edf593d66538c5e530/Lib/platform.py#L1082

`RustPython` can attempt to be `CPython` if "needed".

I'm fine with leaving it out for now though.

python/private/pypi/env_marker_setting.bzl
```diff
@@ -0,0 +1,265 @@
...
+    "@platforms//os:netbsd": "NetBSD",
+    "@platforms//os:openbsd": "OpenBSD",
+    "@platforms//os:osx": "Darwin",
+    "@platforms//os:windows": "Windows",
+    # The value is empty string if it cannot be determined:
+    # https://docs.python.org/3/library/platform.html#platform.machine
+    "//conditions:default": "",
+}
+
+def env_marker_setting(*, name, expression, **kwargs):
```
aignas, line 105: I am wondering if we should create a `config_setting` almost all the time because you always need the `is_yes` or `is_no` associated with this target, or can we use this directly in the `select` statement?

groodt:
> Yeah, extra is just a string and it never is a version, so != and == in theory should be enough. However, expressions with extra should never passed to this marker_setting

My 2c is that extras are more appropriate for to be passed as flags. 

And markers are more appropriate to be mapped from platforms. 

Extras are generally user provided and markers are generally build system / environment provided.

rickeylev:
I pulled out the newly added flags soas not to affect the public API for now. PTAL

I'll send a separate PR to factor out a common "env marker config" thing, as described in https://github.com/bazel-contrib/rules_python/issues/2826#issuecomment-2842970342

python/private/pypi/pep508_env.bzl
```diff
@@ -64,12 +117,45 @@ _sys_platform_values = {
     "osx": "darwin",
     "windows": "win32",
 }
```
aignas, line 119: nit: to keep better compatibility, this could be derived from the map:
```
_sys_platform_values = {
    k.partition(":")[-1]: v
    for k, v in sys_platform_select_map.items()
    if k.startswith("@platform")
}
```

The same with `platform_machine`.

python/private/pypi/env_marker_setting.bzl
```diff
@@ -0,0 +1,186 @@
...
+        expression = expression,
+        os_name = select(os_name_select_map),
+        sys_platform = select(sys_platform_select_map),
+        platform_machine = select(platform_machine_select_map),
+        platform_system = select(platform_system_select_map),
+        platform_release = select({
+            "@platforms//os:osx": "USE_OSX_VERSION_FLAG",
+            "//conditions:default": "",
+        }),
+        **kwargs
```
aignas, line 50: +1 for creating a separate rule with it's own provider which would allow users to override the `env` that is used here.

aignas (commented):
LGTM other than the map/value duplication, but that is reasonably easy to fix, I think.

aignas (approved):
Since factoring out is in the separate PR. Feel free to do the `value/map` deduping in that PR as well.

Issue #2830, opened 2025-04-27
Revert the starlark PEP508 evalutaion


aignas:
This is just to track progress on reverting this for the 1.4.0 release as it was a source of issues.

The list of PRs/things to do are:
- [x] Add the Python implementation back with unit tests.
- [x] Backport the #2793 to the Python codebase.
- [ ] Wire the python code back for the marker evaluation.
- [ ] Wire the python code back for the `deps` generation.

I will be working on this one piece at a time to have steady progress.

aignas:
Reopened #2423 as a feature request that was open to remove Python code that is being added back in here.

aignas:
A release has been cut: https://github.com/bazelbuild/bazel-central-registry/pull/4459

PR #2833, opened 2025-04-27
fix(pypi) backport python_full_version fix to Python

aignas:
Handling of `python_full_version` correctly has been fixed in the Starlark
implementation in #2793 and in this PR I am backporting the changes to handle
the full python version target platform strings so that we can have the same
behaviour for now.

At the same time I have simplified and got rid of the specialization handling
in the Python algorithm just like I did in the starlark, which simplifies the
tests and makes the algorithm more correct.

Summary:
* Handle `cp3x.y_os_arch` strings in the `platform.py`
* Produce correct strings when the `micro_version` is unset. Note, that we use version `0` in evaluating but we use the default version in the config setting. This is to keep compatibility with the current behaviour when the target platform is not fully specified (which would be the case for WORKSPACE users).
* Adjust the tests and the code to be more similar to the starlark impl.

Work towards #2830

### DIFF
--- a/python/private/pypi/whl_installer/platform.py
+++ b/python/private/pypi/whl_installer/platform.py
@@ -292,9 +292,6 @@ class Platform:
             "platform_release": "",  # unset
             "platform_version": "",  # unset
             "python_version": f"3.{minor}",
-            # FIXME @aignas 2024-01-14: is putting zero last a good idea? Maybe we should
-            # use `20` or something else to avoid having weird issues where the full version is used for
-            # matching and the author decides to only support 3.y.5 upwards.
             "implementation_version": f"3.{minor}.{micro}",
             "python_full_version": f"3.{minor}.{micro}",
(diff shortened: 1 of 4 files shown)


python/private/pypi/whl_installer/platform.py
```diff
@@ -292,12 +291,12 @@ def env_markers(self, extra: str) -> Dict[str, str]:
             "platform_system": self.platform_system,
             "platform_release": "",  # unset
             "platform_version": "",  # unset
-            "python_version": f"3.{minor_version}",
+            "python_version": f"3.{minor}",
             # FIXME @aignas 2024-01-14: is putting zero last a good idea? Maybe we should
             # use `20` or something else to avoid having weird issues where the full version is used for
             # matching and the author decides to only support 3.y.5 upwards.
```
aignas, line 297:
```suggestion
```

PR #2846, opened 2025-04-30
fix: Add COVERAGE_RCFILE to accept customized .coveragerc

nlou9:
Revisit this closed issue, https://github.com/bazel-contrib/rules_python/issues/1217. 
We still face the same issue. Inspired by the comments in the issue thread, we add a `COVERAGE_RCFILE` environment variable to be able to pass the customized `.coveragerc` file. 

In our own project, we patched this way, and it resolved our issue.

```
py_test(
    name = "test_utils",
    srcs = ["test_utils.py"],
    data = ["//:.coveragerc"],
    env = {
        "COVERAGE_RCFILE": "$(location //:.coveragerc)",
    },
    deps = [
        ":conftest",
        "//tests/data",
    ],
)
```


aignas:
I see that you are continuously pushing updates to this PR and thought I'd give a few thoughts about the possibility to merge this.

The things that I think this feature would need are:
* Supporting `bootstrap=script` code paths in the same way.
* Having extra documentation in the `docs` folder about this feature.
* Copying the referenced file so that the srcs can be populated by `rules_python`. This is to ensure that coverage report is correct. I am not sure how exactly this could work from the user point of view.
* CHANGELOG updates once we have the design above done/ready.

Open questions about the interface:
* Should we have something like `--@rules_python//python/config:coveragerc` label flag instead of env variables? I personally would prefer that and users could use transitions to force a particular coveragerc value if needed.

There are probably a few other things that are missing in my quick analysis, so happy to hear other thoughts about this.

Issue #2826, opened 2025-04-25
Full pep 508 / pypa dependency specifier support in starlark


rickeylev:
This issue is to track a starlark implementation for the pypa-spec for requirements line parsing and metadata Requires-Dist parsing.

Right now, we have a partial implementation in pep508_evaluate.bzl.

Some relevant docs:

* https://packaging.python.org/en/latest/specifications/dependency-specifiers
* https://packaging.python.org/en/latest/specifications/version-specifiers
* (historical) https://peps.python.org/pep-0508/
* @groodt mentioned pypa/packaging has a set of tests ; maybe we can consume those somehow

TODO
* [x] `>=, <=, <, >` operators for string comparisons
* [x] `===` string operator (I think?)
* [x] `~=` string operator (I think?)
* [x] [data-based version comparison](https://packaging.python.org/en/latest/specifications/version-specifiers/#final-releases:~:text=Date%20based%20release%20segments%20are%20also%20permitted)
* [x] Implement [pypa normalization rules](https://packaging.python.org/en/latest/specifications/version-specifiers/#normalization)
- [x] `python_version` - use the `X.Y` python version
- [x] `python_full_version` - use the `X.Y.Z` python version
- [x] `os_name` - use a value derived from `@platforms//os:<name>` values
- [x] `sys_platform` - use a value derived from `@platforms//os:<name>` values
- [x] `platform_release` - use `""` for the value
- [x] `platform_system` - use a value derived from `@platforms//os:<name>` values
- [x] `platform_version` - uses `""` for the value, missed the fact that we should default to `0`
- [x] `platform_machine` - use a value derived from `@platforms//cpu:<name>` values
- [x] `platform_python_implementation` - Use `CPython` constant.
- [x] `implementation_name` - Use `cpython` constant.
- [x] Support `*` in version comparison (see https://peps.python.org/pep-0440/#compatible-release)

Some keys seem tedious to support

**platform_release, platform_version**

For pip, this is simply the host machine value. For bazel, it would mean whatever the _target_ platform _would_ return. Using the bazel host machine value wouldn't be correct for cross builds, but will probably have to suffice as the first impl and in the short term.

Long term, I think we'll need a `--platform_{release,version}` flag (or toolchain?), which points to a target that can provide the appropriate value.

Notable bits from the spec:

----

> Variables whose value cannot be calculated on a given Python implementation should evaluate to 0 for versions, and an empty string for all other variables.

[...]

PR #2853, opened 2025-05-03
refactor: make env marker config available through target and flag

rickeylev:
This factors creation of (most of) the env marker dict into a separate target and
provides a label flag to allow customizing the target that provides it.

This makes it easier for users to override how env marker values are computed. The
`env_marker_setting` rule will still, if necessary, compute values from the toolchain,
but existing keys (computed from the env marker config target) have precedence.

The `EnvMarkerInfo` provider is the interface for implementing a custom env marker
config target; it will be publically exposed in a subsequent PR.

Along the way, unify how the env dict and defaults are set.


aignas (commented):
LGTM in general, we can talk about exposing the `EnvMarkerInfo` provider in a followup PR that uncomments the docs.

aignas:
Created #2855 to allow enabling the starlark deps paths more easily.

python/private/pypi/flags.bzl
```diff
@@ -97,3 +109,58 @@ This rule allows us to greatly reduce the number of config setting targets at no
...
+    # Similar for iOS:
+    # https://peps.python.org/pep-0730/#platform
+    platform_release = ctx.attr.platform_release
+    if platform_release == "USE_OSX_VERSION_FLAG":
+        platform_release = _get_flag(ctx.attr._pip_whl_osx_version_flag)
+    env["platform_release"] = platform_release
+    env["platform_system"] = ctx.attr.platform_system
+
+    # For lack of a better option, just use an empty string for now.
+    env["platform_version"] = ""
```
aignas, line 143:
```suggestion
    env["platform_version"] = "0"
```

This should be safe here. If you could update the `pep508_env` file for the same, that would tick one of the TODO items in #2826.

rickeylev: Done. I unified how the env creation and defaults are set as part of this.

aignas (approved):

PR #2711, opened 2025-03-31
py_wheel: always generate zip64-capable wheels

devtbi:
Currently, there is no possibility to pass the force zip64 option to the wheel creation. This hinders creation of packages that contain >2Gb files (e.g. large projects with debug symbols).

To fix, always generate zip64 capable wheels. zip64 support is wide spread.

Fixes https://github.com/bazel-contrib/rules_python/issues/2852

PR #2854, opened 2025-05-03
fix: use force_zip64=True in py_wheel to support larger files

### LINKED ISSUE #2852: "File size too large" error when making wheel
# 🐞 bug report

### Affected Rule

The issue is caused by the rule: `py_wheel`

### Is this a regression?

Not sure

### Description

Hello, it appears that when one of the `deps` for `py_wheel` has too large of a file size, building the target fails with

     raise RuntimeError("File size too large, try using force_zip64")

Docs for [ZipFile.open](https://docs.python.org/3/library/zipfile.html#zipfile.ZipFile.open) say

> When writing a file, if the file size is not known in advance but may exceed 2 GiB, pass `force_zip64=True` to ensure that the header format is capable of supporting large files. If the file size is known in advance, construct a [ZipInfo](https://docs.python.org/3/library/zipfile.html#zipfile.ZipInfo) object with [file_size](https://docs.python.org/3/library/zipfile.html#zipfile.ZipInfo.file_size) set, and use that as the name parameter.

Maybe it would make sense to use `force_zip64=True` [here](https://github.com/bazel-contrib/rules_python/blob/1.3.0/tools/wheelmaker.py#L157)?

## 🔬 Minimal Reproduction

(don't have one currently)

## 🔥 Exception or Error

<pre><code>
  ...
  File ".../external/rules_python+/tools/wheelmaker.runfiles/_main/../rules_python+/tools/wheelmaker.py", line 157, in add_file
    with self.open(zinfo, "w") as fdst:
  File "../external/rules_python++python+python_3_11_x86_64-unknown-linux-gnu/lib/python3.11/zipfile.py", line 1201, in close
    raise RuntimeError("File size too large, try using force_zip64")
</code></pre>

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
Linux x86_64
  </code>
</pre>

<pre>
  <code>
8.1.1
  </code>
</pre>

**Rules_python version:**

<pre>
  <code>
1.3.0
  </code>
</pre>

**Anything else relevant?**

tbarrella:
- Before, `py_wheel` throws an error when trying to include files above 2 GB
- After, `py_wheel` uses zip64 to support large files
- Fixes #2852

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,6 +76,8 @@ END_UNRELEASED_TEMPLATE
   multiple times.
 * (tools/wheelmaker.py) Extras are now preserved in Requires-Dist metadata when using requires_file
   to specify the requirements.
+* (packaging) {obj}`py_wheel` now uses zip64 to support larger files.
+  Fixes [#2852](https://github.com/bazel-contrib/rules_python/issues/2852).
 
 {#v0-0-0-added}
 ### Added
--- a/examples/wheel/test_publish.py
+++ b/examples/wheel/test_publish.py
@@ -104,7 +104,7 @@ class TestTwineUpload(unittest.TestCase):
     </head>
     <body>
         <h1>Links for example-minimal-library</h1>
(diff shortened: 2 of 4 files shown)


tbarrella:
#2711 was merged

PR #2855, opened 2025-05-03
feat: add an env variable to toggle pipstar

aignas:
This is a flag to start leveraging of the new code paths. The Starlark
implementation has been added in 1.4 and has been reverted in the latest
release candidates. The `env` variable will be a good way to roll it out more
gradually and get more testing.

For now we are switching only the `whl_library` internals as the
`requirements.txt` files from `uv` may use `*` in `python_full_version` and
`platform_version` that are not yet fully supported (#2826).

Main goals for this is to start using Starlark implementation so that we don't
have any hidden variables. What is more, having this in Starlark is the most
maintainable long-term solution for supporting cross-platform builds.

Work towards #260


rickeylev (approved):
I pushed a fix to wheel_installer_test. CI is passing.

I'll approve so you can merge, but since this is draft, i'll let you trigger merging

aignas:
I expected this to need more work to pass tests, thanks, will merge now.

Issue #2821, opened 2025-04-24
More general support for platform_release


baxelrod-bdai:
# 🐞 bug report

### Description

The `>=` operator should be supported for `platform_release`.

## 🔬 Minimal Reproduction

`requirements.in` has just:
```
keyboard
```

`BUILD.bazel` has:
```
load("@rules_uv//uv:pip.bzl", "pip_compile")

pip_compile(
    name = "generate_requirements",
    extra_args = [
        "--universal",
    ],
    requirements_in = "requirements.in",
    requirements_txt = "requirements.lock",
)
```

Running `bazel run //:generate_requirements` creates a .lock file with lots of entries like:
```
pyobjc-framework-businesschat==11.0 ; platform_release >= '18.0' and sys_platform == 'darwin' \
    --hash=sha256:1f732fdace31d2abdd14b3054f27a5e0f4591c7e1bef069b6aeb4f9c8d9ec487 \
    --hash=sha256:20fe1c8c848ef3c2e132172d9a007a8aa65b08875a9ca5c27afbfc4396b16dbb \
    --hash=sha256:47a2e4da9b061daa89a6367cb0e6bb8cdea0627379dd6d5095a8fd20243d8613
    # via pyobjc
pyobjc-framework-calendarstore==11.0 ; platform_release >= '9.0' and sys_platform == 'darwin' \
    --hash=sha256:40173f729df56b70ec14f9680962a248c3ce7b4babb46e8b0d760a13975ef174 \
    --hash=sha256:67ddc18c96bba42118fc92f1117b053c58c8888edb74193f0be67a10051cc9e2 \
    --hash=sha256:9b310fe66ac12e0feb7c8e3166034bec357a45f7f8b8916e93eddc6f199d08c8
    # via pyobjc
```

Then when you try to `bazel run` anything, you get the crash:
```
ERROR: /home/baxelrod/.cache/bazel/_bazel_baxelrod/223f490fa0c2f2ec38707ddac3dd6272/external/rules_python+/python/private/pypi/pep508_evaluate.bzl:348:20: Traceback (most recent call last):
	File "/home/baxelrod/.cache/bazel/_bazel_baxelrod/223f490fa0c2f2ec38707ddac3dd6272/external/rules_python+/python/private/pypi/extension.bzl", line 611, column 25, in _pip_impl
		mods = parse_modules(module_ctx)
[...]

PR #2856, opened 2025-05-03
fix(pypi): finish PEP508/PEP440 impl for version matching

### LINKED ISSUE #2826: Full pep 508 / pypa dependency specifier support in starlark

aignas:
This reuses the previous work by @vonschultz who implemented
a PEP440 version normalizer. We extend it and use it in the
PEP508 marker evaluation.

Summary:
- Extend the normalization parser to output individual parts
  of the versions to the parsing context.
- Re-implement all of the version comparison calls to use the
  parsed version.
- Add extra validation for `.*` usage in the environment markers
- Fallback to non-version matching in the environment markers
  if one of the sides is not a version.
- Rename the original normalizer file to `version.bzl` because
  as far as Python is concerned this is the only version that
  there can be. We could in theory probably reuse this in other
  code where we are parsing the Python interpreter version many
  times, but this is left for the future.

Fixes #2826
Work towards #2821


rickeylev (commented):
I factored a check_evaluation helper out.

Also added a test to just loop over (expr, want, env) cases. I figure we can copy/paste some real expressions. Also makes it easy to add some arbitrary expression to test.

rickeylev:
I tried fixing a bug with prefix matching exposed by the extra tests. I'm not entirely sure my revision is entirely correct. For fun, I ran the question through Gemini, it took like minutes (:astonished:) of thinking, and ultimately came back with something resembling:

```
if left and right are prefix: ...
elif left is prefix: ...
elif right is prefix: ...
else: ...
```

Which satisfies my confirmation bias, but I didn't look closely or think it through, so yeah, grain of salt.

Also, for version equality, the local part should be ignored, so I commented out that part of `_version_eq`

https://packaging.python.org/en/latest/specifications/version-specifiers/#id5

> Except where specifically noted below, local version identifiers MUST NOT be permitted in version specifiers, and local version labels MUST be ignored entirely when checking if candidate versions match a given version specifier.

I couldn't find an "except where noted" later, so I'm pretty sure the '+blabla` is supposed to be ignored.

aignas:
Oh nice, thank you, I will check it my tomorrow. I knew that the prefix matching is not fully done yet, but ran out of time yesterday.

aignas:
I realized that the prefix matching could be a simple string prefix match so I went that way, the code became easier to reason about and I think it needs to be optimized a little, because I am normalizing the version twice, but I think this direction is promising. Ran out of time today, so I'll leave it for the next time.

aignas:
It seems that this is still not done. I am not sure I can fully grok the spec about what we should do with the version ordering.

The exclusive ordered comparison
https://peps.python.org/pep-0440/#exclusive-ordered-comparison

is talking about `pre-release` versions, but I am not sure if that applies to the `dev` and `local` versions. That said, that part of the pep is about version specifiers and not necessarily comparison of two versions.

Maybe the ordering is answered in
https://peps.python.org/pep-0440/#exclusive-ordered-comparison

aignas:
There is a remaining test case that we need to fix and it is about ensuring that "3.12.0" is equivalent to "3.12" in the expressions so that the equality operator works.

python/private/version.bzl
```diff
@@ -503,17 +508,348 @@ def normalize_pep440(version):
...
+    )
+
+version = struct(
+    normalize = normalize_pep440,
+    parse = parse,
+    # methods, keep sorted
+    key = _version_key,
+    is_compatible = _version_compatible,
+    is_eq = _version_eq,
+    is_eeq = _version_eqq,
```
vonschultz, line 849: Shouldn't we use either `eeq` or `eqq`? Looks strange to use both.

python/private/version.bzl
rickeylev, line 1: Please put a suffix/prefix on this file name. We have like 4 files named "version.bzl" or "versions.bzl" and I find it confusing. parser_version? version_parser? pypa_version_parser?

python/private/version.bzl
```diff
@@ -503,17 +508,348 @@ def normalize_pep440(version):
...
+    and https://peps.python.org/pep-0440/
+
+    Args:
+      version_str: version string to be normalized according to PEP 440.
+      strict: fail if the version is invalid, defaults to True.
+
+    Returns:
+      string containing the normalized version.
+    """
+    version = version_str.strip()  # PEP 440: Leading and Trailing Whitespace
```
rickeylev, line 526: Instead of citing a pep (which are historical documents), link to the pypa specs page.

```suggestion
    # https://packaging.python.org/en/latest/specifications/version-specifiers/#leading-and-trailing-whitespace
    version = version_str.strip()
```

python/private/version.bzl
```diff
@@ -503,17 +508,348 @@ def normalize_pep440(version):
...
+
+        return None
+
+    parser_ctx["is_prefix"] = is_prefix
+    return parser_ctx
+
+def parse(version_str, strict = False):
+    """Parse a PEP4408 compliant version.
+
+    This is similar to `normalize_pep440`, but it parsers individual components to to
```
rickeylev, line 558:
```suggestion
    This is similar to `normalize_pep440`, but it parses individual components to
```

python/private/version.bzl
```diff
@@ -503,17 +508,348 @@ def normalize_pep440(version):
...
+
+    if left_release != right_release:
+        return False
+
+    return (
+        left.pre == right.pre and
+        left.post == right.post and
+        left.dev == right.dev
+        # local is ignored for == checks
+        ##and left.local == right.local
```
rickeylev, line 708:
```suggestion
```

rickeylev (approved):

rickeylev:
While I don't like the "version.bzl" file name, I odn't think it needs to block merging; enqueing.

PR #2857, opened 2025-05-04
fix: configure coverage helpers for test exec group

fmeum:
They are run on the test action's execution platform, which is resolved for the `test` exec group, not the default one.

### DIFF
--- a/python/private/attributes.bzl
+++ b/python/private/attributes.bzl
@@ -397,14 +397,14 @@ COVERAGE_ATTRS = {
     "_collect_cc_coverage": lambda: attrb.Label(
         default = "@bazel_tools//tools/test:collect_cc_coverage",
         executable = True,
-        cfg = "exec",
+        cfg = config.exec(exec_group = "test"),
     ),
     # Magic attribute to make coverage work. There's no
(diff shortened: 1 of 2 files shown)


aignas:
LGHM, but I lack the context behind the reason why we are fixing it. I am +1 for merging but love if you could point me to extra docs.

Does this fix a particular issue or does this just make things more tidy?

fmeum:
The docs don't exist yet, but I'm adding them in https://github.com/bazelbuild/bazel/pull/25996. We can link to that once it lands (and wait for it to get merged if you prefer that).

In a cross-platform build with custom exec constraints on either the `test` or default exec group, using the wrong exec group would potentially result in a failure to run the coverage tool (e.g. if the test runs on Linux, but "build" actions for the `py_test` build on macOS - this would try to run the tool in a macOS JDK on Linux).

rickeylev:
Would this fix https://github.com/bazel-contrib/rules_python/issues/2850 ? The error in that report seems related to the test toolchain

fmeum:
It wouldn't, but I will send a fix for that other issue.

rickeylev (approved):

PR #2862, opened 2025-05-06
fix: Make local runtime repo work on Windows

EthanSteinberg:
fix: Make local runtime repo work on windows

In particular, this fixes two problems:

First, windows requires interface libraries to be linked in.
Second, windows requires an alternative search path (LIBDEST) since LIBDIR is not filled in.


aignas:
Thanks for the PR! PRs improving support on Windows are always welcome. 

Could you please:
* Run `pre-commit run buildifier -a` to fix the buildifier failure in the CI.
* Add a `CHANGELOG.md` item in `Fixed` noting your contribution.

python/private/local_runtime_repo.bzl
```diff
@@ -112,13 +112,22 @@ def _local_runtime_repo_impl(rctx):
...
+
     # In some cases, the value may be empty. Not clear why.
     shared_lib_names = [v for v in shared_lib_names if v]
 
     # In some cases, the same value is returned for multiple keys. Not clear why.
     shared_lib_names = {v: None for v in shared_lib_names}.keys()
     shared_lib_dir = info["LIBDIR"]
 
+    if shared_lib_dir == None:
+        shared_lib_dir = info["LIBDEST"] + "/../libs"
```
rickeylev, line 129: nit: Instead of adding `../libs`, call dirname() and append `/libs`.

python/private/local_runtime_repo_setup.bzl
```diff
@@ -67,6 +82,10 @@ def define_local_runtime_toolchain_impl(
             allow_empty = True,
         ),
         includes = ["include"],
+        deps = select({
+            "@bazel_tools//src/conditions:windows": [":interface", ":abi3_interface"],
```
rickeylev, line 86:
```suggestion
            "@platforms//os:windows": [":interface", ":abi3_interface"],
```

rickeylev (approved):
Yeah, I didn't try any of this with Windows, so thanks for this,  @EthanSteinberg.

Mostly LGTM, but a few small changes requested.

re: adding a test: Yeah, a basic test for this would be good.

The place for the test would be in tests/integration/local_toolchains

Oddly, I don't see any windows exclusions for that? But, I would expect the windows CI to fail if it was running. So there must be a tag or something I'm missing.

In any case -- modifying tests/integration/local_toolchains/MODULE.bazel to have a windows section is the thing to do. You can cd into that directory and run bazel to improve your edit/run loop.

python/private/local_runtime_repo.bzl
```diff
@@ -112,13 +112,22 @@ def _local_runtime_repo_impl(rctx):
         info["INSTSONAME"],
     ]
 
+    if repo_utils.get_platforms_os_name(rctx) == 'windows':
+        shared_lib_names.append("python{major}{minor}.lib".format(**info))
+        shared_lib_names.append("python3.lib")
+
+        interpreter_path = interpreter_path.replace('\\', '/')
```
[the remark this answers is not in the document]
EthanSteinberg: From some investigation comparing a free threading and non-free threading install, it looks like abi_thread is the appropriate flag here. I added it as a first pass, will need to do more testing to make sure this works reliably.

python/private/local_runtime_repo.bzl
EthanSteinberg: Done.

EthanSteinberg:
@aignas @rickeylev Thanks for the suggestions. It's a good call to add a test. I think a test should be added before this is merged. I would be a bit worried about having this fragile functionality without a good CI test.

I have figured out why the current tests aren't working:

1. All integration tests are currently disabled on windows. See https://github.com/bazel-contrib/rules_python/blob/main/.bazelci/presubmit.yml#L149

2. The current local_runtime test does not verify that you can build a C++ Python extension and then load that extension. 

Let me look into both of the above this weekend.

rickeylev:
> integration tests are disabled for windows

Ah, yeah, i see now. Them being disabled in that particular block is WAI. We put the integration tests in their own job.

But yes, common_bazelinbazel_config defines the integration tests, and there are only ubuntu and debian jobs.

Now that the local toolchains can have arbitrary constraints added, we can get rid of that bazel-in-bazel test for local_toolchains. Instead, just make it part of our main config and use a custom flag to activate them (as the existing local_toolchains integration test does). stated another way: copy/paste what's in the local_toolchains into the main project.

> The current local_runtime test does not verify that you can build a C++ Python extension and then load that extension.

That'd be nice, but it's OK if that goes into another PR.

PR #2867, opened 2025-05-08
WIP add custom shebang on py_console_script_binary

chrisirhc:
This functionality came with wheel entry points, but if one uses py_console_script_binary, they're unable to use it with a shebang.

### DIFF
--- a/python/private/py_console_script_binary.bzl
+++ b/python/private/py_console_script_binary.bzl
@@ -52,6 +52,7 @@ def py_console_script_binary(
         entry_points_txt = None,
         script = None,
         binary_rule = py_binary,
+        shebang = None,
         **kwargs):
     """Generate a py_binary for a console_script entry_point.
 
@@ -68,6 +69,8 @@ def py_console_script_binary(
         binary_rule: {any}`rule callable`, The rule/macro to use to instantiate
             the target. It's expected to behave like {any}`py_binary`.
             Defaults to {any}`py_binary`.
+        shebang: [`str`], The shebang to use for the entry point python file.
+            Defaults to empty string.
         **kwargs: Extra parameters forwarded to `binary_rule`.
     """
     main = "rules_python_entry_point_{}.py".format(name)
@@ -81,6 +84,7 @@ def py_console_script_binary(
         out = main,
         console_script = script,
         console_script_guess = name,
+        shebang = shebang,
         visibility = ["//visibility:private"],
     )
 
--- a/python/private/py_console_script_gen.py
+++ b/python/private/py_console_script_gen.py
@@ -44,6 +44,8 @@ import textwrap
 _ENTRY_POINTS_TXT = "entry_points.txt"
 
 _TEMPLATE = """\
+{shebang}
+
 import sys
 
 # See @rules_python//python/private:py_console_script_gen.py for explanation
@@ -87,6 +89,7 @@ def run(
     out: pathlib.Path,
     console_script: str,
     console_script_guess: str,
+    shebang: str = "#!/usr/bin/env python3",
 ):
     """Run the generator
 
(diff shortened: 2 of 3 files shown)


python/private/py_console_script_gen.py
```diff
@@ -87,13 +89,16 @@ def run(
...
     console_script_guess: str,
+    shebang: str,
 ):
     """Run the generator
 
     Args:
         entry_points: The entry_points.txt file to be parsed.
         out: The output file.
         console_script: The console_script entry in the entry_points.txt file.
+        console_script_guess: The string used for guessing the console_script if it is not provided.
```
arrdem, line 100: Unused?

chrisirhc: Hm what do you mean?

chrisirhc:
> I don't fully understand the intention.
> 
> * Doesn't shebang doesn't ever get used because we have the bootstrap script?
> * If we put the actual python path, then it will be non-hermetic, if the shebang is an absolute path?
> * If we are using `/env/usr/bin python` then it is UNIX specific and non hermetic?

Use case: user is setting up the environment for a docker image, and needs a bash executable from the py_console_script (e.g. to run `ray` from command line). User is responsible of setting up the right paths. There's no change in default behavior per this diff. 
Previously, prior to Bazel mod, this was possible and simple through the use of `rules_python_wheel_entry_points` ([per here](https://github.com/bazel-contrib/rules_python/blob/9dfa3abba293488a9a1899832a340f7b44525cad/python/private/pypi/whl_library.bzl#L507)) but these are not reachable now via Bazel mod.

aignas:
Thanks for explaining this! If I understand correctly, maybe we could even replace the legacy `py` file generation with this `py_console_script_binary` usage, which would reduce the footprint of the `whl_library`, which in my books is always a good thing.

I'll have a look at the PR and will leave some comments. :)

python/private/py_console_script_binary.bzl
```diff
@@ -68,6 +69,8 @@ def py_console_script_binary(
         binary_rule: {any}`rule callable`, The rule/macro to use to instantiate
             the target. It's expected to behave like {any}`py_binary`.
             Defaults to {any}`py_binary`.
+        shebang: [`str`], The shebang to use for the entry point python file.
```
aignas, line 72:
```suggestion
        shebang: {type}`str`, The shebang to use for the entry point python file.
```

python/private/py_console_script_binary.bzl
```diff
@@ -52,6 +52,7 @@ def py_console_script_binary(
         entry_points_txt = None,
         script = None,
         binary_rule = py_binary,
+        shebang = None,
```
aignas, line 55:
```suggestion
        shebang = "",
```

Shouldn't this be a string and not `None`?

aignas (commented):
The approach in general LGTM.

chrisirhc:
> If I understand correctly, maybe we could even replace the legacy `py` file generation with this `py_console_script_binary` usage, which would reduce the footprint of the `whl_library`, which in my books is always a good thing.

Ah yes, so that users depend on `py_console_script_binary` instead of the whl_library's entrypoints. 👍  Yep!

chrisirhc:
Rebased to fix conflict with main branch.

docs/_includes/py_console_script_binary.md
```diff
@@ -48,6 +48,26 @@ py_console_script_binary(
...
+systems where the shebang line determines which interpreter is used to execute
+the script, per [PEP441]:
+
+```starlark
+load("@rules_python//python/entry_points:py_console_script_binary.bzl", "py_console_script_binary")
+
+py_console_script_binary(
+    name = "black",
+    pkg = "@pip//black",
+    interpreter = "/usr/bin/env python3",
```
aignas, line 63: Should this be the following instead?
```suggestion
    shebang = "#!/usr/bin/env python3",
```

Issue #2850, opened 2025-05-02
[Bazel CI] Error While resolving toolchains for target //tests/base_rules/py_test. No matching toolchains found for types.



sgowroji:
CI: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/4641#01968ee3-034a-44a1-9599-541d19ea7f77

 Platform: Ubuntu, MacOS, Debian

 Logs: 
```
ERROR: [0m/var/lib/buildkite-agent/builds/bk-docker-vfmd/bazel-org-repo-root/rules_python/tests/base_rules/py_test/BUILD.bazel:18:19: While resolving toolchains for target //tests/base_rules/py_test:test_basic_windows_subject (b27383e): No matching toolchains found for types:
ERROR: [0m/var/lib/buildkite-agent/builds/bk-docker-vfmd/bazel-org-repo-root/rules_python/tests/base_rules/py_test/BUILD.bazel:18:19: While resolving toolchains for target //tests/base_rules/py_test:test_mac_requires_darwin_for_execution_subject (8444979): No matching toolchains found for types:

``` 

 Steps:
  ``` 
git clone https://github.com/bazelbuild/rules_python
git reset ea80366b29219d3dad8c90191c21b77a4525875a  --hard
export USE_BAZEL_VERSION=0b5216ffe029b5fb0d0090f15fc4781522aac6d3
bazel build  ... @rules_python//examples/wheel/... 
 ``` 
 CC Greenteam  @meteorcloudy

rickeylev:
```
No matching toolchains found for types:
  @@bazel_tools//tools/test:default_test_toolchain_type: By default, tests are executed on the first execution platform that matches all constraints specified by the target platform. Either register the target platform (or a platform with at least the same constraints) as an execution platform, or override the default behavior by registering a custom toolchain for @bazel_tools//tools/test:default_test_toolchain_type.
```



We have some analysis tests that force the platform to windows/mac in order to test the windows code paths. Here's one of the failing ones mentioned in the error:

https://github.com/bazel-contrib/rules_python/blob/main/tests/base_rules/py_test/py_test_tests.bzl#L41

It sets config_settings:
```
 "//command_line_option:cpu": "darwin_x86_64",
 "//command_line_option:crosstool_top": CROSSTOOL_TOP,
 "//command_line_option:extra_toolchains": CC_TOOLCHAIN,
 "//command_line_option:platforms": [MAC_X86_64],
```

[...]

PR #2869, opened 2025-05-09
fix: add execution platforms to more analysis tests

fmeum:
Work towards #2850


fmeum:
@rickeylev @aignas I need help here. I am registering the test target platform as an execution platform to ensure compatibility with the new test toolchain, but that runs into an issue I'm planning to resolve in the future by https://github.com/bazelbuild/bazel/pull/26019: every `py_executable` has an exec dependency on a C++ toolchain through `launcher_maker`. 

In this particular case, I could work around this by transitioning to the host platform rather than the fixed platform LINUX_X86_64. Would that be acceptable?

rickeylev:
The launcher_maker is only needed for windows, so that dependency shouldn't be occurring for linux et al. I see it's refering to @bazel_tools//tools/launcher:launcher_maker directly, though. IIRC, bazel_tools has a select() to point that to something no-op. So it shouldn't still be requiring a C++ toolchain?

Would it help if we changed py_executable to point to an alias with select()?

```
alias(
    name = "launcher_maker",
    actual = select({
        "@platforms//os:windows": "@bazel_tools/tools/launcher:launcher_maker",
        "//conditions:default": "//python:none",
    })
```

py_executable has an optional dependency on the cc toolchain regardless, but I'm assuming that's OK.

fmeum:
Yes, it's only needed for Windows, but with the current setup within Bazel, every OS pays the price (the exec dep on a C++ toolchain). 

The selects (both the one that exists in Bazel and the one you suggest) apply after the exec transition and thus can't distinguish between a build that targets Windows and a build that doesn't. That's why I'm planning to introduce a toolchain for the launcher maker, which can match on both exec and target platform simultaneously.

However, this will require Bazel feature detection or a new BCR module for the launcher maker, so it won't help here (at least not soon).

fmeum:
I now realized that this wouldn't work on Windows hosts. What do you think of registering a fake C++ toolchain? I've done that on other analysis tests and it did work there.

rickeylev:
Sgtm. There are a couple fake cc tool chains in tests/support/cc_toolchain that might just work already

rickeylev:
I updated the title and description to better reflect the what/why -- can you please confirm/correct?

CI is passing after your recent change, so I think this is otherwise ready?

After the title/description is confirmed, we can queue to merge.

rickeylev (approved):

aignas:
Does this PR mean that users using `rules_python` will have to depend on a C++ toolchain eventually? Or will this limitation be fixed in the future (https://github.com/bazelbuild/bazel/pull/26019)?

fmeum:
I tweaked this to not require a fake C++ toolchain. The dependency isn't new and the complications around extra execution platforms already affect Python targets today. I will push on the PR to get this improved in Bazel 8.x and 9.x.

PR #2872, opened 2025-05-11
feat(uv): handle credential helpers and .netrc

aignas:
This allows one to download the uv binaries from private mirrors.

The plumbing of the auth attrs allows us to correctly use the `~/.netrc`
or the credential helper for downloading from mirrors that require
authentication.

Testing notes:
* When I tested this, it seems that the dist manifest json may not work
  with private mirrors, but I think it is fine for users in such cases
  to define the `uv` srcs using the `urls` attribute.

Work towards #1975.


aignas:
Marking as draft because of what I've found during testing. This may need some extra polishing before it is ready.

rickeylev (approved):

PR #2873, opened 2025-05-11
feat(pypi): actually start using env_marker_setting

aignas:
Summary:
- `pep508_deps` is now much simpler, because the hard work is done in
  analysis phase
- `whl_library` BUILD.bazel tests now also have a test for the legacy
  flow.

One thing that I noticed is that now we have an implicit dependency on
the python toolchain when getting all of the `whl` target tree. This is
a filegroup target that includes dependent wheels. However, we fallback
to the flag values if we don't have the toolchain, so we should be good
in general.

Overall I like how this is turning out because we don't need to pipe the
`target_platforms` anymore when we enable `PIPSTAR` feature. This means
that we can start creating fewer whl_library instances - e.g. a
`py3-none-any` wheel can be fetched once instead of once per python
interpreter version. I'll leave this optimization for a later time.

Work towards #260


python/private/pypi/generate_whl_library_build_bazel.bzl
```diff
@@ -62,28 +63,44 @@ def generate_whl_library_build_bazel(
...
+            "default_python_version",
         ]
+        dep_template = kwargs.get("dep_template")
+        loads.append(
+            """load("{}", "{}")""".format(
+                dep_template.format(
+                    name = "",
+                    target = "requirements.bzl",
+                ),
+                "all_whl_requirements_by_package",
```
aignas, line 92: TODO: check if this will contain all of the packages or only the ones that are present on all platforms.

We may need to re-include the `config.bzl.tmpl.bzlmod` and put the list of all packages there.

aignas:
Marking as draft due to the TODO note.

CHANGELOG.md
```diff
@@ -97,6 +97,7 @@ END_UNRELEASED_TEMPLATE
 * (pypi) Starlark-based evaluation of environment markers (requirements.txt conditionals)
   available (not enabled by default) for improved multi-platform build support.
   Set the `RULES_PYTHON_ENABLE_PIPSTAR=1` environment variable to enable it.
+  Work towards [#260](https://github.com/bazel-contrib/rules_python/issues/260).
```
rickeylev, line 100:
```suggestion
```

The prior entry about pipstar being experimentally available covers this.

python/private/pypi/pep508_deps.bzl
```diff
@@ -207,37 +149,24 @@ def _resolve_extras(self_name, reqs, extras):
...
+    markers = {}
+    for req in reqs:
+        for x in extras:
+            m = evaluate(req.marker, env = {"extra": x}, strict = False)
+            if m == False:
+                continue
+            elif m == True:
+                _add(deps, deps_select, dep)
                 break
+            else:
```
rickeylev, line 167: I thought evaluate() was only true/false. What's the third return value? I'm not sure env_marker_setting is handling a third return value.

python/private/pypi/whl_library_targets.bzl
```diff
@@ -362,9 +353,19 @@ def _config_settings(dependencies_by_platform, native = native, **kwargs):
...
             * `cp3{minor_version}_{os}_{cpu}`
+        dependencies_with_markers: {type}`dict[str, str]` The markers to evaluate by
+            each dep.
+        rules: used for testing
         native: {type}`native` The native struct for overriding in tests.
         **kwargs: Extra kwargs to pass to the rule.
     """
+    for dep, expression in dependencies_with_markers.items():
+        rules.env_marker_setting(
+            name = "include_{}".format(dep),
```
rickeylev, line 364: I was expecting to a list of extras passed in. But I guess the extras are being handled earlier?

The case I had in mind was a more complicated boolean expression e.g. `(extra == foo and python_version ~= 3.10) or (extra = bar and python_version ~= 3.9)`; something that requires the target config to be evaluated.

That seems rare, though? And can probably be ignored?

rickeylev (approved):

python/private/pypi/pep508_deps.bzl
aignas: If `strict = False` then it will output `False | True` or the remaining expression that we did not have `env` keys to evaluate.

python/private/pypi/whl_library_targets.bzl
aignas: Let me think about it a little more.

aignas: Realized that this should be still fine as is, at least for your example. Leaving the comment unresolved in case something comes along.

PR #2875, opened 2025-05-13
refactor: make python extension generate platform toolchains

rickeylev:
This makes the python bzlmod extension handle generating the platform-specific toolchain
entries ("python_3_10_{platform}"). This is to eventually allow adding additional
toolchains that aren't part of the PLATFORMS mapping in versions.bzl and have
their own custom constraints.

The main things this refactor does are:
1. The bzlmod phase passes the full list of implementation toolchains
   to create (previously, it relied on `hub_repo` to generate the implementation
   names).
2. The name of a toolchain (the toolchain.name arg) and the repo that implements
   it (the py_toolchain_suite.user_repository_repo arg) are separate. This allows
   future work to mixin toolchains that point to arbitrary repos.
3. The platform meta data uses a list of target settings instead of dict of
   flag values. This allows more arbitrary target settings. For now, flag values
   on the platform metadata is still looked for because it's known that users
   patch python/versions.bzl.

Along the way:
* Factor out a platform_info helper in versions.bzl
* Factor out a NOT_ACTUALLY_PUBLIC constants to better denote things that
  are public visibility, but actually internal.
* Add some docs to some internals so we don't have to chase down their definitions.


rickeylev:
I want to cross check the generated BUILD files for toolchains to make sure they're the same semantically, so moving back to draft until I do so. From a cursory look, they look OK, but I changed the order of args, prefix length, etc, so the diff is pretty big. Gonna back those out to make it easier to compare them.

rickeylev:
Manually verified the registrations: they're OK. I tweaked the code a bit so it better matches the original ordering and naming. Ready for review.

aignas (approved):
Outside the scope of this PR, but I would love at some point to start registering the platform values in bazel-mod in the same way that they are done for uv.

PR #2877, opened 2025-05-14
feat: Output the doctrees directory

kaycebasques:
Allows non-hermetic docs builds (e.g. `bazelisk run //docs:docs.run`) to leverage Sphinx's stateful incremental build system. The key is to expose the `doctrees` output artifact alongside the built docs, and then use the `--doctree-dir` flag to notify Sphinx of the existence of the incremental build metadata. This also requires removing the `--fresh-env` and `--write-all`. IMO rules_python should not be forcing those options on all users anyways. When docs are built hermetically, Sphinx has to build everything from scratch anyways. There's no need to explicitly set the flags. Users can set those flags downstream in their own projects, via the `extra_opts` list in their `sphinx_docs()` target.

### DIFF
--- a/sphinxdocs/private/sphinx.bzl
+++ b/sphinxdocs/private/sphinx.bzl
@@ -203,15 +203,19 @@ def _sphinx_docs_impl(ctx):
     per_format_args = {}
     outputs = {}
     for format in ctx.attr.formats:
+        doctree_dir = ctx.actions.declare_directory(paths.join(ctx.label.name, "_build", format + "_doctrees"))
         output_dir, args_env = _run_sphinx(
             ctx = ctx,
             format = format,
             source_path = source_dir_path,
             output_prefix = paths.join(ctx.label.name, "_build"),
             inputs = inputs,
+            doctree_dir = doctree_dir,
         )
         outputs[format] = output_dir
+        outputs[format + "_doctrees"] = doctree_dir
         per_format_args[format] = args_env
+
     return [
         DefaultInfo(files = depset(outputs.values())),
         OutputGroupInfo(**{
@@ -254,7 +258,7 @@ _sphinx_docs = rule(
     },
 )
 
-def _run_sphinx(ctx, format, source_path, inputs, output_prefix):
+def _run_sphinx(ctx, format, source_path, inputs, output_prefix, doctree_dir):
     output_dir = ctx.actions.declare_directory(paths.join(output_prefix, format))
 
     run_args = []  # Copy of the args to forward along to debug runner
@@ -272,10 +276,9 @@ def _run_sphinx(ctx, format, source_path, inputs, output_prefix):
     # Build in parallel, if possible
     # Don't add to run_args: parallel building breaks interactive debugging
     args.add("--jobs", "auto")
-    args.add("--fresh-env")  # Don't try to use cache files. Bazel can't make use of them.
-    run_args.append("--fresh-env")
-    args.add("--write-all")  # Write all files; don't try to detect "changed" files
-    run_args.append("--write-all")
+
+    args.add("--doctree-dir", doctree_dir.path)
+    run_args.extend(("--doctree-dir", doctree_dir.path))
 
     for opt in ctx.attr.extra_opts:
         expanded = ctx.expand_location(opt)
@@ -303,7 +306,7 @@ def _run_sphinx(ctx, format, source_path, inputs, output_prefix):
         executable = ctx.executable.sphinx,
         arguments = [args],
         inputs = inputs,
-        outputs = [output_dir],
+        outputs = [output_dir, doctree_dir],
         tools = tools,
         mnemonic = "SphinxBuildDocs",
         progress_message = "Sphinx building {} for %{{label}}".format(format),


kaycebasques:
@rickeylev

rickeylev:
The docs say https://www.sphinx-doc.org/en/master/man/sphinx-build.html#cmdoption-sphinx-build-d

> Normally, these files are put in a directory called .doctrees under the build directory

So I would expect the `{name}_build` output directory to have the .doctrees directory within it. Are you seeing otherwise? Is it because fresh-env/write-all only apply to the build output directory, and the doctrees, if put elsewhere, are able to persist despite those being set?

I do like that it says the output can be shared between builders -- I wonder if that could be leveraged in a future change.

> Remove fresh-env, write-all

I don't recall why I added those. I can't remember if it was just defensive programming, or if ran across a real issue. Does `--no-fresh-env` or `--fresh-env=false` work? If so, you could try adding that to extra args. If not (annoying), then we should add a remove_opts attribute to deal with those.

Removing fresh-env is possible if Sphinx is using hash-based change detection. If it's using timestamp detection, then it'll act weird because bazel won't reliably set timestamps on files. I don't see a mention in the docs about what strategy it uses, though. I wonder if a plugin could address this if its using timestamps.

Removing write-all: This might be necessary because (i'm pretty sure) bazel clears the output directory every invocation, but if sphinx thinks a file doesn't need to be written (fresh-env=false), it won't write it. And then the output dir is missing files.

kaycebasques:
Re: the location of the `.doctrees` for, I believe it's a sibling of each builder's output dir. E.g. you get `_build/html` and `_build/.doctrees`, not `_build/html/.doctrees`. Let me double-check when I get into the office.

kaycebasques:
Current status is that @rickeylev is helping me investigate whether we can use persistent workers (and a few other things) to get incremental builds working within Bazel's hermetic worldview. Let's leave this PR open for a bit but it may end up getting closed without merge.

kaycebasques:
Following up on the previous discussion of where the doctrees directory is output by default. It's output as a sibling directory, not a child of the build format that you're running.

Ran `sphinx-quickstart` to verify:

```
kayce@kayce0 /t/doctrees [2]> sphinx-quickstart
Welcome to the Sphinx 8.0.2 quickstart utility.

Please enter values for the following settings (just press Enter to
accept a default value, if one is given in brackets).

Selected root path: .

You have two options for placing the build directory for Sphinx output.
Either, you use a directory "_build" within the root path, or you separate
"source" and "build" directories within the root path.
> Separate source and build directories (y/n) [n]: 

The project name will occur in several places in the built documentation.
> Project name: 
* Please enter some text.
> Project name: doctrees
> Author name(s): doctrees
> Project release []: 0.0.0

If the documents are to be written in a language other than English,
you can select a language here by its language code. Sphinx will then
translate text that it generates into that language.

For a list of supported codes, see
https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-language.
> Project language [en]: 

Creating file /tmp/doctrees/conf.py.
Creating file /tmp/doctrees/index.rst.
Creating file /tmp/doctrees/Makefile.
Creating file /tmp/doctrees/make.bat.

Finished: An initial directory structure has been created.

You should now populate your master file /tmp/doctrees/index.rst and create other documentation
source files. Use the Makefile to build the docs, like so:
   make builder
where "builder" is one of the supported builders, e.g. html, latex or linkcheck.

kayce@kayce0 /t/doctrees> make html
Running Sphinx v8.0.2
loading translations [en]... done
making output directory... done
building [mo]: targets for 0 po files that are out of date
writing output... 
building [html]: targets for 1 source files that are out of date
updating environment: [new config] 1 added, 0 changed, 0 removed
reading sources... [100%] index
looking for now-outdated files... none found
pickling environment... done
checking consistency... done
preparing documents... done
copying assets... 
copying static files... done
copying extra files... done
copying assets: done
writing output... [100%] index
generating indices... genindex done
writing additional pages... search done
dumping search index in English (code: en)... done
dumping object inventory... done
build succeeded.

The HTML pages are in _build/html.
kayce@kayce0 /t/doctrees> ls _build
doctrees/  html/
kayce@kayce0 /t/doctrees> ls _build/doctrees
environment.pickle  index.doctree
```

rickeylev:
> remove fresh-env, write-all

Ahh, I see -- those are also being added to the docs.run invocation. I'd be OK removing those from the docs.run invocation. I added those because, without them, you can get weird results if files are removed or the impl of a plugin changes (which can effect the doctree pickle). docs.run isn't intended to be _super_ correct, though; it's primarily just to facilitate direct invocation for whatever reason. It's easy enough to just manually add those args yourself when calling docs.run.

> default location of doctrees is under build/; a sibling of output format dirs
> If the .doctrees dir is shared among all builders then that location makes sense to me.

Thanks for checking.

From a strict bazel perspective, sharing the doctree between the different builders isn't possible. What it basically means is two different actions are both trying to create the same output, which will be an action conflict. In order for them to share, doctree creation would have to be its own action, and that output is then fed into another action that does the output-format building.

Hm, is there a sphinx command to _only_ produce doctree output? I don't see any...maybe a plugin that is a no-op builder backend? Oh hmmm, a plugin to basically copy the doctree files to an output looks fairly easy.

---

I'm not sure if this PR will quite accomplish what you're after, though? It'll give partial incrementality to docs.run, but not fully, and only after a regular `bazel build :docs`.

Yes, this materializes the doctree directory, and docs.run can be pointed to it, however, the doctree directory and its files aren't writable. I didn't get any sphinx errors, though, so sphinx must be recreating the doctree in memory and just ignoring a failed write of the updated doctree file. Because the `run` target doesn't take the doctrees File objects as an inputs (WAI), it won't build those files, so a clean build will have zero doctrees. Since the doctree directory isn't writable, the ones created as part of this don't end up cached.

That'll probably work well enough if you're just modifying a few files, though.

I double checked and `bazel run :docs.run` will create a `.doctrees` in `/tmp/sphinx-out/.doctrees (just wanted to make sure some elsewhere wasn't messing that up)

Another thing is Bazel will entirely clear the declare_directory() output artifact. i.e. changing any source file means all doctrees get deleted. This wouldn't affect `bazel run docs.run` (because the docs.run target doesn't depend on the File objets, just their paths), but would affect the doctree incrementality of `bazel build docs` (even if fresh-env and write-all were removed). So, if you're doing `bazel run :docs.server & ; ibazel build :docs`, then it doesn't help -- the doctrees dir gets nuked every build invocation.

---

Otherwise, my only comment is about putting the doctrees in output groups:

Do they need to be in the output groups? Are you manually accessing them later and wiring them into something else? The docs.run command can still have their paths added to argv even if they're not in the output groups (they're in bazel-out)

I'd OK enough with the format-specific output groups, but don't want them as part of the default outputs. e.g. if i have `formats=html,epub`, then `bazel build :docs` should produce two outputs: the html and epub output directories. Right now, it'll produce 4 (html, html_doctrees, epub, epub_doctrees). i.e. make it so `DefaultInfo(files=...)` doesn't include the `_doctrees` groups.

rickeylev:
> Hm, is there a sphinx command to only produce doctree output? I don't see any...maybe a plugin that is a no-op builder backend? Oh hmmm, a plugin to basically copy the doctree files to an output looks fairly easy.

Ohhhh, wait, i thnk there is! The dummy builder: https://www.sphinx-doc.org/en/master/usage/builders/index.html#sphinx.builders.dummy.DummyBuilder

I think this would work:

```

doctree = <declare dir>
run(sphinx --format=dummy, outputs=[doctree])
for f in formats:
  run(sphinx --format=$ --doctree-dir=$doctree, ...)
```

kaycebasques:
Yes, I don't think this PR as it currently stands is the solution. I'm working on the persistent worker approach but I've still got a lot to figure out.

I will keep the `dummy` builder in mind, thanks for digging that up.

Let's close this PR and I'll start a general issue for all work related to supporting incremental builds.

PR #2880, opened 2025-05-14
feat: allow custom platform when overriding

rickeylev:
This basically allows using any python-build-standalone archive and using it
if custom flags are set. This is done through the `single_version_platform_override()`
API, because such archives are inherently version and platform specific.

Key changes:
* The `platform` arg can be any value (mostly; it ends up in repo names)
* Added `target_compatible_with` and `target_settings` args, which become the
  settings used on the generated toolchain() definition.

The platform settings are version specific, i.e. the key `(python_version, platform)`
is what maps to the TCW/TS values.

If an existing platform is used, it'll override the defaults that normally come
from the PLATFORMS global for the particular version. If a new platform is used,
it creates a new platform entry with those settings.

Along the way:
* Added various docs about internal variables so they're easier to grok at a glance.

### DIFF
--- a/docs/toolchains.md
+++ b/docs/toolchains.md
@@ -236,6 +236,79 @@ existing attributes:
 * Adding additional Python versions via {bzl:obj}`python.single_version_override` or
   {bzl:obj}`python.single_version_platform_override`.
 
+### Registering custom runtimes
+
+Because the python-build-standalone project has _thousands_ of prebuilt runtimes
+available, rules_python only includes popular runtimes in its built in
+configurations. If you want to use a runtime that isn't already known to
+rules_python then {obj}`single_version_platform_override()` can be used to do
+so. In short, it allows specifying an arbitrary URL and using custom flags
+to control when a runtime is used.
+
+In the example below, we register a particular python-build-standalone runtime
+that is activated for Linux x86 builds when the custom flag
+`--//:runtime=my-custom-runtime` is set.
+
+```
+# File: MODULE.bazel
+bazel_dep(name = "bazel_skylib", version = "1.7.1.")
+bazel_dep(name = "rules_python", version = "1.5.0")
+python = use_extension("@rules_python//python/extensions:python.bzl", "python")
+python.toolchain(python_version="3.13.3")
+python.single_version_platform_override(
+    platform = "my-platform",
+    python_version = "3.13.3",
+    sha256 = "01d08b9bc8a96698b9d64c2fc26da4ecc4fa9e708ce0a34fb88f11ab7e552cbd",
+    target_compatible_with = [
+        "@platforms//os:linux",
+        "@platforms//cpu:x86_64",
+    ],
+    target_settings = [
+        "@@//:runtime=my-custom-runtime",
+    ],
+    urls = ["https://github.com/astral-sh/python-build-standalone/releases/download/20250409/cpython-3.13.3+20250409-x86_64-unknown-linux-gnu-install_only_stripped.tar.gz"],
+)
+
+# File: //:BUILD.bazel
+
+load("@bazel_skylib//rules:common_settings.bzl", "string_flag")
+string_flag(
+    name = "custom_runtime",
+    build_setting_default = "",
+)
+
+config_setting(
+    name = "is_custom_runtime_linux-x86-install-only-stripped",
+    flag_values = {
+        ":custom_runtime": "linux-x86-install-only-stripped",
+    },
+)
+```
+
+Notes:
+- While any URL and archive can be used, it's assumed their content looks how
+  a python-build-standalone archive looks.
+- `python.toolchain()` is required if the version is unknown; if the version
+  is already known, it can be omitted.
+- A "version aware" toolchain is registered, which means the Python version flag
+  must also match (e.g. `--@rules_python//python/config_settings:python_version=3.13.3`
+  must be set -- see `minor_mapping` and `is_default` for controls and docs
+  about version matching and selection).
+- The labels in `target_settings` must be absolute; `@@` refers to the main repo.
+- The `target_settings` are `config_setting` targets, which means you can
+  customize how matching occurs.
+
+:::{seealso}
+See {obj}`//python/config_settings` for flags rules_python already defines
+that can be used with `target_settings`. Some particular ones of note are:
+{flag}`--py_linux_libc` and {flag}`--py_freethreaded`, among others.
+:::
+
+:::{versionadded} VERSION_NEXT_FEATURE
+Added support for custom platform names, `target_compatible_with`, and
+`target_settings` with `single_version_platform_override`.
+:::
+
 ### Using defined toolchains from WORKSPACE
 
 It is possible to use toolchains defined in `MODULE.bazel` in `WORKSPACE`. For example
--- a/python/private/platform_info.bzl
+++ b/python/private/platform_info.bzl
@@ -1,0 +1,34 @@
+"""Helper to define a struct used to define platform metadata."""
+
+def platform_info(
+        *,
+        compatible_with = [],
+        flag_values = {},
+        target_settings = [],
+        os_name,
+        arch):
+    """Creates a struct of platform metadata.
+
+    This is just a helper to ensure structs are created the same and
+    the meaning/values are documented.
+
+    Args:
+        compatible_with: list[str], where the values are string labels. These
+            are the target_compatible_with values to use with the toolchain
+        flag_values: dict[str|Label, Any] of config_setting.flag_values
+            compatible values. DEPRECATED -- use target_settings instead
+        target_settings: list[str], where the values are string labels. These
+            are the target_settings values to use with the toolchain.
+        os_name: str, the os name; must match the name used in `@platfroms//os`
+        arch: str, the cpu name; must match the name used in `@platforms//cpu`
+
+    Returns:
+        A struct with attributes and values matching the args.
+    """
+    return struct(
+        compatible_with = compatible_with,
+        flag_values = flag_values,
+        target_settings = target_settings,
+        os_name = os_name,
+        arch = arch,
+    )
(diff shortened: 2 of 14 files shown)


python/private/platform_info.bzl
```diff
@@ -0,0 +1,34 @@
...
+    the meaning/values are documented.
+
+    Args:
+        compatible_with: list[str], where the values are string labels. These
+            are the target_compatible_with values to use with the toolchain
+        flag_values: dict[str|Label, Any] of config_setting.flag_values
+            compatible values. DEPRECATED -- use target_settings instead
+        target_settings: list[str], where the values are string labels. These
+            are the target_settings values to use with the toolchain.
+        os_name: str, the os name; must match the name used in `@platfroms//os`
```
aignas, line 22: The arch and os could be retrieved from constraints list on one hand, on the other hand we could cconstruct the constraints from the os and arch.

python/private/python.bzl
```diff
@@ -288,6 +295,7 @@ def _python_impl(module_ctx):
         loaded_platforms[full_python_version] = python_register_toolchains(
             name = toolchain_info.name,
             _internal_bzlmod_toolchain_call = True,
+            platforms = PLATFORMS | py.config.platform_overrides.get(full_python_version, {}),
```
aignas, line 298: I would prefer using the PLATFORMS global once at the creation of py.config so that we could get rid of the global variable altogether at a later time and we would not need to change the interfaces.

python/private/python_repository.bzl
```diff
@@ -15,7 +15,7 @@
 """This file contains repository rules and macros to support toolchain registration.
 """
 
-load("//python:versions.bzl", "FREETHREADED", "INSTALL_ONLY", "PLATFORMS")
+load("//python:versions.bzl", "FREETHREADED", "INSTALL_ONLY")
```
aignas, line 18: It is so nice to see that gone, but Workspace may still need it?

aignas (approved):

python/private/python.bzl
rickeylev: I had the same thought, but had trouble figuring out how to wire it that way. Did you have something in mind? Do you just mean putting the global on `py.config` (where "defaults" and "platform_overrides" are assigned, in get_toolchain_config)? And then this line changes to `platforms = py.config.platforms | py.config.platform_overrides[version]` ?

python/private/platform_info.bzl
rickeylev: Looking at CI, we might have to change the tag class to accept an os_name and arch value, which I don't really like. I like the simplicity of a list of opaque labels. It is a simple list, though, so we could just look for the common @platforms values and infer them.

The problem CI is showing is, when `{name}_host` goes to be generated, it looks at the global PLATFORMS variable. That's easy to fix, but then it wants to look at the os_name and arch fields of that platform metadata. For these "ad hoc" platforms, those values weren't set, so then it blows up.

I'm going to try and see if I can make it skip generating a host repo for now. I'm still tring to figure out this host-repo code, because it seems like it creates a host repo for every registered version (full version).

python/private/python_repository.bzl
rickeylev: I don't follow -- the only usage of PLATFORMS in this file was on the attribute values. It isn't otherwise used.

rickeylev:
Notes to self:

The basic problem is it's trying to register the 3.13.3 linux-x86-install-stripped runtime as the host runtime. I don't want this to happen, but its pretty insistent. It does this because python.bzl computes a minor_mapping from seen versions and uses the highest patch version. Because the custom 3.13.3 is the highest 3.13 version, it starts treating 3.13 as 3.13.3, which only has the linux-x86-install-stripped runtime.

Thus, when it goes to create the python_3_13_host repo, it tries to use linux-x86-install-stripped. This fails because there's no platform for 3.13.3 that says its compatible with the host platform. That isn't strictly true in this particular case (my host is linux x86), but that's an easy case to make true (Ignas's host is mac, for example). Requiring that I register a runtime for every host my devs use doesn't make sense either -- the purpose of this feature is to add more esoteric versions and platforms; it's likely it'll be a newer python version and/or i don't care that every dev's host can use it.

So, I think the thing to do is change the host-repo creation to go looking a bit harder for a runtime that is compatible with the host os. Move the host_repo() creation out of python_register_toolchains (for bzlmod), and have python.bzl find something to call host_repo() with.

python/private/platform_info.bzl
aignas: Yeah, this is for the pip integration only. (Some users may have some custom repo rules that also rely on this).

python/private/python.bzl
aignas: Yeah I was thinking about something like that. This is very similar how the `uv` version specific things get.

aignas:
Regarding the `minor_mapping`, you can explicitly tell the `python` thing to set a different `minor_mapping` value, I think.

rickeylev:
> manually set minor_mapping

I'll give that a try and see what happens, though I'd like things to work without having to do that.

What's really irking me is that I can't register a runtime unless it has host-compatible variants defined, even if it never gets used in a repository context.

rickeylev:
I'm going to move this back to draft for now. I'm splitting out some separate PRs to refactor the internals that should help this PR and the `python.configure()` style API we talked about.

python/private/python.bzl
rickeylev: Done

rickeylev:
PTAL

tldr of notable changes since last review:

There's a second loop to find a host-compatible runtime if one wasn't created earlier (when python_register_toolchains is calle). Mainly this handles the case where adding a runtime (but only for a subset of platforms) with a higher version causes `@python_3_13_host` to try and use that higher version -- but a host compatible variant doesn't exist. To fix, keep track of all the host-compatible runtimes for the versions and fallback to one of them.

The runtimes within a version get registered in the order of the platform keys. This ends up mattering when custom flags are used for a runtime -- if the regular runtime comes first, it matches (same platform), and the one with custom flags doesn't get looked at. To fix, I made overrides get added as the first key in the platforms dict. (We should transform the map of platforms to a list or something, but something for another PR).

It's now possible to register a runtime and have no host-compatible version available. Thanks for the interpreters.bzl tip -- that was the spot that kept trying to reference things that didn't exist.

Some other misc changes:
* Fixed a bug where overrides (of strip_prefix or urls) would try to modify an immutable dict -- `default["tool_versions"]` had an immutable copy from the global.
* host_compatible_python_repo now takes the backing repo names as inputs (instead of making
  assumptions about the platform being the backing repo name)
* single_version_platform_override accepts os_name/arch args. These populate target_compatible_with (if TCW is empty)

python/private/python.bzl
```diff
@@ -318,27 +357,80 @@ def _python_impl(module_ctx):
...
+                    platform = platform_info,
+                    platform_name = platform_name,
+                    impl_repo_name = repo_name,
+                )
+                host_platforms[platform_name] = host_compat_entry
+                all_host_compatible_impls.setdefault(full_python_version, []).append(
+                    host_compat_entry,
+                )
+                all_host_compatible_impls.setdefault(
+                    full_python_version.rpartition(".")[0],
```
aignas, line 371: nit, using the `version.bzl` to parse the `full_python_version` and then using the `major.minor` from that could be more robust.

But it's fine for this PR as maybe my preference to parse the version here is non warranted.

python/private/python.bzl
```diff
@@ -318,27 +357,80 @@ def _python_impl(module_ctx):
...
+                    [],
+                ).append(host_compat_entry)
+
+        host_repo_name = toolchain_info.name + "_host"
+        if not host_platforms:
```
aignas, line 376: nit: since this has `else` clause, is inverted if easier to read?

aignas (approved):
LGTM

docs/toolchains.md
```diff
@@ -243,6 +243,73 @@ existing attributes:
...
+:::{seealso}
+See {obj}`//python/config_settings` for flags rules_python already defines
+that can be used with `target_settings`. Some particular ones of note are:
+{flag}`--py_linux_libc` and {flag}`--py_freethreaded`, among others.
+:::
+
+:::{versionadded} VERSION_NEXT_FEATURE
+Added support for custom platform names, `target_compatible_with`, and
+`target_settings` with `single_version_platform_override`.
+:::
```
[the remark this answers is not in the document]
rickeylev: Even handier: sphinx has a builder that extract these blocks: https://www.sphinx-doc.org/en/master/usage/builders/index.html#sphinx.builders.changes.ChangesBuilder

python/private/python.bzl
```diff
@@ -265,19 +278,42 @@ def parse_modules(*, module_ctx, _fail = fail):
...
     toolchain_impls = []
 
     # list[str] of the base names of toolchain repos
-    base_toolchain_repo_names = []
+    # list[str] of the repo names for host compatible repos
+    all_host_compatible_repo_names = []
+
+    # todo: The 3.13.1-custom-runtime ends up _last_ in the toolchain
+    # ordering, so it never gets picked.
+    # Need to move it first
```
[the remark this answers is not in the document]
rickeylev: Oops, left over note-to-self; removed. This was fixed by the platforms-map-rebuilding in process_single_Version_platform_override

python/private/python.bzl
rickeylev: Done, though I had to `v.release[0:2]` since there's no .major or .minor attributes of the parsed object.

python/private/python.bzl
rickeylev: Done

python/private/toolchains_repo.bzl
```diff
@@ -380,26 +381,76 @@ def _host_compatible_python_repo(rctx):
...
+            doc = """
+Full python version, Major.Minor.Micro.
+
+Only set in workspace calls.
+""",
+        ),
+        "python_versions": attr.string_dict(
+            doc = """
+If set, the Python version for the corresponding selected platform. Values in
+Major.Minor.Patch format. Keyed by index in `platforms`.
```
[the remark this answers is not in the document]
rickeylev: Done

aignas (approved):

PR #2884, opened 2025-05-16
refactor: have bzlmod pass platforms to python_register_toolchains

rickeylev:
This is to facilitate eventually allowing overrides to add additional platforms.

Instead of the PLATFORMS global being used, the python bzlmod extension passing the
mapping directly to python_register_toolchains, then receives back the subset of
platforms that had repos defined. That subset is then later used when (re)constructing
the list of repo names for the toolchains.

### DIFF
--- a/tests/python/python_tests.bzl
+++ b/tests/python/python_tests.bzl
@@ -149,6 +149,7 @@ def _test_default(env):
         "base_url",
         "ignore_root_user_error",
         "tool_versions",
+        "platforms",
     ])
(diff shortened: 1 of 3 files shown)


tests/python/python_tests.bzl
```diff
@@ -149,6 +149,7 @@ def _test_default(env):
         "base_url",
         "ignore_root_user_error",
         "tool_versions",
+        "platforms",
```
aignas, line 152: Ah, so this does not expose any extra APIs yet, so hence no extra tests here, LGTM.

aignas (approved):

PR #2885, opened 2025-05-16
refactor: make bzlmod directly aware of created toolchain repo names

rickeylev:
This makes python_register_toolchains return the repo names it created, which allows
the bzlmod code to be directly aware of the repos that were created instead of having
to rely on assuming the names via the platform keys.

This is to facilitate python_register_toolchains creating a more arbitrary subset
of platform-specific repos.

### DIFF
--- a/python/private/python.bzl
+++ b/python/private/python.bzl
@@ -267,11 +267,18 @@ def parse_modules(*, module_ctx, _fail = fail):
 def _python_impl(module_ctx):
     py = parse_modules(module_ctx = module_ctx)
 
-    # dict[str version, list[str] platforms]; where version is full
-    # python version string ("3.4.5"), and platforms are keys from
-    # the PLATFORMS global.
-    loaded_platforms = {}
-    for toolchain_info in py.toolchains:
+    # list of structs; see inline struct call within the loop below.
+    toolchain_impls = []
+
+    # list[str] of the base names of toolchain repos
+    base_toolchain_repo_names = []
+
+    # Create the underlying python_repository repos that contain the
+    # python runtimes and their toolchain implementation definitions.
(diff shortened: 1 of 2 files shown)


rickeylev:
Ready for PTAL; dependency PR merged

python/private/python.bzl
```diff
@@ -267,11 +267,18 @@ def parse_modules(*, module_ctx, _fail = fail):
...
+    # list of structs; see inline struct call within the loop below.
+    toolchain_impls = []
+
+    # list[str] of the base names of toolchain repos
+    base_toolchain_repo_names = []
+
+    # Create the underlying python_repository repos that contain the
+    # python runtimes and their toolchain implementation definitions.
+    for i, toolchain_info in enumerate(py.toolchains):
+        is_last = (i + 1) == len(py.toolchains)
```
aignas, line 279:
```suggestion
    all_toolchains = zip(
        [False] * len(py.toolchains),
        py.toolchains,
    )[:-1]
    all_toolchains.append(
        (True, py.toolchains[-1])
    )
    for set_python_toolchain_constraint, toolchain_info in all_toolchains:
```

You could do this, I am curious what is more readable.

Also, I can't remember if the last toolchain is still special because we always know the `python_version` flag setting, so the comment/logic may be outdated. At least when I was bugfixing code in this location I think that the case when the `python_version` does not match the preceding entries never happens in bzlmod.

aignas (approved):
LGTM

python/private/python.bzl
rickeylev: re: if no-version occurs: Good point. It would occur if...hmm...If --python_version was set to a version that wasn't known? This could occur if someone registered a toolchain outside of python.toolchain()

PR #2888, opened 2025-05-16
refactor: make bzlmod create host repos for toolchains

rickeylev:
This moves the creation of the host_toolchain repos into the bzlmod phase.

This is to facilitate future work to allow for when a a particular version
doesn't provide a host-compatible variant

### DIFF
--- a/python/private/python.bzl
+++ b/python/private/python.bzl
@@ -315,6 +315,12 @@ def _python_impl(module_ctx):
                 # The last toolchain is the default; it can't have version constraints
                 set_python_version_constraint = is_last,
             ))
+            if _is_compatible_with_host(mctx, platform_info):
+                host_toolchain(
+                    name = toolchain_info.name,
+                    platforms = [platform_name],
+                    python_version = full_python_version,
+                )
 
     # List of the base names ("python_3_10") for the toolchain repos
     base_toolchain_repo_names = []
@@ -405,6 +411,11 @@ def _python_impl(module_ctx):
         return module_ctx.extension_metadata(reproducible = True)
     else:
         return None
+
+def _is_compatible_with_host(mctx, platform_info):
+    os_name = repo_utils.get_platforms_os_name(rctx)
+    cpu_name = repo_utils.get_platforms_cpu_name(rctx)
+    return platform_info.os_name == os_name and platform_info.arch == cpu_name
 
 def _one_or_the_same(first, second, *, onerror = None):
     if not first:
--- a/python/private/python.bzl
+++ b/python/private/python.bzl
@@ -21,7 +21,7 @@ load(":full_version.bzl", "full_version")
 load(":python_register_toolchains.bzl", "python_register_toolchains")
 load(":pythons_hub.bzl", "hub_repo")
 load(":repo_utils.bzl", "repo_utils")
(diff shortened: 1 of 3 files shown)


python/private/python.bzl
```diff
@@ -286,12 +293,38 @@ def _python_impl(module_ctx):
...
         kwargs.update(py.config.kwargs.get(full_python_version, {}))
         kwargs.update(py.config.default)
-        toolchain_registered_platforms = python_register_toolchains(
+        register_result = python_register_toolchains(
             name = toolchain_info.name,
             _internal_bzlmod_toolchain_call = True,
             **kwargs
         )
-        loaded_platforms[full_python_version] = toolchain_registered_platforms
+        host_compatible = []
```
aignas, line 301: Nice!

python/private/toolchains_repo.bzl
```diff
@@ -375,6 +375,8 @@ def _host_toolchain_impl(rctx):
     if not rctx.delete(python_tester):
         fail("Failed to delete the python tester")
 
+# NOTE: The term "toolchain" is a misnomer for this rule. This doesn't define
+# a repo with toolchains or toolchain implementations.
```
aignas, line 379: What if we rename it to `host_compatible_python_repo`?

aignas (approved):
LGTM, I like the host_compatible solution you did here a lot!

python/private/python.bzl
```diff
@@ -329,31 +362,23 @@ def _python_impl(module_ctx):
...
+        # The target_settings attribute may not be present for users
+        # patching python/versions.bzl.
+        toolchain_ts_map[key] = getattr(entry.platform, "target_settings", [])
+        toolchain_platform_keys[key] = entry.platform_name
+        toolchain_python_versions[key] = entry.full_python_version
+
+        # Repo rules can't accept dict[str, bool], so encode them as a string value.
+        toolchain_set_python_version_constraints[key] = (
+            "True" if entry.set_python_version_constraint else "False"
+        )
```
aignas, line 381: Below here, the host_compatible toolchain list should be passed to the `pythons_hub` so that we keep the right contents in the `interpreters.bzl`.

python/private/toolchains_repo.bzl
rickeylev: Big +1 from me on that!

python/private/python.bzl
rickeylev: Argh, I force-pushed and the code this comment was associated with was nuked.

What what this commenting on?

python/private/python.bzl
```diff
@@ -298,6 +298,7 @@ def _python_impl(module_ctx):
             _internal_bzlmod_toolchain_call = True,
             **kwargs
         )
+        host_compatible = []
```
rickeylev, line 301: (copying prior thread over)

> Below here, the host_compatible toolchain list should be passed to the pythons_hub so that we keep the right contents in the interpreters.bzl.
> This was about the fact that the interpreters.bzl should include only the host_compatible toolchain values and I was not sure if that was the case right now.

I'll check this

rickeylev: Yes, the same interpreters.bzl content is created pre/post this PR.

Thanks for pointing this out. This seems relevant for the final PR in this series that makes a host-compatible repo optional.

PR #2889, opened 2025-05-16
refactor: make bzlmod pass platform mapping to host repo creation

rickeylev:
This makes bzlmod pass the platform metadata to the host_toolchain rule instead of
the host toolchain rule using the fixed PLATFORMS global.

This allows the bzlmod extension to modify the platforms that are available, where the
fixed PLATFORM global can't be changed.


python/private/toolchains_repo.bzl
```diff
@@ -384,6 +386,16 @@ toolchain_aliases repo because referencing the `python` interpreter target from
 this repo causes an eager fetch of the toolchain for the host platform.
     """,
     attrs = {
+        "archs": attr.string_dict(
```
aignas, line 389: nit: 
```suggestion
        "arch_names": attr.string_dict(
```

or `cpu_names`.

aignas (approved):

PR #2895, opened 2025-05-17
sphinxdocs: make xrefs to tag class attributes using attr role work

rickeylev:
The role assigned to attributes within the tag class directive were being given the
role `arg`, when they should be `attr`. This caused xrefs using the attr role to
be unable to find them.

To fix, set them to have the correct role, like the repo rule and regular rule
directives do. Also add a test.


dougthor42 (approved):
I'm kinda surprised that rules_python seems to have such complex sphinx stuff.

My naive thought is that basic ReST and proper sphinx autolinking syntax should be sufficient.

Is a lot of it just so that docs can be built using Bazel? Or because we want to write docs and docstrings as (MyST|markdown) instead of ReST?

Issue #2747, opened 2025-04-07
Improve the `pip.parse` API to allow for incremental building of the configuration


aignas:
Currently the API for defining different parameters by target platform is not ideal and we need to jump over hoops whilst trying to maintain it. The main difficulties are:
* We would like to specify different `pip.parse` attributes per target platform, but we cannot, because some of the parameters need to be labels, label lists or other.
* We cannot have overrides for these things easily and the definitions of what those target platforms are are hard-coded in the code.

The idea that I have is to reuse the same recipe from #2578, where we create a builder for the configuration and it allows us easily define:
* default constraint and flag values for the config settings that we create for different platform variants.
* default values for the pip index url, etc
* default values for which platforms we should support.
* default `auth` configuration for the `bazel_downloader`.
* The list of `python_versions` that are supported #1708.
* build the configuration incrementally for the `pip.parse` invocation allowing per-target-platform configuration of:
    * `extra_pip_args`. #2745 
    * `requirements_lock`. 
    * `constraint_values` and `flag_values`. #2548

What is more the `override` API could be blended in more easily to provide better support for specifying different, patches, etc.

aignas:
Taking the example from https://github.com/bazel-contrib/rules_python/pull/2578#issuecomment-2628784221

We should have something like:
```starlark
pip.defaults(
    hub_name = `pypi`,
    index_url = "https://pypi.org/simple/", # once we want to enable the `experimental_index_url` for everyone or all pip.parse repos regardless where they come from.
    index_url_overrides = {},
)
pip.defaults(
    platform = "cp39_linux_x86_64",
    constraint_values = ["@platforms//os:linux", "@platforms//cpu:x86_64"],
    flag_values = {
        "//python/config_settings:python_version_major_minor": "3.9",
    }
)
pip.defaults(
    platform = "cp313t_linux_x86_64",
    constraint_values = ["@platforms//os:linux", "@platforms//cpu:x86_64"],
    flag_values = {
        "//python/config_settings:python_version_major_minor": "3.13",
        "//python/config_settings:py_freethreaded": "yes",
    }
)
pip.defaults(
    platform = "cp313_linux_x86_64",
    constraint_values = ["@platforms//os:linux", "@platforms//cpu:x86_64"],
    flag_values = {
        "//python/config_settings:python_version_major_minor": "3.13",
        "//python/config_settings:py_freethreaded": "no",
    }
)
# At user site
pip.configure(
    name = "pypi",
[...]

Issue #2548, opened 2025-01-07
pip_parse: requirements_by_platform: allow custom defined platforms


hariom-qure:
# Allow custom platforms to be added to pip_parse

We are trying to install PyTorch and PyTorch has a slightly different way for installing.

The pip command for installing pytorch+linux+cpu:
```bash
pip3 install torch --index-url https://download.pytorch.org/whl/cpu
```
While for a specific GPU version:
```bash
pip3 install torch --index-url https://download.pytorch.org/whl/cu121
```

So we have different requirements files for linux CPU and linux GPU combinations. In the code, I can see that rules_python is using the requirements file depending on the current host that its running, code is [here](https://github.com/bazelbuild/rules_python/blob/main/python/private/pypi/pip_repository.bzl#L95)

How do we pass custom platforms? Something like this
```python
constraint_setting(name = "gpu")

constraint_value(
    name = "gpu_cu_12_1",
    constraint_setting = ":gpu",
)

constraint_value(
    name = "gpu_none",
    constraint_setting = ":gpu",
)

platform(
    name = "lol_x86_64",
    constraint_values = [
        "@platforms//os:linux",
        "@platforms//cpu:x86_64",
        ":gpu_none",
    ],
)

platform(
    name = "linux_x86_64_cu_12_1",
    constraint_values = [
        "@platforms//os:linux",
        "@platforms//cpu:x86_64",
        ":gpu_cu_12_1",
    ],
)
```

# Versions

```
bazel: 6.2.0
rules_python: 0.40.0
```

Thanks!

aignas:
Right now there is no way to do that, similarly to how it is not possible to have switches in the regular Python tooling.

#2449 example shows how people are doing it right now.

hariom-qure:
So the example shown does not have multiple requirements files (I want to support multiple platforms, where torch has different ways for installing in each). Using `requirements_by_platform` is handling most of the cases for me. Only the extra CUDA platform is a problem for me

`pip_parse` all of my requirements files in WORKSPACE
```python
# individual requirements are also parsed with their own names
pip_parse(
    name = "python_deps_linux_x86_64_cu_12_1",
    python_interpreter_target = interpreter,
    requirements_lock = "//third_party/python/locks:requirements_lock_linux_x86_64_cu_12_1.txt",
)

pip_parse(
    name = "python_deps_linux_x86_64",
    python_interpreter_target = interpreter,
[...]

PR #2909, opened 2025-05-18
WIP: a builder API for pip extension 1/n - platform defs

aignas:
This is just some stuff that I wanted to write down before I forget. Feel free
to continue/hack on this.

The goal is to have:
- `platform_version` configuration for the users.
- Support users telling us the preference of which wheels should be used.
- Support users customizing the env markers for the `requirements.txt`
  evaluation. In the future the same API should be good enough for `uv.lock`
  and `py.lock` files.
- Support arbitrary platform definitions - by default `rules_python` will only
  care about the top 2 tiers of the supported platforms, but users should be
  able to customize things.

Work towards #2821
Work towards #2747
Work towards #2548 
Work towards #1708


python/private/pypi/extension.bzl
```diff
@@ -680,6 +757,54 @@ def _pip_impl(module_ctx):
...
+    "index_url_overrides": attr.string_dict(),
+    "simpleapi_skip": attr.string_list(
+        doc = """\
+The list of packages to skip fetching metadata for from SimpleAPI index. You should
+normally not need this attribute, but in case you do, please report this as a bug
+to `rules_python` and use this attribute until the bug is fixed.
+
+EXPERIMENTAL: this may be removed without notice.
+
+:::{versionadded} 1.4.0
```
rickeylev, line 795:
```suggestion
:::{versionadded} VERSION_NEXT_FEATURE
```

rickeylev (commented):
I think I'm following the idea here. Basically, right now you can do:

```
pip.parse(requirements_by_platform={
  "linux*": "requirements_linux.txt",
}, ...)
```

And it uses the PLATFORMS values for what is defined as "linux".

The idea here is to move those definitions into the bzlmod API calls:

```
pip.default(
  platform="my-platform",
  ...
)

pip.parse(requirements_by_platform={
  "my-platform": "requirements_linux.txt",
}, ...)
```

And now instead of using the PLATFORMS global, it'll use that pip.default() information.

Ok, mostly LGTM so far. Under the hood, I can imagine:

```
pip.default(platform="linux_glibc", ...)
pip.default(platform="linux_musl", ...)
pip.parse(requirements_by_platform={
  "linux_glibc": "requirements_a.txt",
  "linux_musl": "requirements_b.txt",
}, ...)
# -> generates a BUILD file like
alias(
  name = "foo"
  actual = select({
    ":is_linux_glibc": "@requirements_a//foo:foo",
    ":is_linux_musl": "@requirements_b//foo:foo",
  })
)
```

Am I imagining the same as you?

Questions:
* What can this express that env markers can't? i.e. why would one do this instead of creating a requirements.txt with `foo; os_name="my-platform-os"` (I might have answered this with my example above -- I don't think "musl" can be expressed by env markers?)
* This seems to create a duplicate mapping of platform names and their metadata (presumably, the python.toolchain() APIs won't be entirely able to not use the platform keys). Is this the part you were mentioning in the recent issue?
* Does this need to feed into the simple-index-url walking? The part I'm thinking of is how it maps wheel names to the config the wheel supports.
* In the example above, all we actually need is a config_setting label. The user could pass that in directly.

aignas:
OK to correctly model the whl selection to allow users to limit the number of selected wheels to minimum requires me to rewrite the `select_whls` function.

I think the current logic is very complex and this could be a chance to simplify it. I think I may back out the code for that from this PR and leave it for the future.

aignas:
I have something that builds the docs (i.e. can download sphinx) but I'll need more time to tidy this up fully.

PR #2911, opened 2025-05-19
fix: Allow PYTHONSTARTUP to define variables

philsc:
With the current `//python/bin:repl` implementation, any variables
defined in `PYTHONSTARTUP` are not actually available in the REPL
itself. I accidentally omitted this in the first patch.

This patch fixes the issue and adds appropriate tests.


aignas (approved):
LGTM

PR #2916, opened 2025-05-20
feat: Support constraints in pip_compile

vihangm:
This adds in support to pass in a constraints file to pip-compile.
This is extremly useful when you want to uprade an indirect/intermediate
dependency to pull in security fixes but don't want to add said dependency to
the requirements.in file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -104,6 +104,9 @@ END_UNRELEASED_TEMPLATE
   Set the `RULES_PYTHON_ENABLE_PIPSTAR=1` environment variable to enable it.
 * (utils) Add a way to run a REPL for any `rules_python` target that returns
   a `PyInfo` provider.
+* (rules) Added support for a constraints file that is passed to pip-compile.
+  Useful when an intermediate dependency needs to be upgraded to pull in
+  security patches.
 
 {#v0-0-0-removed}
 ### Removed
--- a/python/private/pypi/pip_compile.bzl
+++ b/python/private/pypi/pip_compile.bzl
@@ -38,6 +38,7 @@ def pip_compile(
         requirements_windows = None,
         visibility = ["//visibility:private"],
         tags = None,
+        constraints = None,
         **kwargs):
     """Generates targets for managing pip dependencies with pip-compile.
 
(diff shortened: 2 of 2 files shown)


aignas:
The docs say that the `constraints` file can be used more than once: https://pip-tools.readthedocs.io/en/latest/cli/pip-compile/

What is more, I think it is possible to do this today without this PR?
1. Add the `constraints.txt` files to `data` attribute.
2. Use `$(location)` in the args.

Am I missing something here?

vihangm:
> The docs say that the `constraints` file can be used more than once: [pip-tools.readthedocs.io/en/latest/cli/pip-compile](https://pip-tools.readthedocs.io/en/latest/cli/pip-compile/)

Fair point, we could support a list instead of a single file/label instead.

> What is more, I think it is possible to do this today without this PR?
> 
> 1. Add the `constraints.txt` files to `data` attribute.
> 2. Use `$(location)` in the args.
> 
> Am I missing something here?

The data attribute gets popped from the kwargs here, https://github.com/bazel-contrib/rules_python/blob/a746b8fba6472afc935b6e018d5364cc33bad90b/python/private/pypi/pip_compile.bzl#L99
and then the `.update` rule here https://github.com/bazel-contrib/rules_python/blob/a746b8fba6472afc935b6e018d5364cc33bad90b/python/private/pypi/pip_compile.bzl#L160-L165 gets the `$(location)` in the args but the file isn't in it's data deps. So your suggestion doesn't work at the moment.

aignas:
But then it gets reincluded back in via https://github.com/bazel-contrib/rules_python/blob/a746b8fba6472afc935b6e018d5364cc33bad90b/python/private/pypi/pip_compile.bzl#L103

I think I am missing something crucial here. Just by reading the code, this looks like it should work. What is the minimum example that does not work?

aignas:
No, I think the extra attribute for constraints makes sense. The only remaining thing here then is to make it a list.

CHANGELOG.md
```diff
@@ -111,6 +111,9 @@ END_UNRELEASED_TEMPLATE
   and activated with custom flags. See the [Registering custom runtimes]
   docs and {obj}`single_version_platform_override()` API docs for more
   information.
+* (rules) Added support for a using constraints files with `pip_parse`.
```
aignas, line 114:
```suggestion
* (rules) Added support for a using constraints files with `compile_pip_requirements`.
```

aignas (approved):

PR #2922, opened 2025-05-21
fix: When a requirement has more then one possible SHA, only warn that if all SHA returns nothing

stefanpenner:
fix: When a requirement has more then one possible SHA, only warn that if all SHA returns nothing


python/private/pypi/parse_requirements.bzl
```diff
@@ -334,8 +334,8 @@ def _add_dists(*, requirement, index_urls, logger = None):
             sdist = maybe_sdist
             continue
 
-        if logger:
-            logger.warn(lambda: "Could not find a whl or an sdist with sha256={}".format(sha256))
+    if logger and (len(whls) == 0 and sdist == None):
```
aignas, line 337: The vars are undefined here. What is that you are trying to achieve?

stefanpenner: @aignas maybe I'm missing something, but they appear to be defined at:

https://github.com/bazel-contrib/rules_python/pull/2922/files#diff-cea723aaf48fab4acf923b65c3b4fe6c9bb5dec514cacf7c20fb3f92116151d3R311

stefanpenner: Oh `sha256` in the log call isn't defined, let me fix that.

stefanpenner: > What is that you are trying to achieve?
If a requirement has multiple possible SHA, do we need all SHA's to match or just at-least one?

aignas: Ah sorry, somehow I missed it.

However, the intention of the check though was to tell the users that something is wrong andthat the lock file contains hashes that we could not find on the index.

What would be the use case where such behaviour is not desired?

aignas: Ping

aignas:
Closing to cleanup stale PRs.

PR #2924, opened 2025-05-22
docs/refactor: Use python.defaults, not is_default

vonschultz:
When there are multiple Python toolchains, there are currently two ways of setting the default version: the `is_default` attribute of the `python.toolchain()` tag class and the `python.defaults()` tag class. The latter is more powerful, since it also supports files and environment variables. This patch updates the examples and the docs to use `python.defaults()`.

Relates to pull request #2588 and issue #2587.


aignas (approved):
Thank you!

PR #2933, opened 2025-05-24
tests: refactor py_reconfig rules so less boilerplate is needed to add attrs

rickeylev:
Just some minor refactoring of the py_reconfig rule so that it's easier to add attributes
that affect transition state. After this, just two spots have to be modified to
add an attribute (map of attrs, map of attr to transition label).

### DIFF
--- a/tests/support/sh_py_run_test.bzl
+++ b/tests/support/sh_py_run_test.bzl
@@ -13,14 +13,88 @@
 # limitations under the License.
 """Run a py_binary with altered config settings in an sh_test.
 
-This facilitates verify running binaries with different outer environmental
-settings and verifying their output without the overhead of a bazel-in-bazel
-integration test.
+This facilitates verify running binaries with different configuration settings
+without the overhead of a bazel-in-bazel integration test.
 """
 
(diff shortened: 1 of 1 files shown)


tests/support/sh_py_run_test.bzl
```diff
@@ -13,14 +13,88 @@
 # limitations under the License.
 """Run a py_binary with altered config settings in an sh_test.
 
-This facilitates verify running binaries with different outer environmental
-settings and verifying their output without the overhead of a bazel-in-bazel
-integration test.
+This facilitates verify running binaries with different configuration settings
```
dougthor42, line 16: This verbiage is a little odd to me. Is `verify` a command / action? If so, maybe add quotes or a reference?

```
This facilitates "verify" (the action) running binaries with different configuration settings
```

dougthor42 (approved):
I can't say that I fully understand it, but I don't see any glaring issues. So LGTM.

tests/support/sh_py_run_test.bzl
rickeylev: Ah, that's a typo. It should say something more like

> This facilitates running binaries ... and verifying their output ...

PR #2936, opened 2025-05-27
wip: dist-info folders in the venv

aignas:
This adds the necessary `.dist-info` files into the mix
and should get us reasonably close to handling 99% of the cases.

The expected differences from this and a `venv` built by `uv` would be:
* Shared libraries are stored in `<package>.libs` in `uv` venvs. This
  can be achieved in `rules_python` by changing the `installer` settings
  in the `wheel_installer/wheel.py#unzip` function.
* The `RECORD` files are excluded from the `venv`s for better cache hit
  rate in `bazel`, however I am not sure if we should do that for actual
  wheels that are downloaded from the internet.

Tested:
- [x] Building the `//docs` and manually checking the symlinks.
- [ ] Unit tests

Work towards #2156

DO NOT MERGE:
- Stacked on top of #2908


aignas:
I am thinking I should also link other things passed as `data` for completeness.

I think excluding anything outside `site-packages` directory for now would be wise.

----

Tested:
- [x] Building the `//docs` and manually checking the symlinks.
- [x] Ensure that the `.pyi` files get included.
- [x] Ensure that the libs get included.
- [x] Tests that `.dist_info` gets included.
- [x] Tests that ensure that there are files only from one version of the package.
- [x] Remove the topological order usage and document the undefined behaviour.
- [x] document PyInfo changes.

aignas:
I'll push some of the changes later but I have a hard time writing a test that would fullfill some of the requirements in the PR todo list.

tests/modules/other/simple_v2/site-packages/simple.libs/data.txt
```diff
@@ -0,0 +1 @@
+Intentionally empty
```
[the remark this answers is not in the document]
aignas: It is usually libs, but I was not sure how to have a minimal loadable extension for the unit test.

The convention is present on latest wheels published on PyPI, but I haven't seen it in previous wheels, so it is a function of the publisher and not the installer.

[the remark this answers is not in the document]
rickeylev: by libs you mean e.g. `.so` files? Are these python C extensions or regular C libraries? Is there a spec/reference/informational doc that I could read?

TBH, I'm surprised this would be a convention -- why bother putting files in a separate directory _outside_ the package's directory?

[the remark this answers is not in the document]
aignas: https://pypi-browser.org/package/numpy/numpy-2.2.6-cp310-cp310-manylinux_2_17_aarch64.manylinux2014_aarch64.whl

I've found it only on the linux wheel, the mac wheel did not have it, so could be OS specific.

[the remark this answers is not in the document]
aignas: I think the `foo.libs` comes from `auditwheel`: https://github.com/pypa/auditwheel/blob/main/CHANGELOG.md#310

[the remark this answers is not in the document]
aignas: Right now we are *not* using the `RECORD` in this starlark code. The `RECORD` might be (or is) used by the `installer` pkg when extracting the wheel for the first time.

In the future, instead of using `glob` in the `whl_library` we could use the `RECORD` file to generate the `srcs`, `data` and `pyi_srcs` lists, but I am not sure we actually want to validate hashes of each file (though we could).

python/private/py_executable.bzl
```diff
@@ -682,15 +682,30 @@ def _create_venv_symlinks(ctx, venv_dir_map):
 def _build_link_map(entries):
     # dict[str kind, dict[str rel_path, str link_to_path]]
     link_map = {}
+
+    # Here we store venv paths by package
```
rickeylev: The type description makes it clear they're keyed by package, so no need to restate it in prose.

```suggestion
```

python/private/py_executable.bzl
```diff
@@ -682,15 +682,30 @@ def _create_venv_symlinks(ctx, venv_dir_map):
...
         kind_map = link_map.setdefault(kind, {})
-        if entry.venv_path in kind_map:
-            # We ignore duplicates by design. The dependency closer to the
-            # binary gets precedence due to the topological ordering.
-            continue
-        else:
-            kind_map[entry.venv_path] = entry.link_to_path
+
+        # If we detect that we are adding a dist-info for an already existing package
+        # we need to pop all of the previous symlinks from the link_map
```
rickeylev: This changes the semantics to last-wins. We want first-wins so that the topological ordering has meaning.

python/private/py_executable.bzl
```diff
@@ -682,15 +682,30 @@ def _create_venv_symlinks(ctx, venv_dir_map):
...
+        if entry.venv_path.endswith(".dist-info") and entry.src in pkg_map:
+            # dist-info will come always first
+            for kind, dir_paths in pkg_map.pop(entry.src).items():
+                for dir_path in dir_paths:
+                    link_map[kind].pop(dir_path)
+
+        pkg_venv_paths = pkg_map.setdefault(entry.src, {}).setdefault(entry.kind, [])
+        pkg_venv_paths.append(entry.venv_path)
+
+        # We overwrite duplicates by design. The dependency closer to the
```
rickeylev: Similar to above: if an entry is being overwritten, it means the first-wins semantics aren't occurring.

python/private/py_library.bzl
```diff
@@ -206,16 +207,32 @@ Source files are no longer added to the runfiles directly.
...
+            return d
+
+    return None
+
 def _get_imports_and_venv_symlinks(ctx, semantics):
     imports = depset()
     venv_symlinks = depset()
     if VenvsSitePackages.is_enabled(ctx):
-        venv_symlinks = _get_venv_symlinks(ctx)
+        dist_info_metadata = _get_distinfo_metadata(ctx)
```
rickeylev: nit: move the get_distinfo_metadata() call inside the _get_venv_symlinks function. The value isn't used outside that function and that function already has the necessary inputs to call it.

python/private/py_library.bzl
```diff
@@ -254,16 +271,41 @@ def _get_venv_symlinks(ctx):
...
+        # third value into the venv_symlinks, which would be the normalized
+        # package name. This allows us to ensure that we can replace the `dist-info`
+        # directories by checking if the package key is there.
+        dist_info_dir = paths.basename(dist_info_metadata.dirname)
+        package, _, _suffix = dist_info_dir.rpartition(".dist-info")
+        package, _, _version = package.rpartition("-")
+        package = normalize_name(package)
+
+        repo_runfiles_dirname = runfiles_root_path(ctx, dist_info_metadata.short_path).partition("/")[0]
+        venv_symlinks.append(VenvSymlinkEntry(
```
rickeylev: There needs to be a comment somewhere around here that the particular ordering matters -- the dist-info must come before the other symlinks to ensure things are properly respected.

python/private/py_library.bzl
```diff
@@ -254,16 +271,41 @@ def _get_venv_symlinks(ctx):
...
+            kind = VenvSymlinkKind.LIB,
+            link_to_path = paths.join(repo_runfiles_dirname, site_packages_root, dist_info_dir),
+            src = package,
+            venv_path = dist_info_dir,
+        ))
+
+    for src in ctx.files.srcs + ctx.files.data:
```
rickeylev: Something seems wrong here, but I can't quite put my finger on it.

Files in srcs have to have the init.py logic applied to detect namespace package boundaries to auto-detect the proper paths to create.

Files in data shouldn't be part of that logic. If the file is within a directory covered by something in srcs (i.e *.py), then there's no need to look at the data file. If the data file _isn't_ in a directory covered by something in srcs, then it shouldn't treat a .so/.pyi/.pyc file as some sort of boundary marker.

python/private/py_info.bzl
```diff
@@ -67,6 +67,11 @@ the venv to create the path under.
 
 A runfiles-root relative path that `venv_path` will symlink to. If `None`,
 it means to not create a symlink.
+""",
+        "src": """
```
rickeylev: No comment on renaming this field?

rickeylev (changes requested):

tests/modules/other/simple_v2/site-packages/simple.libs/data.txt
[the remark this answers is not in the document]
rickeylev: TIL. Thanks for explaining.

python/private/py_executable.bzl
aignas: Could you describe this in terms of what the end user should be able to do? I think we actually want the last wins? I am a little fuzzy about last vs first here. I added a test to what I think should be possible, so maybe we can discuss there.

python/private/py_library.bzl
aignas: This is no longer true. It can come anywhere, I think.

python/private/py_library.bzl
aignas: Talking about special treatment, I don't think I understand why we need it. We could just treat everything as just files and interleave the directories? There is nothing special in my mind here.

Beforehand we had the auto-generated `namespace pkg __init__.py` files, but if we get rid of them everything is just files, right?

python/private/py_executable.bzl
rickeylev: A binary should be able to add a dependency closer to it to override behavior of further away targets.

"first" refers to topological ordering: the first item is closer. So given this:

```
py_binary(
  name = "bin",
  deps = [
    ":foo_v1",
    ":middle",
  ],
)
py_library(
  name = "middle",
  deps = [":foo_v2"]
)
```

Where foo_v1 and foo_v1 both produce a venv symlink for `foo/lib.py`

Then foo_v1's entry should be used and foo_v2's ignored. This is because the serialized order of bin's targets (i.e venv_symlinks.to_list()) follows topological order: foo_v1, middle, foo_v2

python/private/py_library.bzl
rickeylev: > treat everything as files and interleave the directories

You mean, for every file, create a venv_symlink entry? That won't scale because it doubles the number of files a binary has to materialize, when all that's really needed is e.g. one symlink to a directory.

However, because of namespace packages, we can't simply use the directories directly under `site-packages/`.

python/private/py_executable.bzl
aignas: I've added this sort of test and I've found that the current logic is what I need. That is why I have `simple_v1` and `simple_v2` where `simple_v2` is closer to the `py_binary`. Could you check out the PR and check it please? Maybe my knowledge of depsets is not deep enough.

I think that the `topological` sort is not working right, but I'll need to reed the code once more to fully understand where the `topological` sort should be used.

python/private/py_library.bzl
aignas: What if we treat everything as `namespace` packages? The way I understand this loop is:
* For each file in the sources, symlink their directories.
* If we have common things across `data` and `srcs` then we double the number of directories that we need to process, but I don't see a way to work around the case where we have a big package (e.g. airflow) that have data files under `site-packages/airflow/data-files-without-python-files` that would need to be symlinked in addition to other namespace packages.

I am wondering how we can do this?

python/private/py_library.bzl
```diff
@@ -206,16 +207,32 @@ Source files are no longer added to the runfiles directly.
...
+            continue
+
+        if d.dirname.endswith(".dist-info"):
+            return d
+
+    return None
+
 def _get_imports_and_venv_symlinks(ctx, semantics):
     imports = depset()
     venv_symlinks = depset()
```
aignas: This is missing `order = "topological"`?

python/private/py_library.bzl
```diff
@@ -206,16 +207,32 @@ Source files are no longer added to the runfiles directly.
...
     if VenvsSitePackages.is_enabled(ctx):
-        venv_symlinks = _get_venv_symlinks(ctx)
+        dist_info_metadata = _get_distinfo_metadata(ctx)
+        venv_symlinks = _get_venv_symlinks(
```
aignas: This is overwriting `venv_symlinks` `depset` with a `list`.

python/private/py_library.bzl
aignas: This comment was wrong, this is fine.

python/private/py_info.bzl
aignas: hmmm. I renamed this to `package`, I liked the suggestion.

python/private/py_executable.bzl
aignas: OK, fixed the topological behaviour of the depset.

python/private/py_executable.bzl
aignas: OK, fixed the code, now I think I understand it.

python/private/py_info.bzl
aignas: I have also added a separate field version that a few people have asked in the past for.

aignas:
Hah, I think I know what is happening but I am not sure if the way I am thinking of solving is the easiest approach.

In the unit test if I swap the order of the `simple_v2` with the `user_lib` then the
test passes, otherwise the test fails.

I think the problem is that `buildifier` usually puts the third party deps at
the end of the `deps` list and this means that the topological sort in this
particular case will not be the thing that we want. This is the reason why
"popping" the previous package worked and other things did not.

In order to have the correct topological sort behaviour we have to have the
third party deps come in first when we are constructing the depset. We have to
be careful to only do this sorting of the deps when we have enabled the venv
building and we have `pystar` enabled.

Otherwise, we need to continuously overwrite (and still use the topological sort), but
I think it makes the code less intuitive.

EDIT: ~I think switching to `postorder` that will allow us first going through the transitive dependencies _might_ be the right behaviour here.~

EDIT2: I fixed the logic to what I think should be correct.

aignas:
OK, I think this is in an interesting place right now:
1. I remember people requesting package name and the package version in the provider at some point. This was probably for generating `requires-dist` when packaging wheels or for license stuff.
1. I need to add these fields so that I can merge the depsets in a careful way: first merge the third-party deps and then the first party ones.
1. This allows the correct results irrespective of how the `deps` is sorted.

Though it feels that this is somewhat complicated.

python/private/py_info.bzl
```diff
@@ -205,6 +220,7 @@ Python targets. These are accumulated from the transitive `deps`.
 The order of the depset is not guaranteed and may be changed in the future. It
 is recommended to use `default` order (the default).
 """,
+        "package": "TODO",
```
aignas: Related discussion in Bazel Slack about asking if we should add this field: https://bazelbuild.slack.com/archives/CA306CEV6/p1743702775560119?thread_ts=1743702761.960199&cid=CA306CEV6

rickeylev:
> In the unit test if I swap the order of the simple_v2 with the user_lib then the
test passes, otherwise the test fails.

Topo order doesn't guarantee left-to-right/RTL order, but it shouldn't matter in this particular case because the thing we want is closer regardless of LTR/RTL ordering.

The test has this structure:

```
          venvs_site_packages_libs_test
                      |
                   closer_lib
                  /            \
             user lib           simple_v2
               |
          simple v1
```

The topological ordering should be: closer_lib, (userlib, simplev2), simplev1. Or more precisely, since it's the depsets, not targets, that get merged:

```
# order=topological implied for all depset() calls
simple_v1 = depset(...)
simple_v2 = depset(...)

user_lib = depset(transitive=[simple_v1])

closer_lib = depset(transitive=[user_lib, simple_v2])
venvs_test = depset(transitive=[closer_lib])
```

I put parentheses because Bazel doesn't guarantee left-to-right or right-to-left for topological, just top-down uniqueness. Regardless, simplev2 should occur first.

python/private/py_library.bzl
```diff
@@ -292,12 +338,17 @@ def _get_venv_symlinks(ctx):
...
+            link_to_path = paths.join(prefix, site_packages_root, dirname),
+            package = package,
+            version = version_str,
             venv_path = dirname,
-        ))
-    return venv_symlinks
+        )
+        venv_symlinks.append(entry)
+
+    return depset(venv_symlinks)
```
rickeylev: Missing topological order here. Maybe this is the cause of everything?

Depsets with an explicit order will check their compatibility. However, depsets without an explicit ordering will happily merge with whatever. This behavior is shallowly enforced, from what I recall. i.e. you can merge a postorder with preorder depset by wrapping one in a default order. idk what behavior is exactly when this is done.

```suggestion
    return depset(venv_symlinks, order="topological")
```

rickeylev (commented):
Ok, read through posts since I last looked. I'm not entirely clear on current status, so I just gave it a quick review

aignas:
In my test I get the following `venv_paths` when the topological order deps are converted to list (if I remove the patch I have to sort the deps before merging the depsets):
```
simple
simple-1.0.0.dist-info
simple_v1_extras
simple
simple-2.0.0.dist-info
simple.libs
nspkg/subnspkg/alpha
nspkg/subnspkg/beta
nspkg/subnspkg/delta
nspkg/subnspkg/gamma
__init__.py
single_file.py
with_external_data.py
external_data
```

This means that It is depth first topological rather than breadth first, if that makes sense.

tests/venv_site_packages_libs/depset_tests.bzl
```diff
@@ -0,0 +1,104 @@
...
+            "c2",
+            "b",
+            "c1",
+        ],
+        [
+            "a2",
+            "b",
+            "c1",
+            "c2",
+        ],
```
aignas: OK, so I had an idea that instead of discussing in markdown about theory and findings from `print` debugs, we can have a Starlark test checking the `depset` behaviour.

So as I say, it seems that the topological sorting is DFS.

aignas:
OK, PTAL, have updated docs.

rickeylev:
I revisited what topological ordering means. You're right, essentially. It's not breadth-depth like I originally thought. It's about ensuring that the nodes that point _to_ a node are listed _before_ that node in the linearized ordering. Note that "nodes" means the values in the depset, not depset objects themselves. i.e. `depset(["a"], transitive=["b", "a"])` results in `[b, a]`.

Stated another way: there's no guarantee a value will occur before (or after) another when there is no path between them. This means, in order to get the semantics we want (via depset ordering) -- that simple_v2 can be overridden by another target -- the overriding target _must_ depend on simple_v2, which kinda defeats the purpose of overriding (I shouldn't have to depend on what I want to remove in order to remove it).

So yeah, topological ordering doesn't give the "closet target wins" behavior that I was originally thinking.

Reading https://bazel.build/rules/lib/builtins/depset, preorder might be closer to what we want (direct elements first)? Though, topological ordering could maybe reflect an install order? I'll take the thinking-out-loud to slack.

-----

I'm -1 on trying to classify something as 3P vs 1P. There just isn't a clear definition of what is or isn't. Whether code is made available via pip_parse, or I write an equivalent target in my main repo, or I get it from a custom repository rule, or etc doesn't tell if a particular target is "mine" or "not mine".

-----

I would like to make progress on this PR, though. I'm thinking we just drop the mention of an explicit ordering to allow overriding. Just leave it as undefined for now. i.e., if there's multiple entries for the same path, it won't give an error, but behavior is undefined. This is still experimental, after all. We can try and figure out a better conflict resolution strategy separately.

aignas:
Hmm, I am +1 on strictly mentioning that it is undefined then. We can leave this discussion about how to override targets separately.

I'll cleanup the docs as part of this PR then.

aignas:
OK, PTAL, I think the majority of the stuff should be ready to merge.

python/private/py_library.bzl
```diff
@@ -84,20 +86,24 @@ under the binary's venv site-packages directory that should be made available (i
...
-presence or absense. Stated another way: [pkgutil-style namespace packages](
+presence or absence. Stated another way: [pkgutil-style namespace packages](
 https://packaging.python.org/en/latest/guides/packaging-namespace-packages/#pkgutil-style-namespace-packages)
 won't be understood as namespace packages; they'll be seen as regular packages. This will
 likely lead to conflicts with other targets that contribute to the namespace.
 
 :::{tip}
-This attributes populates {obj}`PyInfo.venv_symlinks`, which is
-a topologically ordered depset. This means dependencies closer and earlier
+This attributes populates {obj}`PyInfo.venv_symlinks` a custom ordered depset - third party dependencies come first when the depsets are merged. This means dependencies closer and earlier
```
rickeylev:
```suggestion
This attributes populates {obj}`PyInfo.venv_symlinks`
```

python/private/py_library.bzl
```diff
@@ -206,16 +213,47 @@ Source files are no longer added to the runfiles directly.
 :::
 """
 
-def _get_imports_and_venv_symlinks(ctx, semantics):
+def _get_package_and_version(ctx):
+    """Return package name and version
+
+    If the package comes from PyPI then it will have a `.dist-info` as part of `data`, which
+    allows us to get the name of the package and its version. This means that we can ensure
+    that package usage closer to the terminal node can override dependencies.
```
rickeylev:
```suggestion
    allows us to get the name of the package and its version.
```

tests/venv_site_packages_libs/depset_tests.bzl
rickeylev, line 1: This was just testing, yes? Please remove

rickeylev:
> should RECORD files be excluded

I don't think it matters? The RECORD file is just a list of files and their hashes. Bazel is already tracking all the files and their hashes. If any change, then RECORD changes, but also Bazel will see them change directly.

arrdem:
+1, the `RECORD` file is stable, is part of the distribution, would be expected in a venv built with any other tool and should be preserved here.

aignas:
> +1, the `RECORD` file is stable, is part of the distribution, would be expected in a venv built with any other tool and should be preserved here.

@arrdem, I think we need to make an additional change in the `whl_library_targets` for that, so I'll leave this out of scope for this PR. Maybe we can do this only for `whl_library` instances that are created from `whls` and not from `sdists`, because I can see how there could be non-reproducibility for sdists based on some environment.

PR #2938, opened 2025-05-28
feat: add persistent worker for sphinxdocs

kaycebasques:
This implements a simple, serialized persistent worker for Sphinxdocs with several optimizations. It is enabled by default.

* The worker computes what inputs have changed, allowing Sphinx to only rebuild what
  is necessary.
* Doctrees are written to a separate directory so they are retained between builds.
* The worker tells Sphinx to write output to an internal directory, then copies it
  to the expected Bazel output directory afterwards. This allows Sphinx to only
  write output files that need to be updated.

This works by having the worker compute what files have changed and having a Sphinx
extension use the `get-env-outdated` event to tell Sphinx which files have changed.
The extension is based on https://pwrev.dev/294057, but re-implemented to be
in-memory as part of the worker instead of a separate extension projects must configure.

For rules_python's doc building, this reduces incremental building from about 8 seconds
to about 0.8 seconds. From what I can tell, about half the time is spent generating
doctrees, and the other half generating the output files.

Worker mode is enabled by default and can be disabled on the target or by adjusting
the Bazel flags controlling execution strategy. Docs added to explain how.

Because `--doctree-dir` is now always specified and outside the output dir,
non-worker invocations can benefit, too, if run without sandboxing. Docs added to
explain how to do this.

Along the way:

* Remove `--write-all` and `--fresh-env` from run args. This lets direct
  invocations benefit from the normal caching Sphinx does.
* Change the args formatting to `--foo=bar` so they are a single element; just
  a bit nicer to see when debugging.

Work towards https://github.com/bazel-contrib/rules_python/issues/2878, https://github.com/bazel-contrib/rules_python/issues/2879


sphinxdocs/private/sphinx.bzl
```diff
@@ -177,6 +178,9 @@ def sphinx_docs(
         **common_kwargs
     )
 
+    common_kwargs_with_manual_tag = dict(common_kwargs)
+    common_kwargs_with_manual_tag["tags"] = list(common_kwargs.get("tags") or []) + ["manual"]
```
kaycebasques, line 182: Pigweed repo is running v0.40.0 of rules_python so I think I'm just accidentally re-introducing some code that has been deleted

sphinxdocs/private/sphinx.bzl
```diff
@@ -243,26 +243,34 @@ _sphinx_docs = rule(
...
-def _run_sphinx(ctx, format, source_path, inputs, output_prefix):
+def _run_sphinx(ctx, format, source_path, inputs, output_prefix, use_cache):
+
     output_dir = ctx.actions.declare_directory(paths.join(output_prefix, format))
 
     run_args = []  # Copy of the args to forward along to debug runner
     args = ctx.actions.args()  # Args passed to the action
 
+    args.add(source_path)
+    args.add(output_dir.path)
```
kaycebasques, line 272: Pretty sure that `sphinx.cmd.build.main` (or whatever function we're using) expects srcdir and outdir to be the first 2 args. The function was failing in weird ways because they weren't the first 2 args.

sphinxdocs/private/sphinx.bzl
```diff
@@ -275,10 +283,14 @@ def _run_sphinx(ctx, format, source_path, inputs, output_prefix):
     # Build in parallel, if possible
     # Don't add to run_args: parallel building breaks interactive debugging
     args.add("--jobs", "auto")
-    args.add("--fresh-env")  # Don't try to use cache files. Bazel can't make use of them.
-    run_args.append("--fresh-env")
-    args.add("--write-all")  # Write all files; don't try to detect "changed" files
-    run_args.append("--write-all")
+
+    if use_cache:
+        args.add("--doctree-dir", paths.join(output_dir.path, ".doctrees"))
```
kaycebasques, line 288: When you spin up a project with `sphinx-quickstart` and then run `make html` (or whatever) to build the docs, the doctree dir gets output as a sibling to the HTML output dir. But when you build from within Bazel, the doctrees dir does indeed get output as a child of the HTML output dir

sphinxdocs/private/sphinx.bzl
```diff
@@ -302,16 +312,41 @@ def _run_sphinx(ctx, format, source_path, inputs, output_prefix):
...
-        arguments = [args],
-        inputs = inputs,
-        outputs = [output_dir],
-        tools = tools,
-        mnemonic = "SphinxBuildDocs",
-        progress_message = "Sphinx building {} for %{{label}}".format(format),
-        env = env,
-    )
+    if use_cache:
+        worker_arg_file = ctx.actions.declare_file(ctx.attr.name + ".worker_args")
```
kaycebasques, line 316: I think this logic could be simplified by using argsfile regardless of whether you're in non-persistent-worker-mode or persistent-worker-mode

rickeylev (commented):
I pushed some changes that I think make this work as a basic incremental builder that relies on the builtin timestamp based detection. Not perfect, but progress!

The key change was to move the doctree directory out of the builder output directory. The builder output directory gets deleted every time the action has to re-run, which would delete the doctree files, too. Now that the doctree files are preserved between worker requests, it's able to reuse them.

rickeylev:
I had an idea for registering an extension without modifying conf.py

`sphinx.application.builtin_extensions` is a list of extensions it always loads. We can implement an extension and add it to that list.

Communicating from the worker loop to the extension is the tricky part. I see in the prototype, digest.json is being used by sticking it in the doctree directory. This is an appealing idea! A regular global would work, too, which would avoid any serialization overhead. Regardless, however it works, the content has to be specific to the particular src/out directory because a worker could service different requests for different doc roots.

Alternatively, the extension could just hash everything itself. This would at least work in a more general sense, better than timestamps.

sphinxdocs/private/sphinx.bzl
rickeylev: Weird. It should have been indifferent due to how things are parsed. In any case, the changes I've made create a reliance on this, so I put a comment to this effect.

sphinxdocs/private/sphinx.bzl
```diff
@@ -104,6 +103,7 @@ def sphinx_docs(
         strip_prefix = "",
         extra_opts = [],
         tools = [],
+        use_cache = False,
```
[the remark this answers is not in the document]
rickeylev: I found it useful, in my testing, too. Renamed to use_persisent_worker

sphinxdocs/private/sphinx.bzl
rickeylev: I synced this PR to head and resolved conflicts.

sphinxdocs/private/sphinx.bzl
rickeylev: Per my other comment: I've refactored the code to put the doctree dir outside of the output dir. This is necessary because bazel will clear the output dir on every action re-run, thus losing the doctrees if they're in it.

sphinxdocs/private/sphinx.bzl
rickeylev: Yes, done.

rickeylev (commented):
I cleaned up the implementation heavily. Two things of note.

The worker installs an in-memory extension that implements the same basic get-env-outdated logic as in the pigweed extension. I'm not 100% sure that simple logic is 100% correct because the timestamp implementation does a lot more stuff. But, my testing didn't show errors.

There was a certain appeal to the digest.json method the original impl / pigweed extension were using, so I had the worker write a similar looking file. It passes the path to the file by using `--define` to set a value in the config file. Extensions can find it via app.config and looking up that config value.

rickeylev:
Another finding: adding no-sandbox will help incremental building. This comes back to the lack of a persistent doctree dir. The env object is pickled into the doctree dir; the env object is what persists information (such as what files already exist, their last modify times, etc).

Even if --doctree-dir is specified, because it's not a declared input, when a sandbox'd execution occurs, the prior run's doctree directory isn't copied into the sandbox. It can't be because it's not a File input. And thus, Sphinx starts with an empty env, sees all files as new, and has to start from scratch.

Setting no-sandbox allows the files from a previous run to become visible. This saved about half the execution time (8s -> 4s) when building rule_python's docs.

Additionally, with the worker-based impl, which is able to fully persist the doctree dir _and_ output dir between runs, incremental building is way faster: 8s -> 0.8 seconds

kaycebasques:
FYI I'm on staycation this week. Will review if I have some free time otherwise can review next week

rickeylev (approved):

rickeylev:
Oh, to clarify, I'm not blocked on needing review. I can just merge it.

PR #2939, opened 2025-05-29
refactor: change site_packages_symlinks to venv_symlinks

rickeylev:
This generalizes the ability to populate the venv directory by adding and additional field,
`kind`, which tells which directory of the venv to populate. A symbolic constant is used
to indicate which directory so that users don't have to re-derive the platform and version
specific paths that make up the venv directory names.

This follows the design described by https://github.com/bazel-contrib/rules_python/issues/2156#issuecomment-2855580026

This also changes it to a depset of structs to make it more forward compatible. A provider
is used because they're slightly more memory efficient than regular structs.

Work towards https://github.com/bazel-contrib/rules_python/issues/2156


python/private/py_executable.bzl
```diff
@@ -618,89 +619,103 @@ def _create_venv(ctx, output_prefix, imports, runtime_details):
...
 
-def _create_site_packages_symlinks(ctx, site_packages):
-    """Creates symlinks within site-packages.
+def _create_venv_symlinks(ctx, venv_dir_map):
+    """Creates symlinks within the venv.
 
     Args:
         ctx: current rule ctx
-        site_packages: runfiles-root-relative path to the site-packages directory
+        venv_dir_map: todo
```
aignas, line 644: nit: only this TODO needs to be replaced.

aignas (commented):
LGTM other than the nit.

aignas (approved):

PR #2941, opened 2025-05-29
fix(pypi): inherit proxy env variables in compile_pip_requirements test

herewasmike:
Bazel does not pass environment variables implicitly (even running test outside of sandbox). This forces compile_pip_requirements test to fail with timeout when attempting to run it behind the proxy. Also changes test_command in dependency_resolver string helper to use dot instead of underscore following deprecation notice


aignas:
Right now the user has to specify the env_inherit by themselves? This would break such users as the PR is right now?

herewasmike:
Yes, you're right, I've pushed the change to merge those lists if user-provided env_inherit variable.

Right now it is possible to specify env_inherit manually, but it's really unclear and requires a good bit of digging to understand that you need to do it

aignas:

PR #2952, opened 2025-06-02
refactor: switch to official go tree-sitter implementation

jbedard:
The https://github.com/smacker/go-tree-sitter library seems dead and we should switch to the official https://github.com/tree-sitter/go-tree-sitter. I know there's already a lot of discussion around tree-sitter in `rules_python/gazelle`, I thought I should bring this into the discussion as well. This PR depends on even more patches then what https://github.com/dougthor42/go-tree-sitter is patching, I'm not sure how you want to deal with that going forward (I've found the latest rules_go solves 50% of it).

The official library has been properly maintained for some time now and has a lot of features that the smacker version is lacking. I was able to delete a decent amount of code in the js/ts gazelle extension when switching (such as all [predicate filtering](https://github.com/aspect-build/aspect-cli/blob/d89f6e83dc94ccd926d718a0506fd944f5826802/gazelle/common/treesitter/filters.go#L19) which is properly implemented in the official go-tree-sitter).


gazelle/MODULE.bazel
```diff
@@ -17,14 +17,25 @@ local_path_override(
 
 go_deps = use_extension("@bazel_gazelle//:extensions.bzl", "go_deps")
 go_deps.from_file(go_mod = "//:go.mod")
+go_deps.module_override(
```
dougthor42, line 20: This is, sadly, a blocker because overrides cannot be used in non-root modules.

If this is included, then no one can add the dep via:

```starlark
# MODULE.bazel
bazel_dep(name = "rules_python_gazelle_plugin")
```

As it will fail with:

```
Using the "go_deps.{tag_class}" tag in a non-root Bazel module is forbidden
```

dougthor42 (changes requested):

gazelle/MODULE.bazel
```diff
@@ -17,14 +17,25 @@ local_path_override(
...
+    patches = ["//patches:go-tree-sitter-cc_library.patch"],
+    path = "github.com/tree-sitter/go-tree-sitter",
+    patch_strip = 1,
+)
+go_deps.module_override(
+    path = "github.com/tree-sitter/tree-sitter-python",
```
jbedard, line 26: One alternative to patching the `tree-sitter-*` libraries (with the `bindings/go/bindings.go` file) is generating a single BUILD ourselves like what I've done in the aspect-cli, see [the http fetch+BUILD](https://github.com/aspect-build/aspect-cli/blob/d89f6e83dc94ccd926d718a0506fd944f5826802/gazelle/common/treesitter/grammars/grammars.bzl#L7-L23) and [associated go](https://github.com/aspect-build/aspect-cli/blob/d89f6e83dc94ccd926d718a0506fd944f5826802/gazelle/common/treesitter/grammars/golang/binding.go).

That's probably nicer then a patch now that I look at it again...?

gazelle/MODULE.bazel
jbedard: You can fork that library as well if you want, or ask users to add these 3 lines themselves. I'm just opening this PR to suggest using the modern `go-tree-sitter` library...

gazelle/MODULE.bazel
jbedard: Using go modules works better when mixing go modules across multiple modules though, such as multiple gazelle languages from different modules. But that requires the patches :/

gazelle/MODULE.bazel
dougthor42: I'm 100% for using the modern go-tree-sitter library. I just think we are blocked from doing so until https://github.com/bazel-contrib/rules_go/pull/4298 is released and we update our `rules_go` version.

Otherwise we get stuck in the "users must apply a patch themselves" or "we have to maintain a fork" situation that we're already in.

I think the path forward is:

1.  Wait for rules_go to release the fix.
2.  Update rules_go
3.  Remove the dougthor42/go-tree-sitter hack and go back to smacker/go-tree-sitter
4.  Migrate to the official tree-sitter repos.

gazelle/MODULE.bazel
dougthor42: Making the BUILD file ourselves is basically what the dougthor42/go-tree-sitter does.

Does manually setting `http_archive.build_file_content` work with both WORKSPACE and bzlmod? If so that might be a way forward. I agree it's cleaner than the patches.

gazelle/deps.bzl
```diff
@@ -197,8 +199,94 @@ def go_deps():
...
+        patch_strip = 1, # keep
+    )
+    go_repository(
+        name = "com_github_tree_sitter_tree_sitter_ruby",
+        importpath = "github.com/tree-sitter/tree-sitter-ruby",
+        sum = "h1:T/NKHUA+iVbHM440hFx+lzVOzS4dV6z8Qw8ai+72bYo=",
+        version = "v0.23.1",
+    )
+    go_repository(
+        name = "com_github_tree_sitter_tree_sitter_rust",
```
dougthor42, line 286: Why does this need all of the non-python languages?

dougthor42 (changes requested):

gazelle/deps.bzl
jbedard: I don't think it needs any of them, but [they are there](https://github.com/tree-sitter/go-tree-sitter/blob/adc13ffd8b2c0b01b878fda9f7c422ce0df5fad3/go.mod#L8-L19), I think [just for tests AFAICT](https://github.com/tree-sitter/go-tree-sitter/blob/adc13ffd8b2c0b01b878fda9f7c422ce0df5fad3/parser_test.go#L20-L31).

I don't think we should care. That's how go.mod works, and deps.bzl is how (legacy) rules_go works.

gazelle/MODULE.bazel
jbedard: This is only for the `tree-sitter-{lang}/bindings/go/bindings.go` files, not the main `go-tree-sitter` go module that has the bigger problem.

gazelle/MODULE.bazel
jbedard: I don't think https://github.com/bazel-contrib/rules_go/pull/4298 fixes every scenario.  After upgrading to rules_go@HEAD I was able to remove my patch to `smacker/go-tree-sitter`, but when switching to `tree-sitter/go-tree-sitter` I had to add almost the same patch again 😢

jbedard:
FWIW I might be blocked on https://github.com/tree-sitter/go-tree-sitter/issues/32 for now so your plan upgrading rules_go and dropping the fork is probably still the next best step.

jbedard:
Changing this to a draft until we have a larger discussion of this is worth it. I found https://github.com/tree-sitter/go-tree-sitter/issues/32 was too noticeable of a perf hit when changing the js gazelle extension.

dougthor42:
ACK, SGTM.

jbedard:
Done a while back in https://github.com/bazel-contrib/rules_python/pull/3069?

PR #2959, opened 2025-06-05
refactor(gazelle) Types for exposed members of `python.ParserOutput` are now all public

yushan26:
Export the members of `python.ParserOutput` struct to make it publicly accessible. This allows other `py` extensions to leverage the Python resolver logic for resolving Python imports, instead of have to duplicate the resolving logic.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,27 @@ BEGIN_UNRELEASED_TEMPLATE
 END_UNRELEASED_TEMPLATE
 -->
 
+{#v0-0-0}
+## Unreleased
+
+[0.0.0]: https://github.com/bazel-contrib/rules_python/releases/tag/0.0.0
+
+{#v0-0-0-changed}
+### Changed
+* (gazelle) Types for exposed members of `python.ParserOutput` are now all public.
+
+{#v0-0-0-fixed}
+### Fixed
+* Nothing fixed.
+
+{#v0-0-0-added}
+### Added
+* Nothing added.
+
+{#v0-0-0-removed}
+### Removed
+* Nothing removed.
+
 {#1-5-0}
 ## [1.5.0] - 2025-06-11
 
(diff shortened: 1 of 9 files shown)


CHANGELOG.md
```diff
@@ -1184,6 +1184,7 @@ Other changes:
   See [#1371](https://github.com/bazel-contrib/rules_python/issues/1371).
 * (refactor) The pre-commit developer workflow should now pass `isort` and `black`
   checks (see [#1674](https://github.com/bazel-contrib/rules_python/issues/1674)).
+* (refactor) Export module as a public struct
```
dougthor42:
```suggestion
* (gazelle) Export `parser.Module` as a public struct.
```

dougthor42: Please move this up to the "unreleased" section.

dougthor42 (commented):
> allows other py extensions to leverage the Python resolver logic

Can you elaborate on what these extensions are or might be?

TBH I'm hesitant to make any more of the gazelle go code truly public. I already am not a fan of the existing public items haha.

I'm hesitant because we don't have many maintainers for the gazelle stuff, and adding more public things means that changes need to be more careful.

I'll defer to @aignas on this, but otherwise the code changes are mostly fine. It's a very mechanical change.

yushan26:
@dougthor42 There may be gazelle extensions that generate other macros that have python dependencies and requires resolving Python imports.
In Go, it's quite common for extensions like to reuse the Go resolver, so it would be nice to be able to use the Python resolver too, instead of duplicating the Python resolver logic.

gazelle/python/file_parser.go
```diff
@@ -41,7 +41,7 @@ const (
 
 type ParserOutput struct {
 	FileName string
-	Modules  []module
+	Modules  []Module
 	Comments []comment
```
aignas: nit: I think that all of the types in this public `ParserOutput` type should be public if the value can be accessed, hence I think the `comment` should be also exposed.

Could you please do that as well? This is a non-functional change and having a git history where such refactors are grouped together may make it easier to read. Having in the CHANGELOG a line that says something like:
> (gazelle) Types for exposed members of `python.ParserOutput` are now all public.

yushan26:
@aignas Yes good point, just update to make them all public.

PR #2968, opened 2025-06-11
revert: change default bootstrap back to system_python

rickeylev:
Switch the default bootstrap back to system_python, per maintainer discussion.

The main reason is downstream consumers are unlikely to be fully ready for the
usage of raw symlinks (declare_symlink artifacts). APIs to detect them aren't
available until Bazel 8, which makes it difficult for packaging rules, such
as rules_pkg, bazel-lib, or tar rules.

This reverts the core part of commit 9f3512fe0cc6d7229170e45724e22e64be0b8300

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,14 +55,6 @@ END_UNRELEASED_TEMPLATE
 {#v0-0-0-changed}
 ### Changed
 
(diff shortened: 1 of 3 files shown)


aignas (approved):
Thank you!

PR #2969, opened 2025-06-11
fix: support pre-release versions and add new toolchain versions

### LINKED ISSUE #2837: Support pre-release python versions for python-build-standalone toolchains
Jax (among others) likes to use Python runtimes from python-build-standalone that are pre-release versions to help verify upcoming behavior and try out new Python features. e.g. right now, they carry a patch to map 3.14 to 3.14.0a6.

The underlying versions.bzl file doesn't support pre-release versions, though, so they patch 3.14.0 to point to a 3.14.0a6 URL. This works, but it's a bit misleading. It's also problematic if one wanted to have configuration for two versions at once, differentiated by their pre-release version.

(Using patches isn't ideal, but I think they're using an older version that doesn't yet have the TOOL_VERSIONS arg/override -- regardless, the same issues are present).

While https://github.com/bazel-contrib/rules_python/issues/2081 would also address this, I figured a more specific feature request is warranted (this is more narrowly scoped and well defined than 2081).

In any case, the desired outcome should be:
- [x] A user can set `--python_version=3.14.0.a6` on the command line, and have the appropriate runtime used 
- [x] A user can have more than one pre-release configured in e.g. TOOL_VERSIONS

This should be fairly easy to implement and test:
* Use py_reconfig_test to set the python_version and customize toolchains
* Have the test assert sys.version
* Start fixing the errors and misbehaviors that ensue.

cc @kanglant

aignas:
Add latest toolchain builds and attempt adding a beta build.

This shows/tests that we can handle pre-release versions just fine
and we are able to test the toolchain matching. Whilst at it it implements
the static advertising of the remaining interpreter information.

Fixes #2837


python/private/hermetic_runtime_repo_setup.bzl
```diff
@@ -195,6 +195,14 @@ def define_hermetic_runtime_toolchain_impl(
         values = {"collect_code_coverage": "true"},
         visibility = ["//visibility:private"],
     )
+    if not version_info.pre:
+        releaselevel = "final"
+    else:
+        releaselevel = {
+            "a": "alpha",
+            "b": "beta",
+            "c": "candidate",
```
rickeylev: I don't think "c" is correct. IIRC, python uses "a", "b", and "rc" suffixes to its version strings.

rickeylev (commented):
I pulled the dict-literal-in-a-ternary out into a regular if-else. I had to reparse it mentally several times otherwise.

python/private/hermetic_runtime_repo_setup.bzl
aignas: Let me double check how the PEP440 normalization is normalizing `rc`. I can't remember if it is `rc` or `c`.

EDIT: it is "rc" https://github.com/bazel-contrib/rules_python/blob/main/python/private/version.bzl#L358

aignas:
```suggestion
            "rc": "candidate",
```

rickeylev (approved):

Issue #2949, opened 2025-06-01
feat: enable pipstar by default


aignas:
This is just a laundry list for the flag to be flipped by default and enable better dependency graph construction using the marker settings.

- [X] Won't do: Add `expressions_any` to the `env_marker_setting` macro so that we can create many targets under the hood so that each `FeatureFlagInfo` producing target would be processing smaller expressions. At the moment we just do `(expr1) or (expr2) or ...` combination, which feel like it may make the tokenizing more nested and we cannot short-circuit early unless we improve the parser.
- [x] Wait for the #2747 to be finished so that users can have levers to pull to configure the env marker evaluation when parsing requirements files.
- [x] Enable `pipstar` by default in `rules_python` CI
- [ ] Should do: Ask users to test `pipstar` in isolation before the next release.
- [X] (optional) add an example in the documentation how to configure `platform_version` and similar?

Am I missing some steps?

FrankPortman:
Posting here after being directed to comment if I find myself needing `RULES_PYTHON_ENABLE_PIPSTAR=0`.

When testing the `1.7.0` RC I am getting errors like:

```shell
===== stderr start =====
ERROR: Could not find a version that satisfies the requirement triton==3.4.0 (from versions: none)
ERROR: No matching distribution found for triton==3.4.0
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/private/var/tmp/_bazel_fp/01684213bb9c693047272ec8f0247481/external/rules_python+/python/private/pypi/whl_installer/wheel_installer.py", line 183, in <module>
    main()
  File "/private/var/tmp/_bazel_fp/01684213bb9c693047272ec8f0247481/external/rules_python+/python/private/pypi/whl_installer/wheel_installer.py", line 168, in main
    subprocess.run(pip_args, check=True, env=env)
  File "/private/var/tmp/_bazel_fp/01684213bb9c693047272ec8f0247481/external/rules_python++python+python_3_12_3_aarch64-apple-darwin/lib/python3.12/subprocess.py", line 571, in run
    raise CalledProcessError(retcode, process.args,
[...]

PR #2987, opened 2025-06-14
feat(pypi): pip.defaults API for customizing pipstar 1/n

aignas:
Parse env markers in pip.parse using starlark

Summary:
- Allow switching to the Starlark implementation of the marker
  evaluation function.
- Add a way for users to modify the `env` for the marker evaluation when
  parsing the requirements. This can only be done by `rules_python` or
  the root module.
- Limit the platform selection when parsing the requirements files.

Work towards #2747
Work towards #2949
Split out from #2909


python/private/pypi/extension.bzl
```diff
@@ -647,10 +781,63 @@ def _pip_impl(module_ctx):
         # NOTE @aignas 2025-04-15: this is set to be reproducible, because the
         # results after calling the PyPI index should be reproducible on each
         # machine.
-        return module_ctx.extension_metadata(reproducible = True)
+        return module_ctx.extension_metadata(reproducible = False)
```
rickeylev: What makes it no longer reproducible?

python/private/pypi/extension.bzl
```diff
@@ -647,10 +781,63 @@ def _pip_impl(module_ctx):
...
+A platform identifier which will be used as the unique identifier within the extension evaluation.
+If you are defining custom platforms in your project and don't want things to clash, use extension
+[isolation] feature.
+
+[isolation]: https://bazel.build/rules/lib/globals/module#use_extension.isolate
+""",
+    ),
+} | {
+    # The values for PEP508 env marker evaluation during the lock file parsing
+    "env_implementation_name": attr.string(
```
rickeylev: Instead of separate attributes, should we have a dict passed?

python/private/pypi/requirements_files_by_platform.bzl
```diff
@@ -123,6 +113,7 @@ def requirements_files_by_platform(
             be joined with args fined in files.
         python_version: str or None. This is needed when the get_index_urls is
             specified. It should be of the form "3.x.x",
+        platforms: TODO
```
rickeylev: nit: doc todo

python/private/pypi/extension.bzl
```diff
@@ -389,6 +488,39 @@ You cannot use both the additive_build_content and additive_build_content_file a
...
+                env_platform_version = tag.env_platform_version,
+                env_sys_platform = tag.env_sys_platform,
+                os_name = tag.os_name,
+                platform = tag.platform,
+                override = mod.is_root,
+                # TODO @aignas 2025-05-19: add more attr groups:
+                # * for AUTH
+                # * for index config
+                # * for whl selection
+                # * for custom platform definition
```
rickeylev: > add attrs for auth

I thought creds helper and bazel downloader would handle that?

> add attrs for index config

e.g. equivs for --index-url and --extra-index? I think a index-per-package (so private packages don't leak to non-private places) might also be good, and/or index-per-platform (for things like torch, which have separate indexes for different backends)

> add attrs for wheel selection

wdym?

> for custom platform definition

Isn't that what this PR does, essentially?

python/private/pypi/evaluate_markers.bzl
```diff
@@ -30,22 +28,26 @@ SRCS = [
...
 ]
 
-def evaluate_markers(requirements, python_version = None):
+def evaluate_markers(*, requirements, platforms):
     """Return the list of supported platforms per requirements line.
 
     Args:
         requirements: {type}`dict[str, list[str]]` of the requirement file lines to evaluate.
-        python_version: {type}`str | None` the version that can be used when evaluating the markers.
+        platforms: {type}`dict[str | struct]` TODO
```
rickeylev: typo? I think you meant `dict[str, struct]` ?

python/private/pypi/evaluate_markers.bzl
```diff
@@ -30,22 +28,26 @@ SRCS = [
...
         req = requirement(req_string)
-        for platform in platforms:
-            if evaluate(req.marker, env = env(platform_from_str(platform, python_version))):
-                ret.setdefault(req_string, []).append(platform)
+        for platform_str in platform_strings:
+            env = platforms.get(platform_str)
+            if not env:
+                fail("Please define platform: '{}'".format(platform_str))
+
+            if evaluate(req.marker, env = env):
```
rickeylev, line 50: We evaluate during the bzlmod phase to basically pre-filter, right?

e.g if you have `{"linux": "requirements_universal.txt"}`, then we're basically stripping out all the `foo; os=="windows"` lines so that we don't bother traversing (in the index) anything for foo for linux, and also avoid generating any targets for foo (for linux, anyways)

python/private/pypi/extension.bzl
```diff
@@ -907,6 +1094,21 @@ the BUILD files for wheels.
 """,
     implementation = _pip_impl,
     tag_classes = {
+        "default": tag_class(
+            attrs = _default_attrs,
+            doc = """\
+This tag class allows for more customization of how the configuration for the hub repositories is built.
+
+This is still experimental and may be changed or removed without any notice.
```
rickeylev: There's an include that will add the warning etc text.

```suggestion

:::{include} /_includes/experimtal_api.md
:::
```

I think that's the right syntax.

python/private/pypi/extension.bzl
aignas: Sorry this is a leftover from my testing.

aignas:
```suggestion
        return module_ctx.extension_metadata(reproducible = True)
```

python/private/pypi/extension.bzl
aignas: Done, much cleaner.

python/private/pypi/extension.bzl
aignas: Added extra comments.

python/private/pypi/evaluate_markers.bzl
aignas: Correct.

PR #2988, opened 2025-06-15
feat(pypi): pip.defaults API for customizing repo selection 2/n

aignas:
WIP: stacked on #2987

This is adding `constraint_values` attribute to `pip.configure` and is
threading it all the way down to the generation of `BUILD.bazel` file of for
config settings used in the hub repository.

Out of scope:
- Passing `flag_values` or target settings. I am torn about it - doing it in
  this PR would flesh out the design more, but at the same time it might become
  harder to review.
- `whl_target_platforms` and `select_whls` is still unchanged, not sure if it
  is related to this attribute addition.

Work towards #2747
Work towards #2548 
Work towards #260


rickeylev:
Haven't looked at code, just saw the PR description.

Passing target settings will be better overall. They're more flexible and easier to customize for users.

I suppose if they can pass a flag setting (the name at least), then it's just as flexible. But if a custom name can be passed, then a custom target_setting should also be passable.

aignas:
Will do this later, because of how much code I would have to change in one go. Hopefully this can be done in 3/n or 4/n. My plan is to roughly spike out the implementation with almost passing tests, little docs and then tidy up each and every one by one.

EDIT: done in https://github.com/bazel-contrib/rules_python/pull/2990

python/private/pypi/extension.bzl
```diff
@@ -640,17 +795,77 @@ def _pip_impl(module_ctx):
...
+    "arch_name": attr.string(
+        doc = """\
+The CPU architecture name to be used.
+
+:::{note}
+Either this or {attr}`env_platform_machine` should be specified.
+:::
+""",
+    ),
+    "constraint_values": attr.label_list(
```
rickeylev, line 815: I think this should be a string_list, not label list?

If it's a label, it'll trigger repo-evaluation, even if it goes unused.

python/private/pypi/hub_repository.bzl
```diff
@@ -83,6 +84,10 @@ hub_repository = repository_rule(
 The list of packages that will be exposed via all_*requirements macros. Defaults to whl_map keys.
 """,
         ),
+        "platform_constraint_values": attr.string_list_dict(
```
rickeylev, line 87: Why does a `{"platform name": [list of constraint labels]}` need to be passed?

When the hub generates selects, it just needs the matching constraints and the label they map to. The "human friendly" platform string isn't needed.

aignas: The `whl_config_setting` right now is using the human readable values. We could in theory just use the `target_settings` directly. I'll think about this a little bit more.

python/private/pypi/extension.bzl
aignas: This will be changed to `target_settings` in the next thing up the stack.

However, when we will have `target_settings` I would have expected that we would need the `attr.label_list` because we want to ensure that the users don't need to use canonical representation of the labels. The fetch would not matter that much, because this affects only the hub repo and the extension itself, both of which are reasonably cheap.

What do you mean about `repo-evaluation`? I thought that we can use `@platforms//os:foo` and does not cause any fetching?

python/private/pypi/config_settings.bzl
```diff
@@ -126,31 +126,26 @@ def config_settings(
             configure config settings for.
         osx_versions (list[str]): The list of OSX OS versions to configure
             config settings for.
-        target_platforms (list[str]): The list of "{os}_{cpu}" for deriving
-            constraint values for each condition.
+        platform_constraint_values: {type}`dict[str, list[str]]` the constraint
+            values to use instead of the default ones.
```
rickeylev:
```suggestion
        platform_constraint_values: {type}`dict[str, list[str]]` the constraint
            values to use instead of the default ones. Key are platform names
            (a human-friendly platform string). Values are lists of
            `constraint_value` label strings.
```

python/private/pypi/extension.bzl
rickeylev: I thought it does cause repo fetching? Though, when I think back, maybe I'm confusing label_list() in repo/bzlmod with `rctx.path(<some label>)`.

In any case, lets double check to verify.

python/private/pypi/hub_repository.bzl
rickeylev: SGTM. I don't mind the name being wired through for intelligibility. The main thing is I don't want the value to have real meaning (e.g. we could just replace them with platform_1, platform_2, etc).

python/private/pypi/extension.bzl
```diff
@@ -788,6 +810,12 @@ The CPU architecture name to be used.
 :::{note}
 Either this or {attr}`env` `platform_machine` key should be specified.
 :::
+""",
+    ),
+    "constraint_values": attr.label_list(
```
rickeylev, line 815: Followup thought for subsequent PR: replace constraint_values and target_settings with simply "config_settings" ?

The custom platform stuff in the toolchains has target_compatible_with (constraints) and target_settings because that's what the toolchain() rule accepts. However, for the pip stuff, it just needs something valid for a select() key.

rickeylev (approved):

python/private/pypi/extension.bzl
aignas: Correct, I like this suggestion. `config_settings` does sound more natural.

PR #2990, opened 2025-06-15
feat: support arbitrary target_settings in our platforms 3/n

aignas:
With this PR we can support arbitrary target settings instead of just
plain `constraint_values`. We still have custom logic to ensure that all
of the tests pass. However, the plan is to remove those tests once we
have simplified the wheel selection mechanisms and the `pkg_aliases`
macro. I.e. if we have at most 1 wheel per platform that the `pypi`
bzlmod extension passes to the `pkg_aliases` macro, then we can just
have a simple `selects.with_or` where we list out all of the target
platform values.

This PR may result in us creating more targets but that is the price
that we have to pay if we want to do this incrementally.

Work towards #2747
Work towards #2548
Work towards #260

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,7 +64,7 @@ END_UNRELEASED_TEMPLATE
 ### Added
 * (pypi) To configure the environment for `requirements.txt` evaluation, use the newly added
   developer preview of the `pip.default` tag class. Only `rules_python` and root modules can use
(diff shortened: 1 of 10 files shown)


rickeylev:
I gave this a skim, but started to lose track of the diff, so I'll hold off on reviewing in depth. Mostly LGTM, though.

PR #3010, opened 2025-06-19
fix(toolchains): use posix-compatible exec -a alternative

rickeylev:
The `exec -a` command doesn't work in dash, the default shell for Ubuntu/debian.

To work around, use `sh -c`, which is posix and dash compatible. This allows changing
the argv0 while invoking a different command. Also adds a test to verify the the
runtime_env toolchain works with bootstrap script.

Fixes https://github.com/bazel-contrib/rules_python/issues/3009


aignas (approved):
LGTM

PR #3011, opened 2025-06-19
feat(pypi): generate filegroup with all extracted wheel files

rickeylev:
Adds a filegroup with all the files that came from the extracted wheel.

This has two benefits over using `whl_filegroup`: it avoids copying the wheel
and makes the set of files directly visible to the analysis phase.

Some wheels are multiple gigabytes in size (e.g. torch, cuda, tensorflow), so
avoiding the copy and archive processing saves a decent amount of time.

Knowing the specific files at analysis time is generally beneficial. The
particular case I ran into was the CC rules were unhappy with a TreeArtifact
of header files because they couldn't enforce some check about who was
properly providing headers that were included (layering check?).

Another example is using the unused_inputs_list optimization, which allows
an action to ignore inputs that aren't actually used. e.g. an action could
take all the wheel's files as inputs, only care about the headers, and then
tell bazel all the non-header files aren't relevant, and thus changes to
other files don't re-run the thing that only cares about headers.

### DIFF
--- a/python/private/pypi/labels.bzl
+++ b/python/private/pypi/labels.bzl
@@ -14,6 +14,7 @@
 
 """Constants used by parts of pip_repository for naming libraries and wheels."""
 
+EXTRACTED_WHEEL_FILES = "extracted_whl_files"
 WHEEL_FILE_PUBLIC_LABEL = "whl"
 WHEEL_FILE_IMPL_LABEL = "_whl"
 PY_LIBRARY_PUBLIC_LABEL = "pkg"
--- a/python/private/pypi/whl_library_targets.bzl
+++ b/python/private/pypi/whl_library_targets.bzl
@@ -24,6 +24,7 @@ load(
     ":labels.bzl",
     "DATA_LABEL",
     "DIST_INFO_LABEL",
+    "EXTRACTED_WHEEL_FILES",
     "PY_LIBRARY_IMPL_LABEL",
     "PY_LIBRARY_PUBLIC_LABEL",
     "WHEEL_ENTRY_POINT_PREFIX",
@@ -101,8 +102,16 @@ def whl_library_targets(
         srcs_exclude = [],
         tags = [],
         filegroups = {
-            DIST_INFO_LABEL: ["site-packages/*.dist-info/**"],
-            DATA_LABEL: ["data/**"],
+            EXTRACTED_WHEEL_FILES: dict(
+                include = ["**"],
+                exclude = ["*.whl"],
+            ),
+            DIST_INFO_LABEL: dict(
+                include = ["site-packages/*.dist-info/**"],
+            ),
+            DATA_LABEL: dict(
+                include = ["data/**"],
+            ),
         },
         dependencies = [],
         dependencies_by_platform = {},
@@ -168,10 +177,11 @@ def whl_library_targets(
     tags = sorted(tags)
     data = [] + data
 
-    for filegroup_name, glob in filegroups.items():
+    for filegroup_name, glob_kwargs in filegroups.items():
+        glob_kwargs = {"allow_empty": True} | glob_kwargs
         native.filegroup(
             name = filegroup_name,
-            srcs = native.glob(glob, allow_empty = True),
+            srcs = native.glob(**glob_kwargs),
             visibility = ["//visibility:public"],
         )
 


python/private/pypi/labels.bzl
```diff
@@ -14,6 +14,7 @@
 
 """Constants used by parts of pip_repository for naming libraries and wheels."""
 
+EXTRACTED_WHEEL_FILES = "extracted_whl_files"
```
aignas, line 17: You still need to modify the `pkg_aliases` macro to include the extra alias.

python/private/pypi/whl_library_targets.bzl
```diff
@@ -101,8 +102,16 @@ def whl_library_targets(
         srcs_exclude = [],
         tags = [],
         filegroups = {
-            DIST_INFO_LABEL: ["site-packages/*.dist-info/**"],
-            DATA_LABEL: ["data/**"],
+            EXTRACTED_WHEEL_FILES: dict(
+                include = ["**"],
```
aignas: nit, in the repo phase we could parse the `RECORD` file and include those files instead? In the case where the whl is built from an sdist, this may include the extra `sdist` file under some configurations. Maybe we should exclude these things? The wheel patching thing has some logic to exclude it correctly.

aignas (changes requested):

python/private/pypi/whl_library_targets.bzl
```diff
@@ -101,8 +102,16 @@ def whl_library_targets(
...
+                exclude = ["*.whl"],
```
aignas: I think the `name` attribute is the wheel filename, so we can be precise with the exclude. If we can pass the sdist filename to this macro, then it would be good as well.
```suggestion
                exclude = [name],
```

python/private/pypi/whl_library_targets.bzl
rickeylev: Ah, I forgot about sdists. And didn't realize they get extracted out into here (and then built into a wheel?).

Rather than parse RECORD, how about making the glob conditional? e.g. `if name.endswith(".whl"): glob-include **`

The RECORD is just a simple manifest of everything in the wheel file, right? i.e. `glob(**)` and RECORD should expand to the same set of files?

python/private/pypi/whl_library_targets.bzl
rickeylev: Is this true for sdists? I guess it doesn't matter -- we'd want to exclude the sdist anyways.

python/private/pypi/labels.bzl
rickeylev: Done.

python/private/pypi/pkg_aliases.bzl
```diff
@@ -151,6 +152,7 @@ def pkg_aliases(
         WHEEL_FILE_PUBLIC_LABEL: WHEEL_FILE_IMPL_LABEL if group_name else WHEEL_FILE_PUBLIC_LABEL,
         DATA_LABEL: DATA_LABEL,
         DIST_INFO_LABEL: DIST_INFO_LABEL,
+        EXTRACTED_WHEEL_FILES: EXTRACTED_WHEEL_FILES,
```
rickeylev, line 155: todo: add a test for this. Just need to have some that references it.

python/private/pypi/whl_library_targets.bzl
aignas: name will always end with `.whl` because the sdist gets built into a wheel. Maybe a better way would be to remove the sdist or just add it to the excludes list.

rickeylev: I like the idea of excluding it from the glob. I think I figured out how to pass the sdist filename (if one exists) along.

python/private/pypi/whl_library_targets.bzl
rickeylev: Done.

python/private/pypi/pkg_aliases.bzl
rickeylev: Actually, the existing tests verify it shows up. Done

python/private/pypi/whl_library_targets.bzl
rickeylev: I plumbed through the sdist filename from whl_library to also be excluded.

rickeylev (commented):
PTAL

aignas (approved):

PR #3029, opened 2025-06-25
fix: delete BUILD et al files from pypi sourced dependencies

rickeylev:
Sometimes wheels mistakenly contain BUILD files (or other special Bazel files).
When extracted, these interfere with how we expected the repo to look. In particular,
globs no longer match correctly because BUILD files create a Bazel package boundary.

To fix, delete these files. They aren't generally usable, since they can't know what
version of Bazel, rule_python, or similar, is consuming them.

Fixes https://github.com/bazel-contrib/rules_python/issues/2782


rickeylev:
Ideas for testing this? Aside from checking in a pre-built whl (I'm very much against any checked in artifacts after what happened to xz)

Ah -- a repository rule to transform a file structure into a zip, to then pass onto whl_library? We can then define an entirely arbitrary layout of files verbatim and have it automatically turned into a wheel for testing.

rickeylev:
I added a helper to create a whl from a directory tree. This is really nice for testing whl_library!

We can now integration test arbitrary wheel file structures and contents without having to download anything.

python/private/internal_dev_deps.bzl
```diff
@@ -28,6 +30,17 @@ def _internal_dev_deps_impl(mctx):
...
     runtime_env_repo(name = "rules_python_runtime_env_tc_info")
 
+    whl_from_dir_repo(
+        name = "whl_with_build_files",
+        root = "//tests/whl_with_build_files:testdata/BUILD.bazel",
+        output = "somepkg-1.0-any-none-any.whl",
+    )
+    whl_library(
+        name = "somepkg_with_build_files",
+        whl_file = "@whl_with_build_files//:somepkg-1.0-any-none-any.whl",
```
aignas, line 40: Finally we can make use of this thing that I've added a long time ago!

python/private/pypi/whl_library.bzl
```diff
@@ -249,6 +249,7 @@ def _whl_library_impl(rctx):
 
     whl_path = None
     if rctx.attr.whl_file:
+        rctx.watch(rctx.attr.whl_file)
```
aignas, line 252: Ah, good call.

tests/support/whl_from_dir/whl_from_dir_repo.bzl
```diff
@@ -0,0 +1,50 @@
...
+            "-0",  # Skip compressing
+            "-X",  # Don't store file time or metadata
+            str(output),
+            "-r",
+            ".",
+        ],
+    )
+    rctx.file("BUILD.bazel", 'exports_files(glob(["*"]))')
+
+whl_from_dir_repo = repository_rule(
```
aignas, line 30: I remember somebody also requested a feature similar to this and wanted to add `PROJECT_ROOT`. Maybe this thing would be useful to them?

Ref: #2517

aignas (approved):
Nice!

Please remember to add a line in the CHANGELOG.

tests/support/whl_from_dir/whl_from_dir_repo.bzl
rickeylev: Yeah, I had a similar thought. I spent some time yesterday thinking about different ways to map custom sources for dependencies into the pip integration. I think https://github.com/bazel-contrib/rules_python/issues/2986 is about the closest we have for this topic, so I'll post details there.

rickeylev:
Changelog updated.

Merging.

PR #3036, opened 2025-06-28
fix(gazelle) Fix dependency added as both deps and pyi_deps

amartani:
Fix an issue in https://github.com/bazel-contrib/rules_python/pull/3014 where a dependency may end up being added in both `deps` and `pyi_deps`, in cases where the regular and the type-checking import refer to different python modules on the same `py_library` target. Other cases are already deduplicated earlier on, but this case can only be deduplicated in the resolve phase.

(No new changelog entry since this is a fix to an unreleased feature that is already in the changelog)


gazelle/python/testdata/type_checking_imports_across_packages/a/bar.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2023 The Bazel Authors. All rights reserved.
```
dougthor42: Nit: copyright 2025 here and elsewhere

aignas: actually, new files don't need the copyright header anymore, so it would be better to leave it out alltogether.

dougthor42: Oh good! When did that change?

amartani: 👍  removed all copyright headers

dougthor42 (approved):

gazelle/python/testdata/type_checking_imports_across_packages/a/bar.py
aignas: @dougthor42, when `rules_python` moved to `bazel-contrib`. I think there was a thread on slack about this.

PR #3045, opened 2025-07-01
fix: Don't let deprecated test targets get matched by '...'

robinlinden:
This fixes "target '//foo_test' is deprecated: Use 'foo.test' instead. The '*_test' target will be removed in the next major release." being warned about once per `compile_pip_requirement` call when running `bazel test ...`.


aignas (approved):
Thanks! LGTM.

PR #3053, opened 2025-07-03
fix: Updating Python toolchains to patch CVE-2025-47273

aaronmaxlevy:
Updating to a slightly newer build (20250612 instead of 20250610) of several Python toolchains that includes a newer version of setuptools that is no longer vulnerable to CVE-2025-47273.

Also added support for Python 3.13.5, since a patched toolchain build for 3.13.4 is not available (and since 3.13.5 was released).

See https://github.com/astral-sh/python-build-standalone/commit/5cc924bf04b73004c57bf438476255c1b5b63e9f#diff-860ea5e06ac2e2191008bbf2de9b216ed11533780f270b5e0cdfc31b37b3b3df and https://github.com/astral-sh/python-build-standalone/releases/tag/20250612


aignas (approved):
Thanks!

PR #3054, opened 2025-07-04
fix(pypi): correctly handle custom names in pipstar platforms

aignas:
Before it seems that we were relying on particular names in the pipstar
platforms. This ensures that we rely on this less. Whilst at it fix a
few typos and improve the formatting of the code.

Work towards #2949
Work towards #2747

### DIFF
--- a/python/private/pypi/extension.bzl
+++ b/python/private/pypi/extension.bzl
@@ -387,12 +387,6 @@ def _configure(config, *, platform, os_name, arch_name, config_settings, env = {
             if key not in _SUPPORTED_PEP508_KEYS:
                 fail("Unsupported key in the PEP508 environment: {}".format(key))
 
-        if not os_name:
-            fail("'os_name' is required")
-
-        if not arch_name:
-            fail("'arch_name' is required")
-
         config["platforms"][platform] = struct(
             name = platform.replace("-", "_").lower(),
             os_name = os_name,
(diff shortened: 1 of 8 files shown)


aignas:
The `requirement_files_by_platform` is still a little broken as we can see in #3063.

python/private/pypi/extension.bzl
```diff
@@ -383,6 +387,12 @@ def _configure(config, *, platform, os_name, arch_name, config_settings, env = {
             if key not in _SUPPORTED_PEP508_KEYS:
                 fail("Unsupported key in the PEP508 environment: {}".format(key))
 
+        if not os_name:
+            fail("'os_name' is required")
```
rickeylev: Were os_name and arch_name effectively required before, or is this new? Though, I think this pip.defaults API is still experimental, right?

rickeylev (approved):

python/private/pypi/extension.bzl
aignas: This is new. It is experimental, but I've been dragging some of the hidden constants from various parts of the PyPI code into the internally defined `pip.default` instances. This has been really useful in that sense. I think maybe I want to revert this requirement. It is not going to be compatible with the builder pattern.

python/private/pypi/extension.bzl
```diff
@@ -383,6 +387,12 @@ def _configure(config, *, platform, os_name, arch_name, config_settings, env = {
...
+
+        if not arch_name:
+            fail("'arch_name' is required")
+
```
aignas:
```suggestion
```

PR #3057, opened 2025-07-05
feat(gazelle): Gazelle plugin generates py_proto_library

shaldengeki:
Fixes https://github.com/bazel-contrib/rules_python/issues/2994.

Please go over this with a fine-toothed comb! This is my first contribution to `rules_python` / the gazelle plugin, and while I've worked in Gazelle before, I'm pretty unfamiliar with the plugin's architecture.

This adds support in the Gazelle plugin for generating `py_proto_library` rules automatically, if there are any `proto_library` rules detected in a given package. We do this via a new Gazelle directive, `python_generate_proto`, which defaults to `true`, and controls whether these rules are generated.

See the tests in `testdata/directive_python_generate_proto` for examples.

By default, we source the `py_proto_library` rule from the `@protobuf` repository. I think this the intended long-term home of the rule? Users are expected to use `gazelle:map_kind` to change this if need be.

I haven't done anything here to support resolution of imports of `py_proto_library`. I think this is worth landing first, to save folks from having to maintain these by hand. But this should lay the foundation for resolving that in https://github.com/bazel-contrib/rules_python/issues/1703.


CHANGELOG.md
```diff
@@ -91,6 +91,8 @@ END_UNRELEASED_TEMPLATE
 * (gazelle) New directive `gazelle:python_generate_pyi_deps`; when `true`,
   dependencies added to satisfy type-only imports (`if TYPE_CHECKING`) and type
   stub packages are added to `pyi_deps` instead of `deps`.
+* (gazelle) New directive `gazelle:python_generate_proto`; when `true` (by default),
```
dougthor42: Let's start this off with a `False` default. Err on the side of caution and whatnot.

gazelle/python/testdata/directive_python_generate_proto/test.yaml
```diff
@@ -0,0 +1,17 @@
+# Copyright 2023 The Bazel Authors. All rights reserved.
```
dougthor42: I found out recently that these license headers are no longer needed. Let's remove them.

gazelle/python/generate.go
```diff
@@ -551,3 +555,49 @@ func ensureNoCollision(file *rule.File, targetName, kind string) error {
...
+	for _, r := range args.File.Rules {
+		if r.Kind() == "py_proto_library" {
+			pyProtoRules[r.Name()] = false
+		}
+	}
+
+	// Now generate a py_proto_library for each proto_library.
+	emptySiblings := treeset.Set{}
+	for _, protoRuleName := range protoRuleNames {
+		pyProtoLibraryName := protoRuleName + "_py_pb2"
```
dougthor42: [OK to punt to a followup Issue/PR] Let's expose this suffix as another directive, just like we have `python_library_naming_convention`, `python_binary_naming_convention`, and `python_test_naming_convention`, we should have `python_proto_naming_convention`.

gazelle/python/testdata/directive_python_generate_proto/test7_removes_when_unnecessary/BUILD.out
```diff
@@ -0,0 +1 @@
+# gazelle:python_generate_proto true
```
dougthor42, line 1: Add test case: "removes just the `py_proto_library` when `python_generate_proto=false` and a `proto_library` exists"

gazelle/python/testdata/directive_python_generate_proto/README.md
```diff
@@ -0,0 +1,9 @@
+# Directive: `python_generate_proto`
+
+This test case asserts that the `# gazelle:python_generate_proto` directive
+correctly:
+
+1.  Uses the default value when `python_generate_proto` is not set.
+2.  Generates (or not) `py_proto_library` when `python_generate_proto` is set, based on whether a proto is present.
```
dougthor42, line 7: Add test case: python code can import generated `py_proto_library` target.

So, check that we resolve the py_library (or py_binary, py_test) dependencies _after_ the py_proto_library target is generated from the proto_library.

This could be done with a "from scratch" test case that starts off with two files: `foo.proto` and `bar.py`, with:

```python
# bar.py

import foo_pb2  # should be deps=[":foo_py_pb2"]
```

gazelle/python/testdata/directive_python_generate_proto/test1_default_with_proto/foo.proto
```diff
@@ -0,0 +1,7 @@
+syntax = "proto3";
+
+package foo;
+
+message Foo {
+    string bar = 1;
+}
```
dougthor42: Nit: LF@EOF

gazelle/README.md
```diff
@@ -224,6 +224,8 @@ Python-specific directives are as follows:
...
 | `# gazelle:python_generate_pyi_deps`                                                                                                                                                                                                                                                            | `false` |
 | Controls whether to generate a separate `pyi_deps` attribute for type-checking dependencies or merge them into the regular `deps` attribute. When `false` (default), type-checking dependencies are merged into `deps` for backward compatibility. When `true`, generates separate `pyi_deps`. Imports in blocks with the format `if typing.TYPE_CHECKING:`/`if TYPE_CHECKING:` and type-only stub packages (eg. boto3-stubs) are recognized as type-checking dependencies. |
+| `# gazelle:python_generate_proto`                                                                                                                                                                                                                                                            | `true` |
```
dougthor42: Please add a detail section down around L485. Example: https://github.com/bazel-contrib/rules_python/blob/2b5e6f54314d2110490724eb707436355b1938fc/gazelle/README.md?plain=1#L459

gazelle/python/kinds.go
```diff
@@ -101,4 +108,10 @@ var pyLoads = []rule.LoadInfo{
 			pyTestKind,
 		},
 	},
+	{
+		Name: "@protobuf//bazel:py_proto_library.bzl",
```
dougthor42: Many packages call this `com_google_protobuf` instead of just `protobuf`. Eg people's `MODULE.bazel` will have:

```
bazel_dep(name = "protobuf", version = <VERSION>, repo_name = "com_google_protobuf")
```

for [compat. with WORKSPACE](https://github.com/protocolbuffers/protobuf?tab=readme-ov-file#bazel-with-bzlmod). Does using `@protobuf` here cause issues with that?

gazelle/python/testdata/directive_python_generate_proto/README.md
dougthor42: > I haven't done anything here to support resolution of imports of `py_proto_library`.

Oh I just re-read this part of your PR description, sorry!

No action needed here.

gazelle/python/kinds.go
shaldengeki: Yeah, this is a good point. The way that the golang extension handles this is by using the `moduleToApparentName` handler, which is supposed to resolve the configured name for a given module name:
https://github.com/bazel-contrib/bazel-gazelle/blob/master/language/go/kinds.go#L178-L181
https://github.com/bazel-contrib/bazel-gazelle/blob/v0.44.0/language/lang.go#L129-L134

I could use that, but we'd have to pick some default in case the `MODULE.bazel` doesn't exist. In that case, I think we should fallback to `com_google_protobuf`, because it's extremely likely the user is using `WORKSPACE` in that case. That sound OK to you?

shaldengeki:
While writing a test covering the bzlmod integration I ran into a small blocker, so I PRed a fix for that separately: https://github.com/bazel-contrib/rules_python/pull/3060

gazelle/python/generate.go
shaldengeki: Ah, TIL -- this is pretty neat. Yeah, happy to land this as a follow-up.

shaldengeki:
@dougthor42 think I'll need your help deciding what to do here, apologies!

I wanted to add some tests that cover the bzlmod use-cases. I can't do that until gazelle gets upgraded, and it sounds like that's quite thorny right now. My inclination is to delete the bzlmod tests (they do pass, fwiw) and land this PR without 'em, to let you plug away at the upgrade at your own pace. Then after the upgrade's landed I can reinstate the tests. Does that sound reasonable to you?

gazelle/python/testdata/directive_python_generate_proto/test7_removes_when_unnecessary/BUILD.out
aignas: Won't this cause an issue with the defaults? The default is `false` and the users may have manually written `py_proto_library` and we would remove them when they update the gazelle plugin.

shaldengeki: Hmmm. I see your point @aignas - we probably want `python_generate_proto=false` to mean "Gazelle just shouldn't manage py_proto_library", and not "Gazelle should ensure that no py_proto_library are ever present". That's, for example, how the golang extension works:

https://github.com/bazel-contrib/bazel-gazelle/blob/master/language/go/generate.go#L537-L541
https://github.com/bazel-contrib/bazel-gazelle/blob/master/language/go/config.go#L231

I'll alter the logic here to make the extension ignore any pre-existing `py_proto_library` when `python_generate_proto=false`, then.

shaldengeki:
All right, I removed the bzlmod tests & reverted the gazelle upgrade in this PR. Happy to restore those later -- let me know what you think.

shaldengeki:
Looks like the build failure is unrelated:
```
(16:04:21) ERROR: /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/ec321eb2cc2d0f8f91b676b6d4c66c29/external/abseil-cpp~/absl/hash/BUILD.bazel:35:11: Compiling absl/hash/internal/hash.cc [for tool] failed: (Exit 1): gcc failed: error executing CppCompile command (from target @@abseil-cpp~//absl/hash:hash)
# ...
external/abseil-cpp~/absl/hash/internal/hash.h:66:10: fatal error: filesystem: No such file or directory
```
Happening, for example, in this PR, too: https://github.com/bazel-contrib/rules_python/pull/3075

gazelle/README.md
```diff
@@ -484,6 +486,32 @@ def py_test(name, main=None, **kwargs):
 )
 ```
 
+#### Directive: `python_generate_proto`:
+When `# gazelle:python_generate_proto true`, Gazelle will generate one `py_proto_library` for each `proto_library`, generating Python clients for protobuf in each package. By default this is turned off. Gazelle will also generate a load for the `py_proto_library` - attempting to detect the configured name for the `@protobuf` / `@com_google_protobuf` repo in your `MODULE.bazel`, and otherwise falling back to `@com_google_protobuf` for compatibility with `WORKSPACE`.
+
+For example, in a package with `# gazelle:python_generate_proto true` and a `foo.proto`, if you have both the proto extension and the Python extension loaded into Gazelle, you'll get something like:
```
dougthor42: Please wrap to ~80 chars.

gazelle/README.md
```diff
@@ -484,6 +486,32 @@ def py_test(name, main=None, **kwargs):
 )
 ```
 
+#### Directive: `python_generate_proto`:
+When `# gazelle:python_generate_proto true`, Gazelle will generate one `py_proto_library` for each `proto_library`, generating Python clients for protobuf in each package. By default this is turned off. Gazelle will also generate a load for the `py_proto_library` - attempting to detect the configured name for the `@protobuf` / `@com_google_protobuf` repo in your `MODULE.bazel`, and otherwise falling back to `@com_google_protobuf` for compatibility with `WORKSPACE`.
```
dougthor42: Nit: "generate a load statement for"

gazelle/README.md
```diff
@@ -484,6 +486,32 @@ def py_test(name, main=None, **kwargs):
 )
 ```
 
+#### Directive: `python_generate_proto`:
```
dougthor42, line 489: Nit: newline after heading

gazelle/python/generate.go
```diff
@@ -88,12 +88,20 @@ func (py *Python) GenerateRules(args language.GenerateArgs) language.GenerateRes
 		}
 	}
 
+	var result language.GenerateResult
+	result.Gen = make([]*rule.Rule, 0)
+
+	pythonProjectRoot := cfg.PythonProjectRoot()
+	visibility := cfg.Visibility()
```
dougthor42: Not that it really matters, but I'm curious as to why these were moved around.

Eg. why not put `if cfg.GenerateProto() {` down near L230 `collisionErrors := singlylinkedlist.New()`?

gazelle/python/generate.go
dougthor42: Could you make an Issue (feature request) for this?

gazelle/python/kinds.go
dougthor42: What you've changed to LGTM, thanks!

dougthor42 (commented):
Thanks for the details in README.md!

> I wanted to add some tests that cover the bzlmod use-cases. I can't do that until gazelle gets upgraded, and it sounds like that's quite thorny right now. My inclination is to delete the bzlmod tests (they do pass, fwiw) and land this PR without 'em, to let you plug away at the upgrade at your own pace. Then after the upgrade's landed I can reinstate the tests. Does that sound reasonable to you?

Yes, the gazelle and rules_go upgrades are a PITA...

Can you elaborate on why you you can't add the bzlmod tests _without bumping gazelle / rules_go / go version_? From what I see in cdeafde0b43884a527d321f78e884d8f543fdbd0, I don't see any reason why the bzlmod tests won't work.

shaldengeki:
> Can you elaborate on why you you can't add the bzlmod tests without bumping gazelle / rules_go / go version? From what I see in https://github.com/bazel-contrib/rules_python/commit/cdeafde0b43884a527d321f78e884d8f543fdbd0, I don't see any reason why the bzlmod tests won't work.

Yep! Should've made this clear earlier. You can reproduce the errors I ran into by checking out https://github.com/bazel-contrib/rules_python/pull/3057/commits/086c6aafa6aec944cbf232608e6cdad7595c3af3 (the last commit before I undid the upgrade & removed the tests), and undoing the upgrade:

```
$ bazel test //...
# ...
FAIL: //python:python_test_directive_python_generate_proto_bzlmod_protobuf_renamed (Exit 1) (see /home/charles/.cache/bazel/_bazel_charles/25da5a3f5a375d8b77b073daf0c17156/execroot/_main/bazel-out/k8-fastbuild/testlogs/python/python_test_directive_python_generate_proto_bzlmod_protobuf_renamed/test.log)
FAIL: //python:python_test_directive_python_generate_proto_bzlmod_protobuf (Exit 1) (see /home/charles/.cache/bazel/_bazel_charles/25da5a3f5a375d8b77b073daf0c17156/execroot/_main/bazel-out/k8-fastbuild/testlogs/python/python_test_directive_python_generate_proto_bzlmod_protobuf/test.log)
INFO: From Testing //python:python_test_directive_python_generate_proto_bzlmod_protobuf:
==================== Test output for //python:python_test_directive_python_generate_proto_bzlmod_protobuf:
--- FAIL: TestGazelleBinary (0.00s)
    --- FAIL: TestGazelleBinary/directive_python_generate_proto_bzlmod_protobuf (0.00s)
        python_test.go:169: expected gazelle exit code: 0
            got: 1
        python_test.go:179: expected gazelle stderr: 
            got: gazelle: -repo_root not specified, and WORKSPACE cannot be found: file does not exist
        python_test.go:144: "" exists
        python_test.go:144: "/directive_python_generate_proto_bzlmod_protobuf" exists
        python_test.go:144: "/directive_python_generate_proto_bzlmod_protobuf/BUILD" exists
        python_test.go:144: "/directive_python_generate_proto_bzlmod_protobuf/MODULE.bazel" exists
        python_test.go:144: "/directive_python_generate_proto_bzlmod_protobuf/README.md" exists
        python_test.go:144: "/directive_python_generate_proto_bzlmod_protobuf/foo.proto" exists
        python_test.go:144: "/directive_python_generate_proto_bzlmod_protobuf/test.yaml" exists
FAIL
================================================================================
INFO: From Testing //python:python_test_directive_python_generate_proto_bzlmod_protobuf_renamed:
==================== Test output for //python:python_test_directive_python_generate_proto_bzlmod_protobuf_renamed:
--- FAIL: TestGazelleBinary (0.00s)
    --- FAIL: TestGazelleBinary/directive_python_generate_proto_bzlmod_protobuf_renamed (0.01s)
        python_test.go:169: expected gazelle exit code: 0
            got: 1
        python_test.go:179: expected gazelle stderr: 
            got: gazelle: -repo_root not specified, and WORKSPACE cannot be found: file does not exist
        python_test.go:144: "" exists
        python_test.go:144: "/directive_python_generate_proto_bzlmod_protobuf_renamed" exists
        python_test.go:144: "/directive_python_generate_proto_bzlmod_protobuf_renamed/BUILD" exists
        python_test.go:144: "/directive_python_generate_proto_bzlmod_protobuf_renamed/MODULE.bazel" exists
        python_test.go:144: "/directive_python_generate_proto_bzlmod_protobuf_renamed/README.md" exists
        python_test.go:144: "/directive_python_generate_proto_bzlmod_protobuf_renamed/foo.proto" exists
        python_test.go:144: "/directive_python_generate_proto_bzlmod_protobuf_renamed/test.yaml" exists
FAIL
================================================================================
```

which reads to me like Gazelle doesn't know about bzlmod, and is just looking for the WORKSPACE.

gazelle/python/generate.go
shaldengeki: Yeah, that's a good point, let me do that.

shaldengeki: (To answer your question: I moved this stuff up here because that's how I started working on this PR. I added an empty method call at first, and pulled things up as I realized I needed them.)

shaldengeki:
@dougthor42 one change I just made -- I double-checked how the golang & java extensions work, and they both generate e.g. `foo_proto` -> `foo_(go|java)_proto`, so I made our default here `foo_py_proto`, instead of `foo_py_pb2` (which came from my own idiosyncratic naming in my personal repo).

I think that makes sense? And it'll make it just a tiny bit easier to resolve imports down the line.

gazelle/python/generate.go
shaldengeki: Filed! https://github.com/bazel-contrib/rules_python/issues/3081

dougthor42:
https://bazel.build/reference/be/protocol-buffer#py_proto_library:

> The convention is to name the `py_proto_library` rule `foo_py_pb2`, when it is wrapping `proto_library` rule `foo_proto`

IIRC this comes from the fact that `protoc` makes `foo_pb2.py` files and they're imported in Python with `import foo_pb2`.

shaldengeki:
Ah, dang it. I knew that came from somewhere.

It is... an unfortunate naming convention, I'll venture? I can put it back, but it's weird because a) it doesn't match other languages, and b) it doesn't even fit the actual naming of the generated file, which is like `foo_pb2.py` (while the convention is `foo_py_pb2`).

Let me know if you'd prefer I revert it.

dougthor42 (commented):
> Let me know if you'd prefer I revert it.

Yeah let's revert 3e0fdcc and we'll use #3081 to support naming things like `foo_py_proto`

Otherwise this LGTM. We just have to figure out that CI failure...

dougthor42:
> which reads to me like Gazelle doesn't know about bzlmod, and is just looking for the WORKSPACE.

Ah yes I see the error and repro'd it locally. Turns out you still need a dummy (empty) WORKSPACE file in the test dir. You can re-add those test cases if you'd like, as long as you're not bumping versions.

With the empty WORKSPACE file, things pass for me locally so hopefully they pass CI too.

shaldengeki:
@dougthor42 Thanks for the review -- really excited to get this in!

I just noticed, as part of doing #3081, that I'm technically not following [the convention you linked](https://bazel.build/reference/be/protocol-buffer#py_proto_library). I'm generating `foo_proto_py_pb2` instead of `foo_py_pb2`. I just pushed a change that fixes this, both in the plugin and also in a bunch of examples where I noticed this as well. Hope that's OK?

dougthor42 (approved):

PR #3058, opened 2025-07-05
fix(pypi): pull fewer wheels with experimental_index_url

### LINKED ISSUE #2759: "Illegal ambiguous match" error with `experimental_index_url`
# 🐞 bug report

### Affected Rule

`pip.parse`

### Is this a regression?

not sure

### Description

When `pip.parse` with `experimental_index_url` uses per-platform requirements files and package versions don't match across files, it's possible (at least on OSX, for some packages) that using the package fails with "Illegal ambiguous match"

## 🔬 Minimal Reproduction

https://github.com/tbarrella/ambiguous

## 🔥 Exception or Error

<pre><code>
Illegal ambiguous match on configurable attribute "actual" in @@rules_python++pip+pypi//debugpy:pkg:
@@rules_python++pip+pypi//_config:is_cp313_py_none_any_osx_aarch64
@@rules_python++pip+pypi//_config:is_cp313_cp313_osx_universal2
Multiple matches are not allowed unless one is unambiguously more specialized or they resolve to the same value. See https://bazel.build/reference/be/functions#select.
</code></pre>

<pre><code>
Illegal ambiguous match on configurable attribute "actual" in @@rules_python++pip+pypi//tornado:pkg:
@@rules_python++pip+pypi//_config:is_cp313_sdist_osx_aarch64
@@rules_python++pip+pypi//_config:is_cp313_abi3_osx_universal2
Multiple matches are not allowed unless one is unambiguously more specialized or they resolve to the same value. See https://bazel.build/reference/be/functions#select.
</code></pre>

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
macOS, M2, 15.3.1
  </code>
</pre>

<pre>
  <code>
Bazelisk version: 1.25.0
Build label: 8.1.1
Build time: Tue Feb 25 18:53:44 2025 (1740509624)
Build timestamp: 1740509624
Build timestamp as int: 1740509624
  </code>
</pre>

**Rules_python version:**

<pre>
  <code>
1.3.0
  </code>
</pre>

**Anything else relevant?**

- The root cause may be the same as #2648
- Workaround is to manually make the versions and hashes match when possible

### LINKED ISSUE #2849: experimental_index_url slows down downloads by 25-50%
# 🐞 bug report

### Affected Rule

pip.parse

### Is this a regression?

No

### Description

Through various issues it has been recommended that I use `experimental_index_url` to get improved behavior for pip.parse. My discovery after attempting to use this in our project is that fetching all deps for the first time slowed down significantly. Here are some of the stats I pulled in our project:

without experimental_index_url:

```
bazel query 'deps(...)' > /dev/null  1.02s user 0.42s system 0% cpu 4:42.53 total
bazel query 'deps(...)' > /dev/null  1.05s user 0.42s system 0% cpu 5:31.24 total
```

with experimental index

```
bazel query 'deps(...)' > /dev/null  1.28s user 0.45s system 0% cpu 6:56.47 total
bazel query 'deps(...)' > /dev/null  1.16s user 0.51s system 0% cpu 7:15.13 total
bazel query 'deps(...)' > /dev/null  1.22s user 0.54s system 0% cpu 7:52.11 total
```

It _seems_ like this is consistent, I tested about 10 times before assuming this was the cause. I imagine this is heavily dependent on what other repo activity exists in the project. I tried messing with the value of `--http_max_parallel_downloads` without any luck.

To test this I did a `rm -rf ~/.cache/pip ~/.cache/bazel` between runs to make sure I was starting completely clean.

## 🔬 Minimal Reproduction

I tried reproducing this in a rules_python example but the difference was not as severe as in our project, which is what makes me think that it's heavily dependent on other http_archives etc that you have.

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
linux x86_64
  </code>
</pre>

<pre>
  <code>
https://github.com/bazelbuild/bazel/commit/2780393d35ad0607cf5e344ae082b00a5569a964
  </code>
</pre>

**Rules_python version:**

<pre>
  <code>
1.4.0-rc
  </code>
</pre>

**Anything else relevant?**

https://bazelbuild.slack.com/archives/CA306CEV6/p1745365100854859

aignas:
Before this we would pull all of the wheels that the user target configuration
would be compatible with and that meant that it was not customizable. This also
meant that there were a lot of footguns in the configuration where the select
statements were not really foolproof.

With this PR we select only those sources that need to be for the declared
configurations.

Freethreaded support should be done by defining extra freethreaded platforms
using the new builder API. It is done as a followup in #3063.

This is also changing the default platforms to be only the fully supported
platforms. This makes the testing easier and avoids us running into
compatibility issues during the rollout.

Work towards #2747
Fixes #2759
Fixes #2849

### DIFF
--- a/tests/pypi/parse_requirements/parse_requirements_tests.bzl
+++ b/tests/pypi/parse_requirements/parse_requirements_tests.bzl
@@ -15,9 +15,10 @@
 ""
 
 load("@rules_testing//lib:test_suite.bzl", "test_suite")
-load("//python/private/pypi:parse_requirements.bzl", "parse_requirements", "select_requirement")  # buildifier: disable=bzl-visibility
-load("//python/private/pypi:pep508_env.bzl", pep508_env = "env")  # buildifier: disable=bzl-visibility
+load("//python/private:repo_utils.bzl", "REPO_DEBUG_ENV_VAR", "REPO_VERBOSITY_ENV_VAR", "repo_utils")  # buildifier: disable=bzl-visibility
 load("//python/private/pypi:evaluate_markers.bzl", "evaluate_markers")  # buildifier: disable=bzl-visibility
+load("//python/private/pypi:parse_requirements.bzl", "select_requirement", _parse_requirements = "parse_requirements")  # buildifier: disable=bzl-visibility
+load("//python/private/pypi:pep508_env.bzl", pep508_env = "env")  # buildifier: disable=bzl-visibility
 
 def _mock_ctx():
     testdata = {
@@ -104,9 +105,22 @@ bar==0.0.1 --hash=sha256:deadb00f
 
 _tests = []
 
+def parse_requirements(debug = False, **kwargs):
+    return _parse_requirements(
+        ctx = _mock_ctx(),
+        logger = repo_utils.logger(struct(
+            os = struct(
+                environ = {
+                    REPO_DEBUG_ENV_VAR: "1",
+                    REPO_VERBOSITY_ENV_VAR: "TRACE" if debug else "INFO",
+                },
+            ),
+        ), "unit-test"),
+        **kwargs
+    )
+
 def _test_simple(env):
     got = parse_requirements(
-        ctx = _mock_ctx(),
         requirements_by_platform = {
             "requirements_lock": ["linux_x86_64", "windows_x86_64"],
         },
@@ -139,7 +153,6 @@ _tests.append(_test_simple)
 def _test_direct_urls_integration(env):
     """Check that we are using the filename from index_sources."""
     got = parse_requirements(
-        ctx = _mock_ctx(),
         requirements_by_platform = {
             "requirements_direct": ["linux_x86_64"],
             "requirements_direct_sdist": ["osx_x86_64"],
@@ -179,7 +192,6 @@ _tests.append(_test_direct_urls_integration)
 
 def _test_extra_pip_args(env):
     got = parse_requirements(
-        ctx = _mock_ctx(),
         requirements_by_platform = {
             "requirements_extra_args": ["linux_x86_64"],
         },
@@ -211,7 +223,6 @@ _tests.append(_test_extra_pip_args)
 
 def _test_dupe_requirements(env):
     got = parse_requirements(
-        ctx = _mock_ctx(),
         requirements_by_platform = {
             "requirements_lock_dupe": ["linux_x86_64"],
         },
@@ -240,7 +251,6 @@ _tests.append(_test_dupe_requirements)
 
 def _test_multi_os(env):
     got = parse_requirements(
-        ctx = _mock_ctx(),
         requirements_by_platform = {
             "requirements_linux": ["linux_x86_64"],
             "requirements_windows": ["windows_x86_64"],
@@ -304,7 +314,6 @@ _tests.append(_test_multi_os)
 
 def _test_multi_os_legacy(env):
     got = parse_requirements(
-        ctx = _mock_ctx(),
         requirements_by_platform = {
             "requirements_linux_download_only": ["cp39_linux_x86_64"],
             "requirements_osx_download_only": ["cp39_osx_aarch64"],
@@ -385,7 +394,6 @@ def _test_env_marker_resolution(env):
         return ret
 
     got = parse_requirements(
-        ctx = _mock_ctx(),
         requirements_by_platform = {
             "requirements_marker": ["cp311_linux_super_exotic", "cp311_windows_x86_64"],
         },
@@ -432,7 +440,6 @@ _tests.append(_test_env_marker_resolution)
 
 def _test_different_package_version(env):
     got = parse_requirements(
-        ctx = _mock_ctx(),
         requirements_by_platform = {
             "requirements_different_package_version": ["linux_x86_64"],
         },
@@ -471,7 +478,6 @@ _tests.append(_test_different_package_version)
 
 def _test_optional_hash(env):
     got = parse_requirements(
-        ctx = _mock_ctx(),
         requirements_by_platform = {
             "requirements_optional_hash": ["linux_x86_64"],
         },
@@ -510,7 +516,6 @@ _tests.append(_test_optional_hash)
 
 def _test_git_sources(env):
     got = parse_requirements(
-        ctx = _mock_ctx(),
         requirements_by_platform = {
             "requirements_git": ["linux_x86_64"],
         },
@@ -539,7 +544,6 @@ _tests.append(_test_git_sources)
 
 def _test_overlapping_shas_with_index_results(env):
     got = parse_requirements(
-        ctx = _mock_ctx(),
         requirements_by_platform = {
             "requirements_linux": ["cp39_linux_x86_64"],
             "requirements_osx": ["cp39_osx_x86_64"],
@@ -626,7 +630,6 @@ _tests.append(_test_overlapping_shas_with_index_results)
 
 def _test_get_index_urls_different_versions(env):
     got = parse_requirements(
-        ctx = _mock_ctx(),
         requirements_by_platform = {
             "requirements_multi_version": [
                 "cp39_linux_x86_64",
@@ -675,14 +678,14 @@ def _test_get_index_urls_different_versions(env):
         evaluate_markers = lambda _, requirements: evaluate_markers(
             requirements = requirements,
             platforms = {
-                "cp39_linux_x86_64": struct(
-                    env = {"python_full_version": "3.9.0"},
-                ),
                 "cp310_linux_x86_64": struct(
                     env = {"python_full_version": "3.10.0"},
                 ),
+                "cp39_linux_x86_64": struct(
+                    env = {"python_full_version": "3.9.0"},
+                ),
             },
-        )
+        ),
     )
 
     env.expect.that_collection(got).contains_exactly([


aignas:
Created this as WIP to signal that I am working on this and to invite feedback. The CI should be mostly happy but there are a few TODOs before we can fully merge and I want to test it a little bit more.

- [x] Freethreaded platform definition should be possible.
- [ ] Probably something else that I need to think of more carefully.

TODO cleanup after this PR is proven to not break things:
- [x] Remove complex `config_setting` generation where we are generating them from whl filenames - it is just too complex to properly maintain and it has not proven to be effective here. In https://github.com/bazel-contrib/rules_python/pull/3065

aignas:
The CI failure is a flake.

aignas:
This would be ready to be reviewed, but because it is stacked on two PRs, let's get those out of the way first.

MODULE.bazel
```diff
@@ -60,9 +60,82 @@ register_toolchains("@pythons_hub//:all")
...
+    env = {"platform_version": "0"},
+    os_name = "windows",
+    platform = "windows_x86_64",
+    platform_tags = ["win_amd64"],
+)
 pip.parse(
     # NOTE @aignas 2024-10-26: We have an integration test that depends on us
     # being able to build sdists for this hub, so explicitly set this to False.
+    #
+    # how do we test sdists? Maybe just worth adding a single sdist somewhere?
```
rickeylev, line 150: Yeah. Create an sdist_from_dir repo rule similar to whl_from_dir. Just something simple to tar.gz everything. Then feed that into the whl_library machinery. Restrict the test to mac/linux if windows is problematic.

examples/bzlmod/MODULE.bazel
```diff
@@ -158,6 +158,23 @@ pip.whl_mods(
...
+        "@platforms//os:windows",
+        "@platforms//cpu:aarch64",
+    ],
+    env = {
+        "platform_version": "0",
+    },
+    os_name = "windows",
+    platform = "windows_aarch64",
+    platform_tags = ["win_amd64"],
+    want_abis = [],  # default to all ABIs
```
rickeylev: I don't like the name "want_abis". "want" is a weak/ambiguous word that implies its expected/common it won't be fulfilled -- does it mean "requires" or "if available" ?

allowed_abis, requires_abi, restricted_to_abis -- better communicates it must be one of the values

preferred_abis -- better communicates none may be chosen.

python/private/pypi/extension.bzl
```diff
@@ -72,11 +72,25 @@ def _platforms(*, python_version, minor_mapping, config):
         version = python_version,
         minor_mapping = minor_mapping,
     )
-    abi = "cp3{}".format(python_version[2:])
 
     for platform, values in config.platforms.items():
+        implementation = values.env["implementation_name"][:2].lower()
```
rickeylev: I think this is the spec of how to transform implementation_name to the abi tag prefix: https://packaging.python.org/en/latest/specifications/platform-compatibility-tags/#python-tag

I was going to suggest `env["abi"]`, but it doesn't look like abi is part of the environment marker language.

python/private/pypi/extension.bzl
```diff
@@ -72,11 +72,25 @@ def _platforms(*, python_version, minor_mapping, config):
...
-        platforms[key] = env(key) | values.env
+
+        env_ = env(struct(
+            abi = abi,
+            os = values.os_name,
+            arch = values.arch_name,
+        )) | values.env
+        platforms[key] = struct(
+            env = env_,
+            want_abis = [
```
rickeylev: What's the difference between the abi computed above, and the want_abis here?

python/private/pypi/extension.bzl
```diff
@@ -863,6 +826,40 @@ This is only used if the {envvar}`RULES_PYTHON_ENABLE_PIPSTAR` is enabled.
 """,
     ),
     # The values for PEP508 env marker evaluation during the lock file parsing
+    "os_name": attr.string(
+        doc = """\
+The OS name to be used.
```
rickeylev, line 907: As a user, how do I know what value to use? e.g. Linux, linux, macos, MacOS, OSX, windows, Windows, Win, etc etc

I think this is the doc? https://packaging.python.org/en/latest/specifications/dependency-specifiers/#environment-markers

python/private/pypi/extension.bzl
```diff
@@ -863,6 +826,40 @@ This is only used if the {envvar}`RULES_PYTHON_ENABLE_PIPSTAR` is enabled.
...
+""",
+    ),
+    "platform_tags": attr.string_list(
+        doc = """\
+A list of `platform_tag` matchers so that we can select the best wheel based on the user
+preference. Per platform we will select a single wheel and the last match from this list will
+take precedence.
+
+The items in this list can contain a single `*` character that is equivalent to `.*` regex match.
+""",
```
rickeylev: I don't understand the purpose/need for platform_tags.

Are "platform tags" a field in wheel metadata or something?

The two things I'm not grokking are:

We have a big map of `platform_name -> <os, cpu, config settings, env markers, etc>` -- how is that not sufficient to determine if a platform and wheel are compatible?

What happens when a wheel matches two different pip.default() platform_tags ?

python/private/pypi/select_whl.bzl
```diff
@@ -0,0 +1,130 @@
+"Select a single wheel that fits the parameters of a target platform."
```
rickeylev: Didn't another PR (or comment in one?) say that trying to match based on filename was problematic and being discarded? This seems like a lot of logic for something that is problematic?

examples/bzlmod/MODULE.bazel
aignas: Ah yes, the `want_abis` is a bad name, forgot to change this. We could call it `abi_tags`, similar to how we have `platform_tags`. This should be the same naming convention.

python/private/pypi/extension.bzl
aignas: Ah yes, I did it correctly elsewhere. This is a good reminder to move the logic to here.

python/private/pypi/extension.bzl
aignas: The `os_name` is just tho `@platforms//os:<name>`. For env markers, I think we could do our best to add the documentation from our internal comments.

python/private/pypi/extension.bzl
aignas: Regarding matching 2 different tags, I think the way it is implemented, it will match the first one that it meets and derive the priority from that, so it is deterministic.

Regarding the `platform_name` -> `os, cpu, config settings, ...`, I want to remove all of that because it was not working reliably. The platform tag may have too many hidden things and it makes things hard to model.

The main thing is that the same platform may be compatible with different platform tags. E.g. somebody asked to match `manylinux` (glibc) only if `musllinux` is not available (muslc). Similarly, `universal2` vs `arch` specific macos wheels - which one should you use?

BTW, the `platform_tag` is the part of the wheel filename (the last segment). It is https://packaging.python.org/en/latest/specifications/platform-compatibility-tags/#platform-tag

python/private/pypi/select_whl.bzl
aignas: I might have not expressed myself clearly enough.

Previous model was: for each requirement file get all of the whls that are compatible with the target platforms (because we don't want to break bazel query). Then map them to config settings. Each config setting is using `flag_value` and uses the select specialization rule to ensure that the most specialized wheel will be selected. However, some target platforms for specific python versions may be using different wheels and therefore we need to have some sort of override method. The selection of the wheels hence need to happen twice - in the repository phase because of overrides and in the analysis phase because of the wheel specialization. 

This model moves the logic to a single function that does not use selects to trim down the list of wheels that we need to use and instead selects just a single one based on the given `abi_tags` and `platform_tags` values, which both define the priority of various ABIs, etc. The selects now do not perform this specialization selection but they only choose which wheel we should take. The unit testing for the platform/wheel matching is now concentrated in a single function and is parameterised.

I am myself convinced that this is probably a better approach because we are left shifting the decision on what gets chosen to the user and we can configure everything through `MODULE.bazel`.

examples/bzlmod/MODULE.bazel
aignas: I'll go for `whl_abi_tags` and `whl_platform_tags`.

keith:
I hit this problem with this change (I don't support windows and use `requirements_by_platform` to exclude it too):

```
% bazel query 'deps(...)'
INFO: Repository rules_python++python+python_3_13_aarch64-pc-windows-msvc instantiated at:
  <builtin>: in <toplevel>
Repository rule python_repository defined at:
  /home/ubuntu/.cache/bazel/_bazel_ubuntu/480ee760ab026be301955dc1adc19e40/external/rules_python+/python/private/python_repository.bzl:271:36: in <toplevel>
ERROR: /home/ubuntu/.cache/bazel/_bazel_ubuntu/480ee760ab026be301955dc1adc19e40/external/rules_python+/python/private/python_repository.bzl:76:34: An error occurred during the fetch of repository 'rules_python++python+python_3_13_aarch64-pc-windows-msvc':
        File "/home/ubuntu/.cache/bazel/_bazel_ubuntu/480ee760ab026be301955dc1adc19e40/external/rules_python+/python/private/python_repository.bzl", line 76, column 34, in _python_repository_impl
                rctx.download_and_extract(
Error in download_and_extract: java.io.IOException: Error extracting /home/ubuntu/.cache/bazel/_bazel_ubuntu/480ee760ab026be301955dc1adc19e40/external/rules_python++python+python_3_13_aarch64-pc-windows-msvc/temp6334258738180209862/cpython-3.13.5+20250702-aarch64-pc-windows-msvc-install_only.tar.gz to /home/ubuntu/.cache/bazel/_bazel_ubuntu/480ee760ab026be301955dc1adc19e40/external/rules_python++python+python_3_13_aarch64-pc-windows-msvc/temp6334258738180209862: Prefix "python/install" was given, but not found in the archive. Here are possible prefixes for this archive: "python".
ERROR: Evaluation of query "deps(...)" failed: preloading transitive closure failed: no such package '@@rules_python++python+python_3_13_aarch64-pc-windows-msvc//': java.io.IOException: Error extracting /home/ubuntu/.cache/bazel/_bazel_ubuntu/480ee760ab026be301955dc1adc19e40/external/rules_python++python+python_3_13_aarch64-pc-windows-msvc/temp6334258738180209862/cpython-3.13.5+20250702-aarch64-pc-windows-msvc-install_only.tar.gz to /home/ubuntu/.cache/bazel/_bazel_ubuntu/480ee760ab026be301955dc1adc19e40/external/rules_python++python+python_3_13_aarch64-pc-windows-msvc/temp6334258738180209862: Prefix "python/install" was given, but not found in the archive. Here are possible prefixes for this archive: "python".
Loading: 2635 packages loaded
```

aignas:
Thank you, I'll take a look later.

Marking this as a draft for now.

aignas:
> I hit this problem with this change (I don't support windows and use `requirements_by_platform` to exclude it too):
> 
> ```
> % bazel query 'deps(...)'
> INFO: Repository rules_python++python+python_3_13_aarch64-pc-windows-msvc instantiated at:
>   <builtin>: in <toplevel>
> Repository rule python_repository defined at:
>   /home/ubuntu/.cache/bazel/_bazel_ubuntu/480ee760ab026be301955dc1adc19e40/external/rules_python+/python/private/python_repository.bzl:271:36: in <toplevel>
> ERROR: /home/ubuntu/.cache/bazel/_bazel_ubuntu/480ee760ab026be301955dc1adc19e40/external/rules_python+/python/private/python_repository.bzl:76:34: An error occurred during the fetch of repository 'rules_python++python+python_3_13_aarch64-pc-windows-msvc':
>    Traceback (most recent call last):
>         File "/home/ubuntu/.cache/bazel/_bazel_ubuntu/480ee760ab026be301955dc1adc19e40/external/rules_python+/python/private/python_repository.bzl", line 76, column 34, in _python_repository_impl
>                 rctx.download_and_extract(
> Error in download_and_extract: java.io.IOException: Error extracting /home/ubuntu/.cache/bazel/_bazel_ubuntu/480ee760ab026be301955dc1adc19e40/external/rules_python++python+python_3_13_aarch64-pc-windows-msvc/temp6334258738180209862/cpython-3.13.5+20250702-aarch64-pc-windows-msvc-install_only.tar.gz to /home/ubuntu/.cache/bazel/_bazel_ubuntu/480ee760ab026be301955dc1adc19e40/external/rules_python++python+python_3_13_aarch64-pc-windows-msvc/temp6334258738180209862: Prefix "python/install" was given, but not found in the archive. Here are possible prefixes for this archive: "python".
> ERROR: Evaluation of query "deps(...)" failed: preloading transitive closure failed: no such package '@@rules_python++python+python_3_13_aarch64-pc-windows-msvc//': java.io.IOException: Error extracting /home/ubuntu/.cache/bazel/_bazel_ubuntu/480ee760ab026be301955dc1adc19e40/external/rules_python++python+python_3_13_aarch64-pc-windows-msvc/temp6334258738180209862/cpython-3.13.5+20250702-aarch64-pc-windows-msvc-install_only.tar.gz to /home/ubuntu/.cache/bazel/_bazel_ubuntu/480ee760ab026be301955dc1adc19e40/external/rules_python++python+python_3_13_aarch64-pc-windows-msvc/temp6334258738180209862: Prefix "python/install" was given, but not found in the archive. Here are possible prefixes for this archive: "python".
> Loading: 2635 packages loaded
> ```

@keith, #3070 should fix your issue, I think. The toolchain pulling in `query` seems to be unrelated to `pip` integration, so if it works after #3070 and #3058 is merged, I think it is safe to say that the issue is resolved.

MODULE.bazel
aignas: To be addressed as part of #3071

python/private/pypi/select_whl.bzl
```diff
@@ -0,0 +1,120 @@
...
+
+        for p in tag.split("."):
+            if not sep and p == head:
+                return priority
+            elif sep and p.startswith(head) and p.endswith(tail):
+                return priority
+
+    return None
+
+def select_whl(*, whls, python_version, platforms, whl_abi_tags, implementation_name = "cpython", limit = 1, logger = None):
```
hartikainen: I tested this from d36610f and see `select_whl` failing with the weirdly named [`glfw` wheels](https://pypi.org/project/glfw/#files) like 
`glfw-2.9.0-py2.py27.py3.py30.py31.py32.py33.py34.py35.py36.py37.py38.p39.p310.p311.p312.p313-none-win_amd64.whl`:

<details>

```console
(15:35:30) ERROR: /private/var/tmp/_bazel_user/1e3721fe4e6a160dd5c2ac3feb003726/external/rules_python+/python/private/version.bzl:577:18: Traceback (most recent call last):
        File "/private/var/tmp/_bazel_user/1e3721fe4e6a160dd5c2ac3feb003726/external/rules_python+/python/private/pypi/extension.bzl", line 794, column 25, in _pip_impl
                mods = parse_modules(module_ctx, enable_pipstar = rp_config.enable_pipstar)
        File "/private/var/tmp/_bazel_user/1e3721fe4e6a160dd5c2ac3feb003726/external/rules_python+/python/private/pypi/extension.bzl", line 655, column 36, in parse_modules
                out = _create_whl_repos(
        File "/private/var/tmp/_bazel_user/1e3721fe4e6a160dd5c2ac3feb003726/external/rules_python+/python/private/pypi/extension.bzl", line 242, column 50, in _create_whl_repos
        File "/private/var/tmp/_bazel_user/1e3721fe4e6a160dd5c2ac3feb003726/external/rules_python+/python/private/pypi/parse_requirements.bzl", line 197, column 33, in parse_requirements
                srcs = _package_srcs(
        File "/private/var/tmp/_bazel_user/1e3721fe4e6a160dd5c2ac3feb003726/external/rules_python+/python/private/pypi/parse_requirements.bzl", line 242, column 30, in _package_srcs
                dist = _add_dists(
        File "/private/var/tmp/_bazel_user/1e3721fe4e6a160dd5c2ac3feb003726/external/rules_python+/python/private/pypi/parse_requirements.bzl", line 418, column 22, in _add_dists
                return select_whl(
        File "/private/var/tmp/_bazel_user/1e3721fe4e6a160dd5c2ac3feb003726/external/rules_python+/python/private/pypi/select_whl.bzl", line 72, column 43, in select_whl
                min_whl_py_version = version.parse(min_version, strict = True)
        File "/private/var/tmp/_bazel_user/1e3721fe4e6a160dd5c2ac3feb003726/external/rules_python+/python/private/version.bzl", line 614, column 19, in parse
                parts = _parse(version_str, strict = strict, _fail = _fail)
        File "/private/var/tmp/_bazel_user/1e3721fe4e6a160dd5c2ac3feb003726/external/rules_python+/python/private/version.bzl", line 577, column 18, in _parse
                _fail(
Error in fail: Failed to parse PEP 440 version identifier '2..py27.py3.py30.py31.py32.py33.py34.py35.py36.py37.py38.p39.p310.p311.p312.p313'. Parse error at '..py27.py3.py30.py31.py32.py33.py34.py35.py36.py37.py38.p39.p310.p311.p312.p313'
(15:35:30) WARNING: errors encountered while analyzing target '//:main_image.load', it will not be built.
```

</details>

aignas: OK, this makes sense, instead of `py2.py3` handling as a special case, I'll change this to actually use the version numbers in a different way.

hartikainen: I just tested the latest commit (82e153251aac435b35ddb0f013dff6418036ea1f) and don't see this issue anymore. Thanks for the quick fix!

aignas:
Just a quick note, this can be reviewed as is, but I want to change up the `whl_platform_tags` behaviour a little in the upcoming week to fully cover all of the bases. The usage would be:
* `manylinux_*_x86_64` means pick the latest `manylinux_x_y_x86_64` package that exists.
* `manylinux_2_31_x86_64` means pick the latest `manylinux_x_y_x86_64` package but no later than `manylinux_2_31_x86_64`.

Right now we just pick the last one in the list, which is a bug. If you have ideas how we can express this idea better, let me know.

aignas:
@keith, @hartikainen, could you please retest this PR if it works with your setup.

aignas:
I am going to split this into smaller PRs and keep this as the top of stack. Marking as draft to signal that this should not be merged.

aignas:
OK, the splitting is done.

PR #3059, opened 2025-07-05
fix(pypi): only generate namespace package shims if implicit namespaces are disabled

rickeylev:
The refactoring to move the pkgutil shim generation to build phase inverted
the logic for when it should be activated. When
`enable_implicit_namespace_pkgs=True`, it means to not generate the pkgutil shims
("respect the Python definition of the namespace package").

To fix, just invert the logic that activates it.

A test will be added in a subsequent PR because the necessary helper isn't
in the 1.5 branch.

Fixes https://github.com/bazel-contrib/rules_python/issues/3038

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -86,6 +86,18 @@ END_UNRELEASED_TEMPLATE
 ### Removed
 * Nothing removed.
 
+{#1-5-1}
+## [1.5.1] - 2025-06-XX
+
+[1.5.1]: https://github.com/bazel-contrib/rules_python/releases/tag/1.5.1
+
+{#v1-5-1-fixed}
+### Fixed
+
+* (pypi) Namespace packages work by default (pkgutil shims are generated
+  by default again)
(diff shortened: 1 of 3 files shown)


rickeylev:
note to self:

* looks like the tests are using some sort of mock that isn't glob-like enough. 
* Argh, the test calls a rule, so can't actually call create_inits(). I think the test needs to be rewritten as an analysis test or something.

aignas (approved):

aignas:
In order to unblock this I modified the unit test to set the flag to disable the consumption of the glob.

CHANGELOG.md
```diff
@@ -86,6 +86,18 @@ END_UNRELEASED_TEMPLATE
 ### Removed
 * Nothing removed.
 
+{#1-5-1}
+## [1.5.1] - 2025-06-XX
```
```suggestion
## [1.5.1] - 2025-07-06
```

aignas:
Thanks for catching this and digging until the end!

PR #3062, opened 2025-07-06
feat: Add windows arm64 python toolchains

AustinSchuh:
The changelog for astral-sh/python-build-standalone release 20250630 says:

  * Add ARM64 Windows builds for Python 3.11+

Lets use them!  This is helpful when using rules_python on arm64 windows

Work towards #2276

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -101,6 +101,18 @@ END_UNRELEASED_TEMPLATE
 ### Removed
 * Nothing removed.
 
+{#1-5-1}
+## [1.5.1] - 2025-07-06
+
+[1.5.1]: https://github.com/bazel-contrib/rules_python/releases/tag/1.5.1
+
+{#v1-5-1-fixed}
+### Fixed
+
+* (pypi) Namespace packages work by default (pkgutil shims are generated
+  by default again)
+  ([#3038](https://github.com/bazel-contrib/rules_python/issues/3038)).
+
 {#1-5-0}
 ## [1.5.0] - 2025-06-11
 
--- a/python/private/pypi/whl_library_targets.bzl
+++ b/python/private/pypi/whl_library_targets.bzl
@@ -331,7 +331,7 @@ def whl_library_targets(
             allow_empty = True,
         )
 
-        if enable_implicit_namespace_pkgs:
+        if not enable_implicit_namespace_pkgs:
             srcs = srcs + getattr(native, "select", select)({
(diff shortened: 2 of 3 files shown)


aignas (changes requested):
Thanks for the PR, almost LGTM.

Please update "3.9" and "3.10" to use the same latest release. Usually they include security and other fixes, so a subset of versions from the release is not what we normally do.

And please change the existing changelog message about the toolchain bump. The users won't see 3.14.0b2 and the previous toolchain bump, so best to have one entry that they can read.

AustinSchuh:
Done!  That's an easy and reasonable request.

I updated `//python/private:print_toolchains_checksums` to print the versions + checksums out in a format which can be copied into `versions.bzl` much easier.  Happy to revert that if you'd rather not merge it.

aignas:
Yes please, reverting the formatting would be easier because I have a PR for that ready.

CHANGELOG.md
```diff
@@ -91,6 +91,10 @@ END_UNRELEASED_TEMPLATE
 * (gazelle) New directive `gazelle:python_generate_pyi_deps`; when `true`,
   dependencies added to satisfy type-only imports (`if TYPE_CHECKING`) and type
   stub packages are added to `pyi_deps` instead of `deps`.
+* (toolchain) Add toolchains for aarch64 windows for
+    * 3.11.13
+    * 3.12.11
```
aignas, line 96:
```suggestion
    * 3.12.11
    * 3.13.5
```

aignas (approved):

python/versions.bzl
```diff
@@ -798,29 +802,33 @@ TOOL_VERSIONS = {
             "s390x-unknown-linux-gnu-freethreaded": "python/install",
             "x86_64-apple-darwin-freethreaded": "python/install",
             "x86_64-pc-windows-msvc-freethreaded": "python/install",
+            "aarch64-pc-windows-msvc": "python/install",
```
aignas, line 805: @AustinSchuh, Is this correct? I think this may need to be

```suggestion
            "aarch64-pc-windows-msvc": "python",
```

aignas: Created #3070

AustinSchuh: That's embarrassing, thank you for catching that and fixing it!  I downloaded the archives and confirmed you are correct.  I also checked the other ones I added.

aignas: No worries, I think the lack of CI is to blame here.

PR #3063, opened 2025-07-06
feat: freethreaded support for the builder API

aignas:
Stacked on #3058

This is a continuation of #3058 where we define freethreaded
platforms. They need to be used only for particular python versions
so I included an extra marker configuration attribute where we
are using pipstar marker evaluation before using the platform.

I think this in general will be a useful tool to configure only
particular platforms for particular python versions

Work towards #2548, since this shows how we can define custom platforms
Work towards #2747

TODO:
- [ ] Fix the remaining expectations in the unit tests. Maybe make the
  tests less brittle and define platforms for unit testing.


python/private/pypi/extension.bzl
```diff
@@ -83,8 +84,13 @@ def _platforms(*, python_version, minor_mapping, config):
             os = values.os_name,
             arch = values.arch_name,
         )) | values.env
+
+        if values.marker and not evaluate(values.marker, env = env_):
+            continue
+
         platforms[key] = struct(
             env = env_,
+            tripple = "{}_{}_{}".format(abi, values.os_name, values.arch_name),
```
rickeylev: typo: triple

MODULE.bazel
```diff
@@ -70,14 +70,22 @@ pip = use_extension("//python/extensions:pip.bzl", "pip")
         config_settings = [
             "@platforms//cpu:{}".format(cpu),
             "@platforms//os:linux",
+            "//python/config_settings:_is_py_freethreaded_{}".format(
+                "yes" if freethreaded else "no",
+            ),
         ],
         env = {"platform_version": "0"},
+        marker = "python_version ~= \"3.13\"" if freethreaded else "",
```
rickeylev: Pretty sure this should be >= 3.13

MODULE.bazel
```diff
@@ -97,16 +109,24 @@ pip = use_extension("//python/extensions:pip.bzl", "pip")
...
         os_name = "osx",
-        platform = "osx_{}".format(cpu),
+        platform = "osx_{}{}".format(cpu, freethreaded),
         platform_tags = [
             "macosx_*_{}".format(suffix)
             for suffix in platform_tag_cpus
         ],
+        want_abis = [
+            "cp{0}{1}t",
```
rickeylev: nit: `{0}{1}` is fine for experimental api, but we should use named params for the final API. Please add a todo, at the least.

python/private/pypi/extension.bzl
```diff
@@ -219,7 +227,14 @@ def _create_whl_repos(
         # spin up a Python interpreter.
         evaluate_markers = lambda module_ctx, requirements: evaluate_markers_py(
             module_ctx,
-            requirements = requirements,
+            requirements = {
+                k: {
+                    # TODO @aignas 2025-07-06: should we leave this as is?
```
rickeylev: leave what as-is?

python/private/pypi/extension.bzl
```diff
@@ -320,6 +331,16 @@ def _create_whl_repos(
                 ))
 
             whl_libraries[repo_name] = repo.args
+            if "experimental_target_platforms" in repo.args:
+                whl_libraries[repo_name] |= {
+                    "experimental_target_platforms": sorted({
+                        # TODO @aignas 2025-07-07: this should be solved in a better way
+                        platforms[candidate].tripple.partition("_")[-1]: None
```
rickeylev: Yeah, isn't this problematic if platform names are user-chosen?

python/private/pypi/extension.bzl
```diff
@@ -406,21 +427,25 @@ def _configure(config, *, platform, os_name, arch_name, config_settings, env = {
             # the lowest priority one needs to be the first one
             platform_tags = ["any"] + platform_tags
 
+        want_abis = want_abis or [
+            "cp{0}{1}",
+            "abi3",
```
rickeylev: Why is abi3 here, but not in the MODULE config?

aignas: Done, incorporated in #3058.

python/private/pypi/extension.bzl
aignas: This is constructed in a way that does not include the user-chosen platform names.

MODULE.bazel
aignas: `~= "3.13"` means `>=3.13 and ==3.*`, so this is technically what we want. There probably won't be Python 4 and the freethreaded by then might not be there. That said, maybe `>=` is more readable.

python/private/pypi/extension.bzl
```diff
@@ -899,6 +922,12 @@ Supported keys:
 ::::{note}
 This is only used if the {envvar}`RULES_PYTHON_ENABLE_PIPSTAR` is enabled.
 ::::
+""",
+    ),
+    "marker": attr.string(
+        doc = """\
+A marker which will be evaluated to disable the target platform for certain python versions. This
```
rickeylev: It wasn't clear to me what "marker" was. It was only after reading the PR description that it was clear it was an environment marker.

Can this be any expression? Is it evaluated as part of the bzlmod phase or analysis (pipstar) phase? Why would I set this instead of env or vice versa?

The docs say "to disable", so if this expression is *true*, then the platform will not match? That logic seems inverted from regular marker expressions. e.g normally `foo; python_version=3.14` means "if the python version is 3.14, then include/use/accept foo". However, here it means the opposite: if the expression is true, then _don't_ include/use/accept foo

aignas: Good points:
* Logic inversion is not there - if it is true, then the platform is active.
* We can call it `marker_expression`, but ideally we could also call the field `enabled_if_true` and then the value is the marker expression.

rickeylev: I think "marker" or "marker_expr" are fine names.

As long as the docs clarify it's an expression and it's evaluation context, then lgtm

aignas: OK, done.

rickeylev (approved):

PR #2962, opened 2025-06-06
deps(gazelle): Bump rules_go from 0.41.0 to 0.55.0

dougthor42:
Bump `rules_go` from 0.41.0 to 0.55.0

Fixes #2956.

+ Run `go get github.com/bazelbuild/rules_go@v0.55.0` for version bump.
+ Run `go mod tidy` to clean up stale deps
+ Run `bazel run //:gazelle_update_repos`
+ Figure out http_archive hashes and apply them for WORKSPACE users.
+ Also bump bazel-gazelle to address a warning:
    + `go get github.com/bazelbuild/bazel-gazelle@v0.40.0`
    + > Mismatch between versions requested for module
      > github.com/bazelbuild/bazel-gazelle
      > Bazel dependency version requested in MODULE.bazel: 0.40.0
      > Go module version requested in go.mod: 0.31.1
      > Please resolve this mismatch to prevent discrepancies between
      > native Go and Bazel builds
+ Update CHANGELOG.md accordingly

PR #3069, opened 2025-07-08
chore: Switch back to smacker/go-tree-sitter

dougthor42:
Finally remove the dougthor42/go-tree-sitter fork, fixing #2630.

Admittedly we could have done this sooner had I figured things
out sooner... but c'est la vie.

Instead of using the BUILD.bazel files in dougthor42/go-tree-sitter,
we basically vendor it via http_archive.

This is different than using patches because non-root Bazel modules
can still make use of the BUILD.bazel files we make.

Background:

The reason we migrated to dougthor42/go-tree-sitter in the first
place was to support python 3.12 grammar. smacker/go-tree-sitter
supported for python 3.12, but made a change to their file structure
that Gazelle was unable to handle. Specifically, the
python/binding.go file indirectly requires a c header file found
in a parent directory, and Gazelle doesn't know how to handle that
for `go_repository` (WORKSPACE) and `go_deps.from_file` (bzlmod).

So the dougthor42/go-tree-sitter created our own BUILD.bazel files
that included the required filegroups and whatnot, thus negating
the need for Gazelle to generate BUILD.bazel files.

Future Work:

This still doesn't resolve the issues with bumping rules_go
and go seen in #2962, but it does simplify that investigation
a bit as it's just one fewer thing to account for. It also
doesn't address the desire to migrate to the official
tree-sitter/go-tree-sitter repo, but @jbedard found some
pref issues with that anyway
(https://github.com/tree-sitter/go-tree-sitter/issues/32).


dougthor42:
Holy shit things passed, yay!

aignas (approved):

gazelle/MODULE.bazel
```diff
@@ -21,14 +21,23 @@ use_repo(
...
     "com_github_stretchr_testify",
     "in_gopkg_yaml_v2",
     "org_golang_x_sync",
 )
 
+http_archive = use_repo_rule("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+http_archive(
+    name = "com_github_smacker_go_tree_sitter",
+    build_file = "@rules_python_gazelle_plugin//:internal/smacker_BUILD.bazel",
```
aignas: nit: I would have expected the following to work too:

```suggestion
    build_file = "@//:internal/smacker_BUILD.bazel",
```

gazelle/deps.bzl
```diff
@@ -175,18 +174,18 @@ def go_deps():
...
         version = "v0.0.0-20190812154241-14fe0d1b01d4",
     )
-    go_repository(
-        name = "com_github_dougthor42_go_tree_sitter",
-        importpath = "github.com/dougthor42/go-tree-sitter",
-        sum = "h1:b9s96BulIARx0konX36sJ5oZhWvAvjQBBntxp1eUukQ=",
-        version = "v0.0.0-20241210060307-2737e1d0de6b",
+    http_archive(
+        name = "com_github_smacker_go_tree_sitter",
+        build_file = "@rules_python_gazelle_plugin//:internal/smacker_BUILD.bazel",
```
aignas: Likewise:
```suggestion
        build_file = Label("//:internal/smacker_BUILD.bazel"),
```

Here is different because users invoke this macro in their workspaces so `//` points elsewhere.

gazelle/MODULE.bazel
dougthor42: All local tests using `@//` seemed to run just fine.

However, when using `@//` in a bzlmod repository that consumes `rules_python` and `rules_python_gazelle_plugin`, I get an error `no repository visible as '@'`:

```console
$ pwd
/usr/local/google/home/dthor/dev/pyle3
$ bazel run //:gazelle
INFO: Invocation ID: f6daed42-b174-43cb-b295-2c05b614b5dd
ERROR: Analysis of target '//:gazelle' failed; build aborted: error creating repo com_github_smacker_go_tree_sitter requested at @@rules_python_gazelle_plugin~//:MODULE.bazel:33:13: no repository visible as '@' to the module 'rules_python_gazelle_plugin@_', but referenced by label '@//:internal/smacker_BUILD.bazel' in attribute 'build_file' of http_archive 'com_github_smacker_go_tree_sitter'.
INFO: Elapsed time: 0.814s, Critical Path: 0.20s
INFO: 1 process: 1 internal.
```

gazelle/deps.bzl
dougthor42: I don't have a _robust_ way to manually test this (all our projects are on bzlmod) but I did set up a dummy WORKSPACE project to with a `local_repository` pointing to this PR and it seemed to work.

Updated.

gazelle/MODULE.bazel
aignas: hmmm, maybe `//:internal/smacker_BUILD.bazel` should be the right thing?

Anyway, this is a minor point here, so feel free to merge. :)

dougthor42: That seems to work; updated!

PR #3072, opened 2025-07-09
feat: Support main_module with bootstrap_impl=system_python

brandonchinn178:
`main_module` was added in #2671, but support was only added for `bootstrap_impl=script`. Despite only supporting `bootstrap_impl=script`, both the CHANGELOG and the documentation didn't mention that limitation, making it seem like `main_module` doesn't work out-of-the-box (given that `bootstrap_impl=system_python` is the default). See https://github.com/bazel-contrib/rules_python/issues/2765#issuecomment-3046428511.

This PR adds support for `main_module` with `bootstrap_impl=system_python`. This is categorized as `feat` because it's new functionality, but it can also be conceptualized as a `fix`, since it's bringing the feature in line with docs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -104,7 +104,7 @@ END_UNRELEASED_TEMPLATE
     * 3.12.11
     * 3.13.5
     * 3.14.0b3
-* (rules) `main_module` now works without needing to set `bootstrap_impl=script`
+* (rules) {obj}`main_module` now works without needing to set {obj}`--bootstrap_impl=script`
 
 {#v0-0-0-removed}
 ### Removed
--- a/tests/bootstrap_impls/BUILD.bazel
+++ b/tests/bootstrap_impls/BUILD.bazel
@@ -124,24 +124,25 @@ py_reconfig_test(
     main = "sys_path_order_test.py",
 )
 
-py_reconfig_test(
-    name = "main_module_test",
+py_library(
+    name = "main_module_lib",
     srcs = ["main_module.py"],
+    imports = ["."],
+)
+
+py_reconfig_test(
+    name = "main_module_script_test",
     bootstrap_impl = "script",
     main_module = "tests.bootstrap_impls.main_module",
     target_compatible_with = SUPPORTS_BOOTSTRAP_SCRIPT,
-)
-
-py_library(
-    name = "main_module_system_python_lib",
-    srcs = ["main_module.py"],
+    deps = [":main_module_lib"],
 )
 
 py_reconfig_test(
     name = "main_module_system_python_test",
     bootstrap_impl = "system_python",
     main_module = "tests.bootstrap_impls.main_module",
-    deps = [":main_module_system_python_lib"],
+    deps = [":main_module_lib"],
 )
 
 sh_py_run_test(


tests/bootstrap_impls/BUILD.bazel
```diff
@@ -127,11 +128,22 @@ py_reconfig_test(
     name = "main_module_test",
     srcs = ["main_module.py"],
     bootstrap_impl = "script",
-    imports = ["."],
```
rickeylev, line 130: This needs to be kept because it ensures the correct entry is added to sys.path. Without it, it can inadvertently rely on the implicit behavior from the flag that adds repos to sys.path and the PYTHONSAFEPATH setting

tests/bootstrap_impls/BUILD.bazel
```diff
@@ -127,11 +128,22 @@ py_reconfig_test(
...
     target_compatible_with = SUPPORTS_BOOTSTRAP_SCRIPT,
 )
 
+py_library(
+    name = "main_module_system_python_lib",
+    srcs = ["main_module.py"],
+)
+
+py_reconfig_test(
+    name = "main_module_system_python_test",
```
rickeylev, line 142: Please refactor the tests to:

* Move main_module into py_library
* rename the lib to main_module_lib
* Add imports=. to the lib
* rename the tests main_module_bootstrap_{script,system_python}_test

This keeps the two tests looking the same.

python/private/main_module_entrypoint_template.py
```diff
@@ -0,0 +1,9 @@
+"""
+A shim to get `main_module` working with `bootstrap_impl=system_python`.
```
rickeylev, line 2: Instead of creating a shim like this, modify the bootstrap template. All it should really need is to populate a %main_module% variable and pass `-m` when invoking python. That should simplify things quite a bit.

CHANGELOG.md
```diff
@@ -104,6 +104,7 @@ END_UNRELEASED_TEMPLATE
     * 3.12.11
     * 3.13.5
     * 3.14.0b3
+* (rules) `main_module` now works without needing to set `bootstrap_impl=script`
```
```suggestion
* (rules) {obj}`main_module` now works without needing to set {obj}`--bootstrap_impl=script`
```

python/private/main_module_entrypoint_template.py
brandonchinn178: hm I could try it again, but it was difficult because then I'd have to update all the locations where `%main%` is used, since there isn't a main file anymore. It seemed easier to say "we're using the main file logic everywhere still, it just points to a shim"

tests/bootstrap_impls/BUILD.bazel
brandonchinn178: I can re-add it, but I don't understand the reasoning. IIUC, `imports` says "add this directory to PYTHONPATH", but we're specifying `main_module` with the full absolute path, so `imports = ["."]` isn't doing anything. Would `imports = []` have the behavior you want?

python/private/main_module_entrypoint_template.py
brandonchinn178: yeah, it seems like the notion of a main file is hardcoded in a lot of places in `python_bootstrap_template.txt`. It's also not immediately clear how the translation to `-m <main_module>` would work for the coverage path. The shim seems like the easiest solution, especially given this is a deprecated codepath and it should have a negligible performance impact

brandonchinn178:
@rickeylev Any updates?

PR #3074, opened 2025-07-10
chore: update python toolchains

aignas:
- use the SHA256SUMS file instead of individual sha256sum files. This
  improves the speed of the tooling and also the old files just
  disappeared for the latest toolchain release.
- update to the latest release.


rickeylev:
CVE-2025-47273 is a vuln in setuptools, how does the interpreter address it?

aignas:
This is because `setuptools` get bundled with the interpreter.

scasagrande:
can we get this cherry picked onto 1.5 to resolve the CVE?

rickeylev:
That's a good idea. 
I think the python.override apis can also be used to switch up which tar.gz is used

scasagrande:
Yeah I finally tracked down where my cve alert was coming from and now I'm using python.override to pick up these new versions.

But releasing a patch version of rules_python will make the fix more likely to be picked up by everyone

PR #3077, opened 2025-07-10
fix(gazelle) Update gazelle to properly process multi-line python imports

yushan26:
A python import may be imported as:
```
from foo.bar.application.\
    pipeline.model import (
    Baz
)
```
However, gazelle fails to resolve this import with the error:
`line 30: "foo.bar.application.pipeline.model\\\n     pipeline.mode.Baz" is an invalid dependency:`

Clean up the imports such that whitespace and \n are removed from the import path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,6 +69,7 @@ END_UNRELEASED_TEMPLATE
 * (toolchain) Python 3.13 now references 3.13.5
 * (gazelle) Switched back to smacker/go-tree-sitter, fixing
   [#2630](https://github.com/bazel-contrib/rules_python/issues/2630)
+* (gazelle) Cleanup multi-line python imports modules
 
 {#v0-0-0-fixed}
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,7 +69,6 @@ END_UNRELEASED_TEMPLATE
 * (toolchain) Python 3.13 now references 3.13.5
 * (gazelle) Switched back to smacker/go-tree-sitter, fixing
   [#2630](https://github.com/bazel-contrib/rules_python/issues/2630)
-* (gazelle) Cleanup multi-line python imports modules
 
 {#v0-0-0-fixed}
 ### Fixed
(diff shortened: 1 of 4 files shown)


gazelle/python/file_parser_test.go
```diff
@@ -291,3 +291,32 @@ def example_function():
...
 	}
 }
+
+func TestParseImportStatements_MultilineWithBackslashAndWhitespace(t *testing.T) {
+	p := NewFileParser()
+	code := []byte(`from foo.bar.\
+    baz import (
+    Something,
+    AnotherThing
+)
```
dougthor42, line 303: Let's test these cases:

1. `from $THING_WITH_NEWLINE import $OTHER` (which you already have)
2. `import $THING_WITH_NEWLINE`
3. The above two cases with:
    1. the escape and newline _before_ the dot.
    2. Windows line endings.

gazelle/python/file_parser.go
```diff
@@ -144,6 +144,14 @@ func parseImportStatement(node *sitter.Node, code []byte) (Module, bool) {
 	return Module{}, false
 }
 
+// cleanImportString removes backslashes and all whitespace from the string.
+func cleanImportString(s string) string {
+	s = strings.ReplaceAll(s, "\\", "")
+	s = strings.ReplaceAll(s, " ", "")
+	s = strings.ReplaceAll(s, "\n", "")
```
dougthor42, line 152: `\r\n` too?

dougthor42 (commented):
Thanks! I'm somewhat surprised that the underlying tree-sitter parser doesn't handle this.

I know you already have the go tests, but lets also add a test case to `gazelle/python/testdata/from_imports/imports_nested_var/__init__.py`:

```diff
 # baz is a variable in foo/bar/baz.py
 from foo.bar.baz import baz
+
+from foo\
+    .bar.\
+        baz import (
+                baz
+                )
```

Given that this is the same import as the previous line, the `BUILD.out` shouldn't change.

CHANGELOG.md
```diff
@@ -69,6 +69,7 @@ END_UNRELEASED_TEMPLATE
 * (toolchain) Python 3.13 now references 3.13.5
 * (gazelle) Switched back to smacker/go-tree-sitter, fixing
   [#2630](https://github.com/bazel-contrib/rules_python/issues/2630)
+* (gazelle) Cleanup multi-line python imports modules
```
dougthor42: Nit1: Move to "fixed" section

Nit2: The current wording makes it sound like we just did some internal refactoring. Consider something like `Multi-line python imports (eg with escaped newlines) are now correctly processed by Gazelle.`.

yushan26:
@dougthor42 Thanks for the review! Can you take another look?

dougthor42 (approved):

PR #3083, opened 2025-07-13
ci: use Ubuntu 22.04

aignas:
I saw most of the PRs fail on RBE with errors and maybe this will fix it.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -91,20 +91,20 @@ tasks:
     <<: *common_workspace_flags_min_bazel
     <<: *minimum_supported_version
     name: "Gazelle: workspace, minimum supported Bazel version"
-    platform: ubuntu2004
+    platform: ubuntu2204
     build_targets: ["//..."]
     test_targets: ["//..."]
     working_directory: gazelle
   gazelle_extension_workspace:
     <<: *common_workspace_flags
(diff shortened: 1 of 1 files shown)


aignas:
Hi, @meteorcloudy, do you know why the RBE CI started failing over the past week? The first guess was that it was because Ubuntu 2004 is too old, but just bumping it does not solve the failure.

meteorcloudy:
We did update the rbe config last week, but not sure if the compile error is caused by that:
```
In file included from external/abseil-cpp~/absl/hash/internal/hash.cc:15:0:
external/abseil-cpp~/absl/hash/internal/hash.h:66:10: fatal error: filesystem: No such file or directory
 #include <filesystem>  // NOLINT
          ^~~~~~~~~~~~
```

Are you able to reproduce the same issue with the `gcr.io/bazel-public/ubuntu2404` docker image?

/cc @coeuvre

meteorcloudy:
I see, you probably need to update the docker container used for the RBE workers, which is currently `ubuntu1804` with a probably very old gcc version

See https://github.com/bazel-contrib/rules_python/blob/5c68ff90030dd5cb8fb4873992a955ac57de7a83/python/private/internal_dev_deps.bzl#L29

Update to `ubuntu2204` could probably work

rickeylev (approved):

PR #3087, opened 2025-07-14
feat: replace /bin/bash with /usr/bin/env bash

fionera:
This allows system which don't have this location (looking at you NixOS) to run the scripts.


aignas (commented):
Could you please also add a note to CHANGELOG? Something like:

> (core) `#!/usr/bin/env bash` is now used as a shebang in the stage1 bootstrap template.

python/private/print_toolchain_checksums.bzl
```diff
@@ -28,6 +28,7 @@ def print_toolchains_checksums(name):
     template = """\
 cat > "$@" <<'EOF'
 #!/usr/bin/env bash
+set -euo pipefail
 
 set -o errexit -o nounset -o pipefail
```
aignas:
```suggestion
set -euo pipefail
```

Ar these not the same? I am happy to leave your option though.

fionera: Oops I was lazy and used GH.dev. Gonna correctly rebase this 😄

fionera: These where added by you btw :P https://github.com/bazel-contrib/rules_python/commit/9555ba8e9ce0902f3d2a315a6e19b8bebe79c0ce#diff-1d13860f11c3e30934c4f69ff153827581f457553c850067514f116cfc7bcc4aR31

aignas: Ah, caught red-handed :D. I thought I added it and then noticed that the stuff was the same and removed. Sorry for this. :D

aignas (approved):
LGTM

Issue #2787, opened 2025-04-18
Support PEP751 lockfile format in bzlmod


aignas:
After `pip lock` lands, this might become a better `requirements.txt` alternative.

References:
* pip (25.1) will export a lockfile - ETA 2 weeks https://github.com/pypa/pip/pull/13213
* packaging has landed the marker resolution (unlocks PDM, poetry, hatch) https://github.com/pypa/packaging/issues/885
* uv is tracking it as an export format https://github.com/astral-sh/uv/issues/12584
* poetry still thinking about it https://github.com/orgs/python-poetry/discussions/10322
* pip-tools keen https://github.com/jazzband/pip-tools/issues/2124

This most likely depends on #2786.

aignas:
#3088 is a duplicate for this.

aignas:
I had a go at this and the number of pieces that have to fall together is roughly as below:
- [ ] Add support for outputing a `pylock.toml` file in our `lock` rule. Fairly trivial, just need to work around the quirk that `uv` will refuse to output the `pylock.toml` file if the filename is not `pylock.toml`.
- [ ] Add support for converting `toml` to `json`. Can be done using the `toml2json` wheels from PyPI, that are pretty lightweight and we could select the host-compatible one at runtime, then we could just `mctx.execute` directly the binary. This or we could execute a python script to do the conversion.
- [ ] Parse the lockfile JSON as per PEP751. This is also trivial.
- [ ] Do not require extras in the lock file to correctly deduce the dependency graph. The `pylock.toml` files do not have this information. We should generate targets for each `provides-extra` in the `whl_library_targets` macro and then expose some of them through the hub repo. Which extra targets we should expose is unclear.

The last one is a blocker. I think we have a ticket for that somewhere. I think the same problem would exist with `uv.lock`.

The only way around that I can see is to:
* Download the wheels (or their METADATA beforehand).
* Write the `Requires-Dist` and `Provides-Extra` contents into the bzlmod lockfile through the new facts mechanism so that this downloading is not needed all the time.
* For each Provides-Extra, add an extra alias.

Alternative could be to use transitions on `py_library`, something like the following:
```
py_library_with_extras(
    name = "my_dep_with_extras",
    dep = "@pypi//actual_dep",
    extras = ["foo", "bar"]
)
```

which would be defined at call site (e.g. the `whl_library` that depends on a package with extras).

[...]

PR #3089, opened 2025-07-14
docs: tell how to emulate dependency groups with pip-compile

rickeylev:
While pyproject.toml is supported by piptools, dependency groups aren't. They can be
emulated by using multiple files. Explain how to do that in the docs and link to
the upstream feature request (https://github.com/jazzband/pip-tools/issues/2062)

Along the way, link to our own feature request for pylock.toml support in pip.parse.

### DIFF
--- a/docs/pypi/lock.md
+++ b/docs/pypi/lock.md
@@ -6,7 +6,7 @@
 :::{note}
 Currently `rules_python` only supports `requirements.txt` format.
 
-#{gh-issue}`3088` tracks `pylock.toml` support.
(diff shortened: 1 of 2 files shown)


docs/pypi/lock.md
```diff
@@ -5,6 +5,8 @@
 
 :::{note}
 Currently `rules_python` only supports `requirements.txt` format.
+
+#{gh-issue}`3088` tracks `pylock.toml` support.
```
aignas:
```suggestion
#{gh-issue}`2787` tracks `pylock.toml` support.
```
#2787 is the right issue

Issue #3082, opened 2025-07-12
chore: Refactor gazelle docs


dougthor42:
The Gazelle docs are fully contained within `gazelle/README.md`. The docs are getting larger and this is starting to become harder to maintain and reference.

I'd like to see the docs added to the readthedocs page, with some logical section splitting.

I propose a structure:

```
./docs/
  ~~gazelle.md~~     # removed; TOC tree in index.md references gazelle/docs/index.md
./gazelle/
  docs/              # new folder
    index.md         # basic configuration, usage
    directives.md    # Details on # gazelle:foo comments found in BUILD.bazel files
    annotations.md   # Details on # gazelle:foo comments found in python source files
    development.md   # source code layout, how to add/run tests, bump versions, write docs, etc.
  README.md          # Most content moved to docs/*.md files; what's left is basically a stub
```

dougthor42:
@aignas @rickeylev Which would you prefer to review?

One big PR that migrates everything at once?

Or a 4 PRs?

1.  Core changes like BUILD files and the high-level Gazelle info (index).
2.  annotations
3.  directives
4.  development

PR #3093, opened 2025-07-15
feat(gazelle): `python_proto_naming_convention` directive controls `py_proto_library` naming

shaldengeki:
Closes https://github.com/bazel-contrib/rules_python/issues/3081.

This adds support in the Gazelle plugin for controlling how the generated `py_proto_library` rules are named; support for these was originally added in https://github.com/bazel-contrib/rules_python/pull/3057. We do this via a new Gazelle directive, `python_proto_naming_convention`, which is similar to `python_library_naming_convention` and the like, except it interpolates `$proto_name$`, which is the `proto_library` rule minus any trailing `_proto`. We default to `$proto_name$_py_pb2`.

For instance, for a `proto_library` named `foo_proto`, the default value would generate `foo_py_pb2`, aligning with [the convention stated in the Bazel docs.](https://bazel.build/reference/be/protocol-buffer#py_proto_library)


gazelle/README.md
```diff
@@ -208,6 +208,8 @@ Python-specific directives are as follows:
...
 | Controls the `py_test` naming convention. Follows the same interpolation rules as `python_library_naming_convention`.                                                                                                                                                                           | |
+| `# gazelle:python_proto_naming_convention`                                                                                                                                                                                                                                                       | `$proto_name$_py_pb2` |
+| Controls the `py_proto_library naming convention. It interpolates `$proto_name$` with the proto_library rule name, minus any trailing _proto. E.g. if the proto_library name is `foo_proto`, setting this to `$proto_name$_my_lib` would render to `foo_my_lib`.                                                                                                                                                                          | |
```
shaldengeki: I didn't add a details section for this below, because none of the other naming convention directives have this, and I think it's pretty self-explanatory?

CHANGELOG.md
```diff
@@ -113,6 +113,8 @@ END_UNRELEASED_TEMPLATE
   dep is not added to the {obj}`py_test` target.
 * (gazelle) New directive `gazelle:python_generate_proto`; when `true`,
   Gazelle generates `py_proto_library` rules for `proto_library`. `false` by default.
+* (gazelle) New directive `gazelle:python_proto_naming_convention`; controls
+  naming of `py_proto_library` rules. See `gazelle/README.md`.
```
dougthor42: nit: please remove "See `gazelle/README.md`." - I'll be refactoring Gazelle docs soon (#3082) and don't want to have to remember to update this :upside_down_face:

gazelle/python/testdata/directive_python_proto_naming_convention/README.md
```diff
@@ -0,0 +1,10 @@
+# Directive: `python_proto_naming_convention`
+
+This test case asserts that the `# gazelle:python_proto_naming_convention` directive
+correctly:
+
+1.  Has no effect on pre-existing `py_proto_library` when `gazelle:python_generate_proto` is disabled.
+2.  Uses the default value when proto generation is on and `python_proto_naming_convention` is not set.
+2.  Uses the provided naming convention when proto generation is on and `python_proto_naming_convention` is set.
```
dougthor42: nit: 3.

dougthor42: Let's add a test case for:

4. existing py_proto_library with name `foo_py_proto` and `python_proto_naming_convention` is set to `$proto_name$_bar`.

Does Gazelle rename the target or leave it alone?

gazelle/python/testdata/directive_python_proto_naming_convention/README.md
```diff
@@ -0,0 +1,10 @@
...
+
+[gh-3081]: https://github.com/bazel-contrib/rules_python/issues/3081
```
dougthor42: nit: unused link. Use it or remove it.

gazelle/README.md
dougthor42: Please add a note that says the python library will always be imported as `foo_pb2` no matter what the `py_proto_library` target name is.

So if we generate `name = "foo_py_pb2"`, then the python import is `import foo_pb2`.

```starlark
proto_library(
    name = "foo_proto",
    srcs = ["foo.proto"],
    ...
)

py_proto_library(
    name = "and_now_for_something_completely_different",
    srcs = [":foo_proto"],
)
```

```python
import foo_pb2  # not "import and_now_for_something_completely_different"
```

dougthor42: I think that Gazelle also can't discover the `py_proto_library` rules yet, so add that note too. E.g.:

> Note that Gazelle is not currently able to map an import `import foo_pb2` to a `py_proto_library` target.

or something like that.

dougthor42: ACK SGTM :+1:.

dougthor42 (commented):
Nice! I tested this on our monorepo and generated 227 `py_proto_library` targets. Hot damn!

There are a variety of nits and requests but otherwise this is looking great, thanks!

gazelle/python/testdata/directive_python_proto_naming_convention/README.md
shaldengeki: Oh, good idea. I'm pretty sure that right now we'd leave the pre-existing target alone and generate a second target following the convention.

*(later: oh, looks like we rename it instead. still, I think the following holds.)*

If I'm considering what we _want_ Gazelle to do, I think such a pre-existing target would interfere with dependency resolution -- we'd have two `py_proto_library` that could potentially fill the same import in that case.

So I don't think we want Gazelle to leave it alone _and generate a new target_; we'd want it to either:
1. rename the target, or
2. keep the preexisting target as the sole `py_proto_library`.

Of these two, I think the most "Gazelle-y" would be to keep the preexisting target around as the sole `py_proto_library`. That's what we appear to do with `py_library`, and that's also what the golang extension does with `go_proto_library` (I tested both by renaming an existing target and re-running Gazelle). I can write a test for that and adjust the logic to match.

dougthor42 (approved):
:rocket:

PR #3096, opened 2025-07-15
fix: Fix whl_library in bazel vendor mode

omar-droubi:
Resolves https://github.com/bazel-contrib/rules_python/issues/3079
- Added PYTHONHOME to whl_library execution environment. Without it, the python interpreter is getting confused where it's running from when bazel --vendor_dir is used
- In _get_toolchain_unix_cflags, the real path for python is used instead of the symlink.


omar-droubi:
Hi @aignas, with regards to the tests, the bug this PR resolves occurs when there is a "bazel clean --expunge" between the "bazel vendor" and the "bazel build" commands. Is it possible to update the "integration_test_bzlmod_ubuntu_vendor" with a "bazel clean" ? I am not sure how impactful that would be on the CI performance.

<code>integration_test_bzlmod_ubuntu_vendor:
    <<: *reusable_build_test_all
    name: "examples/bzlmod: bazel vendor"
    working_directory: examples/bzlmod
    platform: ubuntu2204
    shell_commands:
    - "bazel clean --expunge"
    - "bazel vendor --vendor_dir=./vendor //..."
    - "bazel clean --expunge"
    - "bazel build --vendor_dir=./vendor //..."
    - "rm -rf ./vendor"
</code>

aignas:
I would like to avoid adding `bazel clean` in the CI, maybe we could just add a line that does `bazel vendor` if it is not present yet. If we have regressions in the future, we can look into `bazel clean` usage.

aignas (approved):
Thank you!

PR #3100, opened 2025-07-16
fix: support debian multiarch with local toolchains

rickeylev:
Apparently, there is a "multiarch" style of Python installations, where the shared
libraries can be in a sub-directory. The best docs I could find about this are
https://wiki.debian.org/Python/MultiArch.

To fix, looking at the `MULTIARCH` sysconfig var tells what subdirectory, if any
should be looked in.

Along the way, fix a changelog issue reference url.

Fixes https://github.com/bazel-contrib/rules_python/issues/3099

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -85,10 +85,12 @@ END_UNRELEASED_TEMPLATE
   ([#2503](https://github.com/bazel-contrib/rules_python/issues/2503)).
 * (toolchains) `local_runtime_repo` now checks if the include directory exists
   before attempting to watch it, fixing issues on macOS with system Python
-  ({gh-issue}`3043`).
+  ([#3043](https://github.com/bazel-contrib/rules_python/issues/3043)).
 * (pypi) The pipstar `defaults` configuration now supports any custom platform
   name.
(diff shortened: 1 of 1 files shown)


aignas (approved):

rickeylev:
cc @ekeranen this is slightly different than your patch -- it symlinks `origin -> lib/{multiarch_dirname}/{libname}`. In your original patch, the second symlink call could overwrite the first.

Hrm, though -- maybe that is wrong? because the libs will be in a sub-dir? Which...all the `libpython*` stuff _probably_ wants to be in the same dir. But idk if _both_ the regular and multiarch locations could exist? If they do, then what?

Looking at my local system, /usr/bin/python's MULTIARCH does return a value, however, LIBDIR is already pointing there. This is py 3.12. Is py3.8 doing some older behavior?

ekeranen:
What about implementing this only as a fallback? 

```
if origin.exists: 
    ...
elif origin_multiarch.exists:
   # ... this python installation doesn't set LIBDIR correctly. Fallback to LIBDIR/MULTIARCH. 
   ...
```

rickeylev:
yeah, good idea. done.

PR #3104, opened 2025-07-17
fix: normalize stub_path in repl.bzl

### LINKED ISSUE #3101: REPL invocation fails with "path is not normalized"
# 🐞 bug report

### Affected Rule

The issue is caused by the rule: @rules_python//python/bin:repl 

### Is this a regression?

Not strictly speaking. I do not believe the REPL feature was available prior to 1.5.0. While I've been able to reproduce this issue with both 1.5.0 and 1.5.1, I recall using the REPL feature successfully with a clone of rules_python before the 1.5.0 release.

### Description

When running the example command-line from https://github.com/bazel-contrib/rules_python/blob/main/docs/repl.md I get a an error saying "path is not normalized" and the Python shell is not started.

```sh
$ bazel run @rules_python//python/bin:repl --@rules_python//python/bin:repl_dep=@rules_python//tools:wheelmaker
```

## 🔬 Minimal Reproduction

I ran the above command-line inside a brand new and empty Bazel module with these `MODULE.bazel` contents:

```bzl
module(name="test", version="0.1.0")

bazel_dep(name="rules_python", version="1.5.1")
```

## 🔥 Exception or Error

<pre><code>
Starting local Bazel server (8.3.1) and connecting to it...
INFO: Analyzed target @@rules_python+//python/bin:repl (82 packages loaded, 3620 targets configured).
INFO: Found 1 target...
Target @@rules_python+//python/bin:repl up-to-date:
  bazel-bin/external/rules_python+/python/bin/repl
  bazel-bin/external/rules_python+/python/bin/repl_py.py
INFO: Elapsed time: 11.545s, Critical Path: 0.89s
INFO: 6 processes: 1 action cache hit, 6 internal.
INFO: Build completed successfully, 6 total actions
INFO: Running command line: bazel-bin/external/rules_python+/python/bin/repl
Python 3.11.13 (main, Jun  9 2025, 18:12:55) [Clang 20.1.4 ] on linux
  File "/home/jpwoodbu/.cache/bazel/_bazel_jpwoodbu/4f83e7249d0fd942371631f62dae1282/execroot/_main/bazel-out/k8-fastbuild/bin/external/rules_python+/python/bin/repl.runfiles/rules_python+/python/bin/repl_py.py", line 45, in <module>
    start_repl()
  File "/home/jpwoodbu/.cache/bazel/_bazel_jpwoodbu/4f83e7249d0fd942371631f62dae1282/execroot/_main/bazel-out/k8-fastbuild/bin/external/rules_python+/python/bin/repl.runfiles/rules_python+/python/bin/repl_py.py", line 38, in start_repl
    bazel_runfiles.Rlocation(STUB_PATH),
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jpwoodbu/.cache/bazel/_bazel_jpwoodbu/4f83e7249d0fd942371631f62dae1282/execroot/_main/bazel-out/k8-fastbuild/bin/external/rules_python+/python/bin/repl.runfiles/rules_python+/python/runfiles/runfiles.py", line 176, in Rlocation
    raise ValueError('path is not normalized: "%s"' % path)
ValueError: path is not normalized: "rules_python+/../rules_python+/python/bin/repl_stub.py"
</code></pre>

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 22.04.5 LTS
Release:	22.04
Codename:	jammy

Reproduced on a second machine...

$ lsb_release -a
No LSB modules are available.
Distributor ID:	Debian
Description:	Debian GNU/Linux 12 (bookworm)
Release:	12
Codename:	bookworm
  </code>
</pre>

<pre>
  <code>
$ bazel version
Bazelisk version: v1.20.0
Build label: 8.3.1
Build time: Mon Jun 30 16:23:40 2025 (1751300620)
Build timestamp: 1751300620
Build timestamp as int: 1751300620
  </code>
</pre>

**Rules_python version:**

<pre>
  <code>
1.5.1
  </code>
</pre>

**Anything else relevant?**
I don't think so.

jpwoodbu:
When a REPL target is run from an external Bazel module, the `stub_path` can have path components in it (e.g. "/..") which get rejected by the `Rlocation()` function in `runfiles.py` for not being normalized. This commit normalizes the path before it's passed to `Rlocation()`.

Fixes #3101

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -93,6 +93,8 @@ END_UNRELEASED_TEMPLATE
   ([#3099](https://github.com/bazel-contrib/rules_python/issues/3099)).
 * (pypi) Expose pypi packages only common to all Python versions in `all_requirements`
   ([#2921](https://github.com/bazel-contrib/rules_python/issues/2921)).
+* (repl) Normalize the path for the `REPL` stub to make it possible to use the
+  default stub template from outside `rules_python` ({gh-issue}`3101`).
 
 {#v0-0-0-added}
 ### Added
--- a/python/private/repl.bzl
+++ b/python/private/repl.bzl
@@ -1,12 +1,10 @@
 """Implementation of the rules to expose a REPL."""
 
 load("//python:py_binary.bzl", _py_binary = "py_binary")
-load("@bazel_skylib//lib:paths.bzl", "paths")
 
 def _generate_repl_main_impl(ctx):
     stub_repo = ctx.attr.stub.label.repo_name or ctx.workspace_name
-    unnormalized_path = "/".join([stub_repo, ctx.file.stub.short_path])
-    stub_path = paths.normalize(unnormalized_path)
+    stub_path = "/".join([stub_repo, ctx.file.stub.short_path])
 
     out = ctx.actions.declare_file(ctx.label.name + ".py")
 
--- a/python/private/repl.bzl
+++ b/python/private/repl.bzl
@@ -1,10 +1,12 @@
 """Implementation of the rules to expose a REPL."""
 
 load("//python:py_binary.bzl", _py_binary = "py_binary")
+load("@bazel_skylib//lib:paths.bzl", "paths")
 
 def _generate_repl_main_impl(ctx):
     stub_repo = ctx.attr.stub.label.repo_name or ctx.workspace_name
-    stub_path = "/".join([stub_repo, ctx.file.stub.short_path])
+    unnormalized_path = "/".join([stub_repo, ctx.file.stub.short_path])
+    stub_path = paths.normalize(unnormalized_path)
 
     out = ctx.actions.declare_file(ctx.label.name + ".py")
 
--- a/python/private/repl_template.py
+++ b/python/private/repl_template.py
@@ -5,7 +5,9 @@ from pathlib import Path
 
 from python.runfiles import runfiles
 
-STUB_PATH = "%stub_path%"
+# runfiles.py will reject paths which aren't normalized, which can happen when the REPL rules are
+# used from a remote module.
+STUB_PATH = os.path.normpath("%stub_path%")
 
 
 def start_repl():


aignas:
I saw windows errors had mixed path separators and I wonder if using the bazel skylib helper is causing this.

Maybe trying out os.path.norm would be also an option - the path does not have to exist for it to work, so I think it may be still OK.

jpwoodbu:
> I saw windows errors had mixed path separators and I wonder if using the bazel skylib helper is causing this.

Did you get an error when trying this change? Was the error specific to Windows? Did it work for you on another platform? Would you mind sharing the error message?
 
> Maybe trying out os.path.norm would be also an option - the path does not have to exist for it to work, so I think it may be still OK.

I don't think I understand. Can you elaborate on what you mean by, "the path does not have to exist for it to work"? Are you referring to the path to the stub? If it doesn't exist, wouldn't that cause a fatal error?

aignas:
See the errors in the CI: https://buildkite.com/organizations/bazel/pipelines/rules-python-python/builds/12251/jobs/01981aa5-6b45-4c43-b5bd-8d6dfc408065/artifacts/01981aab-bc3b-4fae-a258-79dd924dfb80

In my tests
```
p = "a/../b"
os.norm.path(p) # returns "b"

pathlib.Path(p).resolve # returns absolute path
```

I meant that `p` does not have to exist on the filesystem for os.norm.path to work the way we want. Obviously for everything to work it will have to be present on the filesystem, but it should be a relative path and not an absolute if we want to pass it to the `runfiles` library.

jpwoodbu:
Understood. Thanks for the reply! I've added another commit which switches to using Python's `normpath()`. Let's see if the tests work now.

I've also tried to clean up the mess I made when I committed the tab completion fix to this same branch. Hopefully everything is better now.

aignas (approved):

Issue #3097, opened 2025-07-15
Gazelle will confuse stdlib packages/modules with sibling modules


dougthor42:
# 🐞 bug report

### Affected Rule

gazelle

### Is this a regression?

It's been like this for as long as I've been using rules_python.

### Description

If there's a file called, say, `typing.py` and a sibling file `foo.py` has `from typing import TYPE_CHECKING` or similar, gazelle will add `:typing` to the `foo` target deps (or to the package `srcs` for package generation mode).

## 🔬 Minimal Reproduction

Dir structure:

```
$ tree gazelle/python/testdata/stdlib_confusion/
gazelle/python/testdata/stdlib_confusion/
├── file_based
│   ├── BUILD.in
│   ├── BUILD.out
│   ├── foo.py
│   ├── __init__.py
│   └── typing.py
├── package_based
│   ├── BUILD.in
│   ├── BUILD.out
│   ├── foo.py
│   ├── __init__.py
│   └── typing.py
├── README.md
├── test.yaml
└── WORKSPACE
```

```
--- FAIL: TestGazelleBinary (0.00s)
    --- FAIL: TestGazelleBinary/stdlib_confusion (0.02s)
        python_test.go:186: stdlib_confusion/file_based/BUILD diff (-want,+got):
              (
                """
                ... // 6 identical lines
                    srcs = ["foo.py"],
                    visibility = ["//:__subpackages__"],
            +       deps = [":typing"],
                )
              
                ... // 5 identical lines
                """
              )
        python_test.go:186: stdlib_confusion/package_based/BUILD diff (-want,+got):
              (
                """
                ... // 5 identical lines
                        "__init__.py",
                        "foo.py",
            +           "typing.py",
                    ],
                    visibility = ["//:__subpackages__"],
                )
                """
              )
```

amartani:
I'm sending https://github.com/bazel-contrib/rules_python/pull/3106 to allow disabling the current sibling import resolution behavior, which would include this situation.

PR #3106, opened 2025-07-19
fix(gazelle): Do not resolve absolute imports to sibling modules

amartani:
Currently, gazelle allows absolute imports to be resolved to sibling modules: an `import foo` statement will resolve to a `foo.py` file in the same folder if such file exists. This seems to be a Python 2 behavior (ie. pre-`from __future__ import absolute_import`), and doesn't work on the current rules_python setup.

This behavior is explicitly tested in the [siblings_import](https://github.com/bazel-contrib/rules_python/tree/cbe6d38d01c14de46d90ea717d0f2090117533fa/gazelle/python/testdata/sibling_imports) test case. However, recreating the exact same repository layout from this test case and running `bazel test //pkg:unit_test`, the test fails with the import failing.

This PR adds a new directive, `gazelle:python_resolve_sibling_imports`, to allow disabling such behavior.

The actual changes are in 3 places:

- In `gazelle/python/target.go`, the directive is added to `if t.siblingSrcs.Contains(fileName) && fileName != filepath.Base(dep.Filepath)`, which is where the import is converted to a full absolute import if it matches a sibling file;
- In `gazelle/python/generate.go`, the handling of `conftest.py` was dependent on this behavior (ie. it added a dependency on the module `conftest`, assuming that it would be resolved to the relative module). That was modified to compute the full absolute module path instead.
- In `gazelle/python/resolve.go`, resolve relative imports even when using file generation mode.

I also explicitly added `gazelle:python_resolve_sibling_imports true` to any test that breaks if the default value of this directive is changed to `false`.


aignas:
Thanks for the PR. Since the current gazelle code would result in un-runnable code, is there a good reason to not default to the correct behaviour and have the flag in case some user wants to do this step-wise?

amartani:
> Thanks for the PR. Since the current gazelle code would result in un-runnable code, is there a good reason to not default to the correct behaviour and have the flag in case some user wants to do this step-wise?

Honestly, my main reason is a feeling of "I must be missing something" about it. The sibling import behavior is not accidental: there's code added that only exists for resolving sibling imports, a test that explicitly tests for this behavior, and multiple other tests that implicitly depend on it. And, as far as I know, it was all written much after python2 was unsupported, which is the only reason that I can think of for why this would be desired. That said, I'd be happy to switch the default if you believe that would be ok.

aignas:
I'll wait for other to chime in on here. @arrdem, your insight from the Aspect.dev point of view would be valuable here as well.

CHANGELOG.md
```diff
@@ -116,6 +116,10 @@ END_UNRELEASED_TEMPLATE
   dep is not added to the {obj}`py_test` target.
 * (gazelle) New directive `gazelle:python_generate_proto`; when `true`,
   Gazelle generates `py_proto_library` rules for `proto_library`. `false` by default.
+* (gazelle) Added directive `# gazelle:python_resolve_sibling_imports` (default
+  `enabled` for backwards-compatibility). When enabled, gazelle will allow absolute
```
dougthor42: `s/enabled/true`

gazelle/python/testdata/sibling_imports_disabled/README.md
```diff
@@ -0,0 +1,17 @@
...
+This test case asserts that imports from sibling modules are NOT resolved as
+absolute imports when the `python_resolve_sibling_imports` directive is
+disabled. It covers 3 different types of imports in `pkg/unit_test.py`:
+
+- `import a` - resolves to the root-level `a.py` instead of the sibling
+  `pkg/a.py`
+- `import test_util` - resolves to the root-level `test_util.py` instead of
+  the sibling `pkg/test_util.py`
+- `from b import run` - resolves to the root-level `b.py` instead of the
+  sibling `pkg/b.py`
```
dougthor42, line 17: Can we add test cases for:

+ relative imports. `import .test_util` resolves to the sibling `pkg/test_util.py`
+ The `# gazelle:python_generation_mode file` case
+ Having a custom `# gazelle:python_root`
+ The "shadow stdlib" case from #3097.

Not all of these need to be separate complete test case directory. I think relative imports and shadowed stdlib checks are pretty easy to add within this test case directory.

aignas:
I'm +1 for keeping the default and creating a ticket to switch the behaviour when we release 2.0.

gazelle/python/testdata/sibling_imports_disabled/README.md
amartani: I added relative imports and shadowing stdlib to this test case, and copied this entire test case into a new "siblings_import_disabled_file_mode" to test all these in file mode.

This surfaced that we were not processing relative imports in file mode - I think the feature was relying on sibling imports behavior. See changes to `gazelle/python/resolve.go` for resolving that.

For the custom `# gazelle:python_root` - since I flipped the default, all the current tests are now being exercised with the new semantics, so I think there's pretty good coverage of that (eg. the `monorepo` test case). Is there any specific case that you think it worth adding a new test for?

amartani:
Changes since the initial PR request:
- Switched default to `python_resolve_sibling_imports false`;
- Re-worded PR title and CHANGELOG entry as a fix instead of feat / added since we are changing the default behavior;
- Adjusted behavior of relative imports when using `gazelle:experimental_allow_relative_imports true` and `gazelle:python_generation_mode file` to also resolve relative imports properly.

dougthor42:
> I'm +1 for keeping the default and creating a ticket to switch the behaviour when we release 2.0.

Oooo sounds like we're about to fight!

Nah I'm kidding of course. I definitely do **not** feel strongly about adjusting the default to the correct behavior. I'm at like a 3 out of 10. So it'll be pretty easy to convince me otherwise.

amartani:
> > I'm +1 for keeping the default and creating a ticket to switch the behaviour when we release 2.0.
> 
> Oooo sounds like we're about to fight!
> 
> Nah I'm kidding of course. I definitely do **not** feel strongly about adjusting the default to the correct behavior. I'm at like a 3 out of 10. So it'll be pretty easy to convince me otherwise.

My understanding of the [documented policy](https://rules-python.readthedocs.io/en/latest/contributing.html#breaking-changes) is that, if this is considered a breaking change but "is not too disruptive" (which I think this would fit), then we want to:

- Merge this with `gazelle: python_resolve_sibling_imports true` (preserve existing behavior);
- After the next _minor_ version (ie. after 1.6 is "cut"), switch default to `false`;
- On 2.0, remove the directive (if we believe there is no reasonable use case for the previous behavior).

aignas:
Ah sorry... :D I misunderstood/misread/did-not-read properly:

> Thus my vote is to adjust the default to the correct behavior: python_resolve_sibling_imports false.

+1 to go with your suggestion. Gazelle does not have the same stability guarantees as rules_python core. This breaking change should not be too disruptive the way I see it.

gazelle/python/testdata/sibling_imports_disabled/README.md
dougthor42: > Is there any specific case that you think it worth adding a new test for?

Nah, what is present LGTM. No action needed.

dougthor42 (approved):
Thanks, new tests LGTM.

I verified the most recent commit a00ec81df95e1a69f8d8ba2db912d946cf3f72e3 on our repo. All :+1: :tada:

PR #3107, opened 2025-07-20
fix(pypi): expose pypi packages only common to all python versions

### LINKED ISSUE #2921: all_requirements fails for multiple python versions
# 🐞 bug report

### Affected Rule

The `all_requirements` constant in the generated requirements.bzl file in the PyPI hub repo.

### Is this a regression?

Not really.

### Description

When running with multiple python version for a single hub, the lock files will generally not contain the exact same dependencies. This means that when using `all_requirements` from the generated `@python_deps//:requirements.bzl` file, it will reference some dependencies which does not exist for all the python versions. In practice this means an error is raised by `@rules_python//python/config_settings:is_not_matching_current_config`.

Instead, `all_requirements` should only return the requirements for the python version matching the current toolchain, OR alternatively only the requirements common to all python versions.

## 🔬 Minimal Reproduction
Steps:
- Create `requirements.in` with:
```
mypy==1.15.0
```
- Generate lock files `requirements_3_10.txt` and `requirements_3_11.txt` for Python 3.10 and Python 3.11, respectively. 

-  Add to MODULE.Bazel:
```
python.toolchain(
    is_default = True,
    python_version = "3.11",
)

pip.parse(
    hub_name = "python_deps",
    python_version = "3.11",
    requirements_linux = ":requirements_lock_3_11.txt",
)

pip.parse(
    hub_name = "python_deps",
    python_version = "3.10",
    requirements_linux = ":requirements_lock_3_10.txt",
)
use_repo(pip, "python_deps")
```
- Now try to use `all_requirements` - e.g. for running checks on all licenses:
```
load("@python_deps//:requirements.bzl", "all_requirements")

license_test(
    name = "all_python_license_check",
    deps = all_requirements,
)
```

## 🔥 Exception or Error

Running the described example on the default toolchain (python 3.11) results in the error:
```
ERROR: {path_to_bazel_cache}/external/rules_python++pip+python_deps/tomli/BUILD.bazel:5:12: configurable attribute "actual" in @@rules_python++pip+python_deps//tomli:_no_matching_repository doesn't match this configuration: No matching wheel for current configuration's Python version.

configuration settings:
    //_config:is_cp310

To determine the current configuration's Python version, run:

For the current configuration value see the debug message above that is
printing the current flag values. If you can't see the message, then re-run the
build to make it a failure instead by running the build with:
    --@@rules_python+//python/config_settings:current_config=fail

However, the command above will hide the `bazel config <config id>` message.


This instance of @@rules_python++pip+python_deps//tomli:_no_matching_repository has configuration identifier ... . To inspect its configuration, run: bazel config ... .

```

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
Linux Ubuntu
  </code>
</pre>

<pre>
  <code>
8.2.1
  </code>
</pre>

**Rules_python version:**

<pre>
  <code>
1.4.1
  </code>
</pre>

dotoleeoak:
Closes #2921

For a single pypi repo with multiple python versions, `all_requirements` fail when a pypi package supports Python version A but not version B. In this case, the pypi package would be included only in requirements lock file for version A, not in one for version B. However, the failure occurs since the package is included in `all_requirements` even for Python version B.
(Minimal reproduction: https://github.com/dotoleeoak/rules-python-2921-repro)

This happens since `packages` parameter for `hub_repository` targets are including all packages across all requirement lock files. Instead of union of packages, intersection of packages for requirement files should be passed to `packages` and exposed to `all_requirements` macro, so that those packages are compatible with all Python versions.

### DIFF
--- a/tests/pypi/extension/extension_tests.bzl
+++ b/tests/pypi/extension/extension_tests.bzl
@@ -331,38 +331,26 @@ new-package==0.0.1 --hash=sha256:deadb00f2
         "pypi": {
             "new_package": {
                 "pypi_316_new_package": [
-                    struct(
-                        config_setting = None,
-                        filename = None,
-                        target_platforms = None,
+                    whl_config_setting(
                         version = "3.16",
                     ),
                 ],
             },
             "old_package": {
                 "pypi_315_old_package": [
-                    struct(
-                        config_setting = None,
-                        filename = None,
-                        target_platforms = None,
+                    whl_config_setting(
                         version = "3.15",
                     ),
                 ],
             },
             "simple": {
                 "pypi_315_simple": [
-                    struct(
-                        config_setting = None,
-                        filename = None,
-                        target_platforms = None,
+                    whl_config_setting(
                         version = "3.15",
                     ),
                 ],
                 "pypi_316_simple": [
-                    struct(
-                        config_setting = None,
-                        filename = None,
-                        target_platforms = None,
+                    whl_config_setting(
                         version = "3.16",
                     ),
                 ],
--- a/tests/pypi/extension/extension_tests.bzl
+++ b/tests/pypi/extension/extension_tests.bzl
@@ -285,6 +285,116 @@ def _test_simple_multiple_requirements(env):
 
 _tests.append(_test_simple_multiple_requirements)
 
+def _test_simple_multiple_python_versions(env):
+    pypi = _parse_modules(
+        env,
+        module_ctx = _mock_mctx(
+            _mod(
+                name = "rules_python",
+                parse = [
+                    _parse(
+                        hub_name = "pypi",
+                        python_version = "3.15",
+                        requirements_lock = "requirements_3_15.txt",
+                    ),
+                    _parse(
+                        hub_name = "pypi",
+                        python_version = "3.16",
+                        requirements_lock = "requirements_3_16.txt",
+                    ),
+                ],
+            ),
+            read = lambda x: {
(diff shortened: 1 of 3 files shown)


tests/pypi/extension/extension_tests.bzl
```diff
@@ -285,6 +285,116 @@ def _test_simple_multiple_requirements(env):
...
+                        target_platforms = None,
+                        version = "3.15",
+                    ),
+                ],
+                "pypi_316_simple": [
+                    struct(
+                        config_setting = None,
+                        filename = None,
+                        target_platforms = None,
+                        version = "3.16",
```
aignas:
```suggestion
                    whl_config_setting(
                        version = "3.16",
                    ),
                ],
            },
            "old_package": {
                "pypi_315_old_package": [
                    whl_config_setting(
                        version = "3.15",
                    ),
                ],
            },
            "simple": {
                "pypi_315_simple": [
                    whl_config_setting(
                        version = "3.15",
                    ),
                ],
                "pypi_316_simple": [
                    whl_config_setting(
                        version = "3.16",
```

aignas (commented):
Logic LGTM, but it would be also good to add a line in CHANGELOG in the Fixed section to ensure that users get informed about the change.

dotoleeoak:
It seems the current CI failures come from GitHub's partial outage
https://www.githubstatus.com/

aignas (approved):

PR #3109, opened 2025-07-21
feat(gazelle): Gazelle resolves module-level imports of `py_proto_library`

### LINKED ISSUE #1703: [Gazelle][proto] How to automatically resolve protobuf target

shaldengeki:
Closes #1703.

This adds support in Gazelle for resolving module-level imports of `py_proto_library` rules. Note that this does _not_ support:

1. Imports of module contents, like messages, services, or enums,
2. Imports of other rules, like GRPC

See the tests for examples of how this works. I would ideally like to implement the first one as well, but I think the "right" way to do that is to use something that's [only present in Gazelle 0.38.](https://github.com/bazel-contrib/bazel-gazelle/pull/1844) So that'll have to wait on an upgrade of Gazelle in this repo; in the meanwhile, I think this is worth landing.


gazelle/python/testdata/resolves_proto_imports/test1_generates_dependency/BUILD.out
```diff
@@ -0,0 +1,8 @@
+load("@rules_python//python:defs.bzl", "py_library")
+
+py_library(
+    name = "test1_generates_dependency",
+    srcs = ["bar.py"],
+    visibility = ["//:__subpackages__"],
+    deps = ["//test1_generates_dependency/foo:test1_generates_dependency_foo_py_pb2"],
+)
```
dougthor42, line 8: Nit: LF@EOF, here and elsewhere.

gazelle/python/testdata/resolves_proto_imports/test2_generates_using_resolve/BUILD.out
```diff
@@ -0,0 +1,10 @@
+load("@rules_python//python:defs.bzl", "py_library")
+
+# gazelle:resolve py test2_generates_using_resolve.bar.bar_pb2 //test2_generates_using_resolve/bar:bar_py_pb2
```
dougthor42, line 3: ~~I'm not sure I understand this test case - can you elaborate?~~ Unless I'm blind, ~~it looks like you're resolving the same thing that gazelle would resolve by default. So... what's that testing?~~

~~I would assume that using a `resolve` directive would typically be for cases where, say, the `bar_py_pb2` `py_proto_library` _doesn't exist_ but we want Gazelle to resolve `import test2_generates_using_resolve.bar.bar_pb2` to something else.~~

Yes, I'm quite blind I guess.

Could you adjust the resolve directive to be easier to ID that it's resolving to a target that doesn't exist? Maybe:

```
# gazelle:resolve py test2_generates_using_resolve.bar.bar_pb2 //some:target
```

gazelle/python/testdata/resolves_proto_imports/test4_generates_imports_for_multiple_proto_srcs/foo/BUILD.out
```diff
@@ -0,0 +1,19 @@
+load("@rules_proto//proto:defs.bzl", "proto_library")
+load("@com_google_protobuf//bazel:py_proto_library.bzl", "py_proto_library")
+
+# gazelle:python_generate_proto true
+
+proto_library(
+    name = "test4_generates_imports_for_multiple_proto_srcs_foo_proto",
+    srcs = [
+        "bar.proto",
+        "foo.proto",
```
dougthor42, line 10: Can we add a test case with `# gazelle:proto file` mode?

One `proto_library` per `.proto` --> one `py_proto_library` per `proto_library` --> the final `py_library` will have multiple `deps`.

gazelle/python/generate.go
```diff
@@ -42,6 +42,8 @@ const (
 	pyTestEntrypointTargetname  = "__test__"
 	conftestFilename            = "conftest.py"
 	conftestTargetname          = "conftest"
+	protoRelKey                 = "_proto_rel"
```
dougthor42, line 45: What's `Rel` in this context?

gazelle/python/resolve.go
```diff
@@ -114,6 +125,30 @@ func importSpecFromSrc(pythonProjectRoot, bzlPkg, src string) resolve.ImportSpec
...
 
+func isProtoLibrary(r *rule.Rule) bool {
+	return r.Kind() == pyProtoLibraryKind
+}
+
+func importsProtoLibrary(cfg *pythonconfig.Config, r *rule.Rule) []resolve.ImportSpec {
+	specs := []resolve.ImportSpec{}
+
+	// Determine the root module and emit an import for that,
+	// i.e. for //foo:foo_py_pb2, we'd get foo.foo_pb2
```
dougthor42, line 136: This is not always true. Depending on `# gazelle:python_root`, any one of these is possible:

```
//some/path/foo:foo_py_pb2
foo.foo_pb2                     # python_root is //some/path
path.foo.foo_pb2                # python_root is //some
some.path.foo.foo_pb2           # python_root is project root, default case.
```

I think we'll need a test case for `# gazelle:python_root`.

```
.
+ src/
  + BUILD.bazel        # gazelle:python_root
  + my_pkg/
    + protos/
      + foo.proto
      + BUILD.bazel
    + subpkg/
    + __init__.py
    + main.py
    + foo.py
    + BUILD.bazel
pyproject.toml
deploy.py
BUILD.bazel
```

In the above tree, `src/my_pkg/main.py` will have:

```python
import my_pkg.foo as foo
import my_pkg.subpkg.bar as bar
import my_pkg.proto.foo_pb2 as foo_pb2
```

While the target will be:

```starlark
# src/my_pkg/BUILD.bazel
py_library(
    name = "main",
    srcs = ["main.py"],
    imports = "..",
    deps = [
        "//src/my_pkg:foo",
        "//src/my_pkg/subpkg:bar",
        "//src/my_pkg/proto:foo_py_pb2",
    ],
)
```

Specifically: the Bazel target will include information (`src`) not found in the python import string.

gazelle/python/resolve.go
```diff
@@ -114,6 +125,30 @@ func importSpecFromSrc(pythonProjectRoot, bzlPkg, src string) resolve.ImportSpec
...
+	protoRelAttr := r.PrivateAttr(protoRelKey)
+	protoSrcsAttr := r.PrivateAttr(protoSrcsKey)
+	if protoRelAttr == nil || protoSrcsAttr == nil {
+		return nil
+	}
+
+	protoRel := protoRelAttr.(string)
+	for _, protoSrc := range protoSrcsAttr.([]string) {
+		generatedPbFileName := strings.TrimSuffix(protoSrc, ".proto") + "_pb2.py"
```
dougthor42, line 145: Future work: if the proto import is guarded with `if TYPE_CHECKING`, then only include the `_pb2.pyi` file in the `pyi_deps` target attribute.

PR #3111, opened 2025-07-21
feat(pypi): implement a new whl selection algorithm

aignas:
DO NOT MERGE: stacked on #3110

This PR only implements the selection algorithm where instead of
selecting all wheels that are compatible with the set of target
platforms, we select a single wheel that is most specialized for a
particular *single* target platform.

What is more, compared to the existing algorithm it does not assume
a particular list of supported platforms and just fully implements the
spec.

Work towards #2747
Work towards #2759
Work towards #2849

### DIFF
--- a/python/private/pypi/select_whl.bzl
+++ b/python/private/pypi/select_whl.bzl
@@ -1,0 +1,230 @@
+"Select a single wheel that fits the parameters of a target platform."
+
+load("//python/private:version.bzl", "version")
+load(":parse_whl_name.bzl", "parse_whl_name")
+load(":python_tag.bzl", "PY_TAG_GENERIC", "python_tag")
+
+_ANDROID = "android"
+_IOS = "ios"
+_MANYLINUX = "manylinux"
+_MACOSX = "macosx"
+_MUSLLINUX = "musllinux"
+
+def _value_priority(*, tag, values):
+    keys = []
+    for priority, wp in enumerate(values):
+        if tag == wp:
+            keys.append(priority)
+
+    return max(keys) if keys else None
+
+def _platform_tag_priority(*, tag, values):
+    # Implements matching platform tag
+    # https://packaging.python.org/en/latest/specifications/platform-compatibility-tags/
+
+    if not (
+        tag.startswith(_ANDROID) or
+        tag.startswith(_IOS) or
+        tag.startswith(_MACOSX) or
+        tag.startswith(_MANYLINUX) or
+        tag.startswith(_MUSLLINUX)
+    ):
+        res = _value_priority(tag = tag, values = values)
+        if res == None:
+            return res
+
+        return (res, (0, 0))
+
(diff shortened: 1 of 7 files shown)


rickeylev:
didn't have time to review, but qq: does it still handle if multiple wheels are compatible?

The type of case I have in mind are things that aren't represented by env markers. e.g. cuda version, free threading, etc.

aignas:
The main idea is that the user can specify the preference for the abi tags and the platform tags. If cuda version or the abi can be expressed by them, then it will work as expected.

On top of this I could implement freethreaded support using this algorithm in #3063, so I am pretty confident that we have the right abstractions in place now, because the configuration of which wheel we should prefer is injected into the extension through the `defaults` API.

rickeylev:
Yeah, I get they can use bzlmod-phase stuff to control which implementations of a library are considered. Big +1 to that.

The case I want to work is multiple implementations in a single build invocation. e.g. `bazel test //foo:foo_test_cuda12 //foo:foo_test_cuda13`. "Multiple implementations" may mean multiple numpy versions, or multiple micro-arch versions, different libcs, etc etc.

I'm guessing this PR does do that (since (1) that's what've talked of doing previous and (2) your mention of free-threading and the 3063 description makes it sound in that same vein) (but still haven't had a chance to review the code).

But this PR description implies that, after bzlmod phase, only _one_ wheel will be chosen?

aignas:
One wheel per target platform. So whatever your set of config settings you want to target are, there should be either 1 or 0 wheels matching that.

python/private/pypi/select_whl.bzl
```diff
@@ -0,0 +1,230 @@
...
+    else:
+        minor = "0"
+        arch = tail
+    version = (int(major), int(minor))
+
+    keys = []
+    for priority, wp in enumerate(values):
+        want_plat, sep, tail = wp.partition("_")
+        if not sep:
+            continue
```
rickeylev, line 56: Why skip? Does this mean it's malformed? Please add comment.

python/private/pypi/select_whl.bzl
```diff
@@ -0,0 +1,230 @@
...
+    ret = {}
+    for whl in whls:
+        parsed = parse_whl_name(whl.filename)
+        priority = None
+
+        # See https://packaging.python.org/en/latest/specifications/platform-compatibility-tags/#compressed-tag-sets
+        for platform in parsed.platform_tag.split("."):
+            platform = _platform_tag_priority(tag = platform, values = whl_platform_tags)
+            if platform == None:
+                if logger:
```
rickeylev: I'm guessing logger might be None for tests? Just pass in a no-op logger instead. It helps avoid NPE type of errors; those bit us in the past with the conditional loggers.

Oh, I see the tests are passing in loggers? Are these checks needed, then?

PR #3114, opened 2025-07-21
fix: repl tab completion

jpwoodbu:
PR Instructions/requirements
  section at the bottom.
* Body text describes:


aignas:
I think #3104 needs to merged into this so that the `os.path.norm` is used instead of what we have now in this PR.

jpwoodbu:
> I think #3104 needs to merged into this so that the `os.path.norm` is used instead of what we have now in this PR.

I believe I've gotten that merge done. I'm surprised by the CI failures. They make it look like the merge wasn't successful, but the content of the commit the tests used looks correct to me.

jpwoodbu:
I'm seeing more clear error messages now. It seems the `readline` module is not available on Windows. As best I can tell, there's maybe nothing in the stdlib that can stand in for readline. I'm think of importing it conditionally.

python/bin/repl_stub.py
```diff
@@ -28,5 +52,19 @@
...
+
+    # TODO(jpwoodbu): Use readline.backend instead of readline.__doc__ once we can depend on having
+    # Python >=3.13.
+    if "libedit" in readline.__doc__:  # type: ignore
+        readline.parse_and_bind("bind ^I rl_complete")
+    elif "GNU readline" in readline.__doc__:  # type: ignore
+        readline.parse_and_bind("tab: complete")
+    else:
+        print("Could not enable tab completion!")
+
```
aignas: Could you print `Could not enable tab completion` on windows as well?

Maybe structuring the code like:
```
try
    import readline
    import rlcompleter

    class DynamicCompleter
        ...

    completer = DynamicCompleter(console_locals)
    readline.set_completer(completer.complete)

    if ...
    else:
        print(...)
except ImportError:
    print(...)
```

jpwoodbu: Done. It's a lot of code inside a try block. Do you have any concerns about maintainability?

aignas: Since it fits into a single screen of text on my laptop, I think it is not too much, but maybe @philsc, could chime in as well? I have no strong opinions right now, but I do think that the code could be changed (maybe the invocation of the completer could be done in a function?

What I like about the current way of writing is that it is very explicit of when the code is executed and I would love to preserve that.

Let me know if you have any ideas.

Maybe something like:

```python
import rlcomplete
try:
    import readline
except ImportError
    readline = None

class DynamicCompleter:
    ...

def setup_completion():
    completer = DynamicCompleter(console_locals)
    readline.set_completer(completer.complete)
    ...
    return True if success else False

if not (readline and setup_completion()):
    print("Could not setup completion")
code.interact()
```

jpwoodbu: I did something similar to what you suggested. I tried to minimize the amount of code within the `readline` import `try` block while leaning on some of the existing control flow to improve the error messages a bit.

Hopefully this preserves the clarity and encapsulation you cited earlier. And the code that's in the `try` block is also not very likely to raise any other exceptions since it's all in-memory configuration.

aignas (commented):
LGTM, could you please add a CHANGELOG.md entry for this? This will be useful for the users to know!

jpwoodbu:
> LGTM, could you please add a CHANGELOG.md entry for this? This will be useful for the users to know!

Thanks! Done.

CHANGELOG.md
```diff
@@ -104,6 +104,8 @@ END_UNRELEASED_TEMPLATE
 
 {#v0-0-0-added}
 ### Added
+* (repl) Default stub now has tab completion where `readline` support is available, see
```
```suggestion
* (repl) Default stub now has tab completion, where `readline` support is available,
  see ([#3114](https://github.com/bazel-contrib/rules_python/pull/3114)). 
```

aignas (approved):
LGTM

PR #3124, opened 2025-07-28
test: Print REPL error message during test failures

philsc:
I noticed that the CI error messages for #3114 are not useful. This
patch aims to help with that by printing the output of the REPL code.
If there's an exception during startup for example, then the test log
will now contain the stack trace.


philsc:
There's a CI failure that I don't know what to do about:
> OSError: [Errno 8] Exec format error: '/tmp/tmp42b4f4hx/bazelci-agent'

https://buildkite.com/bazel/rules-python-python/builds/12315#01985329-ab67-45ed-badf-63ab1af09f97

PR #3128, opened 2025-08-01
fix(gazelle): Rename experimental_allow_relative_imports directive to follow convention

dougthor42:
Prefix `experimental_allow_relative_imports` with `python_` to match the rest of the directives.

1.6.0 hasn't been released yet, so this is a non-breaking change.


dougthor42:
FYI @yushan26, @linzhp in case you are using pre-release code.

PR #3132, opened 2025-08-02
docs(gazelle): Migrate Gazelle docs to ReadTheDocs, part 2/5: installation and usage

dougthor42:
Part of #3082

2nd of probably 5 PRs.

+ Migrate installation and usage info from `gazelle/README.md` to `gazelle/docs/installation_and_usage.md`
+ Slight rewording and reformatting of the `Example` section.
+ Mechanical updates:
  + Wrap at ~80 chars
  + Use MyST directives and roles.


gazelle/docs/installation_and_usage.md
```diff
@@ -0,0 +1,156 @@
...
+
+First, you'll need to add Gazelle to your `MODULE.bazel` file. Get the current
+version of Gazelle from their [releases page][gazelle-releases].
+
+[gazelle-releases]: https://github.com/bazel-contrib/bazel-gazelle/releases/
+
+See the installation `MODULE.bazel` snippet on the `rules_python`
+[releases page][rules-python-releases] in order to configure `rules_python`.
+
+[rules-python-releases]: https://github.com/bazel-contrib/rules_python/releases
```
aignas: I would like to point users to the BCR: https://registry.bazel.build/modules/rules_python_gazelle_plugin

If the instructions are not up-to-date there (but I think they are), then maybe we can clean that up as well?

EDIT: it's just a move, it is fine to leave this as is and make changes in the upcoming PRs.

gazelle/docs/installation_and_usage.md
```diff
@@ -0,0 +1,156 @@
...
+```starlark
+# The following stanza defines the dependency rules_python.
+bazel_dep(name = "rules_python", version = "0.22.0")
+
+# The following stanza defines the dependency rules_python_gazelle_plugin.
+# For typical setups you set the version.
+bazel_dep(name = "rules_python_gazelle_plugin", version = "0.22.0")
+
+# The following stanza defines the dependency gazelle.
+bazel_dep(name = "gazelle", version = "0.31.0", repo_name = "bazel_gazelle")
```
aignas: These are very old, is it on purpose?

EDIT: this is just a move, it's fine to leave it as it is.

gazelle/docs/installation_and_usage.md
```diff
@@ -0,0 +1,156 @@
...
+
+# Imports the pip toolchain generated by the given module extension into the
+# scope of the current module.
+use_repo(pip, "pip")
+```
+
+Next, we'll fetch metadata about your Python dependencies, so that gazelle can
+determine which package a given import statement comes from. This is provided
+by the {bzl:obj}`modules_mapping` rule. We'll make a target for consuming this
+{bzl:obj}`modules_mapping`, and writing it as a manifest file for Gazelle to read.
```
aignas: It seems that the links here don't work - we cannot get to the starlark docs to here in the https://rules-python--3132.org.readthedocs.build/en/3132/gazelle/docs/installation_and_usage.html

Do you have a plan how to make it work in the future?

dougthor42: Sadly I don't have a plan. I have only scratched the surface of our sphinx, MyST, and sphinxdocs/stardoc stuff so I don't currently know how the cross references are actually made.

I removed the broken crossrefs.

gazelle/docs/installation_and_usage.md
dougthor42: Yeah it's just a move, but TBH I doubt I'll go back through these later and update them so I should make sure they're up to date now...

Updated

gazelle/docs/installation_and_usage.md
dougthor42: Same as other comment - I doubt I'll come back to this anytime soon so it's better to update now.

Updated:

+ Reorganized the MODULE.bazel example
  + separated out the gazelle plugin stuff from the core rules_python stuff
  + removed comments from the core rules_python stuff - tell the user to refer to those docs directly instead
  + modernized a smidge
+ Minor changes to the BUILD.bazel example:
  + vertical whitespace for readability
  + Default to supporting multiple languages now that #3057 is merged

gazelle/docs/installation_and_usage.md
aignas: The way to do it would be to define the `bzl_library` targets in the `gazelle` workspace and then include them into the `//docs` target. That should fix the crossreffs.

dougthor42: POC in https://github.com/bazel-contrib/rules_python/pull/3140; will continue discussion there.

PR #3140, opened 2025-08-05
docs(gazelle): POC the bare minimum for linking modules_mapping in docs.

dougthor42:
From https://github.com/bazel-contrib/rules_python/pull/3132#discussion_r2250671361.

While this note exists, the PR description is WIP.

### DIFF
--- a/docs/BUILD.bazel
+++ b/docs/BUILD.bazel
@@ -82,9 +82,15 @@ build_test(
     targets = [":docs"],
 )
 
+alias(
+    name = "gazelle_modules_mapping_def_bzl",
+    actual = "@rules_python_gazelle_plugin//modules_mapping:def_bzl",
+)
+
 sphinx_stardocs(
     name = "bzl_api_docs",
     srcs = [
+        ":gazelle_modules_mapping_def_bzl",
         "//python:defs_bzl",
         "//python:features_bzl",
         "//python:packaging_bzl",
--- a/gazelle/docs/installation_and_usage.md
+++ b/gazelle/docs/installation_and_usage.md
@@ -61,8 +61,8 @@ bazel_dep(name = "rules_python_gazelle_plugin", version = "1.5.1")
 
 Next, we'll fetch metadata about your Python dependencies, so that gazelle can
 determine which package a given import statement comes from. This is provided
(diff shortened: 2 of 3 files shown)


docs/BUILD.bazel
```diff
@@ -82,9 +82,15 @@ build_test(
...
 
+alias(
+    name = "gazelle_modules_mapping_def_bzl",
+    actual = "@rules_python_gazelle_plugin//modules_mapping:def_bzl",
+)
+
 sphinx_stardocs(
     name = "bzl_api_docs",
     srcs = [
+        ":gazelle_modules_mapping_def_bzl",
```
dougthor42, line 93: sphinx_stardocs doesn't like `@rules_python_gazelle_plugin//modules_mapping:def_bzl` with:

```
Error in _stardoc_input_helper:
illegal rule name:
_bzl_api_docs_@rules_python_gazelle_plugin//modules_mapping/def_bzl.primary_bzl_src:
invalid target name '_bzl_api_docs_@rules_python_gazelle_plugin//modules_mapping/def_bzl.primary_bzl_src':
target names may not contain '//' path separators
```

So I tried the alias thing and it worked :shrug:. Having an alias for every `bzl_library` target is not feasible of course. Maybe a filegroup or a meta-target `bzl_library(srcs = ["$ALL_BZL_LIBRARY_TARGETS"])`...

With the alias, we are able to link with `{bzl:obj}`. It goes to: http://localhost:8000/api/rules_python_gazelle_plugin%2B/modules_mapping/def.html#modules_mapping where `%2B` is `+`:

```
rules_python_gazelle_plugin+/modules_mapping/def.html
```

aignas: I think some of the automatic inference withing `sphinxdocs` needs to be updated, but I am not sure where, I'd need to dig, maybe @rickeylev knows of his hat.

rickeylev: The error is due to a macro generating a target with the raw label value; it does this to create some helper targets to extract some info.

sphinxdocs/private/sphinx_stardoc.bzl, function `_name_to_label` -- just needs to also replace `@`, too.

rickeylev (approved):

Issue #824, opened 2022-09-06
Add support for building C extensions


blais:
# 🚀 feature request

* [x] Add support for getting Python headers (Python.h) -- use `@rules_python//python/cc:current_py_cc_headers`
* [x] Provide shared libraries (libpython.so et al) -- use `@rules_python//python/cc:current_py_cc_libs`
* [ ] Provide static libraries (libpython.a)
* [ ] Add a rule/rules to build C extensions
* [ ] Add `PyInfo.cc_info` field to allow C/C++ info (CcInfo) to propagate (directly
  propagating CcInfo it isn't done to prevent a Python target from
  being accidentally accepted by a plain CC target)

### Describe alternatives you've considered

I've cobbled support for these things by hand in the past, was hoping building extension modules would be officially supported at some point.

Some projects, such as tensorflow and pybind11_bazel, have a python_configure.bzl file and/or py_extension rule that attempts to implement this.

(Note: this issue was originally about updating the readme to address the lack of this feature; it's been edited to be a feature request for this feature directly - rickeylev)

rickeylev:
> Can you clarify in the README how this module relates to building extension modules?

Which readme are you referring to, this repo's readme? This repo, pybind11, and tensorflow have different owners. There's some owner-adjacency with this repo and pybind11 (Me and Ralf are on the same team within Google, but don't overlap in work much really), but they're different projects. I think grpc also has a custom solution that mostly-works-except-when-it-doesn't.

Pybind11 is a higher-level way to write C extensions.

Tesorflow's purpose is an ML library -- if it provides a py_extension, then it's almost certainly just incidental. I've talked to various TF folk and they, too, want an official py_extension rule they can use.

I'm pretty certain all of these solutions are incomplete and/or unsound in some way -- I don't think the toolchain provides enough information to implement them entirely correctly.

> hoping building extension modules would be officially supported at some point.

Yes, that's the plan. There aren't really any details beyond that, though. While there is a `py_extension` rule within Google, it's not compatible with Bazel as-is and has some historical baggage that make it unsavory to release as-is.

I'm happy to discuss design and implementation ideas about it, though.

blais:
> > Can you clarify in the README how this module relates to building extension modules?
> 
> Which readme are you referring to, this repo's readme? This repo, pybind11, and tensorflow have different owners. There's some owner-adjacency with this repo and pybind11 (Me and Ralf are on the same team within Google, but don't overlap in work much really), but they're different projects. [...]
[...]

PR #3148, opened 2025-08-06
fix(local_runtime): Improve local_runtime usability in macos / windows

laramiel:
local_runtime fails to handle many variations of python install on Windows and
MacOS, such as:

* LDLIBRARY on MacOS may refer to a file under PYTHONFRAMEWORKPREFIX, not LIBDIR
* LDLIBRARY on Windows refers to pythonXY.dll, not the linkable pythonXY.lib
* LIBDIR may not be correctly set on Windows.
* On windows, interpreter_path needs to be normalized.  Other paths also require this.
* SHLIB_SUFFIX does not indicate the correct suffix.

For examples, see: https://docs.python.org/3/extending/windows.html

In order to resolve this the shared library resolution has been moved into get_local_runtime_info.py, which now does the following:

* Constructs a list of paths to search based on LIBDIR, LIBPL, PYTHONFRAMEWORKPREFIX,
  and the executable directory.
* Constructs a list of libraries to search based on  INSTSONAME, LDLIBRARY,
  pythonXY.lib, etc.
* Checks to see which files exist, partitioning the result into a list of
  "dynamic_libraries" and "static_libraries" 

On Windows and macOS, since SHLIB_SUFFIX does not always indicate the filenames
needed searching, this has been removed from local_runtime_repo_setup and
replaced with an explicit file.

On Windows the interpreter_path and other search paths are now normalized
(`\` converted to `/`).

Additional logging added to local_runtime_repo.

Fixes https://github.com/bazel-contrib/rules_python/issues/3055
Work towards https://github.com/bazel-contrib/rules_python/issues/824


python/private/local_runtime_repo.bzl
```diff
@@ -120,45 +166,44 @@ def _local_runtime_repo_impl(rctx):
...
         micro = info["micro"],
-        interpreter_path = interpreter_path,
-        lib_ext = info["SHLIB_SUFFIX"],
+        interpreter_path = _norm_path(interpreter_path),
+        library_srcs = repr(library_srcs),
         implementation_name = info["implementation_name"],
         os = "@platforms//os:{}".format(repo_utils.get_platforms_os_name(rctx)),
-    ))
+    )
+    logger.debug("BUILD.bazel\n{}".format(build_bazel))
```
aignas: Do you need to have it? You can always do `bazel build <repo>//:BUILD.bazel` to see the file. I think this can be removed and the changes reordering the writing of the files could be also reverted for a smaller diff.

laramiel:
After integrating this some more with tensorstore, I've reworked the library lookup to handle manylinux and other linux installs with no dynamic libraries exposed. 

-----

rickeylev edit to preserve some initial PR info:

Example get_local_runtime_info.py outputs:

```
rules_python:local_runtime_repo(@@local_python3) INFO: GetPythonInfo result:
  INSTSONAME: Python.framework/Versions/3.12/Python
  LDLIBRARY: Python.framework/Versions/3.12/Python
  LIBDIR: /Library/Frameworks/Python.framework/Versions/3.12/lib
  MULTIARCH: darwin
  PY3LIBRARY: <empty string>
  PYTHONFRAMEWORKPREFIX: /Library/Frameworks
  base_executable: /Library/Frameworks/Python.framework/Versions/3.12/bin/python3.12
  implementation_name: cpython
  include: /Library/Frameworks/Python.framework/Versions/3.12/include/python3.12
  major: 3
  micro: 10
  minor: 12
```

Here we see the MacOS framework issue, namely LDLIBRARY is not under LIBDIR, but directly under frameworks.
The new code adds PYTHONFRAMEWORKPREFIX to the search path, so the library is found.

```
rules_python:local_runtime_repo(@@local_python3)  INFO: GetPythonInfo result:
  INSTSONAME: None
  LDLIBRARY: python313.dll
  LIBDIR: T:\build_temp\home\python_4379bdec28bdff81a567a01c9b8cf10e3856c8c966e4fe53945bedea6338b416\tools\libs
  MULTIARCH: None
  PY3LIBRARY: python313.lib
  base_executable: T:\build_temp\home\python_4379bdec28bdff81a567a01c9b8cf10e3856c8c966e4fe53945bedea6338b416\tools\python.exe
  implementation_name: cpython
  include: T:\build_temp\home\python_4379bdec28bdff81a567a01c9b8cf10e3856c8c966e4fe53945bedea6338b416\tools\Include
  major: 3
  micro: 1
  minor: 13
```

Here we see that under windows the python313.dll is reported; python313.lib is added via PY3LIBRARY.


Three more notes on this CL; they may be areas for future improvements.

* On MacOS, when linking against the system python, we may want to use `"-Wl,-framework,Python"`. However this CL does not detect / change those linker options.

* It would be nice to have tests for a minimal C python extension and minimal C python embedded use for the local_runtime as well as the hermetic runtimes; this is similar to tests/cc/current_cc_py_libs

python/private/get_local_runtime_info.py
```diff
@@ -13,9 +13,158 @@
...
+      else:
+        dynamic_libraries[composed_path] = None
+
+  # NOTE: It's possible that including the dynamic libraries currently loaded
+  # by the running python interpreter would be a useful addition.
+
+  return {
+      "dynamic_libraries": list(dynamic_libraries.keys()),
+      "static_libraries": list(static_libraries.keys()),
+  }
```
aignas: There is a mix of 2 and 4 spaces here. Could you please ensure that you have run `pre-commit run black` on this file? You could also install the pre-commit hooks with `pre-commit install`.

CHANGELOG.md
```diff
@@ -101,6 +101,7 @@ END_UNRELEASED_TEMPLATE
   `# gazelle:python_resolve_sibling_imports true`
 * (pypi) Show overridden index URL of packages when downloading metadata have failed.
   ([#2985](https://github.com/bazel-contrib/rules_python/issues/2985)).
+* (toolchains) `local_runtime_repo` better handles variants in MacOS and Windows.
```
aignas: You could also describe that `linux` got improved as well.

aignas (commented):
Consider also referencing the issue in the changelog.

This PR is very exciting, but I gave it a quick glance only since it is reasonably big.

I would love to ask the `local_python` example also be made to work with `bzlmod` - all of the improvements should work there too.

python/private/local_runtime_repo.bzl
laramiel: I can do that, however this is logged at debug level and given my multi-target ci it's helpful to print it in the log.

CHANGELOG.md
rickeylev: I expanded the changelog

python/private/get_local_runtime_info.py
rickeylev: Looks like this was done; resolving

examples/local_python/WORKSPACE
rickeylev, line 1: The integration tests for local toolchains are in tests/integration/local_toolchains. It should already have workspace and bzlmod stuff setup. I think the only thing to do is remove the target_compatible_with constraint that makes it only run on linux/mac.

python/private/get_local_runtime_info.py
```diff
@@ -13,46 +13,163 @@
...
+
+
+def _search_library_names(get_config):
+    """Returns a list of library files to search for shared libraries."""
+    # Quoting configure.ac in the cpython code base:
+    # "INSTSONAME is the name of the shared library that will be use to install
+    # on the system - some systems like version suffix, others don't.""
+    #
+    # A typical INSTSONAME is 'libpython3.8.so.1.0' on Linux, or
+    # 'Python.framework/Versions/3.9/Python' on MacOS.
```
rickeylev, line 82: Just verifying: On mac, INSTSONAME might be a relative path to a dir, not the filename?

python/private/get_local_runtime_info.py
```diff
@@ -13,46 +13,163 @@
...
+            "LDLIBRARY",
+            "PY3LIBRARY",
+            "LIBRARY",
+            "DLLLIBRARY",
+        )
+    ]
+
+    # Set the prefix and suffix to construct the library name used for linking.
+    if _IS_DARWIN:
+        # SHLIB_SUFFIX may be ".so"; always override on darwin to be ".dynlib"
```
rickeylev: Why always override? Because dylib is the idiomatic extension? Or for another reason? Whatever the case, please include it as a comment.

python/private/get_local_runtime_info.py
```diff
@@ -13,46 +13,163 @@
...
+        suffix = ".dylib"
+        prefix = "lib"
+    elif _IS_WINDOWS:
+        # SHLIB_SUFFIX on windows is ".dll"; however the compiler needs to
+        # link with the ".lib".
```
rickeylev: Do you have a reference for this? If so, please include a link. If not, oh well.

python/private/get_local_runtime_info.py
```diff
@@ -13,46 +13,163 @@
...
+    lib_names = [v for v in lib_names if v]
+    return {k: None for k in lib_names}.keys()
+
+
+def _get_python_library_info():
+    """Returns a dictionary with the static and dynamic python libraries."""
+    config_vars = sysconfig.get_config_vars()
+
+    # VERSION is X.Y in Linux/macOS and XY in Windows.
+    if not config_vars.get("VERSION"):
```
rickeylev, line 135: In what case is VERSION missing? I'd think this would always be present. I known, please comment. If this is just defense, that's fine too; just comment its for defense.

laramiel:
> Oddly, I can edit and commit files via the ui, but can't push to laramiel/main on the command line. Is there something special setup with your repo or branch? I was going to move the example over into local toolchains, but it was a bit hard to refactor that via the ui alone.

I wouldn't expect so; I forked using the web ui and haven't changed anything.

python/private/get_local_runtime_info.py
laramiel: Yes. The old get_local_runtime_info returned this on some mac python installs:

     {
        "major": 3,
        "minor": 12,
        "micro": 10,
        "implementation_name": "cpython",
        "base_executable": "/Library/Frameworks/Python.framework/Versions/3.12/bin/python3.12",
        "include": "/Library/Frameworks/Python.framework/Versions/3.12/include/python3.12",
        "LDLIBRARY": "Python.framework/Versions/3.12/Python",
        "LIBDIR": "/Library/Frameworks/Python.framework/Versions/3.12/lib",
        "INSTSONAME": "Python.framework/Versions/3.12/Python",
        "PY3LIBRARY": "",
        "SHLIB_SUFFIX": ".so",
     }
     
     Both LDLIBRARY / INSTSONAME are the same and are relative to the frameworks directory.

python/private/get_local_runtime_info.py
laramiel: Reworked comment & how this works. Added reference to windows in the loop which looks for .lib files.

See, for example: https://docs.python.org/3/extending/windows.html

python/private/get_local_runtime_info.py
laramiel: I believe that I have seen it missing, but I don't have logs for when that happens.

python/private/local_runtime_repo_setup.bzl
```diff
@@ -81,13 +84,15 @@ def define_local_runtime_toolchain_impl(
...
         ),
+        deps = import_deps,
         includes = ["include"],
     )
 
     cc_library(
         name = "_libpython",
         hdrs = [":_python_headers"],
-        deps = _libpython_deps,
+        srcs = [shared_library] if shared_library else [],
```
rickeylev: FWIW, I think srcs can be a mix of static and shared libraries. cc_library will sort them out.

rickeylev (commented):
This is mostly LGTM, just need to remove the example as shown and update tests/integration/local_toolchains instead.

rickeylev:
> > Oddly, I can edit and commit files via the ui, but can't push to laramiel/main on the command line. Is there something special setup with your repo or branch? I was going to move the example over into local toolchains, but it was a bit hard to refactor that via the ui alone.
> 
> I wouldn't expect so; I forked using the web ui and haven't changed anything.

Weird :shrug: . Maybe because your fork's branch is the main branch? sometimes those have additional protections.

examples/local_python/WORKSPACE
laramiel: I don't see which `target_compatible_with` constraint affects tests/integration  Can you drop a reference?

rickeylev: I was mistaken. It looks like we only run tests/integration on Linux, not Mac or Windows. And it only uses bzlmod.

I've setup workspace in that directory and deleted these files.

Enabling it in CI might be tricky because I'm not sure we have enough CI slots left. I've added the jobs configs; lets see how it goes.

rickeylev:
Looks like we have enough slots to run the CI jobs, but they had a lot of unrelated failures. On the plus side, I think they exercised enough of the new local toolchain logic to give some basic validation of the new code.

I don't think we should block this PR on making all of those integration tests pass. I'll reconfig the mac and windows ones to just run the local toolchain test; that looks relatively easy and feasible.

re: adding a py extension test: +1 on such a test (we have need of that elsewhere for testing), but I'm also OK with omitting it for this PR. Test-supporting code goes in `tests/support`

jbms:
I guess `py_extension` is being added for a test now but might later be made into a public rule?

For fvisibility=hidden, really that needs to propagate to all transitive dependencies instead, so should be handled via a transition or just by a global `--copt` option specified by the user.

laramiel:
> I guess `py_extension` is being added for a test now but might later be made into a public rule?

It's added for test purposes, I'll leave it up to @rickeylev or others to make that determination.
It would be nice to have a common py_extension, IMO.

> For fvisibility=hidden, really that needs to propagate to all transitive dependencies instead, so should be handled via a transition or just by a global `--copt` option specified by the user.

For now I've removed the cc_library from py_extension. That seems a little cleaner?

rickeylev:
> I guess `py_extension` is being added for a test now but might later be made into a public rule?

Maybe, but not without some more consideration and evaluation. (e.g. My current day-job team developed their own variation of py_extension, my last team created another one, and I've seen others out there, all with pretty different implementations). #824 is the FR for this (long with lots of details and cases). Your mention of how deps are handled and fvisibility are good examples of the sort of considerations to not forget. I don't want perfect to be the enemy of the good, but also want thoughtfulness put into a public feature. (All that plus I want to keep the PR scope reasonable)

rickeylev:
Woot, CI finally happy!

PR #3156, opened 2025-08-09
feat: Upgrade latest python build versions to 20250808

engnatha:
astral-sh has released [20250808](https://github.com/astral-sh/python-build-standalone/releases/tag/20250808) for python standalone builds. This means we have access to the fix in https://github.com/astral-sh/python-build-standalone/pull/676 which was created to overcome the shortcomings of interactive plotting highlighted by https://github.com/astral-sh/python-build-standalone/issues/129#issuecomment-3164285100.

All latest patch releases supported by this build have been updated in `versions.bzl`. Additionally, `3.14.0rc1` is added opportunistically for those who may be trying out that toolchain.


aignas:
#3116 is already there that performs this update.

PR #3159, opened 2025-08-11
refactor(pypi): use a builder pattern for parsing requirements

aignas:
Whilst trying to finish the configuration API I bumped into the issue
that the API of the `parse_requirements` function does not lend itself
to building the configuration easily.

With this we will be able to build the configuration more easily and
have more customization without passing ever more complex structures
between the functions. What is more, I believe that it may be possible
to have more complex composable builders in the future to build the
`whl_library` args with a single `build` command instead of what we have
now.

Work towards #2747


aignas:
Need to do it again, the `main` diverged too much from the base of this.

PR #3175, opened 2025-08-13
fix: bootstrapping script to not use multiline f-strings

jacky8hyf:
With bootstrap_impl=system_python, after
https://github.com/bazel-contrib/rules_python/pull/2607, host Python 3.6 and above is required because of the use of multiline f-strings. For maximum backwards compatibility (because this is a host dependency), do not use multiline f-strings.

This issue naturally does not apply to bootstrap_impl=script because there are no host Python dependency.

Link: https://github.com/bazel-contrib/rules_python/pull/2607


rickeylev (approved):
Thanks for the PR!

This is a simple change, so LGTM.

Note that there's no guarantees future changes won't introduce incompatibility with old python versions. We won't actively guard against that, but PRs to fix it will likely accepted.

PR #3178, opened 2025-08-14
fix(local_runtime): Search for libs in sys._base_executable when available.

laramiel:
Search directory for libraries should look in the same directory as sys._base_executable. Since sys._base_executable may be unset, fallback to sys.executable

Found this when trying to build using a venv for [tensorstore](https://github.com/google/tensorstore) on Windows:
* Github CI uses nuget to download Python.
* Build sets up a Python venv.

The venv does not include all the lib directories required to link an extension.

Fixes https://github.com/bazel-contrib/rules_python/issues/3172

### DIFF
--- a/python/private/get_local_runtime_info.py
+++ b/python/private/get_local_runtime_info.py
@@ -23,7 +23,7 @@ _IS_WINDOWS = sys.platform == "win32"
 _IS_DARWIN = sys.platform == "darwin"
 
 
-def _search_directories(get_config):
+def _search_directories(get_config, base_executable):
     """Returns a list of library directories to search for shared libraries."""
     # There's several types of libraries with different names and a plethora
     # of settings, and many different config variables to check:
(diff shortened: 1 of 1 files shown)


rickeylev:
> base_executable is missing when building within a venv

But `sys._base_executable` is suppose to point where the venv's interpreter is, that's its purpose.

I wonder if this problem manifests when the underlying interpreter can't be found, but the venv is otherwise functional? Maybe if the venv has a copy, not symlink to the interpreter? IIRC, windows venvs create copies of python.exe and rely on pyvenv.cfg to point to the correct pythonhome location. Hm, maybe hardlinks on linux/mac result in something similar? ISTR macs also have some funny tricks they pull with the python executable.

In anycase, LGTM

laramiel:
I thought that it was trying to be the realpath of sys.executable, usually:

https://github.com/python/cpython/blob/main/Modules/getpath.py

But in any event, we may create nested venvs, and sys._base_executable is used to create those nested venvs, so I think that's the appropriate path to use.  Also the windows copy thing.

PR #3186, opened 2025-08-17
fix(pypi): correctly handle different package versions

### LINKED ISSUE #2797: Support universal requirements for multiple python versions
# 🐞 bug report

### Affected Rule

`pip.parse`

### Is this a regression?

I don't think so

### Description

Some dependencies can vary based on the current python version. In this case for requirements.txt files generated by uv or similar tools, the python version is encoded in the requirements.txt:

```
numpy==2.0.2 ; python_full_version < '3.10'
    # via -r requirements.in
numpy==2.2.5 ; python_full_version >= '3.10'
    # via -r requirements.in
```

In this case if you attempt to use this requirements file with pip.parse you get this error:

```
ERROR: /home/ubuntu/.cache/bazel/_bazel_ubuntu/46f0c9813505c30d178a2aff399354ea/external/rules_python+/python/private/pypi/extension.bzl:224:25: Traceback (most recent call last):
        File "/home/ubuntu/.cache/bazel/_bazel_ubuntu/46f0c9813505c30d178a2aff399354ea/external/rules_python+/python/private/pypi/extension.bzl", line 607, column 25, in _pip_impl
                mods = parse_modules(module_ctx)
        File "/home/ubuntu/.cache/bazel/_bazel_ubuntu/46f0c9813505c30d178a2aff399354ea/external/rules_python+/python/private/pypi/extension.bzl", line 478, column 36, in parse_modules
                out = _create_whl_repos(
        File "/home/ubuntu/.cache/bazel/_bazel_ubuntu/46f0c9813505c30d178a2aff399354ea/external/rules_python+/python/private/pypi/extension.bzl", line 224, column 25, in _create_whl_repos
                fail("Attempting to creating a duplicate library {} for {}".format(
Error in fail: Attempting to creating a duplicate library pip_311_numpy_linux_aarch64_linux_arm_linux_ppc_linux_s390x_linux_x86_64_osx_aarch64_osx_x86_64_windows_x86_64 for numpy
```

Ideally rules_python would know that there should only be a single repo here since these constraints are mutually exclusive

## 🔬 Minimal Reproduction

`bazel build ...` in https://github.com/bazel-contrib/rules_python/pull/2798

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
linux x86_64
  </code>
</pre>

<pre>
  <code>
8.2.1
  </code>
</pre>

**Rules_python version:**

https://github.com/bazel-contrib/rules_python/commit/c981569cc89c76eb57a78f0bbc47f1566211c924

aignas:
After merging #3058, I wanted to try creating a `universal` requirements
file for our `sphinx` setup and have an integration test in that way. It
seems that `alabaster` has different versions for different python
versions and we were not handling this correctly. In #3058 I have
attempted adding a unit test for this but it escaped me that the only
time where this bug manifests itself is when the `parse_requirements` is
used multiple times, once per each `python_version`.

With this I think it is safe to say that the #2797 is fixed, because we
found a bug, where we were not skipping requirements.

As an added counter measure, I have added an extra check elsewhere to
catch a regression.

Fixes #2797


docs/BUILD.bazel
```diff
@@ -186,5 +186,6 @@ lock(
         "--universal",
         "--upgrade",
     ],
+    python_version = "3.9",  # select the lowest version
```
rickeylev: This is just for verification purposes, right? Please update the comment to better reflect that. In practice, we can use whatever version of python we want for our doc build, and a newer would be better.

rickeylev (approved):
Out of scope for this PR, I think, but the `whl_from_dir` was so handy for testing I can't help but bring it up:

How could we test this using a checked-in requirements.txt and avoid hitting the network to talk to pypi?

Maybe a special `bazel://` url that the pip integration translates into `file://` urls (so that pip itself can understand them)?

And we do like:

```
# requirements.txt
foo @ bazel://whatever

# module.bazel
whl_from_dir(name = "whatever")

pip.parse(
  requirements = "requirements.txt"
  custom_packages = {
    "whatever": "@whatever//whatever.whl"
  }
)
```

Or maybe a `pip.custom_index()` that we pass all the whl_from_dir's to, and...can we pass a file:// url to pip for its index?

docs/BUILD.bazel
aignas: This is telling uv to make a lock file that is compatible with Python 3.9 or greater. I'll add a comment here to make it more clear.

aignas:
I am thinking it would be better to do custom index URL. Example:
```
index.from_dir(
    name = "pypi_cache",
    path = "<path-to-dir>",  # similar to local_repository
)

use_repo(index, "pypi_cache")

dev_pip.parse(
    experimental_index_url = "@pypi_cache//index", # a label, but because we are doing this as a dev dep and only supporting dev deps, we can then do `ctx.path("@pypi_cache//index/pkg")`
)
```

Should we support non-dev dependencies for this on-disk PyPI cache?

rickeylev:
Non dev deps: hm. Does isolate solve it instead? That seems cleaner and easier.

Custom index url vs line-level: index level does seem a bit cleaner, I think I like that better

Issue #2762, opened 2025-04-09
Coverage now fails if there are no covered Python files even if test succeeds


phst:
# 🐞 bug report

### Affected Rule

The issue is caused by the rule: py_test

### Is this a regression?

Yes, the previous version in which this bug was not present was: 1.2.0

### Description

If there are no covered source files, the `coverage.py lcov` step fails, causing the entire test to fail when run under coverage (see repro below).

## 🔬 Minimal Reproduction

See https://github.com/phst/py-cover-bug, and the next section.

## 🔥 Exception or Error

After cloning the above repository, run `bazel coverage --test_output=all //a:test`. This fails with an error

```
[...]
INFO: Using default value for --instrumentation_filter: "^//a[/:]".
[...]
INFO: From Testing //a:test:
==================== Test output for //a:test:
Hello world
[...]/coverage/inorout.py:504: CoverageWarning: Module [...]/_main/bazel-out/darwin_arm64-fastbuild/bin/b was never imported. (module-not-imported)
  self.warn(f"Module {pkg} was never imported.", slug="module-not-imported")
[...]/coverage/control.py:894: CoverageWarning: No data was collected. (no-data-collected)
  self._warn("No data was collected.", slug="no-data-collected")
--
Coverage runner: Not collecting coverage for failed test.
The following commands failed with status 1
[...]/bin/a/test.runfiles/_main/a/test
================================================================================
[...]
```

Running this with `bazel test` instead of `coverage` succeeds. Downgrading to rules\_python 1.2.0 also makes it succeed.

Relevant additional output with verbose coverage enabled:

```
[coveragepy] Instrumented Files:
[...]/test.runfiles/_main/bazel-out/darwin_arm64-fastbuild/bin/b/bin
[...]/test.runfiles/_main/bazel-out/darwin_arm64-fastbuild/bin/b/binruntime_objects_list.txt
[coveragepy] Sources:
[...]/test.runfiles/_main/bazel-out/darwin_arm64-fastbuild/bin/b
Coverage entrypoint: [...]/coverage/__main__.py
```

Note that no directories containing Python files are listed in "Sources", so no coverage is generated for them. This is expected given the instrumentation filter, but causes `coverage.py lcov` to err out since it doesn't find any coverage data. My guess is that this is fallout from https://github.com/bazel-contrib/rules_python/pull/2607.

## 🌍  Your Environment

**Operating System:**

macOS

**Output of `bazel version`:**

8.1.1

[...]

PR #3191, opened 2025-08-18
fix(coverage): Disable certain coverage warnings.

phst:
These warnings appear if there are no Python source files in the instrumented directories,
cf. https://github.com/bazel-contrib/rules_python/issues/2762.

Work towards #2762

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,7 +62,8 @@ END_UNRELEASED_TEMPLATE
 
 {#v0-0-0-fixed}
 ### Fixed
-* Nothing fixed.
+* (tests) No more coverage warnings are being printed if there are no sources.
+  ([#2762](https://github.com/bazel-contrib/rules_python/issues/2762))
 
 {#v0-0-0-added}
 ### Added


rickeylev:
Thanks for the PR.

I think we can improve it slighty by using ctx.coverage_instrumented() and ctx.configuration.coverage_enabled. If coverage is enabled and the target isn't instrumented, then disable those warnings. Then we get the best of both worlds: if we expect coverage results, but nothing was generated, then an error (something went wrong). If we don't expect coverage data, then we can ignore it missing.

So what you'll want to do is:
* Add a template variable to the top of the files. You should see some `%blabla%` markers near the top of other things. `COVERAGE_INSTRUMENTED = "%coverage_instrumented%" == "1"`
* Conditionally add the `disable_warnings = ...` if that variable is true
* In py_executable.bzl, pass along an additional key in the template dict. grep for `%python_binary%`, you should find a couple dicts in the stage1_bootstrap and stage2_bootstrap functions.

phst:
Done, sorry for the delay

aignas (approved):
LGTM.

PR #3192, opened 2025-08-18
fix(bootstrap): handle when runfiles env vars don't point to current binary's runfiles

rickeylev:
The stage1 bootstrap script had a bug in the find_runfiles_root
function where it would unconditionally use the RUNFILES_DIR et al
environment variables if they were set.

This failed in a particular nested context: an outer binary
calling an inner binary when the inner binary isn't a data
dependency of the outer binary (i.e. the outer doesn't contain
the inner in runfiles). This would cause the inner binary to
incorrectly resolve its runfiles, leading to failures. Such
a case can occur if a genrule calls the outer binary, which has
the inner binary passed as an arg.

This change adds a check to validate that the script's entry point
exists within the inherited RUNFILES_DIR before using it. If the
entry point is not found, it proceeds with other runfiles discovery
methods. This matches the system_python runfiles discovery logic.

Fixes https://github.com/bazel-contrib/rules_python/issues/3187

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,12 @@ END_UNRELEASED_TEMPLATE
 
 [1.6.0]: https://github.com/bazel-contrib/rules_python/releases/tag/1.6.0
 
+{#v0-0-0-fixed}
+## Fixed
+* (bootstrap) The stage1 bootstrap script now correctly handles nested `RUNFILES_DIR`
+  environments, fixing issues where a `py_binary` calls another `py_binary`
+  ([#3187](https://github.com/bazel-contrib/rules_python/issues/3187)).
+
 {#1-6-0-changed}
 ### Changed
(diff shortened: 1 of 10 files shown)


alexeagle:
@rickeylev is there a plan to turn on `--@rules_python//python/config_settings:bootstrap_impl=script` as default again?

rickeylev:
Eventually, yes. IIRC, the main blocker is Bazel 8+ is required for rules_pkg to be able to process py_binary correctly. https://github.com/bazel-contrib/rules_python/issues/2521 is the tracking issue

alexeagle:
@rickeylev do you think the default could vary based on the Bazel version? Or maybe https://github.com/bazel-contrib/bazelrc-preset.bzl should set it?

aignas:
We technically can set a different default and it may be a good idea, that said, I wonder if we should wait for the packaging rules to properly support symlinks before we do the flip.

PR #3193, opened 2025-08-18
test: Add `_test_simple_multiple_platforms_with_extras`

### LINKED ISSUE #3374: `pip.parse` fails for a package with multiple extras over multiple requirements files
# 🐞 bug report

This has already been discussed in https://github.com/bazel-contrib/rules_python/issues/2797 and https://github.com/bazel-contrib/rules_python/pull/3193 but I wanted to open an issue for tracking purposes.

### Affected Rule

`pip.parse`.

### Is this a regression?

No.

### Description

Installing a package with two different extras in two different platform fails. A simplified example of requirements files that fail:
`requirements.linux_arm64.txt`:
```
package==0.7.0 \
    --hash=sha256:4dd8924f171ed73a4f1a6191e2f800ae1745069989b69fabc45593d6b6504003 \
    --hash=sha256:62833036cbaf4641d66ae94c61c0446890a91b2c0d153946583a0ebe04877a76
```
`requirements.linux_x86_64.txt`:
```
package[extra]==0.7.0 \
    --hash=sha256:62833036cbaf4641d66ae94c61c0446890a91b2c0d153946583a0ebe04877a76
```

This happens, I believe, because `pip.parse` tries to create a duplicate `whl_library` when a package is referenced multiple times with different extras.

## 🔬 Minimal Reproduction

See the `_test_multiple_extras_multiple_requirements` test case in https://github.com/bazel-contrib/rules_python/compare/main...hartikainen:rules_python:bug/multiple-platforms-with-extras-v2.

## 🔥 Exception or Error

<details><summary>Exception</summary>

```console
$ bazel test //tests/pypi/hub_builder:test_multiple_extras_multiple_requirements
ERROR: /Users/user/bazel-contrib/rules_python/main/tests/pypi/hub_builder/BUILD.bazel:3:23: in test_multiple_extras_multiple_requirements_test rule //tests/pypi/hub_builder:test_multiple_extras_multiple_requirements:
        File "/Users/user/.cache/bazel/_bazel_user/07a78f2e19ea2b267dc879919d0b00c5/external/rules_testing+/lib/private/analysis_test.bzl", line 309, column 13, in wrapped_impl
                impl(env, target)
        File "/Users/user/.cache/bazel/_bazel_user/07a78f2e19ea2b267dc879919d0b00c5/external/rules_testing+/lib/unit_test.bzl", line 43, column 40, in lambda
                impl = lambda env, target: impl(env),
        File "/Users/user/bazel-contrib/rules_python/main/tests/pypi/hub_builder/hub_builder_tests.bzl", line 1150, column 22, in _test_multiple_extras_multiple_requirements
                builder.pip_parse(
        File "/Users/user/bazel-contrib/rules_python/main/python/private/pypi/hub_builder.bzl", line 58, column 47, in lambda
                pip_parse = lambda *a, **k: _pip_parse(self, *a, **k),
        File "/Users/user/bazel-contrib/rules_python/main/python/private/pypi/hub_builder.bzl", line 125, column 22, in _pip_parse
                _create_whl_repos(
        File "/Users/user/bazel-contrib/rules_python/main/python/private/pypi/hub_builder.bzl", line 440, column 29, in _create_whl_repos
                _add_whl_library(
        File "/Users/user/bazel-contrib/rules_python/main/python/private/pypi/hub_builder.bzl", line 185, column 13, in _add_whl_library
                fail("attempting to create a duplicate library {} for {}".format(
Error in fail: attempting to create a duplicate library pypi_312_package_py3_none_any_62833036 for package
ERROR: /Users/user/bazel-contrib/rules_python/main/tests/pypi/hub_builder/BUILD.bazel:3:23: Analysis of target '//tests/pypi/hub_builder:test_multiple_extras_multiple_requirements' (config: c4919df) failed
ERROR: Analysis of target '//tests/pypi/hub_builder:test_multiple_extras_multiple_requirements' failed; build aborted
INFO: Elapsed time: 0.078s, Critical Path: 0.00s
INFO: 1 process: 1 internal.
ERROR: No test targets were found, yet testing was requested
```

</details> 

## 🌍  Your Environment

**Operating System:**

```console
$ lsb_release -a

No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 24.04.1 LTS
Release:        24.04
Codename:       noble
```

```console
$ bazel --version
bazel 8.2.1
```

**Rules_python version:**

cf594f780c91f13d48e77faad34df48ac57398da

**Anything else relevant?**

Nope.

hartikainen:
As requested in https://github.com/bazel-contrib/rules_python/issues/2797#issuecomment-3146858689 and https://github.com/aignas/rules_python/pull/2#issuecomment-3196681467. Related to #2797. Closes #3374.

For the sake of full transparency, I used [`gemini-cli`](https://github.com/google-gemini/gemini-cli) to generate most of the most recent solution.


tests/pypi/extension/extension_tests.bzl
```diff
@@ -403,6 +403,90 @@ new-package==0.0.1 --hash=sha256:deadb00f2
...
+        "pypi_312_package_py3_none_any_62833036": {
+            "dep_template": "@pypi//{name}:{target}",
+            "download_only": True,
+            "experimental_target_platforms": ["linux_arm64", "linux_x86_64"],
+            "filename": "package-0.7.0-py3-none-any.whl",
+            "python_interpreter_target": "unit_test_interpreter_target",
+            # NOTE(hartikainen): Perhaps this is part of the problem?
+            # This should say `package[extra]==0.7.0` for `linux_x86_64` platform and
+            # `package==0.7.0` for `linux_arm64`
+            "requirement": "package[extra]==0.7.0",
```
aignas: Hah, actually this is a separate issue and I am not sure how we can fix it. So, a little bit of history here:
1. poetry will export a `requirements.txt` that will have both `package[extra]` and `package` and we were asked to only consider `package[extra]` when doing the installation.
2. However with the advent of `universal` files it means that we should do something differently - first resolve the compatible lines with the target platforms and do the filtering differently.

I remember we had this discussion with in one of @ewianda  PRs, but I can't remember which one it is.

I think the issue that you are facing does not have an open issue yet.

aignas: What is more, we have a test that already tests for this behaviour in the case where the extras differ, but we don't have a test where extra is present in one and in another it is not.

PRs for this are welcome and we should have enough test coverage in places that the fix would be touching.

hartikainen: Thanks for the quick response and context! I'll look into the existing case where extras differ to see if that can be extended to this case. Let's see where I get.

hartikainen:
I think I now have **a** solution for this: creating distinct repository name for each requirement line with extras. I'm not sure if this is the best solution though, because it leads to defining multiple, identical `whl_library` repositories even when different extras point to the same wheel file. This seems inefficient and unnecessary.

After looking around a bit, I realized that maybe following comment by @aignas would be a better solution? https://github.com/bazel-contrib/rules_python/blob/fe45faabeb3dceab8766fb1a67131ec0cc1135dc/python/private/pypi/parse_requirements.bzl#L109-L114

hartikainen:
I spend quite some time trying to wrap my head around the `parse_requirements` code and how to aggregate the extras. I could not figure out a way to make that work without breaking some other tests. I don't think I currently understand the requirement parsing stuff well enough to properly fix the underlying issue.

rickeylev:
Creating another repo for `foo[x]` wouldn't be the thing to do, anyways. The extra doesn't affect the contents of the package (whl file downloaded). A repo is created for each unique (package, version) (i.e. one per unique artifact downloaded).

What the extra does is affect the `Requires-Dist` resolution later, when figuring out the list of dep targets. e.g. given `foo[a] == 1.0`, then the `pypi_foo` repo is created from the `foo-whatever.whl` file. A `py_library(deps=...)` is created based on the `Requires-Dist` entries in the wheel. Those entries can have conditions, like "only if extra X is requested".

Ignas would know this code better, but IIRC, the list of extras is kept and passed around somewhere so that evaluating the Requires-Dist lines has the right context.

aignas:
This is a bit of a mess because historically the `extra` would result in a different set of deps wired in your closure. So in order to fix this there are two approaches:
1. Figure out how to aggregate the requirements without breaking existing tests, if there is no way, change the tests.
2. Enable `pipstar` by default (#2949) - this would use starlark to parse the dependencies and we would not need to do this clever aggregation by extras anymore.

Just to outline the option number 2 (which is a better long term solution), the rough sketch is:

1. enable pipstar by default (I tried it briefly but it was not ready yet). It should be safe to do this in `bzlmod` and in theory it could even work in `WORKSPACE` with extra work on figuring out how the `env_marker_setting` can work.
2. to `whl_library` pass a list of extras for which we should create targets or create `py_library` targets for each `extra` found in the `METADATA`. The naming of the targets could be `pkg` for no extras, `pkg__foo` for `extras==foo`. If there is a dependency on `pkg[bar]` then we would need to translate that to a dependency on `@<hub_name>//pkg:pkg__bar`. If there is a dependency on `pkg[bar,baz]` then this translates to two dependencies - `@<hub_name>//pkg:pkg__bar` and `@<hub_name>//pkg:pkg__baz`. The naming convention of the targets is up for discussion.
3. to the `hub_repository` pass a list of extras for which we should generate extra aliases.
4. The `parse_requirements` code would only need to care about the extras and what need to be present where.

However, there could be still a need to fix the tests to make the behaviour more correct before we go the long way and implement option 2.

tests/pypi/extension/extension_tests.bzl
```diff
@@ -407,6 +407,97 @@ new-package==0.0.1 --hash=sha256:deadb00f2
...
+            read = lambda x: {
+                "requirements.linux_arm64.txt": """\
+package==0.7.0 \
+    --hash=sha256:4dd8924f171ed73a4f1a6191e2f800ae1745069989b69fabc45593d6b6504003 \
+    --hash=sha256:62833036cbaf4641d66ae94c61c0446890a91b2c0d153946583a0ebe04877a76
+""",
+                "requirements.linux_x86_64.txt": """\
+package[extra]==0.7.0 \
+    --hash=sha256:62833036cbaf4641d66ae94c61c0446890a91b2c0d153946583a0ebe04877a76
+""",
```
aignas: I am testing this in `main` and it seems that a universal requirements file might work better in this particular case - I cannot create a failing test case for `parse_requirements` function yet.

aignas: Actually, `parse_requirements` is not handling this correctly.

It returns output that will not work with the current implementation of the `pip.parse`, so it is to blame, but I need to look a little more at it where to do the fix.

hartikainen:
~~I just rebased this branch off of `main` and updated the approach based on the solution two suggested by @aignas's in https://github.com/bazel-contrib/rules_python/pull/3193#issuecomment-3218815034 above. For the sake of full transparency, I used [`gemini-cli`](https://github.com/google-gemini/gemini-cli) to generate most of the solution, and reviewed the code myself. The solution is relatively simple and the tests now pass, at least locally on bzlmod. Any feedback is much welcome!~~

Edit: after a further reading, the current implementation is still quite far from Ignas's comment and what it should be.

aignas:
Hmm, I broke the tests, but I think the fix may address the original issue of the incorrect treatment of packages with extras. I realized that we can do the fix entirely in `parse_requirements` function, so maybe it would be better to change the test to use that instead.

aignas:
Ignore me, I got confused myself in the code a little.

hartikainen:
Closing this in favor of #3385.

PR #3194, opened 2025-08-18
fix(py_wheel): add directories in deterministic order

jwnimmer-tri:
A call to `os.listdir` does not return a deterministic result from one run to the next.  So for example if the output of a skylib [copy_directory](https://github.com/bazelbuild/bazel-skylib/blob/main/docs/copy_directory_doc.md) is added to a wheel, the files will be in a random order.

### DIFF
--- a/tools/wheelmaker.py
+++ b/tools/wheelmaker.py
@@ -152,7 +152,7 @@ class _WhlFile(zipfile.ZipFile):
         """Add given file to the distribution."""
 
         if os.path.isdir(real_filename):
-            directory_contents = os.listdir(real_filename)
+            directory_contents = sorted(os.listdir(real_filename))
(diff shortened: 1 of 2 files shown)


tools/wheelmaker.py
```diff
@@ -152,7 +152,7 @@ def add_file(self, package_filename, real_filename):
         """Add given file to the distribution."""
 
         if os.path.isdir(real_filename):
-            directory_contents = os.listdir(real_filename)
+            directory_contents = sorted(os.listdir(real_filename))
```
rickeylev, line 155:
```suggestion
            # listdir() results are unordered; sort for determinism.
            directory_contents = sorted(os.listdir(real_filename))
```

rickeylev (approved):

PR #3195, opened 2025-08-19
fix(rules): make `py_console_script_binary` compatible with symbolic macros

janwinkler1:
Fixes target naming to follow symbolic macro conventions introduced in Bazel 8.0.

### DIFF
--- a/python/private/py_console_script_binary.bzl
+++ b/python/private/py_console_script_binary.bzl
@@ -53,6 +53,7 @@ def py_console_script_binary(
         script = None,
         binary_rule = py_binary,
         shebang = "",
+        main = None,
         **kwargs):
     """Generate a py_binary for a console_script entry_point.
 
@@ -66,6 +67,8 @@ def py_console_script_binary(
             package as the `pkg` Label.
         script: {type}`str`, The console script name that the py_binary is going to be
             generated for. Defaults to the normalized name attribute.
+        main: {type}`str`, the python file to be generated, defaults to `<name>_entry_point.py` to
+            be compatible with symbolic macros.
         binary_rule: {type}`callable`, The rule/macro to use to instantiate
             the target. It's expected to behave like {obj}`py_binary`.
             Defaults to {obj}`py_binary`.
@@ -73,13 +76,13 @@ def py_console_script_binary(
             Defaults to empty string.
         **kwargs: Extra parameters forwarded to `binary_rule`.
     """
-    main = name + "_entry_point.py"
+    main = main or name + "_entry_point.py"
 
     if kwargs.pop("srcs", None):
         fail("passing 'srcs' attribute to py_console_script_binary is unsupported")
 
     py_console_script_gen(
-        name = name + "_gen",
+        name = name + "_gen__",
         entry_points_txt = entry_points_txt or _dist_info(pkg),
         out = main,
         console_script = script,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -108,6 +108,8 @@ END_UNRELEASED_TEMPLATE
   variants. Setting {obj}`--py_linux_libc=musl` and `--py_freethreaded=yes` now
   activate them, respectively.
   ([#3262](https://github.com/bazel-contrib/rules_python/issues/3262)).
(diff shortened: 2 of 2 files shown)


golithe:
@aignas thank you for your review!

sure, i am looking into it ATM.  

in the [dev-guide](https://rules-python.readthedocs.io/en/latest/devguide.html?utm_source=chatgpt.com#writing-tests) i read that integration tests are discuraged. 

in https://github.com/bazel-contrib/rules_python/commit/c91459a2475c6f8200b82111c2fd87fb0e72122f i quickly tried to add a simpler build test, however that failed in CI, presumably because bazel 7 doesnt support symbolic macros.

so do you think in this case it would be okay to add a new integration test and ignore (if possible) bazel 7 tests there?

python/private/py_console_script_binary.bzl
```diff
@@ -73,13 +73,13 @@ def py_console_script_binary(
...
     """
-    main = "rules_python_entry_point_{}.py".format(name)
+    main = name + "_entry_point.py"
 
     if kwargs.pop("srcs", None):
         fail("passing 'srcs' attribute to py_console_script_binary is unsupported")
 
     py_console_script_gen(
-        name = "_{}_gen".format(name),
+        name = name + "_gen",
```
aignas: This is making the target as `foo_gen` instead of `_foo_gen` that it previously was.

python/private/py_console_script_binary.bzl
```diff
@@ -73,13 +73,13 @@ def py_console_script_binary(
             Defaults to empty string.
         **kwargs: Extra parameters forwarded to `binary_rule`.
     """
-    main = "rules_python_entry_point_{}.py".format(name)
+    main = name + "_entry_point.py"
```
aignas: Just out of curiosity, could you please explain why you've selected to use this?

python/private/py_console_script_binary.bzl
golithe: Yes, this naming change is intentional. The underscore prefix violates Bazel's symbolic macro naming rules. Per the documentation [1], targets created by a symbolic macro must be named either the same as the macro or prefixed by the macro name followed by _, ., or -. The new naming satisfies this requirement.

[1][ https://bazel.build/extending/macros - "Naming conventions for targets created"](https://bazel.build/extending/macros#naming)

python/private/py_console_script_binary.bzl
golithe: This, too, is required for symbolic macro compatibility. The original `.format()` approach created filenames that also violated the naming rules [1].

[1][ https://bazel.build/extending/macros - "Naming conventions for targets created"](https://bazel.build/extending/macros#naming)

python/private/py_console_script_binary.bzl
aignas: The initial implementation was on purpose - we did not want this target to be depended upon. How would you propose we do this moving to the future?

python/private/py_console_script_binary.bzl
aignas: The intention here was to create a file with the same name as the legacy entry point python file within the `whl_library` repository rule. I am OK with this being changed though.

rickeylev:
re: testing this: Is adding target_compatible_with sufficient? Or does bazel 7 failure before it that is processed? If TCW isn't checked early enough, then how about wrapping it in a legacy macro, then checking the bazel version there, and skiping early? There's utility target in rules_testing for generating a no-op skip target for such cases; see skip_test in tests/py_runtime_pair/py_runtime_pair_tests.bzl for an example

re: losing the underscore prefix: We'll just have to live with it. We could add an infix or suffix to help denote its internal. e.g. `foo_gen__ foo__gen foo_internal_gen foo_private_gen` etc

python/private/py_console_script_binary.bzl
aignas:
```suggestion
        name = name + "_gen__",
```

python/private/py_console_script_binary.bzl
aignas: Updated to make it overridable.

Issue #3105, opened 2025-07-18
Feat: Gazelle directive `# gazelle:python_ignore_target target`


dougthor42:
# 🚀 feature request

### Relevant Rules

gazelle

### Description

Gazelle can skip files/directories via `# gazelle:exclude file` and can ignore entire `BUILD.bazel` files with `# gazelle:ignore`. `rules_python_gazelle_plugin` also adds a `# gazelle:ignore import` annotation that can be added to `.py` files.

Gazelle can also force deps to be kept around with the `# keep` comment within a target.

However, there's not currently (that I'm aware of) any way to tell gazelle to ignore an entire _target_.

Ignoring an entire target can be useful when there are hand-written targets or ones that don't follow the rest of the directives (such as additional file-level targets when the generation mode is package). Having such a directive will also unblock the issues I raised on @yushan26's PR #3046, allowing them to expand the scope from just py_binary targets to all targets.

### Describe the solution you'd like

New directive: `# gazelle:ignore_target target`.

Or if we don't want do potentially shadow a future directive from bazel_gazelle, `# gazelle:python_ignore_target target`. We could also bike shed about "ignore" vs "exclude".

If `target` isn't found, the directive has no effect. Maybe it logs a warning, but that's it.

Should wildcards be allowed? For an MVP I don't think so. That adds complexity with, at least for my direct use case, little gain.

### Describe alternatives you've considered

You may ask: "How is this different from a `# keep` comment at the start of a target?"

```starlark
# keep: this cannot be managed by Gazelle because <reason>
py_library(
    ...
)
```

The main difference is that the directive can be set in any parent BUILD.bazel file, thus allowing people to have a central location for all such directives:

```starlark
# gazelle:python_ignore_target //foo/bar:baz
# gazelle:python_ignore_target //foo/bar:baz_test
# gazelle:python_ignore_target //a/b/c/d:e
```

Admittedly it's a pretty weak argument for having a new directive.

There may also be subtle differences with how gazelle collects targets and generates new ones (eg https://github.com/bazel-contrib/bazel-gazelle/issues/1176) but I haven't fully investigated those yet.

PR #3197, opened 2025-08-19
gazelle: Add Gazelle directive # gazelle:python_ignore_target target

yushan26:
Implement a new directive to ignore targets in an existing BUILD.bazel file to prevent gazelle from modifying or cleaning up existing python rules.
```
# gazelle:python_ignore_target //foo/bar:baz
# gazelle:python_ignore_target //foo/bar:baz_test
# gazelle:python_ignore_target //a/b/c/d:e
```
Fixes: #3105

### DIFF
--- a/gazelle/pythonconfig/pythonconfig.go
+++ b/gazelle/pythonconfig/pythonconfig.go
@@ -112,6 +112,8 @@ const (
 	// like "import a" can be resolved to sibling modules. When disabled, they
 	// can only be resolved as an absolute import.
 	PythonResolveSiblingImports = "python_resolve_sibling_imports"
+	// PythonIgnoreTarget represents the directive that controls whether to ignore the specific target.
+	PythonIgnoreTarget = "python_ignore_target"
 )
 
 // GenerationModeType represents one of the generation modes for the Python
@@ -204,6 +206,7 @@ type Config struct {
 	generatePyiDeps                           bool
 	generateProto                             bool
 	resolveSiblingImports                     bool
+	ignoredTargets                            map[string]struct{}
 }
 
 type LabelNormalizationType int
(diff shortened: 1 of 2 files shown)


gazelle/pythonconfig/pythonconfig.go
```diff
@@ -610,6 +615,17 @@ func (c *Config) ResolveSiblingImports() bool {
...
 }
 
+// SetIgnoreTarget sets the target to be ignored.
+func (c *Config) AddIgnoreTarget(target string) {
+	c.ignoredTargets[target] = struct{}{}
+}
+
+// IgnoreTarget returns whether the target should be ignored.
+func (c *Config) IgnoreTarget(target string) bool {
+	_, ignores := c.ignoredTargets[target]
```
dougthor42, line 625: I wonder... should we support wildcards/globbing/regex? And how about the `$python_root$` replacement?

```starlark
# gazelle:python_ignore_target //$python_root$/foo/bar:*_foo

py_library(name = "bar_foo", ...)  # ignored
py_library(name = "bar_bar", ...)  # not ignored
py_library(name = "foo_foo", ...)  # ignored
```

Regex is probably too complex, especially if we want to support the `$python_root$` replacement literal. But simple glob-like matching might be OK. There's precedence for globs, too: `TestFilePattern` uses it and we can make use of `matchesAnyGlob`.

gazelle/pythonconfig/pythonconfig.go
```diff
@@ -610,6 +615,17 @@ func (c *Config) ResolveSiblingImports() bool {
 	return c.resolveSiblingImports
 }
 
+// SetIgnoreTarget sets the target to be ignored.
+func (c *Config) AddIgnoreTarget(target string) {
+	c.ignoredTargets[target] = struct{}{}
+}
+
+// IgnoreTarget returns whether the target should be ignored.
+func (c *Config) IgnoreTarget(target string) bool {
```
dougthor42, line 624: Let's call this `IsIgnoredTarget`.

PR #3210, opened 2025-08-26
cleanup: remove support for extra actions

comius:
This removes the support for Bazel "extra actions".

These have been long deprecated and little to no usage. Because of how long they've
been deprecated, their lack of  use, and how long their replacement (aspects) has
been available, this is not being considered a breaking change.

Fixes https://github.com/bazelbuild/bazel/issues/16455
Fixes https://github.com/bazel-contrib/rules_python/issues/3215

### DIFF
--- a/.github/workflows/check_do_not_merge_label.yml
+++ b/.github/workflows/check_do_not_merge_label.yml
@@ -1,0 +1,20 @@
+name: "Check 'do not merge' label"
+
+on:
+  pull_request_target:
+    types:
+      - opened
+      - synchronize
+      - reopened
+      - labeled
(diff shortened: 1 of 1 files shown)


rickeylev:
@comius : Are extra actions present in Bazel 7 or Bazel 8?

@aignas : I'm inclined to not call this a breaking change. I don't think the extra action stuff was really used externally; even within Google it only had a handful of cases. Thoughts?

aignas (approved):
+1 for not calling it a breaking change.

rickeylev (approved):
Thanks @comius !

PR #3213, opened 2025-08-26
chore: create workflow to check the do-not-merge label

rickeylev:
We have the label, but it doesn't do anything. Add a workflow that can check it, to
be added as a required status check.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -70,12 +70,12 @@ END_UNRELEASED_TEMPLATE
 ### Removed
 * Nothing removed.
 
-{#1-6-0}
+{#v1-6-0}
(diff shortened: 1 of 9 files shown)


rickeylev:
Er, actually, I'll move this back to draft. I had an AI generate this, but didn't verify if it's correct. The `on` conditions look different than our other workflows, so I have to go crosscheck that.

rickeylev:
Ok, consulted the docs. LGTM.

The difference between "pull_request" and "pull_request_target" is simply what code the action sees when it runs (post vs pre PR changes). Doesn't matter in this case, since it just cares about the PR state.

aignas (approved):

PR #3219, opened 2025-08-27
chore: make release tool auto detect next version

rickeylev:
This makes the release tool determine the next version automatically. It does so
by searching for the VERSION_NEXT strings. If VERSION_NEXT_FEATURE is found, then
it increments the minor version. If only patch placeholders are found, then it
increments the patch version.

When the latest version is an RC, an error is raised. This is to protect against
accidentally running it when we're in the middle of the RC phase.


rickeylev:
PTAL

aignas (approved):
LGTM, thanks!

PR #3220, opened 2025-08-27
fix(pypi): add `__init__.py` to site-packages directories

### LINKED ISSUE #3038: namespace packages broken because pkgutil shims not generated by default
# 🐞 bug report

### Affected Rule

I'm assuming there's some issue in `py_binary` or `py_library`. Not sure.

### Is this a regression?

Yes, 1.4.1 appears to work, but 1.5.0-rc4 does not.

### Description

I can no longer import the `kfp.compiler` module.

## 🔬 Minimal Reproduction

Check out the reproduction here: https://github.com/philsc/rules_python-repro/tree/repro-kfp-1.5.0-rc4

The README showcases the issue I'm running into.

## 🔥 Exception or Error

Doing this:
```python
import kfp
print(kfp)

import kfp.compiler
print(kfp.compiler)
```
And having this BUILD target:
```python
py_binary(
    name = "bin",
    srcs = ["bin.py"],
    deps = [
        requirement("kfp"),
    ],
)
```
results in this error:
```
<module 'kfp' from '/data/phil_cache/bazel/_bazel_philipp.schrader/29d62092ea113224b51636420449a2d3/execroot/__main__/bazel-out/k8-fastbuild/bin/bin.runfiles/my_deps_kfp_pipeline_spec/site-packages/kfp/__init__.py'>
  File "/data/phil_cache/bazel/_bazel_philipp.schrader/29d62092ea113224b51636420449a2d3/execroot/__main__/bazel-out/k8-fastbuild/bin/bin.runfiles/__main__/bin.py", line 4, in <module>
    import kfp.compiler
ModuleNotFoundError: No module named 'kfp.compiler'
```

**Rules_python version:**

`1.5.0`

### Workaround

Set `pip.parse.enable_implicit_namespace_pkgs = True` in 1.5

Note that this will have to be changed to False in subsequent versions; the bug is an inverted logic check after some refactoring.

aignas:
Before this change we would be not adding `__init__.py` files for
`site-packages` directories due to an oversight in how the root
directory is handled. As user has reported, we should do this to be
backwards compatible.

Fixes #3038

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,6 +61,10 @@ END_UNRELEASED_TEMPLATE
 * (bootstrap) The stage1 bootstrap script now correctly handles nested `RUNFILES_DIR`
   environments, fixing issues where a `py_binary` calls another `py_binary`
   ([#3187](https://github.com/bazel-contrib/rules_python/issues/3187)).
+* (pypi) Now the `site-packages` will also have the `__init__.py` file generated
+  for `namespace` packages to correctly support cases where one package may be a Python
+  file and the other package may be a proper directory module.
+  Fixes ([#3038](https://github.com/bazel-contrib/rules_python/issues/3038)).
 
 {#v0-0-0-added}
 ### Added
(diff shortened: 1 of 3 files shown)


rickeylev:
This is really weird. I can't make sense how this would fix the described behavior, nor how the existing code results in the described behavior. I'll take further convo to the issue.

aignas:
I wrote it when it was already late, let me close it as I think it is not appropriate.

PR #3225, opened 2025-08-30
feat(pypi): enable pipstar by default

### LINKED ISSUE #2949: feat: enable pipstar by default

aignas:
Before this PR we were using our python helpers to parse the whl METADATA for a
particular host (or experimentally - target) platform and the fixed dependency
list would be written to the `BUILD.bazel` files materialized by the
`whl_library` repository rule.

This PR adds extra plumbing to leverage the Starlark implementation of the
METADATA file parsing (a.k.a. pipstar) which moves the evaluation to analysis
phase as we will get the target platform details via the
`env_marker_config_setting` from the `toolchain` used for the dependencies.

Since users are normally working with a subset of platforms that the METADATA
would pull in, we are writing the list of packages from the requirements file to
a `bzl` file so that `pipstar` knows which packages to consider when parsing the
METADATA. This will ensure that `bazel query` continues to work.

Since just passing the list of packages to the `whl_library` would cause
refetches of all of the `whl_library` dependencies when we add or remove a
package, we pass a path where to load the symbol called `packages` from, which
means that only the analysis phase will be affected because the macros will be
re-evaluated if one adds or removes a package.

We still need to download the correct platform-specific wheels for the right
platform in order to fully resolve the referenced tickets. With this PR we are
deprecating the `experimental_target_platforms` attribute since it has no longer
any effect.

Fixes #2949
Work towards #260
Work towards #2241


aignas:
Need to wait for the #3247 so that I can finish fixing the tests

python/private/internal_dev_deps.bzl
```diff
@@ -53,6 +54,7 @@ def _internal_dev_deps_impl(mctx):
         whl_file = "@implicit_namespace_ns_sub1_whl//:ns_sub1-1.0-any-none-any.whl",
         requirement = "ns-sub1",
         enable_implicit_namespace_pkgs = False,
+        dep_template = "@rules_python//tests/support:support.bzl",
```
rickeylev: This dep_template setting looks wrong?

rickeylev (changes requested):

aignas:
OK, PTAL.

aignas:
TODO:
- [x] revert `requirements.bzl` and `config.bzl` move in `bzlmod`.
- [x] create an extra repo for `config.bzl` storing in `WORKSPACE` named `<prefix>_config` (e.g. `pypi__config`). This does not have the negative effect for passing the `packages` list.

python/private/pypi/whl_config_repo.bzl
```diff
@@ -0,0 +1,46 @@
+"""whl_config_library implementation for WORKSPACE setups."""
```
aignas, line 1: NOTE: we could in theory reuse the `hub_repository.bzl` which is only used in `bzlmod` in here, but I want to leave this out of scope of this PR.

The main realization of this PR is that the WORKSPACE to bzlmod mapping of components is as following:
- bzlmod extension maps to the `pip_parse` repository rule
- the bzlmod hub repo maps to the `config` repository created here
- the spokes in both are the same.

python/private/pypi/whl_library_targets.bzl
```diff
@@ -274,12 +274,18 @@ def whl_library_targets(
     if group_name and "//:" in dep_template:
         # This is the legacy behaviour where the group library is outside the hub repo
         label_tmpl = dep_template.format(
-            name = "_groups",
+            name = "_config",
             target = normalize_name(group_name) + "_{}",
+        ).replace(
+            "//:",
+            "//_groups:",
```
aignas, line 285: Here it is a little more hacky than I would like, but this would disappear together with WORKSPACE, so it is fine for the time being - no need to add another parameter to the macro.

aignas (commented):
Self review.

aignas:
PTAL

python/private/pypi/pip_repository.bzl
```diff
@@ -202,14 +202,31 @@ def _pip_repository_impl(rctx):
...
+                    for pkg in bzl_packages
+                }, value_repr = lambda x: "None"),
+            },
+        )
     rctx.template("requirements.bzl", rctx.attr._template, substitutions = {
-        "    # %%GROUP_LIBRARY%%": """\
-    group_repo = "{name}__groups"
-    group_library(
-        name = group_repo,
+        "    # %%CONFIG_REPO%%": """\
```
rickeylev: Since we're here, please move this triple-quoted chunk to a module-level constant

python/private/pypi/pip_repository.bzl
```diff
@@ -202,14 +202,31 @@ def _pip_repository_impl(rctx):
...
         repo_prefix = "{name}_",
         groups = all_requirement_groups,
-    )""".format(name = rctx.attr.name) if not rctx.attr.use_hub_alias_dependencies else "",
+        whl_map = {{
+            p: ""
+            for p in all_whl_requirements_by_package
+        }},
+    )""".format(name = rctx.attr.name) if not rctx.attr.use_hub_alias_dependencies else """\
+    config_repo = "{name}"
+""".format(name = rctx.attr.name),
```
rickeylev: This multi-line ternary of a triple-quote conditional format with another triple-quote is hard to read. Please split into separate statements.

python/private/pypi/whl_config_repo.bzl
rickeylev: SGTM.

Yeah, without some more broad refactorings, re-using some of the "common" code between workspace and bzlmod is somewhat convoluted.

python/private/pypi/whl_config_repository.bzl
```diff
@@ -0,0 +1,46 @@
+"""whl_config_library implementation for WORKSPACE setups."""
+
+load("//python/private:text_util.bzl", "render")
+load(":generate_group_library_build_bazel.bzl", "generate_group_library_build_bazel")
+
+def _impl(rctx):
```
rickeylev:
```suggestion
def _whl_config_repo_impl(rctx):
```

to match project and broader convention (also makes grepping for a rule's impl function easy)

python/private/pypi/whl_config_repository.bzl
```diff
@@ -0,0 +1,46 @@
...
+    rctx.file("BUILD.bazel", "")
+    rctx.template(
+        "config.bzl",
+        rctx.attr._config_template,
+        substitutions = {
+            "%%PACKAGES%%": render.dict(rctx.attr.whl_map or {}, value_repr = lambda x: "None"),
+        },
+    )
+
+whl_config_repository = repository_rule(
```
rickeylev: nit: whl_config_repo? Just a bit shorter; I think we have about an even number of repo vs repository suffixes, but most new ones we've been using repo

python/private/pypi/whl_library.bzl
```diff
@@ -369,7 +369,10 @@ def _whl_library_impl(rctx):
                 timeout = rctx.attr.timeout,
             )
 
-    if rp_config.enable_pipstar:
+    # NOTE @aignas 2025-09-28: if someone has an old vendored file that does not have the
+    # dep_template set or the packages is not set either, we should still not break, best to
+    # disable pipstar for that particular case.
```
rickeylev, line 374: Should we bother? Revendoring is pretty easy, right?

Also, is this note something worth putting into the docs/changelog somewhere?

python/private/pypi/whl_library_targets.bzl
rickeylev: SGTM. Just put a comment giving context

python/private/pypi/whl_library.bzl
aignas: I found that revendoring requires bazel to run and if the loading/config is broken, it makes it super annoying, I'll if this is breaking or not.

aignas:
This should be a drop-in replacement, oh I think we need to point users to how they can configure the environment used for doing the evaluation.

python/private/pypi/whl_library.bzl
aignas: I thought about this and if the user is using vendored requirements and updates the rules_python version, they might not be able to revendor because the `requirements.bzl` is from somewhere else. So for now it probably is best to not break and accommodate this in a backwards compatible way.

PR #3226, opened 2025-08-30
feat(pypi): support aarch64 windows on pipstar

aignas:
This adds the necessary bits for Windows ARM64 support when evaluating
env markers in requirements files and selecting the right wheels in
`experimental_index_url`.

Related #2276


aignas:
We cannot do this because we have tests using the `//docs` and due to the lack of wheel for `markupsafe`, we cannot include `win_arm64` in the `rules_python` defaults.

TBH, maybe we should treat this as a bug and fix it.

rickeylev:
Yeah, it's something we should fix.

It should be OK for a requirements file to not support the full set of platforms the entire build does. I can see that situation happening in a larger repo.

rickeylev:
Another thought: does this mean if we add a new platform to rules_python's default, downstream user requirements could break? Eek.

aignas:
I think the best way would be to hardcode windows to only mean windows intel, then for linux and osx, do similar things. The users should really be specifying things explicitly...

Otherwise us adding platforms may break bazel quely.

aignas:
Funny enough, but #3307 may fix our CI because `markupsafe` is adding the `windows aarch64` wheels and if we did attempt this before that, we would have not seen this breakage. We have to add a unit test of some sorts before we merge this.

aignas:
Just as I had suspected, the new wheels in sphinx deps masked the failure.

PTAL, I am wondering if the test changes indicate that we should do something differently - add freethreaded variants to default platforms, or other things. Hence the ask for review.

aignas:
Decided to just go forward with adding the platform and see what feedback we get in rc1.

PR #3233, opened 2025-09-03
feat: Allow files in wheels to be installed to directories

Ahajha:
When specifying `data_files` in `py_wheel`, allow just the directory to be specified (with a trailing slash), in which case it will use the existing filename. This avoids duplicating (potentially platform-specific) names. Additionally, targets with multiple files can be installed as a group to a folder, with the same filename-preserving behavior. In general I think this is a better starting point, as I imagine most of the time users would want to preserve the names.

Before, this would result in the file simply not being installed, so this only changes already-broken behavior.

### DIFF
--- a/python/private/py_wheel.bzl
+++ b/python/private/py_wheel.bzl
@@ -206,7 +206,7 @@ filegroup(name = "files", srcs = [":file1.txt", ":file2.txt"])
 
 Allowed paths: {prefixes}
 
-:::{versionchanged} VERSION_NEXT_FEATURE
+:::{{versionchanged}} VERSION_NEXT_FEATURE
 Values can end in slash (`/`) to indicate that all files of the target should
 be moved under that directory.
 :::


Ahajha:
Sorry for the long pause while this sat here, been a busy few months but I got some time to come back to this. Still need to add a few tests + update docs, but the core functionality is ready.

Ahajha:
@aignas Can I get a review on this? We've been using it internally for a while and it's working well for us. Happy to add tests/docs, etc. if you'd like.

Ahajha:
Yup, sure thing!

Ahajha:
Done, done, and done!

rickeylev (approved):
Looks like Ignas's comments have been address, so approving and enqueuing

Ahajha:
Huh, failures look unrelated to my changes (as far as I can tell, happy to be told otherwise). What's the policy when this happens?

rickeylev:
Bazel ci and/or GitHub has been flaky with network reliability lately, so we just requeue and retry

Issue #2652, opened 2025-03-07
Switching to modules causes LoadLibrary to fail in some python extensions due to MAX_PATH length


carpenterjc:
# 🐞 bug report

### Affected Rule
Its within the pip extension. 

```
pip = use_extension("@rules_python//python/extensions:pip.bzl", "pip")
pip.parse( ...
)
```

### Is this a regression?
This works without modules, but doesn't now modules are enabled. 

### Description

Using the module extension to download whl files adds significantly to the path. 
`rules_python~~pip~example_pip_312_charset_normalizer` 
vs:
`example_pip_charset_normalizer`
when running on windows with `--nobuild_python_zip` you can easily exceed MAX_PATH, and any library which uses extensions will fail to load the extensions. 

## 🔬 Minimal Reproduction
Reproduction attached:
[pip_path_length.zip](https://github.com/user-attachments/files/19126008/pip_path_length.zip)

Run: `bazel build //example/...` you get the following error:
```
INFO: Invocation ID: 19b20afe-a6fa-4573-bac9-f74bccb40807
INFO: Analyzed 3 targets (85 packages loaded, 3896 targets configured).
ERROR: D:/bazel_bugs/pip_path_length/example/BUILD.bazel:19:8: Executing genrule //example:run failed: (Exit 1): sh.exe failed: error executing Genrule command (from target //example:run) C:\Program Files\Git\bin\sh.exe -c ... (remaining 1 argument skipped)
Traceback (most recent call last):
  File "c:\dev\_bazel\x3mr6iom\execroot\_main\bazel-out\x64_windows-opt-exec-st-d57f47055a04\bin\example\a_moderatly_long_tool_path.exe.runfiles\rules_python~~pip~example_pip_312_requests\site-packages\requests\compat.py", line 11, in <module>
    import chardet
ModuleNotFoundError: No module named 'chardet'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "c:\dev\_bazel\x3mr6iom\execroot\_main\bazel-out\x64_windows-opt-exec-st-d57f47055a04\bin\example\a_moderatly_long_tool_path.exe.runfiles\_main\example\test.py", line 2, in <module>
    import requests
[...]

PR #3242, opened 2025-09-06
fix(system_python): write import paths to generated file instead of using PYTHONPATH

rickeylev:
This changes the system_python bootstrap to use a 2-stage process like the script
bootstrap does. Among other things, this means the import paths are written to
a generated file (`bazel_site_init.py`, same as boostrap=script) and sys.path
setup is performed by the Python code in stage 2.

Since the PYTHONPATH environment variable isn't used, this fixes the problem on
Windows where the value is too long.

This also better unifies the system_python and script based bootstraps because the
same stage 2 code and bazel_site_init code is used.

Along the way, several other improvements:

* Fixes path ordering for system_python. The order now matches venv ordering
  (stdlib, binary paths, runtime site packages).
* Makes the venv-based solution work when the site module is disabled (`-S`).
* Makes `interpreter_args` attribute and
  `RULES_PYTHON_ADDITIONAL_INTERPRETER_ARGS` env var work with system_python.
* Makes `main_module` work with system_python.
* Progress towards a supportable non-shell based bootstrap (a user requested
  this because their environment doesn't install any shells as a security
  precaution).

Fixes https://github.com/bazel-contrib/rules_python/issues/2652

### DIFF
--- a/python/private/py_executable.bzl
+++ b/python/private/py_executable.bzl
@@ -140,6 +140,9 @@ This is mutually exclusive with {obj}`main`.
 
 :::{versionadded} 1.3.0
 :::
+:::{versionchanged} VERSION_NEXT_FEATURE
+Support added for {obj}`--bootstrap_impl=system_python`.
+:::
 """,
         ),
         "pyc_collection": lambda: attrb.String(
--- a/python/private/py_executable.bzl
+++ b/python/private/py_executable.bzl
@@ -645,11 +645,6 @@ def _create_venv(ctx, output_prefix, imports, runtime_details):
     if pyvenv_cfg:
         files_without_interpreter.append(pyvenv_cfg)
 
-    venv_root = pth.short_path
-    for _ in range(site_packages.count("/") + 1):
-        venv_root = paths.dirname(venv_root)
-    venv_root = runfiles_root_path(ctx, venv_root)
-
     return struct(
         # File or None; the `bin/python3` executable in the venv.
         # None if a full venv isn't created.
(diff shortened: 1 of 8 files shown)


python/private/stage2_bootstrap_template.py
```diff
@@ -66,7 +66,7 @@ def get_windows_path_with_unc_prefix(path):
             break
         except (ValueError, KeyError):
             pass
-    if win32_version and win32_version >= '10.0.14393':
+    if win32_version and win32_version >= "10.0.14393":
```
aignas, line 72: This line remembered me someone posting on #2652, I wonder if this PR would address that as well, where Windows is suffering hitting `MAX_PATH`.

aignas (commented):
Gave a quick look at this and I think the change is really exciting. I'll do a thorough review once it is ready for review.

rickeylev:
fyi

@mering -- non-shell based bootstrap (still needs a system python, so not our coveted native launcher, but doesn't require you to have a shell installed)

@jacky8hyf -- should make bootstrap=script work with site disabled (`-S`)

python/private/stage2_bootstrap_template.py
rickeylev: Possibly, yes. It sounds like LoadLibrary is affected by the environment variables, and if they're too big, fails.

rickeylev:
Ok, ready for review

python/private/py_executable.bzl
```diff
@@ -626,14 +641,28 @@ def _create_venv(ctx, output_prefix, imports, runtime_details):
     }
     venv_symlinks = _create_venv_symlinks(ctx, venv_dir_map)
 
+    files_without_interpreter = [pth, site_init] + venv_symlinks
+    if pyvenv_cfg:
+        files_without_interpreter.append(pyvenv_cfg)
+
+    venv_root = pth.short_path
+    for _ in range(site_packages.count("/") + 1):
+        venv_root = paths.dirname(venv_root)
```
aignas: This is a little mysterious, could you please add a comment about what this code is going for?

rickeylev: I recalled a helper function, py_internal.get_label_repo_runfiles_path, that can directly return the runfiles path using a label as input. I switched to using that. Its much more obvious that what we're doing is figuring out the runfiles path for the root of the venv.

PR #3244, opened 2025-09-06
docs: fix pr doc builds by removing external_version_warning plugin

rickeylev:
Doc builds for PR were failing because the readthedocs_ext.external_version_warning
plugin wasn't handling something correctly. Activating it manually was originally
done to get the warning banners to appear, but it looks like RTD now displays a
warning banner without this special plugin being needed.

Since it's now unnecessary, remove the code that can activate it.


rickeylev:
Turns out the doc build errors in PRs were an actual bug, not a flake. Found and fixed.

aignas (approved):

Issue #2094, opened 2024-07-26
Support a default hub repo concept in bzlmod


aignas:
# 🚀 feature request

### Relevant Rules

`pip.parse`

### Description

Currently the `pip.parse` needs the `hub_name` parameter, which in reality is quite often set to `pypi` or `pip` in the root module. In the non-root modules the users should be also able to use `pip` or `pypi` names and this could be the default `hub_name` value.

### Describe the solution you'd like

Minimum:
* Set the `hub_name` default value to `pypi` (it should not be `pip` because we are not using `pip` here and `pypi` is a more generic term).
* Support isolation of `pip.parse` extension usage via the `isolate` parameter.
* We could force users to use the `isolate` parameter if the `hub_name` is not set, or if it is set to `pip` or `pypi` values

Extra:
* Support passing `pyproject.toml` and/or constraint files to `pip.parse` instead of the lock file itself. Enable users to lock the dependency tree via `@pypi//:lock` target to generate a lock file in the users workspace. An alternative would be to merge the requirements in some way, but that could be harder. This greatly depends on #1975 because `uv` has a lot of different flags to allow overriding dependencies in the constraint files and may enable us to have a working solution. Merging the lock file by just blindly using the lowest version of all could work, but may not be reliable and the solution to properly do the locking may be required.

The lock file merging is used in the `go` ecosystem and requiring users to either name the hub repo that collects the spoke repos or using a default name is the convention of `rules_go/gazelle`, see bazelbuild/bazel-gazelle#1584. Also see bazelbuild/bazel#20186, which talks about stabilizing extension isolation feature. Having `rules_python` supporting it may be a good way to give feedback to bazel devs.

See #2088 for a slightly related topic.

### Describe alternatives you've considered

The main alternative would be to not support it and keep things as they are, but that could make it harder to implement the `isolation` later down the line. What is more, supporting `isolation` of the extensions could make the `pip` code simpler, because we could just ask the user to use `isolate` if we see clashes.

groodt:
> * Set the `hub_name` default value to `pypi` (it should not be `pip` because we are not using `pip` here and `pypi` is a more generic term).

+1 For pypi and not pip

> * Support passing `pyproject.toml` and/or constraint files to `pip.parse` instead of the lock file itself. Enable users to lock the dependency tree via `@pypi//:lock` target to generate a lock file in the users workspace.

This is one additional reason why I've been keen to move dependency specifier lists into Starlark.

[...]

PR #3248, opened 2025-09-07
feat: allow registering arbitrary settings for py_binary transitions

rickeylev:
This implements the ability for users to add additional settings that py_binary, py_test,
and py_wheel can transition on.

There were three main use cases motivating this feature:
1. Making it easier to have multiple pypi dependency closures and shared dependencies.
2. Making it easier to override flags for `py_wheel`.
3. Making it easier to have per-target setting of things like bootstrap_impl, venv
   site packages, etc.

It also adds most of our config settings to the the transition inputs/outputs for those
rules, which allows users to per-target force particular settings without having to
use e.g. `with_cfg` to wrap a target with the desired transition settings. It also
lets use avoid adding dozens of attributes (one per setting); today there are
about 17 flags.

Under the hood, this works by having a bzlmod api that users can pass labels to. These
labels are put into a generated bzl file, which the rules load and add to their
list of transition inputs/outputs. On the target level, the `config_settings` attribute,
which is a `dict[label, str]`, can be set to change the particular flags of interest.

Along the way...
* Create a common_labels.bzl file for the shared label strings
* Remove the defunct py_reconfig code in sh_py_run_test.


docs/howto/common-deps-with-multipe-pypi-versions.md
```diff
@@ -0,0 +1,91 @@
...
+        "//:pypi_hub": "beta",
+    },
+    deps = ["@pypi_beta//requests", ":common"],
+)
+py_library(
+    name = "common",
+    deps = select({
+        ":is_pypi_alpha": ["@pypi_alpha//more_itertools"],
+        ":is_pypi_beta": ["@pypi_beta//more_itertools"],
+    }),
```
aignas, line 84: Should we do this automatically?

We could reserve some `hub` repo name and do this extra wiring automatically (or mostly automatically). Maybe it is worth having a ticket for this if we feel that we should have this eventually.

The #2094 is that ticket.

python/private/builders_util.bzl
```diff
@@ -15,6 +15,41 @@
...
 
 load("@bazel_skylib//lib:types.bzl", "types")
+load(":bzlmod_enabled.bzl", "BZLMOD_ENABLED")
+
+def normalize_transition_inout_values(arg_name, values):
+    """Normalize transition inputs/outputs to canonical label strings."""
+    for i, value in enumerate(values):
+        values[i] = normalize_transition_inout_value(arg_name, value)
+
+def normalize_transition_inout_value(arg_name, value):
```
aignas: `inout` reads very similar to `input`. I read it that way multiple times in my head. Maybe we should rename it or make it `in_out` instead? `io` would also work but has other connotations.

python/private/common_labels.bzl
```diff
@@ -0,0 +1,27 @@
...
+    PRECOMPILE = str(Label("//python/config_settings:precompile")),
+    PRECOMPILE_SOURCE_RETENTION = str(Label("//python/config_settings:precompile_source_retention")),
+    PYTHON_SRC = str(Label("//python/bin:python_src")),
+    PYTHON_VERSION = str(Label("//python/config_settings:python_version")),
+    PYTHON_VERSION_MAJOR_MINOR = str(Label("//python/config_settings:python_version_major_minor")),
+    PY_FREETHREADED = str(Label("//python/config_settings:py_freethreaded")),
+    PY_LINUX_LIBC = str(Label("//python/config_settings:py_linux_libc")),
+    REPL_DEP = str(Label("//python/bin:repl_dep")),
+    VENVS_SITE_PACKAGES = str(Label("//python/config_settings:venvs_site_packages")),
+    VENVS_USE_DECLARE_SYMLINK = str(Label("//python/config_settings:venvs_use_declare_symlink")),
```
aignas, line 26: maybe in a followup we can lowercase this. I am not sure how I feel about `labels.REPL_DEP` usage.

Having said that maybe it is actually pretty good and self explanatory. I'll leave this comment in case it sparks something.

python/private/py_executable.bzl
```diff
@@ -1958,8 +1960,8 @@ def create_executable_rule_builder(implementation, **kwargs):
         ],
         cfg = dict(
             implementation = _transition_executable_impl,
-            inputs = [_PYTHON_VERSION_FLAG],
-            outputs = [_PYTHON_VERSION_FLAG],
+            inputs = TRANSITION_LABELS + [_PYTHON_VERSION_FLAG],
+            outputs = TRANSITION_LABELS + [_PYTHON_VERSION_FLAG],
```
aignas, line 1964: nit: the `_PYTHON_VERSION_FLAG` redefines what is defined in `common_labels.bzl`. This could be also addressed as a followup.

python/private/transition_labels.bzl
```diff
@@ -0,0 +1,33 @@
...
+    labels.PRECOMPILE,
+    labels.PRECOMPILE_SOURCE_RETENTION,
+    labels.PYTHON_SRC,
+    labels.PYTHON_VERSION,
+    labels.PY_FREETHREADED,
+    labels.PY_LINUX_LIBC,
+    labels.REPL_DEP,
+    labels.VENVS_SITE_PACKAGES,
+    labels.VENVS_USE_DECLARE_SYMLINK,
+]
```
aignas: maybe we could have `labels.all`?

python/private/common_labels.bzl
```diff
@@ -0,0 +1,27 @@
+"""Constants for common labels used in the codebase."""
+
+# NOTE: str() is called because some APIs don't accept Label objects
+# (e.g. transition inputs/outputs or the transition settings return dict)
+
+labels = struct(
+    # keep sorted
+    ADD_SRCS_TO_RUNFILES = str(Label("//python/config_settings:add_srcs_to_runfiles")),
+    BOOTSTRAP_IMPL = str(Label("//python/config_settings:bootstrap_impl")),
+    EXEC_TOOLS_TOOLCHAIN = str(Label("//python/config_settings:exec_tools_toolchain")),
```
aignas, line 10: Having
```
_labels = dict(
    ...
)

labels = struct(
    all = _labels.values(),
    **_labels,
)
```

would be nicer, I think.

tests/multi_pypi/beta/pyproject.toml
```diff
@@ -0,0 +1,6 @@
+[project]
+name = "multi-pypi-test-beta"
+version = "0.1.0"
+dependencies = [
+    "more-itertools==9.0.0"
+]
```
aignas: nit: EOF is inconsistent here.

aignas (approved):
LGTM in general.

tests/multi_pypi/beta/pyproject.toml
rickeylev: @gemini-code-assist Suggest a fix to this file to add a trailing newline

docs/howto/common-deps-with-multipe-pypi-versions.md
rickeylev: Yeah, I think its functionality we should more directly fold into the rules themselves. This PR originally started with doing exactly that, but I backed off because 
* I couldn't decide on a label vs string value for the flag, and 
* It felt weird to have an attribute name on the core rules that was only there for the pypi integration. The two things are pretty loosely coupled otherwise. Additionally, this made me realize "pypi" isn't quite the correct term for what it wants to, it's more like "arbitrary_extra_string_flag_usually_to_aid_external_dependency_resolution" or something like that. Which made me realize that what it's actually trying to do is encode a bunch of more specific flags into one flag (e.g. `hub_name = pypi_cuda12_musl` is really trying to represent the idea of the hypothetical flags `--@//:cuda_version=12 --@//:libc=musl` as `//:pypi_hub_name=pypi_cuda12_musl`.

I'll post this thought to 2094

python/private/builders_util.bzl
rickeylev: Done.

python/private/common_labels.bzl
rickeylev: If it helps, I just read it as a class or module name.

python/private/common_labels.bzl
rickeylev: I was intending for this to have more than just flags (e.g. `:current_py_cc_headers` et al) basically any label we end up typing more than 3 times, so I don't think `all()` makes sense in that context.

Hm, REPL_DEP...I don't think it makes sense to have py_binary etc transition on that one. The repl targets are just intended for interactive usage. I've removed that from transition_labels.

python/private/py_executable.bzl
rickeylev: Thanks, yeah, something to address in a follow up.

PR #3255, opened 2025-09-09
fix(pypi): select the lowest available libc version by default

### LINKED ISSUE #3250: version 1.6.0+ with experimental_index_url are fetching incorrect wheels for glibc version
# 🐞 bug report

### Affected Rule

 The issue is caused by the rule:  rules_python 1.6.0+ when using experimental_index_url

### Is this a regression?

 Yes, the previous version in which this bug was not present was: 1.5.4

### Description

When using rules_python 1.6.0 or 1.6.1 and the experimental_index_url feature, incorrect wheels are fetched for the glibc version on the system. For example, systems with glibc 2.28 are attempting to download and execute manylinux_2_34 wheels instead of manylinux_2_28 ones.

## 🔬 Minimal Reproduction

https://github.com/scasagrande/rules-python-glibc-issue

## 🔥 Exception or Error

See repro README

## 🌍  Your Environment

**Operating System:**

I specifically tested on Rocky 8

Seen on both 8.3.1 and 8.4.0, but is likely independent of bazel version

**Rules_python version:**

1.6.0 and 1.6.1 are both exhibiting this behaviour.

aignas:
The #3058 PR has subtly changed the default behaviour of
`experimental_index_url` code path and I think in order to make things
easier by default for our users we should go back to that behaviour.

The selection of the wheels happens in two ways:
1. If the user specifies `whl_platform_tags` value like
   `manylinux_*_x86_64` then we we should just get the lowest available
   versioned wheel. I.e. if `manylinux_2_28_x86_64` and
   `manylinux_2_32_x86_64` are available, we should prefer the former.
2. If the user specifies `whl_platform_tags` value like
   `manylinux_2_32_x86_64` then we we should get the highest version
   that is lower or equal than the specified one. I.e. if
   `manylinux_2_21_x86_64`, `manylinux_2_28_x86_64` and
   `manylinux_2_36_x86_64` are available, we should prefer the
   `manylinux_2_28_x86_64` one.

Fixes #3250

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,6 +79,10 @@ END_UNRELEASED_TEMPLATE
   length errors due to too long environment variables.
 * (bootstrap) {obj}`--bootstrap_impl=script` now supports the `-S` interpreter
   setting.
+* (pypi) We now by default select the highest closest match wheel version if the user is
+  specifying a particular version via the `whl_platform_tags` or if any version is matched via
+  the `*` token, then we select the lowest available version.
+  Fixes [#3250](https://github.com/bazel-contrib/rules_python/issues/3250).
 
 {#v0-0-0-added}
 ### Added
--- a/python/private/pypi/select_whl.bzl
+++ b/python/private/pypi/select_whl.bzl
@@ -46,6 +46,8 @@ def _platform_tag_priority(*, tag, values):
         arch = tail
     version = (int(major), int(minor))
 
+    select_highest = True
+
     keys = []
     for priority, wp in enumerate(values):
         want_os, sep, tail = wp.partition("_")
@@ -65,6 +67,7 @@ def _platform_tag_priority(*, tag, values):
             want_major = ""
             want_minor = ""
             want_arch = tail
+            select_highest = False
         elif os.startswith(_ANDROID):
             # we set it to `0` above, so setting the `want_minor` her to `0` will make things
             # consistent.
@@ -81,7 +84,7 @@ def _platform_tag_priority(*, tag, values):
         # if want_major is defined, then we know that we don't have a `*` in the matcher.
         want_version = (int(want_major), int(want_minor)) if want_major else None
         if not want_version or version <= want_version:
-            keys.append((priority, version))
+            keys.append((priority, version if select_highest else (-version[0], -version[1])))
 
     return max(keys) if keys else None
 
--- a/tests/pypi/select_whl/select_whl_tests.bzl
+++ b/tests/pypi/select_whl/select_whl_tests.bzl
@@ -406,9 +406,10 @@ def _test_multiple_musllinux(env):
     _match(
         env,
         got,
-        # select the one with the highest version that is matching
-        "pkg-0.0.1-py3-none-musllinux_1_1_x86_64.whl",
+        # select the one with the lowest version that is matching because we want to
+        # increase the compatibility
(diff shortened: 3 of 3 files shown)


rickeylev (approved):

rickeylev:
+1 on going back to the previous behavior; I do think that's a regression insofar as users are concerned. For glibc, I think it makes sense to select the oldest (i.e most compatible).

That said, I think the `glibc=* means oldest; glibc=X means newest <= X` semantics seems prone to confusion.

I'm not sure there's any right answer to this problem, though. Minimal version selection is probably one of the better answers. But, if we look at the python package managers, they have various quirks in how they select dependencies (e.g. uv has flags to control oldest/newest selection, uv ignores `>=` python_version settings, etc). Having a lock file that just answers these questions more directly is again sounding pretty appealing.

aignas:
I'll switch to the minimum version selection in both cases and will update the docs as part of this fix. it will be less confusing.

aignas:
OK, this grew a little more than I would have liked but I wanted to improve the docs for the attr to explain the change in behaviour and I thought doing the cleanup now made sense.

I'll leave extra cleanup (e.g. being able to exclude certain version) for a later PR. This should be considered as part of the builder API ticket anyway:
* Allow users making per-wheel changes for the whl filtering (very low priority)
* Allow users excluding very old whl versions.

CHANGELOG.md
```diff
@@ -79,6 +79,12 @@ END_UNRELEASED_TEMPLATE
   length errors due to too long environment variables.
 * (bootstrap) {obj}`--bootstrap_impl=script` now supports the `-S` interpreter
   setting.
+* (pypi) We now use the MVC algorithm to select the right wheel when there are multiple wheels for
+  the target platform (e.g. `musllinux_1_1_x86_64` and `musllinux_1_2_x86_64`). If the user
+  wants to set the minimum version for the selection algorithm, use the
+  {attr}`pip.defaults.whl_platform_tags` attribute to configure that. If `musllinux_*_x86_64` is
+  specified, we will chose the lowest available wheel version.  Fixes
+  [#3250](https://github.com/bazel-contrib/rules_python/issues/3250).
```
rickeylev: @gemini-code-assist "MVC" is a typo for "minimal version select". Fix the typo and use the full phrase, no abbrevation. Also column-wrap to 80 columns.

rickeylev (approved):

python/private/pypi/extension.bzl
```diff
@@ -492,35 +492,37 @@ preference.
...
-latest format.
+MVS matching algorithm).
+
+Common patterns:
+* To select any versioned wheel for an `<os>`, `<arch>`, use `<os>_*_<arch>`, e.g.
+  `manylinux_2_17_x86_64`.
+* To exclude versions up to `X.Y` - **submit a PR supporting this feature**.
+* To exclude versions above `X.Y`, provide the full platform tag specifier, e.g.
+  `musllinux_1_2_x86_64`, which will ensure that no wheels with `musllinux_1_3_x86_64` or higher
+  are selected.
```
rickeylev, line 515: @gemini-code-assist Shorten this doc string to simply describe behavior of the API. Move detailed information about how to make a particular use case work to `docs/pypi/index.md`

rickeylev:
(I was testing to see if gemini can suggest changes to chunks of code that aren't modified or in files not in the PR. Guess it can't. Oh well, good to know.)

rickeylev:
I think PR description needs to be updated? Otherwise lgtm

PR #3261, opened 2025-09-09
wip: see what happens if python_path flag usage removed

rickeylev:
wip checking ci


rickeylev:
Hrm. Failures look to be because of the magic string the runtime env toolchain uses. Even though toolchains are enabled, it basically makes it act like they aren't, so it falls back to using `--python_path`.

I think all this needs is a powershell or batch file version of the runtime_env.sh script that calls python from path.

rickeylev:
Alternative idea: re-purpose the toolchain shebang to act as python_path ? Shebangs don't have any meaning to windows, and it's conceptually the same idea as "specify which python to run a program with"

PR #3264, opened 2025-09-10
chore: add agents guidance for creating bzl_library targets

rickeylev:
I found it was doing a very poor job, so some guidance is needed.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -25,8 +25,7 @@ into the sentence, not verbatim.
 * They should have a single `srcs` file and be named after the file with `_bzl`
   appended.
 * Their deps should be based on the `load()` statements in the source file.
-* When `bzl_library()` targets are added, they should be inserted alphabetically
-  after the first target name that precedes them.
+* `bzl_library()` targets should be kept in alphabetical order by name.
 
 Example:
 


AGENTS.md
```diff
@@ -19,6 +19,28 @@ using a grandoise title.
 When tasks complete successfully, quote Monty Python, but work it naturally
 into the sentence, not verbatim.
 
+### bzl_library targets for bzl source files
+
+* `.bzl` files should have `bzl_library` defined for them.
```
dougthor42, line 24: Do we want to use [bazel-skylib's gazelle plugin](https://github.com/bazelbuild/bazel-skylib?tab=readme-ov-file#gazelle-plugin) for such?

I don't know how feature-complete or well maintained it is though - the core bazel-skylib is good, but I don't know how much love the gazelle plugin gets. It's on my TODO list for internal projects so I figured maybe it's suitable for external too.

rickeylev: If it works, that'd be great. I spent last night arguing with an AI to create a tool to generate bzl_library stuff correctly. It'd be nice to just shut that session down entirely :joy:

rickeylev: Filed https://github.com/bazel-contrib/rules_python/issues/3265

PR #3268, opened 2025-09-14
fix: venv site packages with pkgutil packages

rickeylev:
Currently, an error occurs if one packages files are intended to go into a
sub-directory of another package's directory. This can happen when pkgutil-style
namespace packages are used, which results in multiple distributions wanting
to install the same files (pkgutil `__init__.py` files) into the same top-level
directories. This eventually results in a Bazel error because Bazel detects that
the one output is the prefix of another.

To fix, detect when distributions overlap in their paths and merge their files
manually. Internally, entries are sorted from shorted venv path to longest,
however, that's just an implementation detail.

Along the way, give agents better advice for bzl_library targets.

Fixes https://github.com/bazel-contrib/rules_python/issues/3204


rickeylev:
Good call to factor out the code so it can be unit tested. Waaaay easier to generate test cases and verify behavior.

rickeylev:
So in the code right now, I didn't implement the "link_to_path=None means remove" logic, mostly because I forgot to.

However, when I went back to add it, the semantics are more complicated than when we originally added it: there is now package, version, and kind to account for. Maybe we should just drop that functionality for the time being?

Otherwise: I think the original removal logic has a weird edge case when multiple kinds are involved: only the corresponding kind is removed, not all of them. Which strikes me as a bit weird? e.g given

```
kind=LIB, venv_path=foo, link_to_path=None
kind=BIN venv_path=foo-cli, package=foo, version=1.0, ...
kind=LIB, venv_path=foo, package=foo, version=1.0, ...
```

The result is a `bin/foo-cli` still gets populated, but `site-packages/foo` doesn't. Which strikes me as surprising.

Maybe a more sensible behavior is to match by package first? IDK. There's 4 fields, (kind, venv_path, package, and version), which seems like quite a few to come up with intuitive, sensible, and reasonable behavior. One could write a custom rule to filter the PyInfo objects and propagate them. I'm also thinking, with better options for selecting what dependencies are selected, the reason for omitting entries goes away (multiple versions of something in the dependency closure).

aignas:
I think with the new transition API we can leave the removal of symlinks unimplemented for now on. People will be able to transition to the right hub easily enough.

Otherwise, if we want to do it, we should do the matching by package first.

rickeylev:
Ok, removed the tests and logic for link removal. PTAL

@gemini-code-assist /review

rickeylev:
@gemini-code-assist Give a changelog description for this PR in markdown format, wrapped to 80 columns. Review CHANGELOG.md for existing style and format.

rickeylev:
Neat. It identified and suggested some improvements to the looping we have to do to identify paths and prefixes. I'm gonna hold off on making those changes, though; our test coverage is somewhat modest with these code paths and I'd like to get this fix in.

python/private/common.bzl
```diff
@@ -511,6 +514,37 @@ def target_platform_has_any_constraint(ctx, constraints):
...
+    for _ in range(n):
+        if from_parts[0] == to_parts[0]:
+            from_parts.pop(0)
+            to_parts.pop(0)
+        else:
+            break
+
+    # Impossible to compute a relative path without knowing what ".." is
+    if from_parts and from_parts[0] == "..":
+        fail("cannot compute relative path from '%s' to '%s'", from_, to)
```
aignas, line 543: TIL, that `fail` supports `%s`.

aignas (approved):

PR #3269, opened 2025-09-15
refactor: replace the magic sentinel string in the runtime env toolchain with a powershell wrapper

rickeylev:
This implements a Windows equivalent to the Mac/Linux wrapper script for the
runtime-env toolchain. This helps eliminate a legacy code path where a special
value in the toolchain (aptly named `_magic_pyruntime_sentinel_do_not_use`)
would trigger non-toolchain fallback behavior on Windows and look at the
`--python_path` Bazel builtin flag to figure out what Python to use.

This allows removal of the `--python_path` Bazel builtin flag.

### DIFF
--- a/python/private/runtime_env_toolchain.bzl
+++ b/python/private/runtime_env_toolchain.bzl
@@ -54,12 +54,9 @@ def define_runtime_env_toolchain(name):
         supports_build_time_venv = supports_build_time_venv,
     )
 
-    # This is a dummy runtime whose interpreter_path triggers the native rule
-    # logic to use the legacy behavior on Windows.
-    # TODO(#7844): Remove this target.
     py_runtime(
-        name = "_magic_sentinel_runtime",
-        interpreter_path = "/_magic_pyruntime_sentinel_do_not_use",
+        name = "_runtime_env_py3_runtime_windows",
+        interpreter = "//python/private:runtime_env_toolchain_interpreter.ps1",
         python_version = "PY3",
         visibility = ["//visibility:private"],
         tags = ["manual"],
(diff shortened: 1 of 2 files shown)


rickeylev:
Hrm. From CI:
```
LAUNCHER ERROR: Cannot launch process: "C:/b/bk-windows-s1tx/bazel/rules-python-python/python/private/runtime_env_toolchain_interpreter.ps1" C:\b\w5puohgx\execroot\_main\bazel-out\x64_windows-fastbuild-ST-1b13a798666d\bin\tests\runtime_env_toolchain\bootstrap_script_test.exe.runfiles\_main\tests\runtime_env_toolchain\bootstrap_script_test.zip
Reason: (error: 193): %1 is not a valid Win32 application.
```

Searching indicates this means that CreateProcess() can't be used with non-.exe programs. A powershell script would have to be invoked as e.g. `powershell foo.ps1`.

This _might_ require getting rid of that infernal windows launcher.

Though, maybe if powershell is set as the exe, and then the path to the interpreter is passed in...

rickeylev:
notes to self

The launcher logic is basically:

```
python = GetKey("python_bin_path")
if python_file := GetKey("python_file_short_path"):
  exec $python $python_file $@
elif GetKey("use_zip_file") == "1":
  exec $python ${0/.exe/.zip/} $@
else
  exec $python ${0/.exe//} $@
```

However, the python_file_short_path logic is 8.0 and higher only: https://github.com/bazelbuild/bazel/commit/ae818c43e58a7e32b02371f8683aa5ef454b2ff2

So, we could write powershell.exe to python_bin_path. However, we would then also have to change the stage1 bootstrap to a powershell script.

PR #3274, opened 2025-09-16
refactor(runtime setup): Add separate ABI3 Python headers target

nicholasjng:
Until now, we silently link extensions with both stable and unstable ABI libs, with the latter taking precedence in symbol resolution, because it appears first in the linker command AND, crucially, contains all CPython symbols present in the stable ABI library, thus overriding them.

This has the effect that stable ABI extensions on Windows are useable only with the Python distribution that they were built on. The now introduced separate ABI3 header target fixes this, and should be used for C++ extensions on Windows if stable ABI builds are requested.

Idea as formulated by `@dgrunwald-qt` in https://github.com/nicholasjng/nanobind-bazel/issues/72#issuecomment-3249959583.

-----------------------

This is motivated by https://github.com/nicholasjng/nanobind-bazel/issues/72, which I admittedly have procrastinated on, sorry about that.

This change shifts stable ABI selection on Windows to the extension developer, where it has arguably always been (they had to set the `Py_LIMITED_API` macro). I hope this isn't too much of an issue.  
An upside of this approach, though, is that with a separate target, the question "stable ABI or not" can be decided on an extension-by-extension basis, giving maximum flexibility to developers. This should not influence the wheel platform target, because a wheel is marked ABI3 if and only if all of its extensions are marked as ABI3.

Happy for comments.

### DIFF
--- a/python/private/hermetic_runtime_repo_setup.bzl
+++ b/python/private/hermetic_runtime_repo_setup.bzl
@@ -109,7 +109,26 @@ def define_hermetic_runtime_toolchain_impl(
     cc_library(
         name = "python_headers",
         deps = select({
-            "@bazel_tools//src/conditions:windows": [":interface", ":abi3_interface"],
+            "@bazel_tools//src/conditions:windows": [":interface"],
+            "//conditions:default": None,
+        }),
+        hdrs = [":includes"],
+        includes = [
+            "include",
+        ] + select({
+            _IS_FREETHREADED_YES: [
+                "include/python{major}.{minor}t".format(**version_dict),
+            ],
+            _IS_FREETHREADED_NO: [
+                "include/python{major}.{minor}".format(**version_dict),
+                "include/python{major}.{minor}m".format(**version_dict),
+            ],
+        }),
+    )
+    cc_library(
+        name = "python_headers_abi3",
+        deps = select({
+            "@bazel_tools//src/conditions:windows": [":abi3_interface"],
             "//conditions:default": None,
         }),
         hdrs = [":includes"],


nicholasjng:
@dgrunwald-qt, I would like to hear your opinion and your preferred manner of attribution - I referenced your discussion input in the commit message, but in backticks, so that you do not get flooded with e-mails for every merge commit. Let me know if that works for you.

python/private/hermetic_runtime_repo_setup.bzl
```diff
@@ -109,7 +109,26 @@ def define_hermetic_runtime_toolchain_impl(
...
+            ],
+        }),
+    )
+    cc_library(
+        name = "python_headers_abi3",
+        deps = select({
+            "@bazel_tools//src/conditions:windows": [":abi3_interface"],
             "//conditions:default": None,
         }),
         hdrs = [":includes"],
```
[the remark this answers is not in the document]
nicholasjng: The refactoring bit is fair, the other comment is obviously incorrect.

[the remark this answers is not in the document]
rickeylev: Yeah, that include expression is relatively large, so moving it to a variable SGTM.

rickeylev:
> with the abi-specific taking precedence in symbol resolution, because it appears first in the linker command

This sounds like something to fix? (Not necessarily in this PR). Being _forced_ to into the unstable ABI symbols seems strictly worse. IIUC, the stable API is a subset of the unstable, one, right? If so, lets file an issue for it, at the least.

rickeylev:
For attribution, I think you can put Co-Authored-By (or whatever that header tag is) into the PR description and it'll work as normal. The PR description becomes the commit description upon merge.

python/private/hermetic_runtime_repo_setup.bzl
```diff
@@ -109,7 +109,26 @@ def define_hermetic_runtime_toolchain_impl(
     cc_library(
         name = "python_headers",
         deps = select({
-            "@bazel_tools//src/conditions:windows": [":interface", ":abi3_interface"],
+            "@bazel_tools//src/conditions:windows": [":interface"],
+            "//conditions:default": None,
+        }),
+        hdrs = [":includes"],
```
rickeylev, line 115: I'm not entirely sure, but having the same header files in different targets _might_ result in an ODR violation. It definitely could for `srcs`; I can never remember how `hdrs` gets processed differently, though. I'll ask around.

python/private/hermetic_runtime_repo_setup.bzl
```diff
@@ -109,7 +109,26 @@ def define_hermetic_runtime_toolchain_impl(
...
+                "include/python{major}.{minor}t".format(**version_dict),
+            ],
+            _IS_FREETHREADED_NO: [
+                "include/python{major}.{minor}".format(**version_dict),
+                "include/python{major}.{minor}m".format(**version_dict),
+            ],
+        }),
+    )
+    cc_library(
+        name = "python_headers_abi3",
```
rickeylev: The new target also has to be wired into the toolchain. A convenience target should also be exposed.

python/private/hermetic_runtime_repo_setup.bzl
rickeylev, line 115: Are the headers actually identical for both stable and unstable ABIs? Or, ideally, should stable omit some subset of headers? I'm guessing this is transparently handled by ifdefs, but thought I'd ask for completness/edification.

rickeylev (commented):
LGTM in concept, but needs some cleanup and tests. I can help add some of this.
* Modify py_cc_toolchain to expose python_headers_abi3 as a separate piece
* Add `:current_py_cc_headers_abi3` convenience target.
* Add a basic test to verify things are wired together
* I don't remember if we have a functional test that actually builds and links things together. If so, add an appropriate test.

I read the discussion in https://github.com/nicholasjng/nanobind-bazel/issues/72

Yeah, thoughts from @dgrunwald-qt would be great. It sounds like he has some experience with Windows linking (I don't!). Couple quotes I'm gonna copy here for posterity:

> Normally extension module should be linked against only one of python312.lib/python3.lib.
python312.lib contains a strict superset of the symbols in python3.lib

TIL. If there's a doc or link that helps explain this stuff, that'd be great to reference.

> In our project we hacked rules_python to avoid the python_headers->interface dependencies altogether. Instead we manually add an interface or abi3_interface dependency to our individual extension module targets (with some of our modules using the stable abi3, and others not using that).

Interesting. Something I don't grok is why Windows needs any libs as part of a header's target. I mean, the point of headers is so you _don't_ need the actual implementation at build time. I guess this is some Windows-ism?

I now wonder if the `:libpython` targets have a similar problem. i.e. if users (including linux) of that are linking in the unstable ABI unexpectedly / unnecessarily.

> You could also try the hacky solution of specifying python3.lib first, then python312.lib. If defining Py_LIMITED_API, all symbols should be found in the first file so you should end up with a pure stable ABI extension module. But modules compiled without that flag will end up with a weird mixture of python3.dll and python312.dll references, depending on where the symbols could be found. This would be very unusual but probably works fine?
But referencing only one of the libraries is certainly the cleaner solution.

This answers my earlier question about fixing the order. So, probably worth doing as it'll help prevent issues, but not entirely correct.

rickeylev:
I've added most of the plumbing so the runtime's targets are available via toolchain resolution.

nicholasjng:
Thanks! There is an existing test for ABI3 linkage on Windows with a small test program. I'll adjust that to consume the current_py_cc_headers_abi3 target.

python/private/hermetic_runtime_repo_setup.bzl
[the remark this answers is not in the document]
rickeylev: Fixed.

python/private/hermetic_runtime_repo_setup.bzl
nicholasjng: I think so, yes. On that note, I'm not sure if the `{major}.{minor}m` paths actually exist on Windows - I downloaded some Windows assets from `python-build-standalone`, and none of them had any directories other than `include`.

rickeylev: Thanks for pointing that out. I field https://github.com/bazel-contrib/rules_python/issues/3281 for the m-suffix being defunct.

rickeylev:
Ok, did another batch of cleanup and prep.

I update the windows linkage test to depend on headers_abi3. CI isn't happy, though; says Python.h can't be found? Didn't expect that. The Bazel 9 Ubuntu CI job fails saying the include paths are missing, which sounds like the same sort of category of problem. Maybe it's due to the change to have one cc library depend on the other? And the includes paths aren't propagating?

I've hit my time limit on this PR, so will have to come back to it another time.

rickeylev:
Nice. CI is happy now. I also updated the linking test to run on all platforms.

Is there anything we should add to the linking test that would help verify linkage is correct?

rickeylev (approved):

nicholasjng:
Wow, thanks for pushing this so far! I did not expect this patch to become an affair of this magnitude.

Only thing I could think of is to link the ABI3 headers/libs into a program accessing the unlimited API, and then see if we crash, but I don't know if/how failure tests work in Bazel, and the `current_py_cc_libs` probably contains `libs/python3X.dll` as well, so such a test might not do what we want.

nicholasjng:
It might be good to confirm that python3X.dll is not referenced by an ABI3 extension (with the following taken verbatim from the nanobind_bazel issue above):

```
dumpbin /nologo /DEPENDENTS .\.venv\Lib\site-packages\nanobind_example\nanobind_example_ext.pyd

Dump of file .\.venv\Lib\site-packages\nanobind_example\nanobind_example_ext.pyd

File Type: DLL

  Image has the following dependencies:

    python312.dll
    MSVCP140.dll
    KERNEL32.dll
    VCRUNTIME140.dll
    VCRUNTIME140_1.dll
    api-ms-win-crt-heap-l1-1-0.dll
    api-ms-win-crt-runtime-l1-1-0.dll
    api-ms-win-crt-stdio-l1-1-0.dll
    api-ms-win-crt-string-l1-1-0.dll
    api-ms-win-crt-math-l1-1-0.dll

  Summary

        1000 .data
        1000 .pdata
        5000 .rdata
        1000 .reloc
        A000 .text
```

should read `python3.dll` with this (or a successful) patch only.

rickeylev:
Yeah, the existing test is more like an embedded python program, so it would need the full libs. A simple cc_binary that just depends on the headers, and shell test to grep it, should do the trick, though.

Unfortunately, negative testing in Bazel is nigh impossible.

rickeylev:
Well, that was a fun Windows CI debugging adventure. :melting_face: 

I've added a test that checks the dll depends on python3.dll, but not python3*.dll. (pefile module ftw)

In anycase, _finally_ good to go.

nicholasjng:
that's some dedication doing this on the weekend, much respect! happy to see this go in, sorry I wasn't more helpful in the testing part.

dgrunwald-qt:
> @dgrunwald-qt, I would like to hear your opinion and your preferred manner of attribution

Attribution: this works for me as-is.
PR: I haven't yet had time to test or look more closely.

> Something I don't grok is why Windows needs any libs as part of a header's target. I mean, the point of headers is so you don't need the actual implementation at build time. I guess this is some Windows-ism?

This is ultimately caused by how dynamic linking works: On Linux, ELF .so files have a list of library dependencies and a separate list of imported symbols. `['libc.so.6', 'libdl.so.2'], ['memcpy', 'dlopen', 'Py_Initialize']`. There's effectively one global symbol namespace in each process. This allows importing symbols without having to list which library they're from, as long as that library is loaded early enough.
On Windows, DLLs have a list of (library,symbol) tuples: `[('vcruntime140.dll', 'memcpy'), ('python312.dll', 'Py_Initialize')]`. There is no global symbol namespace; instead there's a separate namespace for each DLL. This means C headers do not contain sufficient information to import symbols on Windows.

This also means the "unresolved symbol" trick that python extension modules use on Linux+MacOS, is unavailable on Windows. Python extension modules must be linked against a python DLL.

Additionally, the Microsoft linker does not support linking directly against a DLL (unlike gcc/ld, which can directly link to `.so` files on linux -- and mingw's gcc can also directly against `.dll` on Windows, via its `dlltool`). Instead, the linker produces an "import library" (.lib file containing a list of (symbol_name,dll_name) tuples) together with the primary .dll output. Then any consumer of the library links against the .lib file at compile-time, whereas at runtime only the .dll file is used.
This split is actually nice in the context of bazel, since changes to a C function body in a library only change the .dll contents, the .lib remains unmodified, so bazel can skip relinking all .exes that use the library.

Now how this works for python is that normal extension modules link to e.g. `python312.dll`; but ABI3 extension modules would like to access the exact same python symbols (well, the stable subset) without having to know that DLL name. So Python additionally provides `python3.dll` for use in ABI3 extension modules. `python3.dll` does not contain any executable code, it just has [export forwarders](https://devblogs.microsoft.com/oldnewthing/20060719-24/?p=30473). So any request to `('python3.dll', 'Py_Initialize')` gets redirected to `('python312.dll', 'Py_Initialize')` by the Windows loader, and the function can be called as normal. (and of course on a different python version, the loader will pick up a different `python3.dll` which can redirect to a different real python dll)

rickeylev:
Thanks for explaining @dgrunwald-qt!

@nicholasjng FYI: In case you missed it, this PR also added feature detection for the new target. You can do the following in your rules to detect if the target exists:

```
load("@rules_python//python:features.bzl", "features")

cc_library(
  name = "foo",
  deps = [
    "@rules_python//python/cc:current_py_cc_headers_abi3"
    if getattr(features, "headers_abi3", False)
    else "@rules_python//python/cc:current_py_cc_headers"
  ]
```

nicholasjng:
Oh that's sweet. Thanks for letting me know! Can I also query the sysconfig extension suffix with that? (see: https://github.com/nicholasjng/nanobind-bazel/issues/61)

rickeylev:
You can compute it yes, though it's not directly available. To build the string `cpython-313-darwin.so`, do something like this example: https://rules-python.readthedocs.io/en/latest/howto/get-python-version.html

```
toolchain = ctx.toolchains["@rules_python//python:toolchain_type"]
runtime = toolchain.py3_runtime
filename = "{implementation}-{major}{minor}-{os}".format(
  implementation = runtime.implementation_name,
  major = runtime.interpreter_version_info.major,
  minor = runtime.interpreter_version_info.minor
  os = ctx.attr.os_name
)
```

For `ctx.attr.os_name`, have an attribute that is set via select, e.g. `os_name = select({"@platforms//os:osx": "darwin", "@platforms//os:linux": "linux", ...})`

nicholasjng:
I see. Since we're probing the context, that probably doesn't work in a macro though, right? Maybe it's finally time for me to start implementing a C extension rule.

PR #3277, opened 2025-09-18
feat(runfiles): support for --incompatible_compact_repo_mapping_manifest

### LINKED ISSUE #3022: Support incompatible_compact_repo_mapping_manifest in runfiles library
Under bzlmod, the repo mapping can become quite large (i.e. tens of megabytes) because its size scales as a factor of the number of repos in the transitive dependencies.

To address this, the `--incompatible_compact_repo_mapping_manifest` flag was introduced. This changes the repo mapping formation to use prefixes (instead of exact repo names) for mapping things.

To make this work with the runfiles library, the code has to be updated to handle these prefixes instead of just exact strings.

```
# old:
+deps+dep1,aaa,xxx
+deps+dep2,aaa,xxx

# new
+deps+*,aaa,xxx
```

Order matters (i.e. an earlier prefix should have precedence).

For reference, the names of these elements are `source_canonical,target_apparent,target_canonical` (i.e. "Within the canonical name `@source`, lookups for `@target_apparent` result in `@target_canonical`")

This looks relatively straight forward and the Java impl should translate over nicely. Dicts are insertion-ordered in Python as of Python 3.7 (long past EOL).

Core algo that needs to go into runfiles.py

```

def parse_repo_map(entries):
  exact = {}
  prefixed = {}
  for s, ta, tc in entries:
    if s.endswith("*"): prefixed[(s, ta)] = tc
    else: exact[(s, ta)] = tc
  return exact, prefixed

def rlocation(source, target):
   if (source, target) in exact: return exact[(source, target])
   for (prefix, ta), tc in prefixed.items():
     if ta == target and source.startswith(prefix): return tc
```

Refs:
* Bazel flag flip: https://github.com/bazelbuild/bazel/issues/26262
* Java impl: https://github.com/bazelbuild/rules_java/pull/301/files

cc @fmeum

jklukas:
Under bzlmod, the repo mapping can become quite large (i.e. tens of megabytes)
because its size scales as a factor of the number of repos in the transitive
dependencies.

To address this, the --incompatible_compact_repo_mapping_manifest flag was introduced.
This changes the repo mapping formation to use prefixes (instead of exact repo names)
for mapping things.

To make this work with the runfiles library, the code has to be updated to handle
these prefixes instead of just exact strings.

Fixes #3022.

### DIFF
--- a/python/runfiles/runfiles.py
+++ b/python/runfiles/runfiles.py
@@ -50,14 +50,13 @@ class _RepositoryMapping:
             prefixed_mappings: Dict mapping (source_prefix, target_apparent) -> target_canonical
         """
         self._exact_mappings = exact_mappings
-        self._prefixed_mappings = prefixed_mappings
 
         # Group prefixed mappings by target_apparent for faster lookups
         self._grouped_prefixed_mappings = defaultdict(list)
         for (
             prefix_source,
             target_app,
-        ), target_canonical in self._prefixed_mappings.items():
+        ), target_canonical in prefixed_mappings.items():
             self._grouped_prefixed_mappings[target_app].append(
                 (prefix_source, target_canonical)
             )
@@ -138,7 +137,7 @@ class _RepositoryMapping:
         Returns:
             True if there are no mappings, False otherwise
         """
-        return len(self._exact_mappings) == 0 and len(self._prefixed_mappings) == 0
+        return len(self._exact_mappings) == 0 and len(self._grouped_prefixed_mappings) == 0
 
 
 class _ManifestBased:
@@ -306,7 +305,8 @@ class Runfiles:
         # Split off the first path component, which contains the repository
         # name (apparent or canonical).
         target_repo, _, remainder = path.partition("/")
-        if not remainder or self._repo_mapping.lookup(source_repo, target_repo) is None:
+        target_canonical = self._repo_mapping.lookup(source_repo, target_repo)
+        if not remainder or target_canonical is None:
             # One of the following is the case:
             # - not using Bzlmod, so the repository mapping is empty and
             #   apparent and canonical repository names are the same
@@ -321,12 +321,10 @@ class Runfiles:
         ), "BUG: if the `source_repo` is None, we should never go past the `if` statement above"
 
         # Look up the target repository using the repository mapping
-        if source_repo is not None:
-            target_canonical = self._repo_mapping.lookup(source_repo, target_repo)
-            if target_canonical is not None:
-                return self._strategy.RlocationChecked(
-                    target_canonical + "/" + remainder
-                )
+        if target_canonical is not None:
+            return self._strategy.RlocationChecked(
+                target_canonical + "/" + remainder
+            )
 
         # No mapping found - assume target_repo is already canonical or
         # we're not using Bzlmod


jklukas:
cc @fmeum for review

python/runfiles/runfiles.py
```diff
@@ -188,25 +305,22 @@ def Rlocation(self, path: str, source_repo: Optional[str] = None) -> Optional[st
...
-            # - path did not contain a slash and referred to a root symlink,
-            #   which also should not be mapped.
+        if not remainder:
+            # path did not contain a slash and referred to a root symlink,
+            # which should not be mapped.
             return self._strategy.RlocationChecked(path)
 
-        assert (
-            source_repo is not None
-        ), "BUG: if the `source_repo` is None, we should never go past the `if` statement above"
```
jklukas, line 203: TODO: This assertion may still be important. I need to reassess above whether we need to match the previous conditional.

jklukas: I don't think an equivalent assertion is needed with the new logic, but I'm open to more thoughts on it.

CHANGELOG.md
```diff
@@ -92,6 +92,12 @@ END_UNRELEASED_TEMPLATE
 
 {#v0-0-0-added}
 ### Added
+* (runfiles) The Python runfiles library now supports Bazel's
+  `--incompatible_compact_repo_mapping_manifest` flag, which uses prefix-based
+  repository mappings to reduce memory usage for large dependency graphs under
+  bzlmod. This allows the repository mapping manifest to be significantly
+  smaller (from tens of megabytes to much less) while maintaining full
+  functionality.
```
```suggestion
* (runfiles) The Python runfiles library now supports Bazel's
  `--incompatible_compact_repo_mapping_manifest` flag.
```

python/runfiles/runfiles.py
```diff
@@ -15,12 +15,126 @@
...
+:::
 """
+import collections.abc
+from collections import defaultdict
 import inspect
 import os
 import posixpath
 import sys
-from typing import Dict, Optional, Tuple, Union
+from typing import Dict, Iterator, Optional, Tuple, Union
```
rickeylev: Please run our pre-commit command on this to format it. There's a lot of trailing whitespace lines (which, annoyingly, github UI doesn't show)

python/runfiles/runfiles.py
```diff
@@ -15,12 +15,126 @@
...
+                prefix = source_canonical[:-1]
+                prefixed_mappings[(prefix, target_apparent)] = target_canonical
+            else:
+                # This is an exact mapping
+                exact_mappings[(source_canonical, target_apparent)] = target_canonical
+
+        return _RepositoryMapping(exact_mappings, prefixed_mappings)
+
+    # Mapping protocol implementation
+    def __getitem__(self, key: Tuple[str, str]) -> str:
```
rickeylev: Please rename this to e.g. `get_canonical` or similar, just not `__getitem__`. This is an internal custom class, so there's not need to add the abstraction or complications of making it look dict-like.

python/runfiles/runfiles.py
```diff
@@ -15,12 +15,126 @@
...
+        # Try prefixed mapping if no exact match found
+        if target_apparent in self._grouped_prefixed_mappings:
+            for prefix_source, target_canonical in self._grouped_prefixed_mappings[target_apparent]:
+                if source_repo.startswith(prefix_source):
+                    return target_canonical
+
+        # No mapping found
+        raise KeyError(key)
+
+    def __iter__(self) -> Iterator[Tuple[str, str]]:
```
rickeylev: Please remove both `__iter__` and `__len__`. Same reason as before: no need to make it dict-like

tests/runfiles/runfiles_test.py
```diff
@@ -525,6 +525,259 @@ def testDirectoryBasedRlocationWithRepoMappingFromOtherRepo(self) -> None:
...
+
+            # Should match the general prefix for non-specific repos
+            self.assertEqual(
+                r.Rlocation("lib/foo/file", "deps+other_repo"),
+                dir + "/lib~general/foo/file",
+            )
+
+    def _createTestRepositoryMapping(self):
+        """Helper method to create a test RepositoryMapping with both exact and prefixed mappings."""
+        from python.runfiles.runfiles import _RepositoryMapping  # buildifier: disable=bzl-visibility
```
rickeylev: Please move this to the top of the file, with the other imports.

Also remove the bzl-visibility comment: it only applies to bzl code, not python.

tests/runfiles/runfiles_test.py
```diff
@@ -525,6 +525,259 @@ def testDirectoryBasedRlocationWithRepoMappingFromOtherRepo(self) -> None:
...
+            ("deps+", "external_dep"): "external_dep~prefix",
+            ("test_deps+", "test_lib"): "test_lib~2.1.0",
+        }
+        
+        return _RepositoryMapping(exact_mappings, prefixed_mappings)
+
+    def testRepositoryMappingLen(self) -> None:
+        """Test __len__ method - should count both exact and prefixed mappings."""
+        repo_mapping = self._createTestRepositoryMapping()
+        self.assertEqual(len(repo_mapping), 5)  # 3 exact + 2 prefixed
```
rickeylev: With `__len__` removed, this test can be removed.

tests/runfiles/runfiles_test.py
```diff
@@ -525,6 +525,259 @@ def testDirectoryBasedRlocationWithRepoMappingFromOtherRepo(self) -> None:
...
+        
+        # Test empty mapping
+        from python.runfiles.runfiles import _RepositoryMapping  # buildifier: disable=bzl-visibility
```
rickeylev: Same as before: move to top of file, remove buildifier comment.

tests/runfiles/runfiles_test.py
```diff
@@ -525,6 +525,259 @@ def testDirectoryBasedRlocationWithRepoMappingFromOtherRepo(self) -> None:
...
+        empty_mapping = _RepositoryMapping({}, {})
+        self.assertEqual(len(empty_mapping), 0)
+
+    def testRepositoryMappingGetItem(self) -> None:
```
rickeylev: Please remove this test or modify it to test esoteric cases that aren't easily tested through the public APIs.

tests/runfiles/runfiles_test.py
```diff
@@ -525,6 +525,259 @@ def testDirectoryBasedRlocationWithRepoMappingFromOtherRepo(self) -> None:
...
+        self.assertEqual(repo_mapping[("deps+some_repo", "external_dep")], "external_dep~prefix")
+        
+        # Test that exact takes precedence over prefix
+        self.assertEqual(repo_mapping[("deps+specific_repo", "external_dep")], "external_dep~exact")
+        
+        # Test KeyError for non-existent mapping (neither exact nor prefix match)
+        with self.assertRaises(KeyError):
+            _ = repo_mapping[("nonexistent", "repo")]
+
+    def testRepositoryMappingIter(self) -> None:
```
rickeylev: Remove this test. `__iter__` doesn't need to be implemented.

tests/runfiles/runfiles_test.py
```diff
@@ -525,6 +525,259 @@ def testDirectoryBasedRlocationWithRepoMappingFromOtherRepo(self) -> None:
...
+        first_three_keys = keys[:3]
+        last_two_keys = keys[3:]
+        
+        for exact_key in exact_keys:
+            self.assertIn(exact_key, first_three_keys)
+        
+        for prefixed_key in prefixed_keys:
+            self.assertIn(prefixed_key, last_two_keys)
+
+    def testRepositoryMappingContains(self) -> None:
```
rickeylev: Remove this test. The underlying class doesn't need to support contains checking.

tests/runfiles/runfiles_test.py
```diff
@@ -525,6 +525,259 @@ def testDirectoryBasedRlocationWithRepoMappingFromOtherRepo(self) -> None:
...
+        self.assertTrue(("", "my_workspace") in repo_mapping)
+        self.assertTrue(("deps+specific_repo", "external_dep") in repo_mapping)
+        
+        # Test prefix-based lookups (consistent with __getitem__)
+        self.assertTrue(("deps+some_repo", "external_dep") in repo_mapping)
+        
+        # Test non-existent mapping
+        self.assertFalse(("nonexistent", "repo") in repo_mapping)
+
+    def testRepositoryMappingGet(self) -> None:
```
rickeylev: Remove this test, same as with getitem

tests/runfiles/runfiles_test.py
```diff
@@ -525,6 +525,259 @@ def testDirectoryBasedRlocationWithRepoMappingFromOtherRepo(self) -> None:
...
+        self.assertEqual(repo_mapping.get(("deps+specific_repo", "external_dep")), "external_dep~exact")
+        
+        # Test prefix-based lookups
+        self.assertEqual(repo_mapping.get(("deps+some_repo", "external_dep")), "external_dep~prefix")
+        
+        # Test non-existent mapping
+        self.assertIsNone(repo_mapping.get(("nonexistent", "repo")))
+        self.assertEqual(repo_mapping.get(("nonexistent", "repo"), "default"), "default")
+
+    def testRepositoryMappingKeysValuesItems(self) -> None:
```
rickeylev: Remove this test; doesn't need to be dict-like

tests/runfiles/runfiles_test.py
```diff
@@ -525,6 +525,259 @@ def testDirectoryBasedRlocationWithRepoMappingFromOtherRepo(self) -> None:
...
+        self.assertIn(("", "my_workspace"), keys)
+        self.assertIn(("deps+", "external_dep"), keys)
+        
+        # Test values()
+        values = list(repo_mapping.values())
+        self.assertEqual(len(values), 5)  # 3 exact + 2 prefixed
+        self.assertIn("_main", values)
+        self.assertIn("external_dep~prefix", values)
+
+    def testRepositoryMappingTruthiness(self) -> None:
```
rickeylev: Remove this test: doesn't need dict semantics

tests/runfiles/runfiles_test.py
```diff
@@ -525,6 +525,259 @@ def testDirectoryBasedRlocationWithRepoMappingFromOtherRepo(self) -> None:
...
+        self.assertTrue(repo_mapping)
+        self.assertTrue(bool(repo_mapping))
+        
+        # Test empty mapping is falsy
+        from python.runfiles.runfiles import _RepositoryMapping  # buildifier: disable=bzl-visibility
+        empty_mapping = _RepositoryMapping({}, {})
+        self.assertFalse(empty_mapping)
+        self.assertFalse(bool(empty_mapping))
+
+    # TODO: Add manifest-based test for compact repo mapping
```
rickeylev: I'm not sure what this is referring to? The manifest and repo mapping files are different.

rickeylev (changes requested):
I was going to push some changes, but noticed that "maintainer can modify" isn't set.

Mostly LGTM, but please remove all the Mapping API compatibility code. As an internal class, there's no need for it to support the wide range of dict operations.

python/runfiles/runfiles.py
rickeylev: Please keep the assert. I'm not entirely sure if we have test coverage of all the cases.

Looking at the code, I don't see why the comment wouldn't still apply. source_repo can end up None here if the repo mapping is empty. (which, according to comments, occurs for workspace mode)

rickeylev (changes requested):

python/runfiles/runfiles.py
jklukas: Restored.

jklukas:
Thank you for the quick review! I believe all comments are now addressed. I removed the `Mapping` compat; now we just have a `lookup` method and an `is_empty` method as needed for the rest of the logic.

python/runfiles/runfiles.py
```diff
@@ -15,14 +15,132 @@
...
+        for line in content.splitlines():
+            source_canonical, target_apparent, target_canonical = line.split(",")
+            if source_canonical.endswith("*"):
+                # This is a prefixed mapping - remove the '*' for prefix matching
+                prefix = source_canonical[:-1]
+                prefixed_mappings[(prefix, target_apparent)] = target_canonical
+            else:
+                # This is an exact mapping
+                exact_mappings[(source_canonical, target_apparent)] = target_canonical
+
```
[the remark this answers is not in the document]
jklukas: It may be best to error out anyway if repo mappings diverged from expected format in the ways described here.

jklukas:
Gentle ping for reviewers to take another look after pushing fixes.

python/runfiles/runfiles.py
```diff
@@ -15,14 +15,132 @@
...
+        prefixed_mappings: Dict[Tuple[str, str], str],
+    ) -> None:
+        """Initialize repository mapping with exact and prefixed mappings.
+
+        Args:
+            exact_mappings: Dict mapping (source_canonical, target_apparent) -> target_canonical
+            prefixed_mappings: Dict mapping (source_prefix, target_apparent) -> target_canonical
+        """
+        self._exact_mappings = exact_mappings
+        self._prefixed_mappings = prefixed_mappings
```
fmeum: This doesn't have to be retained, the grouped dict suffices for lookups and `is_empty`

python/runfiles/runfiles.py
```diff
@@ -188,7 +306,7 @@ def Rlocation(self, path: str, source_repo: Optional[str] = None) -> Optional[st
         # Split off the first path component, which contains the repository
         # name (apparent or canonical).
         target_repo, _, remainder = path.partition("/")
-        if not remainder or (source_repo, target_repo) not in self._repo_mapping:
+        if not remainder or self._repo_mapping.lookup(source_repo, target_repo) is None:
```
fmeum: The result of the lookup should be stored in a variable so that the lookup doesn't have to be repeated below.

jklukas: I also removed the redundant `if source_repo is not None` since this happens immediately after we assert that condition.

python/runfiles/runfiles.py
jklukas: Done

python/runfiles/runfiles.py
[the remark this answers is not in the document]
rickeylev: Agreed. We can ignore this warning for now.

rickeylev:
Thanks for implementing this!

jklukas:
Thanks for the review and getting this landed!

PR #3279, opened 2025-09-19
feat: support per-wheel enable_implicit_namespace_pkgs in whl_mods (WIP)

chrisirhc:
Update: I was [reading a thread](https://bazelbuild.slack.com/archives/CA306CEV6/p1751471945163239) in `#python` and saw that this set to True when setting `--incompatible_default_to_explicit_init_py` is actually desirable and forward-looking.

So perhaps this PR is unnecessary for my use case, unless somebody runs into issues and needs to toggle it for some packages/wheels.

Other note: perhaps the `enable_implicit_namespace_pkgs` documentation can be further clarified.

---

As part of investigating https://github.com/aspect-build/rules_py/issues/611 which requires `enable_implicit_namespace_pkgs = True` as a workaround to unblock PEX creation, I looked into how to only toggle on this flag per wheel, as per https://github.com/aspect-build/rules_py/issues/611#issuecomment-3305979729 .

The goal of this is to reduce reliance on this flag where it's not necessary.

This PR is a proof-of-concept to gather if there's any interest in this feature. The commits aren't cleaned up yet.

Not yet done:
- [ ] Update CHANGELOG.md
- [ ] Update documentation for whl_mods

### DIFF
--- a/python/private/pypi/extension.bzl
+++ b/python/private/pypi/extension.bzl
@@ -49,6 +49,7 @@ def _whl_mods_impl(whl_mods_dict):
                 data = mods.data,
                 data_exclude_glob = mods.data_exclude_glob,
                 srcs_exclude_glob = mods.srcs_exclude_glob,
+                enable_implicit_namespace_pkgs = mods.enable_implicit_namespace_pkgs,
             ))
 
         _whl_mods_repo(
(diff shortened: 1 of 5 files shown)


kwlzn:
any plans to clean this up and land it? it would be helpful to have this capability in the ruleset.

aignas:
Since this has gone stale, let me close it. However PRs are welcome to support it.

PR #3291, opened 2025-09-23
fix: Mark uv extension as reproducible

fmeum:
Reduces the amount of noise in Bazel's default lockfile.


aignas:
This extension would also benefit from the facts feature in bazel. We want to store the URLs for the tool. Right now what is the example noise that this eliminates?

fmeum:
Oh, I didn't notice the download. Let's wait for facts then. 

I'm trying to clean up all the extension entries in the lockfile of an entirely empty Bazel project. I would expect new users to be quite surprised if they find references to (current status) uv, Kotlin and pybind in there without testing why of those.

PR #3295, opened 2025-09-26
feat(python): Support overriding from env vars for `single_version_platform_override`

meteorcloudy:
- Similar to `python.defaults`'s `python_version_env` attribute, we add `python_version_env`, `sha256_env`, `strip_prefix_env`, `url_env` for corresponding attributes of `single_version_platform_override`, allowing reading those values from env vars.
- Allow the python version specified to be a minor version, which will be mapped to a full version with `MINOR_MAPPING` from `//python:versions.bzl`.

This feature will make sure JAX's HERMETIC_PYTHON_* env vars keep working after migrating to Bzlmod.
https://docs.jax.dev/en/latest/developer.html#custom-hermetic-python-examples

### DIFF
--- a/python/private/python.bzl
+++ b/python/private/python.bzl
@@ -1331,7 +1331,7 @@ Arbitrary platform strings allowed.
         ),
         "python_version": attr.string(
             mandatory = False,
-            doc = "The python version to override URLs for. Must be in `X.Y.Z` format.",
+            doc = "The python version to override URLs for. Must be in `X.Y.Z` or `X.Y` format.",
         ),
         "python_version_env": attr.string(
             mandatory = False,


rickeylev:
I'm pretty mixed on this. I'd like to talk in https://github.com/bazel-contrib/rules_python/issues/3293 more to try and figure out how to glue things together better.

As is, what this implies is all the APIs we have that accept some sort of version, url, etc, arg should have a `_env` and `_target` variations of them. (so the pip, python, and uv bzlmod extensions, which encompasses a very large api surface).

Having `python_version_env` for `single_version_platform_override` is also a bit weird -- the API name means it applies to a specific version. Getting the version from an environment variable means its for an arbitrary version.

meteorcloudy:
Yeah, I agree this is more of an API change that specifically address JAX/XLA's needs, happy to wait for a more proper solution.

PR #3301, opened 2025-09-28
chore: use python.defaults to set rules_python default python version

rickeylev:
python.defaults is the modern way to set the default.

Setting it this way also helps avoid a bug where if a root module has a single `python.toolchain()` call (which are implicitly treated as `is_default=True`) and also sets the default using `python.defaults()`, some validation logic gives an error about using both ways to set a default.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -37,13 +37,15 @@ use_repo(
 # We need to do another use_extension call to expose the "pythons_hub"
 # repo.
 python = use_extension("//python/extensions:python.bzl", "python")
+python.defaults(
+    python_version = "3.11",
+)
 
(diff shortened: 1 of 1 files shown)


aignas (approved):

rickeylev:
Hrm. Actually, calling python.defaults() in rules_python would also trigger the bug if the consuming module has a single toolchain call. Fixing this will be a bit more involved.

rickeylev:
Fixing that wasn't so bad.

I moved all the "figure out the default python version" to a separate function. It checks python.defaults, and if that doesn't give an answer, then checks `python.toolchain`.

Also added some tests for some of these cases. That mock module stuff is pretty spiffy!

PR #3304, opened 2025-09-29
feat(toolchains): let local toolchains point to a label

rickeylev:
Currently, the local toolchain code requires using a path (or program name) to find
the Python interpreter. This comes up short when using Bazel to
download an arbitrary runtime (or otherwise manage the creation of it, e.g.
downloading Python and building it from source in a repo rule). In such cases, the
file system location of the interpreter isn't known (it'll be in some Bazel cache
directory).

To fix, add the `interpreter_target` attribute to `local_runtime_repo`, which it
looks up the path for, then continues on as normal. As an example, the test uses
a custom repository rule to download a particular version of Python appropriate
to the OS.


python/private/local_runtime_repo.bzl
```diff
@@ -200,13 +200,37 @@ a system having the necessary Python installed.
...
+:::{seealso}
+The {obj}`interpreter_target` attribute for getting the interpreter from
+a label
+:::
+""",
+            default = "",
+        ),
+        "interpreter_target": attr.label(
+            doc = """
+A label to a Python interpreter executable.
```
aignas, line 222: FYI, this looks a little weird - `local_runtime_repo` is for interpreters that we get from the local system, whereas this tells me that we just can feed it any non-local interpreter as well. It is something in between of hermetic and non hermetic one.

tests/integration/local_toolchains/pbs_archive.bzl
```diff
@@ -0,0 +1,53 @@
+"""A repository rule to download and extract a Python runtime archive."""
```
aignas, line 1: What is pbs?

python/private/local_runtime_repo.bzl
aignas: Read the remainder, it has to be something that is materialized in the repository_rule phase? I.e. it cannot be an artifact from a build action?

rickeylev: Correct. local_runtime_repo runs the provided interpreter to get information about the python install, which it then uses as a platform runtime (i.e. sets py_runtime.interpreter_path, an absolute path to the interpreter)

I was really just meaning to change the test so it didn't rely on whatever the local system had, while still being a local installation. Hence why it does a very simple download and extract of a python-build-standalone runtime (to better replicate its just a directory with files, not something actually setup for Bazel).

But, then I realized if you can feed it an arbitrary repo, then its actually pretty useful as a general feature. Namely, you could build python from source in a repo rule, then point to it directly. Or just have a repo rule that does its own copying/symlinking to some runtime of your choice and feed it into local_runtime_repo, which then handles all the introspection and toolchain generation.

tests/integration/local_toolchains/pbs_archive.bzl
rickeylev: "python build standalone".

aignas (approved):

PR #3310, opened 2025-10-01
fix(toolchains): add musl list of freethreaded runtimes (workspace)

rickeylev:
The musl entry got lost as part of a refactoring. Add it back to the platforms
that generate freethreaded variants.

Fixes https://github.com/bazel-contrib/rules_python/issues/3286

### DIFF
--- a/python/versions.bzl
+++ b/python/versions.bzl
@@ -1060,6 +1060,7 @@ def get_release_info(platform, python_version, base_url = DEFAULT_RELEASE_BASE_U
                     "x86_64-apple-darwin": "pgo+lto",
                     "x86_64-pc-windows-msvc": "pgo",
                     "x86_64-unknown-linux-gnu": "pgo+lto",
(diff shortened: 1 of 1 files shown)


rickeylev:
While creating tests for this, I noticed some weird things with toolchain registration. I think there's a bug in the order that python_register_toolchains() registers things for workspace _and_ in the toolchain constraints, the net effect being you can never match the musl or freethreaded toolchains (I think Yulia filed a bug for this, too).

I think this fix also only affects workspace -- bzlmod is using what's in MODULE.bazel, which I think is also omitting musl. I don't see a musl-freethreaded toolchain showing up there

aignas (approved):

rickeylev:
Ahh, I think I see -- we don't have a musl-freethreaded build built into our versions.bzl file. There are such builds available. JAX must be patching versions.bzl and adding such an entry themselves (or equivalent).

Ugh, versions.bzl is such a mess.

Issue #2937, opened 2025-05-28
`rules_python_publish_deps` causing pip extension resolution to always require network access


meteorcloudy:
# 🐞 bug report

### Affected Rule

https://github.com/bazel-contrib/rules_python/blob/b40d96aba36d675c60b03424aa22f31c09e0ea4f/MODULE.bazel#L63-L75

### Is this a regression?

Yes, after https://github.com/bazel-contrib/rules_python/commit/b9b0948234216cf43bb898eec078b1c71a37c1f9, moving from `pip_internal` to `pip`.

### Description

The `rules_python_publish_deps` python hub uses the same module extension (`pip`) end users use, which causes the module extension resolution to always require network access. In offline build without an up-to-date lockfile, the module extension resolution will fail even though `rules_python_publish_deps` is never required by the end user.

This was discovered in https://github.com/bazelbuild/bazel/pull/26163, where upgrading rules_python broke Bazel's bootstrap build.

## 🔥 Exception or Error

<pre><code>

[1A[K[31m[1mERROR: [0m/tmp/bazel_8RGzY3p4/out/external/rules_python+/python/private/pypi/simpleapi_download.bzl:130:14: Traceback (most recent call last):
	File "/tmp/bazel_8RGzY3p4/out/external/rules_python+/python/private/pypi/extension.bzl", line 615, column 25, in _pip_impl
		mods = parse_modules(module_ctx)
	File "/tmp/bazel_8RGzY3p4/out/external/rules_python+/python/private/pypi/extension.bzl", line 490, column 36, in parse_modules
		out = _create_whl_repos(
	File "/tmp/bazel_8RGzY3p4/out/external/rules_python+/python/private/pypi/extension.bzl", line 155, column 50, in _create_whl_repos
		requirements_by_platform = parse_requirements(
	File "/tmp/bazel_8RGzY3p4/out/external/rules_python+/python/private/pypi/parse_requirements.bzl", line 180, column 36, in parse_requirements
		index_urls = get_index_urls(
	File "/tmp/bazel_8RGzY3p4/out/external/rules_python+/python/private/pypi/extension.bzl", line 470, column 79, in lambda
		get_index_urls = lambda ctx, distributions: simpleapi_download(
	File "/tmp/bazel_8RGzY3p4/out/external/rules_python+/python/private/pypi/simpleapi_download.bzl", line 130, column 14, in simpleapi_download
		_fail("Failed to download metadata for {} for from urls: {}".format(
[...]

PR #3311, opened 2025-10-01
fix(pip): do not use experimental_index_url for publish_deps

aignas:
Currently, experimental_index_url will trigger network access (by the bzlmod extension
logic) as it figures out the wheels. This is problematic because this network
access always occurs, even if a user isn't using the publish tool. Additionally,
failed network accesses can get Bazel into stuck state requiring `bazel shutdown`
to fix it.

To fix, don't  use experimental_index_url for the publish tool deps.

Once bazelbuild/bazel#24777 is merged, we can roll it back,
but then we may need to check-in the `MODULE.bazel.lock` file into git.

This should be no-op for most users because the publisher is usually run
from with `host == exec`, but there is a small chance that people will
hit #2241.

Work towards #2937

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,6 +69,8 @@ END_UNRELEASED_TEMPLATE
 * (bootstrap) For {obj}`--bootstrap_impl=system_python`, the sys.path order has
   changed from `[app paths, stdlib, runtime site-packages]` to `[stdlib, app
   paths, runtime site-packages]`.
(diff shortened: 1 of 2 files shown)


rickeylev:
Running the publisher as a build action doesn't make sense -- a build action isn't appropriate for e.g. uploading release artifacts. So I don't think we have to worry about that.

rickeylev:
I modified the PR description to better describe things ("This" wasn't clear if it was referring to before or after state). Approved; but check the PR description to make sure I understood correctly

PR #3312, opened 2025-10-02
feat/fix(pip): enable local paths for experimental_index_url

adrianimboden:
Hi rules_python team

I use your rules for a long time now. With WORKSPACE style, I use pip as follows:
```
pip_parse(
    name = "py_deps",
    extra_pip_args = [
        "--index-url=/home/user/local_pip_mirror",
        "--no-cache-dir",
    ],
    python_interpreter = "python3",
    requirements_lock = ":requirements_lock.txt",
)
```

The folder `/home/user/local_pip_mirror` gets populated with `pypi-mirror download --requirement requirements_lock.txt`

So I have a nice and clean way to use offline build. I could never get it running directly with `bazel fetch` and stuff. But this solution was very nice because no internet was involved at all.

I am in the way of upgrading to bzlmod. I saw many bug reports for making offline build work, but I honestly gave up with `bazel vendor` and stuff again. The simplest way in my opinion is to just use a local pip mirror. For that I tried the following:

```
pip.parse(         
    experimental_index_url = "/home/thingdust/deps/pip_deps",
    hub_name = "py_deps",
)                        
use_repo(pip, "py_deps")
```

at the moment, `experimental_index_url` must be a https:// url to work.

This MR changes it so that local paths are also a possible `experimental_index_url`.

For my project, the proposed changes are in effect and working great.

It is not perfect, but I think it is an important addition to aid the bzlmod migration.


python/private/pypi/whl_library.bzl
```diff
@@ -295,12 +295,29 @@ def _whl_library_impl(rctx):
...
         urls = rctx.attr.urls
-        result = rctx.download(
-            url = urls,
-            output = filename,
-            sha256 = rctx.attr.sha256,
-            auth = get_auth(rctx, urls),
-        )
+        if len(urls) == 1 and not urls[0].startswith("https://"):
+            local_url = urls[0]
+            local_path = rctx.path(local_url)
```
aignas: Ideally, the `whl_library` should consume locally available paths through labels, because then it makes things easier to thread through and `whl_library` already accepts a label. Would it be possible to create a `repository_rule` that exposes these paths as labels?

aignas:
This looks interesting - we certainly have discussed about using local path for the `experimental_index_url` in the past. A few things that pop to my mind:
* Having a unit test that ensures that this feature works would be great. They can be added in `tests/pypi/simpleapi_download` directory.

How do you update the local mirror?

adrianimboden:
I populate the mirror like this:
```
pypi-mirror download --requirement requirements_lock.txt --download-dir /tmp/download
pypi-mirror create --download-dir /tmp/download --mirror-dir /path/to/mirror --copy
```

The folder looks like this then:
```
├── index.html
├── aiohappyeyeballs
│   ├── aiohappyeyeballs-2.6.1-py3-none-any.whl
│   └── index.html
├── aiohttp
│   ├── aiohttp-3.12.15-cp311-cp311-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
│   └── index.html
├── aiosignal
│   ├── aiosignal-1.4.0-py3-none-any.whl
│   └── index.html
```

I am not sure about the label stuff. Did you think about something like this?
```
new_local_repository = use_repo_rule("@bazel_tools//tools/build_defs/repo:local.bzl", "new_local_repository")

new_local_repository(
    name = "pip_deps_mirror",
    build_file_content = "exports_files(['**'])",
    path = "/home/thingdust/deps/pip_deps",
)

pip.parse(
    experimental_index_url = Label("@pip_deps_mirror"),
    hub_name = "py_deps",
)
use_repo(pip, "py_deps")
```

I am not sure how easy that will be. Seems like a complete new codepath to me when there may be an URL or a label. Or do I miss something?

adrianimboden:
after looking at it again, I saw that there is a much simpler solution to make it better in the meantime.

I wrongly assumed that the download functions don't work for local files. I always had the problem that file:// urls did not work. I found out that this is because the urls get normalized first. A small addition to `strip_empty_path_segments` makes it work with local urls.

Making it work with labels would be nice tough. Probably for another time?

adrianimboden:
It did not work before because strip_empty_path_segments does the following: `file:///path/to/folder` -> `file://path/to/folder`. For local paths, empty segments should not be a problem I presume.

rickeylev:
I really like the idea of being able to point to a local path using a label, for several reasons.

It'd be really convenient for testing our pip integration -- we can easily construct arbitrary index states and have a more end-to-end verification.

It also seems like a really flexible and powerful way for customizing where pip is getting stuff from. You could write a repo rule to make the pip index look however you want, and be populated however you want.

groodt:
Yes, this sort of thing is great. It's often called a "wheel house" and is a very common and useful pattern for offline builds, avoiding sdist in deployment scenarios, etc. Very supportive of this. Tools in a similarish space are: https://github.com/chriskuehl/dumb-pypi

aignas:
Thinking out loud a little bit how this could be designed. This might be a train of thought but I'll just right it out as I think.

1. The idea of passing in a local path or something sounds good, but so far in bazel I've seen this work only if you pass an absolute path or a label. Hence I thought it would be nice to pass a label.
2. If one has labels for each whl file, then we can pass them to `whl_library` `whl_file` attribute: https://rules-python.readthedocs.io/en/latest/api/rules_python/python/private/pypi/whl_library.html#whl_library.whl_file
3. This means that the code in [parse_requirements.bzl](https://github.com/bazel-contrib/rules_python/blob/79f654686646e2ab213741bc4774ea9d69895b54/python/private/pypi/parse_requirements.bzl#L347) needs to inject those labels in some way.
4. In the future we may want to write the URLs into the lock file, so if they have absolute `file:///` in them, this will not age well, so it is best to treat the local index as one that has the right format.
5. `parse_requirements.bzl` is called from `hub_builder` and gets the `get_index_urls` function as a parameter. We could have a separate implementation of that that returns labels instead of URLs, however, the label mapping should be present there is some way.
6. If `pip.parse` can create a local index repository on the fly (i.e. repository where we can access whls by using a scheme of `@local_index_repo_name//<whl_name>:<file_name>.whl`). The extension reads the local directory structure and finds all whl files, then creates a repo and passes the whls as a list of paths/labels. The HTML files are only processed in the extension to avoid the circular dependencies in the extension/starlark evaluation.

So to sum up, the files that would need to be touched:
* `whl_library` - stays the same.
* `hub_builder.bzl` - needs some extra handling of a different `get_index_urls` function. It should handle the case well where the `whl` (or `dist`) struct has `whl_file` but does not have `url` set.
* `parse_requirements.bzl` - needs some minor fixing to accommodate a more generic getting of the wheels.
* `local_whl_repo.bzl` - a new repository that contains the files.
* `simpleapi_local.bzl` - a new file that handles the traversing the local index.html tree.

There are probably ways to optimize this approach.

PR #3321, opened 2025-10-04
poc: singular pypi repo for late binding external dependencies

rickeylev:
POC for having a single `@pypi` repo so that targets can depend on something without
having to choose a particular dependency closure (pip.parse hub_name).

The current impl works in a hub-of-hubs fashion, however, I think that indirection
could be removed, since it's known at bzlmod time the spokes that a pip.parse hub points to.


aignas (commented):
I like this API, this is reasonably easy to understand and is probably easier than doing transitions or anything else that would resolve/override a particular dependency from the hub.

aignas:
This got implemented in #3837

Issue #2951, opened 2025-06-02
Read index urls from the requirements files and instead have a flag


aignas:
Summary from https://github.com/bazel-contrib/rules_python/discussions/2564

```starlark
pip.parse(
    experimental_bazel_downloader = True,
    experimental_index_url_overrides = {"package": "https//baz.com/simple", ...},
    requirements_lock = "requirements.txt",
    ...
```

If `requirements.txt` has:

```
--index_url https://foo/simple
--extra-index-url https://bar/simple
--extra-index-url https://foobar/simple

package==1.2.3 \
    --hash=sha256:abcd1234
```

The `experimental_bazel_downloader = True` API would be functionally equivalent to:

```starlark
pip.parse(
    experimental_index_url = "https://foo/simple",
    experimental_extra_index_urls = ["https://bar/simple", "https://foobar/simple"]
    experimental_index_url_overrides = {"package": "https//baz.com/simple", ...},  # note no "oauth2accesstoken"
    requirements_lock = "requirements.txt",
    ...
```

The net result of such a change is that the `requirements.in` file is now the canonical source for index URLs no matter how the lock file is made:

```mermaid
flowchart TD
    A[req.in] --> B[pip compile<br>uv pip compile]
    A --> C[compile_pip_requirements]
    B --> D[req.lock]
    C --> D
    D --> |effectively sets| E[pip.parse.experimental_index_url<br>pip.parse.experimental_extra_index_urls]
```

### Proposal summary:

1. Remove `experimental_index_url`, `experimental_extra_index_urls`
2. Add boolean `experimental_bazel_downloader` (or similar wording)
3. Have `pip.parse` pull index URLs from the requirements lock file. If not present, default to public PyPI.

I think keeping `experimental_index_url_overrides` is prudent.

#### Alternative:

1. Keep `experimental_index_url`, `experimental_extra_index_urls` but support special options like `DEFER` or similar wording that causes `pip.parse` to pull index URLs from the requirements lock file.

_Originally posted by @dougthor42 in https://github.com/bazel-contrib/rules_python/discussions/2564#discussioncomment-13331380_

martis42:
If a repository is used in different closed off environments, it is possible that each environment uses a different pypi index address (aka the pypi mirror of this environment). Right now this is trivial to model due to 
```starlark
pip-parse(
    hub_name = "some_pip_hub",
    envsubst = ["PIP_INDEX_URL"],
[...]

PR #3322, opened 2025-10-05
feat(pip): global downloader settings

aignas:
Initial API proposal for a stable API for the `experimental_index_url`.

TODO for this PR:
- [ ] Agree on the interface and flow - it should be described either as a
  sketch in docs in how to or in the attributes.
- [ ] Implement
- [ ] Deprecation messages for the `experimental_*index*` flags

WIP

#2951

### DIFF
--- a/python/private/pypi/extension.bzl
+++ b/python/private/pypi/extension.bzl
@@ -115,13 +115,10 @@ def build_config(
                 # extra values that we just add
                 auth_patterns = tag.auth_patterns,
                 netrc = tag.netrc,
-                use_downloader = tag.user_downloader,
-                parallel_download = tag.parallel_download,
-                index_url_overrides = tag.index_url_overrides,
-                downloader_skip = tag.downloader_skip,
-                # TODO @aignas 2025-05-19: add more attr groups:
-                # * for index/downloader config. This includes all of those attributes for
-                # overrides, etc. Index overrides per platform could be also used here.
+                # TODO @aignas 2025-10-05: is this enough?
+                # downloader = tag.downloader,
+                # index_url = tag.index_url,
+                # index_overrides = tag.index_overrides,
             )
 
     return struct(


aignas:
@dougthor42, could you please comment on how you imagine the API work for the downloader configuration?

python/private/pypi/extension.bzl
```diff
@@ -525,7 +526,72 @@ supported from this version without extra handling from the user.
...
+:::
+""",
+    ),
+    "index_overrides": attr.string_dict(
+        doc = """\
+The index URL overrides for each package to use for downloading wheels using
+bazel downloader. This value is going to be subject to `envsubst` substitutions
+if necessary.
+
+The key is the package name (will be normalized before usage) and the value is the
```
dougthor42, line 557: > will be normalized before usage

Clarification: This means that user-provided keys `FrIeNdLy-._.-bArD` and `friendly-bard` will be treated the same (`friendly-bard`), correct? What happens if a post-normalization duplicate key is found?

python/private/pypi/extension.bzl
```diff
@@ -525,7 +526,72 @@ supported from this version without extra handling from the user.
...
+index URL.
+
+This design pattern has been chosen in order to be fully deterministic about which
+packages come from which source. We want to avoid issues similar to what happened in
+https://pytorch.org/blog/compromised-nightly-dependency/.
+
+The indexes must support Simple API as described here:
+<https://packaging.python.org/en/latest/specifications/simple-repository-api/>
+
+If `skip` is used as a value, then we will not use the downloader for a particular package. The
```
dougthor42, line 567: This is interesting to me. Can you elaborate on what problem it's trying to solve?

I think I like the idea, but I'm concerned that it'll cause hard-to-debug issues (such as having correct auth for Bazel downloader but not for the non-Bazel downloader case, or vice-versa).

dougthor42, line 567: Nit:

```suggestion
If `skip` is used as a value, then we will not use the Bazel downloader for a particular package. The
```

python/private/pypi/extension.bzl
```diff
@@ -525,7 +526,72 @@ supported from this version without extra handling from the user.
...
+`index_overrides`. If the package is not found in the index, we will search all of the indexes
+listed in the lock files used to create a particular hub repository.
+
+TODO: experimental and docs in flux.
+
+:::{versionadded} VERSION_NEXT_FEATURE
+:::
+""",
+    ),
+    "index_overrides": attr.string_dict(
```
dougthor42, line 551: Have you thought about cases where packages of a particular version must be downloaded from index A, while others versions must be downloaded from index B?

```
index_overrides = {
    "friendly-bard<=1.0.0,>3.0": INDEX_A,
    "friendly-bard": INDEX_B,
}
```

Or where windows versions are on a different index from linux versions?

TBH I'm not sure how useful such a thing would be, but I could see edge cases requiring it such as when a private index isn't yet updated with the latest version of a package but users must absolutely have the latest version.

python/private/pypi/extension.bzl
```diff
@@ -525,7 +526,72 @@ supported from this version without extra handling from the user.
...
+undefined.
+
+TODO: experimental
+TODO: implement skip
+
+:::{versionadded} VERSION_NEXT_FEATURE
+:::
+""",
+    ),
+    "index_url": attr.string(
```
dougthor42, line 578: Do you plan on dropping `extra_index_urls`?

python/private/pypi/extension.bzl
```diff
@@ -525,7 +526,72 @@ supported from this version without extra handling from the user.
...
+} | AUTH_ATTRS | {
+    # Downloader options
+    "downloader": attr.string(
+        values = ["disabled", "debug", "enabled", "auto"],
+        default = "auto",
+        doc = """\
+Option values:
+* `disabled` - disable downloader.
+* `debug` - stop using parallel downloading for debugging cases.
+* `enabled` - enable downloader for all hubs.
```
dougthor42, line 538: ~~`enabled` would only apply to the hub that it's listed on, right?~~

```starlark
pip.parse(hub_name="foo", downloader="enabled", ...)
pip.parse(hub_name="bar", downloader="disabled", ...)
```

Oh this is the global config as part of `_default_attrs` and the pypi `module_extension`, not the `.parse` tag class. I didn't notice that at first.

So this would be used via:

```starlark
pip = use_extension(@rules_python//python/extensions:pip.bzl, "pip")
pip.default(downloader = "enabled", index_overrides = {...})
pip.parse(...)
```

Correct?

Would the `pip.parse` still support extending/overriding `index_overrides` and indexes?

python/private/pypi/extension.bzl
```diff
@@ -525,7 +526,72 @@ supported from this version without extra handling from the user.
...
-} | AUTH_ATTRS
+} | AUTH_ATTRS | {
+    # Downloader options
+    "downloader": attr.string(
+        values = ["disabled", "debug", "enabled", "auto"],
+        default = "auto",
+        doc = """\
+Option values:
+* `disabled` - disable downloader.
+* `debug` - stop using parallel downloading for debugging cases.
```
dougthor42, line 537: Will this also increase logging levels? Or is that separate?

python/private/pypi/extension.bzl
```diff
@@ -525,7 +526,72 @@ supported from this version without extra handling from the user.
 :::
 """,
     ),
-} | AUTH_ATTRS
+} | AUTH_ATTRS | {
+    # Downloader options
+    "downloader": attr.string(
```
dougthor42, line 531: I would say `bazel_downloader` is better. As is, it kinda sounds like it would modify downloading altogether (similar to `pip.parse.download_only`).

python/private/pypi/extension.bzl
```diff
@@ -525,7 +526,72 @@ supported from this version without extra handling from the user.
...
+        values = ["disabled", "debug", "enabled", "auto"],
+        default = "auto",
+        doc = """\
+Option values:
+* `disabled` - disable downloader.
```
dougthor42, line 536: Nit: mention that `pip` will be used to download things instead.

rickeylev: +1, except instead of pip, be vague and say some external tool

python/private/pypi/extension.bzl
rickeylev: Can you explain the use case? Why would an index have version 1 but not version 2? Or the distributions for windows, but not linux?

The case I can think of is e.g. cuda. Torch has different index urls for different cuda etc builds, but the env marker language can't express that anyways.

python/private/pypi/extension.bzl
aignas: My thinking was that extra index urls can come from pip args and requirements files.

python/private/pypi/extension.bzl
aignas: I am thinking that we can get rid of parallel downloading. Could not think of a better name for this.

python/private/pypi/extension.bzl
dougthor42: A real-life example is airlock/, which has a 2 week soak time for new releases: someone deploys Version 2 to PyPI but it will not show up in Airlock for two weeks - only Version 1 is available until the new version gets vetted and whatnot.

A made-up example is that a package was migrated from INDEX_A to INDEX_B and because $REASON, historical artifacts could not be ported. I highly doubt such a case would occur.

I can't think of any logical reason why an index would only have artifacts of a given OS.

Admittedly this is mostly just theorycrafting. Even in the Airlock case, we just tell people to wait the 2 weeks...

python/private/pypi/extension.bzl
dougthor42: OK, so extending that premise: why not also pull `index_url` from the requirements.txt files too?

python/private/pypi/extension.bzl
dougthor42: I think `debug` is fine for such a case.

Having a dedicated `serial`, `download_serially`, `sequential`, etc. value might be too restrictive.

python/private/pypi/extension.bzl
```diff
@@ -525,7 +526,72 @@ supported from this version without extra handling from the user.
...
+if necessary.
+
+The key is the package name (will be normalized before usage) and the value is the
+index URL.
+
+This design pattern has been chosen in order to be fully deterministic about which
+packages come from which source. We want to avoid issues similar to what happened in
+https://pytorch.org/blog/compromised-nightly-dependency/.
+
+The indexes must support Simple API as described here:
```
dougthor42, line 564: Are url-based requirements specifiers in scope? For example `pip @ https://github.com/pypa/pip/archive/22.0.2.zip` is equivalent to

```starlark
index_overrides = {
    "pip": "https://github.com/pypa/pip/archive/22.0.2.zip"
}
```

Note: IMO this should not be valid, I'm just trying to get clarification.

python/private/pypi/extension.bzl
aignas: The reason is that we have to have a default somewhere if the index is not specified. If it is specified in the lock file, then we can either:
* Use them as extra index_urls.
* Use them and ignore the default that we set.
* Use PyPI as a fallback which gets used if else fails.

At least this is for `requirements.txt`, for PEP751 lock files the index is there always (I think).

python/private/pypi/extension.bzl
aignas: I am wondering about the `index_overrides` - the root module should be able to override anything here. However, what happens if a non-root module is defining this override. To do this blindly is a little sweat-inducing because of all of the recent supply chain attacks the public registries faced. I would say that we should keep `index_overrides` as a root-module-only option. WDYT?

dougthor42: I agree that non-root module overrides is probably a Bad Idea :tm:. I'm more asking about the pip.parse overrides:

```starlark
pip.default(
    downloader = "enabled",
    index_url = AIRLOCK_INDEX_URL,
    index_overrides = {"foo": INDEX_A, "bar": INDEX_A},
)
pip.parse(
    # this parse pulls from 2 indexes: AIRLOCK_INDEX_URL and INDEX_A
    hub_name = "pypi",
    python_version = PYTHON_CURRENT_VERSION,
)
pip.parse(
    # this parse pulls from 3 indexes: AIRLOCK_INDEX_URL for most things,
    # INDEX_A for 'bar' package, and INDEX_B for 'foo' package
    experimental_index_url_overrides = {"foo": INDEX_B},  # same package, different index
    hub_name = "pypi",
    python_version = PYTHON_NEXT_VERSION,
    requirements_lock = "//:requirements_pynext.txt",
)
```

Like elsewhere, this is mostly hypothetical - I personally haven't had a need to use different indexes for different (python versions | hubs | requirement lock files).

python/private/pypi/extension.bzl
dougthor42: > use [`index_url` specified in the lock file] as extra index_urls

I fear that this would risk package name collisions and would be difficult to figure out the index ordering.

> Use them and ignore the default that we set.

This sounds reasonable to me, but then a root module will not be able to override a child module's config (which I believe is a desired feature) without adjusting the child's `requirements.txt` directly:

```
./MODULE.bazel        # sets pip.default(downloader = "enabled", index_url = BAR)
./requirements.txt    # sets --index-url FOO
./child_module/       
    MODULE.bazel      # sets pip.default(downloader = "enabled", index_url = AAA)
    requirements.txt  # sets --index-url BAD
```

If we ignore `pip.default`, the root `MODULE.bazel` will pull from `FOO` and the child will pull from `BAD`. Right? 


> Use PyPI as a fallback which gets used if else fails.

If I set `index_url` in any of my requirements.txt files, I do not want pypi used at all.

```
$ head requirements.txt 
# This file was autogenerated by uv via the following command:
#    uv pip compile --python-version=3.12 --output-file=requirements.txt requirements.in
--index-url https://oauth2accesstoken@us-python.pkg.dev/REDACTED/simple
--extra-index-url https://oauth2accesstoken@us-west2-python.pkg.dev/REDACTED/simple

absl-py==2.3.0 \
```

> for PEP751 lock files the index is there always (I think).

It's [optional](https://packaging.python.org/en/latest/specifications/pylock-toml/#packages-index) :disappointed:

---

I think it's reasonable to put _all_ index config in Bazel and completely ignore any index config set in requirements.txt or PEP751 `pylock.toml` files.

rickeylev:
> airlock has soak time

My day team has run into something similar.

A release candidate will be created, but its hard to consume unless its also uploaded to pypi. Similarly, various projects are wired up to use nightly wheels of important dependencies. Right now, this works by using file:// in requirements files and some repo-rule pre-processing.

> package migrated from index a to index b

iirc, my day team had to change the GCS buckets wheels were being uploaded to. There were a lot of old wheels, and it was generally a pain, so they only copied over recent releases and left the old ones sitting there. The net effect is a partial index based on version. I think this is really niche, though, so don't think its worth trying to add `foo >1.0` support in the override logic.

python/private/pypi/extension.bzl
rickeylev: > non-root index overrides

I agree it's a bad idea to let non-root add indexes. It only makes sense for the isolated=True case, where you want a fully private dependency closure. But I think we should worry about that separately.

> this is mostly hypothetical

An easy non-hypothetical case is pytorch. They have 5 different indexes for cuda, rocm, cpu, etc builds of pytorch. A requirements file just says `torch == whatever`. To make this work on the bazel side, you have to, somehow, specify that, for requirements_cpu.txt, needs torch to come from the pytorch cpu index. Meanwhile, requirements_cuda.txt needs torch to come from the pytorch cuda index.

(Torch indexes are partial, so you can't use them as the sole index. But that's just how they've decided to do their index).

PR #3330, opened 2025-10-07
feat(toolchains): Add 3.14.0

Ahajha:
Per the title: 3.14.0 was released today

Update `CHANGELOG.md` as applicable


rickeylev (approved):
Thank you!

python/versions.bzl
```diff
@@ -810,28 +810,28 @@ TOOL_VERSIONS = {
             "x86_64-unknown-linux-gnu-freethreaded": "python/install",
         },
     },
-    "3.14.0rc1": {
-        "url": "20250808/cpython-{python_version}+20250808-{platform}-{build}.{ext}",
+    "3.14.0": {
+        "url": "20251007/cpython-{python_version}+20251007-{platform}-{build}.{ext}",
```
aignas, line 814: I would have loved to see the other toolchains from the same release to be also added, because that normally brings a number of fixes together. Nevertheless, thank you for contribution.

Ahajha: Ah that would make sense, just bumped this one since it addressed my immediate needs. I might add those, do you usually add them or replace the existing ones?

aignas: Replace the existing versions if they overlap and add a small note in the CHANGELOG for the changes.

Ahajha: Sounds good, will try to get to that tomorrow.

PR #3331, opened 2025-10-08
fix(venv): symlink shared libraries directly 

rickeylev:
It seems `$ORIGIN` resolves prior to symlink resolution. This makes it resolve
differently depending on if the directory or file itself is symlinked.

To fix, special case shared libraries and have them symlinked directly. Since an
explicit file is the target, `VenvSymlinkEntry.link_to_file` is added to hold the
File object that will be linked to.

An unfortunate side-effect of this logic is any package with `lib*.so` files
will be more expensive to build (depset flattened at analysis time, more files
symlinked), but it beats not working at all. Optimizing that can be done in
another change.

Tests added to generate libraries that look like what something from PyPI
does. Manually verified a case using jax and jax plugins.

Fixes https://github.com/bazel-contrib/rules_python/issues/3228


python/private/venv_runfiles.bzl
```diff
@@ -170,7 +173,9 @@ def _merge_venv_path_group(ctx, group, keep_map):
     # TODO: Compute the minimum number of entries to create. This can't avoid
     # flattening the files depset, but can lower the number of materialized
     # files significantly. Usually overlaps are limited to a small number
-    # of directories.
+    # of directories. Note that, when doing so, shared libraries need to
+    # be symlinked directory, not the directory containing them, due to
```
hartikainen: There's probably a typo in "shared libraries need to be symlinked directory". That sentence doesn't make sense to me at least.

rickeylev: Thanks. Fixed.

rickeylev:
@hartikainen Give this a try? I tried this with your repro and things appear to work (I don't have a gpu on my machine, so the jax call to get gpu devices failed, but otherwise everything imported successfully).  This commit has the patch to apply: https://github.com/bazel-contrib/rules_python/commit/add0c262ed6cabe50df9266157ac74aca384f3f9.patch

Otherwise, this is almost ready, just need to do some cleanup for CI.

rickeylev:
PTAL. Cleaned up and ready

python/private/venv_runfiles.bzl
```diff
@@ -172,7 +175,9 @@ def _merge_venv_path_group(ctx, group, keep_map):
     # TODO: Compute the minimum number of entries to create. This can't avoid
     # flattening the files depset, but can lower the number of materialized
     # files significantly. Usually overlaps are limited to a small number
-    # of directories.
+    # of directories. Note that, when doing so, shared libraries need to
+    # be symlinked directly, not the directory containing them, due to
+    # dynamic linker symlink resolution semantics on Linux.
```
aignas, line 180: FYI, the shared libraries quite often (but not necessarily always) will be in a directory named `site-packages/<pkg_name>.libs`. This is true for most of the linux wheels that have been standardized layout after running through `auditwheel`.

https://github.com/pypa/auditwheel

python/private/venv_runfiles.bzl
```diff
@@ -249,13 +254,27 @@ def get_venv_symlinks(ctx, files, package, version_str, site_packages_root):
             continue
         path = path.removeprefix(site_packages_root)
         dir_name, _, filename = path.rpartition("/")
+        runfiles_dir_name, _, _ = runfiles_root_path(ctx, src.short_path).partition("/")
+
+        if _is_linker_loaded_library(filename):
+            entry = VenvSymlinkEntry(
+                kind = VenvSymlinkKind.LIB,
+                # todo: omit setting link_to_path
```
aignas: nit: I guess this will be for a later PR?

python/private/venv_runfiles.bzl
```diff
@@ -278,6 +297,8 @@ def get_venv_symlinks(ctx, files, package, version_str, site_packages_root):
                 files = depset([src]),
             )
             venv_symlinks.append(entry)
+        else:
+            fail("hit", src)
```
aignas: nit: please remove or add a better error message.

tests/support/copy_file.bzl
```diff
@@ -0,0 +1,25 @@
+"""Copies a file to a directory."""
+
+def _copy_file_to_dir_impl(ctx):
```
aignas, line 3: nit: could you add a line why we cannot use `copy_file` from `bazel_skylib`? It has an option to copy instead of symlinking the file.

rickeylev:
@gemini-code-assist /review

python/private/venv_runfiles.bzl
rickeylev: Are you suggesting symlinking the `name.libs` directory? I think that would have the same problem as the (non-.libs) top-level directory being symlinked. What happens is `/runfiles/pypi_foo/site-packages/foo.libs` is gets set as the value of $ORIGIN. Then `$ORIGIN/../torch/libs/libwhateve.so` points to somewhere that doesn't exist.

python/private/venv_runfiles.bzl
rickeylev: Removed the todo

python/private/venv_runfiles.bzl
rickeylev: oops, yes, debug logic, removed.

python/private/venv_runfiles.bzl
```diff
@@ -312,6 +333,22 @@ def get_venv_symlinks(ctx, files, package, version_str, site_packages_root):
...
+    """
+    if not filename.startswith("lib"):
+        return False
+    if filename.endswith((".so", ".dylib", ".lib")):
+        return True
+
+    # Versioned library, e.g. libfoo.so.1.2.3
+    if ".so." in filename:
+        return True
+    return False
```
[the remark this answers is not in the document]
rickeylev: Fixed.

tests/support/copy_file.bzl
rickeylev: Done.

python/private/venv_runfiles.bzl
aignas: No, just a mental note that the so symlinking in the linux case in when they are stored in `.libs` folder is efficient, because this does not necessarily cause `.py` file symlinking.

shayanhoshyari:
Very nice! Thank you.

hartikainen:
Very nice indeed! I just tested this from https://github.com/bazel-contrib/rules_python/compare/main...hartikainen:rules_python:venv-site-packages-investigation and this work as expected. Thank you so much @rickeylev and @aignas 🙌

PR #3334, opened 2025-10-10
build: Starlarkify python flags

mai93:
Add starlark flags for `--python_path`, `--build_python_zip` and `--incompatible_default_to_explicit_init_py`. 

- The transitions logic is updated to set both the native and starlark versions of the flags to allow alternating between them until the native ones are removed.
- `--build_python_zip` is changed to `boolean` instead of `Tristate` with default value set to `True` on `windows` and `False` otherwise.
- `scope = universal` attribute is added to the starlark flags so they can be propagated to exec config on Bazel 9. This required upgrading `bazel_skylib` version to have `scope` attribute defined.

Work towards: https://github.com/bazel-contrib/rules_python/issues/3252

cc @gregestren

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -123,7 +123,7 @@ intersphinx_mapping = {
 extlinks = {
     "gh-issue": (f"https://github.com/bazel-contrib/rules_python/issues/%s", "#%s issue"),
     "gh-path": (f"https://github.com/bazel-contrib/rules_python/tree/main/%s", "%s"),
-    "gh-pr": (f"https://github.com/bazel-contrib/rules_python/pulls/%s", "#%s PR"),
+    "gh-pr": (f"https://github.com/bazel-contrib/rules_python/pull/%s", "#%s PR"),
 }
 
 # --- MyST configuration
--- a/tests/venv_site_packages_libs/BUILD.bazel
+++ b/tests/venv_site_packages_libs/BUILD.bazel
@@ -28,6 +28,7 @@ py_reconfig_test(
     venvs_site_packages = "yes",
     deps = [
         ":closer_lib",
+        "//tests/venv_site_packages_libs/nested_with_pth",
         "//tests/venv_site_packages_libs/nspkg_alpha",
         "//tests/venv_site_packages_libs/nspkg_beta",
         "//tests/venv_site_packages_libs/pkgutil_top",
--- a/tests/venv_site_packages_libs/bin.py
+++ b/tests/venv_site_packages_libs/bin.py
@@ -40,6 +40,8 @@ class VenvSitePackagesLibraryTest(unittest.TestCase):
         self.assert_imported_from_venv("nspkg.subnspkg.delta")
         self.assert_imported_from_venv("single_file")
         self.assert_imported_from_venv("simple")
+        m = self.assert_imported_from_venv("nested_with_pth")
+        self.assertEqual(m.WHOAMI, "nested_with_pth")
 
     def test_data_is_included(self):
         self.assert_imported_from_venv("simple")
--- a/tests/venv_site_packages_libs/nested_with_pth/BUILD.bazel
+++ b/tests/venv_site_packages_libs/nested_with_pth/BUILD.bazel
@@ -1,0 +1,11 @@
+load("//python:py_library.bzl", "py_library")
+
+package(default_visibility = ["//visibility:public"])
+
+py_library(
+    name = "nested_with_pth",
+    srcs = glob(["site-packages/**/*.py"]),
+    data = glob(["site-packages/*.pth"]),
+    experimental_venvs_site_packages = "//python/config_settings:venvs_site_packages",
+    imports = [package_name() + "/site-packages"],
+)
--- a/tests/venv_site_packages_libs/nested_with_pth/site-packages/nested.pth
+++ b/tests/venv_site_packages_libs/nested_with_pth/site-packages/nested.pth
@@ -1,0 +1,1 @@
+nested_sdk
--- a/tests/venv_site_packages_libs/nested_with_pth/site-packages/nested_sdk/nested_with_pth/__init__.py
+++ b/tests/venv_site_packages_libs/nested_with_pth/site-packages/nested_sdk/nested_with_pth/__init__.py
@@ -1,0 +1,1 @@
+WHOAMI = "nested_with_pth"


gazelle/WORKSPACE
```diff
@@ -2,6 +2,19 @@ workspace(name = "rules_python_gazelle_plugin")
 
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
+http_archive(
+    name = "bazel_skylib",
+    sha256 = "6e78f0e57de26801f6f564fa7c4a48dc8b36873e416257a92bbb0937eeac8446",
+    urls = [
+        "https://mirror.bazel.build/github.com/bazelbuild/bazel-skylib/releases/download/1.8.2/bazel-skylib-1.8.2.tar.gz",
```
dougthor42, line 9: Can we version-sync with what's in gazelle/MODULE.bazel?

https://github.com/bazel-contrib/rules_python/blob/5c13539708a80560729424e341603ac59d6ecece/gazelle/MODULE.bazel#L7

Either upgrading MODULE.bazel or downgrading this one.

gazelle/WORKSPACE
```diff
@@ -2,6 +2,19 @@ workspace(name = "rules_python_gazelle_plugin")
 
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
+http_archive(
+    name = "bazel_skylib",
```
dougthor42, line 6: The root `WORKSPACE` doesn't also need this `http_archive`? If not, why not?

gazelle/WORKSPACE
dougthor42: Ah, you said that bazel_skylib needs to have the `scope` attribute, so upgrading gazelle/MODULE.bazel is the way to go.

gazelle/WORKSPACE
mai93: The root `WORKSPACE` gets it from `rules_python_internal_deps` (called [here](https://github.com/bazel-contrib/rules_python/blob/main/WORKSPACE#L28-L30))

gazelle/WORKSPACE
mai93: Done

gazelle/WORKSPACE
dougthor42: Ah right, I'd forgotten about that! Thanks.

examples/build_file_generation/WORKSPACE
```diff
@@ -18,6 +18,19 @@ load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 # Define an http_archive rule that will download the below ruleset,
 # test the sha, and extract the ruleset to you local bazel cache.
 
+http_archive(
+    name = "bazel_skylib",
+    sha256 = "6e78f0e57de26801f6f564fa7c4a48dc8b36873e416257a92bbb0937eeac8446",
+    urls = [
+        "https://mirror.bazel.build/github.com/bazelbuild/bazel-skylib/releases/download/1.8.2/bazel-skylib-1.8.2.tar.gz",
+        "https://github.com/bazelbuild/bazel-skylib/releases/download/1.8.2/bazel-skylib-1.8.2.tar.gz",
```
aignas: What happens if you remove this? I would expect the `python/private/py_repositories.bzl` change to be sufficient and this block should can be removed?

rickeylev: Yeah, the naming of these setup files is understandably confusing 😕 . The file you want to modify is python/private/py_repositories.bzl instead.

python/private/attributes.bzl
```diff
@@ -179,6 +179,13 @@ COMMON_ATTRS = dicts.add(
         # buildifier: disable=attr-license
         "licenses": attr.license() if hasattr(attr, "license") else attr.string_list(),
     },
+    # starlark flags attributes
+    {
+        "_build_python_zip": attr.label(default = "//python/config_settings:build_python_zip"),
```
rickeylev: nit: please suffix the attr names with `_flag`, its a convention we have to help distinguish target-specific attributes vs flags.

rickeylev: All of these are only applicable to tests/binarys. Please move them to py_executable.bzl#EXECUTABLE_ATTRS

python/private/common_labels.bzl
```diff
@@ -7,6 +7,7 @@ labels = struct(
     # keep sorted
     ADD_SRCS_TO_RUNFILES = str(Label("//python/config_settings:add_srcs_to_runfiles")),
     BOOTSTRAP_IMPL = str(Label("//python/config_settings:bootstrap_impl")),
+    BUILD_PYTHON_ZIP = str(Label("//python/config_settings:build_python_zip")),
```
rickeylev, line 10: Please add python_import_all_repositories here, too.

python/private/py_executable.bzl
```diff
@@ -1136,6 +1136,9 @@ def _get_runtime_details(ctx, semantics):
     # TOOD(bazelbuild/bazel#7901): Remove this once --python_path flag is removed.
 
     flag_interpreter_path = read_possibly_native_flag(ctx, "python_path")
+    if not flag_interpreter_path.startswith("python") and not paths.is_absolute(flag_interpreter_path):
```
rickeylev, line 1146: I'm not sure this logic is correct?

On Linux, the value will be empty. Thus we get `True and True`.

Also, I'm not sure if skylib paths.is_absolute handles Windows `C:\` or backslashes.

rickeylev:
@gemini-code-assist /review

python/private/py_executable.bzl
```diff
@@ -1136,6 +1136,9 @@ def _get_runtime_details(ctx, semantics):
...
+        fail("python_path must be an absolute path when it is set.")
```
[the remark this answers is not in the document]
rickeylev: Please fix -- the error message should specify it can be absolute path or program name

examples/build_file_generation/WORKSPACE
mai93: I tried doing this but it needed moving the `rules_python` setup before the `rules_go` one because `rules_go` brings in an older version of `bazel_skylib` ([link](https://github.com/bazel-contrib/rules_go/blob/master/go/private/repositories.bzl#L51-L61)). It seems the current restructure works with CI tests.

python/private/py_executable.bzl
mai93: > On Linux, the value will be empty. Thus we get True and True.

the default value of the flag target is set to `python` ([here](https://github.com/bazel-contrib/rules_python/blob/aed97f631129f91ac1d962b20daf9f852f8ad7b6/python/config_settings/BUILD.bazel#L269)) so if it is not used on linux it should be `python`, is this accepted?

> Also, I'm not sure if skylib paths.is_absolute handles Windows C:\ or backslashes.

from [its implementation](https://github.com/bazelbuild/bazel-skylib/blob/main/lib/paths.bzl#L68), I think it can detect `C:` as an absolute path.

python/private/attributes.bzl
mai93: done

mai93: done

python/private/common_labels.bzl
mai93: done

python/private/py_executable.bzl
rickeylev: Ah, I see. In java, the static default was null, but null was converted to "python" upon access. LGTM

BryceBeagle:
Just commenting for other readers since I encountered this while upgrading `rules_python` to `1.7.0`:

This change means `rules_python` now _requires_ `bazel-skylib>=1.8.0` which introduced `scope` in https://github.com/bazelbuild/bazel-skylib/pull/546

If you're using an older version of `skylib` you will get:
```
❯ bzl build //domains/assistant/apps/apis/assistant_api/internal/...

ERROR: /private/var/tmp/.../external/rules_python/python/config_settings/BUILD.bazel:245:10: @@rules_python//python/config_settings:experimental_python_import_all_repositories: no such attribute 'scope' in 'bool_flag' rule
ERROR: /private/var/tmp/.../external/rules_python/python/config_settings/BUILD.bazel:252:10: @@rules_python//python/config_settings:build_python_zip: no such attribute 'scope' in 'bool_flag' rule
ERROR: /private/var/tmp/.../external/rules_python/python/config_settings/BUILD.bazel:260:10: @@rules_python//python/config_settings:incompatible_default_to_explicit_init_py: no such attribute 'scope' in 'bool_flag' rule
ERROR: /private/var/tmp/.../external/rules_python/python/config_settings/BILD.bazel:267:12: @@rules_python//python/config_settings:python_path: no such attribute 'scope' in 'string_flag' rule
WARNING: Target pattern parsing failed.
ERROR: Skipping '@rules_python//python/config_settings:bootstrap_impl': Error evaluating '@rules_python//python/config_settings:bootstrap_impl': error loading package '@@rules_python//python/config_settings': Package 'python/config_settings' contains errors
ERROR: @rules_python//python/config_settings:bootstrap_impl :: Error loading option @rules_python//python/config_settings:bootstrap_impl: Error evaluating '@rules_python//python/config_settings:bootstrap_impl': error loading package '@@rules_python//python/config_settings': Package 'python/config_settings' contains errors
```

PR #3355, opened 2025-10-14
chore: fix create_archive_and_notes to ignore release tool markers

rickeylev:
The create_archive_and_notes.sh script is incorrectly detecting the version markers
in the tool meant to rewrite them. To fix, ignore those files.


.github/workflows/create_archive_and_notes.sh
```diff
@@ -18,7 +18,13 @@ set -o errexit -o nounset -o pipefail
...
 # find the substring we're looking for in our own file.
 # Exclude CONTRIBUTING.md, RELEASING.md because they document how to use these strings.
-if grep --exclude=CONTRIBUTING.md --exclude=RELEASING.md --exclude-dir=.* VERSION_NEXT_ -r; then
+grep --exclude=CONTRIBUTING.md \
+  --exclude=RELEASING.md \
+  --exclude=release.py \
+  --exclude=release_test.py \
+  --exclude-dir=.* \
+  VERSION_NEXT_ -r
+if $?; then
```
rickeylev: Ah, this is wrong actually. set -errexit above will stop the script if grep above doesn't match anything. Need to figure out how to do this...

PR #3357, opened 2025-10-15
chore(toolchain): use the last build and add 3.15.0a1

aignas:
Updating the toolchain versions to fix a bug in `3.13.8`. This will make
the release better to everyone.


rickeylev (approved):
Since the BCR release is stuck, I think it's fine to put this in 1.7 and do an rc1

aignas:
/gemini adjust the changelog of 1.7 to mention the new versions and the release.

PR #3358, opened 2025-10-15
doc: Fix `pip.default` arguments in multi-platform example

hartikainen:
I was playing around with the [multi-platform PyPI example](https://rules-python.readthedocs.io/en/latest/howto/multi-platform-pypi-deps.html) and noticed that the `pip.default` arguments were wrong.


rickeylev (approved):

rickeylev:
Thanks for the doc fix!

PR #3359, opened 2025-10-16
chore: switch to use publish-to-bcr workflow

rickeylev:
The GitHub App is deprecated, so switch to the more modern workflow.

This will also allow us to eventually use attestations.

Along the way...

* Split up the release workflow into some different sub-jobs
* Run PyPI upload last, as its the most irrevocable step of the process.


rickeylev:
I tried to follow the docs and looked at the rules_lint reference. Haven't tested it end to end, though.

alexeagle (approved):

rickeylev:
Ugh, I botched this rewrite entirely. Feel free to revert in the meantime. Working on a fix.

PR #3365, opened 2025-10-16
chore: make gazelle bcr tests compatible with bcr presubmit environment

rickeylev:
BCR recently changed how it extracts archives in its presubmits: instead of extracting
the whole archive, only the specified portion (`gazelle/` in this case) is extracted.
This broke the gazelle tests because they reference files above the gazelle directory.

To fix, move the module it runs as a test under the gazelle directory.

Because the test module also refers to rules_python, which is above the gazelle
directory, the bcr presubmit has disable that override using `--override_module`.
This means, going forward, the gazelle module, when bcr tests it, will use the
version in the MODULE file (rather than the vendored copy).

Fixes https://github.com/bazel-contrib/rules_python/issues/3364

### DIFF
--- a/gazelle/.bazelignore
+++ b/gazelle/.bazelignore
@@ -1,0 +1,8 @@
+bazel-bin
+bazel-gazelle
+bazel-out
+bazel-testlogs
+examples/bzlmod_build_file_generation/bazel-bin
+examples/bzlmod_build_file_generation/bazel-bzlmod_build_file_generation
+examples/bzlmod_build_file_generation/bazel-out
+examples/bzlmod_build_file_generation/bazel-testlog
--- a/gazelle/.bazelrc
+++ b/gazelle/.bazelrc
@@ -1,3 +1,6 @@
+common --deleted_packages=examples/bzlmod_build_file_generation
+common --deleted_packages=examples/bzlmod_build_file_generation/runfiles
+
 test --test_output=errors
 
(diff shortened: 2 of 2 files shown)


gazelle/examples/bzlmod_build_file_generation/MODULE.bazel
```diff
@@ -13,26 +13,28 @@ module(
...
-    path = "../..",
+    path = "../../..",
 )
 
 # The following stanza defines the dependency rules_python_gazelle_plugin.
 # For typical setups you set the version.
 # See the releases page for available versions.
 # https://github.com/bazel-contrib/rules_python/releases
-bazel_dep(name = "rules_python_gazelle_plugin", version = "0.0.0")
+bazel_dep(name = "rules_python_gazelle_plugin", version = "1.5.0")
```
dougthor42, line 31: Do we want a `LINT.IfChange` wrapper around here, mapped to the plugin release version? I know we don't have any tooling for it, but it's still a good note for reviewers.

Or maybe the example only a min version 1.5 because that's all the features it uses, so we don't need to keep it in sync unless the example starts using newer features.

rickeylev: For the example, the version here doesn't matter because the local_path_override below will override it.

The rules_python version in gazelle's MODULE.bazel will matter, though. It now acts the minimum necessary version of rules_python gazelle needs.

PR #3366, opened 2025-10-17
chore: fix release workflow

rickeylev:
My original change to use the bazel-contrib BCR publishing workflow was horribly broken.
This fixes a variety of issues.

* Fix the workflow call secrets variable name
* Allow getting the BCR publishing token from the workflow arg (the release
  workflow calls it)
* Fix the overall syntax of the workflows. It was just entirely invalid in
  several regards.
* Fix the yaml -> yml file name reference. NGL, it took me longer than I'd
  like to admit to see that, eh.

Along the way, some cleanups and improvements

* Allow specifying a particular tag to release, while using workflow state from
  a different commit. This allows us to make fixes on main, and then use it to
  manually re-trigger a tag to be released.
* Add descriptions for workflow inputs
* Allow bcr token to be passed to release. This will allow e.g. automatically
  scheduled releases.
* Quote shell variables because its good practice.


.github/workflows/publish.yml
```diff
@@ -1,35 +1,34 @@
...
       tag_name: ${{ inputs.tag_name }}
       # GitHub repository which is a fork of the upstream where the Pull Request will be opened.
       registry_fork: bazel-contrib/bazel-central-registry
       attest: false
     permissions:
       contents: write
     secrets:
-      publish_token: ${{ secrets.BCR_PUBLISH_TOKEN }}
+      # Necessary to push to the BCR fork, and to open a pull request against a registry
+      publish_token: ${{ secrets.publish_token || secrets.BCR_PUBLISH_TOKEN }}
```
aignas, line 34: nit: why not have them to be the same thing?

aignas (approved):

.github/workflows/publish.yml
rickeylev: I think to avoid a name clash while allowing the publish workflow to be be called directly (where it uses BCR_PUBLISH_TOKEN) or by another workflow (which uses publish_token).

PR #3367, opened 2025-10-17
chore(deps): update rules_cc and protobuf versions

mai93:
These updates are mainly needed for compatibility with bazel 9:

- Update version of `rules_cc` to [0.2.10](https://github.com/bazelbuild/rules_cc/releases/tag/0.2.10) to fix [the cyclic dependency error](https://buildkite.com/bazel/rules-python-python/builds/13441#019a01d8-d32f-429a-b465-66caee545ab7) while testing with bazel 9 pre-release
- Update version of `protobuf` to [33.0](https://github.com/protocolbuffers/protobuf/releases/tag/v33.0) which has load statements for `rules_cc` needed for bazel 9

### DIFF
--- a/tests/integration/pip_parse/WORKSPACE
+++ b/tests/integration/pip_parse/WORKSPACE
@@ -9,9 +9,11 @@ py_repositories()
 
 # Needed for rules_cc 0.2.10
 load("@bazel_features//:deps.bzl", "bazel_features_deps")
+
 bazel_features_deps()
 
 load("@rules_cc//cc:extensions.bzl", "compatibility_proxy_repo")
+
 compatibility_proxy_repo()
 
 python_register_toolchains(
--- a/examples/build_file_generation/WORKSPACE
+++ b/examples/build_file_generation/WORKSPACE
@@ -31,9 +31,11 @@ py_repositories()
 
 # Needed for rules_cc 0.2.10
 load("@bazel_features//:deps.bzl", "bazel_features_deps")
+
 bazel_features_deps()
 
 load("@rules_cc//cc:extensions.bzl", "compatibility_proxy_repo")
+
 compatibility_proxy_repo()
 
 # We now register a hermetic Python interpreter rather than relying on a system-installed interpreter.
--- a/examples/multi_python_versions/WORKSPACE
+++ b/examples/multi_python_versions/WORKSPACE
@@ -11,9 +11,11 @@ py_repositories()
 
 # Needed for rules_cc 0.2.10
 load("@bazel_features//:deps.bzl", "bazel_features_deps")
+
 bazel_features_deps()
 
 load("@rules_cc//cc:extensions.bzl", "compatibility_proxy_repo")
+
 compatibility_proxy_repo()
 
 default_python_version = "3.9"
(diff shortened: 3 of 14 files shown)


tests/integration/pip_parse/WORKSPACE
```diff
@@ -7,6 +7,13 @@ load("@rules_python//python:repositories.bzl", "py_repositories", "python_regist
 
 py_repositories()
 
+# Needed for rules_cc 0.2.10
+load("@bazel_features//:deps.bzl", "bazel_features_deps")
```
rickeylev, line 11: Instead of copy/pasting this same setup to every downstream workspace, create a second setup function and load and call that.

`python/repositories_deps.bzl`

which does the 4 lines

.bazelci/presubmit.yml
```diff
@@ -43,9 +43,15 @@ buildifier:
   build_flags:
     - "--noenable_bzlmod"
     - "--build_tag_filters=-integration-test"
+    # recent abseil-cpp requires C++17 which is not used by default in Bazel 7
+    - "--cxxopt=-std=c++17"
```
rickeylev, line 80: Can you explain a bit more about this? I'm guessing this is coming from protobuf?

rickeylev (commented):
I pushed an update to the changelog.

Please update the PR description to explain why these bumps were necessary.

Mostly LGTM, sans the question and need to create deps.bzl function

.bazelci/presubmit.yml
mai93: yes, for protobuf 33.0 (and its dependency abseil-cpp) C++17 is the minimum supported version (https://github.com/google/oss-policies-info/blob/main/foundational-cxx-support-matrix.md)

I'm trying to add these flags to make it work with Bazel 7 but it is not straightforward as it is still failing for tests on windows (https://buildkite.com/bazel/rules-python-python/builds/13428#0199f84a-0a7c-4ed7-8b30-6ab8ef898f71) where `-std=c++17` is not recognized. Do you have any suggestions for that?

tests/integration/pip_parse/WORKSPACE
mai93: I tried having a file `load_cc_deps.bzl` that looks like this:
```
load("@bazel_features//:deps.bzl", "bazel_features_deps")
load("@rules_cc//cc:extensions.bzl", "compatibility_proxy_repo")

def add_cc_deps():
  bazel_features_deps()
  compatibility_proxy_repo()
```

but it did not work:
```
ERROR: Failed to load Starlark extension '@@bazel_features_version//:version.bzl'.
Cycle in the workspace file detected. This indicates that a repository is used prior to being defined.
```
because `bazel_features_deps()` call needs to happen before the `
load("@rules_cc//cc:extensions.bzl", "compatibility_proxy_repo")` and we cannot have load statements in starlark functions, or what you're suggesting is different?

aignas:
Playing devil's advocate. This will be a breaking change for bazel 7 users who rely on protobuf. Should we cut a rules_python 2 where we remove protobuf dep instead? We can still merge the PR, but we could do a lightweight version 2 release with cleaning a few things up afterwards.

Ahajha:
Bump on this, this is now also necessary since the 3.8 support drop, the old version needs 3.8 still. I opened another PR which is failing for some reason, but if this one fixes that issue I'm happy to close mine.

PR #3368, opened 2025-10-17
chore: change deleted package flag updater to put one per line

rickeylev:
The `--deleted_package` lines are so long they can't be manually updated.
They were only put onto a single line because older Bazel versions didn't
support accumulative behavior.

Along the way, switch to using `common` command for them, too. The supported
Bazel versions are new enough to support it.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,9 +1,9 @@
-# For bazel-in-bazel testing
-# Trick bazel into treating BUILD files under examples/* as being regular files
+# For bazel-in-bazel testing and ignoring sub-workspaces for ... patterns.
+#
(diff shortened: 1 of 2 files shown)


rickeylev:
It seems like this added the entire gazelle sub-tree, which wasn't all listed before.

I think that's the correct thing to do, but I'm confused how things were working previously if those files weren't ignored. Shouldn't `...` have been matching files under gazelle and giving errors?

In any case, it looks like an exception for `//gazelle/docs` needs to be made -- it gets fed into our doc gen. Or maybe we should change that to a `@gazelle_whatever` reference instead.

aignas:
Shoot, missed this when was proposing #3425. Thanks!

PR #3372, opened 2025-10-20
feat: Add libpython QNX platform support

lalten:
Add libpython QNX platform support.
Note that the qnx py toolchain must come from somewhere else, this doesn't add an interpreter

### DIFF
--- a/python/private/hermetic_runtime_repo_setup.bzl
+++ b/python/private/hermetic_runtime_repo_setup.bzl
@@ -179,20 +179,16 @@ def define_hermetic_runtime_toolchain_impl(
                 "libs/python{major}{minor}t.lib".format(**version_dict),
                 "libs/python3t.lib",
             ],
-            "@platforms//os:linux": [
-                "lib/libpython{major}.{minor}.so".format(**version_dict),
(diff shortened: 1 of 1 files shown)


lalten:
@rickeylev / @aignas could you please retrigger CI? It failed due to yesterday's Docker Hub outage

python/private/hermetic_runtime_repo_setup.bzl
```diff
@@ -184,6 +184,10 @@ def define_hermetic_runtime_toolchain_impl(
                 "lib/libpython{major}.{minor}.so.1.0".format(**version_dict),
             ],
             "@platforms//os:macos": ["lib/libpython{major}.{minor}.dylib".format(**version_dict)],
+            "@platforms//os:qnx": [
```
aignas: Could you please refactor the code so that this is the default? It seems that `linux` and `qnx` have the same code here, so in theory we could just default to that. I wonder if that would incidentally also help FreeBSD, which we don't support, but you never know what PR will come in the future.

@rickeylev, what would you think about this?

lalten: done, I think that's a good idea

PR #3378, opened 2025-10-24
clean: Remove uses of native C++ implementation

comius:
All uses need to be through rules_cc, because C++ will be removed from Bazel.


aignas:
This is related to #3367

PR #3382, opened 2025-10-28
build: change gazelle rules_python override to use bazelrc

rickeylev:
This is because the `--override_module` flag in BCR presubmit configs can't affect local_path_override in MODULE.bazel.

Per https://github.com/bazelbuild/bazel-central-registry/pull/6241#issuecomment-3431422637

### DIFF
--- a/gazelle/examples/bzlmod_build_file_generation/.bazelrc
+++ b/gazelle/examples/bzlmod_build_file_generation/.bazelrc
@@ -8,7 +8,9 @@ common --experimental_enable_bzlmod
 coverage --java_runtime_version=remotejdk_11
 common:bazel7.x --incompatible_python_disallow_native_rules
 
-# NOTE: rules_python MUST be overridden using bazelrc config.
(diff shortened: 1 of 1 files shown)


gazelle/examples/bzlmod_build_file_generation/.bazelrc
```diff
@@ -7,3 +7,8 @@ common --experimental_enable_bzlmod
 
 coverage --java_runtime_version=remotejdk_11
 common:bazel7.x --incompatible_python_disallow_native_rules
+
+# NOTE: rules_python MUST be overridden using bazelrc config.
```
dougthor42: Please clarify that this is only needed for the rules_python repository itself - users should not have to implement such things. A new user will not know that BCR presubmits are specific to the release process.

rickeylev: Good point. Done.

PR #3377, opened 2025-10-23
chore(toolchain): drop all but the latest 3.9 toolchain

aignas:
Summary:
- drop the `3.9` versions one by one.
- add the latest astral toolchain builds.
- add a disclaimer to our changeleg about `3.9`.
- update examples to use the latest `3.9` build (separate PR)
  will switch to a different python version.

Work towards #2704

PR #3384, opened 2025-10-30
chore(toolchain): use 20251028 release

aignas:
Just an update to use the latest release.


aignas:
Will be done together with #3377

Issue #2704, opened 2025-03-26
Remove Python 3.8 and 3.9


aignas:
I have started the removal in #2700, but noticed that a proper removal may require more thinking about how to do this without breaking transitive deps still setting/using 3.8. Probably this ticket will involve bumping protobuf or similar deps to avoid transitive deps using 3.8.

EDIT: Since from 2025 October 3.9 is EOL, we should remove 3.9 as well.

PR #3387, opened 2025-11-01
feat(toolchain): drop 3.8 and print info level messages about it

aignas:
Before this PR we had to have at least one 3.8 toolchain to not break things.
With this we should be good to drop it.

Any python_version 3.8 registrations will be dropped if there are no actual
URLs configured, which means that 3.8 will not be selected. The same with
pip.parse, we will just ignore it and won't add it to the hub. 

In order to ensure that `is_python_3.x` flags continue working, we
just alias them to `@platforms//:incompatible`. No deprecation message is
printed.

Work towards #2704 

Next step for anyone interested and who has more time than me these days:
- [ ] Remove the 3.9 URLs and add them individually to our examples to show
      that one can do that.
- [ ] Update the examples to no longer use 3.9, because it is a maintenance burden.

### DIFF
--- a/python/private/python.bzl
+++ b/python/private/python.bzl
@@ -272,9 +272,11 @@ def _python_impl(module_ctx):
         )
         if not full_python_version:
             logger.info(lambda: (
-                "The actual toolchain for python_version '{}' ".format(toolchain_info.python_version) +
+                "The actual toolchain for python_version '{version}' " +
                 "has not been registered, but was requested, please configure a toolchain " +
                 "to be actually downloaded and setup"
+            ).format(
+                version = toolchain_info.python_version,
             ))
             continue
 


aignas:
hmmm, it looks like the situation is the same, I guess we need to start bumping the `rules_protobuf` or others to ensure that the `3.8` references are no longer there.

aignas:
Coffee thoughts:
* add `is_python_3.8` as an alias to `@platforms//incompatible`.

python/private/pypi/hub_builder.bzl
```diff
@@ -114,9 +114,25 @@ def _pip_parse(self, module_ctx, pip_attr):
             version = python_version,
         ))
 
-    self._platforms[python_version] = _platforms(
-        python_version = python_version,
+    full_python_version = full_version(
```
aignas, line 117: TODO: unit test this.

python/private/full_version.bzl
```diff
@@ -31,6 +32,8 @@ def full_version(*, version, minor_mapping):
     parts = version.split(".")
     if len(parts) == 3:
         return version
+    elif not err:
```
aignas: TODO: unit test

python/private/config_settings.bzl
```diff
@@ -109,13 +115,29 @@ def construct_config_settings(*, name, default_version, versions, minor_mapping,
...
         )
 
+    for minor in range(int(first_minor.partition(".")[-1]) + 1):
+        minor = "3.{}".format(minor)
+        if minor in minor_mapping:
+            break
+
+        # TODO @aignas 2025-11-04: use env-marker-setting with the smallest minor_mapping version
+        native.alias(
+            name = "is_python_{}".format(minor),
```
aignas: We need targets that are for other Python versions, but maybe instead of this hack to make the versions incompatible, we colud do it differently:
```
for minor in range("3.0", max(minor_mapping)):
    native.config_setting()
```

That way we'll get the right behaviour. Another option would be to leave `minor_mapping[3.8] = None`. That way we might get the right behaviour as well.

rickeylev:
> hmmm, it looks like the situation is the same, I guess we need to start bumping the rules_protobuf or others to ensure that the 3.8 references are no longer there.

Is this concern still valid? I see CI is green.

My naive thought is just remove it from versions.bzl. If people need 3.8, they can manually register it back in.

I'm overall +1 on removing an old runtime.

For the is_python3.8 flags: We consider them deprecated, in favor of people writing their own config setting based on the python_version and python_version_major_minor flags. We could start the process of removing those. They're pretty low overhead though; all they are is e.g. `config_setting(flags=":python_ersion_major_minor": "3.*")`. We could just keep an explicit list of such legacy targets and mark them deprecated, then remove them in a subsequent release

One concern I have is: will bzlmod modules break immediately? The particular case I have in mind is e.g. `pip.parse(python_version="3.8")`. I don't care if it breaks in a later phase, but we don't want that breaking during the bzlmod phase. If such is the case, maybe we can silently ignore errors or something?

aignas:
So `rules_protobuf` has a `pip.parse` as a non-dev dependency in their `MODULE.bazel` until recently, so does `rules_fuzzing`. If I just remove references from `versions.bzl`, then the CI is no longer green because of that.

This is because our `bazel mod graph` is as follows:
```
<root> (rules_python@0.0.0)
├───another_module@_
├───bazel_ci_rules@1.0.0
├───bazel_features@1.30.0
│   └───bazel_skylib@1.8.2 (*)
├───bazel_skylib@1.8.2
│   ├───platforms@0.0.11 (*)
│   └───rules_license@1.0.0
├───gazelle@0.40.0
│   ├───bazel_features@1.30.0 (*)
│   ├───bazel_skylib@1.8.2 (*)
│   ├───protobuf@29.0 (*)
│   ├───rules_go@0.55.1 (*)
│   ├───rules_license@1.0.0 (*)
│   ├───rules_shell@0.3.0 (*)
│   └───rules_proto@7.0.2
│       ├───bazel_skylib@1.8.2 (*)
│       ├───protobuf@29.0 (*)
│       ├───rules_cc@0.1.5 (*)
│       └───rules_license@1.0.0 (*)
├───other@_
│   ├───another_module@_ (*)
│   └───bazel_skylib@1.8.2 (*)
├───platforms@0.0.11
│   └───rules_license@1.0.0 (*)
├───protobuf@29.0
│   ├───bazel_features@1.30.0 (*)
│   ├───bazel_skylib@1.8.2 (*)
│   ├───platforms@0.0.11 (*)
│   ├───rules_cc@0.1.5 (*)
│   ├───rules_license@1.0.0 (*)
│   ├───rules_pkg@1.0.1 (*)
│   ├───rules_shell@0.3.0 (*)
│   ├───abseil-cpp@20240116.1
│   │   ├───bazel_skylib@1.8.2 (*)
│   │   ├───platforms@0.0.11 (*)
│   │   └───googletest@1.14.0.bcr.1
│   │       ├───platforms@0.0.11 (*)
│   │       ├───rules_cc@0.1.5 (*)
│   │       └───re2@2023-09-01
│   │           ├───platforms@0.0.11 (*)
│   │           ├───rules_cc@0.1.5 (*)
│   │           └───pybind11_bazel@2.11.1
│   │               ├───platforms@0.0.11 (*)
│   │               └───rules_cc@0.1.5 (*)
│   ├───jsoncpp@1.9.5
│   ├───rules_fuzzing@0.5.2
│   │   ├───abseil-cpp@20240116.1 (*)
│   │   ├───bazel_skylib@1.8.2 (*)
│   │   ├───platforms@0.0.11 (*)
│   │   └───rules_java@8.14.0 (*)
│   ├───rules_java@8.14.0
│   │   ├───abseil-cpp@20240116.1 (*)
│   │   ├───bazel_features@1.30.0 (*)
│   │   ├───bazel_skylib@1.8.2 (*)
│   │   ├───platforms@0.0.11 (*)
│   │   ├───rules_cc@0.1.5 (*)
│   │   ├───rules_license@1.0.0 (*)
│   │   └───zlib@1.3.1.bcr.5 (*)
│   ├───rules_jvm_external@6.3
│   │   ├───bazel_features@1.30.0 (*)
│   │   ├───bazel_skylib@1.8.2 (*)
│   │   ├───platforms@0.0.11 (*)
│   │   ├───rules_java@8.14.0 (*)
│   │   ├───rules_kotlin@1.9.6 (*)
│   │   ├───rules_license@1.0.0 (*)
│   │   └───rules_android@0.1.1
│   ├───rules_kotlin@1.9.6
│   │   ├───bazel_skylib@1.8.2 (*)
│   │   ├───platforms@0.0.11 (*)
│   │   ├───rules_cc@0.1.5 (*)
│   │   ├───rules_java@8.14.0 (*)
│   │   └───rules_proto@7.0.2 (*)
│   └───zlib@1.3.1.bcr.5
│       ├───platforms@0.0.11 (*)
│       ├───rules_cc@0.1.5 (*)
│       └───rules_license@1.0.0 (*)
├───rules_bazel_integration_test@0.27.0
│   ├───bazel_skylib@1.8.2 (*)
│   ├───platforms@0.0.11 (*)
│   ├───rules_shell@0.3.0 (*)
│   ├───buildifier_prebuilt@6.1.2
│   │   ├───bazel_skylib@1.8.2 (*)
│   │   └───platforms@0.0.11 (*)
│   └───cgrindel_bazel_starlib@0.18.0
│       ├───bazel_skylib@1.8.2 (*)
│       ├───buildifier_prebuilt@6.1.2 (*)
│       ├───gazelle@0.40.0 (*)
│       ├───platforms@0.0.11 (*)
│       ├───rules_go@0.55.1 (*)
│       └───stardoc@0.7.2 (*)
├───rules_cc@0.1.5
│   ├───bazel_features@1.30.0 (*)
│   ├───bazel_skylib@1.8.2 (*)
│   └───platforms@0.0.11 (*)
├───rules_go@0.55.1
│   ├───bazel_features@1.30.0 (*)
│   ├───bazel_skylib@1.8.2 (*)
│   ├───gazelle@0.40.0 (*)
│   ├───platforms@0.0.11 (*)
│   ├───protobuf@29.0 (*)
│   ├───rules_proto@7.0.2 (*)
│   └───rules_shell@0.3.0 (*)
├───rules_multirun@0.9.0
│   └───bazel_skylib@1.8.2 (*)
├───rules_pkg@1.0.1
│   ├───bazel_skylib@1.8.2 (*)
│   └───rules_license@1.0.0 (*)
├───rules_python_gazelle_plugin@_
│   ├───bazel_skylib@1.8.2 (*)
│   ├───gazelle@0.40.0 (*)
│   ├───rules_cc@0.1.5 (*)
│   └───rules_go@0.55.1 (*)
├───rules_shell@0.3.0
│   ├───bazel_features@1.30.0 (*)
│   ├───bazel_skylib@1.8.2 (*)
│   └───platforms@0.0.11 (*)
├───rules_testing@0.6.0
│   ├───bazel_skylib@1.8.2 (*)
│   ├───platforms@0.0.11 (*)
│   └───rules_license@1.0.0 (*)
└───stardoc@0.7.2
    ├───bazel_skylib@1.8.2 (*)
    ├───protobuf@29.0 (*)
    ├───rules_java@8.14.0 (*)
    ├───rules_jvm_external@6.3 (*)
    └───rules_license@1.0.0 (*)
```

aignas:
I think this PR shows that we should help `rules_fuzzing` to define the runtime python dependencies in a way that does not require them to hard code the python versions. Maybe users should have a way to define it for all minor versions?

rickeylev:
> a way to not hard code the version

Yeah. Or at least, so that it being in the bzlmod phase doesn't break _everything_ if a version isn't available.

I think the main problem here is the pip based integration that has to do actions at bzlmod and repo phase to answer the question of which urls to get. Which...well, since requirements.txt simply _doesn't_ have that information, there's really no way to answer those questions without activity in those phases :\. It really **requires** something like pylock.toml to do that.

(this is also ignoring the fact that protobuf shouldn't be putting those requirements in there, since it causes multi-version confusing issues. Fixing that would be another way to solve this.)

aignas:
What we will do:
* Add deprecation message to the `is_python_3.8`
* We will strive to not break the `bazel query` and `repository` phase.
* We may want to have a noop toolchain for unsupported versions. Maybe `//python:none` could work.

aignas:
PTAL

python/private/full_version.bzl
```diff
@@ -14,12 +14,13 @@
 
 """A small helper to ensure that we are working with full versions."""
 
-def full_version(*, version, minor_mapping):
+def full_version(*, version, minor_mapping, err = True):
```
dougthor42: Is `err` already a common arg name? If so, no change needed.

If not, I'd recommend `fail_on_err` instead. As is, `err` sounds like the status of whether or not an error occurred.

Or even better, invert the logic to `ignore_err: bool = False` so that the if block on L35 is more readable (positive logic):

```python
elif ignore_err:
    return None
```

python/private/pypi/hub_builder.bzl
```diff
@@ -114,9 +114,29 @@ def _pip_parse(self, module_ctx, pip_attr):
...
+    if not full_python_version:
+        # NOTE @aignas 2025-11-18: If the python version is not present in our
+        # minor_mapping, then we will not register any packages and then the
+        # select in the hub repository will fail, which will prompt the user to
+        # configure the toolchain correctly and move forward.
+        self._logger.info(lambda: (
+            "Ignoring pip python version '{version}' for hub " +
+            "'{hub}' in module '{module}' because there is no registered " +
+            "toolchain for it."
+        ).format(
```
dougthor42, line 131: Probably more readable as a `msg` variable:

```
msg = (
    "Ignoring pipi python ..."
    "'{hub}' in module ..."
    ...
)
self._logger.info(msg.format(...))
```

python/private/pypi/hub_builder.bzl
```diff
@@ -114,9 +114,29 @@ def _pip_parse(self, module_ctx, pip_attr):
...
+        version = python_version,
         minor_mapping = self._minor_mapping,
+        err = False,
+    )
+    if not full_python_version:
+        # NOTE @aignas 2025-11-18: If the python version is not present in our
+        # minor_mapping, then we will not register any packages and then the
+        # select in the hub repository will fail, which will prompt the user to
+        # configure the toolchain correctly and move forward.
+        self._logger.info(lambda: (
```
dougthor42, line 127: Why log with a lambda?

python/private/python.bzl
```diff
@@ -268,7 +268,12 @@ def _python_impl(module_ctx):
         full_python_version = full_version(
             version = toolchain_info.python_version,
             minor_mapping = py.config.minor_mapping,
+            err = False,
         )
+        if not full_python_version:
+            logger.info(lambda: "The actual toolchain for python_version '{}' has not been registered, but was requested, please configure a toolchain to be actually downloaded and setup".format(toolchain_info.python_version))
```
dougthor42: Nit: line length

python/private/pypi/hub_builder.bzl
aignas: I'll leave this as is so that it is not outside the lambda.

python/private/pypi/hub_builder.bzl
aignas: Me and @rickeylev decided to have `lambda` in the logger so that format calls do not cost any more than needed based on the logging level.

python/private/full_version.bzl
aignas: Used `fail_on_err`

python/private/pypi/hub_builder.bzl
dougthor42: Interesting, does Bazel not have the "deferred formatting" that core python logging has?

```python
logger.info("value is: %s", value)
```

The string formatting will not happen unless the log level is INFO or lower.

aignas: Hmmm, I can't remember if I had something like that prototyped earlier or not, maybe we should look into this. should not be too difficult and would definitely remove the need for doing those inline lambdas.

That said, the current way allows us to use the `render.dict` helper to print a nice dictionary, because we are passing a lambda and there will be no penalty for that.

PR #3389, opened 2025-11-03
Add batch_commands to presubmit.yml

meteorcloudy:
Fix bcr presubmit build on windows


aignas:
Thanks!

aignas:
Ah, our codebase needs to be fixed so that we can merge to `main` because (probably) bazel 9.0 release candidate was released.

Issue #3392, opened 2025-11-05
Fix CI to not fail with 9.0.0-rc1


aignas:
It started failing after the release and it was also highlighted in https://github.com/bazelbuild/bazel-central-registry/issues/6402

The `gazelle` plugin depends on `gazelle` itself. After that, we will probably bump into other `dev` dependencies not being compatible, so the best for now would be to ignore the CI failures until we fully fix it.

Definition of Done:
- [x] Unbreak the CI to allow merging by allowing failure.
- [x] Make examples with bazel 9 green (split this if necessary).
- [x] Make regular builds with bazel 9 green.
- [x] Run bazel-in-bazel integration tests with bazel9.
- [ ] Make the BCR pre-submit green.

aignas:
Done in #3393

PR #3394, opened 2025-11-05
chore: Bazel 9 is rc now

alexeagle:
Work towards #3392


aignas:
Thanks, let's merge #3393 first and rebase this and see if it needs a touch-up.

aignas (approved):

aignas:
I am thinking that it would be nice to set this as `last_rc` instead of the explicit version, but then I thought that during the `rc` phase for `9` maybe it would be better to be explicit to ensure that the CI is retriggered.

PR #3399, opened 2025-11-09
fix(local): Fix local_runtime use with free-threaded python

laramiel:
* Return abi_flags from get_local_runtime_info and pass it into the py3_runtime
* Rework how shared-libraries are links are constructed to better meet @rules_cc cc_library.srcs requirements

This improves runtime detection for macos when using a python3.14t framework runtime.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -16,7 +16,7 @@
 buildifier:
   # keep these arguments in sync with .pre-commit-config.yaml
   # Use a specific version to avoid skew issues when new versions are released.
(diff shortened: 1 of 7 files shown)


rickeylev:
Can you clarify what "Rework how shared-libraries are links are constructed to better meet @rules_cc cc_library.srcs requirements"

What requirement does it have that wasn't being properly met?

rickeylev:
Just needs a changelog entry, but otherwise LGTM

rickeylev (approved):

PR #3402, opened 2025-11-10
fix: use runfiles symlinks for venv symlink creation to reduce action count

rickeylev:
When the venv files are materialized, it can result in many symlink actions
being created. Rather than register them as regular symlink actions, batch them
into the runfiles object, which can probably handle large numbers of them more
efficiently.

Work towards https://github.com/bazel-contrib/rules_python/issues/3401

### DIFF
--- a/python/private/py_executable.bzl
+++ b/python/private/py_executable.bzl
@@ -356,7 +356,7 @@ def _create_executable(
             [stage2_bootstrap] + (
                 venv.files_without_interpreter if venv else []
             ),
-        )
+        ).merge(venv.lib_runfiles)
         zip_main = _create_zip_main(
             ctx,
             stage2_bootstrap = stage2_bootstrap,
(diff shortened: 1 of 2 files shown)


aignas (approved):

rickeylev:
As an example: When depending on torch, it reduces the number of actions by about
10,000. 

`bazel aquery --output=summary //docs:sphinx-build` before/after when torch is added as a dep

Prior to this change:
| Action | Before | After |
|--------|--------|--------|
| Symlink | 12,163 | 0 |
| RawSymlink | 98 | 98 | 
| Total actions | 12273 | 110 |

(I omitted action counts that were the same)

Build _might_ be better, but it's hard to tell (initially it seemed like before was half a second slower, but after re-running both many times, the two seem to be converging in how long they take; this is probably more sensitive to the particulars of how the two methods create the files on disk)

PR #3404, opened 2025-11-10
ci: add ci config to test 7 and 8 for bcr like setup

aignas:
Run the examples that BCR uses as tests in a more similar way to BCR.

Unfortunately, this breaks the gazelle plugin on Windows due to #3416, so testing of
it is removed.

Work towards #3392

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,6 +53,12 @@ END_UNRELEASED_TEMPLATE
 
 [0.0.0]: https://github.com/bazel-contrib/rules_python/releases/tag/0.0.0
 
+{#v0-0-0-known-issues}
+### Known Issues
+* (gazelle) Windows support for the Gazelle plugin may be broken. See
+  [#3416](https://github.com/bazel-contrib/rules_python/issues/3416) for
+  details and possible workarounds.
+
 {#v0-0-0-removed}
 ### Removed
 * (toolchain) Remove all of the python 3.8 toolchain support out of the box. Users need


aignas:
OK, getting late, but I think this is a good start.

@rickeylev, do you remember what kind of limitations we have for CI slots?

rickeylev:
There's a limit on the number of jobs. I'm not sure if its hard or soft.

Looking at the recent BCR failure, i'm guess it is covering windows+bazel 8 while we're not

.bazelci/presubmit.yml
```diff
@@ -258,114 +305,6 @@ tasks:
...
-    platform: macos
-    bazel: last_rc
-  integration_test_bzlmod_windows:
-    <<: *reusable_build_test_all
-    # coverage is not supported on Windows
-    name: "examples/bzlmod: Windows"
-    working_directory: examples/bzlmod
-    platform: windows
-    bazel: 7.x
-  integration_test_bzlmod_windows_upcoming:
```
rickeylev, line 324: Well, now I'm a bit confused.
BCR says its failing with 8.x, but this test should have been covering that (at least until the bazel 9 rc came out)?

Yet in this PR's refactored CI run, it's also giving the error BCR is giving.

The only difference seems to be the vendor step?

rickeylev:
re: CI failures: I think we just need `--windows_enable_symlinks`, then it should be OK

aignas:
I remember @meteorcloudy suggesting that we should avoid forcing users to set this flag, because it requires Windows admin permissions. If we need to add it, then the Windows support may not work properly. I think the interesting thing is that we need this flag because we are trying to use our binary in a `sh_binary`, so this requirement may be not related to `rules_python` requirements but to `rules_shell` requirements.

rickeylev:
That's a good point and would explain what we see better. Yeah, the regular tests, which also show up as symlinks on disk (even without the enable symlinks flag), work fine

I tested all the 8.x.0 versions and they all have this same issue. Bazel 7 didn't.

fmeum:
One of the failing runs tries to run `bazel coverage` on Windows with Bazel 7. That certainly isn't supported and if it worked before then only by chance. It may work in Bazel 9.

.bazelci/presubmit.yml
```diff
@@ -91,7 +91,50 @@ buildifier:
...
     - //tests:version_3_14_test
     - //tests:version_default_test
+
+matrix:
+  platform:
+    - ubuntu2204
+    - debian11
+    - macos_arm64
+    - windows
+  bazel: [7.*, 8.*, 9.*]
```
aignas:
```suggestion
  bazel: [7.*, 8.*]
```

For now let's just use 7 and 8

.bcr/gazelle/presubmit.yml
```diff
@@ -15,17 +15,21 @@
 bcr_test_module:
   module_path: "examples/bzlmod_build_file_generation"
   matrix:
-    platform: ["debian11", "macos", "ubuntu2004", "windows"]
-    bazel: [7.x, 8.x]
+    platform: ["debian11", "macos", "ubuntu2204", "windows"]
+    bazel: [7.x, 8.x, 9.x]
```
aignas:
```suggestion
    bazel: [7.*, 8.*]
```

.bcr/presubmit.yml
```diff
@@ -15,17 +15,19 @@
 bcr_test_module:
   module_path: "examples/bzlmod"
   matrix:
-    platform: ["debian11", "macos", "ubuntu2004", "windows"]
-    bazel: [7.x, 8.x]
+    platform: ["debian11", "macos", "ubuntu2204", "windows"]
+    bazel: [7.x, 8.x, 9.x]
```
aignas:
```suggestion
    bazel: [7.*, 8.*]
```

aignas:
OK, I am wondering how we can fix the current build that is saying that we need to update C++ version, but I am not sure where this is handled.

rickeylev:
`--copt=-std=c++17` should do it.

IIRC, the PR that upgrades rules_cc and protobuf ran into this, too?

rickeylev:
Is bumping deps required for rules_python? IIRC, BCR is happy (enough) with rules_python right now.

For the gazelle piece, yes, we had to bump its deps

rickeylev:
Ah, I think I see. the rules_python module depends on the gazelle plugin (dev dep)

I'm going to try and decouple them.

rickeylev:
This is almost passing. It's just gazelle+windows that is failing due to the treesitter compiling bug.

aignas:
Haha:
image
image

aignas:
What should we do?
* Attempt to fix the gazelle building here.
* Comment out the line testing this and move do this part of #3416.

rickeylev:
Lets comment out the line testing it.

rickeylev:
Aha! We can use python format syntax for parameterized parts in the name attribute. e.g. `name: {bazel}`

rickeylev:
fyi @dougthor42 -- we've had to remove testing for windows for gazelle

dougthor42:
> fyi @dougthor42 -- we've had to remove testing for windows for gazelle

ACK and SGTM. The tree sitter compilation issue is super annoying. Sadly I haven't had any more time to dive into it (and having to learn gcc args doesn't help!)

Please make sure to note in the changelog that gazelle windows ci is temporarily disabled and that gazelle windows support is questionable.

.bazelci/presubmit.yml
```diff
@@ -91,7 +91,63 @@ buildifier:
...
+    build_targets:
+      - "//..."
+      - ":modules_map"
+    test_targets:
+      - "//..."
+
+
+matrix:
+  platform:
+    - ubuntu2204
```
dougthor42, line 116: Maybe add Ubuntu 2404 too? Or is it not available in our ci?

.bazelci/presubmit.yml
```diff
@@ -91,7 +91,63 @@ buildifier:
...
+    - debian11
```
dougthor42, line 117: Add debian 12? 13 is already out also.

rickeylev:
> note in changelog

Done

dougthor42 (approved):

aignas:
I see that usually people test a particular version of the OS and I'd like to leave this for some future PR.

Source: spot checked various modules on the bazel central registry a little.

PR #3406, opened 2025-11-12
testing windows + bazel 8

rickeylev:
wip


aignas:
I think we merged a PR recently to fix the builds for BCR, do we still need this?

dougthor42:
> I think we merged a PR recently to fix the builds for BCR, do we still need this?

#3404 removed Gazelle+Windows BCR testing, so I assume yes we still need it?

aignas:
Ah, so this is for windows + bazel 8 testing for the gazelle plugin. SGTM then.

PR #3408, opened 2025-11-12
fix(local) Add api3 targets and additional defines.

laramiel:
Propagate defines and additional dll requirements for local python installs.

In get_local_runtime_info.py:
* detect abi3 vs. full abi libraries.
* Ensure that returned libraries are unique.
* Add additional dlls required by pythonXY.dll / pythonX.dll on windows.
* Add default defines for Py_GIL_DISABLED when the local python is a freethreaded install.
* Add defines (windows) for Py_NO_LINK_LIB to avoid #pragma comment(lib ...) macros

In local_runtime_repo_setup.bzl
* More closely match hermetic_runtime_repo_setup
* Add abi3 header targets.

In local_runtime_repo.bzl
* rework linking to local repository directories to handl abi3 and extra dlls.
* Update parameters passed into local_runtime_repo_setup.bzl

Before these changes, some bazel builds using local Python fail to link properly.
This happens due to a mismatch in the interpreter and the python GIL DISABLED mode, or (on Windows), where both freethreaded and non-freethreaded libraries may attempt to be linked at the same time.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -81,6 +81,8 @@ END_UNRELEASED_TEMPLATE
 * (toolchains) various local toolchain fixes: add abi3 header targets,
   fixes to linking, Windows DLL detection, and defines for free threaded
   runtimes.
+* (toolchains) The `python_headers` target is now compatible with
+  layering_check.
 * (performance) 90% reduction in py_binary/py_test analysis phase cost.
   ([#3381](https://github.com/bazel-contrib/rules_python/pull/3381)).
 * (gazelle) Fix `gazelle_python_manifest.test` so that it accesses manifest files via `runfile` path handling rather than directly ([#3397](https://github.com/bazel-contrib/rules_python/issues/3397)).
--- a/python/private/hermetic_runtime_repo_setup.bzl
+++ b/python/private/hermetic_runtime_repo_setup.bzl
@@ -126,6 +126,7 @@ def define_hermetic_runtime_toolchain_impl(
     )
     cc_library(
         name = "python_headers",
+        hdrs = [":includes"],
         deps = [":python_headers_abi3"] + select({
             "@bazel_tools//src/conditions:windows": [":interface"],
             "//conditions:default": [],


laramiel:
I'm likely to push a commit which mirrors this in tensorstore within a few hours.

Edit: The PR landed https://github.com/google/tensorstore/commit/32caaba48b402a12ec03dd3d31f43d9e72dcb765

python/private/local_runtime_repo_setup.bzl
```diff
@@ -69,44 +79,70 @@ def define_local_runtime_toolchain_impl(
...
+    )
+
+    # header libraries.
+    cc_library(
+        name = "python_headers_abi3",
+        hdrs = [":includes"],
+        includes = ["include"],
+        defines = defines,  # NOTE: Users should define Py_LIMITED_API=3
+        deps = select({
+            "@bazel_tools//src/conditions:windows": [":abi3_interface"],
```
aignas: Should we use `@platforms//os:windows` instead?

python/private/get_local_runtime_info.py
```diff
@@ -190,39 +212,51 @@ def _get_python_library_info(base_executable):
...
-        "interface_libraries": list(interface_libraries.keys()),
-        "shlib_suffix": "" if _IS_WINDOWS else shlib_suffix,
-        "abi_flags": abiflags,
+        "dynamic_libraries": _unique_basenames(dynamic_libraries),
+        "static_libraries": _unique_basenames(static_libraries),
+        "interface_libraries": _unique_basenames(interface_libraries),
+        "abi_dynamic_libraries": _unique_basenames(abi_dynamic_libraries),
+        "abi_interface_libraries": _unique_basenames(abi_interface_libraries),
+        "abi_flags": abi_flags,
+        "shlib_suffix": ".dylib" if _IS_DARWIN else "",
```
rickeylev, line 252: This line looks weird? Shouldn't shlib_suffix be .so for linux, etc ?

Previously, it was empty for windows only. Now its empty for non-darwin

python/private/local_runtime_repo_setup.bzl
```diff
@@ -69,44 +79,70 @@ def define_local_runtime_toolchain_impl(
...
+        name = "libpython",
+        hdrs = [":includes"],
+        defines = defines,
+        srcs = libraries + additional_dlls,
     )
 
+    # runtime configuration
     py_runtime(
-        name = "_py3_runtime",
+        name = "py3_runtime",
```
rickeylev, line 145: is there a particular reason these various targets were made public?

laramiel: They were changed to mimic the hermetic toolchain names. I'm not sure whether there is a benefit to one vs. the other.

python/private/get_local_runtime_info.py
laramiel: It could be.  The goal here is to allow bazel to use the python framework libraries, which don't end in the .dylib suffix (think Frameworks/Python), and need to be symlinked with the correct suffix due to the restrictions in cc_import / cc_library allowed file extensions.  For linux and windows I think that we never have to change the suffix.

PR #3411, opened 2025-11-15
docs: Explain why the `lock` rule has no implicit test target

martis42:
Fixes https://github.com/bazel-contrib/rules_python/issues/3400


aignas (approved):
Thanks!

PR #3415, opened 2025-11-16
refactor(gazelle): Generate a modules map per wheel, then merge

thejcannon:
This change internally splits modules mapping generation to be per-wheel, with a final quick "merge" action at the end.

The idea is to make this process both concurrent and cached (courtesy of Bazel), which can be ideal for codebases with a large set of requirements (as many monorepos end up doing)

Note that the `generator.py` interface changed. This seemed internal, so I didn't mark it breaking (but this change could actually just leave the generator alone, since the current implementation is fine with 1 wheel).

I ran this on the work repo and saw no change in output (but as I edited a single requirement, the overall process was fast ⚡ )

### DIFF
--- a/gazelle/modules_mapping/generator.py
+++ b/gazelle/modules_mapping/generator.py
@@ -104,8 +104,15 @@ class Generator:
                 return True
         return False
 
-    # run is the entrypoint for the generator.
-    def run(self, wheel):
+    def run(self, wheel: pathlib.Path) -> int:
+        """
+        Entrypoint for the generator.
+
+        Args:
+            wheel: The path to the wheel file (`.whl`)
+        Returns:
+            Exit code (for `sys.exit`)
+        """
         try:
             self.dig_wheel(wheel)
         except AssertionError as error:
--- a/gazelle/modules_mapping/merger.py
+++ b/gazelle/modules_mapping/merger.py
@@ -20,7 +20,7 @@ def merge_modules_mappings(input_files: list[Path], output_file: Path) -> None:
         # if there are conflicts
         merged_mapping.update(mapping)
 
-    output_file.write_text(json.dumps(merged_mapping, indent=2, sort_keys=True))
+    output_file.write_text(json.dumps(merged_mapping))
 
 
 if __name__ == "__main__":
--- a/gazelle/modules_mapping/test_merger.py
+++ b/gazelle/modules_mapping/test_merger.py
@@ -7,38 +7,54 @@ from merger import merge_modules_mappings
 
 
 class MergerTest(unittest.TestCase):
-    def test_merger(self):
-        with tempfile.TemporaryDirectory() as tmpdir:
-            d = pathlib.Path(tmpdir)
-            path1 = d / "file1.json"
-            path2 = d / "file2.json"
-            output_path = d / "output.json"
+    _tmpdir: tempfile.TemporaryDirectory
 
-            path1.write_text(
-                json.dumps(
+    def setUp(self) -> None:
+        super().setUp()
+        self._tmpdir = tempfile.TemporaryDirectory()
+
+    def tearDown(self) -> None:
+        super().tearDown()
+        self._tmpdir.cleanup()
+        del self._tmpdir
+
+    @property
+    def tmppath(self) -> pathlib.Path:
+        return pathlib.Path(self._tmpdir.name)
+
+    def make_input(self, mapping: dict[str, str]) -> pathlib.Path:
+        _fd, file = tempfile.mkstemp(suffix=".json", dir=self._tmpdir.name)
+        path = pathlib.Path(file)
+        path.write_text(json.dumps(mapping))
+        return path
+
+    def test_merger(self):
+        output_path = self.tmppath / "output.json"
+        merge_modules_mappings(
+            [
+                self.make_input(
                     {
                         "_pytest": "pytest",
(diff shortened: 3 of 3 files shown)


gazelle/modules_mapping/generator.py
```diff
@@ -106,13 +105,12 @@ def is_excluded(self, module):
...
 
     # run is the entrypoint for the generator.
-    def run(self, wheels):
-        for whl in wheels:
-            try:
-                self.dig_wheel(whl)
-            except AssertionError as error:
-                print(error, file=self.stderr)
-                return 1
+    def run(self, wheel):
```
dougthor42: Nit: arg and return types.

I know they weren't there before, but we might as well leave things better than we found it 🙂

gazelle/modules_mapping/generator.py
```diff
@@ -152,16 +150,13 @@ def data_has_purelib_or_platlib(path):
...
         description="Generates the modules mapping used by the Gazelle manifest.",
-        # Automatically read parameters from a file. Note, the '@' is the same prefix
-        # as set in the 'args.use_param_file' in the bazel rule.
-        fromfile_prefix_chars="@",
     )
     parser.add_argument("--output_file", type=str)
     parser.add_argument("--include_stub_packages", action="store_true")
     parser.add_argument("--exclude_patterns", nargs="+", default=[])
-    parser.add_argument("--wheels", nargs="+", default=[])
+    parser.add_argument("--wheel", type=pathlib.Path)
```
dougthor42, line 164: > Note that the generator.py interface changed. This seemed internal

Hmm... I agree that `generator.py` looks internal but the target is public. I don't see anything depending on it (`bazel query rdeps`), so I'm not sure why it's public. What if we supported both options until 2.0 is released?

If `--wheels` is passed, use old behavior. If `--wheel`, use new behavior. If both, error because the args are mutually exclusive.

Thoughts?

gazelle/modules_mapping/test_merger.py
```diff
@@ -0,0 +1,45 @@
...
+                    {
+                        "_pytest": "pytest",
+                        "_pytest.__init__": "pytest",
+                        "_pytest._argcomplete": "pytest",
+                        "_pytest.config.argparsing": "pytest",
+                    }
+                )
+            )
+
+            path2.write_text(json.dumps({"django_types": "django_types"}))
```
dougthor42: Would you kindly move these file creations into a `setUp()` method?

gazelle/modules_mapping/merger.py
```diff
@@ -0,0 +1,45 @@
...
+        output_file: Path where the merged output should be written
+    """
+    merged_mapping = {}
+    for input_file in input_files:
+        mapping = json.loads(input_file.read_text())
+        # Merge the mappings, with later files overwriting earlier ones
+        # if there are conflicts
+        merged_mapping.update(mapping)
+
+    output_file.write_text(json.dumps(merged_mapping, indent=2, sort_keys=True))
```
dougthor42: The original output file is not formatted nor sorted. Is there a particular reason why you've chosen to do so now?

For even moderate-size json files (1000 keys), doing both sorting and formatting slow things down by a factor of 4.5. This gets worse as the number of keys increases. Our monorepo has ~775 keys in `modules_mapping.json`, so 0.6s to 2.7s is noticeable.

dougthor42 (commented):
Nice! This looks like a great speed increase.

gazelle/modules_mapping/merger.py
dougthor42: Image

Image

gazelle/modules_mapping/generator.py
thejcannon: There's also another option here where we don't change anything at all about `generator.py`.
Or another where we don't do the `--wheel` change, but leave a TODO for it for 2.0.

I'm not really picky about any solution, since it's not my codebase 😉 
Which would you prefer?

gazelle/modules_mapping/merger.py
thejcannon: > Is there a particular reason why you've chosen to do so now?

Habit 😅

gazelle/modules_mapping/test_merger.py
thejcannon: I started to do this, but I actually think this leads to more code, and also leads to coupling.

I think moving the tempdir to `setUp`/`tearDown` makes sense. I'd even be happy to provide a helper method to perform the setup with data.

But otherwise, things like adding more tests is actually much more difficult.

Let me know if you feel strongly.

thejcannon: I ended up refactoring into `setUp`/`tearDown` plus some helpers. I think the result is neat and re-usable without being overbearing.

LMK what you think.

gazelle/modules_mapping/generator.py
dougthor42: Let's leave it as-is.

The only place `generator.py` is used is by the [`_generator` attribute's default](https://github.com/bazel-contrib/rules_python/blob/3440572f5d8bcb45df8db93ecc102475d6508bee/gazelle/modules_mapping/def.bzl#L77-L81). I think it's safe to change.

gazelle/modules_mapping/test_merger.py
dougthor42: Looks fine to me

PR #3425, opened 2025-11-22
chore(bazelrc): fix the update-deleted-packages script

aignas:
Before this we were relying on an upstream script, but because how we
are building our docs we need to customize how it is working. What is
more, the format was hard to maintain and prone to human errors.

This splits out the deleted packages code into a separate rc file that
is much easier to work with.

Split out of #3421


.bazelrc
```diff
@@ -2,16 +2,11 @@
...
+# To update the file, execute
+import .bazelrc.deleted_packages
```
aignas:
```suggestion
import %workspace%/.bazelrc.deleted_packages
```

rickeylev (approved):
Brilliant, I love it

PR #3426, opened 2025-11-22
ci: switch our jobs to mac arm64

aignas:
The arm64 macs have been in the wild for a while and it makes more sense
to test against that platform instead of the Intel macs that very few
people in our user-base are using.

At the same time clarify the supported platforms.


rickeylev (approved):

aignas:
The annoying part is that the local toolchain test is failing on an arm64 due to an incorrect interpreter being pulled in. I suspect that there is some sort of extra stuff going on with the setup of the runners. Should we investigate that? For now I have reverted it to run on intel instead.

rickeylev:
Posterity, link to failure: https://buildkite.com/bazel/rules-python-python/builds/13694/steps/canvas?sid=019aaace-eacc-4858-befc-f59d10348858

Oh, I think this is an easy fix. ISTR hard-coding a os -> python-build-standalone url. That probably just needs to be updated

rickeylev:
I pushed fixes for the local toolchain case.

Issue #2948, opened 2025-06-01
Setup fewer wheel repos in pip.parse


aignas:
Currently we create one repository per python version per wheel repo in any hub that is configured with `experimental_index_url`.

If `pipstar` is enabled by default, we can instead just create a single repository.

TODO:
- [ ] Have a default interpreter that we use to extract wheels if `pipstar` is enabled.
- [ ] Create a single repo for a wheel file when `pipstar` and `index_url` is used no matter how many python versions are used.
- [ ] (optional) extract the wheels without Python dependencies
    - [ ] Extract the wheel using `download_and_extract` (probably depends on the venv code to be stabilized).
    - [ ] Parse the `entry_points.txt` in starlark to create `py_console_script_binary` targets and have everything backwards compatible.

This is just a laundry list of things related, so in the future I may split the ticket further. Just wanted to dump my brain a little and transfer some paper notes to the ticket tracker.

PR #3429, opened 2025-11-23
refactor(pypi): parse entry_points without Python

aignas:
This is a small utility function to get us Python free when wheels are
extracted in the repository phase.

Next is to extract the wheel using `repository_ctx.extract` (#3430).

Whereas patching the wheel after extracting is more involved to be done
without Python because we need to rezip the wheel and that has to be done
with Python for a few reasons (to stay sane). If we want to remove this,
then we would have to create a `whl` file in the build phase, which could
work, but will need to be an exercise for the reader.

Nevertheless, this moves us towards removing any side-effects from Python
interpreter, so changing the default interpreter would not cause us to
refetch everything.

### DIFF
--- a/python/private/pypi/whl_metadata.bzl
+++ b/python/private/pypi/whl_metadata.bzl
@@ -143,10 +143,11 @@ def _parse_entry_points(contents):
         if not start:
             continue
 
-        if start and not line:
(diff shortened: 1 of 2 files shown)


aignas:
Work towards #2948

PR #3430, opened 2025-11-23
refactor(pypi): extract the wheel without python

aignas:
With this we start extracting the wheel without Python and
it becomes a requirement only when patching (we will extract
the wheel without Python, patch it and then re-compress it
which makes a very inefficient process).

This should result in much faster executions because we can start
extracting wheels even before we fetch the entire Python toolchain
and we don't need to fetch it in a wheel-only setup until we are
actually building/executing tests. What is more bazel is faster
in extracting everything.

Work towards #2948

### DIFF
--- a/tests/pypi/hub_builder/hub_builder_tests.bzl
+++ b/tests/pypi/hub_builder/hub_builder_tests.bzl
@@ -1026,7 +1026,6 @@ git_dep @ git+https://git.server/repo/project@deadbeefdeadbeef
             "config_load": "@pypi//:config.bzl",
             "dep_template": "@pypi//{name}:{target}",
             "filename": "plat-pkg-0.0.4-py3-none-linux_x86_64.whl",
-            "python_interpreter_target": "unit_test_interpreter_target",
             "requirement": "plat_pkg==0.0.4",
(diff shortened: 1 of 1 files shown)


rickeylev:
Nice.

A bonus of this is it'll make it easier to use an alternative extractor (it can just extract instead of having to deal with wheel format details)

rickeylev:
I would trigger merge, but this is set as draft -- overall it LGTM. Is there some edge case the python code handles that a plain extract doesn't?

rickeylev:
I renamed the title; this is purely internal and I don't see why it would be user-notable

aignas:
OK, I updated the `hub_builder` code and a few other paths so that we are not even passing the python interpreter to the `whl_library` so that we don't need to fetch the interpreter to extract the whls. The interpreter is only used in the `sdist` building if any of that happens or when we are using `pip`.

Let me know what kind of extra tests we should have here.

tests/pypi/hub_builder/hub_builder_tests.bzl
aignas:
```suggestion
```

PR #3431, opened 2025-11-24
refactor(pypi): use py_console_script_binary in whl_library_targets

aignas:
With this the `whl_library` with pipstar enabled is doing very little logic for
whls in the repo phase. This makes it more sustainable and we can start
trimming the fat away.

TODO:
- [ ] Fix tests

Work towards #2948


aignas:
I will leave this as is because it is low priority, but wanted to get the idea out there. If somebody needs this, let me know.

Issue #3439, opened 2025-12-06
Runfiles files under `.venv/site-packages` can double in size


shayanhoshyari:
# 🐞 bug report

### Affected Rule

`py_binary` with

```
common --@rules_python//python/config_settings:bootstrap_impl=script
common --@rules_python//python/config_settings:venvs_site_packages=yes
common --@rules_python//python/config_settings:venvs_use_declare_symlink=yes
```

### Is this a regression?

Hard to tell. I think it might be unintended behavior.

### Description

For building docker images, it is very common to make .tar files.

Currently, `tar.bzl` (used for aspect_rules_py, rules_oci, ... family) does not support symlinks, but there an ongoing thread about it: <https://github.com/bazel-contrib/tar.bzl/issues/16>

A suggestion by @rickeylev is to use the `File.is_symlink` field to determine if a file is a symlink and just write as-is as symlink and preserve the relative path.

I did implement a basic tar rule equivalent based on this idea at: <https://github.com/bazel-contrib/rules_python/issues/3388#issuecomment-3615363362>

However, with current result, the torch `.so` files return `File.is_symlink = False`, so we will end up writing the file to the tar twice (once the external runfile and once the one under .venv), and it can double the size of the tar (image layer). 

**I can workaround this in my tar rule, but was wondering if it is intended behavior in first place**.

## 🔬 Minimal Reproduction

https://github.com/shayanhoshyari/issue-reports/tree/main/rules_python/venv_tar

## 🔥 Exception or Error

No exception, but files like this are returning `File.is_symlink = False`

```
- path: bazel-out/darwin_arm64-fastbuild/bin/_test.venv/lib/python3.13/site-packages/torch/_C.cpython-313-darwin.so
- short_path: _test.venv/lib/python3.13/site-packages/torch/_C.cpython-313-darwin.so
- is_symlink = False
- target: /private/var/tmp/_bazel_hoshyari/0915edf9aa0fc6e009ba2664be204381/execroot/_main/bazel-out/darwin_arm64-fastbuild/bin/_test.venv/lib/python3.13/site-packages/torch/_C.cpython-313-darwin.so
```

Other files (that are just folders and don't have file resolution) correctly have is_symlink = True
```
- path: bazel-out/darwin_arm64-fastbuild/bin/_test.venv/lib/python3.13/site-packages/filelock
- short_path: _test.venv/lib/python3.13/site-packages/filelock
- is_symlink = True
[...]

PR #3440, opened 2025-12-06
[not for merge] Test a change for #3439

shayanhoshyari:
Experiment for https://github.com/bazel-contrib/rules_python/issues/3439

### DIFF
--- a/python/private/venv_runfiles.bzl
+++ b/python/private/venv_runfiles.bzl
@@ -33,9 +33,6 @@ def create_venv_app_files(ctx, deps, venv_dir_map):
         {type}`struct` with the following attributes:
         * {type}`list[File]` `venv_files` additional files created for
           the venv.
-        * {type}`dict[str, File]` `runfiles_symlinks` map intended for
-          the `runfiles.symlinks` argument. A map of main-repo
-          relative paths to File.
     """
 
     # maps venv-relative path to the runfiles path it should point to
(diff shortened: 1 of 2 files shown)


python/private/venv_runfiles.bzl
```diff
@@ -49,31 +46,29 @@ def create_venv_app_files(ctx, deps, venv_dir_map):
...
-                    # dirname is necessary because a relative symlink is relative to
-                    # the directory the symlink resides within.
-                    from_ = paths.dirname(venv_link_rf_path),
-                    to = link_to,
-                )
-                ctx.actions.symlink(output = venv_link, target_path = rel_path)
-                venv_files.append(venv_link)
+            # Just for demonstration purposes,
+            # there is definitely a better thing to do upstream that removeprefix("../") 
+            link_to_path = link_to.short_path.removeprefix("../") if is_file(link_to) else link_to
```
rickeylev, line 57: If you want to carry a patch, then I suggest calling runfiles_root_path, something like this:
```
link_to = runfiles_root_path(ctx, link_to.short_path) if is_file(link_to) else link_to
```

The case not being handled is if the File being mapped into the venv is coming from the main repository. In that case, short_path is `foo/bar.txt`, but the runfiles root path is e.g. `_main/foo/bar.txt`. Most things that go into the venv come from external repos, so you rarely see entries without `../` in their short path.

shayanhoshyari:
> Per discussion in #3439, we have to keep the runfiles.symlinks logic for performance reasons.

+1

Issue #3434, opened 2025-11-25
Some questions around `experimental_index_url` behavior


FrankPortman:
Opening a ticket to dump in all questions for triage, per suggestion of Slack message.

In this thread https://github.com/bazel-contrib/rules_python/issues/2949#issuecomment-3448011820 it was suggested that `experimental_index_url` would fix any issues I was seeing with `RULES_PYTHON_ENABLE_PIPSTAR=1`. That turned out to be true, but I saw a few things that I wanted clarification on.

### Setup

We use a "universal" lock-file generated via `uv` that looks like this:

```
triton==3.4.0 ; python_full_version < '3.13' and platform_machine == 'x86_64' and sys_platform == 'linux' \
    ...
```

and then in `pip.parse` we don't specify requirements_by_platform as a result:

```python
pip.parse(
    enable_implicit_namespace_pkgs = True,
    experimental_requirement_cycles = { ... bunch of cycles ... },
    hub_name = "core_py_deps",
    python_version = "3.12.3",
    requirements_lock = "//:3rdparty_python_requirements.txt",
)
```

### Questions/Concerns from the switch

Switching allowed my `genquery` based tests to work alongside `pipstar`, but I was hoping to get clarification on the following:

* One of my tests had this line `r.Rlocation("rules_python++pip+core_py_deps_312_attrs/attrs-25.4.0-py3-none-any.whl")` which I had to switch to `r.Rlocation("rules_python++pip+core_py_deps_312_attrs_py3_none_any_adcf7e2a/attrs-25.4.0-py3-none-any.whl")`. I found the new path by manually `cquery`ing in CLI and am not sure why this is happening or what the idiomatic path forward is. Interestingly this new path works on both OSX and Linux.
* On the branch with `experimental_index_url`, I noticed it was downloading all the linux-only torch/cuda wheels. I don't *think* this happens normally, but I am not an expert. The build still works on Mac, but this is still a little concerning since those wheels are massive. Am I thinking about this the wrong way? I can confirm that in a different branch output_base, the `external` folder doesn't have non OSX wheels. But in the `experimental_index_url` branch, it does.
* `bazel query deps(...)` now fails because it is trying to install a Windows-only package (pywinpty). That package is in my universal lockfile with `os_name == 'nt'`

aignas:
[...]

PR #3441, opened 2025-12-06
feat(pip.parse): limit the target platforms we parse requirements for

aignas:
Up until now the users can configure which requirements files to be used
for specific platforms, however, what they cannot configure is what
target platforms should actually be set up.

The difference in the problems is:
1. I want my `bazel build` to work on `osx aarch64` and `linux x86_64`.
1. I want my `bazel build` to build for `linux x86_64` on `osx aarch64`.

With the newly introduced `target_platforms` attribute users can finally
specify their target platforms. To ensure that this also allows users to
specify that they want to support `freethreaded` and `non-freethreaded`
platforms at the same time we support `{os}` and `{arch}` templating in
the strings.

This should fix the `genquery` usage pattern breakage when we previously
enabled `RULES_PYTHON_ENABLE_PIPSTAR=1`.

Work towards #2949
Work towards #3434

### DIFF
--- a/python/private/pypi/extension.bzl
+++ b/python/private/pypi/extension.bzl
@@ -682,7 +682,8 @@ function in general are the same as used in the {obj}`pip.default.platform` attr
 If you only care for the host platform and do not have a usecase to cross-build, then you can put in
 a string `"{os}_{arch}"` as the value here. You could also use `"{os}_{arch}_freethreaded"` as well.
 
-EXPERIMENTAL: this may be removed without notice.
+:::{include} /_includes/experimental_api.md
+:::
 
 :::{versionadded} VERSION_NEXT_FEATURE
 :::


python/private/pypi/extension.bzl
```diff
@@ -667,6 +667,25 @@ EXPERIMENTAL: this may be removed without notice.
...
+only evaluate for a particular platform (e.g. "linux_x86_64"), then put it in here.
+
+If you want `freethreaded` variant, then you can use `_freethreaded` suffix as `rules_python` is
+defining target platforms for these variants in its `MODULE.bazel` file. The identifiers for this
+function in general are the same as used in the {obj}`pip.default.platform` attribute.
+
+If you only care for the host platform and do not have a usecase to cross-build, then you can put in
+a string `"{os}_{arch}"` as the value here. You could also use `"{os}_{arch}_freethreaded"` as well.
+
+EXPERIMENTAL: this may be removed without notice.
```
rickeylev:
```suggestion
:::{include} /_includes/experimental_api.md
:::
```

tests/pypi/extension/pip_parse.bzl
```diff
@@ -41,7 +42,9 @@ def pip_parse(
         envsubst = envsubst,
         experimental_index_url = experimental_index_url,
         experimental_requirement_cycles = experimental_requirement_cycles,
+        # TODO @aignas 2025-12-02: decide on a single attr - should we reuse this?
```
rickeylev, line 45: What's the difference in behavior? Both are a list of platform names (as defined by the pip.platforms() configuration). Is it just that experimental_target_platforms has the `*` matching, and this one doesn't?

rickeylev:
Doc updates would be good to help show how to use this, but that can be done in a separate PR

tests/pypi/extension/pip_parse.bzl
aignas: Correct. `experimental_target_platforms` also is for `pipstar=False` case. I think we should probably remove that attribute altogether when `pipstar=True` by default.

PR #3442, opened 2025-12-06
refactor(core): get_zip_runfiles_path should call startswith less

aignas:
Looking at the investigation in #3380, it seems that we are calling
the startswith many times and I wanted to see if it would be possible
to optimize how it is done.

I also realized that no matter what target we have, we will be calling
the function once with a `__init__.py` path and we can inline this case
as a separate if statement checking for equality instead, which Starlark
optimizer should understand better.

Before this PR for every executable target we would go through the
`legacy_external_runfiles and "__init__.py".startswith("external")` and
this PR eliminates this.

Related to #3380 and #3381

### DIFF
--- a/python/private/py_executable.bzl
+++ b/python/private/py_executable.bzl
@@ -70,6 +70,7 @@ load(":venv_runfiles.bzl", "create_venv_app_files")
 _py_builtins = py_internal
 _EXTERNAL_PATH_PREFIX = "external"
 _ZIP_RUNFILES_DIRECTORY_NAME = "runfiles"
+_INIT_PY = "__init__.py"
 _LAUNCHER_MAKER_TOOLCHAIN_TYPE = "@bazel_tools//tools/launcher:launcher_maker_toolchain_type"
 
 # Non-Google-specific attributes for executables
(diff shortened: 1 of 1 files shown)


aignas:
@tobyh-canva, could you please check the profile for this PR? https://github.com/bazel-contrib/rules_python/pull/3442

I suspect that there should be an improvement, but not sure how large. At least for setups that do not use zip builds and do not have the legacy "external" code path active it should eliminate all appearances of the `startswith` in the trace.

python/private/py_executable.bzl
```diff
@@ -834,14 +835,14 @@ def _create_zip_file(ctx, *, output, zip_main, runfiles):
...
-        "{}=".format(
-            _get_zip_runfiles_path("__init__.py", workspace_name, legacy_external_runfiles),
-        ),
+        "{}=".format(_get_zip_runfiles_path(_INIT_PY, workspace_name)),
     )
 
     def map_zip_empty_filenames(list_paths_cb):
         return [
-            _get_zip_runfiles_path(path, workspace_name, legacy_external_runfiles) + "="
+            # FIXME @aignas 2025-12-06: what kind of paths do we expect here? Will they
```
rickeylev, line 843: I would expect the paths to match what you'd see for a file in `runfiles.files`.

e.g., if a file in the runfiles comes from an external repo (and would have a short path of `external/repo/foo.py` or `../repo/foo.py`), then there would be an empty file with a similar short path.

PR #3446, opened 2025-12-07
fix: Bump minimum protobuf version to 31.0 to fix dropped 3.8 support

Ahajha:
See https://github.com/jvolkman/rules_pycross/issues/209

The old protobuf version still needs Python 3.8, so attempting to load it fails. I see there is already a PR here (https://github.com/bazel-contrib/rules_python/pull/3367) but this is a much smaller change for an isolated issue.

### DIFF
--- a/docs/pypi/download.md
+++ b/docs/pypi/download.md
@@ -113,6 +113,21 @@ This will not work for sdists with C extensions, but pure Python sdists may stil
 approach.
 :::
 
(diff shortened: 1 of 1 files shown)


aignas:
Why is rules_pycross using python3.8? It would be great if you could a little bit explain the breakage to me here.

Ahajha:
Rules_pycross isn't, it's just that it manifested there somehow. The issue is that the older version of protobuf is using 3.8, which was dropped in later versions.

PR #3448, opened 2025-12-09
refactor: avoid conflict merging when shared libraries are present

rickeylev:
Today, when a shared library is present, an extra VenvSymlinkEntry is generated so that
it is linked directly. Unfortunately, this will always have a path overlap conflict with
the rest of the venv symlinks, which triggers the conflict merge logic later in
py_executable. That logic is expensive, as it must flatten all the files and then
link each file individually (essentially doubling the number of files materialized).
For large packages like torch (10k+ files), this can dramatically increase overhead.

To fix, generate VenvSymlinkEntries that don't overlap. The basic logic for how this works
is to identify paths that *must* be directly linked, marking all their parent directories
as not being able to be directly linked, and then grouping what remains into the highest
directly-linkable path.

Along the way, drop the logic that only considers code files and special cases `__init__.py`
files and implicit packages. This is simplify the code and more correctly map the
extracted wheel into the venv.

### DIFF
--- a/python/private/pypi/extension.bzl
+++ b/python/private/pypi/extension.bzl
@@ -14,7 +14,6 @@
 
 "pip module extension for use with bzlmod"
 
-load("@bazel_features//:features.bzl", "bazel_features")
 load("@pythons_hub//:interpreters.bzl", "INTERPRETER_LABELS")
 load("@pythons_hub//:versions.bzl", "MINOR_MAPPING")
 load("@rules_python_internal//:rules_python_config.bzl", rp_config = "config")
@@ -385,13 +384,9 @@ def _pip_impl(module_ctx):
             groups = mods.hub_group_map.get(hub_name),
         )
 
-    if bazel_features.external_deps.extension_metadata_has_reproducible:
-        # NOTE @aignas 2025-04-15: this is set to be reproducible, because the
-        # results after calling the PyPI index should be reproducible on each
-        # machine.
-        return module_ctx.extension_metadata(reproducible = True)
-    else:
-        return None
+    return module_ctx.extension_metadata(
+        reproducible = True,
+    )
 
 _default_attrs = {
     "arch_name": attr.string(


tests/venv_site_packages_libs/app_files_building/app_files_building_tests.bzl
```diff
@@ -294,34 +456,33 @@ def _test_shared_library_symlinking_impl(env, target):
...
-        "bar/x.py": srcs[1],
-        "bar/y.so": srcs[2],
-        "foo": "_main/tests/venv_site_packages_libs/app_files_building/site-packages/foo",
         "foo.libs/libx.so": srcs[3],
-        "root.pth": srcs[-3],
-        "root.py": srcs[-2],
-        "root.so": srcs[-1],
     }
-    env.expect.that_dict(actual[VenvSymlinkKind.LIB]).contains_exactly(expected_libs)
+    env.expect.that_dict(actual[VenvSymlinkKind.LIB]).contains_at_least(expected_libs)
```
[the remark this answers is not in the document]
rickeylev: The test is only interested in knowing that the shared libraries were linked correctly. Other tests verify overall linking.

rickeylev:
re: external files are now skipped: This case wasn't originally intended. It might partially work, but only as long as there's no overlap between the venv paths the repos want to use. If there is, then one repo's files will "disappear"

e.g., given:

```
repo1/foo/a.py
repo2/foo/b.py
```

The old logic would think that "foo" was a common parent directory, and then create a link to the runfiles path for whichever came first (or last, I don't recall). e.g. `foo -> <...>/repo1/foo` (which makes b.py inaccessible).

It's an interesting use case to support because it would allow more efficiently mapping arbitrary extracted wheels into a single location. e.g. `py_library(srcs=["@foo//:all_files", "@foo_plugin//:all_files")` (it's more efficiently because the "conflict merging" happens at the py_library target level, so fewer files and without any depset flattening).

But, it seems somewhat niche.

rickeylev:
re: external files are skipped: While fixing some failures with workspace mode, I mostly brought this back to its prior behavior, I think. External files are no longer explicitly skipped. I _think_ having a mixture of repo sources will work so long as the different groupings are for a single repo. Like before, the new code doesn't make any special effort to handle this case, i.e. its not explicitly supported.

python/private/venv_runfiles.bzl
```diff
@@ -235,109 +241,121 @@ def get_venv_symlinks(ctx, files, package, version_str, site_packages_root):
...
+                else:
+                    break
+
+            optimized_groups.setdefault(venv_path, [])
+            optimized_groups[venv_path].append(src)
+
+    # Finally, for each group, we create the VenvSymlinkEntry objects
+    for venv_path, files in optimized_groups.items():
+        if venv_path in venv_symlinks:
+            fail((
```
aignas: Could you please comment here when this would/could happen? Is it guarding against a bug?

rickeylev: Oops, removed. It was some debugging/guard code from an earlier iteration. I don't see how a duplicate could occur with the current code.

PR #3450, opened 2025-12-10
build: add  flag_alias definition for Starlarkification of python flags

aranguyen:
With this change, when user build with a native python flag, `flag_alias` will point to the starlark version of the flag. Please see design doc [here](https://docs.google.com/document/d/1yOvi4hVV7Ja32ocwVb4lsEUnijftk8nilXPncYm-BH8/edit?tab=t.0#heading=h.qn3unswby87l) for more details.

This feature should work as a no-op in the latest bazel 8.x and 7.x releases.


rickeylev:
Is flag_alias bazel 9+? All the CI jobs are failing because it's not available.

At the least, we need to support 8.x.

We can drop 7.x pretty soon -- I'm thinking the Jan release?

gregestren:
This should work as a no-op in the latest bazel `8.x` and `7.x` releases. 

i.e. they continue to read native flags as usual, so they parse these new entries but just skip over them.

gregestren:
See https://github.com/bazelbuild/bazel/issues/27337 and https://github.com/bazelbuild/bazel/issues/27334.

rickeylev:
> flag_alias() should work with latest 7.x and 8.x releases

Oh ok, that explains. Thanks. I updated the PR and it's using 8.5, but it still fails, error:

```
ERROR: /workdir/MODULE.bazel:460:11: Traceback (most recent call last):
	File "/workdir/MODULE.bazel", line 460, column 11, in <toplevel>
		flag_alias("build_python_zip", "//python/config_settings:build_python_zip")
Error in flag_alias: flag_alias() got unexpected positional argument
```

gregestren:
@aranguyen I think you want to give the parameters names to fix that.

aignas:
Please also add a changelog note stating the minimum supported bazel versions coming with this change.

MODULE.bazel
```diff
@@ -458,3 +458,8 @@ uv_dev = use_extension(
 uv_dev.configure(
     version = "0.6.2",
 )
+
+flag_alias(name = "build_python_zip", starlark_flag = "//python/config_settings:build_python_zip")
+flag_alias(name = "incompatible_default_to_explicit_init_py", starlark_flag = "//python/config_settings:incompatible_default_to_explicit_init_py")
+flag_alias(name = "python_path", starlark_flag = "//python/config_settings:python_path")
+flag_alias(name = "experimental_python_import_all_repositories", starlark_flag = "//python/config_settings:experimental_python_import_all_repositories")
```
shayanhoshyari: will this work for users of rules_python too?
otherwise I think this issue still stands : https://github.com/bazel-contrib/rules_python/issues/3252#issuecomment-3726123681

gregestren: That should work today, even before this PR is submitted.

Bazel 9 hard-codes those aliases right now. Once this PR works we'll remove the hard-coded logic from bazel@head.

The intention is users should have a seamless experience regardless of what's going on behind the scenes.

shayanhoshyari: Thanks! I read https://docs.google.com/document/d/1yOvi4hVV7Ja32ocwVb4lsEUnijftk8nilXPncYm-BH8/edit?tab=t.0#heading=h.mfw8nr4w6amb briefly, and as you say this trickles down to users too, and not just rules_python. Makes sense.

Now another question, by adding this PR, should this code just be removed now? https://github.com/bazel-contrib/rules_python/blob/144984ccbd61f3459dd2a1f86210bd2aab69043f/python/private/flags.bzl#L39-L73  (from https://github.com/bazel-contrib/rules_python/pull/3290) and just the rules_python config be read instead? Because the `flag_alias` is already doing (and even a better job) than what this code is trying to do.

gregestren: That should still be necessary so `rules_python` remains compatible with bazel 7 and 8, which can't resolve the flag aliases. Whenever `rules_python` drops support for pre-9 bazel we can remove this code.

tests/base_rules/py_executable_base_tests.bzl
```diff
@@ -93,7 +93,7 @@ def _test_basic_zip(name, config):
             # platforms.
             # Pass value to both native and starlark versions of the flag until
             # the native one is removed.
-            "//command_line_option:build_python_zip": "true",
+            "//python/config_settings:build_python_zip": "true",
```
shayanhoshyari: @gregestren sorry for ping again. Are you sure this is a correct change on bazel 7 and 8?

My understanding is that 

- you mentioned flag alias is no op on these versions.
- and the code here https://github.com/bazel-contrib/rules_python/blob/144984ccbd61f3459dd2a1f86210bd2aab69043f/python/private/flags.bzl#L39-L73 will always pick the py_fragments values and ignore the rules_python configs for bazel 7 and 8.

So the result is this change to overrides on these bazel versions will make the override stop overriding anything, breaking the test.

I see in CI failures like below that probably is related

```
(18:22:56) ERROR: /workdir/tests/config_settings/transition/BUILD.bazel:6:25: On dependency edge //tests/config_settings/transition:test_py_binary_windows_build_python_zip_false (722907e) -|target|-> //tests/config_settings/transition:test_py_binary_windows_build_python_zip_false_subject: com.google.devtools.build.lib.packages.BuildFileNotFoundException: no such package '@@rules_testing//python/config_settings': BUILD file not found in directory 'python/config_settings' of external repository @@rules_testing. Add a BUILD file to a directory to mark it as a package.
(18:22:56) ERROR: Analysis of target '//tests/config_settings/transition:test_py_binary_windows_build_python_zip_false' failed; build aborted
```

aranguyen: Hi @shayanhoshyari , taking a closer look at the error `no such package '@@rules_testing//python/config_settings': BUILD file not found in directory 'python/config_settings' of external repository @@rules_testing` 

I'm wondering if this is just changing this from `"//python/config_settings:build_python_zip": "true",` to `"@@rules_python//python/config_settings:build_python_zip": "true",`?

The code seems to be looking for this flag definition in package @@rules_testing

aranguyen: and probably update `@@rules_testing` to depends on the newer version of `@rules_python` that has these flag def? https://github.com/bazelbuild/rules_testing/blob/master/MODULE.bazel#L19

rickeylev (approved):

gregestren:
I think I was able to reproduce the previous https://buildkite.com/bazel/rules-python-python/builds/14560 failure in a local client with:

```sh
$  USE_BAZEL_VERSION=last_rc bazelisk build --nobuild //tests/py_zipapp:system_python_zipapp
ERROR: Analysis of target '//tests/py_zipapp:system_python_zipapp' failed; build aborted: no such package '@@rules_python//python/config_settings': The repository '@@rules_python' could not be resolved: Repository '@@rules_python' is not defined
```

But the last commit auto-merged. What're we seeing here?

rickeylev:
Maybe something with auto loading?

gregestren:
Building a `py_binary` has no problem:

```sh
$ USE_BAZEL_VERSION=last_rc bazelisk build //tests/py_zipapp:venv_bin
Target //tests/py_zipapp:venv_bin up-to-date:
  bazel-bin/tests/py_zipapp/venv_bin
```

`system_python_zipapp` above is a `py_zipapp_binary` [with a transition](https://github.com/bazel-contrib/rules_python/blob/c91d9f026c8d10b1807a3ad5cc6dc2027a65e98e/python/private/zipapp/py_zipapp_rule.bzl#L356). I wonder if transition logic is having an issue?

@rickeylev should presubmit have caught that?

rickeylev:
I recall running into a similar error with a bcr gazelle test. The fix was to bump the rules_python version it was using. I don't see how that would happen in this case.

The directory or existing is very strange. It makes me think more that theres a typo or syntax error in a labe somewhere. Or some issue with repo visibility? But usually those errors look very different

gregestren:
` USE_BAZEL_VERSION=last_green` fixes that error. We have some ideas of what might contribute the difference, like https://github.com/bazelbuild/bazel/pull/28506. So we might need to wait for Bazel 9.1.0.

FWIW I couldn't reproduce [the CI result](https://buildkite.com/bazel/rules-python-python/builds/14560/steps/canvas?sid=019c8341-eca0-4bd9-82e4-5998a5db557d&tab=output) in my local workspace. Even if I sync to before this PR (what should be a green baseline) I get a bunch of random errors.

gregestren:
@aranguyen and I did some more Bazel code adjustments that I think will help.

I'm hoping / anticipating Bazel 9.0.1 will look especially clean with this: https://github.com/bazelbuild/bazel/issues/28387.

rickeylev:
> should the presubmit have caught this

Yeah, it should have. I'm seeing it as an error in presubmits now, where 9.0.0rc6 ends up selected for the "upcoming bazel" CI job (bazel version = last_rc).

I'm going to revert the flag_alias() changes here for now, since transitions are used pretty thoroughly in the rules now.

PR #3454, opened 2025-12-11
refactor: optimize venv building for namespace packages

rickeylev:
When implicit namespace packages are used, it's common for multiple distributions
to install into the same directory, triggering the expensive conflict merging
logic. This can be observed wit our doc builds, where `sphinxcontrib` is a
namespace package that 7 distributions install into.

To fix, treat top-level directories that have an importable name and don't have an
`__init__` looking file as implicit namespace packages and mark them as disallowed
from being directly linked. The importable name check is to exclude dist-info 
directories.


rickeylev:
Ah drat. The regex_match function is in Bazel 7, but not exposed. Well, no biggie, we can just check for `.` and `-` if that function isn't present; should work well enough.

rickeylev:
There we go, fixed.

rickeylev:
Can't merge this yet; need to check on a bug it might introduced where top-level files that are merged via conflicts aren't being linked correct anymore i.e.  `nvidia/__init__.py` is being created as a directory that contains all conflicting entries.

rickeylev:
Well, there is a bug, but its not due to this PR. I'm not sure if its pre-existing or was introduced by https://github.com/bazel-contrib/rules_python/pull/3448.

In any case, right now, the conflict merge logic won't produce the correct result if the same file is provided twice. It treats all the entries as directories, so tries to group the conflicting files under a directory named after the file. In practice, this would mostly only affect pkgutil namespace packages with non-trivial logic in `__init__.py`; it's uncommon for different distributions to overlap with other files.

rickeylev:
Sent https://github.com/bazel-contrib/rules_python/pull/3458 to fix that bug

rickeylev:
dependent PR merged, PTAL

aignas (approved):

python/private/venv_runfiles.bzl
```diff
@@ -13,6 +13,7 @@ load(
     "VenvSymlinkEntry",
     "VenvSymlinkKind",
 )
+load(":py_internal.bzl", "py_internal")
```
aignas, line 16: nit: don't we need to add the BUILD.bazel bzl_library dep for this?

rickeylev: Oops, yes. Sent https://github.com/bazel-contrib/rules_python/pull/3462 to fix

Issue #3401, opened 2025-11-10
venv site-packages building requires millions of actions


rickeylev:
On slack, a user reported that using the venv_site_packages=yes flag caused millions of actions to occur. More user details would help verify its due to the flag.

That said but millions of actions is believable if there are many large targets and the conflict detection ends up having to recreate every file under site-packages. e.g. Given 100 tests (reasonable number), and 50,000 files in their deps (large, but not unheard of), then you end up with 5 million actions. No matter what, bazel is going to have to materialize 5 million files in such a case, but if we have to, at the analysis phase, perform 5M symlink() re-declarations, that's a decent amount of overhead.

At the least, we need to optimize the conflict resolution logic. When I fixed it to handle shared libraries better, I just did a simple and naive implementation. A more optimized version would look for a new minimal set of directories/files to symlink instead of just symlinking all of them. Similarly, runfiles.root_symlinks might come in handy here to avoid most symlink()-to-File action calls.

From slack:

> 
> Kris Wilson
>   Oct 19th at 8:55 PM
> this also seems to cause the number of actions for a bazel test //... to shoot up into the multi-millions.. which is strange.  is this creating one action per symlinked file in the venv? (edited)

rickeylev:
I think this is about as optimized as possible for now, so closing. The two possible improvements I can think of are:

1. Apply the same "minimal number or links" logic to the conflict-merge logic (more actions at the expense of fewer total files
2. Have a tool run to create all the symlinks in one action instead of many separate starlark declare_symlink+symlinks action

For 1, conflicts should be much less common now, so idk how much benefit it'll bring.

For 2, idk if it would bring meaningful benefits because we still have to flatten depsets in the analysis phase (this allows better incremental dependency tracking)

A more fuzzy optimization I had in mind was to propagate a smaller depset of known (or likely to be, or expensive to resolve) paths, then have py_library do a smaller conflict resolution to avoid a larger resolution later. The case i had in mind being: say you had a single modular python package (eg a monorepo), but want a site-packages layout.

In any case, optimizations to consider later, if need be.

FrankPortman:
I am not sure whether I am an outlier, but on `2.0.0`, I tried to flip this flag again and the total actions on my polyglot repo went from 18k to millions as well. Is there any diagnostic info I can give you to help triage and potentially further improve performance?

rickeylev:
I think an analysis of a single target is a good place to start.

[...]

PR #3460, opened 2025-12-14
refactor: optimize venv creation for nvidia and pkgutil style namespace packages

rickeylev:
When pkgutil style namespace packages are used, multiple distributions provide
the same venv path (e.g. `foo/__init__.py`). The venv symlink logic then tries to
symlink the `foo/` directory as it looks like the highest linkable directory. When
conflict merging logic runs later, it then has to flatten a depset with all the files
in the conflicting distributions.

To fix, have whl_library() try to guess when a file is a pkgutil namespace package.
These are then pass onto py_library's venv building logic so it can treat the
directories as not directly linkable. A conflict still occurs, but it only
contains the single `__init__.py` file.

Along the way, special case the "nvidia" package name and always treat it as a namespace
package. This is because nvidia packages aren't strictly correct: each has a blank
`__init__.py` file (which marks it as a regular package, not namespace package). Special
casing like this is undesirable, but it greatly reduces the number of conflicts if
e.g. torch is installed, and I couldn't find any other metadata to indicate it's a
namespace package.

Along the way, add some hints to AGENTS.md so they understand repository rules better.

Fixes https://github.com/bazel-contrib/rules_python/issues/3401

### DIFF
--- a/.bazelignore
+++ b/.bazelignore
@@ -19,7 +19,6 @@ examples/bzlmod/other_module/bazel-other_module
 examples/bzlmod/other_module/bazel-out
 examples/bzlmod/other_module/bazel-testlogs
 examples/bzlmod/py_proto_library/foo_external
-examples/bzlmod_build_file_generation/bazel-bzlmod_build_file_generation
 examples/multi_python_versions/bazel-multi_python_versions
 examples/pip_parse/bazel-pip_parse
 examples/pip_parse_vendored/bazel-pip_parse_vendored
--- a/BZLMOD_SUPPORT.md
+++ b/BZLMOD_SUPPORT.md
@@ -11,7 +11,7 @@ In general `bzlmod` has more features than `WORKSPACE` and users are encouraged 
 
 ## Configuration
 
-The releases page will give you the latest version number, and a basic example.  The release page is located [here](/bazel-contrib/rules_python/releases).
+The releases page will give you the latest version number, and a basic example.  The release page is located [here](https://github.com/bazel-contrib/rules_python/releases).
 
 ## What is bzlmod?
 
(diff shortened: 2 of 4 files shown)


python/private/venv_runfiles.bzl
```diff
@@ -270,6 +325,9 @@ def get_venv_symlinks(ctx, files, package, version_str, site_packages_root):
     # List of (File, str venv_path) tuples
     files_left_to_link = []
 
+    # dict[str venv-relative dirname, bool is_namespace_package]
+    namespace_package_dirs = {}
```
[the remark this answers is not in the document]
rickeylev: Fixed

rickeylev:
Doh, forgot to remove some debugging files...

rickeylev:
PTAL. Debug code removed and tests fixed.

python/private/pypi/whl_library.bzl
```diff
@@ -561,6 +568,40 @@ if __name__ == "__main__":
     )
     return contents
 
+def _find_namespace_package_files(rctx, install_dir):
```
aignas: nit, there was a function to generate all of the locations of the namespace package files to be generated, this the inverse, detect them. Should this be in a separate file for easier testability?

AGENTS.md
```diff
@@ -27,6 +27,19 @@ end of the documentation text.
...
+Starlark does not support recursion. Use iterative algorithms instead.
+
+Starlark does not support `while` loops. Use `for` loop with an appropriately
+sized iterable instead.
+
+#### Repository rules
+
+The function argument `rctx` is a hint that the function is a repository rule.
+
+The `repository_ctx` API docs are at: https://bazel.build/rules/lib/builtins/repository_ctx
```
aignas, line 90: The same could be said about mctx and mrctx.

aignas:
Should we close #3401 as part of this?

python/private/pypi/whl_library.bzl
rickeylev: I moved it to pypi_repo_utils.bzl. It's repository-phase code, so I don't want to mix it with the loading-phase code in namespace_packages.bzl

I went to add unit tests, but the AI was just not having it; this logic is covered by the more functional tests, though.

AGENTS.md
rickeylev: Done.

rickeylev:
Yes, marked this as fixing https://github.com/bazel-contrib/rules_python/issues/3401. I'm not sure what optimizations are left action-count wise.

PR #3461, opened 2025-12-14
fix(pipstar): fix whl extraction and flip pipstar=true

### LINKED ISSUE #2949: feat: enable pipstar by default

aignas:
Attempt number 2. This should be smoother this time and should not cause
any breakage because we are not enabling any cross-building by default and
only the host wheels will be present.

Because we also started extracting using starlark APIs, some extra fixups
where needed because some wheels require extracting `.data` files into
correct paths. This also adds the `INSTALLER` file after extracting files to
signify that `pipstar` has installed the file.

Because we have stopped passing hermetic interpreter to the `whl_library` if
pipstar is enabled, we also needed to ensure that the code path is only enabled
if the extraction with pipstar is supported (i.e. bazel >= 8).

Fixes #2949


rickeylev:
CI failure:

```
(00:28:43) ERROR: /workdir/examples/pip_parse/BUILD.bazel:69:8: in env[WHEEL_DATA_CONTENTS] attribute of py_test rule //:pip_parse_test: label '@@rules_python++pip+pypi//s3cmd:data' in $(locations) expression expands to no files. Since this rule was created by the macro 'py_test', the error might have been caused by the macro implementation
(00:28:43) ERROR: /workdir/examples/pip_parse/BUILD.bazel:69:8: Analysis of target '//:pip_parse_test' (config: 5cac5d0) failed
```

rickeylev:
That failure is a bit strange. It would mean there are literally no files under site-packages?

aignas:
The build query is weird, it seems like a regression to me.

aignas:
Somewhat something that I thought might be the case, we can't just extract a wheel, we need to map certain paths in a particular way:
```
$ env USE_BAZEL_VERSION=8.5.0 bazel query "deps(@pypi//s3cmd:data)"
@@rules_python++pip+pypi_39_s3cmd//:data
@@rules_python++pip+pypi_39_s3cmd//:data/share/doc/packages/s3cmd/INSTALL.md
@@rules_python++pip+pypi_39_s3cmd//:data/share/doc/packages/s3cmd/LICENSE
@@rules_python++pip+pypi_39_s3cmd//:data/share/doc/packages/s3cmd/NEWS
@@rules_python++pip+pypi_39_s3cmd//:data/share/doc/packages/s3cmd/README.md
@@rules_python++pip+pypi_39_s3cmd//:data/share/man/man1/s3cmd.1
```

This is what the data contains in the non pipstar case.

aignas:
found it, had it forgotten.

aignas:
@rickeylev, do you wanna do the path remapping at the build time instead of the repo phase? I think that for compatibility reasons we should keep this for now and then we can think about it later on how we can actually extract/link the wheels better.

aignas:
OK, bugs should have been fixed PTAL.

rickeylev:
yes, lets keep it the same for now, for compatibility.

rickeylev (approved):

PR #3463, opened 2025-12-15
fix: allow pypi packages with empty data attribute

rickeylev:
An empty glob under site-packages (after excluding code and pyi files) is a bit strange,
but not actually incorrect. If such a situation occurs, it would cause a failure because
Bazel has since changed to disallow empty globs by default.

To fix, just set allow_empty=True on the data glob.

### DIFF
--- a/python/private/BUILD.bazel
+++ b/python/private/BUILD.bazel
@@ -724,6 +724,7 @@ bzl_library(
     deps = [
         ":common_bzl",
         ":py_info.bzl",
+        ":py_internal.bzl",
         "@bazel_skylib//lib:paths",
     ],
 )


aignas:
I think in reality `dist-info` will always match unless you manually exclude it. So the wheel won't be valid, most likely if you are hitting this.

rickeylev:
Huh, interesting. Gemini warned that setting the attribute would cause a test failure. I wonder if it ran after CI was done and processed that? Or because it grokked the whole code base? In either case, it was right. That was legit impressive.

rickeylev:
Thanks, yeah, good point. I don't see how a valid wheel could trigger this. But it seems relatively innocuous to allow.

PR #3473, opened 2025-12-20
refactor!: create full venv for bootstrap=system_python

rickeylev:
The system_python bootstrap is basically just an alternative stage1 bootstrap now.
Finish unifying it with bootstrap=script by having it support and create a full venv.

While this is the new default behavior, it's disabled for:
* Windows: some difficult to debug failures came up, so disable using it there
* Bazel 7: Bazel 7 doesn't have the `File.is_symlink` APIs, so it's
  problematic to enable it there due to zipapp support.

Switching from non-venv to venv layouts is a significant change and likely to
break something, so marking this as a breaking change.

Additional notes:
* rules_pkg 1.2+ is needed to package a venv-based binary, due to extensive use
  of symlinks.

Work towards https://github.com/bazel-contrib/rules_python/issues/2156

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,9 +69,11 @@ END_UNRELEASED_TEMPLATE
 Other changes:
 * (pypi) Update dependencies used for `compile_pip_requirements`, building
   sdists in the `whl_library` rule and fetching wheels using `pip`.
-* (pypi) We will set `allow_fail` to `False` if the {attr}`experimental_index_url_overrides` is set
-  to a non-empty value. This means that failures will be no-longer cached in this particular case.
-  ([#3260](https://github.com/bazel-contrib/rules_python/issues/3260) and 
+* (pypi) We will set `allow_fail` to `False` if the
+  {attr}`experimental_index_url_overrides` is set
+  to a non-empty value. This means that failures will be no-longer cached in
+  this particular case.
+  ([#3260](https://github.com/bazel-contrib/rules_python/issues/3260) and
   [#2632](https://github.com/bazel-contrib/rules_python/issues/2632))
 
 {#v0-0-0-fixed}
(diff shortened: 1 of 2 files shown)


rickeylev:
This is almost done, but tow issues remain:

1. rules_pkg 1.2 requires bazel 8+ for some reason (not clear why); rules_pkg 1.2 is needed for pkg_tar to work with declare_symlink.
2. Switching system_python to a full build-time venv using declare_symlink breaks packaging using py_wheel and/or pkg_tar (same issue as with bootstrap=script)

I'm not sure if both of these can be fixed. I guess if we drop bazel 7? We'll do that in a month or so anyways. Maybe don't test bazel 7 + packaging?

aignas:
I personally think that it could be best to fail if we detect bazel 7 and we cannot support it. +1 for excluding the tests via `target_compatible_with`. That said, why are we breaking packaging? I think we should exclude the `venv` from packaging altogether?

rickeylev:
The issue with bazel 7 isn't with our rules, but with rules_pkg, in two respects:
1. There is some sort of low-level bzlmod issue with it. Whatever rules_pkg 1.2 is doing, its MODULE doesn't work with bazel 7.
2. Bazel 7 doesn't have the File.is_symlink() apis to understand the venv's we create.

Our venv's work fine on bazel 7; it's just rules_pkg that has an issue.

A py_binary doesn't know if its being consumed by pkg_tar, so it can't automatically disable the symlink-based venv creation.

edit -- if we disable build-time venv creation for bazel 7, that could work. Its somewhat a regression in functionality, though. One only necessary if rules_pkg is used.

rickeylev:
Ah -- how about, for bazel 7, we don't create a full venv for system_python? That would match existing behavior. Bazel 8 + system_python would do full venv. Then we can drop Bazel 7 support In Jan/Feb timeline

rickeylev:
Ok, CI passed. PTAL

I disabled this change for Windows, too. Ran into some weird errors; I think the old zip-based stuff on windows doesn't like the venv (not surprising), but better to deal with that in a separate PR.

Also marking this as breaking because (1) pretty large change and (2) there's already the other breaking change, so might as well bundle it in

aignas:
Is the next release gonna be 2.0?

rickeylev:
Yes

PR #3474, opened 2025-12-22
fix(venv): Fix all .so files missing when py_binary lives at //:BUILD

shayanhoshyari:
When `.label.package` was "" (empty string), doing
`"{}/{}".format(ctx.label.package, bin_venv_path)` created an absolute path,
which caused files to be ignored later, as they no longer looked to be part
of the correct prefix. An empty package name occurs when the target is in
`//:BUILD` with venv on.

To fix, use skylib's `paths` instead, which handles it correctly and won't add the
`/` prefix.

Fixes https://github.com/bazel-contrib/rules_python/issues/3470


rickeylev:
Thanks for the investigation and fix!

PR #3478, opened 2025-12-23
feat: add --debugger flag

rickeylev:
The --debugger flag is useful for injecting a user-specified dependency without
having to modify the binary or test. Similarly, tests now implicitly inherit the
`PYTHONBREAKPOINT` environment variable.

The dependency is only added for the target config because build tools can't be
intercepted for debugging.

### DIFF
--- a/python/py_binary.bzl
+++ b/python/py_binary.bzl
@@ -28,13 +28,13 @@ def py_binary(**attrs):
     * `srcs_version`: cannot be `PY2` or `PY2ONLY`
     * `tags`: May have special marker values added, if not already present.
 
-    Args:
-      **attrs: Rule attributes forwarded onto the underlying {rule}`py_binary`.
-
     :::{versionchanged} VERSION_NEXT_FEATURE
     The `PYTHONBREAKPOINT` environment variable is inherited. Use in combination
     with {obj}`--debugger` to customize the debugger available and used.
     :::
+
+    Args:
+      **attrs: Rule attributes forwarded onto the underlying {rule}`py_binary`.
     """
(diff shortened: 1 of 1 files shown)


docs/api/rules_python/python/config_settings/index.md
```diff
@@ -45,6 +45,27 @@ This flag replaces the Bazel builtin `--build_python_zip` flag.
...
 
+::::{bzl:flag} debugger
+A target for providing a custom debugger dependency.
+
+This flag is roughly equivalent to putting a target in `deps`. It allows
+injecting a dependency into executables (`py_binary`, `py_test`) without having
+to modify their deps. The expectation is it points to a target that provides an
+alternative debugger (pudb, winpdb, debugpy, etc).
+
+* Must provide {obj}`PyInfo`.
```
aignas, line 56: I see that the `debugpy` can be used either as a CLI via `python -m debugpy` which could work well with our `main_module` or via `import debugpy` that would automatically work with this as well. To be truly useful, it would be nice to give at least one example how to setup the debugger. That can be done in a separate PR though.

aignas (approved):

docs/api/rules_python/python/config_settings/index.md
rickeylev: What does `python -m debugpy` do? Just start a debugpy server?

debugpy is particularly interesting because its a client/server design, so it can can actually Just Work with `bazel test` (where stdin is gone).

rickeylev:
re: example for how to use them: I added a basic howto doc.

aignas (approved):
Thanks!

PR #3482, opened 2025-12-27
exp: pypi- allow users to pass any versions

### LINKED ISSUE #3479: "attempting to create a duplicate library" when multiple python versions with same major+minor are used (different patch version)
# 🐞 bug report

### Affected Rule

Maybe `pip.parse`?

### Is this a regression?

It appears to be, yes. I bisected from `1.6.1` to `1.7.0` and found that commit 7b88c87aaab1e4711a8b61d2a47f445052ed6e9a (PR #3243) is the first bad commit.

### Description

We're getting a "duplicate library" error.

```
Error in fail: attempting to create a duplicate library pypi_313_absl_py_py3_none_any_9824a48b for absl_py
```

## 🔬 Minimal Reproduction

**TL;DR: running multiple python versions with the same **minor** version (eg 3.13.4 and 3.13.6) appears to break things.**

I'll see what I can do about making a minimal reproduction.

### Background

We have things set up so that each python target is actually a macro wrapping `py_*` for the current python version and `py_*` for the "pynext" python version. Example:

```starlark
load("@rules_python//python:defs.bzl", _py_binary = "py_binary", _py_test = "py_test")
load("@python_versions//3.13:defs.bzl", _py_binary_next = "py_binary", _py_test_next = "py_test")  # py version kept in sync with MODULE.bazel
def pyle_py_binary(**kwargs):
    _py_binary(**kwargs)
    _py_binary_next(**kwargs)
```

And thus our `MODULE.bazel` file has:

```starlark
PYTHON_VERSION = "3.13.4"
PYNEXT_VERSION = "3.13.6"  # must be different from PYTHON_VERSION. Before 1.7.0, a different **patch** version was sufficient

python.toolchain(
    ...,
    is_default = True,
    python_version = PYTHON_VERSION,
)
python.toolchain(
    ...,
    is_default = False,
    python_version = PYNEXT_VERSION,
)
use_repo(python, PYTHON_VERSION_NAME, PYNEXT_VERSION_NAME, "python_versions")

pip.parse(
    experimental_index_url = ...,
    experimental_index_url_overrides = ...,
    hub_name = "pypi",
    python_version = PYTHON_VERSION,
)
pip.parse(
    experimental_index_url = ...,
    experimental_index_url_overrides = ...,
    hub_name = "pypi",
    python_version = PYNEXT_VERSION,
)
```

You'll notice that we're using the same requirements.txt lock file and `hub_name` for both current and "pynext" versions.

We then use build/test tags to only build/test the current python version or the "pynext" python version:

```shell
# (default: --test_tag_filters=-pynext, only run against the current python version)
bazel test //...

# optionally test against both versions
bazel test --test_tag_filters= //...

# optionally test against only the "pynext" version
bazel test --test_tag_filters=pynext //...
```

## 🔥 Exception or Error

```
 bazel build --nobuild //...
INFO: Invocation ID: 8783fed4-2abe-4061-a080-4b3640de62a1
ERROR: /usr/local/google/home/dthor/.cache/bazel/_bazel_dthor/0f8c52850e7230283fc2f8033149fba2/external/rules_python+/python/private/pypi/hub_builder.bzl:185:13: Traceback (most recent call last):
        File "/usr/local/google/home/dthor/.cache/bazel/_bazel_dthor/0f8c52850e7230283fc2f8033149fba2/external/rules_python+/python/private/pypi/extension.bzl", line 390, column 25, in _pip_impl
                mods = parse_modules(module_ctx, enable_pipstar = rp_config.enable_pipstar)
        File "/usr/local/google/home/dthor/.cache/bazel/_bazel_dthor/0f8c52850e7230283fc2f8033149fba2/external/rules_python+/python/private/pypi/extension.bzl", line 280, column 30, in parse_modules
                builder.pip_parse(
        File "/usr/local/google/home/dthor/.cache/bazel/_bazel_dthor/0f8c52850e7230283fc2f8033149fba2/external/rules_python+/python/private/pypi/hub_builder.bzl", line 58, column 47, in lambda
                pip_parse = lambda *a, **k: _pip_parse(self, *a, **k),
        File "/usr/local/google/home/dthor/.cache/bazel/_bazel_dthor/0f8c52850e7230283fc2f8033149fba2/external/rules_python+/python/private/pypi/hub_builder.bzl", line 125, column 22, in _pip_parse
                _create_whl_repos(
        File "/usr/local/google/home/dthor/.cache/bazel/_bazel_dthor/0f8c52850e7230283fc2f8033149fba2/external/rules_python+/python/private/pypi/hub_builder.bzl", line 440, column 29, in _create_whl_repos
                _add_whl_library(
        File "/usr/local/google/home/dthor/.cache/bazel/_bazel_dthor/0f8c52850e7230283fc2f8033149fba2/external/rules_python+/python/private/pypi/hub_builder.bzl", line 185, column 13, in _add_whl_library
                fail("attempting to create a duplicate library {} for {}".format(
Error in fail: attempting to create a duplicate library pypi_313_absl_py_py3_none_any_9824a48b for absl_py
ERROR: error evaluating module extension @@rules_python+//python/extensions:pip.bzl%pip
INFO: Elapsed time: 26.869s
INFO: 0 processes.
Loading: 363 packages loaded
    currently loading: 
    Fetching module extension @@rules_python+//python/extensions:pip.bzl%pip; Fetch package lists from PyPI index 26s
```

## 🌍  Your Environment

**Operating System:**

```
gLinux (Debian Testing)
```

```
$ bazel version
Bazelisk version: v1.26.0
Build label: 8.3.1
Build time: Mon Jun 30 16:23:40 2025 (1751300620)
Build timestamp: 1751300620
Build timestamp as int: 1751300620
```

**Rules_python version:**

1.7.0

**Anything else relevant?**

Removing any support for "pynext" allows us to bump to 1.7.0. Additionally, bumping `PYNEXT_VERSION` to a different **minor** version than `PYTHON_VERSION` appears to work (eg current=3.13, next=3.14), but I can't fully confirm this yet because our requirements are not compatible with python 3.14.

This might be PEBKAC - if there's a more "correct" way of running multiple python versions please let me know. Our requirements for such are:

+ Do not build/test/run any "pynext" version by default
+ Only need to support versions N and N+1, where N might be a major, minor, or patch version
+ Must use the same requirements.txt lock file.
+ Must use the same pypi hub name
  + Or more accurately: the third-party deps of a target must not change between python versions.

aignas:
This is a quick draft POC to outline how I imagine any versions in `pip.parse` being handled.

TODO:
- [ ] More tests
- [ ] Discuss how to handle the detection of available python interpreters -
  users can specify anything, so we should normalize, but where do we stop?

Fixes #3479.

### DIFF
--- a/python/private/pypi/render_pkg_aliases.bzl
+++ b/python/private/pypi/render_pkg_aliases.bzl
@@ -144,12 +144,14 @@ def _major_minor(python_version):
     minor, _, _ = tail.partition(".")
     return "{}.{}".format(major, minor)
 
-def _major_minor_versions(python_versions):
+def _with_major_minor_versions(python_versions):
     if not python_versions:
         return []
 
     # Use a dict as a simple set
-    return sorted({_major_minor(v): None for v in python_versions})
+    ret = {_major_minor(v): None for v in python_versions}
+    ret.update({v: None for v in python_versions})
+    return sorted(ret)
 
 def render_multiplatform_pkg_aliases(*, aliases, platform_config_settings = {}, **kwargs):
     """Render the multi-platform pkg aliases.
@@ -178,7 +180,7 @@ def render_multiplatform_pkg_aliases(*, aliases, platform_config_settings = {}, 
         **kwargs
     )
     contents["_config/BUILD.bazel"] = _render_config_settings(
-        python_versions = _major_minor_versions(flag_versions.get("python_versions", [])) + flag_versions.get("python_versions", []),
+        python_versions = _with_major_minor_versions(flag_versions.get("python_versions", [])),
         platform_config_settings = platform_config_settings,
         visibility = ["//:__subpackages__"],
     )


python/private/config_settings.bzl
```diff
@@ -184,6 +184,13 @@ def construct_config_settings(
 
 def _python_version_flag_impl(ctx):
     value = ctx.build_setting_value
+    if value:
+        # FIXME @aignas 2025-12-27: is this the right thing to do here? Should we do this only for
+        # the `FeatureFlagInfo`?
+        #
+        # ensure that the values here are normalized
```
rickeylev, line 191: What sort of normalization is this applying that matters?

I think it's fair to say: the provided value is expected to be a valid python version string. If it's not, behavior is undefined.

rickeylev:
The main question I have is: what do different combinations of `pip.parse(python_version=...)` and `--python_version=...` mean? e.g. `pip={3.10.0, 3.10.5} + --python_version=3.10.3` ? Or similar if pip is "3.10" or vice versa. Basically, the matching semantics when it isn't an exact match.

I think i did some matching logic like that in the regular toolchain stuff...I think for the host toolchain? when there wasn't an exact match, it tired to find the "nearest" version (I don't remember if that meant "the highest below it" or "the closest above it").

python/private/config_settings.bzl
aignas: I wanted to make sure that when the user specifies `3.15.0alpha0` it is normalized to `3.15.0a0` and then we depend on normalized values when creating our own config settings.

aignas:
Regarding the matching logic, I was thinking the following:
* User does not specify `python_version` in `pip.parse` - currently unsupported, but we could use the default python version from the toolchains. In `pipstar` this would play well.
* User specifies `python_version = "X.Y"` in `pip.parse`, then we match the `major_minor`. The python interpreter that would be used in the `whl_library` would be the version specified in `MINOR_MAPPING`. In legacy, that could have issues due to METADATA parsing, but with `pipstar` that should be OK.
* User specifies `python_version = "X.Y.Z"` in `pip.parse`, then we match the full normalized version. This could use the `select` property of specificity - we could have both, the `major-minor` and the exact version in the `config_setting` and then the priorities would play well with each other.

aignas:
This will have to be redone with a better design.

PR #3483, opened 2025-12-27
fix(--debugger): Ensure that imports or venv_site_package files are propagated for debugger target

shayanhoshyari:
https://github.com/bazel-contrib/rules_python/commit/a94bd0fdde426bf30efed7c819422d74b404cc18 recently added support for injecting dependencies for easier use of debugger. It allows injecting deps via `--@rules_python//python/config_settings:debugger=<target>`. While the runfiles from `<target>` were inherited in the final binary, the `imports` or `venv_site_packages` were missing. Hence making the debugger target unusable for various corner cases (e.g. when it uses `imports = ...` or when it is coming from pip hub and `venv_site_packages` are on).

This PR fixes that, and extends the unit test to include this situation.

Fixes: https://github.com/bazel-contrib/rules_python/issues/3481


shayanhoshyari:
cc: @rickeylev

rickeylev (approved):
Thanks!

PR #3484, opened 2025-12-28
feat: basic build data with stamping

rickeylev:
This makes build data accessible to binaries. By default, some basic information
is always available: target name and if stamping was enabled.

When `--stamp` is enabled, the full `workspace_status_command` output is included in the
build information.

Build information is made available through a special `bazel_binary_info` module. This
module is created by the bootstrap and injected into sys.modules. It provides an API
for gettting the raw build information that was written. Exposing a module, instead
of a file, is done insulate from the Bazel-ism of runfiles.

When stamping is enabled, the way stamped data works is a bit round about to get
the desired semantics: that the build data reflects the most recent build of a
binary, not the last cached value of the build information (e.g., if I build
a binary, then modify a file in the binary, then build again, the build data
should reflect the timestamp of the second build). Normal Bazel caching, however,
makes that somewhat difficult (the build data file has to take the full transitive
set of files as inputs to properly detect changes, otherwise it'll get cached
on first build of the binary; plus actions can't take runfiles objects as inputs).

To work around this, we use some special APIs to get `ctx.version_file` information
into the output without entirely destroying caching.

* Because `ctx.version_file` doesn't trigger cache invalidation, a special helper,
  `py_internal.copy_without_caching()`, is used to make it _always_ invalidate
  caching. Now we can ensure we have a copy of the most recent invocation data.
* To prevent the above from _always_ rebuilding a binary, another special api,
  `py_internal.declare_constant_metadata_file()` is used to restore the behavior
  of a file never invalidating caching (even if it changes).

This dance is necessary because actions can't take runfiles as direct inputs. It
also avoids having to pass the entire program's transitive set of files as an input
just to do input change tracking.

Note the above only applies when stamping is enabled. Unstamped binaries don't
do any of this.

Along the way...

* The `stamp` attribute now transitions the `--stamp` flag. This was necessary so that
  the dependency on version_info was conditional on stamping.
* Remove a defunct helper from the stage2 bootstrap code.

### DIFF
--- a/python/private/stage2_bootstrap_template.py
+++ b/python/private/stage2_bootstrap_template.py
@@ -52,6 +52,7 @@ BUILD_DATA_FILE = "%build_data_file%"
 
 # ===== Template substitutions end =====
 
+
 class BazelBinaryInfoModule(types.ModuleType):
     BUILD_DATA_FILE = BUILD_DATA_FILE
 
@@ -335,8 +336,11 @@ def _maybe_collect_coverage(enable):
     # We need for coveragepy to use relative paths.  This can only be configured
     # using an rc file.
     rcfile_name = os.path.join(coverage_dir, ".coveragerc_{}".format(unique_id))
-    disable_warnings = ('disable_warnings = module-not-imported, no-data-collected'
-                        if COVERAGE_INSTRUMENTED else '')
+    disable_warnings = (
+        "disable_warnings = module-not-imported, no-data-collected"
+        if COVERAGE_INSTRUMENTED
+        else ""
+    )
     print_verbose_coverage("coveragerc file:", rcfile_name)
     with open(rcfile_name, "w") as rcfile:
         rcfile.write(


CHANGELOG.md
```diff
@@ -68,6 +70,9 @@ END_UNRELEASED_TEMPLATE
 ### Added
 * (binaries/tests) {obj}`--debugger`: allows specifying an extra dependency
   to add to binaries/tests for custom debuggers.
+* (binaries/tests) Build information is now included in binaries and tests.
```
aignas, line 74: We could also tell, that if users want, the could, for example set the `__version__` or other standard Python metadata from the `bazel_binary_info` module. What do you think?

aignas (approved):
LGTM, this is a great feature!

rickeylev:
> We could also tell, that if users want, the could, for example set the __version__ or other standard Python metadata from the bazel_binary_info module. What do you think?

Can you explain your idea more? `__version__` is a python module convention. A distribution's version is gotten through `importlib.metadata`. I don't see how either would be set by what's in this magic bazel_binary_info module? That module only exists when the process was started through the bazel executable. 

One idea I couldn't figure out how to make work was to expose build info through the `.dist-info` / "package metadata" mechanisms. Like, every `py_binary` would show up as if it was a distribution, so would have a `foo.dist-info/` directory.

As an aside, something I couldn't figure out was how to make `run(sys.executable, ...)` stuff work such that it could access build data, too. There'd have to be a module in available on sys.path (e.g. a module in the venv with `import bazel_build_info`), but then I couldn't figure out how a binary with another binary in data, where runfiles get merged, ends up looking.

rickeylev:
I ended up removing the config id because it was causing action conflicts. Apparently, top-level test targets have an implicit configuration transition to set test-only flags _but_ the exec root `bazel-bin/k8-fastbuild` etc) isn't changed (e.g. to `ST-whatever`). Stated another way, doing `bazel build //:foo_test //:genrule_that_uses_foo_test` will try to create the file `bazel-bin/foo_test.build_data.txt` with different content (the config id). I removed the config_mode to play it safe. Annoying, but NBD to omit it. Also a known bazel issue: https://github.com/bazelbuild/bazel/issues/21410

aignas:
> That module only exists when the process was started through the bazel executable.

Ah, yes, then it would not work.

The thing that `rules_go` has is `x_defs` which does template replacement in `.go` files that is really useful to include stamped metadata. What I would do as a user, to include stamped info is to create a `__init__.py` file with `__version__ = "%rules_python_build_stamp%"` and expect rules_python to do the template substitution.

That said, this can also be done by `genrule` or similar when one wants to brew their own.

> One idea I couldn't figure out how to make work was to expose build info through the .dist-info / "package metadata" mechanisms. Like, every py_binary would show up as if it was a distribution, so would have a foo.dist-info/ directory.

This is an interesting idea, maybe we can pass this via a provider `DistInfo` or something similar, that could later be used in `py_wheel` to write the `dist-info`? I think some of the necessary templating might be there.

EDIT: added a thing I remembered about `rules_go` and here is an issue where somebody is trying to use it, so might be a good way to add context: https://github.com/bazel-contrib/rules_go/issues/3847

PR #3492, opened 2026-01-02
fix: prevent a 404 error when serving Sphinx docs and Bazel is configured with a --symlink_prefix option

laurenshobert:
When Bazel is configured with the [`--symlink_prefix`](https://bazel.build/reference/command-line-reference#build-flag--symlink_prefix) option, the sphinxdocs `.serve` target fails to serve files from the correct directory. This happens because the directory layout under `execroot` and the workspace no longer match, causing the target to miss the generated HTML files.

This change updates the logic to use rpathlocation instead, ensuring the correct path is resolved regardless of symlink configuration.

Fixes https://github.com/bazel-contrib/rules_python/issues/3410

### DIFF
--- a/sphinxdocs/private/sphinx_server.py
+++ b/sphinxdocs/private/sphinx_server.py
@@ -11,6 +11,9 @@ from python.runfiles import Runfiles
 def main(argv):
     r = Runfiles.Create()
     serve_directory = r.Rlocation(argv[1])
+    if not serve_directory:
+        print(f"Error: could not find runfile for '{argv[1]}'", file=sys.stderr)
+        return 1
 
     class DirectoryHandler(server.SimpleHTTPRequestHandler):
         def __init__(self, *args, **kwargs):


aignas (approved):

aignas:
Thanks!

shayanhoshyari:
Not related to this PR, just some notes that others might find interesting...

Looking at this PR, I ended up going down a rabbit hole of why Bazel has so amny different `$(...)` macros for paths https://bazel.build/reference/be/make-variables#predefined_label_variables : `$(location)`, `$(rlocationpath)`, `$(rootpath)`, `$(execpath)`.

Digging turned out:
- First there was just `location`, but it turned out problematic as there was no way to distinguish when runfile path is needed vs. when path under workspace: https://github.com/bazelbuild/bazel/issues/2475#issuecomment-339318016, so `execpath`, `rootpath` were added. The names are not really helpful to tell one which is which.
- Then when bazel mod came out, the existence of external repos broke things + workspace, and they had to add `rlocationpath`: https://github.com/bazelbuild/bazel/issues/15029. Though onces workspaces go away I think we have  `rlocation path = os.path.join(_main, rootpath)`

Hindsight is 20/20 😅

schenker:
Thanks!

PR #3498, opened 2026-01-06
feat(gazelle): Add ancestor conftest.py files

### LINKED ISSUE #3497: Bug/regression: Ancestor conftest inclusion
👋 (I searched `conftest` in the issues, closed and open, and didn't find a relevant issue)

# 🐞 bug report

### Affected Rule

 The issue is caused by the rule: `py_test` (and `py_library`) (and Gazelle)

### Is this a regression?

 Kinda yes, kinda no? The previous version in which this bug was not present was: 1.5.4 (pre 1.6.X)

### Description

I'm having some heartburn updating past 1.6.0 due to https://github.com/bazel-contrib/rules_python/pull/2998 (I'll probably dump those in another issue), but one that sticks out is how Gazelle handles (or doesn't handle) ancestor `conftest.py`.

(See the repro for code structure)

- Pytest loads ancestor `conftest.py` files (not just the ones in the test file's directory). See [the docs](https://docs.pytest.org/en/stable/reference/fixtures.html#conftest-py-sharing-fixtures-across-multiple-files)
- Old Gazelle rules had all the test targets in the `tests/` dir, meaning they got the `:conftest` dependency (even if they were in descendant directories)
- New Gazelle rules have the test in the test's parent directory (this is a good thing 👍) but now those targets no longer get the automatic ancestor `conftest` dependency

## 🔬 Minimal Reproduction

Given this structure (all files empty, except for `projects/repro/BUILD.bazel` which has `# gazelle:python_root`)
```
<root>/
└── projects/
    └── repro/
        ├── BUILD.bazel
        └── tests/
            ├── BUILD.bazel
            ├── conftest.py
            ├── test_root_dir.py
            └── subdir/
                └── test_subdir.py
```

- On `1.5.4`: Gazelle will put everything in `projects/example/tests/BUILD.bazel`, where specifically `test_subdir.py` has a dependency on `:conftest`
- On 1.6.0: Gazelle puts `conftest.py` and `test_root_dir.py` targets in `projects/example/tests/BUILD.bazel` (with the test depending on `:conftest`) but `projects/example/tests/subdir:test_subdir` doesn't depend on `conftest.py` anymore)

## 🔥 Exception or Error

(If you put fixtures/hooks in `conftest.py`, especially ones with automagically execute, they won't work)

## 🌍  Your Environment

N/A

**Rules_python version:**

`1.6.3`

**Anything else relevant?**

thejcannon:
Re-introduces the gazelle behavior of adding conftest targets to tests so that
they inherit parent directory configs.

Fixes #3497

### DIFF
--- a/gazelle/python/generate.go
+++ b/gazelle/python/generate.go
@@ -507,7 +507,7 @@ func (py *Python) GenerateRules(args language.GenerateArgs) language.GenerateRes
 				pyTestTarget.addModuleDependency(
 					Module{
 						Name: importSpecFromSrc(pythonProjectRoot, conftestPkg, conftestFilename).Imp, 
-						Filepath: filepath.Join(conftestPkg, conftestFilename)
+						Filepath: filepath.Join(conftestPkg, conftestFilename),
 					},
 				)
 			}


gazelle/python/generate.go
```diff
@@ -66,6 +66,24 @@ func matchesAnyGlob(s string, globs []string) bool {
 	return false
 }
 
+// findConftestPaths returns package paths containing conftest.py, from currentPkg
+// up through ancestors, stopping at pythonProjectRoot (or repo root if empty).
```
dougthor42: N.B.: There are valid cases where there exists one or more conftest.py files that are above `pythonProjectRoot`:

```
MODULE.bazel
conftest.py
src/             # gazelle:python_root
  package1/
    subpackage/
      conftest.py
      foo.py
      foo_test.py
  package2/
    ...
```

I think we should ignore the `pythonProjectRoot` and only stop at the Bazel module root when searching.

rickeylev:
@dougthor42 : I see you removed it from the merge queue. Do you want me to undo my review and let you handle this PR? I just approved because I saw it requested for 1.8-rc2. I'm fine with omitting this from RC2 if you want time to review.

gazelle/python/generate.go
```diff
@@ -481,14 +499,15 @@ func (py *Python) GenerateRules(args language.GenerateArgs) language.GenerateRes
...
-				pyTestTarget.addModuleDependency(conftestModule)
+		shouldAddConftest := pyTestTarget.annotations.includePytestConftest == nil ||
+		*pyTestTarget.annotations.includePytestConftest
+
+		if shouldAddConftest {
+			for _, conftestPkg := range findConftestPaths(args.Config.RepoRoot, args.Rel, pythonProjectRoot) {
+				conftestModule := Module{
+					Name: importSpecFromSrc(pythonProjectRoot, conftestPkg, conftestFilename).Imp,
+				}
+				pyTestTarget.deps.Add(conftestModule)
```
dougthor42: Can you explain why you changed from `pyTestTarget.addModuleDependency()` to `pyTestTarget.deps.Add()`?

rickeylev:
I'm tagging RC2 now, so I'll just omit this PR for now.

thejcannon:
I'll get to the comments tonight hopefully when I crack open my laptop.

It's also no rush on my end, I've already patched my way to using it at $work.

rickeylev:
As an FYI, I'll probably be creating an RC3 in the next day or so.

gazelle/python/generate.go
thejcannon: `addModuleDependency` was calling `importSpecFromSrc`, which we're already doing on line 508.
This was making it so the `.Imp` looked like the neighbor `conftest.py`, so the parent never got added.

This is only the case if we're resolving sibling modules.

If you make the change to use `addModuleDependency(Module{Name: importSpecFromSrc(pythonProjectRoot, conftestPkg, conftestFilename).Imp})` then `//python:python_test_simple_test_with_conftest` will fail with

```
    --- FAIL: TestGazelleBinary/simple_test_with_conftest (0.01s)
        python_test.go:186: simple_test_with_conftest/bar/BUILD diff (-want,+got):
              (
                """
                ... // 23 identical lines
                        ":bar",
                        ":conftest",
            -           "//:conftest",
                    ],
                )
                """
              )
```

---

There's probably a way to fix the `if` inside `addModuleDependency` but I couldn't grok it...

thejcannon: Ahh with fresh eyes I realize I was missing `FilePath`

thejcannon:
Little late, but got the code updated

PR #3505, opened 2026-01-11
fix (venv_site_packages): Fix wrong runfiles.symlinks when py_binary is not in root module

shayanhoshyari:
When: 1) venv_site_packages is on 2) we have a py_binary in a non-root module (e.g. a tool used in rules), and it uses python deps that result in usage of runfiles.symlinks, the `ctx.runfile` based symlinks end up going in the wrong folder (`_main`), while the `ctx.actions.symlink(...)` files go in the right place. This results in an invalid `venv` and import errors.

The reason is that `actions.symlinks` always go to the `_main` module, as [Bazel docs explain](https://bazel.build/extending/rules#runfiles_symlinks). To send symlinks to other modules, one needs to use root_symlinks and prefix them with the right module name. 

Fixes: https://github.com/bazel-contrib/rules_python/issues/3503

### DIFF
--- a/tests/venv_site_packages_libs/py_binary_other_module_test.sh
+++ b/tests/venv_site_packages_libs/py_binary_other_module_test.sh
@@ -1,50 +1,6 @@
-#
-# Test that for a py_binary from a dependency module, we place links created via runfiles(...)
-# in the right place. This tests the fix made for issues/3503
-#
+# Test that for a py_binary from a dependency module, we place links created via
+# runfiles(...) in the right place. This tests the fix made for issues/3503
 
 set -eu
-
-# Helper to check links exist and where they point to
-check_link() {
-  link=$1
-
-  # 1) Must exist
-  if [ ! -e "$link" ]; then
-    return 1
-  fi
-
-  # 2) Must be a symlink
-  if [ ! -L "$link" ]; then
-    return 1
-  fi
-}
-
-ensure_link() {
-  if ! check_link $1; then
-    echo "ERROR: $link does not exist"
-    return 1
-  fi
-}
-
-cd ${RUNFILES_DIR}
-
-# First do a tree of whole runfiles for easier debugging in case of failure
-echo "[*] Runfile tree"
-find .
-
-# Sanity check that invalid files don't exist.
-echo "[*] Sanity check our ensure_link function"
-if check_link ./__I_DO_NOT_EXIST__; then
-  echo "Check link function is broken"
-  exit 1
-fi
-
-# Check the links exist in the correct place.
-echo "[*] Testing existence of symlinks in the right place"
-ensure_link ./other+/_venv_bin.venv/lib/python3.13/site-packages/nspkg/subnspkg/delta/__init__.py
-ensure_link ./other+/_venv_bin.venv/lib/python3.13/site-packages/nspkg/subnspkg/gamma/__init__.py
-
-# Finally, test that running the binary works, i.e imports are resolved.
 echo "[*] Testing running the binary"
-./other+/venv_bin
+"$VENV_BIN"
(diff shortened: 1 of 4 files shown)


python/private/venv_runfiles.bzl
```diff
@@ -65,7 +65,8 @@ def create_venv_app_files(ctx, deps, venv_dir_map):
             bin_venv_path = paths.join(base, venv_path)
             if is_file(link_to):
                 # use paths.join to handle ctx.label.package = ""
-                symlink_from = paths.join(ctx.label.package, bin_venv_path)
+                # ctx.label.repo_name should be prepended as we use runfiles.root_symlinks
+                symlink_from = paths.join(ctx.label.repo_name, ctx.label.package, bin_venv_path)
```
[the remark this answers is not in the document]
shayanhoshyari: Gemini is half right ctx.label.repo_name is empty for root.
But `.lstrip("@")` is not needed. `label.repo_name` is already canonic.

rickeylev:
Thanks for the fix. Overall it LGTM. I added the basics of a test to verify it, but have run out of time to fully incorporate the test. I was manually testing by running `bazel run @other//:venv_bin`.

I think all that's left is to change the venv_bin to a py_test, then put a test_suite() somewhere that refers to it (tests/venvs_site_packages_libs would be a good a place)

shayanhoshyari:
Thanks! I was also doing a similar thing, good it means I was on the right track.
Let me see if I can finish it tonight.

shayanhoshyari:
Thanks for the quick response.

~@rickeylev I pushed a test in https://github.com/bazel-contrib/rules_python/pull/3505/commits/1b6fe66c000bafbd140f29eaa8a85082d978f3ac~

~I used a `sh_test` so I can also actually test running the binary. Let me know if you have any concern, and I can switch to a analysis_test (test_suite).~

Nevermind, needs a analysis_test due to https://github.com/bazel-contrib/rules_python/pull/3505#discussion_r2679324002. I will try to do it tomorrow.

tests/venv_site_packages_libs/py_binary_other_module_test.sh
```diff
@@ -0,0 +1,50 @@
...
+echo "[*] Sanity check our ensure_link function"
+if check_link ./__I_DO_NOT_EXIST__; then
+  echo "Check link function is broken"
+  exit 1
+fi
+
+# Check the links exist in the correct place.
+echo "[*] Testing existence of symlinks in the right place"
+ensure_link ./other+/_venv_bin.venv/lib/python3.13/site-packages/nspkg/subnspkg/delta/__init__.py
+ensure_link ./other+/_venv_bin.venv/lib/python3.13/site-packages/nspkg/subnspkg/gamma/__init__.py
```
shayanhoshyari: Seems bazel 7 calls it `other/` rather than `other+/`
https://storage.googleapis.com/bazel-untrusted-buildkite-artifacts/019bac0c-4189-4c07-8d12-8904cd464fa0/tests/venv_site_packages_libs/py_binary_other_module_test/test_attempts/attempt_1.log

Ok really need an analysis test here for this part. Will do.

rickeylev: You can create an analysis test if you really want (I think there's several in the venv_site_packages_libs directory somewhere you can base upon; you can just pass the venv_bin binary as the input and inspect its runfiles).

That said, I would just check that the modules are importable instead of their exact paths. The exact paths are somewhat just an implementation detail. The key attribute we care about is if things are importable. If you still want to check the paths then just do a looser match, e.g. `other*/*/site-packages/nspkg/subnspkg/gamma/__init__.py`

shayanhoshyari: > I would just check that the modules are importable instead of their exact paths

Oh even easier. Ok I will do this. Since the execution of the binary checks this already, I think I just need to delete these lines that check the files. And add some better error message if imports fail in the binary.

rickeylev (approved):

rickeylev:
I've cleaned up and simplified the test.

shayanhoshyari:
> I've cleaned up and simplified the test.

music to my ears. thanks.

PR #3506, opened 2026-01-11
chore: remove defunct runtime distinction logic

rickeylev:
This removes some defunct logic that would distinguish how a runtime was found (using
toolchain resolution or the `--python_top` flag). With the python_top flag gone,
there's only one way to get the runtime.


aignas (approved):

aignas:
Thanks for the cleanup!

PR #3508, opened 2026-01-12
chore: remove py_proto_library from bzlmod example

rickeylev:
With the py_proto_library rules removed from rules_python, rules_python is no
longer the appropriate place to have examples for using the py_proto_library
rules. Having such example also causes some maintenance burden, as apparently
it incurs a nodejs dependency, which fails under newer bazel versions.

Fixes https://github.com/bazel-contrib/rules_python/issues/3362


aignas:
Should we remove `py_proto` dependency and mark the next release as 2.0?

PR #3514, opened 2026-01-14
feat: enable pyproject.toml as single source of truth for Python version

golithe:
- Enables using pyproject.toml as single source of truth for Python version via 
`python.defaults(pyproject_toml=...)`, `pip.default(pyproject_toml=...)`, and `pip.parse(pyproject_toml=...)`.
- The version is read from the requires-python field and used as the default toolchain version, so anything relying on the default (f.e. `py_binary`/`py_test`, `pip.parse`, `compile_pip_requirements`) picks it up with no explicit `python_version` in BUILD files:

```starlark
# MODULE.bazel
python.defaults(pyproject_toml = "//:pyproject.toml")
pip.default(pyproject_toml = "//:pyproject.toml")

# BUILD.bazel (version comes from default, no need to restate it)
compile_pip_requirements(
    name="requirements",
    requirements_txt = "requirements.txt"
 )
```
- Relates to #1708 (lifts the mandatory `python_version` restriction on `pip.parse`)

Changes:
- Added `pyproject_toml` attribute to `python.defaults()` and `pip.default()` and `pip.parse()`, reading the version from `requires-python` 
- Parses `pyproject.toml` with the pure-Starlark `@toml.bzl (toml.decode)` (no host Python interpreter required)
- Auto-registers a toolchain for the default version if no explicit `python.toolchain()` call exists
- Made `python_version` optional in `pip.parse()`: the version can come from `pip.parse(pyproject_toml=...)` directly (which also works for non-root modules) or fall back to `pip.default(pyproject_toml=...)`
- Precedence order: `ENV > pyproject_toml > python_version_file > python_version`

Requirements:
- `requires-python = "==X.Y.Z"` format in pyproject.toml

Thanks to @aignas  and @rickeylev  for the review and design guidance


aignas:
Hey, thank you for the contribution.

In general I am +1 on having the python version read from `pyproject.toml` but I have concerns on depending on Python for doing that. It is trivial to do it in Python but we have a layering violation, because we need to use some python interpreter before the interpreters are setup and that would mean that we would have to use the Python interpreter from the system.

> * Creates `python_version_from_pyproject` repository for BUILD file usage:
> 
> ```python
> load("@python_version_from_pyproject//:version.bzl", "PYTHON_VERSION")
> compile_pip_requirements(python_version = PYTHON_VERSION, ...)
> ```

What is the benefit of this? If we are setting the default version through the `pyproject.toml` interpreter, the `python_version` should not need to be set anymore in this way.

> * Makes `config_settings` optional in `pip.default()`. Previously, users had
>   to pass an empty list (`config_settings = []`) even when not using platform-specific
>   configurations. Now it can be omitted entirely when not needed.

+1 on this change. If you'd like to merge it earlier than we decide on the various `pyproject.toml` processing specific details, please submit a separate PR.

> * Added `pyproject_toml` attribute to `python.defaults()` and `pip.default()`

LGTM, but I am wondering if this is the simplest possible interface. We could instead reuse, continue using the machinery that is used by `.python-version` and do detection based on extension. If it is called a `pyproject.toml`, then we know for sure what it can contain.

> * Reads Python version from `requires-python` field (must be `==X.Y.Z` format)

Do we need to restrict the user to `==X.Y.Z`? Should we also support `X.Y`? What about `>=X.Y`? We do have python version parsing reimplemented in Starlark, so we process this information statically.

> * Made `python_version` optional in `pip.parse()` when using `pip.default(pyproject_toml=...)`

Is the idea to set the default value for `//python/config_settings:python_version` from `pyproject.toml`? If so, should we do this change in 2 steps - make the `pip.parse` `python_version` optional and fallback to the default version?

rickeylev:
I, too, am supportive of this.

> layering violation, using python to parse it

I'm not worried about that. We should just pick a python runtime to use for repo-phase logic that rules_python uses. It simplifies us being able to assume what python is being used and its capabilities. There's also the `config` bzlmod extension we could use to give some ability to customize what python we use during the repo phase.

> what syntax of requires-python should we support, how do they work

I'd say lets just go with what the author is willing to contribute. If someone later wants `>`, then they can send a PR.

golithe:
hi @aignas and @rickeylev, thank you for your comments! i am currently taking a vacation, will address your comments nexy week! cheerio

python/private/pypi/extension.bzl
```diff
@@ -86,12 +87,23 @@ def build_config(
...
     defaults = {
         "platforms": {},
+        "python_version": None,
     }
     for mod in module_ctx.modules:
         if not (mod.is_root or mod.name == "rules_python"):
             continue
 
         for tag in mod.tags.default:
+            pyproject_toml = getattr(tag, "pyproject_toml", None)
```
rickeylev: The pyproject tag will always be present, so this can use normal `.` for accessing.

python/private/pypi/extension.bzl
```diff
@@ -407,7 +425,7 @@ Either this or {attr}`env` `platform_machine` key should be specified.
 """,
     ),
     "config_settings": attr.label_list(
-        mandatory = True,
+        mandatory = False,
```
rickeylev: I think a recent PR fixed this?

python/private/pypi/extension.bzl
```diff
@@ -469,6 +487,22 @@ If you are defining custom platforms in your project and don't want things to cl
 [isolation] feature.
 
 [isolation]: https://bazel.build/rules/lib/globals/module#use_extension.isolate
+""",
+    ),
+    "pyproject_toml": attr.label(
+        mandatory = False,
+        allow_single_file = True,
```
rickeylev: allow_single_file can be omitted. It has no meaning for repo/bzlmod rules

python/private/pypi/extension.bzl
```diff
@@ -469,6 +487,22 @@ If you are defining custom platforms in your project and don't want things to cl
...
+        doc = """\
+Label pointing to pyproject.toml file to read the default Python version from.
+When specified, reads the `requires-python` field from pyproject.toml and uses
+it as the default python_version for all `pip.parse()` calls that don't
+explicitly specify one.
+
+The version must be specified as `==X.Y.Z` (exact version with full semver).
+This is designed to work with dependency management tools like Renovate.
+
+:::{versionadded} 1.8.0
```
rickeylev: Change to VERSION_NEXT_FEATURE instead of 1.8.0

python/private/pypi/extension.bzl
```diff
@@ -649,7 +683,7 @@ find in case extra indexes are specified.
             default = True,
         ),
         "python_version": attr.string(
-            mandatory = True,
+            mandatory = False,
```
rickeylev, line 923: update python_version attr doc to indicate the other option

add e.g.

```
:::{seealso}
The {obj}`pyproject_toml` attribute for getting the version from a project file.
:::
```

python/private/BUILD.bazel
```diff
@@ -29,7 +29,10 @@ package(
 
 licenses(["notice"])
 
-exports_files(["runtime_env_toolchain_interpreter.sh"])
+exports_files([
+    "runtime_env_toolchain_interpreter.sh",
+    "pyproject_version_extractor.py",
+])
```
rickeylev, line 37: pyproject_version_extractor doesn't need to be exported. repo rules don't respect visibility settings. They're just label-looking syntax for file-system paths.

python/private/pyproject_version_extractor.py
rickeylev, line 1: Ah, this is a good opportunity to add a toml2json tool, there's another pending PR that needs it, too.

Copy the toml2json tool from https://github.com/bazel-contrib/rules_python/pull/3557/changes and use it in this PR for parsing the pyproject.toml file.

* Copy tools/toml2json; move it to tools/private/toml2json
* Copy its tests.
* Use convert_uv_lock_to_json in python/private/pypi/uv_lock.bzl as inspiration for creating a pyproject_to_json bzl function. Notably, invoke the helper toml2json script using the utility functions it uses. These ensure the script is executed correctly/safely.

python/private/python.bzl
```diff
@@ -869,12 +903,22 @@ def _compute_default_python_version(mctx):
                 defaults_attr.python_version_file,
                 onerror = _fail_multiple_defaults_python_version_file,
             )
-        if default_python_version_file:
+
+        # Priority order: pyproject_toml > python_version_file > python_version_env > python_version
```
rickeylev: This logic needs to be partially reverted, as its changing behavior.

The ENV attribute has highest precedence (this allows overriding on the command line), followed by version file. I'm not sure if `.python-version` or pyproject.toml should have higher precedence; I'm think pyproject.toml should be higher, though.

python/private/python.bzl
```diff
@@ -915,11 +959,29 @@ def _create_defaults_attr_struct(*, tag):
...
         python_version_file = getattr(tag, "python_version_file", None),
+        pyproject_toml = getattr(tag, "pyproject_toml", None),
     )
 
-def _create_toolchain_attr_structs(*, mod, config, seen_versions):
+def _create_toolchain_attr_structs(*, mod, config, seen_versions, module_ctx, logger):
     arg_structs = []
 
+    # Check if pyproject_toml was specified in defaults
+    # If so, register a toolchain for it
```
rickeylev: The diff is hiding a lot of the content -- is this new logic specific to pyproject? It would make sense that this occurs for whatever default is used, not just if pyproject is.

python/private/python.bzl
```diff
@@ -959,6 +1021,18 @@ def _create_toolchain_attrs_struct(
...
     attrs = {
+        "pyproject_toml": attr.label(
+            mandatory = False,
+            allow_single_file = True,
+            doc = """\
+Label pointing to pyproject.toml file to read the default Python version from.
+When specified, reads the `requires-python` field from pyproject.toml.
+The version must be specified as `==X.Y.Z` (exact version with full semver).
+
+:::{versionadded} 1.8.0
```
rickeylev: Specify VERSION_NEXT_FEATURE

tests/pyproject/compile_requirements_test/MODULE.bazel
rickeylev, line 1: These tests are OK, but they're integration tests. We only add integration tests when absolutely necessary because we have a limited number of jobs we can run (and we're near that limit already).

I think there are some more unittest-style things with pieces mocked out.

Otherwise, we'll just have to leave the code paths as untested 🤷

rickeylev (commented):
Core logic is OK, but some of the impl details and tests need to change

python/private/pypi/extension.bzl
golithe: Kept `getattr` because the existing unit tests in `tests/pypi/extension/` create mock tag structs that don't include `pyproject_toml`. Using `tag.pyproject_toml` breaks `test_simple`. Happy to update the test mocks instead if preferred.

python/private/pypi/extension.bzl
golithe: @rickeylev do you happen  to know which one? i havent found it.. could also split this out as a separate PR if you'd prefer it merged independently.

python/private/python.bzl
golithe: PTAL

python/private/python.bzl
golithe: I removed the pyproject-specific auto-registration. `_create_toolchain_attr_structs` now receives the already-computed `default_python_version` and auto-registers a toolchain for it if no explicit `python.toolchain()` exists for that version. This is generic and works for any default source (pyproject, version file, env, static string). WDYT?

golithe:
@aignas: 
> Do we need to restrict the user to ==X.Y.Z? Should we also support X.Y? What about >=X.Y? We do have python version parsing reimplemented in Starlark, so we process this information statically.

i decided to start with ==X.Y.Z only as the simplest useful subset. the usecase i had in mind was mainly that tools like renovate can easily update it with exactly  pinned versions.  IMO Support for >=, X.Y, could be added in follow-up PRs. WDYT?

> Is the idea to set the default value for //python/config_settings:python_version from pyproject.toml? If so, should we do this change in 2 steps - make the pip.parse python_version optional and fallback to the default version?

`python_version` is now optional in `pip.parse()` and falls back to the version from `pip.default(pyproject_toml=...)`. If neither is set, it fails with a clear error message.

python/private/pypi/extension.bzl
aignas: +1 to update the mocks. In there setting it to `None` is good enough. There is a function where that would be easy to do.

python/private/pypi/extension.bzl
aignas: Yeah, in `main` it is `True`.

python/private/pypi/extension.bzl
```diff
@@ -707,6 +744,14 @@ The list of labels to use as SRCS for the marker evaluation code. This ensures t
 code will be re-evaluated when any of files in the default changes.
 """,
         ),
+        "_toml2json": attr.label(
+            default = "//tools/private/toml2json:toml2json.py",
+        ),
+        # TODO: Set a proper default once interpreter bootstrapping is resolved.
+        # See https://github.com/bazel-contrib/rules_python/pull/3557
```
aignas: What does this mean? Should we have some better default here?

golithe:
any update on this @rickeylev ?

golithe:
@aignas any update here?

python/private/pyproject_utils.bzl
```diff
@@ -0,0 +1,66 @@
...
+
+    pyproject_path = module_ctx.path(pyproject_label)
+    module_ctx.read(pyproject_path, watch = "yes")
+
+    toml2json = module_ctx.path(_TOML2JSON)
+    result = module_ctx.execute([
+        "python3",
+        str(toml2json),
+        str(pyproject_path),
+    ])
```
aignas: We no longer need this, we can use a pure starlark library to parse toml into JSON.

python/private/pyproject_utils.bzl
```diff
@@ -0,0 +1,66 @@
...
+    parts = bare_version.split(".")
+    if len(parts) != 3:
+        fail("requires-python must be in X.Y.Z format, got: {}".format(bare_version))
+    for part in parts:
+        if not part.isdigit():
+            fail("requires-python must be in X.Y.Z format, got: {}".format(bare_version))
+
+    return bare_version
+
+def read_pyproject_version(module_ctx, pyproject_label, logger = None):
```
aignas: Maybe it would be good to have `read_pyproject` function instead of this. Then it could return a struct with important values:
```starlark
struct(
    version = "3.13.9"
)
```

python/private/pyproject_utils.bzl
```diff
@@ -0,0 +1,66 @@
...
+    Returns:
+        The bare version string (e.g. "3.13.9").
+    """
+    if not requires_python.startswith("=="):
+        fail("requires-python must use '==' for exact version, got: {}".format(requires_python))
+
+    bare_version = requires_python[2:].strip()
+
+    # Validate X.Y.Z format
+    parts = bare_version.split(".")
```
aignas: please use the `python/private/version.bzl` to parse and normalize the version here. It will ensure that the code fails in the right way.

python/private/pyproject_utils.bzl
```diff
@@ -0,0 +1,66 @@
...
+    """Parse and validate the requires-python field.
+
+    Args:
+        requires_python: The raw requires-python string from pyproject.toml.
+
+    Returns:
+        The bare version string (e.g. "3.13.9").
+    """
+    if not requires_python.startswith("=="):
+        fail("requires-python must use '==' for exact version, got: {}".format(requires_python))
```
aignas: We could in theory just take the string instead of the version. Then we can use something like:
`evaluate("{python_version}{requires_python}".format(...))`

That will save us from a refactor in the future. In my experience it is best to read files and return `requires_python` as is and then delegate the logic of this to some other file.

CHANGELOG.md
```diff
@@ -145,6 +145,8 @@ and [#1975](https://github.com/bazel-contrib/rules_python/issues/1975).
 * (toolchains) Support dynamically fetching and registering Python runtimes
   from a python-build-standalone manifest file using
   `python.override(add_runtime_manifest_urls = ..., runtime_manifest_sha = ...)`.
+* (pip,python) Added `pyproject_toml` attribute to `pip.default()` and `python.defaults()`
+  to read Python version from pyproject.toml `requires-python` field (must be `==X.Y.Z` format).
```
aignas: todo: a reminder to switch to `news/3514.added.md` file for this snippet.

python/private/pyproject_repo.bzl
```diff
@@ -0,0 +1,80 @@
...
+        ),
+    },
+    doc = """Repository rule that reads Python version from pyproject.toml.
+
+This rule creates a repository with a `version.bzl` file that exports
+`PYTHON_VERSION` constant.
+
+Example:
+```python
+    load("@python_version_from_pyproject//:version.bzl", "PYTHON_VERSION")
```
aignas: not sure we need this, we could just set the default version in the extension.
Then the following should be using the default version:
```
    compile_pip_requirements(
        name = "requirements",
        requirements_txt = "requirements.txt",
    )
```

Or am I missing something?

aignas (commented):
Sorry it took a while to get to this.

Leaving some comments as `rules_python` has moved ahead a little since you created this PR.

My thinking is that the following structure would make sense:
* Create a helper that reads `pyproject.toml` and parses the contents.
* Pass `pyproject.toml` to the `python.bzl` to parse and detect the default here. This will take care of everything during runtime. Since the `DEFAULT_PYTHON_VERSION` can be retrieved, no need for the extra repository. Also this plays well with only the root module being able to set the python version.
* Pass `pyproject.toml` and get the version for the `hub_builder` and friends. Again, no need to create a separate repo. In the future we can use the `requires_python` to support dynamic selection/registration. This is especially useful since we have introduced `uv.lock` support.

Just to reiterate, I think this feature is really good and a welcome addition to `rules_python`. Thank you for your work!

golithe:
@aignas 

> Sorry it took a while to get to this. 

dont worry about the timing, thank you for your thorough review

>  I think this feature is really good and a welcome addition to rules_python

thank you!

python/private/pyproject_utils.bzl
golithe: just tbc, do you want me to:

1. just structure read_pyproject to return the raw requires_python string and keep the current effective behavior (single exact version for now) e.g. leaving full specifier support as a future change?
1. actually wire up evaluate(...) for arbitrary specifiers as part of this PR?

my read is the the first but happy to go further if you'd prefer

golithe:
@aignas, i restructured per your suggestions. 

- i dropped the repo rule (version is now the default so `compile_pip_requirements` picks it up automatically)
- swapped `toml2json` for pure-Starlark `@toml.bzl`
- `read_pyproject()` now returns a struct parsed via `version.bzl`
- changelog moved to a news fragment

on `requires_python`: i went with the raw string plus `version.bzl` (current exact-version behavior), leaving `evaluate()`-based specifier support as the future step you mentioned.

(checks fail with the same macOS jobs as on `main`)

python/private/pypi/extension.bzl
```diff
@@ -359,6 +367,10 @@ You cannot use both the additive_build_content and additive_build_content_file a
 
     for mod in module_ctx.modules:
         for pip_attr in mod.tags.parse:
+            python_version = pip_attr.python_version or config.python_version
+            if not python_version:
+                _fail("pip.parse() requires either python_version attribute or pip.default(pyproject_toml=...) to be set")
```
aignas: Since there is another PR #3755, which can take the pyproject.toml for other things, I thin we could support specifying `pyproject.toml` as part of `pip.parse` as well. That way the users can specify that instead of `python_version`. I think that that way would allow us to let non-root modules do that too, so that could be a nice long term solution.

Then the implementation of #3755 can build on the generic structure of the code implemented here.

python/private/pypi/extension.bzl
```diff
@@ -699,6 +712,21 @@ If you are defining custom platforms in your project and don't want things to cl
...
+    "pyproject_toml": attr.label(
+        mandatory = False,
+        doc = """\
+Label pointing to pyproject.toml file to read the default Python version from.
+When specified, reads the `requires-python` field from pyproject.toml and uses
+it as the default python_version for all `pip.parse()` calls that don't
+explicitly specify one.
+
+The version must be specified as `==X.Y.Z` (exact version with full semver).
+This is designed to work with dependency management tools like Renovate.
```
aignas, line 732:
```suggestion
:::{note}
The version must be specified as `==X.Y.Z` (exact version with full semver).
This is designed to work with dependency management tools like Renovate.
:::
```

python/private/pypi/extension.bzl
```diff
@@ -699,6 +712,21 @@ If you are defining custom platforms in your project and don't want things to cl
 [isolation] feature.
 
 [isolation]: https://bazel.build/rules/lib/globals/module#use_extension.isolate
+""",
+    ),
+    "pyproject_toml": attr.label(
```
aignas, line 722: +1 for also allowing users specify `pyproject_toml` via the `pip.parse`.

python/private/pypi/extension.bzl
```diff
@@ -870,14 +898,18 @@ find in case extra indexes are specified.
...
 The Python version the dependencies are targetting, in Major.Minor format
 (e.g., "3.11") or patch level granularity (e.g. "3.11.1").
 
 If an interpreter isn't explicitly provided (using `python_interpreter` or
 `python_interpreter_target`), then the version specified here must have
 a corresponding `python.toolchain()` configured.
+
+:::{seealso}
+The {obj}`pyproject_toml` attribute for getting the version from a project file.
+:::
```
aignas, line 934:
```suggestion
:::
:::{versionchanged} VERSION_NEXT_FEEATURE
No longer mandatory if {attr}`pyproject_toml` or {obj}`pip.default.pyproject_toml` is specified.
:::
```

#1708 is also an important issue to mention in this PR if we lift this restriction.

python/private/pyproject_utils.bzl
```diff
@@ -0,0 +1,48 @@
...
+    validated and normalized via {obj}`//python/private:version.bzl` so that
+    malformed input fails in a consistent way. Broader specifier support
+    (e.g. `>=`, `X.Y`) can be layered on here in the future.
+
+    Args:
+        requires_python: {type}`str` the raw `requires-python` value.
+
+    Returns:
+        {type}`str` the normalized version string (e.g. `"3.13.9"`).
+    """
```
aignas, line 40:
```suggestion
    """
    if not requires_python:
        fail("`requires-python` must be specified")

```

python/private/python.bzl
```diff
@@ -463,7 +465,14 @@ def _python_impl(module_ctx):
         )
 
     if bazel_features.external_deps.extension_metadata_has_reproducible:
-        return module_ctx.extension_metadata(reproducible = True)
+        # Build the list of direct dependencies
+        root_direct_deps = ["pythons_hub", "python_versions"]
```
aignas: I can't remember this anymore, but doesn't this make it so that bazel mod tidy will include these repos in `use_repo` statements if not already present? If yes, then we should not change that behaviour, if not, then this change is good.

aignas (commented):
One last comment here.

python/private/python.bzl
golithe: good catch, reverted.

python/private/pypi/extension.bzl
golithe: added

python/private/pypi/extension.bzl
golithe: added #1708 in PR desc

python/private/pypi/extension.bzl
golithe: done. 
- added a `pyproject_toml` attribute to `pip.parse` that resolves the version from the file when `python_version` is not set
- since parse tags are not root-only, non-root modules can use it too
- kept `read_pyproject` and `version_from_requires_python` generic so #3755 can build on the same structure

wdyt?

aignas (approved):

golithe:
\o/
/cc @gzm0 @JeroenSchmidt

JeroenSchmidt:
@golithe thank you for doing this. Do you have any suggestions on how best to extend your work to support python ranges? This is something I might need to implement in the future to reduce biforcation between bazel and pyproject.toml in my repos.

I'd also love to hear your [thoughts on how pyproject.toml can be leveraged more ](https://github.com/bazel-contrib/rules_python/discussions/3914 )

aignas:
@JeroenSchmidt, the best way would be to use the `requires_python` expression and iterate across all of the available Python versions:
```starlark
for mod in mctx.modules:
    for attr in mod.parse:
        # It is important to only iterated registered versions.
        for python_version in MINOR_MAPPING_REGISTERED:
            if evaluate('"{}"{}'.format(python_version, pyproject_toml.requires_python)):
                hub_builder.add(python_version, attr)
```

I hope that helps you get on the right track.

golithe:
@JeroenSchmidt 
> thank you for doing this

youre welcome

> I'd also love to hear your https://github.com/bazel-contrib/rules_python/discussions/3914

i will look into it


@aignas thanks for your suggestion

i think that leaves the question open about the selection of the version in the toolchain mechanism.

either `python.defaults(pyproject_toml )` would stay exact-only or, if it accepts a range, there needs to be an explicit rule (f.e. "highest-compatible-registered" , correct?.

what are your thoughts on this @JeroenSchmidt , @aignas?

JeroenSchmidt:
Hi @golithe besides "default" i was a thinking that a range of toolchains would be setup for you based on what you have in the pyproject.toml. 

This would be quite useful if you have a library that you want to be validated against multiple versions of Python (as defined in your pyproject.toml). 

`requirments` is already able to switch between different lock files depending on the lock file as long as they are on the same hub. 

This assumes you either use a uv lock file (which is universal) or generate a lock file per Python version and assign it to one hub. 

Essentially; it would allow for a turn key solution to do this: 
```bash
bazel test --config=py310 //...
bazel test --config=py311 //...
```

PR #3516, opened 2026-01-14
Add missing #3046 reference to 1.8.0 changelog

dougthor42:
I've had to reference this a couple times now and keep forgetting that it's not in the changelog haha. I figured I'd remedy that.

### DIFF
--- a/python/private/pypi/pep508_deps.bzl
+++ b/python/private/pypi/pep508_deps.bzl
@@ -156,6 +156,7 @@ def _add_reqs(deps, deps_select, dep, reqs, *, extras):
             return
 
     markers = {}
+    found_unconditional = False
     for req in reqs:
         for x in extras:
             m = evaluate(req.marker, env = {"extra": x}, strict = False)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -216,6 +216,9 @@ END_UNRELEASED_TEMPLATE
   from the [20251014] release.
 * (deps) (bzlmod) Upgraded to `bazel-skylib` version
   [1.8.2](https://github.com/bazelbuild/bazel-skylib/releases/tag/1.8.2)
+* (gazelle) Gazelle will now delete targets that have an invalid `srcs`
+  attribute (typically because the only `srcs` file was deleted).
+  [#3046](https://github.com/bazel-contrib/rules_python/pull/3046)
```
shayanhoshyari: Was this 1.7.0 or 1.8.0? Seems it is already under 1.8.0

> (gazelle) Remove {obj}py_binary targets with invalid srcs. This includes files that are not generated or regular files.

https://github.com/dougthor42/rules_python/blob/main/CHANGELOG.md?plain=1#L138-L140

Screenshot 2026-01-16 at 1 18 32 PM

same PR, https://github.com/bazel-contrib/rules_python/pull/3046, added this line to changelog.

I actually remember it was not in 1.7.0, and when I updated to 1.8.0-rc1 it was there. Because it did break some of my targets and I had to add some `#gazelle:...` directives to fix things.

dougthor42: Ah, good catch! Updated.

dougthor42: Well no wonder I was never able to find it... I was always looking in the 1.7 change list haha.

rickeylev (approved):

PR #3522, opened 2026-01-19
chore (py_internal): Remove roundtrip of putting py_internal in the config repo

shayanhoshyari:
I was reading on `py_internal` and found this opportunity for simplification.

When there was support for Bazel `< 7` `py_internal` was in `@rules_python_internal`, because it had to be set to None depending on Bazel version. 

https://github.com/bazel-contrib/rules_python/pull/3282 removed Bazel `< 7` support shims, and made `py_internal` not depend on Bazel version.  So now there no need to put it in `@rules_python_internal`.

### DIFF
--- a/python/private/py_internal.bzl
+++ b/python/private/py_internal.bzl
@@ -18,8 +18,10 @@ Re-exports the restricted-use py_internal helper under its original name.
 These may change at any time and are closely coupled to the rule implementation.
 """
 
-# The py_internal object needs to be accessed only from "tools/build_defs/python", so
-# we need to first put it there with a different name.
+# The native `py_internal` object is only visible to Starlark files under
+# `tools/build_defs/python`. To access it from `//python/private`, we use an
+# indirection through `//tools/build_defs/python/private/py_internal_renamed.bzl`,
+# which re-exports it under a different name.
 load("//tools/build_defs/python/private:py_internal_renamed.bzl", "py_internal_renamed")  # buildifier: disable=bzl-visibility
 
 py_internal = py_internal_renamed


shayanhoshyari:
cc: @aignas since you did the original simplification (https://github.com/bazel-contrib/rules_python/pull/3282)

aignas:
Is there any benefit in runtime performance, etc, or is it just a cleanup of the code? I am not sure I understand the implications by just reading the comments.

shayanhoshyari:
> Is there any benefit in runtime performance, etc, or is it just a cleanup of the code? I am not sure I understand the implications by just reading the comments.

Just cleanup the code. 

Performance and code behavior should remain the same.

I was trying to wrap my head around why [py_internal](https://github.com/bazelbuild/bazel/blob/master/src/main/starlark/builtins_bzl/common/python/py_internal.bzl) is going through so many loops. With this change next person reading the code hopefully does not have to.

---

**Before**

```mermaid
flowchart TB
A["//tools/build_defs/python/private:py_internal_renamed.bzl<br>py_internal_renamed"]
B["@rules_python_internal//:py_internal.bzl<br>py_internal_impl"]
C["@rules_python//private:py_internal.bzl<br>py_internal"]

A --> B --> C
```


**Now**
```mermaid
flowchart TB
A2["//tools/build_defs/python/private:py_internal_renamed.bzl<br>py_internal_renamed"]
C2["@rules_python//private:py_internal.bzl<br>py_internal"]

A2 --> C2
```


`rules_python_internal//:py_internal.bzl` usage was only needed for Bazel 6 or earlier, since for those `py_internal` was not defined, and it was set to None for them depending on Bazel version, like [when <= supported](https://github.com/bazel-contrib/rules_python/blob/f2668295be2f73623ea0ed62f2942ecfd448c8d2/python/private/internal_config_repo.bzl#L118-L123). But [now](https://github.com/bazel-contrib/rules_python/blob/1.8.0/python/private/internal_config_repo.bzl#L129-L130) that it is unconditionally available, no need for this.

aignas:
Thanks!

PR #3531, opened 2026-01-22
feat: add package metadata for wheel libraries

### LINKED ISSUE #2054: Expose a `rules_license` `PackageInfo` from imported dependencies
`pip_parse` allows us to import third party dependencies, but the imports lack enough information for us to generate an SBOM. It would be useful if targets imported from third party python deps were annotated with a `PackageInfo` from `rules_license` (notably, the `purl` is incredibly useful for generating CycloneDX format SBOMs).

While it might be possible to add this information in a custom way, adopting `rules_license` allows SBOMs to be generated without adding special logic to each ruleset.

stevebarrau:
Add package_metadata rule to generated BUILD files for wheel libraries to track package provenance using PURL (Package URL) format.

This is then picked up by [supply_chain_tools](https://registry.bazel.build/modules/supply_chain_tools) to produce SBOM for python target using external dependencies.

Fixes #2054


python/private/pypi/whl_library.bzl
```diff
@@ -506,6 +506,11 @@ def _whl_library_impl(rctx):
             entry_points[entry_point_without_py] = entry_point_script_name
 
         namespace_package_files = pypi_repo_utils.find_namespace_package_files(rctx, rctx.path("site-packages"))
+        purl = "pkg:pypi/{}@{}".format(
```
aignas: What happens if we have private packages in here? Is `pkg:pypi/{}@{}` only a type of registry or the actual public PyPI?

python/private/pypi/whl_library.bzl
```diff
@@ -526,14 +531,21 @@ def _whl_library_impl(rctx):
                 "pypi_version={}".format(metadata["version"]),
             ],
             namespace_package_files = namespace_package_files,
+            purl = purl,
```
aignas: This also needs to be added above, in the other branch of the code.

aignas:
What are other things that are needed for #2054 to be resolved?

python/private/pypi/whl_library.bzl
stevebarrau: IIUC registry information is not yet present in the `metadata` object, right? IIUC the most common approach is to have an `--index-url` instruction in the `requirements_lock` file. DO you know where I can fish out this information from within the rules?

python/private/pypi/whl_library.bzl
stevebarrau: Added in 9673f8c9.

stevebarrau:
> What are other things that are needed for #2054 to be resolved?

IIUC, rules_license was the previous attempt at doing SBOM and package level metadata. The new approach would be supply_chain_tool (i.e. this PR).

CC @aiuto @shs96c

MODULE.bazel
```diff
@@ -6,8 +6,9 @@ module(
 
 bazel_dep(name = "bazel_features", version = "1.21.0")
 bazel_dep(name = "bazel_skylib", version = "1.8.2")
-bazel_dep(name = "rules_cc", version = "0.1.5")
+bazel_dep(name = "package_metadata", version = "0.0.6")
```
aiuto: 0.0.7 is out now

python/private/pypi/generate_whl_library_build_bazel.bzl
```diff
@@ -41,6 +41,12 @@ _TEMPLATE = """\
 
 package(default_visibility = ["//visibility:public"])
 
+package_metadata(
+    name = "package_metadata",
+    purl = {purl},
+    visibility = ["//:__subpackages__"],
```
aiuto, line 46: Why limited visibility when the rest is public?

stevebarrau: I followed go from the gazelle repo: https://github.com/bazel-contrib/bazel-gazelle/blob/7b9d7f36a9278df5637cdb82f660c46b8df2aae4/internal/go_repository.bzl#L432-L438

aiuto: Ah... but I wonder if they did it right :-)

Usually, tools pick up the metadata by aspect traversal, so visibility is not needed.
But... we are building out an override system that lets you apply alternate metadata for a package. People will certainly do things like splice in license and copyright declarations when they have researched it and the wheel doesn't contain it. I could make a case that those overrides should point to the label of the the thing the aspect will find.  But we *could* do string of the target.

@yannic: you did the go thing.  Why limited visiblity?

python/private/pypi/whl_library.bzl
aignas: Yes, the `--index-url` may be present in the `extra_pip_args` attribute of `whl_library` but if we know the URL to download the wheel from, we don't pass the `extra_pip_args`.

Given that sometimes people chose to use a fall-through cache mirror (Artifactory, et all) for public PyPI packages, I feel like the `purl` should be still pointing to the public PyPI and only for things not found on PyPI we should point to a particular registry.

What is the spec here that you are implementing against?

python/private/pypi/whl_library.bzl
```diff
@@ -297,6 +297,16 @@ def _extract_whl_py(rctx, *, python_interpreter, args, whl_path, environment, lo
         logger = logger,
     )
 
+def _to_purl(metadata):
```
TheGrizzlyDev: what if it's a wheel library from outside of pypi? I've worked on this before and I took this approach: https://github.com/bazel-contrib-supply-chain-extras/rules_python/pull/1/changes

python/private/pypi/whl_library.bzl
stevebarrau: https://github.com/package-url/purl-spec/blob/main/types-doc/pypi-definition.md

python/private/pypi/whl_library.bzl
stevebarrau: We could set the address of the wheel archive in the extra attributes of the purl.

darkrift:
@stevebarrau @aignas Anything missing to push this through ?

aignas:
Once the CI passes I can review.

aignas:
I'll wait until anyone comes forward and tells me if this works, also please let me know where I can find docs on how to make this useful for generating an SBOM.

darkrift:
@aignas check the sbom target example here on how to extract that information : https://github.com/bazel-contrib/supply-chain/blob/main/examples/server/BUILD

Doc is here : https://github.com/bazel-contrib/supply-chain/blob/main/docs/metadata/README.md

aignas:
Thank you for the links!

It would be nice to add a similar difftest for an sbom somewhere within our examples. Maybe examples/bzlmod would be a good place?

I won't be able to write any code next week, so anyone feel free to take over.

aignas:
Since the CI is green and this does not seem to add a lot of overhead, I'll go ahead and merge it.

aignas (approved):

PR #3535, opened 2026-01-22
feat(python): add arm64e-apple-darwin platform support

ma-oli:
MacOS supported architectures include x86_64, arm64 and arm64e. While unusual, there are certain edge cases that require setting the host platform to arm64e, in which case we probably want the python toolchain resolution to happen properly. So we're adding a new platform in this commit that will match the same arm64 toolchain on arm64e platform.

### DIFF
--- a/python/versions.bzl
+++ b/python/versions.bzl
@@ -1127,6 +1127,14 @@ def _generate_platforms():
             os_name = LINUX_NAME,
             arch = "aarch64",
         ),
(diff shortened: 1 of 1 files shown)


aignas:
Could you please run buildifier on the changes?

ma-oli:
Oops; sorry about that. Done

aignas:
Thank you!

Issue #2586, opened 2025-01-27
Create py_zipapp_binary to for building zip from py_binary


rickeylev:
Right now, py_binary has an implicit output (`{name}.zip}`) and output group (`python_zip_file`) that builds a zip file of the py_binary. This zip file is a "zipapp", basically a zip file with a top-level `__main__.py` file. See https://peps.python.org/pep-0441/ and https://docs.python.org/3/library/zipapp.html for details.

If the Bazel builtin flag `--build_python_zip` is set, then the zip output changes slightly: it is turned into a self-executable zip file by prepending some shell code to the zip file.

Having this part of py_binary itself is problematic for four main reasons:
1. It makes maintaining py_binary harder. The logic for creating the zip is subtle and introduces several branches on code that is otherwise straight forward.
2. A zipapp is just one of many ways to create a "distributable" py_binary. It doesn't make sense to bake one directly into py_binary.
3. A zipapp only works if it's pure-python code. If there are C library dependencies, then the zip _must_ be extracted, which then breaks the logic for handling zipapp-based startup.
4. A zip file doesn't have native support for symlinks. This makes it incompatible with bootstrap=script and how it creates a venv with symlinks to the desired python interpreter. Working around this requires hacks being applied at runtime with their own problems.

What to do:

* [ ] Create a `py_zipapp` rule. As input, they take a `py_binary` or `py_test`. As output, they produce a zipapp zip file.
* [ ] Create a `py_zipapp_binary` and `py_zipapp_test` rule. As input, they take a `py_zipapp` target. As output, they produce a self-executable zip file.
* [ ] As part of the above: create a tool to generate zip files that contain symlinks. Use `File.is_symlink` to detect if inputs should be symlinks or not. For a python-based implementation, how to create a symlink: https://stackoverflow.com/questions/35782941/archiving-symlinks-with-python-zipfile . It doesn't necessairily have to be implemented in Python. Another language is fine, however, if it's a compiled language, then we'll need to update rules_python handle building and consuming prebuilt artifacts.

I'll note that there are rule sets, e.g. rules_pkg that support somewhat arbitrary creation of zip files. Using one of those would also be fine.

arrdem:
This is almost exactly the strategy that my rules_zapp implementation used, for all it was the first ruleset I wrote. I'd be open to contributing something along those lines if that's of interest.

rickeylev:
Yes, that sounds pretty promising!

rickeylev:
[...]

PR #3539, opened 2026-01-25
feat: add py_zipapp_binary and test rules for zipapp support

rickeylev:
This adds Python zipapp support as a separate rule that works with a py_binary.

This will eventually replace the zipapp logic that is directly part of py_binary.

The high level behavior should be the same, but there's a few differences and improvements:

Changes:
* Output name is `{name}.pyz`
* Whether the output is executable is controlled by attribute, not flag.

Improvements:
* Compression level can be explicitly controlled.
* Re-unzipping optimization: only unzip if the zip's hash changed
* Zip output is more deterministic (input order doesn't affect output)
* runfiles.symlinks, runfiles.root_symlinks, and raw symlinks are supported
* Zipping venv-based binary is supported
* Lower analysis-phase cost
* Self-execution doesn't require python itself

For now, this lacks Windows support for lack of time (it requires plumbing in the
"windows exe maker" stuff or equivalent for self-executable zip support).

Along the way:
* Add PyExecToolsInfo.exec_runtime: an easier way to get the RBE-compatible runtime
* Add PyExecutableInfo fields to aid packaging a binary.
* Set `--verbose_failures` in CI config to give more info when actions fail.
* Make `--visible_for_testing` a command line flag. This makes it a bit easier to
  debug toolchain issues by showing the source of the toolchain.
* Add `actions_run` helper: this makes it easier to use py scripts as build actions.

Work towards https://github.com/bazel-contrib/rules_python/issues/3324 https://github.com/bazel-contrib/rules_python/issues/2586


rickeylev:
@gemini-code-assist review

rickeylev:
OK, ready for review. PTAL

python/private/zipapp/py_zipapp_rule.bzl
```diff
@@ -0,0 +1,317 @@
...
+    return zip_main_py
+
+def _map_zip_empty_filenames(list_paths_cb):
+    return ["rf-empty|" + path for path in list_paths_cb().to_list()]
+
+def _map_zip_runfiles(file):
+    return "rf-file|" + _is_symlink(file) + "|" + file.short_path + "|" + file.path
+
+def _map_zip_symlinks(entry):
+    return "rf-symlink|" + _is_symlink(entry.target_file) + "|" + entry.path + "|" + entry.target_file.path
```
aignas, line 54: nit: I remember that `"%s" % "foo"` being faster than `"{}".format`. Is `+` even better to use than `"%s"` or are they equivalent?

python/private/zipapp/py_zipapp_rule.bzl
```diff
@@ -0,0 +1,317 @@
...
+        output = preamble,
+        substitutions = substitutions,
+        is_executable = True,
+    )
+    return preamble
+
+def _create_self_executable_zip(ctx, preamble, zip_file):
+    pyz = ctx.actions.declare_file(ctx.label.name + ".pyz")
+    args = ctx.actions.args()
+    args.add(preamble)
```
aignas, line 165: nit: why not use `add_all`?

python/private/zipapp/py_zipapp_rule.bzl
```diff
@@ -0,0 +1,317 @@
...
+    "_zipper": attr.label(
+        cfg = "exec",
+        default = "//tools/zipapp:zipper",
+    ),
+}
+_TOOLCHAINS = [EXEC_TOOLS_TOOLCHAIN_TYPE]
+
+py_zipapp_binary = rule(
+    doc = """
+Packages a `py_binary` as a Python zipapp.
```
aignas, line 293: Is the `zipapp` the same as https://docs.python.org/3/library/zipapp.html?

python/private/py_exec_tools_info.bzl
```diff
@@ -39,11 +39,27 @@ toolchain.
...
+
+:::{versionadded} VERSION_NEXT_FEATURE
+
+In prior versions, the equivalent can be obtained using:
+```
+exec_runtime = (
+    ctx.toolchains["@rules_python//python:exec_tools_toolchain_type"].
+        exec_tools.exec_interpreter[platform_common.ToolchainInfo].py3_runtime
+    )
+```
```
aignas, line 61: This is a great addition!

python/private/py_exec_tools_toolchain.bzl
```diff
@@ -104,6 +110,12 @@ def _current_interpreter_executable_impl(ctx):
     # re-exec. If it's not a recognized name, then they fail.
     if runtime.interpreter:
         executable = ctx.actions.declare_file(runtime.interpreter.basename)
+
+        # NOTE: Using ctx.actions.symlink() here doesn't always work with RBE
+        # because it's not guaranteed that it will materialize as a symlink, but
+        # we rely on it being a symlink so that Python can find its actual
```
aignas, line 116: what does this mean to me as a maintainer. Do some RBE implementation not work with our code? Does it not work sometimes? Do we have the RBE support for this disabled because of this?

tools/private/zipapp/exe_zip_maker.py
```diff
@@ -0,0 +1,40 @@
+import hashlib
```
aignas, line 1: Is this private to the `zipapp`? I would personally have it in `python/private/zipapp/exe_zip_maker.py` until somebody comes along and asks for this to be exposed.

tools/private/zipapp/zipper.py
```diff
@@ -0,0 +1,230 @@
+import argparse
```
aignas, line 1: ditto

CHANGELOG.md
```diff
@@ -82,6 +82,17 @@ END_UNRELEASED_TEMPLATE
   `true`, a `py_*` target's `pyi_srcs` attribute will be set if any `.pyi` files
   that are associated with the target's `srcs` are present.
   ([#3354](https://github.com/bazel-contrib/rules_python/issues/3354)).
+* (zipapp) {obj}`py_zipapp_binary` and {obj}`py_zipapp_test` rules added. These
+  will replace `--build_python_zip` and the zip output group of
```
aignas, line 86: when do you want to replace these? Next major version? Should we add some deprecation messages?

tests/py_zipapp/BUILD.bazel
```diff
@@ -0,0 +1,66 @@
...
+    },
+    main = "main.py",
+    target_compatible_with = NOT_WINDOWS,
+    deps = ["@dev_pip//absl_py"],
+)
+
+py_zipapp_binary(
+    name = "system_python_zipapp",
+    binary = ":system_python_bin",
+    target_compatible_with = NOT_WINDOWS,
```
aignas, line 56: What about windows and zipapps? If I understand correctly, Windows most of the time is building zips under the hood?

It would be good to have a comment somewhere about plans/strategy of this.

aignas (approved):
In general LGTM. Thank you for easy to read code!

python/private/zipapp/py_zipapp_rule.bzl
rickeylev: I think + was slightly faster, but don't entirely recall.

python/private/zipapp/py_zipapp_rule.bzl
rickeylev: No particular reason.

python/private/zipapp/py_zipapp_rule.bzl
rickeylev: It's a different underlying implementation, but yes, it uses the same spec. (basically a zip with `__main__.py` defined at the root)

python/private/py_exec_tools_toolchain.bzl
rickeylev: It's more of an interaction between the RBE protocol and how some programs (Python included) special case symlinks.

The contract of `declare_file("foo"); symlink("foo", "actual")` is "make sure there is a file 'foo', and when reading its content, the content is the same as 'actual'" -- notably, it doesn't require foo to be a symlink or not. In RBE, it makes it a regular file that is a copy of the actual file. Usually NBD.

Enter Python. It reads through the symlinks, e.g. given `/home/user/python3 --> /usr/bin/python3`, Python will read through the symlink and use /usr as PYTHONHOME, so then it can find its stdlib etc.

Enter Bazel's executable rule restriction: Bazel requires that the DefaultInfo.executable file be an output of the target returning DefaultInfo. So what do you do if you're forwarding something along, such as when you want to forward the interpreter obtained via the toolchain? Use a symlink.

> Do some RBE not work with our code

RBE works with our code. The couple places where Python is needed as a build action, but can't be done as a normal py_binary (i.e. actions that are part of the py_binary implementation), requires special care to point to the actual underlying runtime paths, not an symlinked indirection.

tests/py_zipapp/BUILD.bazel
rickeylev: Yeah, windows defaults to --build_python_zip=true. I don't know why, though.

I added some todo tasks to #2586 and a todo here. The rough plan is to get py_zipapp working for windows and not default windows to zipapps.

tools/private/zipapp/exe_zip_maker.py
rickeylev: Yeah, this is all private. I'll move it to tools/private.

tools/private/zipapp/zipper.py
rickeylev: Will move to tools/private

CHANGELOG.md
rickeylev: Yeah, eventually, it will need a breaking change and major version bump.

It's possible to refactor the py_binary macro to call the py_zipapp_binary macro (this would get the logic out of py_binary directly, and just have some extra target defs in py_binary). IDK if I'm willing and able to do that rigmarole. Ultimately, the flag and implicit outputs are destined for removal.

> add deprecation messages

Yes, good idea.

tools/private/zipapp/exe_zip_maker.py
rickeylev: Move to tools/private

tools/private/zipapp/zipper.py
rickeylev: Moved to tools/private

CHANGELOG.md
rickeylev: Added a "add deprecation notice" to tracking issue todo list

PR #3547, opened 2026-01-26
docs (debugger): Update using debuggers how to guide on using debugpy (e.g. vscode)

shayanhoshyari:
Thanks to help from @rickeylev in the discussion in https://github.com/bazel-contrib/rules_python/issues/3481 I was able to do a debugger integration running bazel test or bazel run straight in vscode (since it uses debugpy I suspect PyCharm should also work). Added this workflow to the docs, hoping others might also find it useful.

Added page: https://rules-python--3547.org.readthedocs.build/en/3547/howto/debuggers.html 

Related to  https://github.com/bazel-contrib/rules_python/discussions/3485

### DIFF
--- a/docs/howto/debuggers.md
+++ b/docs/howto/debuggers.md
@@ -140,9 +140,13 @@ For the remainder of this document, we assume you are using vscode.
         ]
 
         if bzl_args[1:]:
-            # Append extra arguments (only applicable for 'run' mode).
-            cmd.append("--")
-            cmd.extend(bzl_args[1:])
+            if args.mode == "run":
+                # Append extra arguments for 'run' mode.
+                cmd.append("--")
+                cmd.extend(bzl_args[1:])
+            elif args.mode == "test":
+                # Append extra arguments for 'test' mode.
+                cmd.extend([f"--test_arg={arg}" for arg in bzl_args[1:]])
 
         env = {
             **os.environ.copy(),


shayanhoshyari:
> PYDEVD_RESOLVE_SYMLINKS and IDE_PROJECT_ROOTS

Here are the "docs" (in quotes since they are just github comments)

[PYDEVD_RESOLVE_SYMLINKS](https://github.com/microsoft/debugpy/issues/743#issuecomment-1112580721)

[IDE_PROJECT_ROOTS](https://github.com/microsoft/debugpy/issues/824#issuecomment-1018518479)


> Any idea how closely coupled the VS Code side and target-under-test side are? 

In theory the vscode and target under test python versions don't have to be the same. I tried 3.10 (vscode python) and 3.12 (target under test) and it looked okay. That being said it is possible to hook `launch.json` to use the bazel python, by adding  these at the cost of more boilerplate 🙈 
```
# launch.json
{
    ...
   "python" : ".vscode/python",
    ...
}

# .vscode/python
set -e

cd "$(realpath $(dirname $0)/..)"
exec bazel --quiet run \
    --run_in_cwd \
    --show_result 0 \
    @rules_python//python/bin:python \
    -- \
    "$@"
```


> My overall goal is to simplify things so that the typical bazel-bin/foo_test or bazel test //:foo_test can Just Work

I thought a bit about this, but could not come up with a good answer.

#### Pip package

If we put launcher script in a pypi package. Then we can do

```json
{
    ...   
   "module" : "bazel_shim.main", # replace .vscode/launch.py with this
}
```

But still how to get this module is a challenge. Because it needs to exist in the vscode python. Putting it in rules_python's requirements.txt won't help :( . 

#### Write a vscode plugin

Or just go to bazel's vscode plugin. https://github.com/bazel-contrib/vscode-bazel 

This probably can work out of the box, we can vendor the launcher script, and add a new debug target type. Cons are that: It is vscode specific (might be okay), and well more work as we need to write typescript and contribute to another repo.

aignas:
I have been maintaining (very low effort) pytest-bazel PyPI package and I am convinced that for pytest it is the right level of abstraction. Maybe debugpy-bazel could be similar?

PR #3553, opened 2026-01-29
fix: use powershell.exe instead of pwsh.exe for build_data_writer

### LINKED ISSUE #3552: whl_library.bzl requires PowerShell 7 (pwsh.exe for build_data_writer.ps1) but stock Windows normally has PowerShell 5.1
# 🚀 feature request

### Relevant Rules

[whl_library.bzl](https://github.com/bazel-contrib/rules_python/blob/main/python/private/pypi/whl_library.bzl#L410), [whl_extract.bzl](https://github.com/bazel-contrib/rules_python/blob/main/python/private/pypi/whl_extract.bzl#L16) and [repo_utils.bzl](https://github.com/bazel-contrib/rules_python/blob/main/python/private/repo_utils.bzl#L440)

### Description

py_executable.bzl invoked from [repo_utils.bzl](https://github.com/bazel-contrib/rules_python/blob/main/python/private/repo_utils.bzl#L440) hardcodes pwsh.exe as the executable for running build_data_writer.ps1 on Windows. PowerShell 7+ (pwsh.exe) is not installed by default on Windows — only Windows PowerShell 5.1 (powershell.exe) ships with the OS.

This causes builds to fail on stock Windows with:
```
  CreateProcessW("pwsh.exe" -File external/rules_python+/python/private/build_data_writer.ps1):
  The system cannot find the file specified. (error: 2)
```

This came up during testing of #3543 / #3549. The extraction fix works, but the pwsh.exe requirement is a separate hurdle for Windows users without PowerShell 7 installed.

  Ref: https://github.com/bazel-contrib/rules_python/issues/3543 (last two comments)

Change pwsh.exe to powershell.exe in py_executable.bzl. The script's cmdlets are compatible with Windows PowerShell 5.1, so no script changes are needed.

No alternatives for Windows 11 or earlier as the PowerShell 5.1 is shipped with stock OS.

vadikmironov:
Stock Windows ships with Windows PowerShell 5.1 (powershell.exe) but not PowerShell 7+ (pwsh.exe). Using pwsh.exe causes builds to fail on machines without PowerShell 7 installed. Also adds -ExecutionPolicy Bypass since Windows PowerShell defaults to Restricted policy which blocks .ps1 script execution.

Fixes #3552

### DIFF
--- a/python/private/py_executable.bzl
+++ b/python/private/py_executable.bzl
@@ -1369,8 +1369,12 @@ def _write_build_data(ctx):
     action_args = ctx.actions.args()
     writer_file = ctx.files._build_data_writer[0]
     if writer_file.path.endswith(".ps1"):
+        # powershell.exe is used for broader compatibility
+        # It is installed by default on most Windows versions
         action_exe = "powershell.exe"
         action_args.add_all([
+            # Bypass execution policy is needed because,
+            # by default, Windows blocks ps1 scripts.
             "-ExecutionPolicy",
             "Bypass",
             "-File",


python/private/py_executable.bzl
```diff
@@ -1369,9 +1369,13 @@ def _write_build_data(ctx):
     action_args = ctx.actions.args()
     writer_file = ctx.files._build_data_writer[0]
     if writer_file.path.endswith(".ps1"):
-        action_exe = "pwsh.exe"
-        action_args.add("-File")
-        action_args.add(writer_file)
+        action_exe = "powershell.exe"
```
rickeylev, line 1374:
```suggestion
        # powershell.exe is used for broader compatibility
        # It is installed by default on most Windows versions
        action_exe = "powershell.exe"
```

python/private/py_executable.bzl
```diff
@@ -1369,9 +1369,13 @@ def _write_build_data(ctx):
...
+        action_args.add_all([
+            "-ExecutionPolicy",
```
rickeylev, line 1378:
```suggestion
            # Bypass execution policy is needed because,
            # by default, Windows blocks ps1 scripts.
            "-ExecutionPolicy",
```

rickeylev:
The CI failure looks bogus -- buildifier can't find the version we aren't asking for? I'll just submit passed it.

rickeylev:
Thanks for the PR, too!

vadikmironov:
Thanks a lot @rickeylev and @aignas! One other thing that came up as part of this change - there is a small discrepancy in CONFIG_ID handling by .sh and .ps1 scripts. Here is the summary:

- build_data_writer.ps1 [writes CONFIG_ID $env:CONFIG_ID](https://github.com/bazel-contrib/rules_python/blob/main/python/private/build_data_writer.ps1#L4)
- build_data_writer.sh does not write CONFIG_ID
- py_executable.bzl does not pass CONFIG_ID in the env dict to the action

Minor thing of course, but either the .ps1 should drop the CONFIG_ID line to match .sh, or both scripts should include it and py_executable.bzl should pass the value. What 's your view on this? PR either way should be trivial.

rickeylev:
Oops, that's a bug. I had to remove writing CONFIG_ID because it caused action conflicts; I must have forgotten to update the ps1 script.

vadikmironov:
#3556

PR #3555, opened 2026-01-30
fix(pip): add read permissions when extracting wheels

### LINKED ISSUE #3554: (pipstar) Wheel install fails when wheel has bogus permissions
# 🐞 bug report

### Affected Rule 

 The issue is caused by the rule: pip.parse

### Is this a regression?

 Yes, the previous version in which this bug was not present was: 1.7

### Description

When building ag-ui-adk==0.4.2 via pip.parse, _find_namespace_package_files fails with a permission denied error

## 🔬 Minimal Reproduction
MODULE.bazel:
```
bazel_dep(name = "rules_python", version = "1.8.3")
python.toolchain(python_version = "3.11.9")

pip.parse(
    hub_name = "pypi",
    python_version = "3.11.9", 
)
use_repo(pip, "pypi")
```
requirements.in:
```
ag-ui-adk==0.4.2
```

BUILD
```

compile_pip_requirements(
    name = "requirements",
    src = "requirements.in",
)
```

Steps to reproduce:
```
bazel run :requirements.update        
bazel build @pypi//...
```

## 🔥 Exception or Error

<pre><code>

ERROR: /Users/tkaplan/Library/Caches/bazel/_bazel_tkaplan/dfc5910e1204e65cb55de615cec627b5/external/rules_python+/python/private/pypi/pypi_repo_utils.bzl:188:28: An error occurred during the fetch of repository 'rules_python++pip+pypi_311_ag_ui_adk':
	File "/Users/tkaplan/Library/Caches/bazel/_bazel_tkaplan/dfc5910e1204e65cb55de615cec627b5/external/rules_python+/python/private/pypi/whl_library.bzl", line 433, column 79, in _whl_library_impl
		namespace_package_files = pypi_repo_utils.find_namespace_package_files(rctx, install_dir_path)
	File "/Users/tkaplan/Library/Caches/bazel/_bazel_tkaplan/dfc5910e1204e65cb55de615cec627b5/external/rules_python+/python/private/pypi/pypi_repo_utils.bzl", line 188, column 28, in _find_namespace_package_files
		content = rctx.read(init_py)
Error in read: com.google.devtools.build.lib.vfs.FileAccessException: /Users/tkaplan/Library/Caches/bazel/_bazel_tkaplan/dfc5910e1204e65cb55de615cec627b5/external/rules_python++pip+pypi_311_ag_ui_adk/site-packages/ag_ui_adk/__init__.py (Permission denied)
ERROR: no such package '@@rules_python++pip+pypi_311_ag_ui_adk//': com.google.devtools.build.lib.vfs.FileAccessException: /Users/tkaplan/Library/Caches/bazel/_bazel_tkaplan/dfc5910e1204e65cb55de615cec627b5/external/rules_python++pip+pypi_311_ag_ui_adk/site-packages/ag_ui_adk/__init__.py (Permission denied)
ERROR: /Users/tkaplan/Library/Caches/bazel/_bazel_tkaplan/dfc5910e1204e65cb55de615cec627b5/external/rules_python++pip+pypi/ag_ui_adk/BUILD.bazel:5:12: @@rules_python++pip+pypi//ag_ui_adk:dist_info depends on @@rules_python++pip+pypi_311_ag_ui_adk//:dist_info in repository @@rules_python++pip+pypi_311_ag_ui_adk which failed to fetch. no such package '@@rules_python++pip+pypi_311_ag_ui_adk//': com.google.devtools.build.lib.vfs.FileAccessException: /Users/tkaplan/Library/Caches/bazel/_bazel_tkaplan/dfc5910e1204e65cb55de615cec627b5/external/rules_python++pip+pypi_311_ag_ui_adk/site-packages/ag_ui_adk/__init__.py (Permission denied)

</code></pre>

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
macOS 15.7.3
  </code>
</pre>

<pre>
  <code>
9.0.0
  </code>
</pre>

**Rules_python version:**

<pre>
  <code>
1.8.3
  </code>
</pre>

**Anything else relevant?**

thirtyseven:
Fix for wheels like https://files.pythonhosted.org/packages/ad/39/da8b5c0f875ccb1770349caaecd87a253949ccbcdc2c869929919d744551/ag_ui_adk-0.4.2-py3-none-any.whl where the contents do not have the read bit set.

Fixes #3554

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -243,7 +243,6 @@ use_repo(
     "rules_python_runtime_env_tc_info",
     "somepkg_with_build_files",
     "whl_library_extras_direct_dep",
-    "whl_perms_bad_perms_pkg",
     "whl_with_build_files",
 )
 


tests/pypi/whl_permissions/bad_perms_pkg-1.0-py3-none-any.whl
rickeylev, line 1: Sorry, but binary files aren't allowed.

We have a help repo-rule to construct wheels from a directory. See `python/private/internal_dev_deps.bzl`, the `whl_from_dir_repo` usages, for an example. Basially, create a directory with the files you need, then point that rule to it.

I'd be fine with having the test done in a followup PR, too.

aignas:
Hmm, well this is a bummer. Having to chmod files when extracting is not ideal.

I would say that is a missing feature in bazel's extract function. What is the reason for keeping such incorrect permissions when extracting?

I remember some users mentioned that they have no chmod on their runners, so we should think about using rusts uutils for this. However, what happens for really large wheels?

Should we just disable extracting of the wheel with Starlark for now?

thirtyseven:
This might also be a bug in the release process for this wheel, I can't see a good reason to *publish* a wheel that is broken like this. If it's not widespread and we can't assume chmod is available, maybe it can just be WAI.

rickeylev:
Lets just make the chmod call optional. If chmod is present, great, it lets us handle this edge case. If not, oh well. i.e. change the logic to ignore errors.

I'd be +1 on more advanced solutions, e.g.
* If system chmod is missing, use a tool from a bazel reference
* Have a custom extract tool instead of Bazel's extract. (I'd be particularly +1 on this because bazel's extract is known to perform poorly with large archives compared to some other tools)

But, I don't think we need to block a fix that captures the majority case. 

> Should we just disable extracting of the wheel with Starlark for now?

You mean disable pipstar? I'll defer to you on that, but I'd lean towards no.

aignas:
The extraction of the wheel and the parsing of the METADATA feature flags can be toggled independently.

Regarding a long term fix, I agree, a separate binary that is fast and performant would be better in this case.

tests/pypi/whl_permissions/bad_perms_pkg-1.0-py3-none-any.whl
rickeylev: Testing this was rather annoying. Making the files read-only meant the tool to zip them couldn't read them. So a custom tool to set permissions when written to the zip is needed. Which then added a bunch of extra stuff.

I've just removed the tests. This is a low-risk fix for a malformed input, so I'm not too worried about it.

PR #3557, opened 2026-01-31
uv.lock support

rickeylev:
This is a basic impl of adding uv.lock support to pip.parse.

It was largely vibed up by gemini. The details definitely need some work, but the
general logic of it looks pretty sound.

### DIFF
--- a/python/extensions/uv_external_deps.bzl
+++ b/python/extensions/uv_external_deps.bzl
@@ -1,0 +1,217 @@
+"""A module extension for uv external dependencies.
+
+This extension allows users to define external dependencies using uv.
+"""
+
+load("@bazel_skylib//lib:selects.bzl", "selects")
+load("//python/private:repo_utils.bzl", "REPO_DEBUG_ENV_VAR", "repo_utils")
+load("//python/private:text_util.bzl", "render")
+load("//python/private/pypi:env_marker_setting.bzl", "env_marker_setting")
+load("//python/private/pypi:pypi_repo_utils.bzl", "pypi_repo_utils")
+load("//python/private/pypi:wheel_tags_setting.bzl", "wheel_tags_setting")
+
+def _convert_uv_lock_to_json(mrctx, attr, logger):
+    python_interpreter = pypi_repo_utils.resolve_python_interpreter(
+        mrctx,
+        python_interpreter_target = attr._python_interpreter_target,
+    )
+    toml2json = mrctx.path(attr._toml2json)
+    src_path = mrctx.path(attr.srcs[0])
+
+    stdout = pypi_repo_utils.execute_checked_stdout(
+        mrctx,
+        logger = logger,
+        op = "toml2json",
+        python = python_interpreter,
+        arguments = [
+            str(toml2json),
+            str(src_path),
+        ],
+        srcs = [toml2json, src_path],
+    )
+    return stdout
+
+def _wheel_repo_impl(rctx):
+    rctx.download(rctx.attr.urls, output = "output.zip")
+    rctx.file("BUILD.bazel", "exports_files(glob(['**']))")
+
+wheel_repo = repository_rule(
+    implementation = _wheel_repo_impl,
+    attrs = {
+        "urls": attr.string_list(),
+    },
+)
+
+_PACKAGE_BUILD_TEMPLATE = """
+
+load("@rules_python//python/extensions:uv_external_deps.bzl", "define_targets")
+
+package(
+    default_visibility = ["//visibility:public"]
+)
+exports_files(glob(["**"]))
+
+define_targets(
+    name = "{package}",
+    selectors = {selectors}
+)
+"""
+
+def _hub_package_build_file(rctx, package, selectors):
+    rctx.file(
+        "{}/BUILD.bazel".format(package),
+        _PACKAGE_BUILD_TEMPLATE.format(
+            package = package,
+            selectors = render.list(selectors),
+        ),
+    )
+
+def _hub_repo_impl(rctx):
+    hub_selectors = json.decode(rctx.attr.selectors)
+
+    for package, selectors in hub_selectors.items():
+        _hub_package_build_file(rctx, package, selectors)
+
+    rctx.file("BUILD.bazel", "")
+
+hub_repo = repository_rule(
+    implementation = _hub_repo_impl,
+    attrs = {
+        "selectors": attr.string(),
+    },
+)
+
+def wheel_tags_from_wheel_url(url):
+    _, _, basename = url.rpartition("/")
+    basename = basename.removesuffix(".whl")
+    distro, _, tail = basename.partition("-")
+    version, _, tail = tail.partition("-")
+    py_tag, _, tail = tail.partition("-")
+    abi, _, tail = tail.partition("-")
+    platform, _, tail = tail.partition("-")
+    return {
+        "python_tag": py_tag,
+        "abi_tag": abi,
+        "platform_tag": platform,
+    }
+    #charset_normalizer-3.4.4-cp310-cp310-macosx_10_9_universal2.whl",
+
+def _uv_external_deps_extension_impl(mctx):
+    logger = repo_utils.logger(mctx, "uvextdeps")
+
+    # sources[distro][version][type][url] = any_of_conditions
+    # url -> info
+    sources = {}
+    hub_name = None
+    for mod in mctx.modules:
+        for hub in mod.tags.hub:
+            out = _convert_uv_lock_to_json(mctx, hub, logger)
+            lock_info = json.decode(out)
+            break
+
+    # Basically what we're doing is:
+    # URLs are given name R (repo name). Thus, we never duplicate a download.
+    # Repo R is used if conditions C are met (wheel tags, marker, and
+    # custom config settings for that lock file).
+    url_to_repo = {}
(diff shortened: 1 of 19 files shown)


python/extensions/uv_external_deps.bzl
```diff
@@ -0,0 +1,152 @@
+"""A module extension for uv external dependencies.
```
rickeylev, line 1: Ignore this file, this is a bunch of earlier prototype code and isn't used

python/private/pypi/extension.bzl
```diff
@@ -272,6 +272,7 @@ You cannot use both the additive_build_content and additive_build_content_file a
     exposed_packages = {}
     extra_aliases = {}
     whl_libraries = {}
+    uv_selectors = {}
```
rickeylev, line 275: This should probably be removed and folded into...however the env markers are parsed out of the requirements.txt lines.

python/private/pypi/hub_builder.bzl
```diff
@@ -167,6 +176,97 @@ def _pip_parse(self, module_ctx, pip_attr):
         enable_pipstar_extract = self._config.enable_pipstar_extract or self._get_index_urls.get(pip_attr.python_version),
     )
 
+def _process_uv_lock(self, module_ctx, pip_attr):
```
rickeylev, line 179: The general logic of this looks sound, but I think it's duplicating a bunch of the logic done by create_whl_repos?

python/private/pypi/hub_builder.bzl
```diff
@@ -167,6 +176,97 @@ def _pip_parse(self, module_ctx, pip_attr):
...
+    for name, pkgs in packages_by_name.items():
+        package_settings = []
+        for pkg in pkgs:
+            version = pkg["version"]
+            wheels = pkg.get("wheels", [])
+            if not wheels:
+                continue
+
+            # Use first wheel
+            wheel = wheels[0]
```
rickeylev, line 214: I just told it to use the first wheel for simplicity, but that isn't strictly correct.

I think this needs the platform-sniffing logic in uv_external_deps.bzl and/or the config_settings logic from pip

python/private/pypi/hub_builder.bzl
```diff
@@ -167,6 +176,97 @@ def _pip_parse(self, module_ctx, pip_attr):
...
+                src = src,
+                whl_library_args = whl_library_args,
+                download_only = pip_attr.download_only,
+                netrc = self._config.netrc or pip_attr.netrc,
+                auth_patterns = self._config.auth_patterns or pip_attr.auth_patterns,
+                python_version = _major_minor_version(pip_attr.python_version),
+                is_multiple_versions = False,
+                use_downloader = True,
+                interpreter = interpreter,
+                enable_pipstar = False,
```
rickeylev, line 246: Probably hardcode enable_pipstar=True ?

python/private/pypi/hub_builder.bzl
```diff
@@ -167,6 +176,97 @@ def _pip_parse(self, module_ctx, pip_attr):
...
+                enable_pipstar_extract = False,
```
rickeylev, line 247: Same here: force enable pipstar extract?

python/private/pypi/hub_builder.bzl
```diff
@@ -167,6 +176,97 @@ def _pip_parse(self, module_ctx, pip_attr):
...
+
+            markers = pkg.get("resolution-markers", [])
+            if not markers:
+                package_settings.append((repo_name, None))
+            else:
+                for m in markers:
+                    package_settings.append((repo_name, m))
+
+        if package_settings:
+            self._uv_selectors[name] = json.encode(package_settings)
```
rickeylev, line 262: Not quite sure why it's doing this list of tuples encoding thing...

python/private/pypi/hub_repository.bzl
```diff
@@ -45,6 +45,68 @@ def _impl(rctx):
     # `requirement`, et al. macros.
     macro_tmpl = "@@{name}//{{}}:{{}}".format(name = rctx.attr.name)
 
+    if rctx.attr.uv_selectors:
```
rickeylev, line 48: Ah, definitely need to refactor this to fold it into the normal code paths. This is converting the env markers into select expression, but I'm pretty sure we have that logic elsewhere already.

python/private/pypi/hub_repository.bzl
```diff
@@ -45,6 +45,68 @@ def _impl(rctx):
...
+            # repo points to the whl_library which has these targets.
+            # So actual should be repo + "//:pkg" etc.
+            
+            def make_select(suffix):
+                d = {}
+                for k, v in select_dict.items():
+                    d[k] = v + suffix
+                return render.dict(d)
+
+            content = """
```
rickeylev, line 72: And pull this out. This is generating the hub BUILD file.

python/private/pypi/pip_repository_attrs.bzl
```diff
@@ -70,4 +70,20 @@ True will become default in a subsequent release.
...
+    "uv_lock": attr.label(
+        allow_single_file = True,
+        doc = """\
+The uv.lock file to use for resolving dependencies.
+""",
+    ),
+    "_toml2json": attr.label(
+        default = Label("//tools/toml2json:toml2json.py"),
+        allow_single_file = True,
+        cfg = "exec",
```
rickeylev, line 83: lol, remove this and executable=true. They have no meaning for repo-phase attributes.

python/private/pypi/wheel_tags_setting.bzl
```diff
@@ -0,0 +1,116 @@
...
+            "//conditions:default": "other",
+        }),
+        arch = select({
+            "@platforms//cpu:x86_64": "x86",
+            "//conditions:default": "other",
+        }),
+        **wheel_tags
+    )
+
+def _wheel_tags_setting_impl(ctx):
```
rickeylev, line 38: I swear I thought we had a rule for this? This is basically evaluating a env marker expression, taking info from the toolchain etc.

tests/tools/toml2json/toml2json_test.py
```diff
@@ -0,0 +1,62 @@
...
+        fd, path = tempfile.mkstemp(suffix=".toml", dir=self.temp_dir.name)
+        with os.fdopen(fd, "wb") as f:
+            f.write(content)
+        return path
+
+    def test_basic_conversion(self):
+        toml_content = b"""
+[owner]
+name = "Tom Preston-Werner"
+dob = 1979-05-27T07:32:00-08:00
```
rickeylev, line 27: I wonder if there really is a Tom Preston-Werner born 1979-05-27 at 7:32 AM? 😂

tools/toml2json/toml2json.py
rickeylev, line 1: The toml2json tool is super simple. Just evalutations toml to a python object, then dumps it back as json. The only special thing is handling datetime objects, which is a simple 1-liner fix.

rickeylev (commented):
Question: how to pass uv.lock? Or pylock.toml. Separate attrs, or use existing requirements_lock/requirements_by_platform ?

Question: Shall I split out toml2json to a separate PR?

@aignas request: Any pointers of what code should be folded into where. Even simple hints like which function name / attributes to fold something into; Gemini has been doing pretty well with being pointed in the vague direction

shayanhoshyari:
> gemini

I have been using gemini (through gemini cli) and claude (claude code and cursor agent)

claude just eats gemini cli for lunch in terms of getting things right

python/private/pypi/hub_builder.bzl
aignas: The `parse_requirements.bzl` at the end of the logic uses `select_whl` function to select the right wheel. We should do here too.

python/private/pypi/hub_builder.bzl
```diff
@@ -111,6 +114,7 @@ def _build(self):
         },
         exposed_packages = sorted(self._exposed_packages),
         whl_libraries = self._whl_libraries,
+        uv_selectors = self._uv_selectors,
```
aignas, line 117: Ideally, the `uv_selectors` should instead be repleaced with `whl_config_setting` instances.

python/private/pypi/hub_repository.bzl
aignas: `render_pkg_aliases` or something similar is what should be used.

tests/tools/toml2json/toml2json_test.py
aignas: Surprise: https://en.wikipedia.org/wiki/Tom_Preston-Werner

aignas: And the primary source: https://github.com/toml-lang/toml?tab=readme-ov-file#example

python/private/pypi/wheel_tags_setting.bzl
aignas: `select_whl` is what you need.

aignas:
Also, there should be no need to change the `hub_repository` repository rule, all of the stuff can be done in a single function.

The platform logic that we have in `parse_requirements` function should be reused in the `uv.lock` file.

The API that I think we should have is:
```
pip.parse(
    # all of the same attrs except for `requirements*`
    uv_lock = <my_file>
)
```

PR #3559, opened 2026-02-01
feat(pypi): store necessary facts fetched from SimpleAPI

### LINKED ISSUE #2731: Write pip extension metadata to the MODULE.bazel.lock file
This has been split from #2434.

The idea of this is to leverage the feature from https://github.com/bazelbuild/bazel/issues/24777 when it is available in order to make the SimpleAPI querying more efficient.

aignas:
This allows us to cache what we store from the SimpleAPI in between the
runs.

Summary:
- Refactor the existing code to use an `await` helper to handle parallelism
  better when we need to wait for the download of the Simple API contents and
  then do some processing.
- Filter the return values to only the packages that are needed for the files
  instead of everything that we find on the index.
- This in turn allows us to store the values as facts in the lock file.

TODO:
- [ ] Add unit tests for facts code.

Fixes #2731


rickeylev:
@aignas did you want to move this forward? If I understand correctly, the facts API is how we can avoid extraneous network calls during bzlmod phase?

aignas:
Yeah, I am working on this slowly. The main logic is there but still need to finish writing extra tests for `read_simple_api` and `read_simple_api_with_facts` (and the `await` helper) functions, because they got a little more complex.

I personally think the experiment was a success and in general this speeds up the execution well enough to make it an improvement.

I also think we could add a similar facts module to fetch all of the information about existing standalone builds to make them available just how `rules_js` did this recently.

The same would go for `uv` builds.

rickeylev:
Nice, that sounds fantastic!

aignas:
I'll close this PR, because too many parts of the code need to be touch, so I'll move bit by bit in this direction.

PR #3561, opened 2026-02-02
feat(zipapp): add windows support

rickeylev:
This adds Windows support to the py_zipapp rules.

It works the same as what py_binary does: the Bazel builtin launcher.exe is
populated with the path to the zip, and the exe then handles starting Python
and running the zip.

This should be the last of the implementation, so all that remains is
deprecating the code in py_binary and removing it.

Work towards https://github.com/bazel-contrib/rules_python/issues/2586


rickeylev:
@gemini-code-assist review

rickeylev:
PTAL

PR #3576, opened 2026-02-07
chore: disable build_python_zip in bazelrcs

rickeylev:
This sets `--build_python_zip=false` in our bazelrc files to better reflect the future
state of the zipapp support and flush out issues.

Work towards https://github.com/bazel-contrib/rules_python/issues/2586


rickeylev:
note to self: closer. The edge case this revealed is `--enable_runfiles=false`, which is the default on windows. That can be partially supported / worked around by looking in the manifest for the bootstrap parts. However, fundamentally, a venv layout requires runfiles to be enabled to materialize the necessary file structure. Alternatively, I think we can default enable_runfiles=True for windows via a transition.

PR #3582, opened 2026-02-10
fix(pip): preserve PEP 508 URL-based requirements when extract_url_srcs=False

jsharpe:
pip_parse (via pip_repository) passes extract_url_srcs=False to parse_requirements. The _package_srcs() function silently dropped PEP 508 URL-based requirements (pkg @ https://...) in this mode because _add_dists() returns can_fallback=False for URL requirements, causing them to fall through to the `continue` statement.

Fix the elif condition to also accept the case where extract_url_srcs is False but a valid URL dist exists, falling back to pip to handle the URL requirement directly.


jsharpe:
CI failures seem unrelated to this change.

PR #3588, opened 2026-02-11
fix(pypi): normalize extras in requirement strings per PEP 685

### LINKED ISSUE #3587: Pipstar extras not matched due to hyphen vs underscore normalization
# 🐞 bug report

### Affected Rule

`pip.parse` / `whl_library` (Pipstar extras resolution)

### Is this a regression?

Yes, this worked in `rules_python` 1.6.3 (non-Pipstar path). The extras fix in #3468 correctly passes extras down the call stack, but a normalization mismatch prevents them from being matched against `Requires-Dist` markers.

### Description

PR #3468 fixed the issue of extras not being passed down in Pipstar. However, extras containing **hyphens** in `requirements.txt` are not normalized to **underscores** before being matched against the wheel's `Requires-Dist` metadata, which uses the PEP 685-normalized form (underscores).

For example, with `sqlalchemy[asyncio,postgresql-asyncpg,postgresql-psycopg2binary]==2.0.36` in `requirements.txt`, the generated `BUILD.bazel` for the sqlalchemy `whl_library` shows:

```python
extras = [
    "asyncio",
    "postgresql-asyncpg",        # hyphens from requirements.txt
    "postgresql-psycopg2binary",
],
requires_dist = [
    # ...
    "asyncpg ; extra == 'postgresql_asyncpg'",             # underscores in wheel metadata
    "psycopg2-binary ; extra == 'postgresql_psycopg2binary'",  # underscores in wheel metadata
    # ...
],
```

The extras list uses hyphens (`postgresql-asyncpg`) while `Requires-Dist` markers use underscores (`postgresql_asyncpg`). The comparison fails silently, so `psycopg2-binary` and `asyncpg` are never added as transitive dependencies of `sqlalchemy`.

Note that `asyncio` (no hyphens) works correctly because it doesn't need normalization.

**Workaround**: Use underscores instead of hyphens in the extras specifier in `requirements.in`:
```
sqlalchemy[asyncio,postgresql_psycopg2binary,postgresql_asyncpg]==2.0.36
```

## 🔬 Minimal Reproduction

`requirements.in`:
```
sqlalchemy[asyncio,postgresql-psycopg2binary,postgresql-asyncpg]==2.0.36
psycopg2-binary~=2.9.9
asyncpg~=0.30.0
```

`MODULE.bazel`:
```python
pip.parse(
    hub_name = "pypi",
    python_version = "3.11",
)
use_repo(pip, "pypi")
```

Then query:
```bash
bazel query 'kind("py_library", deps(@pypi//sqlalchemy))' | grep -E 'psycopg|asyncpg'
# Returns empty — psycopg2_binary and asyncpg are missing
```

On `rules_python` 1.6.3 (non-Pipstar), the same query correctly returns:
```
@pypi//psycopg2_binary:pkg
@pypi//asyncpg:pkg
```

## 🔥 Exception or Error

<pre><code>
No error at build time. At runtime, tests fail with:
ModuleNotFoundError: No module named 'psycopg2'
</code></pre>

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
macOS 15.3.1 (darwin 25.2.0, arm64)
  </code>
</pre>

<pre>
  <code>
bazel 8.1.1
  </code>
</pre>

**Rules_python version:**

<pre>
  <code>
1.8.4
  </code>
</pre>

**Anything else relevant?**

The root cause appears to be that `extras` parsed from the requirement string in `whl_library.bzl` (via `requirement(rctx.attr.requirement).extras`) preserves the original hyphenated form from `requirements.txt`, but the `Requires-Dist` markers in wheel `METADATA` use PEP 685-normalized names (hyphens → underscores). The matching logic needs to normalize extras before comparison.

Related: #3352, #3468

kevinpark1217:
## Summary

Extras parsed from requirement strings (e.g., from `requirements.txt`) were not being normalized, causing mismatches when evaluating PEP 508 marker expressions. For example, `sqlalchemy[postgresql-psycopg2binary]` would fail to resolve `psycopg2-binary` as a transitive dependency because the wheel METADATA marker expression `extra == "postgresql_psycopg2binary"` uses the underscore-normalized form (per PEP 685), while the extras set retained the original hyphenated form from the requirement string.

## Before

```
# requirements.txt
sqlalchemy[postgresql-psycopg2binary]==2.0.36

# Parsed extras: ["postgresql-psycopg2binary"]
# Marker evaluation: "postgresql-psycopg2binary" != "postgresql_psycopg2binary" -> MISS
# Result: psycopg2-binary NOT included as a dependency
```

## After

```
# requirements.txt
sqlalchemy[postgresql-psycopg2binary]==2.0.36

# Parsed extras: ["postgresql_psycopg2binary"]  (normalized)
# Marker evaluation: "postgresql_psycopg2binary" == "postgresql_psycopg2binary" -> MATCH
# Result: psycopg2-binary correctly included as a dependency
```

## Changes

- **`python/private/pypi/pep508_requirement.bzl`**: Apply `normalize_name()` to each extra during requirement parsing, consistent with how the package name is already normalized.
- **`tests/pypi/pep508/requirement_tests.bzl`**: Updated existing test expectation for case normalization and added test case for hyphenated extras (`sqlalchemy[asyncio,postgresql-psycopg2binary,postgresql-asyncpg]`).
- **`tests/pypi/pep508/deps_tests.bzl`**: Added `test_extras_with_hyphens_are_normalized` integration test confirming that dependencies gated behind hyphenated extras are correctly resolved.
- **`CHANGELOG.md`**: Added entry under Unreleased > Fixed.

Fixes #3587

### DIFF
--- a/python/private/pypi/BUILD.bazel
+++ b/python/private/pypi/BUILD.bazel
@@ -262,6 +262,7 @@ bzl_library(
     name = "pep508_deps_bzl",
     srcs = ["pep508_deps.bzl"],
     deps = [
+        ":pep508_env_bzl",
         ":pep508_evaluate_bzl",
         ":pep508_requirement_bzl",
         "//python/private:normalize_name_bzl",
@@ -272,6 +273,7 @@ bzl_library(
     name = "pep508_env_bzl",
     srcs = ["pep508_env.bzl"],
     deps = [
+        "//python/private:normalize_name_bzl",
         "//python/private:version_bzl",
     ],
 )


python/private/pypi/pep508_requirement.bzl
```diff
@@ -45,7 +45,7 @@ def requirement(spec):
     extras_unparsed, _, _ = extras_unparsed.partition("]")
     for char in _STRIP:
         requires, _, _ = requires.partition(char)
-    extras = extras_unparsed.replace(" ", "").split(",")
+    extras = [normalize_name(e) for e in extras_unparsed.replace(" ", "").split(",") if e]
     name = requires.strip(" ")
     name = normalize_name(name)
```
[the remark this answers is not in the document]
kevinpark1217: Thanks for the suggestion! I've already addressed this in the latest push — the `[""]` fallback was removed entirely. The downstream consumer (`_resolve_extras` in `pep508_deps.bzl`) already handles the empty-extras case with `extras or [""]`, so the fallback in `requirement()` was unnecessary. The current code is simply:

```starlark
extras = [normalize_name(e) for e in extras_unparsed.replace(" ", "").split(",") if e]
```

tests/pypi/pep508/deps_tests.bzl
```diff
@@ -218,6 +218,37 @@ def test_span_all_python_versions(env):
...
+    but wheel METADATA uses underscores in marker expressions
+    (e.g., extra == "db_backend") per PEP 685, the deps should still resolve.
+
+    Args:
+        env: the test environment.
+    """
+    requires_dist = [
+        "bar",
+        'baz-lib; extra == "db_backend"',
+        'qux-async; extra == "async_driver"',
```
aignas:
```suggestion
        'qux-async; extra == "async-driver"',
```

aignas (commented):
Thank you for the fix. What do you think about changing the test just a little to Ensure that the code that does dependency resolution is always correct? I am thinking that we probably don't need to test un-normalized `extras` passing to the `deps` function in the test, but having the `requires_dist` un-normalized would be good.

tests/pypi/pep508/deps_tests.bzl
kevinpark1217: Good catch — applied the suggestion and went further: the `requires_dist` marker expressions now use hyphens in both lines (`extra == "db-backend"` and `extra == "async-driver"`), and I added PEP 685 normalization in `marker_expr()` in `pep508_evaluate.bzl` so the literal string side of `extra` comparisons is normalized before the comparison. This follows the same pattern as the existing `_ENV_ALIASES` normalization for platform variables. The env side (from `requirement()` parsing) is already normalized, so only the marker literal needs it.

All 24 pep508 tests pass including the updated integration test.

python/private/pypi/pep508_evaluate.bzl
```diff
@@ -316,6 +323,10 @@ def marker_expr(left, op, right, *, env, strict = True):
             # See the note above on normalization
             right = env.get(_ENV_ALIASES, {}).get(var_name, {}).get(right, right)
 
+        # See the note above on PEP 685 normalization
+        if var_name == "extra":
+            right = normalize_name(right)
+
```
kpark-hrp: @aignas Not sure if there is a more cleaner way to do this. Open to suggestions here.

aignas: So best would be to do it:
* normalize before passing here: https://github.com/bazel-contrib/rules_python/blob/main/python/private/pypi/pep508_deps.bzl#L53 and after receiving.
* normalize here: https://github.com/bazel-contrib/rules_python/blob/main/python/private/pypi/pep508_deps.bzl#L47

I think you have already done the second part.

Then as you pointed out, the marker expressions can include the non-normalized version, which means that we have to normalize the marker expression in some way and the best that I can see is what you did. We probably want to normalize in the comparison function both the left and right in there. Maybe then we don't have to worry too much about normalizing in the other places. Let me take a look at that code and add further comments.

python/private/pypi/pep508_evaluate.bzl
```diff
@@ -307,6 +308,12 @@ def marker_expr(left, op, right, *, env, strict = True):
             # The following normalizes the values
             left = env.get(_ENV_ALIASES, {}).get(var_name, {}).get(left, left)
```
aignas: So this function is about transforming values to make it easier to transform the `left` hand side. Maybe we could abstract/generalize this to do it for `extra`.

So instead of the concept of env aliases, we could have a concept of env normalization functions. Something like:
```
if _ENV_ALIASES in env:
    left = env.get(_ENV_ALIASES, {}).get(var_name, lambda x: x)(x)
```

aignas: The same applies for `right`.

python/private/pypi/pep508_evaluate.bzl
kevinpark1217: Implemented in 3e48fee. Refactored `_ENV_ALIASES` from dict-based lookups to function-based normalization as you suggested:

- `pep508_env.bzl`: `create_env()` now registers `normalize_name` as the `"extra"` normalizer and wraps `platform_machine_aliases` in a lambda, both as `(value) -> normalized_value` functions
- `pep508_evaluate.bzl`: `marker_expr()` now calls `env.get(_ENV_ALIASES, {}).get(var_name, lambda x: x)(left)` — a single generalized pattern for all variable normalization, no more separate `if var_name == "extra"` blocks
- `pep508_deps.bzl`: `_evaluate_any` and `_add_reqs` call `create_env()` inline (matching the codebase pattern in `flags.bzl` and `env_marker_setting.bzl`) so `_aliases` normalization is available during marker evaluation

All 24 pep508 tests pass.

python/private/pypi/pep508_evaluate.bzl
kevinpark1217: Done — adopted exactly this pattern in 3e48fee:

```starlark
left = env.get(_ENV_ALIASES, {}).get(var_name, lambda x: x)(left)
```

And for `right` in the else branch. The normalization functions are registered in `create_env()` in `pep508_env.bzl`:

```starlark
"_aliases": {
    "extra": normalize_name,
    "platform_machine": lambda x: platform_machine_aliases.get(x, x),
},
```

This unifies platform alias handling and PEP 685 extra normalization into a single mechanism.

aignas:
It seems that the upcoming bazel version is failing with weird errors. No action item here, just the observation of the CI state at the moment.

PR #3593, opened 2026-02-13
fix(pypi): update pypi tooling deps to setuptools 82, packaging 26

### LINKED ISSUE #3532: `packaging==24.0` is incompatible with `setuptools==78.1.1`
# 🐞 bug report

### Affected Rule

When updating requirements I hit the following error:
```
 ImportError: Cannot import `packaging.licenses`.
                Setuptools>=77.0.0 requires "packaging>=24.2" to work properly.
                Please make sure you have a suitable version installed.
```

### Is this a regression?

Yes, but I am not sure when it was introduced.

### Description

In `//python/private/pypi/deps.bzl` there is a list of pypi packages with `packaging-24.0` and `setuptools-78.1.1`, which clearly doesn't satisfy the setup tools requirement described in the error. 

I have a branch where I have run `bazel run //tools/private/update_deps:update_pip_deps` to update all the packages, and that seems to work fine, but I can see from the git history that `packaging` was manually downgraded in commit 309ee59 with commit message:
```
revert: Updated pip and packaging versions to work with free-threading packages (https://github.com/bazel-contrib/rules_python/pull/2514) (https://github.com/bazel-contrib/rules_python/pull/2584)
This reverts commit https://github.com/bazel-contrib/rules_python/commit/fbf8bc10a466c498fc80c6b58c939a87a8d9e929 (https://github.com/bazel-contrib/rules_python/pull/2514)

Also, update the CHANGELOG about the reverting.

Fixes https://github.com/bazel-contrib/rules_python/issues/908, which is about the `pip-compile` not using the right files
for performing the locking. It seems that the `pip` upgrade regressed
this error.
```

I don't know if this is still a problem, but if not a simple solution could be to just update all the packages or only `packaging` to a new enough release.

## 🔬 Minimal Reproduction

Since the problem seemed clear I haven't made a minimal reproduction, please let me know if that is still desired. 

## 🔥 Exception or Error

```
 ImportError: Cannot import `packaging.licenses`.
                Setuptools>=77.0.0 requires "packaging>=24.2" to work properly.
                Please make sure you have a suitable version installed.
```

## 🌍  Your Environment

**Operating System:**

Both macOS and Linux.

9.0.0

**Rules_python version:**

1.8.1

**Anything else relevant?**

No

Danielkonge:
This fixes the incompatibility between `packaging==24.0` and `setuptools==78.1.1` by just updating the requirements. 

This works as intended in my own repositories (where I had problems with requirement updates before), but I am not sure if there are still some issues like what caused the original revert of the `packaging` update. I checked and I don't see any absolute paths in the compiled lock files (as originally mentioned in #908), but I don't know if there are any other known issues.

Another solution is to just pick compatible versions of `packaging` and `setuptools` some other way. 

This closes #3532.

Note: I haven't updated the changelog, since I was not sure if this solution is even desirable? Or if you prefer to only change `packaging` and `setuptools` versions?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,7 +58,8 @@ END_UNRELEASED_TEMPLATE
 
 {#v0-0-0-changed}
 ### Changed
-* (pypi) Updated the pypi tooling dependencies.
+* (pypi) Update dependencies used for `compile_pip_requirements`, building
+  sdists in the `whl_library` rule and fetching wheels using `pip`.
 
 {#v0-0-0-fixed}
 ### Fixed


Danielkonge:
> ## Code Review
> This pull request attempts to update internal PyPI dependencies. However, a critical security vulnerability has been identified: the updated dependency versions and package formats are highly suspicious and do not align with official PyPI releases, suggesting a potential supply chain attack. Specifically, versions for `pip`, `packaging`, `setuptools`, `tomli`, and `zipp` are significantly ahead of their official counterparts. The `tomli` update is particularly concerning as it introduces a platform-specific wheel for a pure-python package, which is a common indicator of malicious intent, and also points to an invalid URL that will break builds. It is strongly recommended to reject this PR and thoroughly investigate the environment that generated these updates.

I can see that `tomli` got the wrong version (a mac version), so I will fix that. I am not really sure about the other notes?

rickeylev:
For lack of another option, I don't see why to not update. @aignas any objection?

examples/pip_parse/requirements_windows.txt
```diff
@@ -240,7 +240,7 @@ urllib3==1.26.20 \
     --hash=sha256:0ed14ccfbf1c30a9072c7ca157e4319b70d65f623e91e7b32fadb2853431016e \
     --hash=sha256:40c2dc0c681e47eb8f90e7e27bf6ff7df2e677421fd46756da1161c39ca70d32
     # via
-    #   -c ./constraints_urllib3.txt
```
aignas, line 243: I wonder if this is OK or if this is a sign of potential regressions.

TBH, I would love to potentially replace the `compile_pip_requirements` with our `uv` rule. It may be easier to work with and bump deps for.

Then these deps would be only for `sdist` in the repository phase. I am fine to bump this.

Danielkonge:
Should I add anything to the changelog, or do you think this is fine to merge as is?

aignas:
Yes please, add a changelog item about the change.

CHANGELOG.md
```diff
@@ -80,6 +80,7 @@ END_UNRELEASED_TEMPLATE
   `@rules_python//command_line_option:enable_runfiles=INHERIT` to the
   `config_settings` attribute. NOTE: `enable_runfiles=true` will
   soon become **required for Windows**.
+* (pypi) Updated the pypi tooling dependencies.
```
aignas: Let's create a new section for this change (in the unreleased), because we would like to not be shipping this with 1.9, which we have already made a release candidate for.

Danielkonge: I moved the updated part to a new section now (assuming I understood the template correctly).

Danielkonge: I am not really sure how my last commit broke CI, but if I used the template wrong, feel free to push to this branch or let me know, what I need to change.

CHANGELOG.md
```diff
@@ -47,6 +47,27 @@ BEGIN_UNRELEASED_TEMPLATE
...
+
+[0.0.0]: https://github.com/bazel-contrib/rules_python/releases/tag/0.0.0
+
+{#v0-0-0-removed}
+### Removed
+* Nothing removed.
+
+{#v0-0-0-changed}
+### Changed
+* (pypi) Updated the pypi tooling dependencies.
```
```suggestion
* (pypi) Update dependencies used for `compile_pip_requirements`, building
  sdists in the `whl_library` rule and fetching wheels using `pip`.
```

aignas (approved):
LGTM

PR #3599, opened 2026-02-14
perf(py_wheel): defer depset expansion to execution time

bd-dstodolsky:
This is the `py_wheel` counterpart to the analysis-time performance work done
for `py_binary`/`py_test` in #3381 and #3442,   deferring depset
expansion to execution time.  Tested under Bazel 8.5.1 and 9.0.0.


rickeylev (changes requested):
Thanks for the optmization, +1 to it in general.

This looks entirely AI generated, so I'm not looking at it particularly deeply.

Trim the PR description. It becomes the commit description, so it needs to be concise and relevant for future people reading through the commit log. If there's information to keep, post it as a comment.

Remove the tests that rely on timing. CI is not a reliable environment for performance tests.

Do not create shell tests that invoke bazel as a subprocess. Such integration tests are expensive and difficult to debug.

If any test is created, it should be an analysis test, using rules_testing, that verifies the args are passed to the action as a manifest file.

I'm fine with no tests here; how the content is passed is an implementation detail, so existing tests should verify correctness.

bd-dstodolsky:
Thanks. Tests removed and comments shortened. Note the performance test was tagged manual so it would not run in CI, but happy to remove; I had originally done it just as a benchmark but then decided to convert to a test, but happy to eliminate. Quick review appreciated!

rickeylev:
Thanks, looks good!

PR #3600, opened 2026-02-15
chore: add a script to easily checkout prs

rickeylev:
This adds a script I've had for awhile to make it easier to grab PRs and work on them.

It makes grabbing and editing a user's PR much easier, especially with some aliases
setup:

```
git pr-checkout <PR url>
...edit files...
git pr-push
```

While things like `gh pr checkout` exist, it doesn't well handle the "triangle"
workflow and some other edge cases.


aignas:
I use the gh CLI and you just need to do gh pr checkout to checkout a particular PR.

rickeylev:
I've had endless trouble trying to use plain `gh pr checkout`. Checkout out usually worked, but updating and pushing would frequently break. :shrug:

PR #3611, opened 2026-02-18
fix: better bootstrap manifest handling

rickeylev:
Changes to the bootstrap to better handle RUNFILES_MANIFEST_FILE mode

DO NOT MERGE: await #3610

### DIFF
--- a/python/private/python_bootstrap_template.txt
+++ b/python/private/python_bootstrap_template.txt
@@ -10,13 +10,13 @@ import sys
 import os
 import subprocess
 import uuid
-# runfiles-relative path
 # NOTE: The sentinel strings are split (e.g., "%stage2" + "_bootstrap%") so that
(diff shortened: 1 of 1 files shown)


python/private/python_bootstrap_template.txt
```diff
@@ -165,6 +165,29 @@ def print_verbose(*args, mapping=None, values=None):
         else:
             print("bootstrap: stage 1:", *args, file=sys.stderr, flush=True)
 
+def maybe_find_in_manifest(rf_root_path, show_lines=False):
+    if not os.environ.get("RUNFILES_MANIFEST_FILE"):
```
dougthor42, line 169: Nit: mixing 4-space indent and 2-space indent, here and a couple other places.

It looks like most of the file is 2-space.

PR #3620, opened 2026-02-20
feat: also export `lib` symlink from python host repository

DoDoENT:
This PR adds `lib` to the list of exported symlinks in the generated host toolchain. This is useful when a Python host interpreter is used as a script runner during the build of other toolchains, where simple `py_binaries` cannot be used.

An example where I need this can be seen in [this emscripten issue](https://github.com/emscripten-core/emsdk/issues/1675).

If you think there is a better way of solving this problem, please let me know.


rickeylev:
I'm pretty sure this isn't what you want to do. I don't think it work reliably, either, since "lib" is a directory, and passing those around directly isn't well supported by Bazel.

The `@python_3_13_host//:python` target is only appropriate for usage during the repository phase. The "host" it refers to working with is the one Bazel is running on.

It more sounds like your case is running a build tool during the normal build phase, i.e. ctx.actions.run. For that to be sound, it needs to find python through toolchain resolution, which ensures a python compatible with where bazel chooses to run the tool.

Try doing something like done here: https://github.com/bazel-contrib/rules_python/blob/589c7eea22a11fc455ae2833dde7961c351d3cdb/python/private/common.bzl#L607-L610

* Lookup up the exec runtime via toolchain resolution (L605)
* Add the interpreter and its files to the action inputs (L610)

DoDoENT:
> Try doing something like done here:

I tried that first, and it didn't work. It kept searching for the WASM version of Python, i.e., it always resolved the toolchain for the WASM platform instead of the exec platform. I then introduced a transition to force resolution in the exec platform, but I still couldn't get a runfiles for the interpreter in the exec platform that need to be given to the toolchain config in the target platform. See [this](https://github.com/DoDoENT/emsdk/blob/fix/system-python/bazel/remote_emscripten_repository.bzl#L65) - this is a repository rule, so it needs a "repository context" Python interpreter's runfiles that need to be injected as "compiler files" for the `cc_toolchain` that is being built. I also tried creating a `py_binary` and injecting that, but `create_cc_toolchain_config_info` [accepts](https://github.com/DoDoENT/emsdk/blob/main/bazel/emscripten_toolchain/toolchain.bzl#L1142) only string to string map for tools, where each string refers to the path within the C++ toolchain's repository and treats it as a native executable, so `py_binary` labels can't be used here. 

Namely, Emscripten is just a bunch of Python scripts that impersonate a C++ compiler. So, the `rules_python` here is a kind of "toolchain for a toolchain", and I need the interpreter and its runfiles at the repository phase, which makes `python_3_14_host//:python` just the right thing.

However, even if I were wrong and the emscripten issue could be resolved differently, I still think that this fix is necessary, as even during the repository phase, the interpreter may have a need to import built-in Python modules, which is currently not available as the `lib` folder is not exposed.

> I don't think it work reliably, either, since "lib" is a directory, and passing those around directly isn't well supported by Bazel.

Regarding this, everywhere I tested, the `lib` was always a symlink, and exposing this symlink was enough. Technically, I could write glob of `lib/**`, but that would just introduce a lot of files into Bazel's dependency graph, which would make analysis much slower. I consider this Python's `lib` symlink/folder equivalent to `sysroot` in C++ toolchains, which is often addressed as a directory or symlink node in Bazel, to reduce the stress on the analysis.

I tested that this change works reliably on both Linux and MacOS. You can test yourself by cloning [this branch of my playground repository](https://github.com/DoDoENT/bazel-playground/tree/feature/bazel-9) and running `bazel test //:wasm-all --config=wasm` on both Linux x86_64 and MacOS arm64 machines.

rickeylev:
> It resolved to the wasm platform instead of exec platform

Make sure its using the [exec tools toolchain type](https://rules-python.readthedocs.io/en/latest/toolchains.html#exec-tools-toolchain-type), not the "target toolchain type". The difference is the exec one matches on the exec platform instead of target platform.

If that isn't working, then it'd be best to provide a minimal reproduction showing that a custom rule isn't able to look up the exec toolchain independent of the target platform. It would mean some sort of bug in the rules_python toolchain settings. e.g something looking like:

```
def repor_rule_impl(ctx):
  <look up toolchain>
  <put path into .sh file>
  <add python runtime files to action inputs>
  <run shell script as a build action using ctx.actions.run>

repro_rule = rule(..., toolchains=<exec toolchaintype>)

bazel build --platforms=<not your host platform> :foo
```

> the lib dir being a symlink is just as fine as `lib/**`

The symlink working is likely just sandboxing (or lack there of) leaking. Bazel doesn't do anything special with symlinks or directories. They're just treated as regular files; they don't imply any automatic recursive globbing.

> See this - this is a repository rule, so it needs a "repository context" Python

Like I mentioned on slack: this isn't a repository rule. It's regular loading-phase code. The telltale sign is `native.filegroup()`, which doesn't exist in the repository phase.

DoDoENT:
I'll try that approach next week and get back if it doesn't work.

So far, I've never used Bazel's toolchain API on such a low level.

DoDoENT:
@rickeylev , thank you for your help. I managed to create [this commit](https://github.com/emscripten-core/emsdk/commit/dae73970c0b516a107197ad6d822fbed58ddb7f1) which solves my problem with non-hermetic python usage in emscripten, without the need for this PR.

I'm closing it now.

PR #3628, opened 2026-02-23
chore: factor release note checking into separate script

rickeylev:
This is to make it easier to check for missing version markers by running the script directly

### DIFF
--- a/.github/workflows/create_archive_and_notes.sh
+++ b/.github/workflows/create_archive_and_notes.sh
@@ -26,8 +26,10 @@ if [ -z "$TAG" ]; then
 fi
 # If the workflow checks out one commit, but is releasing another
 git fetch origin tag "$TAG"
-# Update our local state so the grep command below searches what we expect
+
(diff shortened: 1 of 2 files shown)


.github/workflows/create_archive_and_notes.sh
```diff
@@ -29,22 +29,6 @@ git fetch origin tag "$TAG"
 # Update our local state so the grep command below searches what we expect
 git checkout "$TAG"
 
-# Exclude dot directories, specifically, this file so that we don't
```
dougthor42, line 32: [Optional, non-blocking] It might be prudent to call `check_version_markers.sh` from this script.

+ Removes the extra step in the release flow (I'm not sure what the GH step overhead is)
+ There's no risk of forgetting to check for `VERSION_NEXT` markers (if this create_archive_and_notes.sh` script is ever called outside of CI).

rickeylev: Done

PR #3630, opened 2026-02-23
chore: silence annoying compile warnings

rickeylev:
Protobuf was causing spammy C++ compile warnings. To fix them, silence them in copts.


rickeylev:
Ugh. I forgot windows uses a different flag style. And there isn't a way to vary the flags with the global copt flag. Closing.

PR #3636, opened 2026-02-24
fix(bootstrap): manual runfiles path construction when using submodules

faximan:
This fixes https://github.com/bazel-contrib/rules_python/issues/3563. Verified by running the repro in https://github.com/mering/reproduction_rules_python_1_7.

The identified regression in https://github.com/bazel-contrib/rules_python/commit/b8e32c454a1158cd78ce4ecaef809b99bef4e5da is problematic because prepending `ctx.workspace_name` to `short_path` results in paths containing `..` (e.g., `_main/../sub+/path/to/file.py`) when building from a root module that includes other modules. This causes the `_find_runfiles_root` [logic](https://github.com/faximan/rules_python/blob/eb6ac472eb86cc263acc336a2b73982043069aae/python/private/site_init_template.py#L73), which _counts slashes_, to incorrectly calculate the runfiles root.

The fix is simply using the available `runfiles_root_path` function instead. In the example above, this makes the path simply `sub+/path/to/file.py`.


python/private/py_executable.bzl
```diff
@@ -510,10 +510,7 @@ def _create_zip_main(ctx, *, stage2_bootstrap, runtime_details, venv):
         substitutions = {
             "%python_binary%": python_binary,
             "%python_binary_actual%": python_binary_actual,
-            "%stage2_bootstrap%": "{}/{}".format(
```
aignas, line 513: Would it be possible to add a test somewhere? For example `examples/bzlmod` has submodules in it which could ensure that we can still import things.

@rickeylev Any other ideas how we could do this?

faximan: I tried and failed to write a test here tests the specific fix in this PR. As I'm very unfamiliar with this code base in general, I am afraid I'd need support to land that.

shayanhoshyari:
Related to https://github.com/bazel-contrib/rules_python/pull/3605

faximan:
> Related to #3605

Indeed! It is not sufficient to fix https://github.com/bazel-contrib/rules_python/issues/3563 though (I just verified), but it is definitely the same idea.

faximan:
@rickeylev any chance you could take a look at this? Thanks!

python/private/py_executable.bzl
faximan: @aignas any way we can move this forward?

aignas: I'd like to have a test before we can merge this. Even a happy path test case that previously did not work and now works would be better than nothing. Because we have `examples/bzlmod` which has submodule `other_module`, it should be doable IMHO. Having a `py_binary` defined in the `other_module` and the test in the `examples/bzlmod` would be the way to go.

faximan: Alright, thanks for the suggestions. I have added the most minimal test I could come up with that fails on `main` but passes with the fix. Please take a look!

faximan: The test fails on Windows for unrelated reasons to the PR. I'm not familiar with Windows usage, so please let me know if you see any obvious way to fix.

Otherwise I could disable it on Windows like this I think?


```
        "@platforms//os:windows": ["@platforms//:incompatible"],
        "//conditions:default": [],
    }),
```

faximan: Done.

faximan:
I assume this is an unrelated transient error?

```
(05:25:44) ERROR: /workdir/python/runfiles/BUILD.bazel:31:11: //python/runfiles:runfiles depends on @@+python+python_3_14_1_x86_64-unknown-linux-gnu//:python_runtimes in repository @@+python+python_3_14_1_x86_64-unknown-linux-gnu which failed to fetch. no such package '@@+python+python_3_14_1_x86_64-unknown-linux-gnu//': java.io.IOException: Error downloading [https://github.com/astral-sh/python-build-standalone/releases/download/20251202/cpython-3.14.1+20251202-x86_64-unknown-linux-gnu-install_only.tar.gz] to /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/ec321eb2cc2d0f8f91b676b6d4c66c29/external/+python+python_3_14_1_x86_64-unknown-linux-gnu/temp871320864538462325/cpython-3.14.1_20251202-x86_64-unknown-linux-gnu-install_only.tar.gz: GET returned 502 Bad Gateway or Proxy Error
```

PR #3639, opened 2026-02-27
refactor(pypi): factor out a simple implementation of the PyPI cache

aignas:
We want to keep a `dict` like interface and later we would like to use
the same interface to also do more things. I expect the cache key to
become different in the future (i.e. include requested versions in it)
so that we can check if we have the right versions in the
MODULE.bazel.lock file or if we should actually call to PyPI.

Work towards #2731

### DIFF
--- a/python/private/pypi/pypi_cache.bzl
+++ b/python/private/pypi/pypi_cache.bzl
@@ -8,16 +8,18 @@ In the future the same will be used to:
 - Store PyPI index query results as facts in the MODULE.bazel.lock file
 """
 
-def pypi_cache():
+def pypi_cache(store = None):
     """The cache for PyPI index queries."""
-    self = struct(
-        store = {},
-    )
 
-    return struct(
+    # buildifier: disable=uninitialized
+    self = struct(
+        _store = store or {},
         setdefault = lambda key, parsed_result: _pypi_cache_setdefault(self, key, parsed_result),
         get = lambda key: _pypi_cache_get(self, key),
     )
+
+    # buildifier: enable=uninitialized
+    return self
 
 def _pypi_cache_setdefault(self, key, parsed_result):
     """Store the value if not yet cached.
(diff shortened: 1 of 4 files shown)


python/private/pypi/pypi_cache.bzl
```diff
@@ -0,0 +1,45 @@
...
+    """Return the parsed result from the cache.
+
+    Args:
+        self: {type}`struct` The self of this implementation.
+        key: {type}`str` The cache key, can be any string.
+
+    Returns:
+        The {type}`struct` or `None` based on if the result is in the cache or not.
+    """
+    return self.store.get(key)
```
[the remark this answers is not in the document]
rickeylev: The top-level methods with self saves indentation and avoids leakage of scope

python/private/pypi/pypi_cache.bzl
```diff
@@ -0,0 +1,45 @@
...
+  for the same package for different hub repos.
+
+In the future the same will be used to:
+- Store PyPI index query results as facts in the MODULE.bazel.lock file
+"""
+
+def pypi_cache():
+    """The cache for PyPI index queries."""
+    self = struct(
+        store = {},
```
rickeylev: Should this be `_store` to indicate its internal, and that the setdefault/get methods re supposed to be used?

rickeylev (approved):
This change looks too simple?

The cache is changing from a dict to custom object. How is this working? Ah, is it because it just happens that only e.g. get() and setdefault() are used, not `[]` ?

python/private/pypi/pypi_cache.bzl
aignas: OK, will do that.

aignas:
Yeah, I wanted to first create a seam and change the usage to not assume a dict.

Later this will grow more meat.

PR #3640, opened 2026-03-01
chore: cleanup the usage of rctx.getenv

aignas:
Remove most of the usage of os.environ.get and use getenv instead.
The question is if we should just replace all usages blindly?


aignas:
I checked the PR again and it seems that only in one place we have `PATH`, but that was the case previously, so the behaviour is not changed. For now it is good enough.

PR #3644, opened 2026-03-04
fix(bootstrap) handle when the runfiles env vars are not correct

### LINKED ISSUE #3518: RUNFILES_DIR inheritance causes issues when non-Python process invokes py_binary
# 🐞 bug report

### Affected Rule

The issue is caused by the rule: `py_console_script_binary` (and likely `py_binary` in general) when used as a tool invoked by another process that has inherited `RUNFILES_DIR` from a different `py_binary`.

The specific file affected is `_bazel_site_init.py` generated by the bootstrap, specifically the `_find_runfiles_root()` function.

### Is this a regression?

Yes, the previous version in which this bug was not present was: **1.6.3**

This appears to be a variant of [the issue that was fixed in 1.7.0](https://github.com/bazel-contrib/rules_python/issues/3187) (*"The stage1 bootstrap script now correctly handles nested RUNFILES_DIR environments, fixing issues where a py_binary calls another py_binary"*), but the fix doesn't cover the case where a **non-Python process** sits between the two Python binaries.

### Description

When using multiple `py_binary` tools in a Bazel `genrule`, where one tool (e.g., `grpcio-tools`) invokes a non-Python binary (e.g., `protoc`) that then invokes another `py_binary` as a plugin (e.g., `protoc-gen-mypy`), the second Python binary fails with `ModuleNotFoundError` because it inherits the wrong `RUNFILES_DIR` from the first Python binary.

**The chain of execution is:**
1. `grpcio-tools` (py_binary) → sets `RUNFILES_DIR` to its own runfiles
2. `grpcio-tools` invokes `protoc` (C++ binary) → inherits `RUNFILES_DIR` unchanged
3. `protoc` invokes `protoc-gen-mypy` (py_console_script_binary) as a plugin → inherits wrong `RUNFILES_DIR`
4. `protoc-gen-mypy` bootstrap sees `RUNFILES_DIR` pointing to `grpcio-tools.runfiles`, correctly determines it doesn't contain its own files
5. Fallback to `__file__`-based path calculation fails because `__file__` resolves to the **symlink target** (outside `.runfiles/`) rather than the symlink path (inside `.runfiles/`)
6. `sys.path` gets populated with non-existent paths → `ModuleNotFoundError`

## 🔬 Minimal Reproduction

```python
# BUILD.bazel
load("@rules_python//python/entry_points:py_console_script_binary.bzl", "py_console_script_binary")

py_console_script_binary(
    name = "protoc-gen-mypy",
    pkg = "@pypi//:mypy-protobuf",
)

py_binary(
    name = "grpcio-tools",
    srcs = ["grpcio_tools.py"],
    deps = ["@pypi//:grpcio-tools"],
)

genrule(
    name = "generate_protos",
    srcs = ["example.proto"],
    outs = ["example_pb2.py", "example_pb2.pyi"],
    cmd = """
        $(location :grpcio-tools) \
            --plugin=protoc-gen-mypy=$(location :protoc-gen-mypy) \
            --mypy_out=$(GENDIR) \
            $(location example.proto)
    """,
    tools = [
        ":grpcio-tools",
        ":protoc-gen-mypy",
    ],
)
```

The key aspect is that `grpcio-tools` invokes `protoc` (C++ binary), which then invokes `protoc-gen-mypy` as a subprocess plugin. The non-Python `protoc` binary passes through the environment unchanged.

## 🔥 Exception or Error

<pre><code>
ModuleNotFoundError: No module named 'mypy_protobuf'
--mypy_out: protoc-gen-mypy: Plugin failed with status code 1.
</code></pre>

With `RULES_PYTHON_BOOTSTRAP_VERBOSE=1`, the root cause is visible:

<pre><code>
bootstrap: stage 2: initial environ: RUNFILES_DIR='.../grpcio-tools.runfiles'

bazel_site_init: runfiles_root: .../bin/tool/src/protos
# ^^^ WRONG! Should be .../protoc-gen-mypy.runfiles

bazel_site_init: append sys.path: .../bin/tool/src/protos/sfc_rules_python++pip+pypi/.../mypy-protobuf@3.6.0/site-packages
# ^^^ This path doesn't exist - the actual files are in protoc-gen-mypy.runfiles/...
</code></pre>

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
Linux 5.10.238 (aarch64)
  </code>
</pre>

<pre>
  <code>
Build label: 8.2.1-sf5
  </code>
</pre>

**Rules_python version:**

<pre>
  <code>
1.7.0
  </code>
</pre>

**Anything else relevant?**

**Root Cause:** The `_find_runfiles_root()` function in `_bazel_site_init.py` has a fallback that calculates the runfiles root from `__file__`:

```python
num_dirs_to_runfiles_root = _SELF_RUNFILES_RELATIVE_PATH.count("/") + 1
runfiles_root = os.path.dirname(__file__)
for _ in range(num_dirs_to_runfiles_root):
    runfiles_root = os.path.dirname(runfiles_root)
```

However, when Python imports a module through a symlink, `__file__` is set to the **symlink target**, not the symlink path. The symlink target is:
```
.../bin/tool/src/protos/_protoc-gen-mypy.venv/.../site-packages/_bazel_site_init.py
```

But the symlink path (where the file is actually accessed from) is:
```
.../protoc-gen-mypy.runfiles/_main/tool/src/protos/_protoc-gen-mypy.venv/.../site-packages/_bazel_site_init.py
```

The symlink target doesn't have `.runfiles/_main/` in its path, so the directory traversal lands in the wrong location.

**Workaround:** Create wrapper scripts that unset `RUNFILES_DIR` before invoking the plugins:

```bash
cat > wrapper/protoc-gen-mypy << 'EOF'
#!/bin/bash
unset RUNFILES_DIR
exec "$ACTUAL_BINARY" "$@"
EOF
```

This forces the bootstrap to use the `__file__`-based fallback without the incorrect `RUNFILES_DIR` validation failing first... but then the symlink issue means we'd still have a problem. Actually, unsetting `RUNFILES_DIR` works because when `RUNFILES_DIR` is unset, the bootstrap uses a different code path that correctly finds the runfiles via `RUNFILES_MANIFEST_FILE` or other means.

I want to disclaim that I've used AI to find this issue and the possible solution, as I'm not too familiar with Bazel yet. However; its solution seems to have fixed our issue locally and the explanation seems to match the other linked issue and its accompanying PR.

adhoc-bobcat:
This change addresses an issue that has existed since 1.7.0 where Python binaries launched as subprocesses could incorrectly inherit and use runfiles environment variables (`RUNFILES_DIR`, `RUNFILES_MANIFEST_FILE`) from the parent process.

## Why this change is needed:

When a Python binary spawns another Python binary, the child process inherits the environment variables. If the parent had runfiles-related environment variables set, the child would attempt to use the parent's runfiles tree, which is incorrect and leads to import errors if the child has different dependencies or a different runfiles layout.

## Behavior Before:

A child Python process would trust and use any existing `RUNFILES_DIR` and `RUNFILES_MANIFEST_FILE` environment variables. If these were set by a parent Python process, they would point to the parent's runfiles, causing the child to fail when trying to load its own resources or dependencies.

## Behavior After:

The Python bootstrap scripts now include a check to validate the inherited runfiles environment variables. If the runfiles variables exist but do not point to the correct location for the current binary, the bootstrap script will unset both `RUNFILES_DIR` and `RUNFILES_MANIFEST_FILE` from the environment. This allows the subsequent fallback logic in the bootstrap process to correctly locate the runfiles for the current process, ensuring dependencies are resolved properly. This modification has been applied to both bootstrap template files.

This change ensures that even nested Python binary calls correctly find their respective runfiles.

Fixes #3518

### DIFF
--- a/tests/bootstrap_impls/bin_calls_bin/BUILD.bazel
+++ b/tests/bootstrap_impls/bin_calls_bin/BUILD.bazel
@@ -1,7 +1,7 @@
 load("@rules_shell//shell:sh_test.bzl", "sh_test")
+load("//python:py_library.bzl", "py_library")
 load("//tests/support:py_reconfig.bzl", "py_reconfig_binary")
 load("//tests/support:support.bzl", "NOT_WINDOWS", "SUPPORTS_BOOTSTRAP_SCRIPT")
-load("//python:py_library.bzl", "py_library")
 
 py_library(
     name = "inner_lib",
@@ -23,10 +23,10 @@ py_reconfig_binary(
 py_reconfig_binary(
     name = "inner_bootstrap_system_python",
     srcs = ["inner.py"],
-    deps = [":inner_lib"],
     bootstrap_impl = "system_python",
     main = "inner.py",
     tags = ["manual"],
+    deps = [":inner_lib"],
 )
 
 genrule(
@@ -59,10 +59,10 @@ sh_test(
 py_reconfig_binary(
     name = "inner_bootstrap_script",
     srcs = ["inner.py"],
-    deps = [":inner_lib"],
     bootstrap_impl = "script",
     main = "inner.py",
     tags = ["manual"],
+    deps = [":inner_lib"],
 )
 
 py_reconfig_binary(


aignas (approved):
LGTM, but I'd like for @rickeylev  to also take a look. I vaguely remember something similar in the past and am wondering if clearing the variables is a red herring or not.

That said, my thinking is:
* CI is green.
* You have added a test for the usecase - thank you!
* If this breaks someone's workflow and they come along, they can add one more test and fix it.

Could you please add a CHANGELOG.md item to add a note about what is fixed please?

adhoc-bobcat:
> Could you please add a CHANGELOG.md item to add a note about what is fixed please?

Done.

aignas (approved):

PR #3647, opened 2026-03-05
ci: disable soft-fail jobs to unbreak ci

rickeylev:
BazelCI has a bug where soft_fail jobs will break the entire CI pipeline.

Commenting them out seems to avoid the bug

See also: https://github.com/bazelbuild/continuous-integration/pull/2487


rickeylev:
Looks like bazel ci fixed this; closing

PR #3650, opened 2026-03-06
tests: add bazel 9 testing, use latest released (not rc) for bazel

rickeylev:
This runs the default test jobs using the test matrix so that Bazel 9 is better
covered by testing.

The main thing needed is setting `--incompatible_strict_action_env=false`. This is
enabled in Bazel 9, but breaks Windows. See
https://github.com/bazel-contrib/rules_python/issues/3655 for details.

It also changes to using `N.x` instead of `N.*` for specifying the Bazel version.
The difference is `x` matches the latest released version, while `*` matches
the latest released _or release candidate_. Since release candidates can have
regressions, and we have a separate job for RCs, use the latest released version.

Along the way, mark the last_rc job as soft-fail

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -198,6 +198,7 @@ tasks:
     platform: ubuntu2204
     bazel: last_rc
     # This is an advisory job; doesn't block merges
+    # RCs may have regressions, so don't fail our CI on them
     soft_fail:
       - exit_status: 1
       - exit_status: 3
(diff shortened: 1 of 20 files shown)


rickeylev:
Ready for review.

Main change is having to disable strict action env. https://github.com/bazel-contrib/rules_python/issues/3655 details why, but the short version is we'll have to get off of the Bazel launcher to make windows Just Work without extra global flags set

PR #3652, opened 2026-03-07
refactor(pypi): move absolute_url to whl_library

aignas:
With this PR we move the processing of the `index_url` to the
`whl_library` as a preparatory step for easier `facts` implementation.
The motivation is many-fold:
1. Do not have too much duplication in the facts file by potentially
   naturally eliminating the `index_url` prefix from the `whls` if it
   appears like so on the index contents.
2. Avoid doing `envsubst` too early and have logic that has to deal with
   it.
3. Make the cache just return fact values from the lock file in the
   future instead of needing to change to an absolute URL and do
   envsubst on it.
4. We should have a better performance because we should be doing way
   fewer calls to make the URL absolute during parsing of the index.
5. With the `index_url` passed to the `whl_library`, we can help out the
   `purl` construction as what has been discussed in #3531 about wheels
   from non-public indexes.

Summary:
- Attempt to put the `index_url` in the fewest structs possible.
- Extract the `urllib` utilities file for manipulation of the URLs.
- Simplify tests testing the `absolute_url` logic.

Work towards #2731

### DIFF
--- a/tests/pypi/urllib/urllib_tests.bzl
+++ b/tests/pypi/urllib/urllib_tests.bzl
@@ -19,6 +19,14 @@ def _test_absolute_url(env):
     env.expect.that_str(urllib.absolute_url("https://example.com//a/b//", "../../file.whl")).equals("https://example.com/file.whl")
     env.expect.that_str(urllib.absolute_url("https://example.com//a/b//", "/file.whl")).equals("https://example.com/file.whl")
 
+    # Relative URLs
+    env.expect.that_str(urllib.absolute_url("https://example.com/relative", "file.whl")).equals("https://example.com/relative/file.whl")
+    env.expect.that_str(urllib.absolute_url("https://example.com/relative/", "file.whl")).equals("https://example.com/relative/file.whl")
+    env.expect.that_str(urllib.absolute_url("https://example.com/relative/", "../relative/file.whl")).equals("https://example.com/relative/file.whl")
+
+    # Relative URL for files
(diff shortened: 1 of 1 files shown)


rickeylev:
I saw you thumbed up a couple gemini comments, so i'm assuming you wanted to address those, so I'll let you queue for merge when you're ready.

tests/pypi/urllib/urllib_tests.bzl
```diff
@@ -19,6 +19,14 @@ def _test_absolute_url(env):
     env.expect.that_str(urllib.absolute_url("https://example.com//a/b//", "../../file.whl")).equals("https://example.com/file.whl")
     env.expect.that_str(urllib.absolute_url("https://example.com//a/b//", "/file.whl")).equals("https://example.com/file.whl")
 
+    # Relative URLs
+    env.expect.that_str(urllib.absolute_url("https://example.com/relative", "file.whl")).equals("https://example.com/relative/file.whl")
+    env.expect.that_str(urllib.absolute_url("https://example.com/relative/", "file.whl")).equals("https://example.com/relative/file.whl")
+    env.expect.that_str(urllib.absolute_url("https://example.com/relative/", "../relative/file.whl")).equals("https://example.com/relative/file.whl")
+
+    # Relative URL for files
+    env.expect.that_str(urllib.absolute_url("file://{PYPI_BAZEL_WORKSPACE_ROOT}", "vendor/distro/file.whl")).equals("file://{PYPI_BAZEL_WORKSPACE_ROOT}/vendor/distro/file.whl")
```
aignas, line 28: This potentially would be a really good thing for #1689. Once we have full `facts` support, we can just provide user with ways to load wheel files from anywhere.

aignas: Related PR that I actually wanted to link: https://github.com/bazel-contrib/rules_python/pull/3312

PR #3653, opened 2026-03-07
fix: Fix zipapp compression support for new `py_zipapp_binary` target

nate-england:
The compression level isn't being correctly passed by the zipapp rule to the zipper
program, resulting in an error if custom compression is specified.

To fix, pass the arg correctly to the zipper.

`venv_zipapp_test.py` is updated to also confirm zipapps are properly compressed when configured in the BUILD target.

Fixes https://github.com/bazel-contrib/rules_python/issues/3646

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,7 +63,8 @@ END_UNRELEASED_TEMPLATE
 
 {#v0-0-0-fixed}
 ### Fixed
-* Nothing fixed.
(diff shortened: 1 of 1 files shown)


aignas:

This LGTM. Could you please add a line in the CHANGELOG as well to indicate that an issue got fixed similar to how we used to do this?

nate-england:
Done!

rickeylev:
Thanks for the fix!

PR #3654, opened 2026-03-08
feat(pypi): store PyPI results as facts v2

### LINKED ISSUE #2731: Write pip extension metadata to the MODULE.bazel.lock file

aignas:
This PR adds functionality needed to write data that we find useful on the
SimpleAPI responses to the lock file. I.e. this will no longer connect to
the network if it can find the necessary information in the lock file.

Superseeds #3559
Fixes #2731

### DIFF
--- a/python/private/pypi/pypi_cache.bzl
+++ b/python/private/pypi/pypi_cache.bzl
@@ -10,23 +10,26 @@ In the future the same will be used to:
 
 load(":version_from_filename.bzl", "version_from_filename")
 
+# This value should be changed whenever the storage format changes.
+# Changing it simply means the information cached in the lockfile has to be
+# recomputed.
 _FACT_VERSION = "v1"
 
-def pypi_cache(module_ctx = None, store = None):
+def pypi_cache(mctx = None, store = None):
     """The cache for PyPI index queries.
 
     Currently the key is of the following structure:
     (url, real_url, versions)
 
     Args:
-        module_ctx: The module context
+        mctx: The module context
         store: The in-memory store, should implement dict interface for get and setdefault
 
     Returns:
         A cache struct
     """
     mcache = memory_cache(store)
-    fcache = facts_cache(getattr(module_ctx, "facts", None))
+    fcache = facts_cache(getattr(mctx, "facts", None))
 
     # buildifier: disable=uninitialized
     self = struct(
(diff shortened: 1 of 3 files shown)


python/private/pypi/parse_simpleapi_html.bzl
```diff
@@ -53,16 +53,37 @@ def parse_simpleapi_html(*, content):
 
     # Each line follows the following pattern
     # <a href="https://...#sha256=..." attribute1="foo" ... attributeN="bar">filename</a><br />
+    #
+    # Sometimes the lines may be split, so we should seek until `<br />`
```
aignas: What I meant here is that yank reason sometimes spans multiple lines. It would be best to build a small tokenizing parser here.

python/private/pypi/pypi_cache.bzl
```diff
@@ -52,4 +74,163 @@ def _pypi_cache_get(self, key):
...
+
+    root_url, _, distribution = index_url.rstrip("/").rpartition("/")
+    distribution = distribution.rstrip("/")
+    root_url = root_url.rstrip("/")
+
+    for sha256, d in (value.sdists | value.whls).items():
+        facts.setdefault("dist_hashes", {}).setdefault(root_url, {}).setdefault(distribution, {}).setdefault(d.url, sha256)
+        if not d.url.endswith(d.filename):
+            facts.setdefault("dist_filenames", {}).setdefault(root_url, {}).setdefault(distribution, {}).setdefault(d.url, d.filename)
+        if d.yanked:
```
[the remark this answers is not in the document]
aignas:
```suggestion
        if not d.yanked:
```

aignas:
OK, this is done, PTAL.

rickeylev:
Preserving part of the original description
----

Improvements that could be done but I don't have time right now for (but
contributions are welcome):
* Improve the code to use `sha256` in addition to `version` to do the filtering
  what we need to pull from the lock file. Whilst theoretically this could improve
  how quickly we find things, not sure how better it would be in practise.
* Pull the METADATA contents from the PyPI SimpleAPI if the `metadata_url` is there
  and write the `Requires-Dist` to the `MODULE.bazel.lock` file as facts.
* Shorten the package URLs to remove common prefix.

python/private/pypi/pypi_cache.bzl
```diff
@@ -52,4 +74,204 @@ def _pypi_cache_get(self, key):
...
+            k: sorted(v)
+            for k, v in sha256s_by_version.items()
+        },
+    )
+
+def facts_cache(known_facts, facts_version = _FACT_VERSION):
+    """The facts cache.
+
+    Here we have a way to store things as facts and the main thing to keep in mind is that we should
+    not use the real_url in case it contains credentials in it (e.g. is of form `https://<username>:<password>@<host>`).
```
rickeylev, line 161: Containing the username would probably be OK. But yeah, not the password.

Should we strip out the username/password from the cache key? Normally that info wouldn't be present in the url if they were passed via headers.

PR #3657, opened 2026-03-10
feat(pypi): first check index contents before downloading metadata about distributions

### LINKED ISSUE #2632: pypi downloads blocked by experimental_downloader_config fail silently
# 🐞 bug report

### Affected Rule

`pip.parse` with the bazel downloader enabled

### Is this a regression?

not sure

### Description

with some config like this:

```bzl
pip.parse(
    experimental_extra_index_urls = [
        "https://download.pytorch.org/whl/cu121",
    ],
    experimental_index_url = "https://pypi.org/simple",
    experimental_index_url_overrides = {
        "torch": "https://download.pytorch.org/whl/cu121",
        "torchaudio": "https://download.pytorch.org/whl/cu121",
        "torchvision": "https://download.pytorch.org/whl/cu121",
    },
...
)
```

If your remote downloader config doesn't allow pytorch.org, rules_python can fail with this error:

```
Error in fail: Failed to download metadata for ["torch", "torchaudio", "torchvision"] for from urls: ["https://pypi.org/simple", "https://download.pytorch.org/whl/rocm6.2"]
```

Instead of the normal:

```
Error in download: java.io.IOException: Configured URL rewriter blocked all URLs: [https://download.pytorch.org/whl/rocm6.2/torch/]
```

I believe this is because of `allow_fail = True` https://github.com/bazelbuild/rules_python/blob/fcf7221c1e079307ff13d32239b7782d2f1dc48c/python/private/pypi/simpleapi_download.bzl#L208 and it makes it hard to spot that it's a misconfiguration on your downloader config.

## 🌍  Your Environment

**Operating System:**

linux aarch64

95bbed03758a7651916d0708a892f5342a0421c5

**Rules_python version:**

https://github.com/bazelbuild/rules_python/commit/fcf7221c1e079307ff13d32239b7782d2f1dc48c

### LINKED ISSUE #3260: SimpleAPI download failures are cached and require bazel clean
# 🐞 bug report

### Affected Rule

`pip.parse()`

### Is this a regression?

Probably not.

### Description

We have a downloader config that rewrites URLs to a private Artifactory instance that's only accessible via our VPN. If I try to build without the VPN, I get an expected failure:
```
Error in fail: Failed to download metadata for ["nh3", "idna", "rich", "zipp", "mdurl", "twine", "certifi", "keyring", "pkginfo", "rfc3986", "urllib3", "docutils", "pygments", "requests", "jaraco-classes", "jaraco-context", "markdown-it-py", "more-itertools", "readme-renderer", "jaraco-functools", "backports-tarfile", "requests-toolbelt", "charset-normalizer", "importlib-metadata", "cffi", "jeepney", "pycparser", "cryptography", "secretstorage", "pywin32-ctypes"] for from urls: ["https://pypi.org/simple"].
If you would like to skip downloading metadata for these packages please add 'simpleapi_skip=[
    "nh3",
    "idna",
    "rich",
    "zipp",
    "mdurl",
    "twine",
    "certifi",
    "keyring",
    "pkginfo",
    "rfc3986",
    "urllib3",
    "docutils",
    "pygments",
    "requests",
    "jaraco-classes",
    "jaraco-context",
    "markdown-it-py",
    "more-itertools",
    "readme-renderer",
    "jaraco-functools",
    "backports-tarfile",
    "requests-toolbelt",
    "charset-normalizer",
    "importlib-metadata",
    "cffi",
    "jeepney",
    "pycparser",
    "cryptography",
    "secretstorage",
    "pywin32-ctypes",
]' to your 'pip.parse' call.
```

However, even after turning on the VPN, I still get this error until I do `bazel clean`. It seems like failed downloads are cached, and shouldn't be?

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>

  </code>
</pre>

<pre>
  <code>
Bazelisk version: 1.25.0
Build label: 8.2.1
Build time: Thu Apr 17 18:31:45 2025 (1744914705)
Build timestamp: 1744914705
Build timestamp as int: 1744914705
  </code>
</pre>

**Rules_python version:**

<pre>
  <code>
1.4.1
  </code>
</pre>

**Anything else relevant?**

aignas:
The overall changes to the architecture are:
* First check which packages are on which index.
* Then write these details as facts for future reuse.
* Then use the `index_url_overrides` to ensure that we are pulling things from the right index for the packages.
* Then download everything.

Notes on implementation:
* This will pull index contents at most once for each index.
* This will make the initial download times longer, but because we have `MODULE.bazel.lock` file and the facts written there, it should be OK.
* This allows us to just parse the `index_url` and `extra_index_urls` from the lock files and use that without printing any warning messages. If we don't see any regressions in testing, I think this code path could be enabled for everyone by default. So `experimental_index_url` will no longer be experimental. I think this might have been the last thing holding us from flipping the switch.

Fixes #2632
Fixes #3260

### DIFF
--- a/python/private/pypi/parse_simpleapi_html.bzl
+++ b/python/private/pypi/parse_simpleapi_html.bzl
@@ -72,11 +72,16 @@ def parse_simpleapi_html(*, content, parse_index = False):
         if start_tag == -1:
             break
 
-        # Find the end of the opening tag and the closing </a>
-        tag_end = content.find(">", start_tag)
-        end_tag = content.find("</a>", tag_end)
-        if tag_end == -1 or end_tag == -1:
+        # Find the closing </a> tag first, then find the end of the opening
+        # <a ...> tag using rfind. This correctly handles attributes that
+        # contain > characters, e.g. data-requires-python=">=3.6".
+        end_tag = content.find("</a>", start_tag)
+        if end_tag == -1:
             break
+        tag_end = content.rfind(">", start_tag, end_tag)
+        if tag_end == -1 or tag_end <= start_tag:
+            cursor = end_tag + 4
+            continue
 
         # Extract only the necessary slices
         filename = content[tag_end + 1:end_tag].strip()
--- a/tests/pypi/parse_simpleapi_html/parse_simpleapi_html_tests.bzl
+++ b/tests/pypi/parse_simpleapi_html/parse_simpleapi_html_tests.bzl
@@ -88,7 +88,7 @@ def _test_sdist(env):
             struct(
                 attrs = [
                     'href="https://example.org/full-url/foo-0.0.1.tar.gz#sha256=deadbeefasource"',
-                    'data-requires-python="&gt;=3.7"',
+                    'data-requires-python=">=3.7"',
                     "data-yanked",
                 ],
                 filename = "foo-0.0.1.tar.gz",
@@ -105,7 +105,7 @@ def _test_sdist(env):
             struct(
                 attrs = [
                     'href="https://example.org/full-url/foo-0.0.1.tar.gz#sha256=deadbeefasource"',
-                    'data-requires-python="&gt;=3.7"',
+                    'data-requires-python="<=3.7"',
                     "data-yanked=\"Something &#10;with &quot;quotes&quot;&#10;over two lines\"",
                 ],
                 filename = "foo-0.0.1.tar.gz",


python/private/pypi/pypi_cache.bzl
```diff
@@ -89,6 +89,9 @@ def _pypi_cache_get(self, key):
     if not cached and versions:
         # Could not get from in-memory, read from lockfile facts
         cached = self._facts.get(index_url, versions)
+    else:
+        # TODO @aignas 2026-03-20: add a test here
```
aignas: TODO: split into a separate PR, because it is a bug in the current implementation.

python/private/pypi/parse_simpleapi_html.bzl
```diff
@@ -73,18 +79,23 @@ def parse_simpleapi_html(*, content):
...
 
-        # 3. Efficient Attribute Parsing
         attrs = _parse_attrs(attr_part)
         href = attrs.get("href", "")
         if not href:
             continue
 
+        if parse_index:
+            pkg_name = filename
+            packages[normalize_name(pkg_name)] = href
```
aignas, line 100: Check if we have to `normalize` all of the package names here. It might be a little bit of throw away work that we are doing here.

python/private/pypi/simpleapi_download.bzl
```diff
@@ -71,110 +75,116 @@ def simpleapi_download(
...
-                normalize_name(pkg),
-                index_urls,
-            )
-            for pkg in failed_sources
+            pkg: index_url_overrides.get(pkg, index_urls)
+            for pkg in sources
         }
 
+        # TODO @aignas 2026-03-20: we haven't found these pkgs on the index, so we can
+        # print a warning, or we can fallback to PyPI. For now let's fail
```
aignas: Probably remove the fail here.

python/private/pypi/parse_simpleapi_html.bzl
aignas: We do this here because then it is easy to cache the results and the PyPI response is massive.

aignas:
OK, done - PTAL.

aignas:
OK, PTAL, I am changing the behaviour in a way that will make it easier to default `index_url` to `pypi.org` and parse/override params from the lock files as is asked in #2951. In the next PR I plan to just flip the switch by: defaulting the `index-url` internally to `pypi.org` and then parsing the requirements files to get the values. I think it is going to be a good thing for 2.0 and if it is breaking, it should be minimally so. I think we should also rename the `experimental_index_url_overrides` to `index_url_overrides`.

Given that there are usages of `experimental_index_url` in the bazel-central-registry, I think it is best to just make the parameter a noop. All of the values are actually using `pypi.org/simple`.

PR #3661, opened 2026-03-12
fix: prevent stdlib pyc files from invalidating runtime repos

rickeylev:
The runtime repositories are being constantly invalidated due to pyc creation under Bazel 9
because, starting in Bazel 9, `glob()` functions implicitly register `repository_ctx.watch()`
calls on the files and directories they match. Thus, the directories where `__pycache__`
directories are created end up being considered changed (either directly because their
mtimes change, or indirectly, because their directory listing changes), which then invalidates
the repo, causing it to re-run.

This glob-induced-watching seems to occur even if an `exclude` would have excluded the file.

Note that this only seems to occur if `reproducible=False`, which generally wouldn't occur,
but could occur if a user is registering their own runtime and doesn't care about the sha.
Regardless, this still seems worthwhile because it allows pyc to be more safely be
generated without causing repo invalidations, while allowing them to be persisted between
repo-phase invocations.

To fix, create `__pycache__` directories ahead of time and symlink them to a location that Bazel
isn't watching, i.e. outside the repository's directory. I tried creating a separate top-level
folder that wasn't matched by any globs and symlinking to it, but Bazel would read through
the symlinks and watch the underlying locations.

This also has a side-bonus that allows pyc files to be re-used in between
repository-phase invocations.

Fixes https://github.com/bazel-contrib/rules_python/issues/3643

### DIFF
--- a/python/private/python_repository.bzl
+++ b/python/private/python_repository.bzl
@@ -69,7 +69,9 @@ def _get_pycache_root(rctx):
         return repo_utils.mkdir(rctx, res)
 
     # Suffix for cases 2-4
-    suffix = "rules_python_{}/{}".format(hash(str(rctx.workspace_root)), rctx.name)
+    # The first level directory is static and documented so that it is easy to
+    # use with e.g. --sandbox_add_mount_pair=/tmp/rules_python_pycache
+    suffix = "rules_python_pycache/{}/{}".format(hash(str(rctx.workspace_root)), rctx.name)
 
     # 2. XDG_CACHE_HOME
     res = rctx.getenv("XDG_CACHE_HOME")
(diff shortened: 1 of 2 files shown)


python/private/python_repository.bzl
```diff
@@ -52,6 +52,95 @@ def is_standalone_interpreter(rctx, python_interpreter_path, *, logger = None):
...
+    """Calculates and creates the pycache root directory.
+
+    Returns:
+        {type}`path | None` The path to the pycache root, or None if it couldn't
+        be created.
+    """
+    os_name = repo_utils.get_platforms_os_name(rctx)
+    is_windows = os_name == "windows"
+
+    # 1. RULES_PYTHON_PYCACHE_DIR
```
aignas, line 65: Please document the env vars that we use here somewhere plus don't forget the changelog notes. Might be great to mention.

rickeylev: Done

PR #3662, opened 2026-03-12
chore: use bazel 9 by default

rickeylev:
Updates the project bazel version and bzlmod examples to use Bazel 9

Along the way, upgrade rules_go to 0.60.0; this is a dev-only dependency. This is needed
because Bazel 9 removed CcInfo, and earlier rules_go versions refer to the builtin symbol.

Also remove the test to check the bazel version. It doesn't do much useful.


aignas (approved):

rickeylev:
I've gone and remove the assert_bazelversion test. I don't see the value in test that validates the .bazelversion file against the currently running bazel. We have a bunch of cases where they wouldn't be the same (I.e. two thirds of our CI jobs)

PR #3666, opened 2026-03-15
fix(pypi): propagate fails if overrides are passed only one index is used

aignas:
Before this PR there would be confusing failures when downloader config
is set to disallow certain values or when the authentication is not
setup properly. This is a small fix towards a better goal state where
we set `allow_fail = False` in cases where we know that we have to
succeed to download metadata from that particular URL.

The use-cases covered:
- Only one index_url is passed to `pip.parse`.
- `index_url_overrides` are passed which means that we should fail if
  there are insufficient overrides.

The downside to this is that it is really hard to return custom error
messages telling the user what to do, but on the flip side, the failures
coming from bazel itself might be more descriptive in the case of
outh-misconfiguration or bazel downloader configuration settings.

Work towards #2632 and #3260.

### DIFF
--- a/tests/pypi/simpleapi_download/simpleapi_download_tests.bzl
+++ b/tests/pypi/simpleapi_download/simpleapi_download_tests.bzl
@@ -24,11 +24,7 @@ def _test_simple(env):
     calls = []
 
     def read_simpleapi(ctx, url, versions, attr, cache, get_auth, block, allow_fail):
-        _ = ctx  # buildifier: disable=unused-variable
-        _ = attr
-        _ = cache
-        _ = get_auth
-        _ = versions
+        _ = ctx, attr, cache, get_auth, versions  # buildifier: disable=unused-variable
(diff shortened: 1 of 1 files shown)


tests/pypi/simpleapi_download/simpleapi_download_tests.bzl
```diff
@@ -159,15 +159,86 @@ If you would like to skip downloading metadata for these packages please add 'si
...
     ])
 
 _tests.append(_test_fail)
 
+def _test_allow_fail_single_index(env):
+    calls = []
+    fails = []
+
+    def read_simpleapi(ctx, *, url, versions, attr, cache, get_auth, block, allow_fail):
+        _ = ctx  # buildifier: disable=unused-variable
```
rickeylev: Nit: assign them all to underscore in a single line. Just saves a bit of vertical space

PR #3691, opened 2026-04-11
feat!(pypi): enable bazel downloader by default

### LINKED ISSUE #260: "Cross compilation" of py_binary/py_image/py_library targets

### LINKED ISSUE #1357: Use bazel's downloader for downloading pip packages

### LINKED ISSUE #2241: Third party dependencies are incorrect when using RBE because host != exec

### LINKED ISSUE #2951: Read index urls from the requirements files and instead have a flag

aignas:
Summary:
- change: Set `https://pypi.org/simple` as the default index.
- refactor: Leave the code for the legacy behaviour intact in case we
  need to do more work on the configuration to exclude the cases where
  the downloader should be used.
- add: add `index_url` configuration option for setting the defaults.
- add: add a small utility for parsing the arguments.
- fix: downloader will not be used if there is no url associated with
  the source irrespective of what the index_url setting is.
- fix: ensure all of the URLs are normalized when used.

Fixes #260
Fixes #1357
Fixes #2241
Fixes #2951


aignas:
@dougthor42, FYI, this may be interesting to test, but it can wait until rc0 is released.

rickeylev (approved):

rickeylev:
is this a breaking change? I see some of the `experiemtnal_xxx` args are doc'd as no-ops, but it looked like the code used them as a fallback.

If so, that's fine. Lets just document it and add `!` to the summary line.

PR #3694, opened 2026-04-11
feat(runfiles): create pathlib api for runfiles library

### LINKED ISSUE #3296: runfiles: add pathlib.Path or importlib.Traversable API
The current runfiles library accepts runfile strings and returns file system strings.

This works fine for looking up specific resources, but can break when trying to list resources in a directory. The particular case is when a runfiles manifest is used without runfiles being materialized. In such a case, two problems occur: (1) there is no actual directory, and (2) files logically part of a runfiles directory may be in different underlying file system directories.

To fix this, we should have an API that returns an object that represents a path instead of the underlying filesystem path. The two candidates in the stdlib are pathlib.Path and importlib.Traversable.  Traversable has a subset of Path methods.

IMHO, I think Path is slightly better -- Traversable is specific to the import system, while runfiles aren't.

Related slack discussion: https://bazelbuild.slack.com/archives/CDCMRLS23/p1758750938143029

rickeylev:
This adds a pathlib.Path-based API for interacting with runfiles. This
makes it easier to interact with runfiles.

Fixes https://github.com/bazel-contrib/rules_python/issues/3296

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,14 @@ BEGIN_UNRELEASED_TEMPLATE
 END_UNRELEASED_TEMPLATE
 -->
 
+## Unreleased
+
+### Added
+* (runfiles) Added a {obj}`Path` class that implements the `pathlib.PurePath` API
+  for idiomatic path manipulation of runfiles. It is created by calling the
+  {obj}`Runfiles.root` instance method. The {obj}`Path` class also has a
+  {obj}`Path.runfiles_root` method to return the root of the runfiles tree.
+
 {#v2-0-0}
 ## [2.0.0] - 2026-04-09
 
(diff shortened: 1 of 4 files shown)


tests/runfiles/pathlib_test.py
```diff
@@ -0,0 +1,105 @@
...
+class PathlibTest(unittest.TestCase):
+    def setUp(self) -> None:
+        self.tmpdir = tempfile.TemporaryDirectory(dir=os.environ.get("TEST_TMPDIR"))
+        # Runfiles paths are expected to be posix paths internally when we construct the strings for assertions
+        self.root_dir = pathlib.Path(self.tmpdir.name).as_posix()
+
+    def tearDown(self) -> None:
+        self.tmpdir.cleanup()
+
+    def testPathApi(self) -> None:
```
aignas: nit, maybe it is personal preference but itwould be nice to have the tests named in snake case here except for the special methods.

CHANGELOG.md
```diff
@@ -47,6 +47,15 @@ BEGIN_UNRELEASED_TEMPLATE
 END_UNRELEASED_TEMPLATE
 -->
 
+## Unreleased
```
aignas, line 51: Please use the template :)

aignas (approved):

PR #3696, opened 2026-04-12
chore: add some skills and update agents doc

rickeylev:
Some misc instructions for agents to follow.

* Tell them about the multiple modules. I've found they can get
  confused trying to run code in different sub-modules.
* Tell them to never run expunge. This hard-wipes caches and trashes
  build performance, not to mention takes many minutes itself
* Tell them to use the fast-tests config, which captures more than
  just integration tests now.

Add skills for interacting with build kite, which makes automating
PRs a bit easier.

### DIFF
--- a/.agents/skills/hello/SKILL.md
+++ b/.agents/skills/hello/SKILL.md
@@ -1,0 +1,7 @@
+---
+name: hello
+description: say hello
+---
+
+Say hello with a made up name and greeting.
+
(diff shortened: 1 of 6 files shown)


.agents/skills/hello/SKILL.md
```diff
@@ -0,0 +1,7 @@
+---
+name: hello
+description: say hello
+---
+
+Say hello with a made up name and greeting.
```
aignas: Do you want this?

aignas (approved):

PR #3701, opened 2026-04-13
chore: have no-matching-distribution error message hint about target_platforms vs requirements_by_platform

rickeylev:
Based on the slack discussion where a cross-build was failing because
`requirements_by_platform` was set, but `target_platforms` wasn't
set, have the error message hint at the distinction to better point
users for how to fix it.


python/private/pypi/pkg_aliases.bzl
aignas:
```suggestion
No matching wheel for current configuration's Python version and platform.
```

aignas (approved):

aignas (approved):

PR #3706, opened 2026-04-14
tests: better simulate BCR environment

rickeylev:
Better simulate the BCR environment by removing files that aren't
extracted when BCR extracts the release archive.


.bazelci/isolate.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
```
aignas, line 1: This is a great idea!

aignas (approved):

PR #3710, opened 2026-04-16
fix(pypi): skip index lookups when all package overrides are specified

### LINKED ISSUE #3709: simpleapi_download: root /simple/ fetch crashes on registries that don't serve PEP 503 root index (e.g. Google Artifact Registry)
# 🐞 bug report
### Affected Rule
`simpleapi_download.bzl` — specifically `_read_simpleapi` and `_get_dist_urls`
### Is this a regression?
Yes, this works in 1.7.0. The root `/simple/` fetch was introduced in #3657 (commit 900d557).
### Description
We're upgrading from rules_python 1.7.0 to 2.0.0-rc2 and hit an issue with a private PyPI registry on Google Artifact Registry (GAR).
In 2.0.0-rc2, `simpleapi_download.bzl` fetches the root `/simple/` page of each index URL to discover which packages are on which index (introduced in #3657). GAR returns 404 for the root page, it only serves per-package pages like `/simple/six/`. `ctx.download` throws `IOException` on the 404, which kills the entire pip resolution.
We confirmed this is a GAR limitation, not an auth issue. Authenticated requests also get 404 on the root while per-package pages return 200.
We were able to work around this by patching `_read_simpleapi` to pass `allow_fail=parse_index` to `ctx.download`, and added a `if not result.success: continue` guard in `_get_dist_urls`, but I'm unsure how this fares for cases with `experimental_extra_index_urls` set.
## 🔬 Minimal Reproduction
1. Set up a Google Artifact Registry Python repository
2. Configure `pip.parse` with `experimental_index_url` pointing to the GAR Simple API URL, or pass `--index-url` via `extra_pip_args`
3. Run any `bazel build` that triggers pip resolution
The root cause can be verified independently:
```bash
# Per-package page works (200)
curl -s -o /dev/null -w "%{http_code}" \
  "https://<region>-python.pkg.dev/<project>/<repo>/simple/six/" \
  -H "Authorization: Bearer $(gcloud auth print-access-token)"
# Root page does not (404)
curl -s -o /dev/null -w "%{http_code}" \
  "https://<region>-python.pkg.dev/<project>/<repo>/simple/" \
  -H "Authorization: Bearer $(gcloud auth print-access-token)"
```

## 🔥 Exception or Error
<pre><code>ERROR: .../simpleapi_download.bzl:258:39: Traceback (most recent call last):
  ...
  File ".../simpleapi_download.bzl", line 164, column 43, in _get_dist_urls
    results[index_url] = download.wait()
  File ".../simpleapi_download.bzl", line 258, column 39, in lambda
    result = download.wait(),
Error in wait: java.io.IOException: Error downloading
  [https://<region>-python.pkg.dev/<project>/<repo>/simple/]
  GET returned 404 Not Found
</code></pre>

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
  macOS 24.5.0 (darwin arm64) / Linux (CI)
  </code>
</pre>

<pre>
  <code>
8.5.1
  </code>
</pre>

**Rules_python version:**

<pre>
  <code>
2.0.0-rc2
  </code>
</pre>

**Anything else relevant?**

aignas:
When index_url_overrides is provided for all packages, we no longer need to
call the index at all. This improves performance and aligns with the expected
behavior where overrides should be sufficient.

Just to note, currently Pytorch, PyPI, Artifactory have the root index pages, whilst GAR does not.

Fixes #3709

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -81,7 +81,8 @@ Other changes:
   we will from now on fetch the lists of available packages on each index. The
   used package mappings will be written as facts to the `MODULE.bazel.lock` file
   on supported bazel versions and it should be done at most once. As a result,
-  per-package {obj}`experimental_index_url_overrides` is no longer needed . What
+  per-package {obj}`experimental_index_url_overrides` is no longer needed, but 
+  if specified, it needs to be provided for all packages not on the default index. What
   is more, the flags for `--index_url` and `--extra-index-url` now behave in the
   same way as in `uv` or `pip`, i.e. we default to `--index-url` if the package
   is not found in `--extra-index-url`.  Fixes
--- a/python/private/pypi/simpleapi_download.bzl
+++ b/python/private/pypi/simpleapi_download.bzl
@@ -43,8 +43,7 @@ def simpleapi_download(
         attr: Contains the parameters for the download. They are grouped into a
           struct for better clarity. It must have attributes:
            * index_url: str, the index, or if `extra_index_urls` are passed, the default index.
-           * index_url_overrides: dict[str, str], the index overrides for
-             separate packages.
+           * index_url_overrides: dict[str, str], the index overrides for separate packages.
            * extra_index_urls: Will be looked at in the order they are defined and the first match
                 wins. This is similar to what uv does, see
                 https://docs.astral.sh/uv/concepts/indexes/#searching-across-multiple-indexes.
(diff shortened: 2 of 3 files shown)


rickeylev (approved):

rickeylev:
Does this need to be put into the 2.0 release?

PR #3716, opened 2026-04-18
feat(runfiles): implement runfiles.Path io methods

### LINKED ISSUE #3296: runfiles: add pathlib.Path or importlib.Traversable API

rickeylev:
The runfiles.Path class is currently a PurePath subclass, which means
it only does string manipulation for constructing paths. This change
adds functionality to actually access files.

Fixes https://github.com/bazel-contrib/rules_python/issues/3296

### DIFF
--- a/python/runfiles/runfiles.py
+++ b/python/runfiles/runfiles.py
@@ -29,7 +29,7 @@ import pathlib
 import posixpath
 import sys
 from collections import defaultdict
-from typing import Dict, List, Optional, Tuple, Union
+from typing import Dict, Generator, Iterable, List, Optional, Tuple, Union
 
 
 class _RepositoryMapping:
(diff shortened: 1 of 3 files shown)


aignas:
The Self type annotation was only introduced in 3.13, this should be made compatible with lower Python versions too. :)

aignas (approved):

PR #3718, opened 2026-04-20
feat(windows): site-packages venv support

rickeylev:
This implements Windows support for library files being populated
in the venv site-packages directory (i.e. `--venv_site_packages=yes`).

This is mostly just an expansion of the work done to make venvs work
on Windows. It mostly just adds the site-packages paths to the list
of (venv_path, symlinks) values that is currently used for the
interpreter files.

The main notable change is that creation of the site-packages
directory also requires traversing the whole site-packages directory
and re-creating its structure in the temporary venv. This isn't ideal,
but there isn't much other option.

Work towards https://github.com/bazel-contrib/rules_python/issues/3245


rickeylev:
I think this is the last piece to bring venv support on windows on par
with linux.

aignas (approved):

PR #3722, opened 2026-04-21
doc: add notes on the design

aignas:
Just to start a page on design notes to tell users better what they are getting.


rickeylev (approved):
I did some heavy copy editing and rewrite. LGTM on the main topics, though.

PR #3726, opened 2026-04-23
feat(venv): support data, include, and scripts schemes

rickeylev:
Currently, the files from the data, headers, and scripts portions of a
wheel don't end up in the proper sub-directories of the venv. This
means the full files of a distribution aren't available at the typical
location in the venv, making it harder to integrate with standard
tools.

To fix, simply map the directories to paths in the venv and give them
similar treatment as site-packages.

The spec says certain first-level directories of the `.data` directory
map to specific scheme paths, which `whl_library` already handles.
Here's a listing of the
`wheel data directory -> install scheme path key -> whl_library directory`
relationships

* purelib -> purelib -> site-packages
* platlib -> platlib -> site-packages
* headers -> include -> include
* scripts -> scripts -> bin
* data -> data -> data

Relevant reading:
* Packaging specification:
  https://packaging.python.org/en/latest/specifications/binary-distribution-format
* Posix install scheme paths:
  https://docs.python.org/3/library/sysconfig.html#posix-prefix
* Windows install scheme paths:
  https://docs.python.org/3/library/sysconfig.html#nt

The whl_library rule uses posix names for extracting. When
materialized into a binary's venv, platform specific names are used:

* bin -> (posix) bin; (Windows) Scripts
* include -> (posix) include; (Windows) Include (capital i)
* data -> venv root directory

Along the way ...

* The data files (files under the "data" scheme of a whl) are now
  always included as part of depending on the library. They would be
  in included in venv_site_packages=yes mode, so this better aligns
  behavior of the two modes.
* Rename `is_venv_site_packages` to `_is_venv_site_packages_yes` to
  better represent the purpose and visibility of it.
* Make whl_from_dir support Windows. Testing of venvs relies on
  using it for testing various special cases.


aignas:
Be sure you check the whl_library extraction starlark code, because it is handling some of the mapping already.

python/config_settings/BUILD.bazel
```diff
@@ -229,7 +229,7 @@ string_flag(
 )
 
 config_setting(
-    name = "is_venvs_site_packages",
+    name = "_is_venvs_site_packages",
```
aignas: This is different from the PR description and the constant elsevhere.

rickeylev: Thanks, good catch. Fixed.

AI went on a spree and I missed this after cleaning up after it.

rickeylev:
@gemini-code-assist review

rickeylev:
Alrighty, I think this is ready. Some more work to make the things in bin actually runnable is needed, but this at least gets files into the right directory

python/private/venv_runfiles.bzl
```diff
@@ -452,19 +495,36 @@ def get_venv_symlinks(
 
     # Finally, for each group, we create the VenvSymlinkEntry objects
     for venv_path, files in optimized_groups.items():
+        if venv_path.startswith("data/"):
+            out_venv_path = venv_path[len("data/"):]
+            kind = VenvSymlinkKind.DATA
+            prefix = ""
+        elif venv_path.startswith("include/"):
```
aignas, line 502: What about windows wheels? Do we not need to check for capitalized includes?

tests/whl_with_build_files/BUILD.bazel
```diff
@@ -1,9 +1,9 @@
 load("//python:py_test.bzl", "py_test")
-load("//tests/support:support.bzl", "SUPPORTS_BZLMOD_UNIXY")
+load("//tests/support:support.bzl", "SUPPORTS_BZLMOD")
 
 py_test(
     name = "verify_files_test",
     srcs = ["verify_files_test.py"],
-    target_compatible_with = SUPPORTS_BZLMOD_UNIXY,
+    target_compatible_with = SUPPORTS_BZLMOD,
```
aignas, line 7: nit: would it be possible to do this `target_compatible_with` at the `somepkg_with_build_files`? Or are we running them on non-bzlmod as well?

Can be done in a separate PR.

aignas (approved):
LGTM with minor questions.

python/private/venv_runfiles.bzl
rickeylev: No, because the wheel extraction always uses unix names. Platform-specific renaming happens at the py_binary level.

Admittedly, I'm not entirely happy with this block of logic because it assumes how whl_library extracted the contents. Just another good reason we need to create a `py_extracted_wheel_library` rule

tests/whl_with_build_files/BUILD.bazel
rickeylev: To set TCW on somepkg_with_build_files, it would need to be set by whl_library.

A recent PR added some of the whl_from_dir stuff to workspace. The trick was to put it in internal_dev_setup.bzl -- whl_library relies on some earlier rules_python repo generation.

python/private/venv_runfiles.bzl
aignas: The rule would be taking what as an input? The wheel or extracted contents of the wheel? You mean the moving of the files around could be done in the build phase? I would definitely +1 that. With pipstar being the only code path that might be doable, but might require a 3.0 release.

rickeylev: Whl_library would read entry point.txt and generate a target for each entry point. This target generates a platform aware/specific file (which shows up as a KIND: BIN entry in PyInfo), but . This file is venv agnostic (by way of the polyglot trick or equivalent). Py binary can then just copy/symlink that file into the venv bin directory.

That's my rough thinking anyways

PR #3730, opened 2026-04-23
fix(gazelle): handle auto-included __init__.py when generating py_binary targets.

### LINKED ISSUE #3729: Gazelle: conflict with per-file generation, `python_generation_mode_per_file_include_init`, and `py_binary` targets
# 🐞 bug report

### Affected Rule

Gazelle

### Is this a regression?

No

### Description

When using the following configuration:
```
# gazelle:python_generation_mode file
# gazelle:python_generation_mode_per_file_include_init true
```

Gazelle attempts to generate a `py_binary` containing each module with an `if __name__ == '__main__'`, as well as a `py_library` with the same name containing just `__init__.py`.

See here: https://github.com/bazel-contrib/rules_python/blob/ccdfc8a7e6aa888e88bdbaee141eceef7a7611e4/gazelle/python/generate.go#L276-L281

Generation logic removes the executable module from `srcs`, relying on `srcs` being empty later on to skip generating a library. But with `__init__.py` added, `srcs` is never empty.

## 🔬 Minimal Reproduction

I will post a PR with an updated test case and a fix.

## 🔥 Exception or Error

Gazelle complains about conflicting targets with the same name.

jvolkman:
Fixes binary target generation when using per-file mode and automatically 
including __init__.py files.

Fixes #3729

### DIFF
--- a/gazelle/python/generate.go
+++ b/gazelle/python/generate.go
@@ -258,6 +258,15 @@ func (py *Python) GenerateRules(args language.GenerateArgs) language.GenerateRes
 	// Create a validFilesMap of mainModules to validate if python macros have valid srcs.
 	validFilesMap := make(map[string]struct{})
 
+	// Determine whether we have an __init__.py file in this package and whether we'll implicitly include it in srcs.
+	var hasPopulatedInit bool
+	var autoIncludeInit bool
+	if cfg.PerFileGeneration() {
+		var hasInit bool
+		hasInit, hasPopulatedInit = hasLibraryEntrypointFile(args.Dir)
+		autoIncludeInit = cfg.PerFileGenerationIncludeInit() && hasInit && hasPopulatedInit
+	}
+
 	appendPyLibrary := func(srcs *treeset.Set, pyLibraryTargetName string) {
 		allDeps, mainModules, annotations, err := parser.parse(srcs)
 		for name := range mainModules {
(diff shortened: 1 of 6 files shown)


gazelle/python/generate.go
```diff
@@ -258,6 +258,10 @@ func (py *Python) GenerateRules(args language.GenerateArgs) language.GenerateRes
 	// Create a validFilesMap of mainModules to validate if python macros have valid srcs.
 	validFilesMap := make(map[string]struct{})
 
+	// Determine whether we have an __init__.py file in this package and whether we'll implicitly include it in srcs.
+	hasInit, nonEmptyInit := hasLibraryEntrypointFile(args.Dir)
+	autoIncludeInit := cfg.PerFileGenerationIncludeInit() && hasInit && nonEmptyInit
```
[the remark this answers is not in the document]
jvolkman: Done. Thanks Gemini!

gazelle/python/generate.go
```diff
@@ -258,6 +258,15 @@ func (py *Python) GenerateRules(args language.GenerateArgs) language.GenerateRes
 	// Create a validFilesMap of mainModules to validate if python macros have valid srcs.
 	validFilesMap := make(map[string]struct{})
 
+	// Determine whether we have an __init__.py file in this package and whether we'll implicitly include it in srcs.
+	var nonEmptyInit bool
```
dougthor42: I've never been a fan of this negative boolean logic. Let's switch to positive booleans:

```go
var populatedInit bool
```

dougthor42 (commented):
Oh, and please add a line to `CHANGELOG.md`

gazelle/python/generate.go
jvolkman: Named it `hasPopulatedInit` to match the existing `hasInit`.

dougthor42 (approved):

PR #3731, opened 2026-04-23
feat(toolchains): add add_target_settings to python.override

### LINKED ISSUE #3673: feature: avoid a fallback to the default-registered toolchains when using config_settings
# 🚀 feature request

### Relevant Rules

This applies to the toolchain registration extensions.

### Description

We have a product with multiple parallel python ecosystems, and I'm trying to make them more separate mentally, for our sanity. We're using the idiom described in https://github.com/bazel-contrib/rules_python/issues/3671#issuecomment-4101798143 to declare a custom family of toolchains, like:

```python
string_flag(
    name = "python_toolchain_family", 
    build_setting_default = "prebuilt", 
    values = ["prebuilt", "custom"]) 
```

and then we attached the `custom` config_setting to our custom toolchains (via target_settings) like:

```python

    native.toolchain(
        ...
        target_settings = [
            "@rules_python//python/config_settings:is_python_3.9",
            "//thirdparty:python_toolchain_family_custom",
        ],
        toolchain = "//thirdparty/Python39:runtime_pair",
        toolchain_type = "@rules_python//python:toolchain_type",
    )
```

So far so good. However, the problem is, when we request an illegal combination, like:

```python
py_binary(
    python_version = "3.11", 
    config_settings = {
        "//:python_toolchain_family": "custom"
    }
)
``` 

(when we didn't register 3.11 as a custom), we then get the default-registered python 3.11 toolchain, instead of getting a toolchain election error. 

I'd really like a way to unconditionally attach the config to all of the other registered toolchains, too, like 

```python
python.override(toolchain_target_settings = ["//:python_toolchain_family_prebuilt"])
```

So there's a clean separation between our various custom ones and the prebuilt ones. I know this is possible manually with `python.single_version_platform_override`, but in order for this to work I have to attach it to the entire transitive graph of toolchains.

I'm not super familiar with the rules_python internals, but it might also make sense to attach it to something like `config.add_transition_setting`. I'm sure there are also probably ways to do this that I haven't thought of :)

novas0x2a:
Users who register custom toolchain families (e.g. gated behind a
string_flag like "custom" vs "prebuilt") have no way to prevent the
default-registered toolchains from acting as a silent fallback. When a
user requests a custom toolchain at a version they didn't register,
Bazel's toolchain resolution quietly falls back to the default prebuilt
toolchain instead of producing an error.

This adds a `toolchain_target_settings` attribute to `python.override`
that appends the given `config_setting` labels to the `target_settings`
of every toolchain registered by the module extension:

```starlark
python.override(
    toolchain_target_settings = ["@@//:python_toolchain_family_prebuilt"],
)
```

These settings are appended to the `target_settings` of all toolchains
registered by the extension, including any that already have settings
from `python.single_version_platform_override`. Toolchains registered
outside the extension (e.g. via `local_runtime_toolchains_repo`) are not
affected.

Note: `toolchain_target_settings` is popped from the config `default`
dict before it reaches `python_register_toolchains()`, so it doesn't
leak into the `python_repository` kwargs.

Integration test passes across the full Bazel version matrix (7.7.0,
8.5.1, 9.0.0rc1). Verified the test fails without the fix (default
toolchain silently resolves as fallback).

Fixes https://github.com/bazel-contrib/rules_python/issues/3673

### DIFF
--- a/python/private/python.bzl
+++ b/python/private/python.bzl
@@ -1173,7 +1173,7 @@ These settings are appended to the `target_settings` of all toolchains
 registered by the extension, including any that already have settings
 from `python.single_version_platform_override`.
 
-:::{versionadded} VERSION_NEXT
+:::{versionadded} VERSION_NEXT_FEATURE
 :::
 """,
         ),
--- a/python/private/python.bzl
+++ b/python/private/python.bzl
@@ -403,6 +403,10 @@ def _python_impl(module_ctx):
     # the PLATFORMS global for this toolchain
     toolchain_platform_keys = {}
 
+    # Extra target_settings to add to every registered toolchain, e.g. for
+    # gating the default toolchains behind a custom config_setting.
+    global_toolchain_target_settings = py.config.toolchain_target_settings
+
     # Split the toolchain info into separate objects so they can be passed onto
     # the repository rule.
     for entry in toolchain_impls:
@@ -414,7 +418,7 @@ def _python_impl(module_ctx):
 
         # The target_settings attribute may not be present for users
         # patching python/versions.bzl.
-        toolchain_ts_map[key] = getattr(entry.platform, "target_settings", [])
+        toolchain_ts_map[key] = getattr(entry.platform, "target_settings", []) + global_toolchain_target_settings
         toolchain_platform_keys[key] = entry.platform_name
         toolchain_python_versions[key] = entry.full_python_version
 
@@ -702,6 +706,9 @@ def _process_global_overrides(*, tag, default, _fail = fail):
 
         default["minor_mapping"] = tag.minor_mapping
 
+    if tag.toolchain_target_settings:
+        default["toolchain_target_settings"] = list(tag.toolchain_target_settings)
+
     forwarded_attrs = sorted(AUTH_ATTRS) + [
         "base_url",
         "register_all_versions",
@@ -809,6 +816,7 @@ def _get_toolchain_config(*, modules, _fail = fail):
     )
 
     register_all_versions = default.pop("register_all_versions", False)
+    toolchain_target_settings = default.pop("toolchain_target_settings", [])
     kwargs = default.pop("kwargs", {})
 
     versions = {}
@@ -834,6 +842,7 @@ def _get_toolchain_config(*, modules, _fail = fail):
         minor_mapping = minor_mapping,
         default = default,
         register_all_versions = register_all_versions,
+        toolchain_target_settings = toolchain_target_settings,
     )
 
 def _compute_default_python_version(mctx):
@@ -1144,6 +1153,30 @@ The values in this mapping override the default values and do not replace them.
             default = {},
         ),
         "register_all_versions": attr.bool(default = False, doc = "Add all versions"),
+        "toolchain_target_settings": attr.string_list(
+            mandatory = False,
+            doc = """\
+A list of `config_setting` labels to add to the `target_settings` of every
+toolchain registered by this module extension. This is useful for creating
+separate "families" of toolchains gated behind custom build settings.
+
+For example, to ensure the default prebuilt toolchains are only resolved when
+a `prebuilt` config setting is active:
+
+```starlark
+python.override(
+    toolchain_target_settings = ["@@//:python_toolchain_family_prebuilt"],
+)
+```
+
+These settings are appended to the `target_settings` of all toolchains
+registered by the extension, including any that already have settings
+from `python.single_version_platform_override`.
+
+:::{versionadded} VERSION_NEXT
+:::
+""",
+        ),
     } | AUTH_ATTRS,
 )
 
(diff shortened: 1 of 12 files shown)


python/private/python.bzl
```diff
@@ -1144,6 +1153,31 @@ The values in this mapping override the default values and do not replace them.
...
+```starlark
+python.override(
+    toolchain_target_settings = ["@@//:python_toolchain_family_prebuilt"],
+)
+```
+
+Any user-registered toolchains (via `python.single_version_platform_override`
+with `target_settings`) already have their own settings and are unaffected by
+this attribute; this only appends to the settings derived from the platform
+definitions.
```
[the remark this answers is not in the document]
novas0x2a: agreed; fixed in latest push.

tests/integration/toolchain_target_settings/MODULE.bazel
```diff
@@ -0,0 +1,37 @@
...
+)
+
+python = use_extension("@rules_python//python/extensions:python.bzl", "python")
+python.toolchain(python_version = "3.13")
+
+# Gate ALL default-registered toolchains behind the "prebuilt" config setting.
+# This prevents them from being a silent fallback when a different toolchain
+# family is requested.
+python.override(
+    toolchain_target_settings = ["@@//:is_prebuilt"],
```
aignas:
```suggestion
    add_target_settings = ["@@//:is_prebuilt"],
```

It would be good to bikeshed on the name a little before this gets merged. What would be a good name here?
* `add_target_settings`
* `add_toolchain_target_settings`

python/private/python.bzl
```diff
@@ -1144,6 +1153,30 @@ The values in this mapping override the default values and do not replace them.
             default = {},
         ),
         "register_all_versions": attr.bool(default = False, doc = "Add all versions"),
+        "toolchain_target_settings": attr.string_list(
+            mandatory = False,
+            doc = """\
+A list of `config_setting` labels to add to the `target_settings` of every
+toolchain registered by this module extension. This is useful for creating
+separate "families" of toolchains gated behind custom build settings.
```
aignas: It would be great if the docs showed how to create those families. Do we use the `python` extension multiple times, but with `isolate = True`? Do we need to define the toolchains ourselves?

An alternative to this is to just have a global flag that is built-into our config settings that is called `//python/config_settings:use_hermetic_toolchain=<yes|no|auto>`.

I'll wait for @rickeylev to also chime in here.

novas0x2a: I believe the approach we take today for creating the family concept is exactly the approach you mentioned in https://github.com/bazel-contrib/rules_python/issues/3673, though the actual _creation_ of those toolchains is a bit fraught. In our particular case, we have a custom-built toolchain using rules_foreign_cc, which we then promote into a full toolchain with a dance that looks like:
```
    native.toolchain(
        name = name,
        target_compatible_with = ["@platforms//os:linux"],
        target_settings = [
            "@rules_python//python/config_settings:is_python_3.9",
            "//thirdparty:python_toolchain_family_custom",
        ],
         toolchain = "//thirdparty/Python39:runtime_pair",
         toolchain_type = "@rules_python//python:toolchain_type",
```

The actual toolchain we build internally is almost certainly not an example that anyone would want to follow, but I'd be glad to add some less-weird docs as part of this, I just don't know what to add :laughing:

tests/integration/toolchain_target_settings/MODULE.bazel
novas0x2a: yeah, something with "add_" in it definitely seems better than my first cut, since it's definitely an append

python/private/python.bzl
aignas: Ah, OK, understood, then the family docs are not needed here in the docstring. Let's just decide on the name.

rickeylev: > have a use_hermetic_toolchains=yes|no

I'm mixed on this idea. I like it because it makes it easy to enable/disable on a per-target or command line basis Out Of The Box. But, it doesn't disable it by default (what nova wants to do), and controlling the config setting implementation gives the max flexibility to the user. Since we have a PR already for this, lets go with this.

tests/integration/toolchain_target_settings/MODULE.bazel
rickeylev: +1 to add_target_settings. The `python` extension is all about toolchan stuff, so no need for the "toolchain_" prefix. I agree about "add" to clarify its additional; the single_platform_override api uses it in a "clobber" sense, in comparison.

rickeylev (commented):
This is great. Thanks for the clean, well thought out and reasoned PR!

rickeylev:
I've pushed some prs to cleanup and rename. LGTM, assuming CI is happy (though it seems to be very flaky network today)

rickeylev (approved):

PR #3732, opened 2026-04-24
feat: gazelle add python_strip_import_prefix directive for src-layout packages

taowang487:
## Details
Adds a new `# gazelle:python_strip_import_prefix <prefix>` directive that strips a path segment from the filesystem path before computing Python import specs.

  This solves the src-layout problem: projects using pyproject.toml with `packages = ["src/mylib"]` (Hatch, setuptools, etc.) have a `src/` directory that is stripped at install time and is not part of the Python import path. Without this directive, Gazelle indexes `src/mylib` as the import spec src.mylib, causing all cross-package dependency resolution to fail and requiring dozens of manual `# gazelle:resolve` workarounds.

  Usage

  In a `BUILD.bazel` alongside `# gazelle:python_root`:
```
  # gazelle:python_root
  # gazelle:python_strip_import_prefix src
```

  This tells Gazelle to strip `src/` when computing import specs, so `src/mylib/__init__.py` is indexed as `mylib` (not `src.mylib`), and consumers that `import mylib` resolve correctly to `//src/mylib` automatically.

## Testing
Added a unit test

Also tested in a toy project, `# gazelle:python_strip_import_prefix src` can fix the error below
```
gazelle: ERROR: failed to validate dependencies for target "@gazelle_test//consumer":

"consumer/app.py", line 1: "mylib.greet" is an invalid dependency: possible solutions:
        2. Use the '# gazelle:resolve py mylib.greet TARGET_LABEL' BUILD file directive to resolve to a known dependency.
        3. Ignore it with a comment '# gazelle:ignore mylib.greet' in the Python file.

"consumer/app.py", line 1: "mylib" is an invalid dependency: possible solutions:
        2. Use the '# gazelle:resolve py mylib TARGET_LABEL' BUILD file directive to resolve to a known dependency.
        3. Ignore it with a comment '# gazelle:ignore mylib' in the Python file.
```


aignas:
Why not have the python_root directive within the src/BUILD file? Is it so that you can keep the test files in the test folder at the python_root folder? Please add more text about the motivation for this new directive.

taowang487:
Hi @aignas, thanks for the review! I am using `python_root` in my `src/BUILD`. My original motivation was to eliminate the many `# gazelle:resolve py` directives I needed in my codebase.  After further investigation I realized that I can fix it the other way and `python_root` should be more than enough. Closing this PR. Thanks again, and have a great weekend!

dougthor42:
+1, `python_root` should be sufficient. I'd be interested in seeing what your current `resolve` directives are.

For example, we structure our code like so:

```
MODULE.bazel
pyproject.toml
rust/
  BUILD.bazel
  crate1/
    BUILD.bazel
    src/
  crate2/
    BUILD.bazel
    src/
src/             # pretty much only python code
  BUILD.bazel    # contains `#gazelle:python_root`
  python_package_1/
    BUILD.bazel
    file.py
    subpackage/
  python_package_2/
    BUILD.bazel
    other_file.py
tools/bazel/
  BUILD.bazel
  repositories.bzl
  private/
    BUILD.bazel
    repositories.bzl
```

And have not had to include any `resolve` directives to support that layout.

PR #3735, opened 2026-04-24
chore(entry_point): remove all repository_rule entry_point code and docs

aignas:
This PR remove any mention or supporting code for the legacy
(and long removed) `entry_point` macro.

Fixes #3642

PR #3743, opened 2026-04-27
feat(venv): make wheel scripts runnable in venv

### LINKED ISSUE #3202: Support non-entry point declared executables in wheels
While helping a user, I was pointed to an older dtschema wheel: https://pypi-browser.org/package/dtschema/dtschema-2021.10-py3-none-any.whl

Instead of having formally declared entry_points, it has e.g. `{name}.data/scripts/{script-name}` entries e.g. `dtschema-2021.10.data/scripts/dt-doc-validate`. I'm guessing its some older way of packaging entry points?

These are `#!python` shebang scripts, so obviously they're expecting an installer to rewrite them. They have custom startup logic.

Normally we point users to py_console_script_binary, however, using it in this case is tought because of the custom startup logic. Similarly, re-exporting the file directory doesn't help because of the shebang.

I think, ideally, we could detect these during the repo phase and have a rule that wraps or transforms them. All it really has to do is copy the content into a .py file, then stick that in a py_binary.

More short term, some docs or helper rule for how to get these to be executable and runnable.

rickeylev:
Currently, `#!python` based scripts in a wheel are put into the venv
as-is, so aren't actually runnable. Additionally, wheel entry points
aren't installed into the venv.

To fix, rewrite the scripts to re-exec themselves with the `python3`
binary in the same directory.

Also adds support for wheel entry points in venvs. This is done by
having the repo-phase parse entry_points.txt and defining a target for
each entry point. These targets run during build phase, so are able
to generate platform-specific outputs.

Cross-building with Windows is also added, supporting both Windows
as a target or exec platform to generate outputs for the target
platform.

Fixes https://github.com/bazel-contrib/rules_python/issues/3202


rickeylev:
Oh neat, it looks like entry points already work? I'm guessing something in whl_library is generating something from the entry points config?

rickeylev:
Hm, after poking around a bit, i'm guessing the entry points stuff is actually relying on repo-phase code generating the scripts. It shouldn't be too hard to replace that with a build-time generation rule.

aignas:
I have actually removed the entry point parsing/generation at head with #3735. If you need the code feel free to restore some of it. Should be easy to do a partial revert and just plumb through the entry point script names to create py_console_script_binary targets for.

aignas:
That said, I was thinking that it would be best to return the console scripts via the PyInfo provider (or some other provider) so that all we have to do in py_console_script_binary is get the file from the provider instead of parsing an entry_points.txt file.

I think it is very doable to also use the same provider to expand the console script template on the fly in the venv.

rickeylev:
> I just removed the entry points generated

:joy: :sob: lol, ah, what timing. But yeah, shouldn't be hard to re-add it. All it needs to do is read the entry points config and generate a platform-specific file.

> return console scripts on PyInfo or some other provider for py_console_script_binary

Two questions:
1. What problem does that solve / functionality it enable?
2. Does it affect the py_console_script_binary API?

For (2), I think we had some hacky label-manipulation in the macro to pick the entry points script? A provider would allow figuring that out at analysis phase instead.

aignas:
>     1. What problem does that solve / functionality it enable?
> 
Currently the `py_console_script_binary` is building a full runfiles tree and for the usecase that you are implementing in here - a `py_binary` is just calling utilities within the same `venv` that is a little bit too heavy. I want to separate the usecases for `py_console_script_binary` to be only something where we actually want to build a full separate `venv` for the said scripts.


>     2. Does it affect the py_console_script_binary API?
> 
> 
> For (2), I think we had some hacky label-manipulation in the macro to pick the entry points script? A provider would allow figuring that out at analysis phase instead.

It should not affect the external API and I agree, the implementation would be way cleaner as we would only depend on the `:pkg` `py_library` or `py_whl_library` target.

python/private/pypi/venv_entry_point_template.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
```
aignas, line 1: Is this compatible with nix? I remember that `/bin/bash` should be replaced with `/usr/bin/env bash` on those setups, so I am wondering if it is similar for sh.

python/private/pypi/whl_extract.bzl
```diff
@@ -20,22 +20,8 @@ def whl_extract(rctx, *, whl_path, logger):
...
-            rctx,
-            op = "Fixing wheel permissions {}".format(whl_path),
-            arguments = ["chmod", "-R", "a+rX", str(install_dir_path)],
-            logger = logger,
-        )
-        if result.return_code != 0:
-            # It's possible chmod is not available or the filesystem doesn't support it.
-            # This is fine, we just want to try to fix permissions if possible.
-            logger.warn(lambda: "Failed to fix file permissions: {}".format(result.stderr))
+    _maybe_fix_permissions(rctx, whl_path = whl_path, logger = logger)
```
aignas, line 23: BTW, this permissions fixing is only needed on bazel 8.5 and below. The later versions do read-only permission fixes during extracting.

python/private/pypi/whl_extract.bzl
```diff
@@ -70,17 +56,36 @@ def whl_extract(rctx, *, whl_path, logger):
...
+    # Ensure that there is no data dir left
         rctx.delete(data_dir)
 
+def _maybe_fix_permissions(rctx, *, whl_path, logger):
+    # Fix permissions on extracted files. Some wheels have files without read permissions set,
+    # which causes errors when trying to read them later.
+    # We apply this to the root directory to ensure that everything in bin/, site-packages/,
+    # etc. is readable and executable where appropriate.
+    os_name = repo_utils.get_platforms_os_name(rctx)
+    if os_name != "windows":
```
aignas, line 75: BTW, this permissions fixing is only needed on bazel 8.5 and below. The later versions do read-only permission fixes during extracting. This `if` statement could include the version checks as well.

python/private/pypi/whl_library_targets.bzl
```diff
@@ -180,6 +190,28 @@ def whl_library_targets(
...
+        ep_name = kwargs.pop("name")
+        ep_target_name = "bin/{}".format(ep_name)
+        rules.venv_entry_point(
+            name = ep_target_name,
+            **kwargs
+        )
+        data.append(ep_target_name)
+
+    # NOTE: We assume there is no overlap between the rewrite-bin file names,
+    # bin/ source files and generated bin/ files.
```
aignas: We could just verify this assumption and print warnings or fail.

python/private/repo_utils.bzl
```diff
@@ -522,6 +522,38 @@ def _extract(mrctx, *, archive, supports_whl_extraction = False, **kwargs):
...
+    Args:
+        mrctx: module_ctx or repository_ctx object
+        src: {type}`path` the source path
+        dest: {type}`path` the destination path
+    """
+    if hasattr(mrctx, "rename"):
+        mrctx.rename(src, dest)
+        return
+
+    # Fallback for Bazel < 8.0
```
aignas, line 539: With this, we may remove the `wheel.py` file because we have the right fallbacks.

aignas (approved):
This is really good, thank you very much!

python/private/pypi/venv_entry_point_template.sh
rickeylev: Searching around a bit, everything I find says /bin/sh is preferred for nix

python/private/pypi/whl_extract.bzl
rickeylev: I've added a comment to that effect on the maybe_fix_permisisons function

python/private/pypi/whl_extract.bzl
rickeylev: I'll leave that for a separate PR

python/private/repo_utils.bzl
rickeylev: I think we'd need this regardless? The starlark extract logic still has to look through the bin/ directory and and move files to rewrite-bin

python/private/pypi/whl_library_targets.bzl
rickeylev: I've added guard logic.

rickeylev:
> Currently the py_console_script_binary is building a full runfiles tree and for the usecase that you are implementing in here - a py_binary is just calling utilities within the same venv that is a little bit too heavy. I want to separate the usecases for py_console_script_binary to be only something where we actually want to build a full separate venv for the said scripts.

Ah, I think I get what you mean. Do you want to file an issue for further discussion? entrypoints/scripts/things-in-bin seem distinct enough they warrant _something_ in a provider, but it's not clear to me how a consumer can make effective use of it via a py_binary.

rickeylev:
I got everything working except pythonw scripts under bazel 7. The root cause for this is wheel.py. We've hard-coded it to use a unix install format, so it rewrites #!pythonw to #!python. Later, when the rewrite tool goes to fixup the command line, it doesn't see pythonw, so can't generate the correct pythonw.exe part.

Since we're going to remove wheel.py, I've just skipped that particular test.

python/private/repo_utils.bzl
aignas: Well as far as what the wheel.py is doing, we have reimplemented everything in starlark a while ago. The only reason why we still needed to have something like the wheel.py is to install sdists, or use pip. The extraction logic can be done fully in starlark.

PR #3744, opened 2026-04-27
fix(pypi): harden the WORKSPACE python detection in pip_repository

aignas:
A fixup to #3737

### DIFF
--- a/python/private/pypi/pip_repository.bzl
+++ b/python/private/pypi/pip_repository.bzl
@@ -90,9 +90,13 @@ def _pip_repository_impl(rctx):
         python_interpreter = rctx.attr.python_interpreter,
         python_interpreter_target = rctx.attr.python_interpreter_target,
     )
-    result = rctx.execute([python_interpreter, "-S", "-c", "import sys; print(sys.version.split()[0])"])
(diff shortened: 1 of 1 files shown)


aignas:
PTAL

python/private/pypi/pip_repository.bzl
```diff
@@ -90,11 +90,10 @@ def _pip_repository_impl(rctx):
         python_interpreter = rctx.attr.python_interpreter,
         python_interpreter_target = rctx.attr.python_interpreter_target,
     )
-    result = rctx.execute([python_interpreter, "--version"])
-    if result.stdout:
-        python_version = result.stdout.strip().split(" ")[-1]
-    else:
-        fail("Could not determine Python version")
+    result = rctx.execute([python_interpreter, "-S", "-c", "import sys; print(sys.version.split()[0])"])
```
rickeylev: Isn't there a helper in pypi_repo_utils for running python? Do you want to use that instead?

At the least, use repo_utils.checked_execute (or w/e its called), since that does logging and error reporting for you.

Do you want -S, or -I?

In any case, add a comment about why -S is added

rickeylev (approved):
lgtm other than the 1 comment

PR #3748, opened 2026-04-30
chore: always use starlark-based extraction for wheels

rickeylev:
Currently, the Python `installer` library is used for extracting a wheel when Bazel 7 is used. This library only outputs platform-specific extractions, and it was hard-coded to use a unixy style install. This ends up hiding Windows `pythonw`-based scripts and entry points because those are normalized to simply `python` on unix.

To fix, use the Starlark-based extraction logic instead. This is a simpler implementation that extracts the wheels mostly as-is, which allows build-phase logic to handle platform-specific differences.

This also makes the Starlark based extraction work for Bazel 8.3 and earlier, which previously didn't support it.


rickeylev:
Am I missing something? This seems _too_ easy to remove :sweat_smile: 

Ah there we go, CI flagged something...

rickeylev:
@aignas re: the gemini comment that its incorrectly marked reproducible even if an sdist is built: 

I think it's right, but that bug looks to be pre-existing. Right now, pipstar extraction is used for bazel 8+, so it'll always be marked reproducible, even if an sdist was used. I'm going to leave that alone in this PR.

aignas (approved):
LGTM, please also cleanup the python sources that handle extraction and the config repo of unused config flags.

rickeylev:
> LGTM, please also cleanup the python sources that handle extraction and the config repo of unused config flags.

Done! Killed wheel.py and its supporting code.

PR #3752, opened 2026-05-01
fix(pypi): include RECORD file in installed wheel targets

rickeylev:
Currently, `importlib.metadata.files()` returns `None` for packages installed via `rules_python`'s wheel rules. This happens because the `RECORD` file in the `.dist-info` directory is explicitly excluded from the `data` attribute of the generated `py_library` targets.

To fix this, remove `**/*.dist-info/RECORD` from the `_data_exclude` list in `whl_library_targets.bzl`. This ensures that the `RECORD` file is preserved in the runfiles and available at runtime, enabling `importlib.metadata.files()` to correctly list the files in the package.

Work towards https://github.com/bazel-contrib/rules_python/issues/3024

### DIFF
--- a/python/private/pypi/whl_library_targets.bzl
+++ b/python/private/pypi/whl_library_targets.bzl
@@ -374,10 +374,6 @@ def whl_library_targets(
             "**/*.py",
             "**/*.pyc",
             "**/*.pyc.*",  # During pyc creation, temp files named *.pyc.NNNN are created
-            # RECORD is known to contain sha256 checksums of files which might include the checksums
-            # of generated files produced when wheels are installed. The file is ignored to avoid
-            # Bazel caching issues.
-            "**/*.dist-info/RECORD",
         ]
         for item in data_exclude:
             if item not in _data_exclude:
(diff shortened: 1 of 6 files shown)


python/private/pypi/whl_library_targets.bzl
```diff
@@ -374,10 +374,6 @@ def whl_library_targets(
             "**/*.py",
             "**/*.pyc",
             "**/*.pyc.*",  # During pyc creation, temp files named *.pyc.NNNN are created
-            # RECORD is known to contain sha256 checksums of files which might include the checksums
-            # of generated files produced when wheels are installed. The file is ignored to avoid
-            # Bazel caching issues.
-            "**/*.dist-info/RECORD",
```
aignas, line 380: This should be removed only if the sdist_filename is None.

aignas (approved):

python/private/pypi/whl_library_targets.bzl
rickeylev: Done

rickeylev:
I think we should do the same for sdist cases, but since this PR is good to go, I'm merging it now.

PR #3755, opened 2026-05-07
feat(pypi): restrict pip.parse hub visibility

### LINKED ISSUE #3413: `pip.parse` should allow hiding transitive dependencies
# 🚀 feature request

### Relevant Rules

`pip.parse`

### Description

Given a `requirements.in` file containing only `foo==4.2.0` from which I create a lock file with content
```
foo==4.2.0 --hash=<some_hash>
dep_of_foo==1.33.7 --hash=<some_hash>
```

When using `pip.parse` to crate a hub `@pypi` from this lock file users can access all Python modules from the lock file. Concretely, `@pypi//foo` and `@pypi//dep_of_foo`.

I consider `dep_of_foo` an implementation detail which no user should depend on. When changing the version of `foo`, then `dep_of_foo` might vanish or change drastically as side effect. If I wanted users to access `dep_of_foo`, I would have added it to the `requirements.in` file to make it an explicit and desired direct dependency of my project.

It would be great if there were an option enforcing that transitive dependencies are not available to users.

Ideally `pip.parse` would offer an attribute `restrict_visibility_to` (or any other name) which takes a file list. Then, one could provide one or multiple `requirements.in` files to this attribute. `pip.parse` can then read those files, extract the Python module names and ensure only those are public targets in the pip hub.

The implementation would be easier if `restrict_visibility_to` takes a list of strings and the user explicitly states which Python modules should be public. However, I consider this inferior, as it increases the maintenance burden whenever changing the `requirements.in` files.

I implemented the described behavior locally as a workspace rule, which creates a new hub with alias targets pointing to the hub created by `pip.parse`. It is trivial to do so, not much logic is required.

However, this means one has to teach people not to use the original hub created by `pip.parse`. Or one has to write yet another piece of custom code for a BUILD file checker ensuring this rule.
Overall, it would be much nicer if this behavior would be a feature of upstream `pip.parse`.

Harshal96:
Add a restrict_visibility_to attribute for bzlmod pip.parse so hub aliases are public only for packages listed in direct requirement files.

Keep all lockfile packages available to generated wheel repositories so transitive dependencies continue to resolve internally.

Fixes #3413

PR Instructions/requirements
  section at the bottom.
* Body text describes:


python/private/pypi/parse_requirements.bzl
```diff
@@ -62,6 +63,9 @@ def parse_requirements(
             the platforms stored as values in the input dict. Returns the same
             dict, but with values being platforms that are compatible with the
             requirements line.
+        exposed_requirements: List of requirements files. When present, only
```
aignas: Should we key this by_platform as well?

python/private/pypi/parse_requirements.bzl
```diff
@@ -251,6 +273,19 @@ def parse_requirements(
 
     return ret
 
+def _exposed_package_names(ctx, exposed_requirements):
+    """Parse the requirement files that define hub-exposed package names."""
+    if not exposed_requirements:
+        return None
+
+    exposed = {}
```
aignas: This acts like a union. Is this the right thing to do here?

From the modeling perspective, the requirements may be different based on the target platform. That said, I think this approach is OK in the parse_requirements internals. Since we are asking the user to provide the source lock files.

python/private/pypi/extension.bzl
```diff
@@ -786,6 +786,21 @@ The Python version the dependencies are targetting, in Major.Minor format
 If an interpreter isn't explicitly provided (using `python_interpreter` or
 `python_interpreter_target`), then the version specified here must have
 a corresponding `python.toolchain()` configured.
+""",
+        ),
+        "restrict_visibility_to": attr.label_list(
```
aignas: I like adding this, but I imaging that we could repurpose this set of files for more labels.
1. Restrict visibility.
2. Create a `@pypi//:lock.update` target automatically that includes the `srcs` passed here as `srcs` passed to the lock file. This plumbing would be similar to what `rules_jvm` have with their `repin`.

We should also assume that we will need to support `pyproject.toml` files/file via the same interface in the future.

python/private/pypi/hub_builder.bzl
```diff
@@ -510,6 +510,7 @@ def _create_whl_repos(
         extra_pip_args = pip_attr.extra_pip_args,
         get_index_urls = self._get_index_urls.get(pip_attr.python_version),
         evaluate_markers = _evaluate_markers(self, pip_attr),
+        exposed_requirements = pip_attr.restrict_visibility_to,
```
aignas: Should we create a separate function for this?
```
    exposed_requirements = parse_dep_srcs(module_ctx, pip_attr.restrict_visibility_to)
```

That way we make the code more maintainable and easier to test.

python/private/pypi/parse_requirements.bzl
```diff
@@ -231,17 +236,34 @@ def parse_requirements(
         #     for p in dist.target_platforms
         # ]
 
+        normalized_name = normalize_name(name)
+        is_exposed = len(requirement_target_platforms) == len(requirements)
+        if exposed_package_names != None and normalized_name not in exposed_package_names:
+            is_exposed = False
```
aignas: I would like this logic to be somewhere outside the `parse_requirements.bzl` file. Would it be possible to move it to `hub_builder.bzl`? The output of `parse_requirements` have this `is_exposed` flag (that is an internal implementation detail) and if we make the `lock sources` mandatory, then we would be able to drop that implementation detail, which actually makes the code a little bit more complex than it needs to be.

python/private/pypi/parse_requirements.bzl
```diff
@@ -231,17 +236,34 @@ def parse_requirements(
         #     for p in dist.target_platforms
         # ]
 
+        normalized_name = normalize_name(name)
+        is_exposed = len(requirement_target_platforms) == len(requirements)
+        if exposed_package_names != None and normalized_name not in exposed_package_names:
```
aignas: I don't like the double negatives, maybe the following is easier to read?

```suggestion
        if not (exposed_package_names and normalized_name in exposed_package_names):
```

aignas (commented):
Thank you for the contribution.

aignas:
My suggestion for this would be to:
* Rebase and keep in mind the design in #3514.
* Add a list of `deps` to the fields stored in the `pyproject.toml` struct.
* Use that to set the visibility.

Leave out the locking out of scope for now to keep this PR focused.

PR #3760, opened 2026-05-08
fix(logger): do not output WARN level logs for non-root modules

### LINKED ISSUE #3749: Performance degradation and warning messages after upgrading to rules_python 2.0.0 (was fine in 1.9.0)
# 🐞 bug report

### Affected Rule

Likely related to pip_deps/pip extension handling in rules_python.

### Is this a regression?

 Yes, the previous version in which this bug was not present was: 1.9.0

### Description

After upgrading from rules_python 1.9.0 to 2.0.0, every shutdown and build results in multiple warnings about the requirements file being generated without hashes (see below). In addition, there is a recurring performance regression observed. With 1.9.0, the build completes in about 10 seconds, but with 2.0.0 it takes about 5 seconds longer (~15s). These warnings and performance issues occur on every shutdown/build.

**Warning log:**
```
rules_python:pypi:hub:pip_deps WARNING: requirement file has been generated without hashes, will use all hashes for the given version  that could find on the index:
    []
INFO: Analyzed 4 targets (101 packages loaded, 4766 targets configured).
INFO: Found 4 targets...
INFO: Elapsed time: 14.471s, Critical Path: 0.34s
INFO: 31 processes: 28 internal, 1 linux-sandbox, 2 local.
INFO: Build completed successfully, 31 total actions

long running task 
# Fetching module extension @@rules_python+//python/extensions:pip.bzl%pip; Fetch package lists from PyPI index 5s

With rules_python 1.9.0 it takes ~5s less and no warnings.
```

**Commands to reproduce:**
```
sudo rm -rf $(dirname $(bazelisk info output_base)); bazelisk clean --expunge; bazelisk shutdown; time bazelisk build :all --disk_cache= --remote_cache=
```

### Tested on
- RHEL 9 (x86_64)
- Fedora Workstation latest (x86_64)

## 🔬 Minimal Reproduction

Sample minimal project:
```diff
# python-test/.bazelversion
+8.3.1
# python-test/BUILD.bazel
+load("@rules_python//python:defs.bzl", "py_binary")
+load("@rules_python//python:pip.bzl", "compile_pip_requirements")
+
+compile_pip_requirements(
+    name = "requirements",
+    requirements_in = "requirements.in",
+    requirements_txt = "requirements_lock.txt",
+)
+
+py_binary(
+    name = "main",
+    srcs = ["main.py"],
+    deps = ["@pypi//certifi"],
+)
# python-test/MODULE.bazel
+module(
+    name = "example",
+    version = "0.0.1",
+)
+
+bazel_dep(name = "rules_python", version = "1.9.0")
+
+python = use_extension("@rules_python//python/extensions:python.bzl", "python")
+python.toolchain(python_version = "3.11.9")
+
+pip = use_extension("@rules_python//python/extensions:pip.bzl", "pip")
+pip.parse(
+    hub_name = "pypi",
+    python_version = "3.11.9",
+    requirements_lock = "//:requirements_lock.txt",
+)
+use_repo(pip, "pypi")
# python-test/main.py
+print("Hello world!")
# python-test/requirements.in
+certifi
# python-test/requirements_lock.txt
+#
+# This file is autogenerated by pip-compile with Python 3.11
+# by the following command:
+#
+#    bazel run //:requirements.update
+#
+certifi==2026.4.22 \
+    --hash=sha256:3cb2210c8f88ba2318d29b0388d1023c8492ff72ecdde4ebdaddbb13a31b1c4a \
+    --hash=sha256:8d455352a37b71bf76a79caa83a3d6c25afee4a385d632127b6afb3963f1c580
+    # via -r requirements.in
```

## 🔥 Exception or Error

See warning output above.

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
RHEL 9 (x86_64)
Fedora Workstation latest (x86_64)
  </code>
</pre>

<pre>
  <code>
Bazelisk 8.3.1
  </code>
</pre>

**Rules_python version:**
<pre>
  <code>
1.9.0 (works fine)
2.0.0 (shows warnings and is slower)
  </code>
</pre>

**Anything else relevant?**

Regression appears with rules_python 2.0.0 and is not present in 1.9.0; please investigate the hash-handling regression and performance difference.

---

aignas:
A user has reported issues which actually resulted from a particular
`bazel` version being used and once he updated, the issues disappeared
because of changes in implicit non-root module dependencies.

In order to reduce the warning spam of non-root modules, set the default
level to `ERROR` for non-root module parsing context and `WARN` for root
module context. This means that the users will not get console output
that they have no way to fix without patching upstream dependencies.

At the same time ensure that the name of the module is printed together
with the logs to better understand where something is coming from.

Fixes #3749


rickeylev (approved):

python/private/repo_utils.bzl
```diff
@@ -50,17 +51,27 @@ def _logger(mrctx = None, name = None, verbosity_level = None, printer = None):
...
-        verbosity_level = env_var_verbosity or verbosity_level
+        verbosity_level = env_var_verbosity or default_verbosity_level
 
     verbosity = {
         "DEBUG": 2,
+        "ERROR": -1,
+        "WARN": 0,
         "FAIL": -1,
         "INFO": 1,
         "TRACE": 3,
```
aignas, line 76:
```suggestion
        "DEBUG": 2,
        "ERROR": -1,
        "FAIL": -1,
        "INFO": 1,
        "TRACE": 3,
        "WARN": 0,
```

PR #3762, opened 2026-05-08
fix(pip): Pass --abi flag when downloading wheel for freethreaded python

meteorcloudy:
When `pip download` is used for fetching pip dependency, we need to pass `--abi` flag to fetch the correct wheel file for freethreaded python.

Essentially making https://github.com/google-ml-infra/rules_ml_toolchain/blob/1c2352bde159ee14cc5050a871c437622fe25e1f/py/python_init_pip.bzl#L69-L71 unnecessary.

### DIFF
--- a/python/private/pypi/parse_requirements.bzl
+++ b/python/private/pypi/parse_requirements.bzl
@@ -43,6 +43,8 @@ def parse_requirements(
         get_index_urls = None,
         evaluate_markers = None,
         extract_url_srcs = True,
+        download_only = False,
+        python_version = None,
         logger):
     """Get the requirements with platforms that the requirements apply to.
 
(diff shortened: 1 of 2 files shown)


python/private/pypi/parse_requirements.bzl
```diff
@@ -114,7 +116,14 @@ def parse_requirements(
...
                     reqs_with_env_markers.setdefault(requirement_line, []).append(plat)
-            options[plat] = pip_args
+            plat_args = list(pip_args)
+            if download_only and plat.endswith("freethreaded") and python_version:
+                if not any([a.startswith("--abi") for a in plat_args]):
+                    major, _, tail = python_version.partition(".")
+                    minor, _, _ = tail.partition(".")
+                    if major and minor:
+                        plat_args.append("--abi=cp{}{}t".format(major, minor))
+            options[plat] = plat_args
```
aignas, line 126: With 2.0.0 and above this might not be needed, because by default we are using the downloader to fetch the wheels and the freethreaded wheels should be supported by default.

meteorcloudy: Ah, nice, is experimental_index_url now enabled by default in 2.0.0? Is there any case we'll fallback to `pip download` where this flag is still needed?

aignas: Not that I can see.

meteorcloudy: OK, then closing this PR

PR #3766, opened 2026-05-10
feat(coverage): warn when bundled coverage tool has no wheel for requested python_version/platform

Syndic:
Addresses review feedback on #3764:

~~NOTE: Starlark unit tests have no way to capture stdout/stderr, so we can't directly assert that the warning is printed. I am not sure if the refactoring-for-testability I have considered would be appreciated, so this PR is a minimal approach. (I will also prepare an alternative PR that includes a small refactor and some tests, in case that is preferred.)~~ A logger has been added, and is used in testing.

Previously, when `configure_coverage_tool = True` was set but the bundled `coverage.py` wheel set had no entry for the requested (python_version, platform), `coverage_dep` returned None silently. The result was that `bazel coverage` produced empty per-test lcov files for `py_test` targets with no signal to the user that coverage was unconfigured.

Print a WARNING in that path so the misconfiguration is visible. Preserves the existing silent return for the windows branch, which is intentionally quiet because the upstream coverage wrapper does not support windows.

### DIFF
--- a/python/private/coverage_deps.bzl
+++ b/python/private/coverage_deps.bzl
@@ -17,6 +17,7 @@
 
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 load("@bazel_tools//tools/build_defs/repo:utils.bzl", "maybe")
+load("//python/private:repo_utils.bzl", "repo_utils")
 load("//python/private:version_label.bzl", "version_label")
 
 # START: maintained by 'bazel run //tools/private/update_deps:update_coverage_deps <version>'
@@ -166,7 +167,7 @@ _coverage_deps = {
 
 _coverage_patch = Label("//python/private:coverage.patch")
 
-def coverage_dep(name, python_version, platform, visibility):
+def coverage_dep(name, python_version, platform, visibility, logger = None):
     """Register a single coverage dependency based on the python version and platform.
 
     Args:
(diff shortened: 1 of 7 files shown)


python/private/coverage_deps.bzl
```diff
@@ -188,7 +188,15 @@ def coverage_dep(name, python_version, platform, visibility):
     url, sha256 = _coverage_deps.get(abi, {}).get(platform, (None, ""))
 
     if url == None:
-        # Some wheels are not present for some builds, so let's silently ignore those.
+        # buildifier: disable=print
+        print((
```
aignas: Use `logger.warn` with an optional logger passed in.

aignas (changes requested):

python/private/coverage_deps.bzl
aignas: Please also take a look how to pass that from the layers above. For bzlmod we'll have the module extension context, so it should be relatively easy. For macros for WORKSPACE we should pass in a mock context because we are executing it in a macro if I remember correctly.

aignas (approved):
Thanks! The general direction is good, when I have time I'll look at merging this.

aignas (approved):
Thank you!

honnix:
This seems to generate unrelated warnings for platforms we don't care about. Is possible to filter those out?

PR #3768, opened 2026-05-11
feat(pypi): add native local wheel override support to bzlmod pip.parse

meteorcloudy:
This feature allows developers to easily test locally built Python wheels (e.g., wheels built locally before running integration tests) without needing to publish them to an index or manually edit `requirements_lock.txt`.

This is a proper implementation of the local_wheel_dist_folder feature used in JAX/TF/XLA.

- https://github.com/jax-ml/jax/blob/d72694890a8dd1911435edc8f5732af0c58054f5/WORKSPACE#L50-L62
- https://github.com/google-ml-infra/rules_ml_toolchain/blob/11bbb592cb4d7cbc8ae07d6a4e5ad58ee38605f2/py/python_repo.bzl

### How to Use

In your `MODULE.bazel` file, configure `pip.parse` with the new `local_wheel_*` attributes:

```starlark

pip.parse(
    hub_name = "pip",
    python_version = "3.15",
    experimental_index_url = "https://pypi.org/simple",
    local_wheel_dir = "dist",
    local_wheel_pkgs = ["libtpu"],
)
```

### Details

* All child wheels in the directory belonging to the specified packages are scanned and selected to find the optimal wheel for the target platform.
* Discovery the workspace root via `@@//:MODULE.bazel`
* Effective only when `pip.parse` is evaluated by the root module.
* Effective only when `index_url` is used (which is now by default)


python/private/pypi/extension.bzl
```diff
@@ -758,6 +759,30 @@ is not required. Each hub is a separate resolution of pip dependencies. This
...
+If the path contains a wildcard `*` (e.g. `dist/libtpu-*.whl`), matching `.whl`
+candidates are discovered and the newest version is selected by comparing
+basenames lexicographically.
+
+If a file is missing on disk or no files match the wildcard pattern, the
+override is silently ignored and Bazel falls back to the remote PyPI index.
+
+Overrides apply when bazel downloader is used and only take effect in the root module.
+
+:::{versionadded} 2.1.0
```
aignas:
```suggestion
:::{versionadded} VERSION_NEXT_FEATURE
```

tests/support/mocks/mocks.bzl
```diff
@@ -11,12 +11,60 @@ def _path_new(path, mock_files = None):
         {type}`MockPath` A struct mocking a path object.
     """
     mock_files = mock_files or {}
-    return struct(
-        exists = path in mock_files,
-        basename = path.split("/")[-1],
-        dirname = "/".join(path.split("/")[:-1]),
-        _path = path,
-    )
+    path_str = str(path)
```
aignas, line 19: This is great to have, but having a few comments would be good.

python/private/pypi/hub_builder.bzl
```diff
@@ -696,3 +712,61 @@ def _use_downloader(self, python_version, whl_name):
         normalize_name(whl_name),
         self._get_index_urls.get(python_version) != None,
     )
+
+def _collect_local_wheels(module_ctx, pip_attr, is_root = False):
```
aignas: Please add a docstring with typing information

python/private/pypi/hub_builder.bzl
```diff
@@ -681,9 +689,17 @@ def _whl_repo(
     # targets to each hub for each extra combination and solve this more cleanly as opposed to
     # duplicating whl_library repositories.
     target_platforms = src.target_platforms if is_multiple_versions else []
+    repo_name = whl_repo_name(src.filename, src.sha256, *target_platforms)
+
+    if local_wheel:
+        repo_name += "_local_override"
+        path_str = local_wheel._path if hasattr(local_wheel, "_path") else str(local_wheel)
+        args["urls"] = ["file://" + path_str]
```
aignas: Nit: Is this `path_str` absolute? We could as well use `whl_file` attribute and pass a label if we were ever able to construct one.

Note to myself - because we are not getting this from `pypi_cache.bzl`, this will never go into the lock file, which means that whether it is absolute or not does not matter that much.

python/private/pypi/hub_builder.bzl
```diff
@@ -696,3 +712,61 @@ def _use_downloader(self, python_version, whl_name):
...
+            if target_dir.exists:
+                candidates = target_dir.readdir()
+            else:
+                candidates = []
+
+            for candidate in candidates:
+                if not candidate.basename.endswith(".whl"):
+                    continue
+                if _wildcard_match(candidate.basename, pattern):
+                    if not matched_wheel or matched_wheel.basename < candidate.basename:
```
aignas: This relies on lexicographic sorting, but it may be better to:
1. parse the wheel name (with //python/private:parse_whl_name.bzl or similar)
2. get the version component
3. parse the version component (with //python/private:version.bzl)
4. Do the version comparison.

What do we do if we match multiple abis, platforms? Should we use the `select_whl` function from `parse_requirements.bzl` to actually match the right wheel based on target platform?

Maybe the right way would be to actually override the `get_index` function in case it is a root module to replace the wheels found on the internet with the wheels found locally and then let the wheel selection algorithm do the right thing instead of doing the wild card matching here and this comparison.

aignas (changes requested):
I have some concerns about this short-circuiting the wheel selection algorithm used in `parse_requirements` so left a few ideas.

tests/support/mocks/mocks.bzl
meteorcloudy: Done

python/private/pypi/hub_builder.bzl
meteorcloudy: Thanks for the feedback. I've replaced `local_wheels` with `local_wheel_dir` and `local_wheel_pkgs` and replies on the wheel selection to find the right wheel file. PTAL.

Issue #2950, opened 2025-06-01
chore: cleanup WORKSPACE


aignas:
This is a task related to eventual cleanup of WORKSPACE support.

Q: when should that happen? I personally think that having that done when bazel 9 is released and a stable bazel version no longer has `WORKSPACE` is a good time.

The following is a list of random things that we would need to do:
- [ ] Remove entry_points in `whl_library`.
- [ ] `IS_BZLMOD_ENABLED` has to go.
- [ ] `TOOL_VERSIONS` has to probably go and we can migrate the config to `MODULE.bazel` instead.

PR #3780, opened 2026-05-16
chore!: drop WORKSPACE support from rules_python

aignas:
All of the bazel versions that we support support `bzlmod` and the support for
`WORKSPACE` is being dropped in other rulesets as well. At some point new
versions of our dependencies will make it more and more difficult to carry
support for both, so in order to make the codebase easier to maintain, let's
drop the support.

Regarding whether we should call the next version 2.1 or 3.0, I am not sure.
For `bzlmod` users nothing is changing, so Semver is respected, whereas for
`WORKSPACE` users, `rules_python` cannot be used anymore, it's as if 2.1 no
longer exists for `WORKSPACE` users, so I think it is fine to call the release
2.1.

Summary:
- Remove examples.
- Remove docs.
- Start dropping symbols and support as well.

NOTE: this may need to wait until we discuss out what we are doing with our downstream users:
- https://github.com/protocolbuffers/protobuf/blob/main/WORKSPACE
- add more

Work towards #2950


gazelle/python/testdata/from_imports/import_nested_var/__init__.py
```diff
@@ -13,8 +13,4 @@
 # limitations under the License.
 
 # baz is a variable in foo/bar/baz.py
-from foo\
-    .bar.\
-        baz import (
-            baz
-        )
+from foo.bar.baz import baz
```
dougthor42: Looks like autoformatting snuck in.

Either wrap in `# fmt: off` `# fmt: on` comments, or tell formatter to ignore this file. This one is intentionally asserting some python AST parsing.

aignas:
The `toolchain_target_settings_test` fails on Bazel 7.7.0 because the `config.add_transition_setting` + `apply_config_settings_attr` interaction relies on label string formatting that is inconsistent across Bazel versions.

**Root cause:**

Bazel 8.0 release notes state:

> [Incompatible] All labels in Bazel error messages, log output, Build Event Protocol, etc. are now prefixed with double-at (`@@`) instead of single-at (`@`) where applicable, to properly denote that they contain canonical repo names.

The flow:

1. `config.add_transition_setting(setting = "//:family")` calls `str(tag.setting)` in `config.bzl`
2. That string lands in `EXTRA_TRANSITION_SETTINGS` -> `TRANSITION_LABELS` -> transition `inputs`/`outputs`
3. `apply_config_settings_attr` calls `str(key)` on the `Label` from `config_settings = {"//:family": "custom"}`
4. The transition output key from step 3 must **exactly match** the input key from step 2

On Bazel 7, `str(Label(...))` produces single-`@` labels (e.g. `@//:family`), but `str(tag.setting)` for `attr.label` in a module extension tag can produce a different form (e.g. `@module_under_test//:family`). This mismatch means the transition output key does not match any input key -> the `family` flag stays at its default (`"prebuilt"`) -> the `is_prebuilt` config_setting matches -> the gated default toolchain resolves -> the build succeeds when it should fail.

Bazel 8 fixed this with the `@@` canonicalization change, making label strings consistent across all APIs. This is tracked by the umbrella issue [bazelbuild/bazel#11128](https://github.com/bazelbuild/bazel/issues/11128) "Support workspace label prefixes in Starlark config and platform APIs", which encompassed related sub-issues (#9065, #10152, #10499, #10927).

**Fix:** Restrict `toolchain_target_settings_test` to `bazel_versions = ["8.5.1", "9.1.0"]` where cross-repo transition labels resolve consistently.

aignas:
Sorry for the LLM analysis, but I was a little baffled why this started failing now. The rest has been crafted with Artisanal Intelligence.

rickeylev:
> at some point our deps will drop it and it will be difficult to support

Is there such a case today? Or a dependency we want to use, but it's bzlmod only so using it is tough. I would find that very compelling.

Because Bazel 8 supports workspace, and [Bazel 8 is in support until Dec 2027](https://bazel.build/release#support-matrix), support for workspace is still technically around for another 1.5 years.

But yeah, the oldest bazel version supports bzlmod now. And AFAIK, Bazel 7 bzlmod is decent; decent enough one can use bzlmod with Bazel 7.

> 2.1 or 3.0

Eh, mild preference for 3.0 because we aren't being forced to drop support due to a dependency. I'm fine either way, though.

> may need to wait until we discuss what we're doing with downstream users

I asked the protobuf people what plans they have. If they respond back, I'll post back.

aignas:
If we agree that this is a 3.0 change, then let's look at the other things we should do for 3.0 and do it at once. I am OK with it being a 3.0 thing.

rickeylev:
In talking with the protobuf people, they were somewhat alarmed that rules_jvm_external and the others mentioned in the [slack thread](https://bazelbuild.slack.com/archives/CDCMRLS23/p1778690322572269) had dropped workspace support already.

Protobuf said they plan to drop workspace Q1 2027, so basically just after Bazel 7 is out of support (but before bazel 8 is out of support).

aignas:
> they were somewhat alarmed that rules_jvm_external and the others mentioned

Let's see how it plays out. I don't want to drop `WORKSPACE` just to make our lives more painful and force users to not be able to update to `2.1` and above.

dougthor42:
+1 to holding off until _at least_ the protobuf team drops WORKSPACE support, if not until Bazel 8 is out of support. Even though I'd love to have it gone, haha.

+1 to dropping WORKSPACE being a 3.0 change.

PR #3785, opened 2026-05-16
feat(pypi): support importing uv.lock file

aignas:
The strategy for this is:
* First add a way for us to create a `uv.lock` file from the `lock` rule.
* Then add a `uv.lock` reader via the bazel toml parser.
* Then plug the code into the `parse_requirements` function so that we
  can reuse the most of code already there.
* Add some sample docs for the `uv.lock`.

Extra things that we could do:

* Call the PyPI index to understand if the packages are yanked or not - lock
  file does not have that information.
* Read the `pyproject.toml` file to get the index values for each package.
* Add e2e tests from `rules_py` test suite.

Closes #3557
Work towards #2787


aignas:
Addressed all review comments. Please re-review when you get a chance. @gemini-code-assist

aignas:
Addressed all review comments and added Python 3.14 toolchain + integration test. Please re-review. @gemini-code-assist

jvolkman:
The [latest release](https://github.com/jvolkman/toml.bzl/releases/tag/v0.4.0) of toml.bzl makes bazel_lib a dev dependency. It now uses skylib for the `bzl_library` rules.

python/private/pypi/extension.bzl
```diff
@@ -818,6 +820,12 @@ a string `"{os}_{arch}"` as the value here. You could also use `"{os}_{arch}_fre
 
 :::{versionadded} 1.8.0
 :::
+""",
+        ),
+        "uv_lock": attr.label(
+            doc = """\
+(label, optional): A label pointing to the uv.lock file. If provided,
```
aignas, line 917: TODO: use a style similar to other docstrings

python/private/pypi/hub_builder.bzl
```diff
@@ -506,10 +506,12 @@ def _create_whl_repos(
             ),
             logger = logger,
         ),
+        uv_lock = pip_attr.uv_lock,
```
aignas, line 506: Add a test to ensure this is tested.

MODULE.bazel
```diff
@@ -51,9 +52,13 @@ python.defaults(
 python.toolchain(
     python_version = "3.11",
 )
+python.toolchain(
+    python_version = "3.14",
+)
```
aignas: rm

MODULE.bazel
```diff
@@ -51,9 +52,13 @@ python.defaults(
...
 use_repo(
     python,
     "python_3_11",
+    "python_3_14",
```
aignas: rm

python/uv/private/lock.bzl
```diff
@@ -253,9 +276,15 @@ def _lock_run_impl(ctx):
         return shell.quote(arg.replace("/", path_sep))
 
     info = ctx.attr.lock[_RunLockInfo]
+
+    if ctx.attr.output.endswith(".lock"):
```
rickeylev: is the `.lock` suffix the right thing to check for, or should it be `uv.lock` ? idk what the convention is for if one has to have multiple uv lock files.

python/uv/private/lock.bzl
```diff
@@ -253,9 +276,15 @@ def _lock_run_impl(ctx):
...
+        template = ctx.files._uv_lock_template[0]
+    else:
+        template = ctx.files._template[0]
```
rickeylev: nit: rename `_template` so it has a name that represents its purpose (pip_lock_template?)

python/private/pypi/parse_requirements.bzl
```diff
@@ -64,28 +66,224 @@ def parse_requirements(
             requirements line.
         extract_url_srcs: A boolean to enable extracting URLs from requirement
             lines to enable using bazel downloader.
+        uv_lock: {type}`str | None` an optional label/file path to the uv.lock
+            file. The ctx.read function will be used to read the contents.
+            If provided, the function will use the uv.lock file as the primary
+            source for package metadata and perform a consistency check against
```
rickeylev, line 69: Why do a consistency check?

python/private/pypi/parse_requirements.bzl
```diff
@@ -64,28 +66,224 @@ def parse_requirements(
...
-           be passed to {obj}`whl_library`.
-         * `whls`: {type}`list[struct]` The list of whl entries that can be
-           downloaded using the bazel downloader.
-         * `sdist`: {type}`list[struct]` The sdist that can be downloaded using
-           the bazel downloader.
-
-        The second element is extra_pip_args should be passed to `whl_library`.
+         * `is_multiple_versions`: {type}`bool` `True` if multiple versions have been
+           specified for this package.
+         * `index_url`: {type}`str` The index URL used to download the package.
```
rickeylev, line 82: Is there just one possible index url?

python/private/pypi/parse_requirements.bzl
```diff
@@ -64,28 +66,224 @@ def parse_requirements(
...
+
+def _get_all_platforms(requirements_by_platform):
+    """Get the set of all platform names from requirements_by_platform."""
+    all_platforms = {}
+    for plats in requirements_by_platform.values():
+        for p in plats:
+            all_platforms[p] = None
+    return sorted(all_platforms)
+
+def _parse_requirements_with_uv_lock(
```
rickeylev, line 118: Why does this function have the uv_lock arg as optiona, and fall back to requirements parsing? Its name would indicate it is for parsing uv lock files

python/private/pypi/parse_requirements.bzl
```diff
@@ -64,28 +66,224 @@ def parse_requirements(
...
+            _, _, filename = url.rpartition("/")
+            key = (filename, sha256)
+            if key not in seen:
+                seen[key] = None
+                entry["src_entries"].append(struct(
+                    version = version,
+                    sha256 = sha256,
+                    url = url,
+                    filename = filename,
+                    # NOTE @aignas 2026-05-17: we don't know if it is yanked because we are not
```
rickeylev: Does uv not handle filtering out yanked versions for us?
Though I suppose a version is likely yanked after uv.lock is generated.

What does pypi do if you try to download a yanked version? error or give it to you?

python/private/pypi/parse_requirements.bzl
```diff
@@ -64,28 +66,224 @@ def parse_requirements(
...
+            url = sdist["url"]
+            _, _, filename = url.rpartition("/")
+            entry["src_entries"].append(struct(
+                version = version,
+                sha256 = sha256,
+                url = url,
+                filename = filename,
+                yanked = None,
+            ))
+        elif pkg.get("source", {}).get("git"):
```
rickeylev: Do file:// urls need special handling?

python/private/pypi/parse_requirements.bzl
```diff
@@ -97,14 +295,8 @@ def parse_requirements(
         logger.trace(lambda: "Using {} for {}".format(file, plats))
         contents = ctx.read(file)
 
-        # Parse the requirements file directly in starlark to get the information
-        # needed for the whl_library declarations later.
         parse_result = parse_requirements_txt(contents)
 
-        # Save parsed results from ALL files, even those with no matching
```
rickeylev, line 97: This comment seems relevant still?

Many of the removed comments here seem relevant

python/private/py_repositories.bzl
```diff
@@ -72,6 +72,12 @@ def py_repositories(transition_settings = []):
         strip_prefix = "rules_cc-0.1.5",
         urls = ["https://github.com/bazelbuild/rules_cc/releases/download/0.1.5/rules_cc-0.1.5.tar.gz"],
     )
+    http_archive(
```
rickeylev, line 75: fwiw, i think its fine to make it bzlmod only. I thought the uv.lock rule was already bzlmod only anyways?

rickeylev (approved):
Overall LGTM. Some minor questions and nits. But overall, this fit in rather nicely with the pipstar code you've written, nice!

python/uv/private/lock.bzl
aignas: Yeah, let's go for a safer option and do `uv.lock`.

python/private/pypi/parse_requirements.bzl
aignas: Stale comment, will remove.

python/private/pypi/parse_requirements.bzl
aignas: Yes, only one default, others are `extra`.

python/private/py_repositories.bzl
aignas: This is only to make the loads work easier. The `parse_requirements` is used in `WORKSPACE` as well.

rickeylev:
(agent comment) Addressed the second review cycle comments (enforcing `uv.lock` suffix, renaming `_template` attribute, restoring requirement parsing comments, and registering missing Stardoc TOML dependencies). Please re-review @gemini-code-assist @rickeylev

rickeylev:
Addressed the remaining unresolved review threads (adopting `str.removeprefix()`, syncing `toml.bzl` to 0.4.1 in `py_repositories.bzl`, adding absolute path check in `lock.bzl`, formatting `uv_lock` docstrings with `VERSION_NEXT_FEATURE`, and removing temporary Python 3.14 toolchains). Please re-review @gemini-code-assist @rickeylev

python/private/pypi/parse_requirements.bzl
rickeylev: Gemini says: when file:// are used, it shows up as `pkg["source"]["url"]`

Added a branch and helper function here to handle that.

python/private/py_repositories.bzl
```diff
@@ -72,6 +72,12 @@ def py_repositories(transition_settings = []):
...
+        name = "toml.bzl",
+        sha256 = "63633c762bdb4d836add5a9a81deeeae93c4b2edbd62ac032733020b5652b90a",
+        strip_prefix = "toml.bzl-0.4.0",
+        url = "https://github.com/jvolkman/toml.bzl/releases/download/v0.4.0/toml.bzl-v0.4.0.tar.gz",
+    )
```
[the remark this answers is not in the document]
rickeylev: workspace config updated to use 0.4.1

python/uv/private/lock.bzl
```diff
@@ -84,40 +84,75 @@ def _lock_impl(ctx):
...
+        progress_message = "Creating a requirements.txt with uv: %{label}"
+
+    if is_uv_lock:
+        src_dir = ctx.files.srcs[0].dirname if ctx.files.srcs else "."
+        python_path = getattr(python, "path", python)
+        command = 'export UV_PYTHON_PATH="$(pwd)/{python}" && "$@" --python "$UV_PYTHON_PATH" && cp "{src_dir}/uv.lock" "{output}"'.format(
+            python = python_path,
+            src_dir = src_dir,
+            output = output.path,
+        )
```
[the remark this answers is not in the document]
rickeylev: Fixed

python/private/pypi/extension.bzl
rickeylev: Fixed.

python/private/pypi/parse_requirements.bzl
rickeylev: Stale comment removed

rickeylev (commented):
Fixed a variety of comments

rickeylev:
@aignas PTAL -- CI is happy! But the unresolved comments from Gemini I think might be valid?

aignas:
test this please

aignas:
It seems that the extras are defined in a different way than `requirements.txt` files and it means that we should handle it slightly differently from what we have today. The good news is that it will benefit us in many other ways, but the drawback is that it makes the uv.lock implementation a little bit more work.

The gist is that each `package` in the lock file has its dependencies encoded per extra. The good news is that we can just feed this dependency tree to the hub repo, then do the cycles resolution and then we don't need the `requirements_cycles` definition. #1975 had code how we could deal with this sort of stuff.

Since the dep graph is coming from `uv.lock` itself, we have the following options:
- Generate the requirements lines based on what extras we are using in the uv.lock. This would allow us to easily plug into what we already have. However, implementing that is almost the same as export a `uv.lock` file to `requirements.txt` file. Which I assume may have some challenges. We could just invoke the `uv` in the repo phase to do this.
- Generate one `whl_library` per Python dependency. Then we would need to do some smart stuff in order to handle cycles in the hub repository, but the main idea would be to actually have the dependency graph in the hub repository so that we can correctly select the minimum set of `whl_library` targets that have to be present here.

We could already do this if we read the `whl_metadata` from the SimpleAPI and pulled the `RequirementsDist` values for each package. That way we would not need to depend on the `uv.lock` format (which could change many times in the future). That would also mean that we could work the same way if `py.lock` file was used with `pip.parse`.

So this `uv.lock` file consumption only works if we:
* Don't have extras usage in the lock file
* convert the `uv.lock` file into `requirements.txt` in starlark or execute uv itself.

rickeylev:
ai analysis of ci failure: path instead short_path should be used.

I think that's right? The logic has moved around a bit, but it looks like the is_windows branch of building the args previously did e.g. `hasattr(arg, "path")` and now it uses short_path

rickeylev:
CI is happy again

python/private/pypi/parse_requirements.bzl
```diff
@@ -59,28 +62,392 @@ def parse_requirements(
...
+            whl_abi_tags = platform.whl_abi_tags,
+            implementation_name = platform.env.get("implementation_name", "cpython"),
+            limit = 1,
+            logger = logger,
+        )
+        if matched != None:
+            result.append(p)
+    return result
+
+def _resolve_version_wheels(src_entries, distribution, extra_pip_args, extra_str, platforms, logger):
```
aignas: We should use the `select_whl` function instead of duplicating here.

rickeylev: i.e. remove this function entirely and use select_whl() instead?

aignas: At least from the design point of view that is why the `select_whl` was created.

rickeylev: told slopbot to remove that function and use select_whl instead. how's this look?

aignas:
I think the last thing is to ensure that the code does not assume that platforms can be nothing. The code that would call the parse_requirements.bzl entry point would pass the platforms in any case, so we can simplify the code a little bit here. However, I see that we need to repay a little debt in the test setup department.

PR #3786, opened 2026-05-17
fix(gazelle): replace cgo tree-sitter binding with gotreesitter

odvcencio:
Potential fix for #1913.

The Gazelle Python parser currently depends on `github.com/smacker/go-tree-sitter`, which brings in the C tree-sitter runtime through cgo. That makes the Gazelle binary harder to cross-compile, especially for Darwin targets, even though the parser is only used as an implementation detail of the Python extension.

This replaces that binding with `github.com/odvcencio/gotreesitter` v0.19.1, a pure-Go tree-sitter-compatible parser/runtime, and updates the Gazelle dependency metadata to use the generated Python grammar from gotreesitter. The parser behavior stays tree-sitter based; the before/after difference is that Gazelle no longer needs the cgo tree-sitter dependency for Python source parsing.

The implementation uses a shared gotreesitter parser pool and releases parsed trees after Gazelle extracts imports/comments/main metadata, so repeated package scans do not retain parser arenas longer than needed.

No CHANGELOG entry is included because this is an implementation dependency change for the Gazelle extension rather than a new user-facing API.

Validation:

- `bazel run //:gazelle_update_repos`
- `bazel test //python:all --test_output=errors`
- `bazel build --platforms=@io_bazel_rules_go//go/toolchain:darwin_arm64 //python:gazelle_binary`

Local performance check:

- On gotreesitter's top-50 Python smoke benchmark, gotreesitter v0.19.1 is in the same range as the C runtime: gotreesitter ~6.0-8.1 us/op vs C ~6.6-7.3 us/op in a Docker cgo harness run.
- On gotreesitter's synthetic larger Python full-parse benchmark, v0.19.1 is still slow: roughly 9.9-10.2 ms/op, about 1.9 MB/s on this machine. This should remain draft while larger-file Python parser/runtime performance is evaluated.


odvcencio:
Updated the Gazelle gotreesitter pin to v0.18.0, including `gazelle/deps.bzl`.\n\nValidation run locally:\n- `bazel run //:gazelle_update_repos`\n- `bazel test //python:default_test`\n\nNote: `--config=fast-tests` is not defined in the Gazelle submodule, so the focused Bazel commands were run without that config.

PR #3795, opened 2026-05-22
feat: expose interpreter files-to-run on PyRuntimeInfo

jsun-splunk:
Rules that execute a py_runtime interpreter in an action need the interpreter executable together with its runfiles metadata. The existing PyRuntimeInfo fields identify the interpreter file and runtime files, but do not preserve the target's FilesToRunProvider for executable interpreter targets.

Add PyRuntimeInfo.interpreter_files_to_run for runtimes created from an executable interpreter target, and validate that direct provider construction keeps the FilesToRunProvider executable aligned with the interpreter field. Direct file interpreters and platform runtimes continue to leave this field unset, preserving existing py_runtime behavior.

Document the new public provider field and add focused analysis-test coverage for executable, file-only, platform, and invalid constructor cases.

One example is [rules_pycross](https://github.com/jvolkman/rules_pycross) wheel building with an in-build executable Python runtime. pycross needs to run the selected Python interpreter in an action, but PyRuntimeInfo previously exposed only the interpreter File and runtime files, not the interpreter target's FilesToRunProvider. For executable in-build runtimes, the FilesToRunProvider is the Bazel-native handle that carries both the executable and the runfiles metadata needed to stage it as an action tool. Exposing it lets pycross consume the runtime interpreter directly for wheel-build actions instead of reconstructing or approximating the interpreter's runtime closure from separate fields.


jsun-splunk:
@rickeylev, @aignas - any chance of a review?

python/private/py_runtime_rule.bzl
```diff
@@ -39,25 +39,44 @@ def _py_runtime_impl(ctx):
...
         interpreter_di = interpreter[DefaultInfo]
-
-        if interpreter_di.files_to_run and interpreter_di.files_to_run.executable:
+        interpreter_file = None
+
+        # Direct file targets also expose files_to_run.executable. They should
+        # keep py_runtime's file-only behavior and not populate
+        # interpreter_files_to_run. Rule targets have OutputGroupInfo; direct
+        # file targets do not.
+        is_file_target = OutputGroupInfo not in interpreter
```
matt-sm: Could this key off `interpreter_file.is_source` instead of the `OutputGroupInfo` check? The case you're guarding against is really a source file — that's the only non-executable input with a `files_to_run.executable` — and `is_source` doesnt rely on which providers Bazel attaches to rule vs
file targets.

python/private/py_runtime_rule.bzl
```diff
@@ -39,25 +39,44 @@ def _py_runtime_impl(ctx):
...
+        if _is_singleton_depset(interpreter_di.files):
+            interpreter_file = interpreter_di.files.to_list()[0]
+
+        if is_file_target and interpreter_file:
```
matt-sm: Is it worth noting somewhere that source-file interpreter now takes this path instead of the executable branch?

eg. py_runtime(interpreter = ...) pointing at a downloaded/extracted file @python_x86_64//:bin/python3 rather than a *_binary 

This used to hit the executable branch and pick up the runfiles.merge/runtime_files expansion, and now skip it.

python/private/py_runtime_rule.bzl
jsun-splunk: you are correct, .is_source is the better choice.

python/private/py_runtime_rule.bzl
jsun-splunk: You caught an unintended behavioural change. I've restructured the if statement so source-file interpreter goes through the old route that collects the runfiles too.

python/private/py_runtime_rule.bzl
```diff
@@ -39,25 +39,46 @@ def _py_runtime_impl(ctx):
...
             fail("interpreter_path must be an absolute path")
     else:
         interpreter_di = interpreter[DefaultInfo]
+        interpreter_file = None
 
-        if interpreter_di.files_to_run and interpreter_di.files_to_run.executable:
+        if _is_singleton_depset(interpreter_di.files):
+            interpreter_file = interpreter_di.files.to_list()[0]
+
+        if interpreter_file and interpreter_file.is_source:
```
matt-sm: This now treats any singleton source file output as the direct-source-file path. That fixes direct file labels, but it may also include non- executable wrapper targets/filegroups whose single output is a source file. 

Those used to fall through to the singleton-file branch without merging default runfiles into `runtime_files`; now their default runfiles get folded in.

Should this branch also require `interpreter_di.files_to_run.executable == interpreter_file`, so it only catches the direct file-label case? Or should we add a test for new behaviour?

jsun-splunk: I do think the old path made more sense. Extra files in the singleton scenario can be added with `py_runtime(files = [...])`. So I will add the suggested guard.

I've also added a couple of analysis test to keep that behaviour too.

python/private/py_runtime_info.bzl
```diff
@@ -239,6 +250,19 @@ The Python implementation name (`sys.implementation.name`)
 If this is an in-build runtime, this field is a `File` representing the
 interpreter. Otherwise, this is `None`. Note that an in-build runtime can use
 either a prebuilt, checked-in interpreter or an interpreter built from source.
+""",
+        "interpreter_files_to_run": """
+:type: None | FilesToRunProvider
```
matt-sm, line 255: Storing FilesToRunProvider as a field inside PyRuntimeInfo is unusual — providers are normally returned in parallel, not nested within each other.

Would it be cleaner for py_runtime to return both providers as siblings:

```
return [PyRuntimeInfo(...), FilesToRunProvider(...)]
```

and have py_toolchain explicitly thread both through ToolchainInfo:

```
ToolchainInfo(
    py_runtime = ctx.attr.runtime[PyRuntimeInfo],
    interpreter_files_to_run = ctx.attr.runtime[FilesToRunProvider],
)
```
so consumers access them as peers rather than reaching into PyRuntimeInfo for a nested provider object?

jsun-splunk: FilesToRunProvider is a bit special. It cannot be created directly with `FilesToRunProvider(...)`, it is implicitly created with a executable target for bazel and accessed through `DefaultInfo.files_to_run`. So you can't do something like:

```
return [PyRuntimeInfo(...), ctx.attr.tool[DefaultInfo].files_to_run]`
```

If you tried that you get an error like:

```
at index 1 of result of rule implementation function, got element of type FilesToRunProvider, want Info
```

Also, the primary API for consuming python toolchain is through the PyRuntimeInfo provider. Deviating from it means every consumer will need to add abstraction details to deal with the deviation..

matt-sm (approved):

python/private/py_runtime_rule.bzl
```diff
@@ -111,6 +140,7 @@ def _py_runtime_impl(ctx):
     py_runtime_info_kwargs = dict(
         interpreter_path = interpreter_path or None,
         interpreter = interpreter,
+        interpreter_files_to_run = interpreter_files_to_run,
```
rickeylev: Pass this in the py_runtime_info_kwargs.update() call on L155. This first construction are args only passed to the legacy builtin PyRuntimeInfo object

Actually, because of the pop on L152, this field isn't being passed anywhere.

rickeylev (approved):
Thanks for this! It's bugged me that a FilesToRun-providing target could be fed into py_runtime, but it wasn't well preserved after passing through it.

Mostly LGTM. The interpreter_files_to_run isn't being passed into PyRuntimeInfo from py_runtime. Please fix (a test to verify that would be nice, too, but not going to block on that).

For my own edification, how might the code in https://github.com/bazel-contrib/rules_python/blob/main/python/private/common.bzl#L643 (`_actions_run`, the `if PyInterpreterProgramInfo ...` code path) be updated to use this? (that some internal helper code we have for running via an action) If I'm understanding right, something like this?

```
exec_runtime = ctx.toolchains[EXEC_TOOLS_TOOLCHAIN_TYPE].exec_tools.exec_runtime
...
ctx.actions.run(
  executable = [exec_runtime.interpreter_files_to_run],
  ...
)
```

Or it could be put into `tools` for a more complex case.

jsun-splunk:
> Thanks for this! It's bugged me that a FilesToRun-providing target could be fed into py_runtime, but it wasn't well preserved after passing through it.
> 
> Mostly LGTM. The interpreter_files_to_run isn't being passed into PyRuntimeInfo from py_runtime. Please fix (a test to verify that would be nice, too, but not going to block on that).
> 
> For my own edification, how might the code in https://github.com/bazel-contrib/rules_python/blob/main/python/private/common.bzl#L643 (`_actions_run`, the `if PyInterpreterProgramInfo ...` code path) be updated to use this? (that some internal helper code we have for running via an action) If I'm understanding right, something like this?
> 
> ```
> exec_runtime = ctx.toolchains[EXEC_TOOLS_TOOLCHAIN_TYPE].exec_tools.exec_runtime
> ...
> ctx.actions.run(
>   executable = [exec_runtime.interpreter_files_to_run],
>   ...
> )
> ```
> 
> Or it could be put into `tools` for a more complex case.

You are correct, except you don't need to wrap files_to_run in a list. So it will be

```
ctx.actions.run(
  executable = exec_runtime.interpreter_files_to_run,
  ...
)
```

adding the same to `tools` will make the executable and its runfiles available in the action space too.

python/private/py_runtime_rule.bzl
jsun-splunk: Thanks for reviewing.

Its actually the other way around, L152 pops it from `builtin_py_runtime_info_kwargs` while keeps in `py_runtime_info_kwargs` for PyRuntimeInfo. But the pop was silly, adding it in L155 will have the same effect.

I've updated it now, see https://github.com/bazel-contrib/rules_python/compare/8b47e8d4f1bf96bec756e430552e5fd35fbdc902..2d2912de3543c841a64203e82298d4f50c305c89.

rickeylev: Ahhh, I must have read that line a dozen times and didn't notice the pop was on the other dict. Thanks. LGTM

PR #3797, opened 2026-05-27
fix: avoid ln race condition in bootstrap script

rickeylev:
Avoid startup failures when multiple processes attempt to create the same symlinks simultaneously.

Ensure symlinks are created and ignore errors if they already exist, which can happen in a race condition during startup.


rickeylev:
Argh, somehow it picked the wrong base commit or something. I gotta go futz with this...

rickeylev:
Alright, fixed the commits. Ready for review

aignas (approved):

PR #3798, opened 2026-05-27
chore: use direct @dev_pip targets instead of requirement()

rickeylev:
Use @dev_pip//<pkg> format instead of the requirement() macro. This aligns with modern bzlmod patterns and simplifies the BUILD file.


rickeylev:
I pulled this out of your uv.lock PR. Those requirement() calls have bugged me for awhile, so i figured a good time to kill them.

aignas (approved):

PR #3800, opened 2026-05-29
fix(uv): allow environment setting for the update action

### LINKED ISSUE #3405: Support authentication in lock rule in sandbox
Consider this feature.

There is prior art to passing around credentials in https://github.com/astral-sh/uv/issues/1520

Example from a Slack thread

> Write auth information to a file, say /tmp/auth
> Set --sandbox_add_mount_pair=/tmp/auth:/tmp/auth

aignas:
With this PR we allow users to set extra environment variables using `.bazelrc`
and to ensure that we can pass UV extra parameters. This should enable users use
this with private indexes in a `diff_test` usage scenario.

Whilst at it, add integration tests to actually verify that passing works via
env vars. Whilst at it also fix the Windows support for the `uv` lock rule.

Fixes #3405


docs/pypi/download.md
```diff
@@ -334,6 +334,26 @@ into whatever HTTP(S) request it performs against `example.com`.
 
 See the [Credential Helper Spec][cred-helper-spec] for more details.
 
+### Using a credential helper with the lock rule
```
rickeylev: FWIW, I'm +1 on changing the `bazel run //foo.lock` target to run uv directly so that it uses the user's permissions. For casual use, its more convenient than having to setup the mount pair etc

docs/pypi/download.md
```diff
@@ -334,6 +334,26 @@ into whatever HTTP(S) request it performs against `example.com`.
...
+
+The same credential helper pattern can be used with the {obj}`@rules_python//python/uv:lock.bzl`
+`lock` macro. Instead of embedding credentials in `UV_EXTRA_INDEX_URL`, set
+`UV_CREDENTIAL_HELPER` to point to a credential helper script and use
+`--sandbox_add_mount_pair` to mount it into the sandbox:
+
+```
+# .bazelrc
+build --sandbox_add_mount_pair=/path/to/cred_helper.sh=/cred_helper.sh
+build --action_env=UV_CREDENTIAL_HELPER=/cred_helper.sh
```
rickeylev: Saw a clever alternative: Use a custom workspace status to put an extra key which is the path to what you want. Then pass version_file to the action and parse out the value. The nominal benefit is it doesn't become part of the cache key.

docs/pypi/download.md
```diff
@@ -334,6 +334,26 @@ into whatever HTTP(S) request it performs against `example.com`.
...
+### Using a credential helper with the lock rule
+
+The same credential helper pattern can be used with the {obj}`@rules_python//python/uv:lock.bzl`
+`lock` macro. Instead of embedding credentials in `UV_EXTRA_INDEX_URL`, set
+`UV_CREDENTIAL_HELPER` to point to a credential helper script and use
+`--sandbox_add_mount_pair` to mount it into the sandbox:
+
+```
+# .bazelrc
+build --sandbox_add_mount_pair=/path/to/cred_helper.sh=/cred_helper.sh
```
rickeylev: Does the LHS have to be an absolute path? If its relative, what is it relative to? does it support e.e. %workspace% interpolation?

If you don't know the answer, thats OK. Mostly just for my own edification.

python/uv/private/lock.bzl
```diff
@@ -142,6 +142,7 @@ def _lock_impl(ctx):
             uv,
             python_files,
         ],
+        use_default_shell_env = True,
```
rickeylev, line 209: What makes this necessary? Please add comment

docs/pypi/download.md
```diff
@@ -334,6 +334,26 @@ into whatever HTTP(S) request it performs against `example.com`.
 
 See the [Credential Helper Spec][cred-helper-spec] for more details.
 
+### Using a credential helper with the lock rule
+
+The same credential helper pattern can be used with the {obj}`@rules_python//python/uv:lock.bzl`
+`lock` macro. Instead of embedding credentials in `UV_EXTRA_INDEX_URL`, set
+`UV_CREDENTIAL_HELPER` to point to a credential helper script and use
```
rickeylev: is UV_CREDENTIAL_HELPER part of uv itself? I don't see it in the uv docs, nor the rules_python code.

rickeylev (approved):

docs/pypi/download.md
aignas: I think it does not need to be an absolute path.

python/uv/private/lock.bzl
aignas:
```suggestion
        # User reported being unable to add `--action_env` and get it to work.
        # Without this flag.
        #
        # Ref: https://app.slack.com/client/TA4K1KQ87/CA306CEV6
        use_default_shell_env = True,
```

docs/pypi/download.md
aignas: hmmm I think it might be our own, I think I should review this once more.

docs/pypi/download.md
aignas: This is also interesting: https://docs.astral.sh/uv/concepts/authentication/cli/#using-credentials-with-external-tools

docs/pypi/download.md
aignas: I agree, but if you want a `diff_test` here, then everything runs as an action, so we would then need to mount the credential helper inside. Maybe the docs should be updated to reflect that better?

rickeylev: Ah doh, `//requirements.run` already works the way I described.

Actually, re-reading the uv docs you linked to, since uv has its own credential store, is what needs to be mounted in the sandbox the uv credential store? And then uv just has to be told where that is.

aignas:
OK, PTAL, I think this can be merged if there are no extra comments.

PR #3802, opened 2026-05-31
feat(toolchains): support runtime registration from manifest

rickeylev:
Currently, all supported Python runtime versions and their
platform-specific metadata (URLs, SHA256s, strip_prefix) must be
hardcoded in `python/versions.bzl`. This makes it slow and difficult
to adopt new Python versions or custom builds without updating
`rules_python` itself.

This PR introduces the ability to dynamically fetch and register
Python runtimes from a remote python-build-standalone (PBS) manifest
file (e.g., `SHA256SUMS`).

This is supported via two new attributes in `python.override`:
- `add_runtime_manifest_urls`: A list of URLs pointing to manifest
  files  to parse and register.
- `runtime_manifest_sha`: The SHA256 hash of the manifest file.

The manifest file format is the python-build-standalone SHA256SUMS
format (`SHA FILENAME`), extended to allow arbitrary URLs for the
filename (to allow more arbitrary locations).


docs/toolchains.md
```diff
@@ -310,6 +312,65 @@ Added support for custom platform names, `target_compatible_with`, and
...
+release manifest:
+
+```
+# File: MODULE.bazel
+python = use_extension("@rules_python//python/extensions:python.bzl", "python")
+python.override(
+    add_runtime_manifest_urls = [
+        "https://github.com/astral-sh/python-build-standalone/releases/download/20260414/SHA256SUMS",
+    ],
+    runtime_manifest_sha = "ce18fdfd47c66830a40ea9b9e314a14b1636bbfd684501bc5ca1fc6d55a7933f",
```
aignas, line 341: This is really good. Maybe we could dog-food the API and update the `python/versions.bzl`.

docs/toolchains.md
```diff
@@ -310,6 +312,65 @@ Added support for custom platform names, `target_compatible_with`, and
...
+In the example below, we register all runtimes available in a specific PBS
+release manifest:
+
+```
+# File: MODULE.bazel
+python = use_extension("@rules_python//python/extensions:python.bzl", "python")
+python.override(
+    add_runtime_manifest_urls = [
+        "https://github.com/astral-sh/python-build-standalone/releases/download/20260414/SHA256SUMS",
+    ],
```
aignas, line 339: I would love to support labels or files that users define themselves. That way they can be in the repo and we don't need to have a custom dictionary in the `python/versions.bzl` anymore, we could just have a list of files.

python/private/pbs_manifest.bzl
```diff
@@ -0,0 +1,116 @@
...
+    if plus:
+        python_version = left
+        build_version, sep, rest = tail.partition("-")
+        if not sep:
+            return None
+    else:
+        python_version, sep, rest = left.partition("-")
+        if not sep:
+            return None
+        build_version = ""
```
aignas, line 38: What about using our own PEP440 compliant parser?

The format `3.10.20+20260414` is compliant.


We could the use the version comparison function to do the ordering.

python/private/python.bzl
```diff
@@ -741,10 +742,71 @@ def _override_defaults(*overrides, modules, _fail = fail, default):
...
+            "riscv64-unknown-linux-gnu",
+            "s390x-unknown-linux-gnu",
+            "x86_64-apple-darwin",
+            "x86_64-pc-windows-msvc",
+            "x86_64-unknown-linux-gnu",
+            "x86_64-unknown-linux-musl",
+        ]
+        is_full = entry.flavor.endswith("-full")
+        if expects_full != is_full:
+            continue
```
aignas: If there is `install_only` and `full`, which one will we prefer? I think we should prefer `install_only` due to size.

python/private/python.bzl
```diff
@@ -741,10 +742,71 @@ def _override_defaults(*overrides, modules, _fail = fail, default):
...
+        else:
+            urls = ["{}/{}".format(base_url, location) for base_url in base_download_urls]
+
+        v_dict = available_versions.setdefault(py_version, {})
+        v_dict.setdefault("sha256", {})[matched_platform] = sha256
+        v_dict.setdefault("url", {})[matched_platform] = urls
+        if is_full:
+            v_dict.setdefault("strip_prefix", {})[matched_platform] = "python/install"
+        else:
+            v_dict.setdefault("strip_prefix", {})[matched_platform] = "python"
```
aignas: We should do elif here because this is `install_only` path. other types should be ignored.

aignas (commented):
Thank you. This was on my mind for some time, so thanks for taking a stab at it.

What do you think about doing the following:
* Registering the manifests we use in our `MODULE.bazel` file
* Changing the rule that prints the contents that we should paste to `python/versions.bzl` to instead update the file with what is defined in the `MODULE.bazel` file. This way we know the manifest files are interpreted in exactly the same way as we would normally do it.

docs/toolchains.md
rickeylev: Oh good idea. Adding a add_runtime_manifest_files label_list should be pretty easy.

python/private/pbs_manifest.bzl
rickeylev: Use the PEP440 parser on the whole file name, or just the X.Y.Z+build_version component?

Unless there's a python-build-standalone reference that says their naming format is PEP440 compliant, I'm -1 on using it for the whole filename; that seems just incidental.

Using it just for the "X.Y.Z+build_version" -- I guess so? So you mean return the full object/struct the PEP440 parser returns?

tests/python_bzlmod_ext/parse_sha_manifest_tests.bzl
```diff
@@ -0,0 +1,155 @@
...
+        "python_version": "3.10.20",
+        "vendor": "unknown",
+    })
+
+    # 3. Freethreaded
+    parsed3 = parse_filename("cpython-3.13.13+20260414-aarch64-apple-darwin-freethreaded+pgo+lto-full.tar.zst")
+    env.expect.that_dict(parsed3).contains_exactly({
+        "arch": "aarch64",
+        "build_version": "20260414",
+        "flavor": "freethreaded+pgo+lto-full",
```
rickeylev: note to self: freethreaded should be parsed out and captured as a separate field

python/private/python.bzl
rickeylev: Yeah, we should prefer install_only.

Long term, I'd like to add a build flag to capture the "flavor" dimension, but I haven't fully figured out if it should be further broken up into discrete fields (e.g. should lgo and pgo be separate fields? should "fullness" be a separate field?) etc

rickeylev (commented):
> what do you think of (1) register our manifests in module.bazel, (2) changing rule to update the manifest specified by (2)

SGTM. I wasn't sure how we should use this ourselves. I like your idea of allowing label() to point to a file.

python/private/python.bzl
rickeylev: Done

docs/toolchains.md
rickeylev: Are you ok doing this in a separate PR? It doesn't look too bad, but its enough of a change in how the default runtimes are registered i think it deserves its own change. [proof of concept](https://github.com/rickeylev/rules_python/compare/feat.toolchain.manifest...rickeylev:rules_python:register-builtin-runtimes-manifest?expand=1)

rickeylev:
PTAL

docs/toolchains.md
rickeylev: Yeah, lets defer this to another PR. Converting our "list" of runtimes into the manifest format is a bit more involved because of workspace mode and the TOOL_VERSIONS format. See https://github.com/bazel-contrib/rules_python/pull/3812 for the details

python/private/pbs_manifest.bzl
aignas: I would have said that we can use it for the whole version `X.Y.Z+build_version` only was my suggestion.

aignas (approved):

PR #3808, opened 2026-06-01
refactor(pypi): don't use Python to patch wheels

aignas:
1. Refactored patch_whl.bzl — Extracted fix_record_content() as a public
   pure function (no rctx dependency) so it can be unit tested directly.

2. Unit tests (patch_whl_tests.bzl) — 5 new rules_testing tests for fix_record_content:
   - Adds missing files to RECORD
   - Returns None when nothing is missing
   - Preserves quoting style of existing entries
   - Skips excluded files (INSTALLER, RECORD*, REQUESTED)
   - Skips .whl files

3. Integration test (patch_whl_patch_test.py) — End-to-end test that:
- Creates a wheel from testdata/pkg/ via whl_from_dir_repo
- Applies a patch via whl_library(whl_patches=...)
- Verifies pkg.PATCHED is True (was False) and pkg.DATA is unchanged

The integration test follows the established pattern: test data in
tests/pypi/patch_whl/testdata/, registered in internal_dev_deps.bzl,
wired into MODULE.bazel, and built as a py_test. All 205 pypi tests
pass.

With this we can potentially more easily separate whl and sdist path
ways.


python/private/pypi/patch_whl.bzl
```diff
@@ -12,27 +12,26 @@
...
-
-If we were silently regenerating the RECORD file, we may be vulnerable to supply chain
-attacks (it is a very small chance) and keeping the RECORD patches next to the
-other patches ensures that the users have overview on exactly what has changed
-within the wheel.
+"""A small utility to patch a file in the repository context natively.
+
+This replaces the previous approach that depended on running Python to
+repack patched wheels. Instead, after extracting and patching, we fix
+the RECORD file natively in Starlark and repack using system zip.
```
rickeylev, line 19: I forget -- why do we have to extract the wheel, patch it, rezip it, then extract it again later? Why not just extract and patch?

python/private/pypi/patch_whl.bzl
```diff
@@ -98,43 +88,154 @@ def patch_whl(rctx, *, python_interpreter, whl_path, patches, **kwargs):
...
+        op = "PatchWhl",
+        arguments = [
+            "zip",
+            "-0",
+            "-X",
+            str(output),
+            "-r",
+            ".",
+        ],
+    )
```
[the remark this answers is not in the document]
rickeylev: Pretty sure this isn't accurate. Before repacking, the original .whl file is deleted.

python/private/pypi/patch_whl.bzl
```diff
@@ -98,43 +88,154 @@ def patch_whl(rctx, *, python_interpreter, whl_path, patches, **kwargs):
...
-    record_patch = rctx.path("RECORD.patch")
+    _fix_record(rctx)
+
     whl_patched = patched_whl_name(whl_input.basename)
 
-    pypi_repo_utils.execute_checked(
-        rctx,
-        python = python_interpreter,
-        srcs = [
-            Label("//python/private/pypi:repack_whl.py"),
```
rickeylev, line 108: Can we delete repack_whl.py with this PR, too?

rickeylev (approved):

python/private/pypi/patch_whl.bzl
aignas: It was like that due to the `whl_installer.py` failing to install that. Now there is no such need anymore. I am wondering what we should do with the implicit API of:

> We will provide you with a patched `.whl` file to be consumed via the `filegroup` if needed.

Given that most of the usecases of patching are to make things work better in bazel, I am not sure there is a strong argument to keep that behaviour.

FrankPortman: I don't use this functionality but I use something "adjacent" to it (a bit hard to explain in a short comment 😅), but it sounds like if you want to get rid of this functionality, it can probably be recovered by an orthogonal rule/macro layer that uses py_wheel, py_package, and maybe a little bit of glue in Starlark.

Just a driveby comment from somebody that uses this ruleset for packaging/wheels, in case you wanted a nudge for or against removing the API you're talking about.

aignas: @FrankPortman, could you add more details about how you are using our ruleset? We don't want to get rid of packaging rules, we want to get rid of repository_rule repackaging of the files and printing the RECORD patch.

FrankPortman: Yes that's my understanding of your comment. I am essentially corroborating that even if you got rid of what you're saying, it should be straightforward to make a generic rule/macro to get something similar back via py_wheel and py_package.

rickeylev: Thanks, Frank!

rickeylev (approved):

PR #3811, opened 2026-06-06
fix(uv): respect uv.tool settings in pyproject.toml

### LINKED ISSUE #3807: uv settings are ignored in pyproject.toml
@aaronsky from Slack:

> Another thing I noticed: it seems as though `[tool.uv]` settings in my `pyproject.toml` aren’t being considered when using the lock rule, so I have to duplicate settings as flags into args. This works as-expected in rules_uv’s pip_compile rule, despite the fact that neither it nor rules_python’s implementation pass `--project/--config-file`, nor set `UV_PROJECT/UV_CONFIG_FILE` in the environment. So I’m not sure what’s going on there.

aignas:
With this we auto-detect the location of the project file based on
where the first pyproject.toml file is located. Whilst this may work
for majority of the cases there could be a case where the user wants
to leverage the workspaces, we iterate through all pyproject.toml
files and choose the one with the shortest directory path. If this
does not work, we you can just override it manually.

Along the way, give some advice to agents for handling flakey CI and
pushing changes to PRs.

Fixes #3807


python/uv/private/lock.bzl
```diff
@@ -99,6 +100,25 @@ def _lock_impl(ctx):
         args.add("--generate-hashes")
     if not ctx.attr.strip_extras:
         args.add("--no-strip-extras")
+
+    project = None
+
+    # Autodetect the project based on the `pyproject.toml` location - it will be the first src that
+    # we see that is named "pyproject.toml"
+    for src in srcs:
```
rickeylev: Only do the auto-detection logic if ctx.attr.project isn't set

```
project = ctx.attr.project

if not project:
  for src in srcs: ...

if not project:
  project = pkg
```

python/uv/private/lock.bzl
```diff
@@ -275,6 +295,17 @@ modifications and the locking is not done from scratch.
...
         ),
+        "project": attr.string(
+            doc = """\
+Overrides the `--project` directory passed to `uv pip compile`.
+If not set, the project directory is auto-detected: when
+`pyproject.toml` files are in {obj}`lock.srcs`, the one with the
+shortest directory path is selected. This makes `uv` read
+`[tool.uv]` settings (e.g. `no-build-isolation`,
+`exclude-dependencies`) from that `pyproject.toml`.
+{versionadded}VERSION_NEXT_FEATURE
```
rickeylev:
```suggestion

:::{versionadded} VERSION_NEXT_FEATURE
:::
```

tests/uv/lock/pyproject_toml/BUILD.bazel
```diff
@@ -0,0 +1,44 @@
+# Copyright 2025 The Bazel Authors. All rights reserved.
```
rickeylev: Remove bazel copyright

rickeylev (approved):

python/uv/private/lock.bzl
rickeylev: Done

tests/uv/lock/pyproject_toml/BUILD.bazel
rickeylev: Done

rickeylev (approved):

rickeylev:
last error: CI says lock file on windows differs, but doesn't give details about the diff?

PR #3812, opened 2026-06-06
refactor(toolchains): register runtimes using manifest

rickeylev:
This changes the list of runtimes that are registered to come from a file instead of
being logic within Starlark code.

Along the way...

* Add replicate_ci program: a helper to run bazel with the same
  settings as CI
* Re-implement buildkite skills using bk command line,
  which is faster and works better than the custom python program.
* Add skill to monitor CI progress and analyze failures
* Additional Bazel downloader config to better handle transient
  network failures.
* Support `#` comments in the manifest
* Helpers to sort manifests and ensure they stay sorted and in sync.


rickeylev:
I really like this listing of it. It makes it really obvious and apparent how many runtimes are being registered (600+ !). It also just seems much easier to add/remove entries.

The paths in the manifest are a bit funny: `<DATE>/<filename>`. This is somewhat expected because we're mixing and matching across different build dates.

It looks like CI is failing on the workspace tests. I'm not keen to port all this functionality to workspace. I'm thinking, for workspace, we just replace some internals so it reads from the manifest file, but don't expose a `add_runtime_manifest_{file,url}` equivalent.

Q: Should we have one manifest file, or split it along some axis? e.g. python version, platform, build date, etc.

Q: Should we publish a runtime manifest as part of releases? The thought is it lets one easily "upgrade" the runtimes (treating rules_python as a trusted source) without having to upgrade rules_python itself. The counter argument is: one can always add a custom manifest.

python/versions.bzl
```diff
@@ -57,1208 +57,7 @@ _ASTRAL_PREFIX = "https://releases.astral.sh/github/python-build-standalone/rele
 # It is possible to provide lists in "url". It is also possible to provide patches or patch_strip.
```
aignas: Please remove the supporting `print_toolchain_checksum.bzl` docs above.

python/runtimes_manifest.txt
```diff
@@ -0,0 +1,617 @@
+# Standalone runtimes manifest catalog
```
aignas: Please a comment on how this is generated.

python/runtimes_manifest.txt
```diff
@@ -0,0 +1,572 @@
...
+00bb2d629f7eacbb5c6b44dc04af26d1f1da64cee3425b0d8eb5135a93830296  20250317/cpython-3.13.2+20250317-x86_64-unknown-linux-musl-install_only.tar.gz
```
dougthor42: This sorting makes it difficult to know if a given python version is in the manifest.

Could there be another 1st column which is just the file name sans build date? And then things get sorted by that column instead of the hash.

```
cpython-3.11.5+20230826-x86_64-pc-windows-msvc-shared-install_only 00f00226... 20230826/cpython-3.11.5+20230826-x86_64-pc-windows-msvc-shared-install_only.tar.gz
cpython-3.12.9+20250317-aarch64-unknown-linux-gnu-install_only 00c6bf9a... 20250317/cpython-3.12.9+20250317-aarch64-unknown-linux-gnu-install_only.tar.gz
cpython-3.12.11+20250808-aarch64-pc-windows-msvc-install_only 00bf7d7... 20250808/cpython-3.12.11+20250808-aarch64-pc-windows-msvc-install_only.tar.gz
```

I guess it would remove the nice alignment that hash-as-first-col gives. And would need a natsort to handle `.9` being sorted before `.11` and whatnot.

gazelle/downloader_config.cfg
```diff
@@ -14,3 +17,7 @@ rewrite ^github\.com/bazelbuild/bazel-skylib/(.*) mirror.bazel.build/github.com/
 rewrite ^github\.com/bazelbuild/platforms/(.*) mirror.bazel.build/github.com/bazelbuild/platforms/$1
 rewrite ^github\.com/bazelbuild/rules_kotlin/(.*) mirror.bazel.build/github.com/bazelbuild/rules_kotlin/$1
 rewrite ^github\.com/bazelbuild/rules_shell/(.*) mirror.bazel.build/github.com/bazelbuild/rules_shell/$1
+rewrite ^github\.com/bazelbuild/rules_java/(.*) mirror.bazel.build/github.com/bazelbuild/rules_java/$1
+rewrite ^github\.com/bazelbuild/stardoc/(.*) mirror.bazel.build/github.com/bazelbuild/stardoc/$1
+
+
```
dougthor42: nit: extra newlines, here and in other downloader_config files.

dougthor42: All 3 downloader_config files are identical - can/should they be deduped? In a separate PR.

dougthor42 (commented):
> Q: Should we have one manifest file, or split it along some axis? e.g. python version, platform, build date, etc.

I'd vote for one file as long as the sorting issue in my other comment can be addressed.

If not, then split along python minor version (3.9 to 3.15; 7 files) makes sense to me. Then dropping support for a major version is just deleting the file.

> Q: Should we publish a runtime manifest as part of releases? The thought is it lets one easily "upgrade" the runtimes (treating rules_python as a trusted source) without having to upgrade rules_python itself. The counter argument is: one can always add a custom manifest.

Eh. If it's trivial to do, sure, but at least for me I don't see much use for it. How often are people adding their own or customizing runtimes? I don't know.

python/runtimes_manifest.txt
rickeylev: I agree. Right now it's just sorted by hash because I was diffing manifests for debugging purposes.

the `hash filename` format comes from python-build-standalone's SHA256SUMS file, so we more-or-less have to support that. We can support other formats, too, though. I find the hash being first annoying, too.

rickeylev:
Apparently there's a pre-existing bug with some of the toolchain registrations. CI was giving an error that 3.15.0a8-freethreaded wasn't matching (and I noticed that there are ~700 toolchains created, while only ~600 urls listed). It turns out the 3.15.0a8-freethreaded toolchain is downloading a non-freethreaded implementation.

```
bazel mod show_repo @@+python+python_3_15_0a8_x86_64-unknown-linux-gnu-freethreaded

python_repository(
  ...
  urls = [
    ".../cpython-3.15.0a8+20260414-x86_64-unknown-linux-gnu-install_only.tar.gz", 
    ...
  ]
)
```

I'm not entirely clear why this occurs (it doesn't affect all 3.15 or freethreaded).

In any case, now that I've finally figured _that_ out, I just need to add the correct urls to the manifest.

aignas:
> > Q: Should we have one manifest file, or split it along some axis? e.g. python version, platform, build date, etc.
> 
> I'd vote for one file as long as the sorting issue in my other comment can be addressed.
> 
> If not, then split along python minor version (3.9 to 3.15; 7 files) makes sense to me. Then dropping support for a major version is just deleting the file.
> 

Doing a single file for now sounds good to me. Either that or vendored multiple files by release from the upstream.

> > Q: Should we publish a runtime manifest as part of releases? The thought is it lets one easily "upgrade" the runtimes (treating rules_python as a trusted source) without having to upgrade rules_python itself. The counter argument is: one can always add a custom manifest.
> 
> Eh. If it's trivial to do, sure, but at least for me I don't see much use for it. How often are people adding their own or customizing runtimes? I don't know.

I'd say we can publish them, or from the 3.0 release default to the latest python build standalone release, i.e. we would be testing only one release in our CI. But maybe it would be best to pin that.

python/private/python.bzl
```diff
@@ -742,6 +742,17 @@ def _override_defaults(*overrides, modules, _fail = fail, default):
 
             override.fn(tag = tag, _fail = _fail, default = default)
 
+def _manifest_entry_sort_key(entry):
+    flavor_rank = {"full": 3, "install_only": 1, "install_only_stripped": 2}.get(entry.archive_flavor, 4)
```
aignas, line 746: nit: printing unknown flavour and unknown microarch might be useful when debugging.

python/versions.bzl
```diff
@@ -1530,3 +329,67 @@ def gen_python_config_settings(name = ""):
...
+    microarch = entry.microarch
+    if not microarch:
+        microarch_rank = 0
+    elif microarch.startswith("v") and microarch[1:].isdigit():
+        microarch_rank = int(microarch[1:])
+    else:
+        microarch_rank = 999
+    return (flavor_rank, microarch_rank)
+
+def tool_versions_from_manifest_entries(entries, base_url = DEFAULT_RELEASE_BASE_URL):
```
aignas: Is this supposed to be used only in this file? Do we actually want to expose this?

aignas (approved):
LGTM. This is really exciting!

python/runtimes_manifest.txt
```diff
@@ -563,30 +608,11 @@ fb1caac917d7b6497bb6f5950da5f1e48d05c43a498948dd97f85760c4382d9f  20251031/cpyth
...
 fc7e1fb553c47b831ed7fa529575145207f000f967513f7b9ea809cce006ed79  20260325/cpython-3.13.12+20260325-riscv64-unknown-linux-gnu-install_only.tar.gz
 fca340d8fb7a05cd90e216ce601b25d492ed8c1a3b6a6d77703e0f15ab3711a7  20251031/cpython-3.14.0+20251031-riscv64-unknown-linux-gnu-install_only.tar.gz
 fd5a9e0f41959d0341246d3643f2b8794f638adc0cec8dd5e1b6465198eae08a  20240107/cpython-3.12.1+20240107-x86_64-pc-windows-msvc-shared-install_only.tar.gz
 fdfa86c2746d2ae700042c461846e6c37f70c249925b58de8cd02eb8d1423d4e  20241205/cpython-3.13.1+20241205-aarch64-unknown-linux-gnu-install_only.tar.gz
 fe459da39874443579d6fe88c68777c6d3e331038e1fb92a0451879fb6beb16d  20230826/cpython-3.11.5+20230826-s390x-unknown-linux-gnu-install_only.tar.gz
 fee80e221663eca5174bd794cb5047e40d3910dbeadcdf1f09d405a4c1c15fe4  20230726/cpython-3.10.12+20230726-aarch64-unknown-linux-gnu-install_only.tar.gz
 ffb6af51fbfabfc6fbc4e7379bdec70c2f51e972b1d2f45c053493b9da3a1bbe  20251209/cpython-3.13.11+20251209-s390x-unknown-linux-gnu-install_only.tar.gz
-e39a5118b9c3590d2c7fb4539cf70ad6f54c4c2014a164ca635f7eee61f502c5  20260414/cpython-3.15.0a8+20260414-aarch64-apple-darwin-freethreaded+pgo+lto-full.tar.zst
```
aignas: BTW, we have received a new bump of the toolchains, but we probably don't need to update it here: https://github.com/astral-sh/python-build-standalone/releases/tag/20260610

python/runtimes_manifest.txt
aignas: We could have a pre-commit hook that sorts the file in a particular way (e.g. based on the order preference order. We could do it as a followup PR to keep this PR scoped well.

rickeylev:
Blarg, another workspace issue: in order to load the information from a file, a repo rule has to run to parse the manifest and make it loadable in starlark. I initially had rules_python_internal_config do this. However, CI reports Bazel crashing with an NPE.

I think what's happening is the the "repos can only be seen after the next WORKSPACE-level statement" behavior. in the repositories.bzl code it looks like this:

```
load("python:versions.bzl", "TOOL_VERSIONS")

def py_repositories():
  internal_config_repo(...)
  pythons_hub(tool_versions=TOOL_VERSIONS)
```

TOOL_VERSIONS requires the internal_config_repo to be initialized, since it loads the parsed manifest entries from there.

Moving the internal_config_repo() call to WORKSPACE fixes this. Other than introducing another "setup repositories" call (which is fine IMHO), I don't know how else this could be worked around.

edit: I can think of another work around: duplicate the manifest in a .bzl file to make workspace mode easier

rickeylev:
woo, success!

I need to do some cleanup (later, its 1am), but this is finally passing.

For the workspace problem: I just embedded a duplicate of runtimes_manifest.txt as a bzl file so it's easy to load. I don't want to put a lot of effort into workspace mode. Maybe a future change could add the ability to use custom manifests that get parsed at repo phase.

tests/support/mocks/mocks.bzl
```diff
@@ -102,10 +102,31 @@ def _module_new(name, *, is_root = False, **tags):
...
+bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb  20251031/cpython-3.14.0+20251031-x86_64-unknown-linux-gnu-install_only.tar.gz
+cccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc  20260414/cpython-3.15.0a8+20260414-x86_64-unknown-linux-gnu-install_only.tar.gz
+dddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddd  20241205/cpython-3.13.1+20241205-x86_64-unknown-linux-gnu-install_only.tar.gz
+"""
+
 def _mctx_read(self, x, watch = None):
     _ = watch  # @unused
     path_str = x._path if hasattr(x, "_path") else str(x)
     if path_str not in self.mock_files:
+        if "runtimes_manifest.txt" in path_str:
```
rickeylev: the generic mock shouldn't have the runtimes_manifest.txt hard coded into it.

tests/toolchains/transitions/transitions_tests.bzl
```diff
@@ -142,7 +141,7 @@ def _test_full_version(name):
         name = name,
         tests = {
             v.replace(".", "_"): (v, v)
-            for v in TOOL_VERSIONS
+            for v in MINOR_MAPPING.values()
```
rickeylev: Need to revert this test back to using TOOL_VERSIONS instead of MINOR_MAPPING. This might be tricky since TOOL_VERSIONS isn't populated in bzlmod mod anymore, and populating it in workspace mode is problematic.

Ah, have a different object: tool_versions will be workspace mode, and we'll load another that is generated by bzlmod. Since this is a test,
we can have it load a special repo with this info, which shouldn't cause the phasing issues as originally seen.

gazelle/downloader_config.cfg
rickeylev: Done

rickeylev: Done -- added a pre-commit check that syncs the sub-workspaces to the main workspace.

python/private/python.bzl
rickeylev: Since we don't strictly control the filename format, I'd rather it silently fallback. Otherwise, if you use some custom file name format, it'll print an annoying error.

python/runtimes_manifest.txt
rickeylev: I added a brief comment about the origin it came from.

Generating this was somewhat complicated, and a combination of scripts and some hand-modifications. Some hand modifications were needed because of bugs in how the TOOL_VERSIONS would convert things into urls

python/runtimes_manifest.txt
rickeylev: Done; added a script and pre-commit hook

python/versions.bzl
rickeylev: Done

python/versions.bzl
rickeylev: Done; made private

tests/support/mocks/mocks.bzl
rickeylev: Done

tests/toolchains/transitions/transitions_tests.bzl
rickeylev: Done

python/private/print_toolchain_checksums.bzl
rickeylev, line 1: I think this whole print_toolchain_checksums script can be deleted?

rickeylev: Deleted

python/runtimes_manifest.txt
rickeylev: Acknowledged

rickeylev:
FYI: This also adds several skills. I had pretty good results with them (it successfully fetched, diagnosed, reproduced, and fixed several CI failures on the first try and without going off the rails)

PR #3814, opened 2026-06-08
build: configure secondary mirror fallback

rickeylev:
Currently, transient GitHub network issues can cause builds to fail
when pulling e.g. platforms.

To fix, configure Bazel's downloader to try GitHub first and fall back
to mirror.bazel.build.


dougthor42 (approved):
FTR, we resolved this by using [BuildBarn remote asset](https://github.com/buildbarn/bb-remote-asset) and a bazelisk wrapper that installs it for users. LMK if you'd like to know more.

Question: These configs do not apply to non-root bazel modules, correct? Meaning this will have no impact on any consumers of rules_python?

rickeylev:
> use build barn remote asset

Hm, interesting, yeah i'd like to learn more. I'll send you a chat.

> only applies to ruleS_python

Correct

PR #3827, opened 2026-06-17
chore: generate release changelog from news entries

rickeylev:
Currently, we update the change log directly when user-visible changes
are made. This works, but makes backporting patches into release
branches prone to unnecessary conflicts. These conflicts can sometimes
be quite confusing, making it easy to accidentally merge incorrect
content.

Instead, what we can do is keep track of an upcoming release's change
log entries in separate files under the `news/` directory. When
creating a release, all the news entries are assembled into a section
in the change log and the news entries deleted. When backporting a
change, after cherry picking, the news entries are merged into the
release's change log section.


rickeylev:
Moving to draft to refine this a bit more

rickeylev:
Ok, I think this is ready now.

aignas (approved):

PR #3833, opened 2026-06-19
chore: automate releases

rickeylev:
This introduces workflows and tools to automate releases. The basic
idea it to have an issue with tasks, and workflows invoke tools that
parse and update the tasks in the issue.

The basic release steps are:
1. Prepare a release (update changelog via PR)
2. Create a release branch after preparation
3. Backport changes into the release branch
4. Tag RCs
5. Promote to final

Each step has a workflow. For now, most of the workflows require a
manual trigger, to help prevent workflows from running wild as we
flush out bugs and edge cases.

The workflows almost entirely delegate to the `release` tool. There's
basically one command per workflow. This makes it easy to encode
specialized logic and run it locally for testing or reproduction.


.github/workflows/process_backports.yml
```diff
@@ -0,0 +1,40 @@
...
+    # Always gate GHA runs to ensure we are operating on a type:release labeled issue if metadata is queried
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Setup Bazel
+        uses: bazel-contrib/setup-bazel@0.0.8
```
aignas: We already are at 0.19.0: https://github.com/bazel-contrib/setup-bazel

Please double check the GH action versions generated by the slopbot, usually they are way out-dated.

.github/workflows/cut_release_branch.yml
```diff
@@ -0,0 +1,37 @@
...
+  issues: write
+
+jobs:
+  cut_branch:
+    # Run only if the issue has the type:release label
+    if: contains(github.event.issue.labels.*.name, 'type:release')
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
```
aignas: The same for the checkout action, it is at v6 right now.

aignas (approved):
LGTM other than the versions of the GH actions used in here.

.github/workflows/cut_release_branch.yml
rickeylev: Done

.github/workflows/process_backports.yml
rickeylev: Done. Also added an agent rule for github yaml files

PR #3837, opened 2026-06-19
feat: unified pypi hub repository

rickeylev:
This implements a pypi hub that is the union of all pypi hubs.

The basic design is:
* The `pip` extension _always_ creates a `@pypi` repo unless the name
  is already taken by another hub definition.
* The `--pypi_hub` flag dispatches to one of the hubs. If not set,
  then it uses a default one (first, or as configured)

The set of packages and targets the unified hub exposes is a union
of all other hubs. If the unified hub routes to a hub that doesn't
support such a target, then it points to a target that fails at
execution time. This is to allow query and cquery to work even if
some targets don't exist in some hubs.


rickeylev:
Ready for review.

I think the main win of this, in comparison to using `pip.parse.config_settings` and custom pip.parse platforms, is it creates a well-defined global name so that code doesn't have to worry about custom hub names. e.g. the protobuf rules, pytest rules, sphinxdoc rules, can refer to a pypi dependency without having to know what root-level hub_name was used for pypi stuff.

.agents/plans/pypi_hub_proxy_feature.md
```diff
@@ -0,0 +1,287 @@
...
+### Automatic Proxy Construction & Collision Logic
+
+During the evaluation of the `pip` extension across the dependency graph:
+1.  **Unconditional Creation**: The extension will **always** synthesize a
+    proxy repository rule with the apparent name `pypi`, even if zero
+    `pip.parse` concrete hubs are defined in the dependency graph (in which
+    case the proxy is completely valid but empty).
+2.  **Collision Prevention**: If a user explicitly defines a concrete hub
+    named `pypi` (`pip.parse(hub_name = "pypi")`), the automatic proxy
+    synthesis is skipped so the user maintains absolute control over that
```
aignas, line 30: In this particular case we could:
* Print a warning that it is skipped.
* Make the hub `<module_name>.pypi` in this case and tell this to the user. Since this is a special name, we should not get us in a situation where we break if there is a non-root module using this name.
* If the user defines a `pip.parse` with `pypi`, then set that as default automatically.

.agents/plans/pypi_hub_proxy_feature.md
```diff
@@ -0,0 +1,287 @@
...
+string_flag(
+    name = "pypi_hub",
+    build_setting_default = "auto", # Default value is "auto"
+    visibility = ["//visibility:public"],
+)
+```
+
+In `python/private/common_labels.bzl`:
+```starlark
+    PYPI_HUB = str(Label("//python/config_settings:pypi_hub")),
```
aignas: I think it would be nicer to have something like `venv` instead of `pypi_hub` because that could be something that is more familiar to users.

docs/pypi/download.md
```diff
@@ -50,6 +50,73 @@ You can use the pip extension multiple times. This configuration will create
...
+)
+use_repo(pip, "pypi_a")
+
+# Define concrete hub 'pypi_b'
+pip.parse(
+    hub_name = "pypi_b",
+    python_version = "3.11",
+    requirements_lock = "//:requirements_b.txt",
+)
+use_repo(pip, "pypi_b")
```
aignas: Why do we need to `use_repo` of this specific hub?

python/private/pypi/extension.bzl
```diff
@@ -28,8 +28,27 @@ load(":pip_repository_attrs.bzl", "ATTRS")
...
+
+_whl_mods_repo = repository_rule(
+    doc = """\
+This rule creates json files based on the whl_mods attribute.
+""",
+    implementation = _whl_mods_repo_impl,
+    attrs = {
+        "whl_mods": attr.string_dict(
+            mandatory = True,
+            doc = "JSON endcoded string that is provided to wheel_builder.py",
```
aignas: nit: please update the docstring.

python/private/pypi/extension.bzl
```diff
@@ -404,8 +423,19 @@ You cannot use both the additive_build_content and additive_build_content_file a
         hub_group_map[hub.name] = out.group_map
         hub_whl_map[hub.name] = out.whl_map
 
+    default_hub = None
+    for mod in module_ctx.modules:
```
aignas: This should be part of `build_config`

.agents/plans/pypi_hub_proxy_feature.md
```diff
@@ -0,0 +1,287 @@
...
+    a target depending on an unrepresented missing package succeeds, while
+    `bazel run` on that target gracefully fails during execution with the exact
+    synthesized error message.
+5.  **Unionized Extra Hub Aliases**: Author a test asserting that a binary
+    successfully runs using a custom `extra_hub_aliases` target resolved
+    through the `@pypi` proxy.
+
+## 4. Execution Steps
+
+1.  **Phase 1**: Define `pypi_hub` `string_flag` and register it in
```
aignas:
```suggestion
1.  **Phase 1**: Define `venv` `string_flag` and register it in
```

`aspect/rules_py` is using this term for some things so maybe it would be good if we used it as well to make it easier for the users to understand the feature parity between the rulesets.

python/private/pypi/unified_hub_setup.bzl
```diff
@@ -0,0 +1,102 @@
...
+            flag_values = {"@rules_python//python/config_settings:pypi_hub": hub},
+        )
+
+_STANDARD_ALIASES = [
+    "pkg",
+    "whl",
+    "data",
+    "dist_info",
+    "extracted_wheel_files",
+]
```
aignas, line 32: This should be imported from somewhere.

.agents/plans/pypi_hub_proxy_feature.md
rickeylev: Thanks for coming up with some ideas. My initial thinking was to skip the logic to ensure existing behavior wasn't affected, then figure out how we could transition to `@pypi` being a pip-extension owned name.

The idea to make `name=pypi` an implicit default is appealing. I'm a bit concerned it may over-complicate how a default is selected. But...I do like it. Feels like a pretty reasonable behavior. So lets do that.

What do you think of:

Print a warning if the name collision occurs. If an env var is set (`RULES_PYTHON_PYPI_HUB_RESERVED=1`), then the hub is silently renamed module_name.pyi and is used as the default hub (if pip.default wasn't used). In a future release, we flip the default.

docs/pypi/download.md
rickeylev: We do not; removed.

python/private/pypi/extension.bzl
rickeylev: how so?

i think this is a slop diff. I don't see why it moved this function all the way to the top of the file. I've moved it back down to where it was.

.agents/plans/pypi_hub_proxy_feature.md
rickeylev: Done

.agents/plans/pypi_hub_proxy_feature.md
rickeylev: Done.

python/private/pypi/unified_hub_setup.bzl
rickeylev: Done, imported from the pypi labels.bzl file

python/private/pypi/extension.bzl
rickeylev: done

.agents/plans/pypi_hub_proxy_feature.md
aignas: SGTM

aignas (approved):

.agents/plans/pypi_hub_proxy_feature.md
rickeylev: Done

PR #3838, opened 2026-06-20
ci: enable bazel-diff

rickeylev:
Enable bazel-diff to improve CI performance by skipping targets that aren't affected by a PR


.bazelci/presubmit.yml
```diff
@@ -25,6 +25,8 @@ buildifier:
   bazel: 7.x
   skip_in_bazel_downstream_pipeline: "Bazel 7 required"
 .reusable_config: &reusable_config
+  environment:
+    USE_BAZEL_DIFF: "1"
```
dougthor42, line 29: Where is this env var used? Is it baked into `.bazelci` config outside of the rules_python repo or something?

I was trying to look up what exactly it would end up doing. I know about https://github.com/Tinder/bazel-diff but we don't have it as a `bazel_dep`.

jvolkman: Looks like it's here: https://github.com/bazelbuild/continuous-integration/blob/36dd157dc727d2787c4f8f515825e63bdb52e4be/buildkite/bazelci.py#L539

dougthor42: :tada:

dougthor42 (approved):

rickeylev:
I think this broke CI. I think that what happened is the first enabling of bazel-diff skips running it entirely. Subsequent runs use it, but it tries to run targets with target_compatible_with set, then fails.

PR #3839, opened 2026-06-21
fix: run bazel-diff in cquery mode under bazelci

rickeylev:
This PR fixes the CI breakage on main where bazel-diff ran in plain query
 mode, ignoring target_compatible_with and attempting to run
platform-incompatible tests.

BazelCI has an `EXP_USE_CQUERY` env variable that uses cquery to identify
targets instead. Try that, which should respect TCW and filter out
targets that would be passed to bazel-diff.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -27,6 +27,7 @@ buildifier:
 .reusable_config: &reusable_config
   environment:
     USE_BAZEL_DIFF: "1"
+    EXP_USE_CQUERY: "true"
   build_targets:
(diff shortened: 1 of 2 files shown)


rickeylev:
ai got pretty creative in figuring out how to call bazel-diff in cquery mode :laughing:

ignore this for now

rickeylev:
there we go, much more sane

rickeylev:
since CI is broken and changes to presubmit.yml will cause bazel-diff to be skipped entirely, i'm going to force submit this.

PR #3850, opened 2026-06-26
feat(pypi): add pip.dep to declare abstract pypi dependencies

rickeylev:
Introduce the pip.dep tag class to allow modules to declare abstract
PyPI dependencies. These declarations ensure the target structure for
a PyPI package is created, but provide a no-op implementation that
passes analysis time, but fails at execution time.

Along the way, create an agent rule for `*.bzl` files to help guide it
in creating better bzl files.


tests/pypi/extension/extension_tests.bzl
```diff
@@ -435,6 +443,35 @@ def _test_default_hub_precedence(env):
...
+    pypi = _parse_modules(
+        env,
+        module_ctx = _pypi_mock_mctx(
+            _mod(
+                name = "my_module",
+                dep = [
+                    _dep(
+                        name = "declared-pkg",
+                        extra_targets = ["declared-alias"],
+                    ),
```
aignas, line 456: Please add a unit test to ensure that if we declare a dep and then it is provided by a pip.parse, that it works as expected.

rickeylev: Done

tests/integration/unified_pypi_test.py
```diff
@@ -74,6 +74,27 @@ def test_invalid_default_hub_fails_evaluation(self):
...
+    def test_unimplemented_declared_dep_fails_build(self):
+        # Even though cquery succeeds:
+        self.run_bazel("cquery", "//:bin_declared_only")
+
+        # Build must fail because the package is not implemented by any concrete hub
+        result = self.run_bazel("build", "//:bin_declared_only", check=False)
+        self.assertNotEqual(result.exit_code, 0)
+        self.assert_result_matches(
+            result,
+            'ERROR: PyPI package "declared_only_pkg" is not available when building under PyPI hub "pypi_b".',
```
aignas: A helpful error message pointing users what to do here would be really good.

aignas (approved):

tests/integration/unified_pypi_test.py
rickeylev: Done

Issue #3265, opened 2025-09-11
Use tool to manage bzl_library generation


rickeylev:
It's become a bit of a pain to manage all the bzl_library targets. A tool to deal with it would be much nicer.

I made a quick try and using https://github.com/bazelbuild/bazel-skylib?tab=readme-ov-file#gazelle-plugin

It mostly worked, but had the following problems:
1. Our local_repository() hack in workspace broke it. I had to comment that out. This is a hack for Bazel 9+workspace. Could possibly remove it, since workspace may be dropped in bazel 9
2. It rewrote **_all_** our build files. It's convention is `{name}` for bzl_library, not `{name}_bzl`. Renaming our internal targets is feasible, but we have to retain the public targets naming.
3. It added go_proto_library everywhere. We don't need or want that.
4. I think it rewrote various other targets. I'm not against that, per se, however, I'm not a fan of the giant comment blocks configuring gazelle. I'm also wary of gazelle changing its preferred naming behavior and churning a lot.
5. The normal setup has us put a load for gazelle in our root build file; we can't do this as it would make gazelle a prod dependency, when it needs to be a dev-only dependency. Maybe if we just move it to tools/private?
The functionality we need is:

As an alternative, I have a tool that mostly handles bzl_library stuff almost done. It just does two things: generating bzl_library, and sorting names in build files.

cc @dougthor42

rickeylev:
Fixed by https://github.com/bazel-contrib/rules_python/pull/3852

PR #3852, opened 2026-06-26
refactor: use gazelle for bzl_library management

rickeylev:
To make it easier to manage the many bzl_library targets we have, switch
to using
gazelle to do so.

Unfortunately, gazelle has strong opinions about target names: it forces
`{foo}` names and doesn't allow `{foo}_bzl` names. Trying to make it do
so requires
quite a bit of gazelle directives.

Instead, rename all the internal targets, but create aliases for public
targets.

Along the way...
* Delete unused py_args.bzl
* Add agent rule to avoid inappropriate copyright
* Add a skill for creating rules


tests/support/whl_from_dir/BUILD.bazel
```diff
@@ -0,0 +1,22 @@
+# Copyright 2024 The Bazel Authors. All rights reserved.
```
dougthor42: 1. I thought we didn't need copyrights anymore?
2. Year is wrong

dougthor42 (commented):
> Unfortunately, gazelle has strong opinions about target names: it forces `{foo}` names and doesn't allow `{foo}_bzl` names. Trying to make it do so requires quite a bit of gazelle directives.

I found similar issues when trying to gazelle-ify our `bzl_library` stuff in QAI 😞.

General question: Did you delete most files and have Gazelle remake them? Ordering has changed and whatnot.

Looks like a lot of build failures, too.

aignas:
I saw one `bzl_library` `alias` created with visibility `@rules_python//:__subpackages__`. I think that restricted visibility deprecated aliases should be just removed.

aignas (approved):

rickeylev:
Yeah, slopbot is having a hard time with this. Feel free to drop random comments on things you see so they don't get forgotten about.


> remove visibility restricted aliases

+1

> wrong copyrights

Ugh, I know. It really insists on adding them. agents.md says not to. Very open to ideas on how to stop it from doing this.

tests/support/whl_from_dir/BUILD.bazel
rickeylev: Fixed

aignas:
> Ugh, I know. It really insists on adding them. agents.md says not to. Very open to ideas on how to stop it from doing this.

Creating a separate file to list the copyrights for each file that we have from the legacy times and remove the copyright headers from the code and have a pre-commit hook to check for that.

aignas:
Please reference #3265 in this PR :)

PR #3854, opened 2026-06-26
fix: avoid stage1 bootstrap stdlib shadowing

gleyba:
The stage 1 system-python bootstrap imports standard-library modules before it
re-execs into the configured runtime. If the target output directory contains
files named like standard-library modules, such as `shutil.py` or `types.py`,
the bootstrap can import those files instead and fail before user code starts.

This change removes Python's unsafe script-directory prepend before those early
imports, while preserving `sys.path[0]` when Python has already suppressed that
prepend through `-P`/`PYTHONSAFEPATH` or isolated mode. Stage 2 now uses the same
isolated-mode guard when removing or recreating the main-directory path.

It also pins the Windows venv entry point template to LF line endings so the
compile-pip CI dirty-worktree check is stable across checkout settings.

Before this change, a generated target output could shadow bootstrap stdlib
imports, and isolated-mode runs on older Python versions could lose a required
stdlib path.

After this change, bootstrap stdlib imports resolve from the interpreter's
standard library, isolated mode preserves the interpreter-provided path, and the
line-ending-sensitive CI check stays clean.

Tests:

```shell
bazel test --config=fast-tests \
  //tests/bootstrap_impls:stdlib_shadowing_system_python_test

bazel test --config=fast-tests \
  //tests/bootstrap_impls:stdlib_shadowing_system_python_test \
  //tests/bootstrap_impls:interpreter_args_test \
  //tests/bootstrap_impls:sys_path_order_bootstrap_script_test
```


python/private/python_bootstrap_template.txt
```diff
@@ -5,13 +5,31 @@ from __future__ import absolute_import
...
+# Python 3.11 introduced PYTHONSAFEPATH (-P), which disables the unsafe prepend.
+# Isolated mode (-I) also disables it, including on older interpreters without
+# safe_path. In either case, sys.path[0] is not the script directory and should
+# be preserved.
+if (
+    not getattr(sys.flags, "safe_path", False) and
+    not getattr(sys.flags, "isolated", False) and
+    sys.path
+):
+  del sys.path[0]
```
aignas, line 25: I remember the `coverage` changing the first item in the `sys.path` - we should have a test to ensure that the instrumentation there is not going to break during `bazel coverage` runs.

.gitattributes
```diff
@@ -1,6 +1,7 @@
 python/features.bzl export-subst
 tools/publish/*.txt linguist-generated=true
 tests/uv/lock/testdata/requirements.txt text eol=lf
+python/private/pypi/venv_entry_point_template.bat text eol=lf
```
aignas: Is this an intended change?

aignas (commented):
Please also add a `news/3854.fixed.md` file to describe the fix.

python/private/python_bootstrap_template.txt
```diff
@@ -5,13 +5,25 @@ from __future__ import absolute_import
...
+# By default, Python prepends the directory containing this script to
+# sys.path. The stage 1 bootstrap only needs stdlib modules before it
+# re-execs into the configured runtime, so avoid resolving imports from the
+# target's output or runfiles package directory. This matters when that
+# directory contains files with stdlib names, such as shutil.py or types.py.
+#
+# Python 3.11 introduced PYTHONSAFEPATH (-P), which disables the unsafe prepend.
+# In that case, sys.path[0] is not the script directory and should be preserved.
+if not getattr(sys.flags, "safe_path", False) and sys.path:
+  del sys.path[0]
```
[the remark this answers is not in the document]
gleyba: done

.gitattributes
gleyba: removed

python/private/python_bootstrap_template.txt
gleyba: I tried:
```
bazel coverage --config=fast-tests --nocache_test_results //tests/bootstrap_impls:stdlib_shadowing_system_python_test
```
seems to work, what else should i test?

aignas (approved):

PR #3855, opened 2026-06-27
feat: stop using chmod on Bazel 8.6+

### LINKED ISSUE #3585: Stop using `chmod` on bazel 8.6 or 9.0.1 and above
This is a followup to #3554. Since bazel has added a fix for this to the future versions, we should make the usage of `chmod` only on particular bazel versions.

aignas:
## Summary

Add an `extract_needs_chmod` config flag that controls whether `chmod`
is run after extracting wheel files. The flag is `False` for Bazel
>= 8.6 (where the fix for file permissions is built into Bazel) and
`True` for older versions.

Move `_maybe_fix_permissions` from `whl_extract.bzl` into
`repo_utils.bzl` and have `_extract` call it conditionally based
on the config flag. This ensures both `whl_extract` and `patch_whl`
benefit from the chmod fix when needed.

Whilst at it migrate tests to common mocks.

Fixes #3585
Closes #3860


tests/repo_utils/repo_utils_test.bzl
```diff
@@ -6,6 +6,48 @@ load("//tests/support:mocks.bzl", "mocks")
 
 _tests = []
 
+def _make_rctx(*, os_name, mock_extracts = None, mock_files = None):
```
rickeylev, line 9: Please use the mock from tests/support/mocks/mocks.bzl#mocks.rctx instead

rickeylev (approved):
lgtm with comment

news/3855.changed.md
```diff
@@ -0,0 +1 @@
+(pypi) chmod after wheel extraction is now skipped on Bazel 8.6+
```
rickeylev: Is this worth putting in news? I don't think there's a user visible change in behavior.

Issue #3791, opened 2026-05-19
Remote repo contents cache is known to not work correctly


aignas:
In our repository we are using remote contents cache metadata in the following repositories:
* `whl_library` - symlinks from the venv are added to files from this repo if we are using the site-packages venv layout, which is not the default yet.
* `uv_repository` - it is unaffected
* `python_repository` - symlinks from the venv are added to files from this repo

https://github.com/search?q=repo%3Abazel-contrib%2Frules_python%20repo_metadata&type=code

Because of how we are using symlinks to create a `venv` as part of the build actions, there may be failures, like:

```
ERROR: Python interpreter not found: /home/buildbuddy/workspace/output-base/execroot/_main/bazel-out/k8-fastbuild/bin/libs/artifacts/cli.runfiles/_main/libs/artifacts/_cli.venv/bin/python3
```

The relevant `.bazelrc` is:
```
startup --experimental_remote_repo_contents_cache
```

A potential workaround is to avoid creating symlinks
```
common --@rules_python//python/config_settings:venvs_use_declare_symlink=no
```

[Flag documentation](https://rules-python.readthedocs.io/en/latest/api/rules_python/python/config_settings/index.html#venvs_use_declare_symlink)

DoDoENT:
This issue [seems](https://github.com/DoDoENT/bazel-playground/actions/runs/26340354076/job/77541188104?pr=36) to be backported now also to `rules_python v1.9.1`. 

EDIT: `1.9.0` is [also affected](https://github.com/DoDoENT/bazel-playground/actions/runs/26340532360/job/77541680746?pr=36) - it happened to work for me by accident

DoDoENT:
This issue is actually bigger than it appears... Without the `experimental_remote_repo_contents_cache`, `toolchains_llvm` are always extracted on CI machines, making builds very, very slow.

travisdowns:
> This issue is actually bigger than it appears... Without the `experimental_remote_repo_contents_cache`, `toolchains_llvm` are always extracted on CI machines, making builds very, very slow.

We are hitting this. Is it filed upstream? New for bazel 9?

aignas:
See: https://github.com/bazelbuild/bazel/pull/29767

fmeum:
Please test with https://github.com/bazelbuild/bazel/pull/29767 if you can!

DoDoENT:
[...]

Issue #2530, opened 2024-12-24
Use pip repos with different names between different python versions


michaeljs1990:
# 🐞 bug report

### Affected Rule

This seems to be an "issue" when you using py_library and add a dependency to it that is in a pip repo not using your default python version. Hopefully I explained that correctly but I also opened a PR to make it easy to reproduce https://github.com/bazelbuild/rules_python/pull/2529.

### Is this a regression?

I don't think so I have another version of rules python in use 0.32.2 I believe and it has the same issue.

### Description

https://github.com/bazelbuild/rules_python/pull/2529 described better here but when using multiple repos and python versions and trying to use the library that contains a requirement from the non default and building it things blow up and look like this.

```
ERROR: /private/var/tmp/_bazel_schuettm/e053d0eaf0a843e9f76bdb61766af4b7/external/rules_python++pip+pypi_3_10_only/flask/BUILD.bazel:5:12: configurable attribute "actual" in @@rules_python++pip+pypi_3_10_only//flask:pkg doesn't match this configuration: No matching wheel for current configuration's Python version.

The current build configuration's Python version doesn't match any of the Python
wheels available for this distribution. This distribution supports the following Python
configuration settings:
    //_config:is_python_3.11

To determine the current configuration's Python version, run:
    `bazel config <config id>` (shown further below)

and look for one of:
    @@rules_python+//python/config_settings:python_version
    @@rules_python+//python/config_settings:pip_whl
    @@rules_python+//python/config_settings:pip_whl_glibc_version
    @@rules_python+//python/config_settings:pip_whl_muslc_version
    @@rules_python+//python/config_settings:pip_whl_osx_arch
    @@rules_python+//python/config_settings:pip_whl_osx_version
    @@rules_python+//python/config_settings:py_freethreaded
    @@rules_python+//python/config_settings:py_linux_libc

If the value is missing, then the default value is being used, see documentation:
https://rules-python.readthedocs.io/en/latest/api/rules_python/python/config_settings

This instance of @@rules_python++pip+pypi_3_10_only//flask:pkg has configuration identifier e62ee83. To inspect its configuration, run: bazel config e62ee83.

For more help, see https://bazel.build/docs/configurable-attributes#faq-select-choose-condition.

Use --verbose_failures to see the command lines of failed build steps.
ERROR: Analysis of target '//tests:my_lib_3_10_only_test' failed; build aborted: Analysis failed
INFO: Elapsed time: 0.197s, Critical Path: 0.02s
[...]

PR #3856, opened 2026-06-27
perf(pypi): extract wheels once and reuse

### LINKED ISSUE #2948: Setup fewer wheel repos in pip.parse

aignas:
Summary:
- Extract wheels in a repository A.
- Then parse METADATA in a repository B and C and reuse A everywhere.
- We also alias the targets for each entry point that is created in whl_library.
- The wheel repository names will be shorter in general that should help on systems
  like Windows, where path length is sometimes an issue.

Limitations:
- If users use the whl_mods API we cannot reuse the extracted whls.
- If users patch wheels, those are not reused.
- Reused only within the context of a single hub repo, we could do better with
  reuse across all of the hubs, but then we need to solve the "how to tell users
  that we need to have a single index per package? how to customize it?" problem.
- If we find the same wheel downloaded from different indexes in the same hub repo
  then we stop reuse for that particular wheel file. This should be rare?

Extra thoughts on the design:
- This is a step towards automatic cycle resolution.
- With uv.lock we can point directly to the extracted sources because
  the dep graph can be passed to the hub repo.

Once #3791 is resolved this should bring reasonable speedups, especially if there the same wheel used in multiple places.

Fixes #2948
Work towards #2530


.agents/plans/single_dep_whl_library.md
```diff
@@ -0,0 +1,131 @@
...
+`docs/readthedocs_build.sh` file so that we are exercising the code. And document this in
+the `docs/environment-variables.md` file.
+If the flag is off, then it is equivalent to the current
+`main` branch behaviour, if it is on, then it is equivalent to turning all of the behaviour
+implemented here on.
+
+Since this is a refactor for users not using private APIs, the env variable will be switched/removed
+later together with old code.
+
+All of the code should be written in a TDD style where we add a failing test for a feature and then
```
rickeylev: Interesting idea. It it working ok?

.agents/plans/single_dep_whl_library.md
```diff
@@ -0,0 +1,131 @@
...
+  to `+pip+dev_pip_requests_py3_none_any_2a0d60c1`
+* from `+pip+dev_pip_314_roman_numerals_py3_none_any_647ba99c`
+  to `+pip+dev_pip_roman_numerals_py3_none_any_647ba99c`
+* from `+pip+dev_pip_314_markupsafe_cp314_cp314t_manylinux_2_17_x86_64_fed51ac4`
+  to `+pip+dev_pip_markupsafe_cp314_cp314t_manylinux_2_17_x86_64_fed51ac4`
+
+So the naming convention after the change is:
+`<prefix>_<name>_<py_tag>_<abi_tag>_<platform_tag>_<sha256[:8]>`
+
+Where `<prefix>` is the hub repository name and the rest of the segments come from the wheel name
```
rickeylev: If we could do away with the hub name in the whl_library name, this would allow different hubs to benefit from the optimization, too. (any progress on de-duplication would be good, though!)

.agents/plans/single_dep_whl_library.md
```diff
@@ -0,0 +1,131 @@
+# Single dep for spoke repos per `whl_name`.
```
rickeylev: idk how much of this plan is hand-written vs generated.

fwiw, i've had best results with having it generate the plan, then go through a few rounds of review and tweaking.

python/private/pypi/render_pkg_aliases.bzl
```diff
@@ -87,7 +87,25 @@ package(default_visibility = ["//visibility:public"])
         extra_loads = extra_loads,
     )
 
-def render_pkg_aliases(*, aliases, requirement_cycles = None, extra_hub_aliases = {}, **kwargs):
+def _render_extra_alias(*, name, repo, target):
+    return """\
+package(default_visibility = ["//visibility:public"])
```
rickeylev: this package() line probably doesn't belong here and will cause weird errors

rickeylev (commented):
saw this was draft, so didn't look too thoroughly, just for things that looked like the bot was obviously wrong

.agents/plans/single_dep_whl_library.md
aignas: I find that the tests are sometimes better quality and it helps with the iterative process.

.agents/plans/single_dep_whl_library.md
aignas: For this to work we would have to have the dep graph in the bu repo. Not not-doable, but requires the `whl_library` to not do any depgraph generation. The reasons are 2 fold:
1. The hubs may declare cyclic dependency graphs. The handling of that is done at the hub level, not the spoke, but in the spoke we need to exclude them. We could potentially reuse `whl_library` nodes that do not have any cycles.
2. We are not passing `target_platforms` to the `whl_library` and instead we are passing the packages that the hub contains, so that we can avoid `bazel query` errors. This was done in the hopes that `whl_library` may not need to know `target_platforms` to generate the `deps`. In the end, this may be not a valid assumption.

For us to be able to reuse the `whl_library` in all places we have to:
* Write the `target_platforms` into the `unified_hub` as some sort of `venv -> list[target_platform]` map.
* Load that via `load("unified_hub")` in the `whl_library` and for each value we would generate an appropriate `deps` field. This would happen in a macro, so would not trigger refetching of the repos.
* Load cycles per venv from the unified hub repo. This allows us to exclude the deps in the cycle from the `whl_library` context. Again, the strategy should be done as above.

I think this is technically doable with the current layout and it could become way faster that way, but I'd like to first untangle the per-venv usage.

.agents/plans/single_dep_whl_library.md
aignas: I have done the same - write some of the plan and then go a few rounds of review and tweaking, but I did the review via AI - i.e. does it understand what to implement? We'll see how it works. I'll move the file to the `.agents` dir.

python/private/pypi/render_pkg_aliases.bzl
aignas: I'll let AI find this out via TDD :D Thanks for highlighting this.

aignas:
Again, ran out of free tokens, so will postpone for another day.

aignas:
* 2 types of whl_library repos - one with extracted files, one wiring the deps.

aignas:
So it seems that it is much better to just separate the repositories - one with extracted whl sources, one with the parsed METADATA. This way the change is surgical and very easy to reason about. This was spiked by hand and then vibed and debugged. The build should work but there are many things missing.

TODO:
- [ ] split the new repo into a separate file.
- [ ] Create a separate function for the BUILD.bazel file generation and add unit tests.
- [ ] Update existing hub_builder.bzl unit tests.
- [ ] Remove the confabulated "fixes" to make the code easier to read.

python/private/pypi/extension.bzl
```diff
@@ -441,14 +441,21 @@ You cannot use both the additive_build_content and additive_build_content_file a
...
     extra_aliases = {}
     whl_libraries = {}
+    whl_library_deps_map = {}
     for hub in pip_hub_map.values():
         out = hub.build()
 
         for whl_name, lib in out.whl_libraries.items():
             if whl_name in whl_libraries:
-                fail("'{}' already in created".format(whl_name))
+                print("'{}' already in created".format(whl_name))
```
aignas: TODO: add back the fail.

python/private/pypi/hub_builder.bzl
```diff
@@ -308,14 +316,12 @@ def _add_whl_library(self, *, python_version, whl, repo):
...
-    # we are using pipstar and we are downloading the wheel using the downloader
-    #
-    # However, for that we should first have a different way to reference closures with
-    # extras. For example, if some package depends on `foo[extra]` and another depends on
-    # `foo`, we should have 2 py_library targets.
-    repo_name = "{}_{}_{}".format(self.name, version_label(python_version), repo.repo_name)
+    whl_repo_name = "whl_{}".format(repo.whl_repo_name)
+
+    # TODO @aignas 2026-07-03: filter out the config_load from the args here
+    self._whl_libraries[whl_repo_name] = repo.args
```
aignas: Check that there are no unnecessary args.

python/private/pypi/hub_builder.bzl
```diff
@@ -344,6 +352,20 @@ def _add_whl_library(self, *, python_version, whl, repo):
     else:
         mapping[repo.config_setting] = repo_name
 
+def _add_whl_library_deps(self, repo, whl_repo_name, deps_repo_name):
+    args = repo.args
+
+    deps_args = {}
+    for key in ("config_load", "dep_template", "group_deps", "group_name", "annotation", "pip_data_exclude"):
```
aignas: Pass annotations here and handle everything.

aignas: pass visibility here.

python/private/pypi/extension.bzl
```diff
@@ -22,6 +22,7 @@ load("//python/private:auth.bzl", "AUTH_ATTRS")
 load("//python/private:normalize_name.bzl", "normalize_name")
 load("//python/private:pyproject_utils.bzl", "read_pyproject", "version_from_requires_python")
```
rickeylev, line 23: more

/review

?

aignas:
This should now work.

EDIT: just realized that merging `main` into this branch should not have affected it.

aignas:
OK, I personally like the fact that everything is done as part of `whl_library` macro, which allows us to change less code.

aignas:
I'll add some unit tests for piece of mind.

aignas:
Hmmmm. This is an interesting one:
```
    common: {
        "requirement": "alabaster==0.7.13",
        "index_url": "https://pypi.org/simple/alabaster",
        "urls": ["https://files.pythonhosted.org/packages/64/88/c7083fc61120ab661c5d0b82cb77079fc1429d3f913a456c1c82cf4658f7/alabaster-0.7.13-py3-none-any.whl"],
        "integrity": "sha256-HuGayoAburtbo/XyWORCLfqG+C8+nO+whZsoPN1/YqM=",
        "filename": "alabaster-0.7.13-py3-none-any.whl",
    }
    missing: {
        "envsubst": ["PIP_INDEX_URL"],
    }
```

I think in this particular case, we could just add the `envsubst` to the `ignore_keys` part. However, if the `index_url` and the `urls` differ before interpolation with envsubst, but not after, then maybe we should handle this in some particular way... Some ideas:
* Add the extra URL for each wheel.
* For each URL, we try the URL synchronously.
* For each URL we also add the index_url to ensure that we explicitly state which is which.
* We change the URLs to a dict, where the index_url is the key and we use the `setdefault` to populate it. For any index URL the value should then be the same and we avoid the error.

aignas:
Extra things that we should do:
* parse the wheel name and do not use the full wheel name but only use: distribution, py_tag, abi_tag, platform_tag, and a little bit of sha256? Since we start using sub-resource shas, maybe we should not use the sha256 at all?

aignas:
Given the recent refactors, I've did it again in a more surgical way.

PR #3859, opened 2026-06-28
build(ci): consolidate PR guardrails and block draft files

rickeylev:
Create a workflow for lighter weight PR checks. These just check basic things
about a PR, e.g. markers to prevent submit, blocking files that shouldn't
be checked in.


aignas (approved):

.github/workflows/basic-checks.yml
```diff
@@ -0,0 +1,52 @@
...
+  pull_request_target:
+    types:
+      - opened
+      - synchronize
+      - reopened
+      - labeled
+      - unlabeled
+
+jobs:
+  basic-checks:
```
aignas: Basic checks does not say much. Maybe we should consolidate this into the `ci.yaml` file?

My thinking is that in addition to buildkite:
* PR should have the basic checks, ruff, mypy ran before each PR.
* main should have `check for blacked files, ruff, mypy`, maybe some extra things.

rickeylev: I went to merge them into ci.yaml, but the agent raised a good point: doing so would make the mypy and ruff checks run when labels are changed, which is wasteful.

I renamed it to pr-metadata-checks, which is more descriptive.

PR #3865, opened 2026-06-28
fix: add explicit `data` attribute to `compile_pip_requirements` and forward it to the generated `py_binary`

### LINKED ISSUE #3858: `compile_pip_requirements` does not forward `data` attribute to the generated `py_binary`
# 🐞 bug report

### Affected Rule

The issue affects the `compile_pip_requirements` rule/macro.

### Is this a regression?

I believe this is not a regression.

### Description

I'm trying to pass a custom certificate file to `compile_pip_requirements` via `extra_args` using `--cert=$(location ...)`.

The `compile_pip_requirements` macro accepts a `data` attribute, but it looks like this attribute is not forwarded to the internal `py_binary` target generated for the `.update` target.

As a result, using `$(location :pip_cert)` inside `extra_args` fails during analysis because the generated `py_binary` does not declare `:pip_cert` as a prerequisite.

## 🔬 Minimal Reproduction

```

filegroup(
    name = "pip_cert",
    srcs = ["pip.cert"],
)

compile_pip_requirements(
    name = "requirements",
    src = "requirements.in",
    extra_args = ["--cert=$(location :pip_cert)"],
    data = [":pip_cert"],
)
```

## 🔥 Exception or Error

<pre><code>
ERROR: .../BUILD.bazel:8:25: in args attribute of py_binary rule //:requirements.update: label '//:pip_cert' in $(location) expression is not a declared prerequisite of this rule. Since this rule was created by the macro 'pip_compile', the error might have been caused by the macro implementation
ERROR: .../BUILD.bazel:8:25: Analysis of target '//:requirements.update' (config: 7f8856b) failed
ERROR: Analysis of target '//:requirements.update' failed; build aborted
</code></pre>

## 🌍  Your Environment

**Operating System:** Ubuntu/MacOS.

<pre>
  <code>
$ bazel version
Build label: 8.5.1-homebrew
Build time: Tue Jan 01 00:00:00 1980 (315532800)
Build timestamp: 315532800
Build timestamp as int: 315532800
  </code>
</pre>

**Rules_python version:** tested with rules_python from the main branch, approximately version 2.0.3.

**Anything else relevant?**

I see two possible ways to address this:
1. Forward the `data` attribute to the generated `py_binary` target.
2. Add a dedicated label attribute, e. g. `ssl_cert`, and let the rule/macro generate the corresponding `--cert=$(location ...)` argument internally.

The first option seems important regardless of whether a dedicated certificate attribute is added. The `compile_pip_requirements` macro accepts `data`, so I would expect files listed there to be available to the generated executable target. Without that, using `data` together with `extra_args` is confusing.

The second option would make the certificate use case easier, but it would not fully solve the problem. `extra_args` may need to reference other files too, not only certificates. For that reason, forwarding `data` still seems necessary.

granatam:
As stated in #3858, the `compile_pip_requirements` macro accepts a `data` attribute, but this attribute is not forwarded to the internal `py_binary` target generated for the `.update` target.

As a result, using `$(location :pip_cert)` inside `extra_args` fails during analysis because the generated `py_binary` does not declare `:pip_cert` as a prerequisite.

Before:
```
ERROR: /.../BUILD.bazel:8:25: in args attribute of py_binary rule //:requirements.update: label '//:pip_cert' in $(location) expression is not a declared prerequisite of this rule. Since this rule was created by the macro 'pip_compile', the error might have been caused by the macro implementation
ERROR: /.../BUILD.bazel:8:25: Analysis of target '//:requirements.update' (config: 7f8856b) failed
ERROR: Analysis of target '//:requirements.update' failed; build aborted
```

After:
```
INFO: Analyzed target //:requirements.update (87 packages loaded, 4078 targets configured).
INFO: Found 1 target...
Target //:requirements.update up-to-date:
  bazel-bin/requirements.update
INFO: Elapsed time: 6.240s, Critical Path: 0.11s
INFO: 1 process: 11 action cache hit, 1 internal.
INFO: Running command line: bazel-bin/requirements.update '--src=rules_python_pip_parse_example/requirements.in' rules_python_pip_parse_example/requirements_lock.txt //:requirements '--resolver=backtracking' --allow-unsafe --generate-hashes '--cert=./pip.cert'
```

Closes #3858


python/private/pypi/pip_compile.bzl
```diff
@@ -100,11 +102,11 @@ def pip_compile(
     # for a requirements file that does `-r ../other/requirements.txt`
     native.filegroup(
         name = name,
-        srcs = kwargs.pop("data", []) + [requirements_txt],
+        srcs = data + [requirements_txt],
         visibility = visibility,
     )
 
-    data = [name, requirements_txt] + srcs + [f for f in (requirements_linux, requirements_darwin, requirements_windows) if f != None] + constraints
+    data = data + [name, requirements_txt] + srcs + [f for f in (requirements_linux, requirements_darwin, requirements_windows) if f != None] + constraints
```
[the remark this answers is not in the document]
granatam: Done

.bazelci/presubmit.yml
```diff
@@ -561,6 +561,15 @@ tasks:
...
     - "git diff --exit-code"
 
+  integration_test_compile_pip_requirements_update_data:
+    <<: *reusable_build_test_all
+    name: "compile_pip_requirements: .update data forwarding"
+    working_directory: tests/integration/compile_pip_requirements
+    platform: ubuntu2204
+    shell_commands:
+    - "bazel query 'labels(data, //:requirements.update)' | grep -Fx '//:requirements.in'"
+    - "bazel query 'labels(data, //:requirements.update)' | grep -Fx '//:requirements_extra.in'"
```
aignas: Could we use `genquery` for this target and not have the integration test at all?

python/private/pypi/pip_compile.bzl
```diff
@@ -95,16 +97,18 @@ def pip_compile(
...
```
aignas: So to better understand how this was not working previously was because we were including the `data` via the `filegroup` target via `[name, ...]` and not directly via files via `data`?

Out of curiosity, would the usecase have worked if the parameters did not use `$(location)`?. I would have thought that the files would have been included but the `$(location)` function in the `args` would not have worked correctly.

If my analysis is correct, please update the `news` file to hint about why thee fix was done.

granatam: > So to better understand how this was not working previously was because we were including the data via the filegroup target via [name, ...] and not directly via files via data?

Yep, you're right.

Btw, do we need `name` in the generated `py_binary` data at all? As I understand it, `name` is the default target used by other `compile_pip_requirements` rules. Since `requirements_txt` is already added separately, maybe it is enough to pass `data` and `requirements_txt` directly. I may be missing another reason for keeping `name` there though.

> Out of curiosity, would the usecase have worked if the parameters did not use `$(location)`?. I would have thought that the files would have been included but the `$(location)` function in the args would not have worked correctly.

Yes, that matches my understanding. Without `$(location)`, the files could still be available at runtime through the generated `filegroup`. The issue is that `$(location :pip_cert)` needs `:pip_cert` to be passed directly to the generated `py_binary`.

I'll update the `news` entry.

.bazelci/presubmit.yml
granatam: Yep, that would be better. 

Also, `genquery` revealed another issue: the `data` attribute does not allow duplicate labels. In the test `BUILD` file we have this:

```
compile_pip_requirements(
    name = "os_specific_requirements",
    src = "requirements_os_specific.in",
    data = [
        "requirements_extra.in",
        "requirements_os_specific.in",
    ],
    requirements_darwin = "requirements_lock_darwin.txt",
    requirements_linux = "requirements_lock_linux.txt",
    requirements_windows = "requirements_lock_windows.txt",
)
```

After forwarding the user-provided `data` directly to the generated `py_binary`, `requirements_os_specific.in` appears twice: once from `src` and once from `data`:

```
ERROR: /.../rules_python/tests/integration/compile_pip_requirements/BUILD.bazel:69:25: 
Label '//:requirements_os_specific.in' is duplicated in the 'data' attribute of rule 'os_specific_requirements.update'
```

I am not sure whether we should deduplicate these labels in the macro, or treat this as a user error and let Bazel report it. Currently, I lean toward deduplicating them in the macro.

aignas (approved):

aignas:
Could you please take a look at the build failures?

granatam:
> Could you please take a look at the build failures?

Fixed

rickeylev (approved):

Issue #3867, opened 2026-06-29
Release 2.2.0


rickeylev:
# Release tasks
- [x] Prepare Release | status=done pr=#3874 commit=ee0acbdc
- [x] Create Release branch | status=done branch_url=https://github.com/bazel-contrib/rules_python/tree/release/2.2 commit=ee0acbdc
- [x] Tag RC0 | status=done tag=2.2.0-rc0 commit= f7751e43
- [x] Tag RC1 | status=done tag=2.2.0-rc1 commit= b1fb23bc
- [x] Tag RC2 | status=done tag=2.2.0-rc2 commit= 721ca37b
- [x] Tag RC3 | status=done tag=2.2.0-rc3 commit= df4acbd8
- [x] Tag RC4 | status=done tag=2.2.0-rc4 commit= be0ae538
- [x] Tag Final | status=done tag=2.2.0 commit= be0ae538

## Backports

- [x] #3854 | status=done rc=rc1 commit= b1fb23bc
- [x] #3850 | status=done rc=rc2 commit= efa67e94
- [x] #3887 | status=done rc=rc3 commit= df4acbd8
- [x] #3890 | status=done rc=rc4 commit= be0ae538

---
*Maintainers: Automation will react to changes on this issue.*

<details>
<summary><b>Available Commands</b></summary>

Maintainers can trigger automation by running manual workflows:
- [Process Backports Workflow](https://github.com/bazel-contrib/rules_python/actions/workflows/process_backports.yml)
- [Generate RC Tag Workflow](https://github.com/bazel-contrib/rules_python/actions/workflows/generate_rc.yml)
- [Promote RC to Final Release Workflow](https://github.com/bazel-contrib/rules_python/actions/workflows/promote_rc.yml)
</details>

rickeylev:
edit: 2.2.0-rc0 was not actually tagged; script had a bug

rickeylev:
edit: nope, 2.2.0-rc0 was tagged, but release workflow failed

rickeylev:
**New Release Candidate Tagged!** 🐍🌿

Release Candidate **2.2.0-rc0** has been successfully generated and tagged on branch `release/2.2`.

- View Tag: [2.2.0-rc0](https://github.com/bazel-contrib/rules_python/releases/tag/2.2.0-rc0)
[...]

PR #3872, opened 2026-06-30
Prepare release v2.2.0

rickeylev:
Work towards #3867


rickeylev:
Woot. This one is correct, but going to close it to try out the workflow method of creation

PR #3851, opened 2026-06-26
feat: updates to the py_extension rule for building C extensions

rsartor-cmd:
This builds on the existing `py-extension` branch. It adds support for platform and abi tags in the resulting library filename. The `:py_extension_test`, `:py_extension_analysis_tests`, and `:py_limited_api_tests` test targets in `//tests/cc/py_extension` now build and pass.

Relates to https://github.com/bazel-contrib/rules_python/issues/3283

PR #3875, opened 2026-06-30
feat: updates to the py_extension rule for building C extensionsPy extension

rsartor-cmd:
Per feedback from #3851, this PR re-works the platform detection logic to rely on platform constraints instead of the Python runtime/toolchain. 

### Description

This PR contains updates to the experimental `py_extension` implementation. It re-works the platform tag detection to rely on modern platform constraints, and refactors the compilation and linking mechanism
to delegate to Bazel's native `cc_shared_library` and `cc_library` rules.

This PR targets the main repository's `py-extension` branch (not `main`).

#### 1. Platform Tag & ABI Tag Derivation
*   **Introduced `abi_tag` to `PyCcToolchainInfo`**: Added the `abi_tag` field to the `PyCcToolchainInfo` provider, populated by the `py_cc_toolchain` rule. It defaults to deriving the tag from
`python_version` (e.g. `cpython-311`) for backward compatibility.
*   **Constraint-Based Detection**: Replaced legacy parsing of `cc_toolchain` CPU names with direct lookup in rules_python's central `PLATFORMS` registry using platform constraints.
*   **Limited API Configuration**: Updated how limited API (`.abi3.so`) suffixes are appended based on the toolchain configuration.

#### 2. Compilation & Linking Delegation (Refactor to `cc_shared_library`)
*   **Macro Fusion**: Replaced the custom linking logic in the `py_extension` rule with a macro of the same name. It now accepts C/C++ source/header files directly (`srcs`, `hdrs`, `copts`, `defines`),
implicitly wrapping them in a private `cc_library` under the hood.
*   **Consolidated Attributes**: Consolidated static linkage under standard `deps` (removing the redundant `static_deps` attribute) and aligned linker arguments with `cc_shared_library`'s `user_link_flags`.
*   **Wrapper Rule (`_py_extension_wrapper`)**: Added a lightweight, private rule that wraps the `cc_shared_library` output to:
*   Query the Python toolchain for platform/ABI tags to resolve PEP 3149 compliant filenames (e.g. `.cpython-311-x86_64-linux-gnu.so` or `.abi3.so`).
*   Create a cheap `symlink` from the CSL output to the PEP 3149 name.
*   Propagate `PyInfo` (for python rules) and `CcSharedLibraryInfo` (for dynamic C++ dependencies).
*   **ODR Validation**: Leveraging native `cc_shared_library` ensures strict analysis-time validation against One Definition Rule (ODR) violations (e.g., duplicate static linkage in dynamic chains).

#### 3. Runfiles and Data Support
*   **Runfiles Propagation**: Re-implemented runfiles collection using the repository's standard `builders.RunfilesBuilder` to support runtime assets (`data` attribute) and dynamic library dependency
propagation.

### Verification & Testing
*   Added parity test cases in `dependency_graph_tests.bzl` to verify dynamic dependency chains and static sharing behavior.
*   Added runfiles validation in `py_extension_tests.bzl` to verify data asset propagation.
*   Updated limited API tests to match CSL output structure.
*   All 14 tests in `//tests/cc/py_extension/...` are compiling and passing.
──────
TAG=agy
CONV=23f8a5e8-2d99-401a-9903-1256b7d42b0e


rsartor-cmd:
The mypy CI failure appears to be caused by the check action using Python 3.9, but the newest mypy only supporting 3.10+

```
Run docker build /home/runner/work/_actions/jpetrucciani/mypy-check/master --build-arg "pyver=3.9" -t mypy
```

```
#11 3.773 ERROR: Ignored the following versions that require a different python version: 1.20.0 Requires-Python >=3.10; 1.20.1 Requires-Python >=3.10; 1.20.2 Requires-Python >=3.10; 2.0.0 Requires-Python >=3.10; 2.1.0 Requires-Python >=3.10
#11 3.774 ERROR: Could not find a version that satisfies the requirement mypy==2.1.0 (from versions: 0.1, 0.11, 0.12, 0.13, 0.14, 0.15, 0.16, 0.17, 0.18, 0.19, 0.20, 0.21, 0.221, 0.222, 0.223, 0.224, 0.225, 0.226, 0.227, 0.228, 0.229, 0.230, 0.231, 0.232, 0.233, 0.234, 0.235, 0.236, 0.237, 0.238, 0.239, 0.240, 0.241, 0.250, 0.251, 0.252, 0.253, 0.254, 0.255, 0.256, 0.470, 0.471, 0.501, 0.510, 0.511, 0.520, 0.521, 0.530, 0.540, 0.550, 0.560, 0.570, 0.580, 0.590, 0.600, 0.610, 0.620, 0.630, 0.641, 0.650, 0.660, 0.670, 0.700, 0.701, 0.710, 0.711, 0.720, 0.730, 0.740, 0.750, 0.760, 0.761, 0.770, 0.780, 0.781, 0.782, 0.790, 0.800, 0.812, 0.900, 0.901, 0.902, 0.910, 0.920, 0.921, 0.930, 0.931, 0.940, 0.941, 0.942, 0.950, 0.960, 0.961, 0.971, 0.981, 0.982, 0.990, 0.991, 1.0.0, 1.0.1, 1.1.1, 1.2.0, 1.3.0, 1.4.0, 1.4.1, 1.5.0, 1.5.1, 1.6.0, 1.6.1, 1.7.0, 1.7.1, 1.8.0, 1.9.0, 1.10.0, 1.10.1, 1.11.0, 1.11.1, 1.11.2, 1.12.0, 1.12.1, 1.13.0, 1.14.0, 1.14.1, 1.15.0, 1.16.0, 1.16.1, 1.17.0, 1.17.1, 1.18.1, 1.18.2, 1.19.0, 1.19.1)
#11 3.775 ERROR: No matching distribution found for mypy==2.1.0
#11 ERROR: process "/bin/sh -c pip install -U pip &&     pip install -r requirements.txt" did not complete successfully: exit code: 1
```

-----

For what it's worth, I ran `mypy --show-column-numbers --hide-error-context python/runfiles` and `mypy --show-column-numbers --hide-error-context tests/runfiles` locally and it found no issues with 3 files.

python/private/cc/py_extension_rule.bzl
```diff
@@ -51,26 +55,21 @@ def _py_extension_impl(ctx):
         user_link_flags.append("-Wl,--allow-shlib-undefined")
 
     ext = _get_extension(cc_toolchain)
-    use_py_limited_api = ctx.attr.py_limited_api and ctx.attr.py_limited_api != "none"
+    use_py_limited_api = bool(ctx.attr.py_limited_api)
```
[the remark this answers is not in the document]
rsartor-cmd: Yes, this is intended. Previously, `"none"` was the default used to indicate that the attribute was not set, and that the limited API should not be used. That default value has been changed to the empty string, which resolves to `False` when passed to `bool()`.

python/private/cc/py_extension_rule.bzl
```diff
@@ -159,30 +158,45 @@ PY_EXTENSION_ATTRS = COMMON_ATTRS | {
...
+            "@platforms//os:windows",
+            "@platforms//cpu:x86_64",
+            "@platforms//cpu:aarch64",
+            "@platforms//cpu:armv7",
+            "@platforms//cpu:i386",
+            "@platforms//cpu:ppc",
+            "@platforms//cpu:riscv64",
+            "@platforms//cpu:s390x",
+        ],
     ),
```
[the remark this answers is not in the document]
rsartor-cmd: Fixed.

python/private/cc/py_extension_rule.bzl
```diff
@@ -235,161 +236,72 @@ def _get_extension(cc_toolchain):
...
+            if c_label in constraints_map:
+                c_val = constraints_map[c_label]
+                if not ctx.target_platform_has_constraint(c_val):
+                    match = False
+                    break
+            else:
+                match = False
+                break
+        if match:
+            return _derive_pep3149_tag(platform, info)
```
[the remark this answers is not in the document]
rsartor-cmd: Fixed.

tests/cc/py_extension/BUILD.bazel
```diff
@@ -96,18 +96,18 @@ cc_library(
 
 py_extension(
     name = "ext_limited",
+    py_limited_api = "3.8",
     static_deps = [":ext_limited_impl"],
```
rickeylev: please rename static_deps to deps (can be done separate pr)

tests/cc/py_extension/dependency_graph_tests.bzl
```diff
@@ -0,0 +1,223 @@
...
+        dynamic_deps = [":" + name + "_cslB", ":" + name + "_cslC"],
+    )
+    analysis_test(
+        name = name,
+        target = name + "_cslA",
+        impl = _csl_dynamic_deps_test_impl,
+    )
+
+def _csl_dynamic_deps_test_impl(env, target):
+    env.expect.that_target(target).has_provider(CcSharedLibraryInfo)
```
mkruskal-google, line 76: This test seems a little light.  Don't we need to check more to ensure it actually linked everything correctly?  Same for the others

rsartor-cmd: Yes, I can add tests for that. These tests were in the earlier stages, when I was working out what providers were necessary to return.

rsartor-cmd: @mkruskal-google Done. How does that look? https://github.com/bazel-contrib/rules_python/pull/3875/changes/eb4b8a8981125b77ac7df1a859933fa8b5d38357

python/private/cc/py_extension_rule.bzl
```diff
@@ -134,88 +61,45 @@ def _py_extension_impl(ctx):
             transitive_sources = depset([py_dso]),
             imports = depset([import_path]),
         ),
-        propagated_cc_info,
+        csl_target[CcSharedLibraryInfo],
```
rickeylev: Why do we propagate CcSharedLibraryInfo? It's so that the py extension can be given as a dynamic dep to another cc shared library? I'm not sure that makes sense -- lets discuss in chat.

tests/cc/py_extension/BUILD.bazel
```diff
@@ -14,34 +15,47 @@ package(
...
-    name = "ext_shared",
-    dynamic_deps = [
-        ":add_one_shared",
-    ],
-    static_deps = [
-        ":ext_shared_impl",
-    ],
+    # An extension that also depends on a data file
+    name = "ext_with_data",
+    data = ["test_symbols.h"],
```
rickeylev: Lets use a non-code type of file for a data dependency, i.e. somedata.txt

tests/cc/py_extension/BUILD.bazel
```diff
@@ -94,25 +102,67 @@ cc_library(
...
+    srcs = ["ext_init_in_dep.c"],
+    deps = [
+        "@rules_python//python/cc:current_py_cc_headers",
+    ],
+)
+
+#####
+
+py_extension(
+    # An extension with its PyInit_* function in a dynamic dependency
```
rickeylev: Don't support this case, or rather, don't add a test for it -- it doesn't make any sense. The PyInit symbol is supposed to be provided by the foo.so module being loaded. If its provided by a shared library, that would imply that others are supposed to load that shared library and also get that symbol -- but that symbol is supposed to be unique to the loaded foo.so module

python/private/cc/py_extension_rule.bzl
```diff
@@ -134,88 +61,45 @@ def _py_extension_impl(ctx):
...
+    ),
+    "src": lambda: attrb.Label(
+        mandatory = True,
+        providers = [CcSharedLibraryInfo],
+        doc = "The cc_shared_library target to wrap.",
+    ),
+    "_constraints": lambda: attrb.LabelList(
+        default = sorted({
+            c: None
+            for info in PLATFORMS.values()
```
rickeylev: The PLATFORMs global is legacy and shouldn't be used. The main problem is it requires every platform to be defined in rules_python, but that doesn't scale.

python/private/cc/py_extension_rule.bzl
```diff
@@ -1,129 +1,56 @@
...
         )
     else:
-        py_toolchain = ctx.toolchains[TARGET_TOOLCHAIN_TYPE]
-        py_runtime = py_toolchain.py3_runtime
-        cc_toolchain = ctx.toolchains["@bazel_tools//tools/cpp:toolchain_type"].cc
-        platform_tag = _get_platform(cc_toolchain)
-        output_filename = "{module_name}.{pyc_tag}{abi_flags}-{platform}.{ext}".format(
+        py_toolchain = ctx.toolchains[PY_CC_TOOLCHAIN_TYPE]
+        py_cc_toolchain = py_toolchain.py_cc_toolchain
+        platform_tag = _get_platform(ctx)
```
rickeylev, line 26: I'm not sure the platform part of the infix calculation here correct.

It's constructing a name like `cpython-311t-win_arm64`, but I don't see the last portion, e.g. "win_arm64", specified as part of PEP 3149. The OS/Arch is part of the wheel filename formats for sure, but this is the filename for C module, not wheel.

I'm pretty sure i've seen e.g. `gnu-linux` in filenames python searches for, however, I can't seem to find a spec/pep that describes that behavior and why it looks for the names it does.

python/private/cc/py_extension_macro.bzl
```diff
@@ -1,33 +1,96 @@
...
+        )
+        csl_deps.append(":" + impl_lib_name)
+
+    # 3. If no static deps or sources were specified, use empty target for CSL requirement
+    if not csl_deps:
+        csl_deps.append("//python/private/cc:empty")
+
+    # 4. Create the underlying cc_shared_library
+    csl_name = "_" + name + "_csl"
+    csl_kwargs = {}
```
rickeylev: There's a small collection of attributes (visiblity, compatible_with, some others) that need to be copied into depended-upon generated targets. I think there's a helper function in utils for this; a todo for now is fine.

python/private/cc/py_extension_rule.bzl
rickeylev, line 74: Also set the "has shared libraries" field (I forget its exact name)

python/private/cc/py_extension_rule.bzl
```diff
@@ -1,129 +1,56 @@
...
-    # to treat it as undefined causes it to be retained.
-    user_link_flags.append("-Wl,--undefined=PyInit_{module_name}".format(
-        module_name = module_name,
-    ))
-
-    if ctx.attr.external_deps:
-        user_link_flags.append("-Wl,--allow-shlib-undefined")
+    import_path = repo_name
+    if ctx.label.package:
+        import_path = repo_name + "/" + ctx.label.package
```
rickeylev: Setting import path here is incorrect. It's going to add the directory the .so file is in to the python sys.path search, which makes it directly importable. However, this is wrong for two reasons:
1. A foo.so file can be part of package. In whih case, `import pkg.foo` is how to import it, and `import foo` would be incorrect
2. If it is part of package, all py files in the directory become directly importable as top-level modules. e.g `foo/bar.py` becomes importable as `import bar`. This is incorrect, as it breaks the identify modules have, causes duplicate code loading, and can shadow other top-level imports.

python/private/py_cc_toolchain_rule.bzl
```diff
@@ -45,7 +45,14 @@ def _py_cc_toolchain_impl(ctx):
     else:
         headers_abi3 = None
 
+    abi_tag = ctx.attr.abi_tag
+    if not abi_tag:
+        # Derive default: cpython-XX
```
rickeylev: This calculation is incomplete: it also needs to take into account freethreadedness.

You can copy how py_runtime computes it. I think it has an abiflags attribute? I can't remember exactly.

python/private/cc/py_extension_macro.bzl
```diff
@@ -1,33 +1,96 @@
...
+            "@rules_python//python/config_settings:_is_py_linux_libc_glibc": "glibc",
+            "@rules_python//python/config_settings:_is_py_linux_libc_musl": "musl",
+            "//conditions:default": "glibc",
+        })
+
+    if data != None:
+        kwargs["data"] = data
+
+    # 6. Wrap with py_extension_wrapper for PEP 3149 naming & PyInfo
+    py_extension_wrapper(
```
rickeylev, line 121: Since we now have an internal-only wrapper rule we can pass extra arbitrary internal-use-only args, I think we can simplify how the platform info is passed in: have os and arch attributes, then use select. This also makes it easy to map bazel's conditions to the necessary python value

```
os = select({
  "@platforms//os:windows": "win",
  "@platforms//os:osx": "darwin",
})
arch = select({
  "@platforms//cpu:x86_64": ...,
  ...
})
```

python/private/cc/py_extension_rule.bzl
rsartor-cmd: Previously, it was returning CcInfo as a provider, which didn't work because 1) a cc_shared_library doesn't provide that anyways, and 2) no-one depending on a py_extension would need to know the compiler details or object code because it's not statically linked into other libraries. So we changed it from CcInfo to CcSharedLibraryInfo, which is what the cc_shared_library provides. I don't recall everything that we said about CcSharedLibraryInfo. But yes, the de-duplication is necessary in certain cases. The AI informs me that some large libraries (e.g. numpy, scipy) will have a central core extension plus several satellite extension libraries (e.g. linalg, fft, etc). In that situation, if we're exporting a CcSharedLibraryInfo, the py_extensions will participate in the de-duplication across the transitive dependency graph.

tests/cc/py_extension/BUILD.bazel
rsartor-cmd: Already done.

tests/cc/py_extension/BUILD.bazel
rsartor-cmd: Makes sense. Done.

tests/cc/py_extension/BUILD.bazel
rsartor-cmd: Removed.

python/private/cc/py_extension_rule.bzl
rsartor-cmd: PEP 3149 only covers platform tags for Unix .so files. My AI said something about the Windows platform tags being defined in a GitHub issue, but I suspect that was a hallucination. I'm struggling to find an official reference that documents windows platform tags. I just checked EXT_SUFFIX on a windows machine running CPython 3.14 and it says `.cp14-win_amd64.pyd`, so it's definitely a thing. I just haven't found the official decision yet.

python/private/cc/py_extension_rule.bzl
rsartor-cmd: I think I've fixed it. I added a couple of parameters to the wrapper rule for os and CPU, and used a `select()` to get a value based on the relevant constraints.

It's possible that we could get os and cpu directly from the tool chain, instead of the constraints. Would that be valuable?

Also, I could add a parameter to the macro so users could specify the platform tags they want, in those rare cases when they're targeting some os or cpu that we don't have defined yet.

python/private/cc/py_extension_rule.bzl
rickeylev: I asked a python core dev -- there's not standard for this. So just match what we observe and we can try to make it as clean as make sense

rsartor-cmd:
Attempting a rebase to fix the CI failures.

rsartor-cmd:
Un-rebased

rsartor-cmd:
@rickeylev Would it be okay if I rebased your py-extension branch (the base of this PR) onto main? That, together with rebasing my branch would fix most if not all of the CI failures, I think. I already did a dry-run locally and there was only a couple of conflicts which were basically nothing.

python/private/py_cc_toolchain_macro.bzl
```diff
@@ -30,4 +30,23 @@ def py_cc_toolchain(**kwargs):
...
+            "@platforms//os:macos": "macos",
+            "@platforms//os:windows": "windows",
+            "//conditions:default": "linux",
+        })
+    if "cpu" not in kwargs:
+        kwargs["cpu"] = select({
+            "@platforms//cpu:aarch64": "aarch64",
+            "@platforms//cpu:x86_32": "x86_32",
+            "//conditions:default": "x86_64",
+        })
```
rickeylev, line 45: For os and cpu, lets expose the python-specific terminology and values for them on the toolchain (I think this is why PyRuntimeInfo does). There's several (os.name, platform.something, sys.platform, sysconfig values etc). I'm not sure which of them is most approrpriate, but we can defer that to another PR

python/private/cc/py_extension_rule.bzl
```diff
@@ -119,86 +129,21 @@ def _get_extension(cc_toolchain):
...
 
     Returns:
         The platform tag, e.g. "x86_64-linux-gnu" or "win_amd64"
     """
-    platform_tag = _get_platform_from_constraints(ctx)
-    if platform_tag:
-        return platform_tag
+    py_toolchain = ctx.toolchains[PY_CC_TOOLCHAIN_TYPE]
+    py_cc_toolchain = py_toolchain.py_cc_toolchain
+    if hasattr(py_cc_toolchain, "platform_tag") and py_cc_toolchain.platform_tag:
```
rickeylev, line 143: Remove the hasattr check. You might see such checks on the PyruntimeInfo, but that's only because there's a legacy bazel-builtin definition that sometimes gets used. The PyccToolchain doesn't have that complication, so we can assume the shape is consistent.

rickeylev (approved):
just a couple small changes requested, but since we need to rebase atop main and this has been out awhile, lets just merge now. it ok if ci is failing in this WIP branch

PR #3887, opened 2026-07-02
fix: use locale-neutral Windows ACL for build data

### LINKED ISSUE #3886: Windows build data generation fails on localized installs due to "Everyone" ACL lookup
# 🐞 bug report

### Affected Rule

`py_binary` / `py_test` on Windows when rules_python writes build data via
`python/private/build_data_writer.ps1`.

### Is this a regression?

Introduced by PR #3604 - and first rolled out in 1.9.0

### Description

On localized Windows installations, build data generation can fail because
`python/private/build_data_writer.ps1` creates a file ACL rule using the English
localized account name `"Everyone"`.

That works on English Windows, but fails on German and other localized Windows
installations where the well-known Everyone principal has a localized display
name. On German Windows, the identity is displayed as `Jeder`.

The script currently does:

```powershell
$AccessRule = New-Object System.Security.AccessControl.FileSystemAccessRule("Everyone", "Read", "Allow")
```

This should use the language-neutral well-known SID for Everyone instead of the
localized display name.

## 🔬 Minimal Reproduction

On a German Windows installation, run a target that causes rules_python to write
build data, for example in the rules_python repository:

```powershell
bazel test //tests/build_data:build_data_test --config=fast-tests
```

The failure can also be reproduced directly with the ACL snippet from
`python/private/build_data_writer.ps1` on a German Windows host.

## 🔥 Exception or Error

<pre><code>
Ausnahme beim Aufrufen von "SetAccessRule" mit 1 Argument(en):  "Manche oder alle Identitätsverweise konnten nicht
übersetzt werden."
In
&lt;redacted&gt;\execroot\_main\external\rules_python+\python\private\build_data_writer.ps1:25
Zeichen:1
+ $Acl.SetAccessRule($AccessRule)
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [], MethodInvocationException
    + FullyQualifiedErrorId : IdentityNotMappedException
</code></pre>

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
Windows with German locale
  </code>
</pre>

<pre>
  <code>
Bazel 9.1.1
  </code>
</pre>

**Rules_python version:**

<pre>
  <code>
current main / 2.1.0
  </code>
</pre>

**Anything else relevant?**

The Windows well-known Everyone principal can be referenced using the
language-neutral SID `S-1-1-0`, or in .NET via
`System.Security.Principal.WellKnownSidType.WorldSid`.

Using `WorldSid` preserves the existing behavior, granting read access to
Everyone, while avoiding localized account-name lookup failures.

alex-the-third:
On localized Windows installations, `build_data_writer.ps1` could fail when
creating the output file ACL because it used the English localized `Everyone`
principal name.

This changes the ACL rule to use the language-neutral `WorldSid` identity
instead, preserving the existing Everyone read permission without depending on
the display language of Windows.

Before: German and other localized Windows installs could fail with
`IdentityMappedException` / `IdentityNotMappedException`.

After: build data generation succeeds on localized Windows installs.

Fixes #3886

Tests:
`bazelisk test //tests/build_data:build_data_test --config=fast-tests`

### DIFF
--- a/news/3886.fixed.md
+++ b/news/3886.fixed.md
@@ -1,0 +1,1 @@
+(windows) Fixed build data generation on localized Windows installations.
--- a/python/private/build_data_writer.ps1
+++ b/python/private/build_data_writer.ps1
@@ -21,7 +21,15 @@ $Utf8NoBom = New-Object System.Text.UTF8Encoding $False
 [System.IO.File]::WriteAllLines($OutputPath, $Lines, $Utf8NoBom)
 
 $Acl = Get-Acl $OutputPath
-$AccessRule = New-Object System.Security.AccessControl.FileSystemAccessRule("Everyone", "Read", "Allow")
+$EveryoneSid = New-Object System.Security.Principal.SecurityIdentifier(
+    [System.Security.Principal.WellKnownSidType]::WorldSid,
+    $null
+)
+$AccessRule = New-Object System.Security.AccessControl.FileSystemAccessRule(
+    $EveryoneSid,
+    "Read",
+    "Allow"
+)
 $Acl.SetAccessRule($AccessRule)
 Set-Acl $OutputPath $Acl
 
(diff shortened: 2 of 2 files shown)


rickeylev:
TIL.

Thanks for the fix!

rickeylev (approved):

PR #3917, opened 2026-07-10
fix(sphinxdocs): invalidate cache on branch change

kaycebasques:
We had to disable the `sphinxdocs` persistent worker in Pigweed (https://pwbug.dev/493749800) because the cache invalidation logic seemed incomplete with regards to Git branches. E.g. in branch `feat` you create a new doc `foo.rst`. You switch back to branch `main` and try to build Sphinx, but Sphinx warns about needing to ignore unreadable document `foo.rst`. Because Pigweed builds Sphinx with `--fail-on-warning` this would break our docs build. This PR adds the bug fix and also starts a suite of tests that will be focused on persistent worker correctness.


sphinxdocs/tests/persistent_worker/git_test.py
```diff
@@ -0,0 +1,213 @@
...
+)
+""")
+
+    def test_git_branch_workflow(self):
+        with tempfile.TemporaryDirectory() as tmpdir:
+            repo = pathlib.Path(tmpdir)
+            self._create_test_workspace(repo)
+
+            # 1. Initialize git repo and make initial commit on main
+            self._run_cmd(["git", "init"], cwd=repo)
```
rickeylev: Let's remove using git. Insofar as bazel and sphinx are concerned, all its doing it change the content of files. That can be easily done by just directly modifying the files.

kaycebasques:
temporary artifact (`build.log`) needs to be deleted

sphinxdocs/tests/persistent_worker/BUILD.bazel
```diff
@@ -0,0 +1,40 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+load("@rules_python//python:py_test.bzl", "py_test")
+
+py_test(
+    name = "git_test",
```
rickeylev: Lets create a top-level integration_tests directory instead for heavy weight integration tests. This just makes it easier to the fast, regular tests using `bazel test //tests/...` and skip any integration tests

sphinxdocs/tests/persistent_worker/git_test.py
```diff
@@ -0,0 +1,213 @@
...
+            "common --disk_cache=\n"
+            "build --noremote_accept_cached\n"
+            "common --http_timeout_scaling=10.0\n"
+            "common --experimental_repository_downloader_retries=10\n"
+            f"common --downloader_config={downloader_cfg}\n"
+            "common --lockfile_mode=off\n"
+        )
+
+        sphinxdocs_root_str = str(self.sphinxdocs_root).replace("\\", "/")
+        (repo / "MODULE.bazel").write_text(f"""\
```
rickeylev: Use rules_bazel_integration_test (i think its called) instead. rules_python has some examples. we can just copy the code from there. It's a bit nicer because the sub-workspace are regular files, so you can cd directly into them and run builds, which makes it easier to reproduce issues. The outer test then becomes just a driver that modifies state and run commands in the sub-workspace.

rickeylev (changes requested):

rickeylev (approved):
I've refactor a bunch to address my comments
* Switched it to rules_bazle_integration_test
* removed using git
etc

kaycebasques:
Woooo thanks @rickeylev

PR #3919, opened 2026-07-11
fix: make executable implicit defaults public so other modules can use rule builders

13steinj:
`create_executable_rule_builder()` (used by `py_binary_rule_builder()` /
`py_test_rule_builder()`, the public `python/api/executables.bzl` API)
bundles three implicit attrs whose defaults point at private targets
under `//python/private`: `build_data_writer`,
`debugger_if_target_config`, and `uncachable_version_file`. These
targets had no explicit visibility, so they inherited the package's
`default_visibility` (`//:__subpackages__`), which only covers packages
inside the rules_python repo itself.

Because the builder API is designed to let external modules call
`rule()` themselves (via `builder.build()`), Bazel checks visibility of
these attr defaults from the *calling* module's package, not from
rules_python's. Any external repo constructing a rule via
`py_binary_rule_builder()` / `py_test_rule_builder()` therefore fails at
analysis time with a visibility error.

This has been broken since the builder API's introduction; there's prior
art in this same file for exactly this situation (e.g.
`stage1_bootstrap_template`, among others)

### DIFF
--- a/.bazelrc.deleted_packages
+++ b/.bazelrc.deleted_packages
@@ -7,6 +7,7 @@ common --deleted_packages=examples/bzlmod/entry_points/tests
 common --deleted_packages=examples/bzlmod/libs/my_lib
 common --deleted_packages=examples/bzlmod/other_module
 common --deleted_packages=examples/bzlmod/other_module/other_module/pkg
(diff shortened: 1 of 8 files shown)


rickeylev:
This looks great. Thanks for the fix!

aignas (approved):

PR #3933, opened 2026-07-17
fix(sphinxdocs): clean stale worker output dir and doctrees on first request

MartinNeudecker:
The Sphinx worker keeps doctrees outside declared outputs so that incremental
builds survive across invocations. A newly started worker has no digest history,
so it treats every input as changed and re-reads all docs. Re-reading against
the stale environment left by a previous worker can produce spurious warnings,
e.g. duplicate labels, which become build failures when `--fail-on-warning`
is enabled.

On the first request, delete the worker output directory and doctree directory
so the worker starts from a clean Sphinx environment.

### DIFF
--- a/sphinxdocs/sphinxdocs/private/sphinx_build.py
+++ b/sphinxdocs/sphinxdocs/private/sphinx_build.py
@@ -128,6 +128,7 @@ class Worker:
 
         incoming_digests = {}
         current_digests = self._digests.setdefault(srcdir, {})
+        is_first_request = not current_digests
         changed_paths = []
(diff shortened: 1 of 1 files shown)


rickeylev:
Ah, I think I understand now. I was confused because bazel clears output directories, _but_ the doctree dir isn't the output directory. It's `--doctree-dir={output_dir}_doctrees` (sibling of the directory). Hence its state is undefined in between worker starts or invocations.

(ideally there'd be a way to restore the state, but i think clearing things after a restart makes sense. It already does that in another case if exit code 2 comes back from sphinx)

rickeylev (approved):
Thanks for the fix!

PR #3935, opened 2026-07-18
fix(pypi): handle absolute file URLs in wheel downloader

aignas:
This fixes the normalization function to return correctly normalized
file URLs which is the minimum that we need to get absolute file
URLs handled correctly.

Related #3312


rickeylev (approved):
ai review comments look legit

PR #3939, opened 2026-07-20
test(tools): auto-patch release CLI helpers in conftest.py

rickeylev:
Add an autouse fixture in conftest.py for tests/tools/private/release that automatically patches run_cmd, Git._run_git, and GitHub._run_gh using pytest-mock. This prevents tests from executing command line tools that could have side-effects.

### DIFF
--- a/tests/tools/private/release/conftest.py
+++ b/tests/tools/private/release/conftest.py
@@ -3,18 +3,34 @@ import pytest
 pytest_plugins = ["tests.tools.private.release.release_test_helper"]
 
 
+@pytest.fixture(name="mock_run_cmd")
+def fixture_mock_run_cmd(mocker):
+    """Fixture to patch shell.run_cmd and its imports in git and gh modules."""
+    mock = mocker.patch("tools.private.release.shell.run_cmd")
+    mocker.patch("tools.private.release.git.run_cmd", mock)
+    mocker.patch("tools.private.release.gh.run_cmd", mock)
+    return mock
+
+
+@pytest.fixture(name="mock_run_git")
+def fixture_mock_run_git(mocker):
+    """Fixture to patch Git._run_git."""
+    return mocker.patch("tools.private.release.git.Git._run_git")
+
+
+@pytest.fixture(name="mock_run_gh")
+def fixture_mock_run_gh(mocker):
+    """Fixture to patch GitHub._run_gh."""
+    return mocker.patch("tools.private.release.gh.GitHub._run_gh")
+
+
 @pytest.fixture(name="auto_patch_cmd_helpers", autouse=True)
-def fixture_auto_patch_cmd_helpers(mocker):
+def fixture_auto_patch_cmd_helpers(mock_run_cmd, mock_run_git, mock_run_gh):
     """Automatically patches run_cmd, Git, and GitHub CLI helpers.
 
     This prevents tests from executing command line tools that could have
     side-effects.
     """
-    mock_run_cmd = mocker.patch("tools.private.release.shell.run_cmd")
-    mocker.patch("tools.private.release.git.run_cmd", mock_run_cmd)
-    mocker.patch("tools.private.release.gh.run_cmd", mock_run_cmd)
-    mock_run_git = mocker.patch("tools.private.release.git.Git._run_git")
-    mock_run_gh = mocker.patch("tools.private.release.gh.GitHub._run_gh")
     return {
         "run_cmd": mock_run_cmd,
         "run_git": mock_run_git,
--- a/tests/tools/private/release/conftest.py
+++ b/tests/tools/private/release/conftest.py
@@ -1,6 +1,18 @@
+import dataclasses
+from unittest.mock import MagicMock
+
 import pytest
 
 pytest_plugins = ["tests.tools.private.release.release_test_helper"]
+
+
+@dataclasses.dataclass
+class AutoPatchCmdHelpers:
+    """Dataclass holding mocked command helpers."""
+
+    run_cmd: MagicMock
+    run_git: MagicMock
+    run_gh: MagicMock
 
 
 @pytest.fixture(name="mock_run_cmd")
@@ -31,8 +43,8 @@ def fixture_auto_patch_cmd_helpers(mock_run_cmd, mock_run_git, mock_run_gh):
     This prevents tests from executing command line tools that could have
     side-effects.
     """
-    return {
-        "run_cmd": mock_run_cmd,
-        "run_git": mock_run_git,
-        "run_gh": mock_run_gh,
-    }
+    return AutoPatchCmdHelpers(
+        run_cmd=mock_run_cmd,
+        run_git=mock_run_git,
+        run_gh=mock_run_gh,
+    )
--- a/tests/tools/private/release/conftest.py
+++ b/tests/tools/private/release/conftest.py
@@ -1,0 +1,22 @@
+import pytest
+
+pytest_plugins = ["tests.tools.private.release.release_test_helper"]
+
+
+@pytest.fixture(name="auto_patch_cmd_helpers", autouse=True)
+def fixture_auto_patch_cmd_helpers(mocker):
+    """Automatically patches run_cmd, Git, and GitHub CLI helpers.
+
+    This prevents tests from executing command line tools that could have
+    side-effects.
+    """
+    mock_run_cmd = mocker.patch("tools.private.release.shell.run_cmd")
+    mocker.patch("tools.private.release.git.run_cmd", mock_run_cmd)
+    mocker.patch("tools.private.release.gh.run_cmd", mock_run_cmd)
+    mock_run_git = mocker.patch("tools.private.release.git.Git._run_git")
+    mock_run_gh = mocker.patch("tools.private.release.gh.GitHub._run_gh")
+    return {
+        "run_cmd": mock_run_cmd,
+        "run_git": mock_run_git,
+        "run_gh": mock_run_gh,
+    }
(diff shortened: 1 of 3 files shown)


rickeylev:
@aignas any thoughts on conftest.py integration? In this PR, a conftest.py is added to a py_library that defines a fixture that should be automatically used, and things just magically work. Which is kind of neat.

Originally, though, I was thinking that pytest_test would glob conftest.py by default. I think that better matches pytest semantics?

aignas (approved):

aignas:
I agree, maybe just globbing `conftest.py` is indeed a good idea. Though `gazelle` has some logic of including the `conftest.py`, so the two would be clashing. It would be good to ask @dougthor42 about this as well.

tests/tools/private/release/conftest.py
```diff
@@ -0,0 +1,22 @@
+import pytest
+
+pytest_plugins = ["tests.tools.private.release.release_test_helper"]
+
+
+@pytest.fixture(name="auto_patch_cmd_helpers", autouse=True)
+def fixture_auto_patch_cmd_helpers(mocker):
```
dougthor42: Optional Nit: More idiomatic `pytest` would be to have a single fixture for each item, and then a 4th fixture to collect them all.

tests/tools/private/release/conftest.py
```diff
@@ -0,0 +1,22 @@
...
+
+    This prevents tests from executing command line tools that could have
+    side-effects.
+    """
+    mock_run_cmd = mocker.patch("tools.private.release.shell.run_cmd")
+    mocker.patch("tools.private.release.git.run_cmd", mock_run_cmd)
+    mocker.patch("tools.private.release.gh.run_cmd", mock_run_cmd)
+    mock_run_git = mocker.patch("tools.private.release.git.Git._run_git")
+    mock_run_gh = mocker.patch("tools.private.release.gh.GitHub._run_gh")
+    return {
```
dougthor42: I recommend returning a dataclass instead of dict - helps with type checking (do we do type checking in this project? 🤣 )

tests/tools/private/release/conftest.py
dougthor42: Might be out of scope, but why use `pytest_mock` over just plain-ol `unittest.mock` API? Or pytest's built in [`monkeypatch` API](https://docs.pytest.org/en/latest/how-to/monkeypatch.html)?

tests/tools/private/release/gh_test.py
```diff
@@ -59,3 +59,23 @@ def test_resolve_pr_number_invalid(mocker, gh):
     with pytest.raises(ValueError, match="Could not resolve PR reference"):
         gh.resolve_pr_number("invalid-ref")
     mock_run_cmd.assert_not_called()
+
+
+def test_auto_patched_helpers_prevent_real_execution(auto_patch_cmd_helpers):
+    from tools.private.release.gh import GitHub
+    from tools.private.release.git import Git
+    from tools.private.release.shell import run_cmd
```
dougthor42: The `gh_test` target needs to be updated to include these as deps, no?

dougthor42 (approved):
>> I was thinking that pytest_test would glob conftest.py by default
>
> Though gazelle has some logic of including the conftest.py, so the two would be clashing

Agreed, gazelle already has [`python_include_ancestor_conftest`](https://rules-python.readthedocs.io/en/latest/gazelle/docs/directives.html#python-include-ancestor-conftest) so it would be a little strange for `pytest_test` to glob all of them.

---

For the record, we (Quantum) have actively been moving _away_ from using pytest's conftest magic precisely because goes against the "explicit dependencies" Bazel idiom.

We set both directives `python_include_ancestor_conftest=False` and `include_pytest_conftest=False` because our pre-Bazel conftest files had accumulated loads of cruft, and would pollute the dependency tree: a change in file A would bust the conftest.py cache and then bust the cache of every downstream target, even if those targets didn't use the fixtures defined in that conftest.

The [docs for `python_include_ancestor_conftest`](https://rules-python.readthedocs.io/en/latest/gazelle/docs/directives.html#python-include-ancestor-conftest) shows an example of 4 different conftest files getting added as dependencies. This mimics what pytest does under the hood, but it also means that `my_test.py`'s cache is much more fragile. If `my_test.py` doesn't use any of the fixtures defined in `//:conftest` (even auto-used ones), then why does it depend on `//:conftest`?

---

This PR looks reasonable to me. All comments are optional / questions.

tests/tools/private/release/conftest.py
aignas: There is a `mypy` test for the `runfiles` but not more, I think.

tests/tools/private/release/conftest.py
rickeylev: With the unittest.mock decorators, I can never keep track of which decorator call goes to which arg. When using the inline `with patch(...)`, it adds extra indents which is annoying. If you don't use the context manager, then you risk forgetting to unmock.

With monkeypatch, it looked like a more basic patching API, not one that could handle mocks (e.g. that could capture args, assert them later, etc).

Using the mocker object, it started to resemble the `mock = self.enter_context(patch(...))` code as seen in absltest, which I really like (auto cleanup and no lose of identation)

rickeylev: > Optional Nit: More idiomatic `pytest` would be to have a single fixture for each item, and then a 4th fixture to collect them all.

Done

tests/tools/private/release/conftest.py
rickeylev: > I recommend returning a dataclass instead of dict - helps with type checking (do we do type checking in this project? 🤣 )

Done

tests/tools/private/release/gh_test.py
rickeylev: > The `gh_test` target needs to be updated to include these as deps, no?

Argh, yeah. and imports moved to top. dumb bot. Done.

rickeylev:
> auto-including conftest and including parent conftest is more idiomatic but makes bazel cache fragile

Hrm, yeah, these are good points. I guess the ideal solution would be to factor the directory structure into a more well-defined tree? Which, well, that's just a fancy way of saying "auto-include; cache fragility is the user's fault"

e.g.

```
squares/
  conftest.py # fixture_quadralateral, fixture_square
  rectangles/
    conftest.py # fixture_rectangle

-> change to

shapes/
  conftest.py # fixture_quadralateral
  squares/
    conftest.py # fixture_square
  rectangles/
    conftest.py # fixture_rectangle
```

I wonder if `pytest_conftest_library()` is needed? (though I guess that's basically just py_library)

Something I'm not clear on is what target `conftest.py` is supposed to go into. Right now, it all works because conftest.py is in the release_test_helper target, which is a dependency of all (supposedly) the tests. I guess more ideally it should be its own py_library target, and then every test should directly depend on it?

tests/tools/private/release/conftest.py
dougthor42: ACK. I agree with the mock.patch decorators - the decorator to arg mapping is a PITA. Personally I don't mind the indents of the CM.

But yes that's solid reasoning, thanks for explaining.

dougthor42 (commented):
> factor the directory structure into a more well-defined tree? 

Yeah factoring the dir structure is one solution, but IME is typically not very feasible. It's really easy to end up over-defined and not know where to put anything new or where to put something that needs **both** `fixture_square` and `fixture_rectangle` (which is probably a sign you need to refactor your fixtures).

> I wonder if `pytest_conftest_library()` is needed?

I don't think so. As you said, it's just a normal `py_library`. If it were to exist, how would it be different than `py_library`? If the only difference is "magic to add the other conftests as deps" then I'd be pretty strongly opposed because that would hide the explicit deps.

> Where should conftest.py go?

A `py_library` as a dependency of any `py_test` that actually needs it. Old-school gazelle default behavior was to add the dep to any sibling `py_test` targets, but as mentioned above this leads to extraneous deps. Personally I use the `# gazelle:include_dep :conftest` annotation instead.

I'm also a big fan of "one file, one target", so I'm partial to a dedicated `py_library` for conftest. Given that you have `autouse=True` inside, all sibling and child python test files implicitly depended on it. So I would say yes each `py_test` target should have it as a **direct dependency** instead of indirect through `release_test_helper`. Use Bazel to turn the implicit dep to explicit.

aignas:
+1 for having `testonly = True` `conftest`-only `py_library`.

rickeylev:
> py_library(conftest) ... direct dep of test instead of helper library

I'll open a PR for this. I'm fine with this, but it does make me ask:

Adding the conftest.py lib to every test is somewhat tedious and easy-ish to forget or overlook. Aside from using gazelle to manage everything, is there anything we could do in the rules to make it less error-prone to forget?

To which I answer...I guess that leads back to auto-inclusion? I suppose a `native.existing_rule()` could check for a `conftest` target and add it to deps or something?  Just thinking out loud.

> refactoring the dir structure isn't always feasible

Yeah, understood and I agree.

aignas:
There are 2 possibilities here:
1. Have it explicitly included, which is tedious, but is not magical.
2. Have it implicitly included, which is magical, but less tedious.

Depending on whether it is a big project with strict rules or a smaller project with more fluidity, the preference will be different. Then maybe we should support both things as a config option? But then what are you configuring? We could have a label flag for this behaviour to include or exclude the `existing_rule("conftest")`. That said, this is a package definition concern, so it should not be dynamically configurable but statically configurable.

If it is statically configurable and it is not gazelle then I see only to options:
1. bool parameter.
2. Rely on users using [with_cfg.bzl](https://github.com/fmeum/with_cfg.bzl)
3. different loads depending on whether you want to use inclusive or exclusive version.

I would vote for 2, but this is just a version of 3 and 1 is not super good.

PR #3945, opened 2026-07-22
tests: convert private tool tests to pytest_test and pytest style

rickeylev:
Convert private tool tests (`tests/tools/zipapp/` and
`tools/private/update_deps/`) to `pytest_test` and pytest style to
be more idiomatic Python tests.

Also converts to use pytest fixtures.


rickeylev:
Nothing notable about using pytest rules in this

aignas (approved):

PR #3946, opened 2026-07-22
tests: add workspace-compatible unified pypi hub so @pypi refs work

rickeylev:
Currently, the unified pypi hub is bzlmod only. While this is WAI,
it makes using the references in tests that would be good to run
under workspace unable to run.

To fix, have our workspace config create a unified hub repo with
the underlying pypi hubs that it needs to map to.

This is kept as an internal-only helper because it doesn't
generalize well to arbitrary projects.

### DIFF
--- a/tests/tools/private/release/BUILD.bazel
+++ b/tests/tools/private/release/BUILD.bazel
@@ -1,5 +1,5 @@
 load("//python:py_library.bzl", "py_library")
-load("//tests/support:support.bzl", "SUPPORTS_BZLMOD")
+load("//tests/support:support.bzl", "NOT_WINDOWS")
 load("//tests/support/pytest_test:pytest_test.bzl", "pytest_test")
 
 py_library(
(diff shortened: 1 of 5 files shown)


tests/tools/private/release/BUILD.bazel
```diff
@@ -1,5 +1,5 @@
 load("//python:py_library.bzl", "py_library")
-load("//tests/support:support.bzl", "SUPPORTS_BZLMOD")
+load("//tests/support:support.bzl", "NOT_WINDOWS")
```
aignas, line 2: Why is windows not supported?

aignas (approved):

tests/tools/private/release/BUILD.bazel
rickeylev: Because we only run the release tools on linux (with mac added since mac laptops are common for dev work). Adding windows pulls in additional complications for dev deps etc.

PR #3948, opened 2026-07-24
refactor(pypi): make whl_library a macro and split impl into whl_archive and pip_archive

aignas:
Before this PR the `whl_library` would be a do-all repository rule. Whilst it
is convenient to reuse the code, it is actually really difficult to maintain
and make it more performant. Side effect here is that the python dependencies 
(like `setuptools`, etc) will no longer be downloaded for whl-only extracts, 
it makes it a tiny bit faster.

With this split we can drop certain dependencies from the whl extraction and
optimize the common path - whl extraction where the URL for downloading the
wheel is known. This also allows us to start handling the sdists in an entirely
different way.

In a followup PR I plan to split the part which just extracts the wheel to
lay a more surgical foundation to #3856.

Foundation work for #2410.
Split out of #3856.
Work towards #2948.


rickeylev:
argh, what the heck, it went totally awol and borked this. I'll reset and fix it.

rickeylev (approved):

python/private/pypi/whl_library.bzl
```diff
@@ -619,10 +652,75 @@ Instantiated from pip_repository and inherits config options from there.
...
 """,
-    implementation = _whl_library_impl,
+    implementation = _pip_archive_impl,
     environ = [
         "RULES_PYTHON_PIP_ISOLATED",
         REPO_DEBUG_ENV_VAR,
     ],
 )
+
+whl_archive = repository_rule(
```
nate-england, line 668: This change broke specifying `timeout` in `pip.parse`. Users get the following error now
```
(00:01:57) ERROR: /home/nathanael/.cache/bazel/_bazel_nathanael/f057a3c2a08c6c9f1c59ae3fa9073646/external/rules_python+/python/private/pypi/extension.bzl:615:20: @@rules_python+//python/extensions:rules_python++pip+pypi_312_absl_py_py3_none_any_88476fd8: no such attribute 'timeout' in 'whl_archive' rule
(00:01:57) ERROR: /home/nathanael/.cache/bazel/_bazel_nathanael/f057a3c2a08c6c9f1c59ae3fa9073646/external/rules_python+/python/private/pypi/whl_library.bzl:815:20: Traceback (most recent call last):
	File "/home/nathanael/.cache/bazel/_bazel_nathanael/f057a3c2a08c6c9f1c59ae3fa9073646/external/rules_python+/python/private/pypi/extension.bzl", line 615, column 20, in _pip_impl
		whl_library(name = name, **args)
	File "/home/nathanael/.cache/bazel/_bazel_nathanael/f057a3c2a08c6c9f1c59ae3fa9073646/external/rules_python+/python/private/pypi/whl_library.bzl", line 815, column 20, in whl_library
		whl_archive(name = name, **kwargs)
Error in repository_rule: failed to instantiate 'whl_archive' from this module extension
```

nate-england: Reported in https://github.com/bazel-contrib/rules_python/issues/4077

PR #3951, opened 2026-07-27
sphinxdocs: create sphinxdocs-specific pyproject and requirements

rickeylev:
To further separate sphinxdocs from rules_python, give it its own
pyproject and requirements files.

Also switch it to using uv.lock instead of requirements.lock


rickeylev:
as an aside, while doing this, i found that uv_lock can't be set on its own. I sent https://github.com/bazel-contrib/rules_python/pull/3952 to try and fix that

PR #3957, opened 2026-07-27
build: move requirements and lock files out of docs/ to dev/

rickeylev:
build(deps): move root requirements and lock files out of docs/ to dev/

Currently, the root development requirements and lock files reside in the `docs/` directory, which is confusing as they contain development and testing dependencies for the entire repository, not just documentation.

To fix this, create a `dev/` directory and move `pyproject.toml`, `requirements.txt`, `uv.lock`, and their associated lock build rules into it. Update repository references and regenerate the lockfiles accordingly.


rickeylev:
fwiw, I'm not strongly tied to the `dev/` name. At the least, though, we should move these out of docs/, since they aren't just for the docs stuff.

aignas (approved):
Another place to put them would be `private`, but `dev` or the root of the repo is good.

PR #3958, opened 2026-07-27
docs: merge external sphinxdocs dependencies into lockfile generation

rickeylev:
Currently, to express the starting points of the PyPI dependency graph for documentation lockfiles, all dependencies must be manually listed in `docs/pyproject.toml`. Because `sphinxdocs` is an external dependency not published to PyPI, its dependencies previously had to be manually duplicated in `docs/pyproject.toml` to be included in dependency resolution.

This change automates synthesizing a unified `pyproject.toml` during lockfile generation by merging additional `pyproject.toml` dependency lists into the main file before passing the result to the lock targets.

### DIFF
--- a/docs/BUILD.bazel
+++ b/docs/BUILD.bazel
@@ -209,10 +209,29 @@ sphinx_build_binary(
     ],
 )
 
+py_binary(
+    name = "merge_pyproject_deps",
+    srcs = ["merge_pyproject_deps.py"],
+    python_version = "3.14",
+    target_compatible_with = _TARGET_COMPATIBLE_WITH,
+)
+
+genrule(
+    name = "merge_pyproject",
+    srcs = [
+        "pyproject.toml",
+        "@sphinxdocs//:pyproject.toml",
+    ],
+    outs = ["merged/pyproject.toml"],
+    cmd = "$(location :merge_pyproject_deps) --output $@ $(location pyproject.toml) $(location @sphinxdocs//:pyproject.toml)",
+    target_compatible_with = _TARGET_COMPATIBLE_WITH,
+    tools = [":merge_pyproject_deps"],
+)
+
 # Run bazel run //docs:requirements.update
 lock(
     name = "requirements",
-    srcs = ["pyproject.toml"],
+    srcs = [":merge_pyproject"],
     out = "requirements.txt",
     args = [
         "--emit-index-url",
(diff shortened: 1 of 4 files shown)


docs/requirements.txt
```diff
@@ -5,11 +5,11 @@
 absl-py==2.3.1 ; python_full_version < '3.10' \
     --hash=sha256:a97820526f7fbfd2ec1bce83f3f25e3a14840dac0d8e02a0b71cd75db3f77fc9 \
     --hash=sha256:eeecf07f0c2a93ace0772c92e596ace6d3d3996c042b2128459aaae2a76de11d
-    # via rules-python-docs (docs/pyproject.toml)
+    # via rules-python-docs (bazel-out/k8-fastbuild-ST-501889d92340/bin/docs/merged/pyproject.toml)
```
aignas, line 8: This is no-go, IMHO. This is not strictly reproducible.

docs/BUILD.bazel
```diff
@@ -209,10 +209,29 @@ sphinx_build_binary(
...
+    name = "merge_pyproject",
+    srcs = [
+        "pyproject.toml",
+        "@sphinxdocs//:pyproject.toml",
+    ],
+    outs = ["merged/pyproject.toml"],
+    cmd = "$(location :merge_pyproject_deps) --output $@ $(location pyproject.toml) $(location @sphinxdocs//:pyproject.toml)",
+    target_compatible_with = _TARGET_COMPATIBLE_WITH,
+    tools = [":merge_pyproject_deps"],
+)
```
aignas, line 229: Ideally the `pyproject.toml` uses `uv` workspaces to create a `lock` file. Then we include the `pyproject.toml` via data and it just works.

I think it would be good to figure out how to use https://docs.astral.sh/uv/concepts/projects/dependencies/ for this.

We even have a test that checks that `pyproject.toml` inclusion is working: https://github.com/bazel-contrib/rules_python/tree/main/tests/uv/lock/workspaces

The very least we can do is use `copy_file` to bring the `pyproject.toml` into the current module scope and then include it using workspaces as in the example. The best we could do is to just include it.

aignas (changes requested):

docs/requirements.txt
rickeylev: Ah, yeah. "not strictly reproducible" is a good way of putting it. Having the merge tool rewrite these would solve that.

docs/BUILD.bazel
rickeylev: Ohh, interesting. Yeah, workspaces sounds very promising.

Maybe we should add a workspaces arg to the lock() rule? Or something like that. I haven't finished reading the docs, but -- yeah copy_file or something -- I suspect some starlark logic to arrange files so the paths in pyproject are compatible with the bazel invoked uv.

aignas: Yeah, +1 to adding an arg.

PR #3962, opened 2026-07-27
feat(rules): emit `RunfilesGroupInfo` from `py_binary`, `py_test`, and `py_runtime`

### LINKED ISSUE #3757: Emit RunfilesGroupInfo in py_binary targets
# 🚀 feature request

### Relevant Rules

I want `py_binary` to return an additional provider, [`RunfilesGroupInfo`](https://github.com/hermeticbuild/rules_runfiles_group).
`py_library` could contribute by also returning this provider, but that's more of an implementation detail.

I'd be open to work on this if you agree that this is a good idea.

### Description

I'd like to make packaging of `py_binary` targets in container images more efficient. To that end, I have designed a set of providers and described their semantics in [a new Bazel module](https://github.com/hermeticbuild/rules_runfiles_group).

`py_library` targets should forward `RunfilesGroupInfo` and `RunfilesGroupMetadataInfo` from their deps and add their own runfiles (`srcs`, `data`) to a group.
`py_binary` targets should aggregate runfiles groups from their deps and emit `RunfilesGroupInfo` and `RunfilesGroupMetadataInfo` that contain at least:

* A group for the interpreter + standard library
* One group per third party dependency

Additionally, it could also make sense to produce groups for each `py_library` target in the transitive dependency graph.

Doing nothing would mean we continue creating (very large) container image layers that contain the full runfiles tree of a `py_binary` target.

Writing custom packaging rules just for rules_python is unsustainable in my opinion.

FrankPortman:
Implements #3757 against rules_runfiles_group 0.1.0 (the new entries-depset API).

When enabled, binaries emit these groups:

* `rules_python#runtime`: interpreter + stdlib
* `rules_python#pypi/<name>`: one per PyPI package, stable across whl repo names
* `rules_python#<label>`: one `Label` named group per `py_library`
* `rules_python#venv`: the binary-specific venv symlinks and support files
* `rules_python#app`: the binary's own code, executable, bootstrap, and build data (`executable_group = True`)
* `data#<label>`: rule targets from data attributes

- Only py_binary/py_test/py_runtime emit the public provider, since only they can satisfy the union == default_runfiles contract (a py_library's srcs travel via PyInfo, not its own runfiles). 
  - There is also a flag I didn't fully understand here (`--srcs_as_runfiles`) that I feel like is somehow implicated. Libraries propagate entries through a private provider. 
  - The consequence is a foreign rule that takes in `py_library` may not get RGI behavior for free without its own shim. We can change this if we are willing to accept a change in semantics around `union == default_runfiles` (I think).
- Gated by the upstream `--@rules_runfiles_group//runfiles_group:enabled flag`; `--//python/config_settings:runfiles_groups` defaults to following it and can override either way.
- A binary overriding `pyc_collection` away from the config default, or using legacy implicit `__init__.py`, falls back to coarse runtime/venv/app grouping (dep entries are computed against the config, so the union would otherwise break).
  - Claude helped me find these edge cases - they were not in my initial understanding or sketch.

I had Claude patch the Bazel 9 vs Bazel 8 autoloads stuff just to get CI working, since I figure y'all will have a preference for how that is handled, and I can just adapt accordingly.

Fixes #3757

---

To implement this PR, I first looked at the reference implementation and browsed the `rules_python` enough to form an opinion on what a sketch might look like. I then steered Claude Fable pretty heavily in plan-mode, given a sketch/pseudocode of how I think it should look. Finally, I used Claude Fable to put finishing touches on comments, documentation, help me think of new test cases, and then stress test some edge cases.

I believe I understand this code well enough to respond to feedback and "own" the changes, conceptually, but I want to be transparent about LLM assistance.


news/3757.added.md
```diff
@@ -0,0 +1,10 @@
+(rules) Added experimental support for the `RunfilesGroupInfo` provider from
+[`rules_runfiles_group`](https://github.com/bazel-contrib/rules_runfiles_group).
+When enabled (via `--@rules_runfiles_group//runfiles_group:enabled` or
+{obj}`--@rules_python//python/config_settings:runfiles_groups`), `py_binary`
```
aignas, line 4: I think we should advertise only one - the `//python/config_settings` one.

aignas (commented):
In general I quite like the solution, thank you for submitting the PR. I think @rickeylev is doing some experiments with the venv building so I'll defer to him for final approval. The only thing is that I would love to keep the `@rules_runfiles_group//runfiles_group:enabled` an internal target and force the users to go through the main flag in `rules_python`. THis makes it easier to reason about the API surface.

For my understanding:
* If this is disabled, the overhead would be minimal right? It's basically just an extra `if` check.
* This is disabled by default?
* Where does the `rules_runfiles_group` dependency comes from?

One last thing, if we could link to upstream docs on how one can create this layering rule for container image, that would be a big +1 from me.

FrankPortman:
Thanks for the review!

Re the flag, I deliberately built an interaction with the upstream flag but this definitely up for discussion. The upstream `:enabled target` is from `rules_runfiles_group`'s ecosystem-wide switch (the `rules_java` integration in bazelbuild/rules_java#368 reads the same one). The Python flag's `auto` default follows it so a polyglot repo can turn on layering everywhere with one flag when a packaging rule needs it. Otherwise explicit `enabled`/`disabled` work as you might expect:

| `--//python/config_settings:runfiles_groups` | upstream `:enabled=false` (default) | upstream `:enabled=true` |
|---|---|---|
| `auto` (default) | no | **yes** |
| `enabled` | **yes** | **yes** |
| `disabled` | no | no |

My thought was that it would be nice to not have to opt individual rulesets in, if we're comfortable doing a top-down `:enabled=true`, but since this feature is new and experimental I could see it being reasonable to ONLY respect the `rules_python` flags. LMK which way y'all prefer, just wanted to clarify current behavior since it's buried in a long PR.

Your questions:
- Overhead when disabled: yes, minimal - two implicit label attrs (the flag targets) plus a couple of if checks per target. No providers, depsets, or runfiles objects are created. Bazel 7/8 might be even more minimal but I don't fully understand the autoloads stuff there.
- Disabled by default: **yes**. auto resolves to the upstream flag's default, which is off.
- The dep comes from the BCR: `bazel_dep` on `rules_runfiles_group` 0.1.0 (the module that @malt3 linked in #3757).

 I'll add pointers to the `rules_runfiles_group` README ("for users" / "for packaging rule authors") and `rules_img` to the flag docs once that code is settled.

aignas:
I'd love @malt3 have a look as well, does this look idiomatic usage of the `rules_runfiles_group`?

malt3:
Hey there. I'd like to give this a proper review. This week I'm on the go without a laptop. I'll be able to review early next week.

FrankPortman (commented):
_Posted by Fable 5.1 on behalf of @FrankPortman requesting a Claude review._

Adversarial review of the current head (`c37ba6dd`), findings inline. Summary and the parts that don't fit on a line:

**Verdict.** Not mergeable as-is: the public provider's core contract (union of groups == `default_runfiles`) is violated for any rule target in `data` whose default outputs aren't in its `data_runfiles` (verified empirically on Bazel 9.1.1; details on the `collect_data_entries` comment), and the test suite cannot detect it. The design choices themselves check out: the private `PyRunfilesGroupsInfo` seam, using `data_runfiles` for `data`, both coarse fallbacks (verified to hold via cquery), and the version-gated shim.

**Structural feedback / simplifications.**
1. The groups are built by re-enumerating every runfiles contribution in parallel with the code that builds `DefaultInfo` runfiles, so the invariant holds by bookkeeping rather than by construction. The robust shape is to build the runtime/venv/app/dep pieces once in `_get_base_runfiles_for_binary` / `_create_executable` and derive both `DefaultInfo` and the groups from the same objects. This is the one item worth a maintainer opinion before acting, since it touches core runfiles construction.
2. Fixing the blocker is a simplification: `data` handling collapses to `files` merged with `data_runfiles` for every target, and the singleton heuristic disappears.
3. Deduplicate the pyc-selection rule with `PycCollectionAttr` and extract one shared "PyInfo files a binary adds" helper used by both the binary runfiles loop and the groups code.
4. Drop the duplicated rank constants, inline the single-use provider wrapper, skip empty entries, and declare the flag attrs only when the feature is available.

**Performance.** Disabled mode: two extra implicit label attrs per py target plus a couple of branches, and one avoidable runfiles allocation per binary (`venv_files_runfiles`). Enabled mode: per-target cost is O(direct deps) with one entry and one depset node retained (linear, not quadratic, thanks to the 0.1.0 depset shape); no new actions. The place it can bite at scale is consumer-side: per-library granularity plus the empty entries mean a binary with N transitive libraries yields roughly N+5 groups, and a packager enforcing a layer budget pays upstream's pairwise merge, roughly O(G^2) in group count. Skipping empty entries helps now; a coarser (e.g. by-repo) granularity option is a reasonable follow-up.

**Test gaps** (listed inline on the test file): both coarse fallbacks, py_test, py_runtime, pyi_deps, precompile-enabled selection, py_binary-in-deps, foreign `RunfilesGroupInfo` deps, non-PyInfo deps, rule targets in `srcs`, `--build_python_zip`.

CI is currently red on the ruff format check (one file).

python/private/runfiles_groups.bzl
```diff
@@ -0,0 +1,517 @@
...
+          to the collecting target's own group.
+    """
+    direct = []
+    transitive = []
+    own_files = []
+    for target in data:
+        if RunfilesGroupInfo in target:
+            transitive.append(target[RunfilesGroupInfo].entries)
+            continue
+        runfiles = target[DefaultInfo].data_runfiles
```
FrankPortman, line 454: _Posted by Fable 5.1 on behalf of @FrankPortman requesting a Claude review._

**blocker / correctness.** Union of groups != `default_runfiles` for the common `data` shape. With Bazel's default `--incompatible_always_include_files_in_data`, `ctx.runfiles(collect_default = True)` adds a `data` rule target's **default outputs** in addition to its `data_runfiles`; this only adds `data_runfiles` (plus the singleton special case below), and the docstring asserts default outputs "contribute nothing". Verified on Bazel 9.1.1 with a rule returning `DefaultInfo(files = depset([a, b]))` and a rule with `files != runfiles` in `data`: their files land in `default_runfiles` and in no group, via py_binary `data`, py_library `data`, and inherited through py_binary-in-deps. Upstream's `runfiles_groups.data_entry()` gets this right (`files` merged with runfiles).

Suggested fix: for every target, `content = ctx.runfiles(transitive_files = files).merge(data_runfiles)`; the singleton heuristic then becomes unnecessary. The pre-existing comment in `common.bzl` (~L301-306) that this mirrors is stale and should be fixed too.

tests/runfiles_groups/runfiles_groups_tests.bzl
```diff
@@ -0,0 +1,393 @@
...
+            for path in group.symlinks
+        ]),
+    ).contains_exactly(
+        _unique_sorted([
+            s.path
+            for s in default_runfiles.symlinks.to_list()
+        ]),
+    )
+
+def _test_py_binary_groups(name):
```
FrankPortman, line 130: _Posted by Fable 5.1 on behalf of @FrankPortman requesting a Claude review._

**major / tests.** The only union-equality test uses `genrule` and `filegroup` in `data`, both of which put their outputs in `data_runfiles`, so it cannot detect the blocker above. Add a Starlark rule with a multi-file `DefaultInfo(files = ...)` and no runfiles, and one with `files != runfiles`, in both py_binary `data` and py_library `data`, asserting union equality.

python/private/py_executable.bzl
```diff
@@ -1322,8 +1327,119 @@ def py_executable_base_impl(ctx, *, semantics, is_test, inherited_environment =
...
+            extra_deps = extra_deps,
+            required_py_files = required_py_files,
+            required_pyc_files = required_pyc_files,
+            implicit_pyc_files = implicit_pyc_files,
+            implicit_pyc_source_files = implicit_pyc_source_files,
+        ))
+
     return providers
 
+def _create_runfiles_groups(
```
FrankPortman, line 1348: _Posted by Fable 5.1 on behalf of @FrankPortman requesting a Claude review._

**major / structure.** Groups are produced by re-enumerating every runfiles contribution (required files, pyc variant, `extra_default_outputs`, stage2, build data, cc/native runfiles, venv pieces) independently of the code that builds the `DefaultInfo` runfiles (`_get_base_runfiles_for_binary`, `_create_executable`). The invariant holds today only by parallel bookkeeping; any future addition to `app_runfiles` / `extra_runfiles` silently breaks it with no test coverage.

Suggested direction: build the runtime/venv/app/dep pieces once in `_get_base_runfiles_for_binary` / `_create_executable`, return them in `runfiles_details` / `exec_result`, and derive both the `DefaultInfo` runfiles and the groups from those same objects.

tests/runfiles_groups/runfiles_groups_tests.bzl
```diff
@@ -0,0 +1,393 @@
...
+    "runfiles_groups",
+)
+load("@rules_testing//lib:analysis_test.bzl", "analysis_test")
+load("@rules_testing//lib:test_suite.bzl", "test_suite")
+load("@rules_testing//lib:util.bzl", rt_util = "util")
+load("//python:py_binary.bzl", "py_binary")
+load("//python:py_library.bzl", "py_library")
+load("//python/private:runfiles_groups.bzl", "PyRunfilesGroupsInfo")  # buildifier: disable=bzl-visibility
+
+_tests = []
```
FrankPortman, line 34: _Posted by Fable 5.1 on behalf of @FrankPortman requesting a Claude review._

**major / tests.** Untested paths: both coarse fallbacks (`legacy_create_init = 1`, `pyc_collection` override), `py_test`, the `py_runtime` provider, `pyi_deps`, precompile enabled (implicit pyc vs source selection), py_binary-in-deps, a foreign `RunfilesGroupInfo` dep, a non-PyInfo dep in py_library (`include_plain_dep_py_files`), rule targets in `srcs`, `--build_python_zip`. The coarse paths were verified to hold via cquery, but nothing in the suite does. Each should get an analysis test asserting `_assert_groups_cover_default_runfiles`.

python/private/py_library.bzl
```diff
@@ -205,8 +200,67 @@ def py_library_impl(ctx):
...
+    own_files.add(required_pyc_files)
+    if pyc_collection_enabled:
+        own_files.add(implicit_pyc_files)
+    else:
+        own_files.add(implicit_pyc_source_files)
+    own_files.add(data_entries.own_files)
+
+    pypi_package, _ = _get_package_and_version(ctx)
+    own_entries = struct(
+        direct = [library_entry(ctx, pypi_package, own_files.build())],
```
FrankPortman, line 253: _Posted by Fable 5.1 on behalf of @FrankPortman requesting a Claude review._

**minor / correctness + performance.** Every py_library emits an own entry even when it contributes nothing, and the non-PyInfo fallback emits entries with empty runfiles. Verified: every py_binary carries `@@rules_python+//python/private:empty rank=-1 n=0` (the debugger alias resolving to a py_library); umbrella libraries and `cc_library` deps do the same. That's noise for packagers (empty layers unless the consumer filters) and pads the group count that feeds `layer_budget` merging. Skip own/fallback entries whose content is empty, as `collect_src_entries` already does.

python/private/runfiles_groups.bzl
```diff
@@ -0,0 +1,517 @@
...
+# pressure, packagers prefer merging groups that share an affinity, keeping
+# Python groups together instead of interleaving them with other rulesets'
+# groups.
+_AFFINITY = "rules_python"
+
+# Ranks: runtime < pypi deps < venv < first-party libraries < app.
+# Literal values of the runfiles_groups.RANK_* anchors: they can't be read at
+# load time because the shim stubs the module out on Bazel versions where the
+# feature is unavailable.
+_RUNTIME_RANK = -1000  # runfiles_groups.RANK_FOUNDATION
```
FrankPortman, line 94: _Posted by Fable 5.1 on behalf of @FrankPortman requesting a Claude review._

**minor / terseness.** `_RUNTIME_RANK` / `_PYPI_RANK` / `_APP_RANK` duplicate upstream `RANK_FOUNDATION` / `RANK_SHARED_DEPS` / `RANK_EXECUTABLE`. The "can't read them at load time" justification doesn't hold: they're only needed inside the `*_entry()` constructors, which never run when the shim is stubbed. Use `runfiles_groups.RANK_*` inside the constructors and keep only `_VENV_RANK` / `_LIBRARY_RANK` as local constants.

python/private/runfiles_groups.bzl
```diff
@@ -0,0 +1,517 @@
...
+    value = ctx.attr._runfiles_groups_flag[BuildSettingInfo].value
+    if value == RunfilesGroupsFlag.ENABLED:
+        return True
+    if value == RunfilesGroupsFlag.DISABLED:
+        return False
+
+    # auto: follow the ecosystem-wide switch.
+    return runfiles_groups.is_enabled(ctx)
+
+def pyc_collection_enabled_by_default(ctx):
```
FrankPortman, line 139: _Posted by Fable 5.1 on behalf of @FrankPortman requesting a Claude review._

**minor / structure.** `pyc_collection_enabled_by_default` re-implements `PycCollectionAttr.is_pyc_collection_enabled` (attributes.bzl) with the attr fixed to `inherit`, and `_py_info_files` re-implements the dep loop in `_get_base_runfiles_for_binary`. Two copies of the pyc-selection rule to drift apart. Suggested: give `is_pyc_collection_enabled` a `pyc_collection =` parameter, and extract one `py_info_runfiles_files(info, pyc_collection_enabled)` helper used by both the binary runfiles loop and this file.

python/private/py_executable.bzl
```diff
@@ -519,6 +519,11 @@ WARNING: Target: {}
         # depset[ExplicitSymlink]None; symlinks that should be created in
         # the venv to augment app_runfiles
         venv_app_symlinks = venv.lib_symlinks if venv else None,
+        # runfiles|None; the non-interpreter venv files: site-packages
+        # symlinks and small generated files (pth, site init, pyvenv.cfg).
+        venv_files_runfiles = (
```
FrankPortman, line 524: _Posted by Fable 5.1 on behalf of @FrankPortman requesting a Claude review._

**minor / performance.** `venv_files_runfiles` allocates an extra runfiles object (and nested set) for every binary on every build, even with the feature disabled. Return `venv.files_without_interpreter` / `venv.lib_runfiles` here and build the runfiles lazily inside `_create_runfiles_groups`.

python/private/py_library.bzl
```diff
@@ -116,31 +128,15 @@ This allows optimizing the generation of symlinks to be cheaper at analysis time
         "_add_srcs_to_runfiles_flag": lambda: attrb.Label(
             default = labels.ADD_SRCS_TO_RUNFILES,
         ),
+        "_runfiles_group_enabled": lambda: attrb.Label(
```
FrankPortman, line 131: _Posted by Fable 5.1 on behalf of @FrankPortman requesting a Claude review._

**minor / idiom + performance.** On Bazel 7/8 both `_runfiles_group_enabled` (resolving to `//python:none`) and `_runfiles_groups_flag` are still declared (here, in py_executable.bzl and py_runtime_rule.bzl), adding two dependency edges to every py target even though `is_enabled` short-circuits on `RUNFILES_GROUPS_AVAILABLE` before reading either. Define the attrs dict in `runfiles_groups.bzl` as `{}` when unavailable and `dicts.add` it, so unsupported Bazel versions pay nothing.

news/3757.added.md
```diff
@@ -0,0 +1,10 @@
+(rules) Added experimental support for the `RunfilesGroupInfo` provider from
+[`rules_runfiles_group`](https://github.com/bazel-contrib/rules_runfiles_group).
+When enabled (via `--@rules_runfiles_group//runfiles_group:enabled` or
```
FrankPortman, line 3: _Posted by Fable 5.1 on behalf of @FrankPortman requesting a Claude review._

**minor / docs.** Advertises `--@rules_runfiles_group//runfiles_group:enabled` as an enable path; @aignas's open comment asks to advertise only the `//python/config_settings` flag. Mention only `--@rules_python//python/config_settings:runfiles_groups=enabled` here and keep the upstream flag solely in the `auto` value description of the reference docs.

docs/api/rules_python/python/config_settings/index.md
```diff
@@ -242,6 +242,55 @@ The `auto` value
...
+* One group per other `py_library` dependency, named by its label.
+* `rules_python#venv`: the binary-specific venv symlinks and support files.
+* `rules_python#app`: the binary's own code, executable, and bootstrap files.
+* One group per rule target in `data` attributes, named by its label.
+
+`py_library` participates through a private rules_python provider rather
+than by emitting `RunfilesGroupInfo` itself; its sources reach a binary via
+{obj}`PyInfo`, not through its own runfiles, so only a binary can produce
+groups whose union equals its `DefaultInfo.default_runfiles` (the public
+provider's contract). That exactness has two exceptions, in which only the
```
FrankPortman, line 269: _Posted by Fable 5.1 on behalf of @FrankPortman requesting a Claude review._

**minor / docs.** "That exactness has two exceptions" is inaccurate: the union contract still holds in coarse mode (verified); what degrades is granularity. Reword to "falls back to coarse grouping". Relatedly, the `collect_data_entries` docstring's "Known approximations" list describes a hypothetical gap while omitting the real one (default outputs of `data` rule targets).

tests/integration/whl_library/test_contents.py
```diff
@@ -120,13 +120,17 @@ def _normalize_label(label: str) -> str:
...
             "WHL_DEPS": "@whl_archive//:pkg",
-            "WHL_TARGET_DEPS": "@whl_deps_repo//:pkg",
+            "WHL_TARGET_DEPS": "@whl_deps_library//:pkg",
         }.items():
             self.assertEqual(
                 {
                     self._normalize_label(x)
                     for x in self._read_file(var)
                     if not x.endswith("toolchain_type")
+                    # The runfiles_groups shim points this attribute at
```
FrankPortman, line 130: _Posted by Fable 5.1 on behalf of @FrankPortman requesting a Claude review._

**minor / CI.** `ruff format --check` fails on this file ("Would reformat: tests/integration/whl_library/test_contents.py"), so the ruff check is red. Run `ruff format` on it.

python/private/internal_config_repo.bzl
```diff
@@ -111,14 +137,22 @@ def _internal_config_repo_impl(rctx):
         supports_whl_extraction = str(supports_whl_extraction),
         extract_needs_chmod = str(extract_needs_chmod),
         builtin_py_cc_link_params_provider = builtin_py_cc_link_params_provider,
-        modules_using_explicit_initpy = str(explicit_init_py_modules),
         bazel_8_or_later = str(bazel_major_version >= 8),
         bazel_9_or_later = str(bazel_major_version >= 9),
         bazel_10_or_later = str(bazel_major_version > 9),
     ))
 
+    if bazel_major_version >= 9:
```
FrankPortman, line 145: _Posted by Fable 5.1 on behalf of @FrankPortman requesting a Claude review._

**minor / hermeticity (acceptable as-is, flagging for the record).** The shim is keyed on `bazel_major_version >= 9` rather than on autoload availability: a Bazel 8 build with `--incompatible_autoload_externally=` (autoloads off) could load rules_runfiles_group but is stubbed anyway. Fine if documented as "Bazel 9+"; a version-independent key would need Bazel to expose the autoload setting to repository rules.

python/private/runfiles_groups.bzl
```diff
@@ -0,0 +1,517 @@
...
+    return runfiles_groups.entry(
+        name = RUNTIME_GROUP,
+        content = runfiles,
+        kind = "foundation",
+        rank = _RUNTIME_RANK,
+        do_not_merge = True,
+        merge_affinity = _AFFINITY,
+    )
+
+def venv_entry(runfiles):
```
FrankPortman, line 176: _Posted by Fable 5.1 on behalf of @FrankPortman requesting a Claude review._

**nit.** The docstring says the venv group goes "late in the ordering", but rank -10 sorts *before* the per-library groups at rank -1, contradicting the module docstring's stated order (runtime < pypi < venv < libraries < app). Either fix the wording or swap the two ranks.

python/private/runfiles_groups.bzl
```diff
@@ -0,0 +1,517 @@
...
+        depset of entries with the order the protocol requires.
+    """
+    direct = []
+    transitive = []
+    for c in collected:
+        direct.extend(c.direct)
+        transitive.extend(c.transitive)
+    return runfiles_groups.entries(direct = direct, transitive = transitive)
+
+def create_runfiles_group_info(entries, *, executable_group = None):
```
FrankPortman, line 486: _Posted by Fable 5.1 on behalf of @FrankPortman requesting a Claude review._

**nit / terseness.** `create_runfiles_group_info` is a one-line wrapper around the provider constructor with a single caller; inline it.

python/private/runfiles_groups.bzl
```diff
@@ -0,0 +1,517 @@
...
+        default_info = dep[DefaultInfo]
+        if py_files != None:
+            content = ctx.runfiles(transitive_files = py_files).merge(
+                default_info.default_runfiles,
+            )
+        elif include_plain_dep_py_files:
+            # Mirrors create_py_info: `.py` files of non-PyInfo deps are
+            # added to the library's transitive_sources. create_py_info
+            # already flattens this same depset in the same analysis, so
+            # the to_list() here is a cache hit, not a second flattening.
```
FrankPortman, line 339: _Posted by Fable 5.1 on behalf of @FrankPortman requesting a Claude review._

**nit.** The comment claims this `to_list()` is "a cache hit, not a second flattening"; depset flattening isn't memoized per call in a way that makes that reliably true (unverified against Bazel source). Drop the claim and just note the cost mirrors `create_py_info` for the same rare non-PyInfo dep.

malt3:
Hi @FrankPortman. Thank you very much for this PR.
I think we should hold off with merging this until the requirements for adding support for `rules_runfiles_group` in rules_java are finalised.
We are currently discussing whether the current design is good enough, or if we change the implementation to use a generic aspect that uses callbacks offered by the various language rules. You can follow the discussion in https://github.com/bazelbuild/rules_java/pull/368.

FrankPortman:
@malt3 understood. Will wait to see how that shakes out. I glanced at the proposal and it does look like it has the potential to be pretty concise and lead to simpler adoption from different rulesets.

PR #3971, opened 2026-07-28
refactor: updates to the py_extension rule for building C extensions

rsartor-cmd:
Per feedback from #3875 , this PR makes a few changes:
* Removed an unnecessary hasattr check on py_cc_toolchain
* Renamed the arbitrary "os" and "cpu" identifiers to terms standardized in PEP 508.

Also marks `py_extension` as experimental.


rickeylev (approved):
This is great, thank you!

PR #3973, opened 2026-07-29
feat(cc): introduce py_extension rule for C/C++ Python extension modules

### LINKED ISSUE #3283: Implement py_extension rule
This is a tracking issue to design and implement a rule, `py_extension` whose purpose is to create Python C extension objects.

A Python C extension itself is fairly simple: a shared library with a `PyInit_<name>` symbol. What makes it complicated are the myriad of different C++ linkages the extension may be used with.

I've written [rules_python py_extension design](https://docs.google.com/document/d/1aXZ6ejeGSRhfGFEC6AuGlIWW4gJGbOUsgYv16qqGeFc/edit?tab=t.0) to go into detail about the cases and design I'm currently thinking.

In short, there are 6 types of linkages, which come from two ways a terminal binary links its external dependencies and three ways a py_extension links its external dependencies, and how overlapping external dependencies are handled when those interact. (what constitutes "external" depends on the situation, but is usually code it doesn't own)

The most relevant use case is:
* A prebuilt Python binary (i.e. RPATH is unknown)
* A dynamically linked extension (whose RPATH we can control)
* In a virtual environment site-packages layout (i.e. the relative location of libraries is assumed)

rickeylev:
Building C/C++ Python extension modules currently requires manually
configuring complex `cc_binary` or `cc_shared_library` targets and
setting up platform, ABI3, and extension tags. The new
`py_extension` rule provides a high-level Starlark API for compiling
native extensions that integrate seamlessly with `rules_python`
toolchains and `py_library` dependencies.

At a high level, the implementation is fairly simple:
Users can provide raw C/C++ `srcs`, which are put into a `cc_library`.
That `cc_library` is then given to `cc_shared_library`, which performs
the linking to create e.g. `libfoo.so`. That artifact is then given
to a custom rule that handles rename it according to Python needs,
e.g. `foo.abi3.so`.

`cc_shared_library` is used because it provides various facilities
to control creation of a shared library, namely, control over what
is and isn't linked into the shared library.

Credit to @rsartor-cmd for finishing the prototype, expanding test
coverage, and verifying behavior.

Closes #3283.


rickeylev:
cc @rsartor-cmd : something that you might run into integrating on the protobuf side: https://github.com/bazel-contrib/rules_python/pull/3973/changes/65886db2d430da2b4a622ab3831d4794658aabf7  tldr, some flags necessary for building on mac are missing (that commit adds them)

rickeylev:
OK! Finally ready.

python/private/py_cc_toolchain_rule.bzl
```diff
@@ -45,7 +82,36 @@ def _py_cc_toolchain_impl(ctx):
...
+    if not abi_tag:
+        # Derive default ABI tag:
+        # On POSIX: cpython-XX[t] (PEP 3149 / PEP 703)
+        # On Windows: cpXX[t] (PEP 3149 / PEP 703, CPython issue & commit):
+        #   - https://peps.python.org/pep-3149/
+        #   - https://peps.python.org/pep-0703/
+        #   - https://github.com/python/cpython/issues/67169
+        #   - https://github.com/python/cpython/commit/03a144bb6ac3d7631a3bdb895e2a1f2d021fb08b
+        version_parts = ctx.attr.python_version.split(".")
+        prefix = "cp" if ctx.attr.sys_platform == "win32" else "cpython-"
```
aignas, line 101: Should we have the `implementation` as well? `cp` and `cpython` seem to be well known names.

python/private/py_cc_toolchain_rule.bzl
```diff
@@ -45,7 +82,36 @@ def _py_cc_toolchain_impl(ctx):
...
+        abi_tag = "{}{}{}{}".format(prefix, version_parts[0], version_parts[1], abi_flags)
+
+    libc = ctx.attr.libc or LibcFlag.get_value(ctx)
+
+    platform_tag = _get_platform_tag(
```
aignas, line 106: Just a reminder that we have the `EnvMarkerInfo` provider, that might be useful.

tests/cc/py_extension/py_limited_api/py_limited_api_tests.bzl
```diff
@@ -0,0 +1,140 @@
+# Copyright 2025 The Bazel Authors. All rights reserved.
```
aignas, line 1: I would vote to remove these headers everywhere, because they are unmaintainable.

aignas (approved):

tests/cc/py_extension/py_limited_api/py_limited_api_tests.bzl
rickeylev: Yeah, that's why we don't add them ourselves. This one just slipped through the original review.

However, since this code was contributed by Google/Bazel, they can decide to put their copyright on it. We generally shouldn't modify a copyright from a contributor once we accept their contribution (it just doesn't mean anything; us changing the code doesn't change their actual copyright claim on it).

IANAL, of course, but that's the basic gist of what's been explained to me.

python/private/py_cc_toolchain_rule.bzl
rickeylev: Ahhh is _that_ where these strings come from? Yeah, we should add an implementation field; I'll put together a PR for that.

python/private/py_cc_toolchain_rule.bzl
rickeylev: Oh yeah. Maybe that should go into a toolchain? Something I was pondering is a "python runtime metadata" provider that has the fields in common between PyRuntime and PyccTooolchain.

PR #3974, opened 2026-07-29
feat(pypi): support non-sha256 --hash pins and index digests

### LINKED ISSUE #3972: pip.parse: support non-sha256 --hash pins (AWS CodeArtifact serves sha512)
pip.parse only recognizes `--hash=sha256:` pins in requirements files — [`index_sources.bzl`](https://github.com/bazel-contrib/rules_python/blob/main/python/private/pypi/index_sources.bzl) splits requirement lines on that literal string, so pins using any other algorithm are silently dropped.

This matters because AWS CodeArtifact's Simple API advertises **sha512** digests for hosted wheels, and lockers like uv record the strongest digest offered. A `uv export` lockfile for a CodeArtifact index therefore pins those wheels as:

```
my-package==1.2.3 \
    --hash=sha512:1266ed31df77bd2c9939b224ade65b25db25b56ccd2e9c94d4008a656dea1519d176895b9746eef3586d30e444977354440305eb900ac84e2394b1acecfe9935
```

With `experimental_index_url`, pip.parse treats the entry as hash-less and falls back to trusting whatever the index lists:

```
DEBUG: /home/runner/.cache/bazel/_bazel_runner/.../external/rules_python+/python/private/repo_utils.bzl:101:16:
rules_python:my_module:pypi:hub:my_pip WARNING: requirement file has been generated without hashes, will use all hashes for the given version 1.2.3 that could find on the index:
    ["b0fd6065e31278409509c593276c4fe66948ad6a16800c7dec93f9b7d92ac267"]
```

The lockfile's pin is ignored, so the artifact is no longer verified against what was locked.

**Feature request:** parse `--hash=<algo>:<digest>` for the other PEP 503/691 algorithms (at least sha512) and match them against the index metadata, which already carries multi-algorithm digests in the PEP 691 `hashes` dict.

Observed on rules_python 2.0.3; the parsing is unchanged on `main`.

**Current workaround:** post-process the exported lockfile, replacing each sha512 pin with the wheel's sha256 computed from the downloaded bytes.

abcdabcd987:
Previously only `--hash=sha256:` pins and `#sha256=` URL fragments were
recognized. Pins using any other hash algorithm were silently dropped: the
requirement was treated as hash-less, and in the pip fallback the hashes
were stripped from the requirement line entirely, so the artifact was no
longer verified against what was locked.

Now a distribution digest is carried through the code as one canonical
`<algo>:<hex digest>` string (the same shape as pip `--hash` and uv.lock
`hash` values) for any `hashlib` algorithm:

- a new `hash.bzl` module (single `hash` symbol) centralizes digest
  parsing, validation and conversion to SRI;
- `parse_simpleapi_html` accepts any PEP 503 `#<algo>=<digest>` fragment
  and keys `whls`/`sdists` by the canonical digest; `sha256s_by_version`
  is renamed to `hashes_by_version`;
- matching pins against the index is a direct dict lookup; when the index
  only advertises a different algorithm the requirement falls back to pip
  with the pins kept in the requirement line, so nothing installs
  unverified (pip verifies sha256/384/512 pins and rejects requirement
  lines pinned only with other algorithms);
- `whl_library` gains an `integrity` attribute, and the extension now
  always passes the digest as an SRI `integrity` value instead of
  `sha256`, so the lock entries use one format for all algorithms; a
  malformed (non-hex) sha256/384/512 digest fails loudly at extension
  evaluation instead of silently disabling download verification;
- the lock file facts store `<algo>:<digest>` values (`_FACT_VERSION`
  bumped to `v2`; old facts are refetched once); this also fixes the
  latent bug where `dist_filenames` was written keyed by URL but read
  back keyed by digest, so stored filenames were never found;
- uv.lock `hash` values pass through for any algorithm instead of being
  mangled by a `sha256:` string replace.

Repo names are unchanged for sha256 digests, and sha256 stays the
preferred algorithm when a direct-URL requirement has several pins.
Digests whose algorithm has no SRI form (e.g. md5) still match the index
but are downloaded without downloader-side verification, with a warning.

Fixes #3972


python/private/pypi/hashes.bzl
```diff
@@ -0,0 +1,142 @@
...
+        use an SRI supported algorithm.
+    """
+    for algo in _SRI_ALGOS:
+        sri = hex_to_sri(algo, hashes.get(algo, ""))
+        if sri:
+            return sri
+
+    return ""
+
+def preferred_digest(hashes):
```
aignas: For utility libs like this, it would be good to name the file `hash.bzl` and have a single exported symbol `hash`, where it has 3 attributes that are functions: `hash.preferred_digest`, `hash.integrity`, `hash.hex_to_sri`.

python/private/pypi/hub_builder.bzl
```diff
@@ -677,6 +678,12 @@ def _whl_repo(
 
     args["urls"] = [src.url]
     args["sha256"] = src.sha256
+    if not src.sha256:
```
aignas: Please refactor to actually use `integrity` internally all of the time. This will make the code easier to maintain.

python/private/pypi/hub_builder.bzl
```diff
@@ -685,7 +692,7 @@ def _whl_repo(
     target_platforms = src.target_platforms if is_multiple_versions else []
 
     return struct(
-        repo_name = whl_repo_name(src.filename, src.sha256, *target_platforms),
+        repo_name = whl_repo_name(src.filename, src.sha256 or preferred_digest(src.hashes), *target_platforms),
```
aignas: here as well, lets just use `preferred_digest` all of the time.

python/private/pypi/parse_requirements.bzl
```diff
@@ -82,7 +83,7 @@ def parse_requirements(
          * `index_url`: {type}`str` The index URL used to download the package.
          * `srcs`: {type}`list[struct]` A list of per-distribution source entries, each
            containing: `distribution`, `extra_pip_args`, `requirement_line`,
-           `target_platforms`, `filename`, `sha256`, `url`, `yanked`.
+           `target_platforms`, `filename`, `sha256`, `hashes`, `url`, `yanked`.
```
aignas: let's use `integrity` as the name instead of `hashes`. In bazel-land it is more common.

python/private/pypi/parse_simpleapi_html.bzl
```diff
@@ -101,15 +108,26 @@ def parse_simpleapi_html(*, content, parse_index = False):
...
+        dist_url, _, fragment = href.partition("#")
+        algo, sep, digest = fragment.partition("=")
+        algo = algo.lower()
+        if not (sep and digest and algo in HASH_ALGOS):
+            if fragment:
+                # Not a hash fragment, keep it as part of the URL.
+                dist_url = "{}#{}".format(dist_url, fragment)
+            algo = ""
+            digest = ""
+        sha256 = digest if algo == "sha256" else ""
```
aignas: I think we should stop setting `sha256` internally in here and just use `digest` everywhere.

python/private/pypi/parse_simpleapi_html.bzl
aignas: Please change the name of attr to `hashes_by_version`

python/private/pypi/pypi_cache.bzl
```diff
@@ -251,16 +257,17 @@ def _get_from_facts(facts, known_facts, index_url, requested_versions, facts_ver
         else:
             dists = known_sources.setdefault("sdists", {})
 
-        known_sources.setdefault("sha256s_by_version", {}).setdefault(version, []).append(sha256)
+        known_sources.setdefault("sha256s_by_version", {}).setdefault(version, []).append(digest)
```
aignas: Also we should change sha256s_by_version here.

python/private/pypi/pypi_cache.bzl
```diff
@@ -318,7 +325,9 @@ def _store_facts(facts, fact_version, index_url, value):
     #   "dist_hashes": {
     #     "<index_url>": {
     #       "<last segment>": {
-    #         "<dist url>": "<sha256>",
+    #         # A bare sha256 hex digest or "<algo>:<digest>" when the index
```
aignas: lets do <algo>:<digest> all the time to simplify the code.

aignas (commented):
Thanks a lot for this cleanup/refactor/feature addition.

Overall I am in favour of this, but to make it more maintainable, please switch to using integrity all the time and make the MODULE.bazel.lock file entries more normalized, i.e. use the same format all the time, not just when it is not sha256. If it makes sense, we could use the SRI integrity value to write to the facts if that makes the code any easier to reason about.

Also please change the PR description to be more in style of a commit message. This is too large.

python/private/pypi/hashes.bzl
abcdabcd987: Done. Renamed to `hash.bzl` with a single exported `hash` struct following the
`version.bzl` style: `hash.preferred_digest`, `hash.integrity`,
`hash.hex_to_sri`, plus `hash.digest(algo, hex)` — a validating constructor
for the canonical `<algo>:<hex>` form that everything now uses internally
(this is where the algorithm allow-list check lives) — and `hash.ALGOS`,
which the docstrings cross-reference.

python/private/pypi/hub_builder.bzl
abcdabcd987: Done. `_whl_repo` no longer sets `sha256` at all; it always sets
`integrity = hash.integrity(src.digest)`, so the `whl_library` entries use one
format for every algorithm. `hash.integrity()` does `fail()` loudly when a
sha256/384/512 digest is not valid hex, so a corrupted pin cannot silently
disable verification.

python/private/pypi/hub_builder.bzl
abcdabcd987: Done, and it ended up simpler than that: src entries now carry a single
normalized `digest` string, so this line is just
`whl_repo_name(src.filename, src.digest, ...)`. `whl_repo_name` strips the
`<algo>:` prefix internally, so repo names stay byte-for-byte identical for
sha256 digests. `hash.preferred_digest` is still used at the one place where
several pins can exist for a single artifact (direct-URL requirements with
multiple `--hash` options), with sha256 preferred to keep names stable.

python/private/pypi/parse_requirements.bzl
abcdabcd987: I went with `digest` rather than `integrity` here and would like your take.
The src entries now carry one field holding the normalized
`<algo>:<hex digest>` value (the `sha256` + `hashes` pair is gone). I avoided
calling it `integrity` because in bazel-land that name specifically means the
SRI `<algo>-<base64>` format accepted by `ctx.download(integrity = ...)`, and
this field is deliberately not that: keeping the hex form makes index
matching, repo naming and lock-file diffs line up with pip's `--hash` and
uv.lock's `hash` values with zero conversions. A field named `integrity` that
cannot be passed to the `integrity` attribute seemed like a footgun. The SRI
conversion happens in exactly one place — `hash.integrity(src.digest)` when
building the `whl_library` args. Happy to rename if you still prefer
`integrity`.

python/private/pypi/parse_simpleapi_html.bzl
abcdabcd987: Done. The dist structs now only carry `digest` (`<algo>:<hex>`, or empty when
the index advertises none) — the `sha256` and `hashes` fields are gone. The
`whls`/`sdists` dicts are keyed by the same canonical value, so requirement
pins (already in that format) match with a direct dict lookup and the
prefix-stripping special case in `_add_dists` is deleted.

python/private/pypi/parse_simpleapi_html.bzl
abcdabcd987: Done — renamed here, in `pypi_cache.bzl` and in
`simpleapi_download._with_index_url`. The values are canonical
`<algo>:<hex>` strings (empty when the index advertises no digest, as
before), so the no-pins fallback path goes through exactly the same lookup as
pinned requirements.

python/private/pypi/pypi_cache.bzl
abcdabcd987: Done, together with the rename above — no occurrences of
`sha256s_by_version` remain.

python/private/pypi/pypi_cache.bzl
abcdabcd987: Done. `dist_hashes` values and `dist_yanked` keys are now always
`<algo>:<digest>` (or empty when the index advertises no digest, as before),
which deleted both the bare-sha256 legacy branch on read
and the stored-hash reconstruction loop on write. Since this changes the
stored format, I bumped `_FACT_VERSION` to `v2` — existing facts are treated
as a cache miss and refetched/rewritten once. I considered writing SRI values
into the facts as you suggested in the review summary, but `<algo>:<digest>`
keeps the hex (needed for repo naming and matching) and matches pip/uv lock
formats byte-for-byte, so it avoids base64 round-trips at every boundary.

abcdabcd987:
Thanks for the review! All addressed:

* Everything internal now flows through a single canonical
  `<algo>:<hex digest>` string per artifact (the pip/uv format); the sha256
  special-casing for distribution digests is gone from
  `parse_simpleapi_html`, `parse_requirements`, `pypi_cache` and
  `hub_builder` (the PEP 714 `metadata_sha256` handling is untouched).
* `whl_library` repos always get the digest as an SRI `integrity` value now —
  `sha256` is no longer set by the extension, so the lock entries use the same
  format for every algorithm.
* Facts always store `<algo>:<digest>` and the fact version is bumped to `v2`
  (one-time refetch of the cached index info).
* `hashes.bzl` → `hash.bzl` with a single exported `hash` struct.
* PR description rewritten as a commit-message-style summary.

One naming question left inline: I used `digest` for the `<algo>:<hex>` form
and kept `integrity` strictly for SRI values — see the
`parse_requirements.bzl` thread.

abcdabcd987:
CI is back to green now.

aignas (approved):
Thank you!

PR #3987, opened 2026-08-01
docs: convert raw URLs in generated docs and docstrings to hyperlinks

rickeylev:
Plain text URLs in documentation Markdown files and Starlark/Python docstrings were rendered as non-clickable text in the generated HTML output, making navigation cumbersome.

To fix this, raw URLs are converted into MyST autolinks (<https://...>), Markdown links, and {gh-issue} roles. A custom {pep} Sphinx role is also introduced in docs/conf.py to dynamically resolve PEP numbers to https://peps.python.org/pep-XXXX/, and existing PEP references in docstrings are updated to use {pep}.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,6 +1,9 @@
 # Configuration file for the Sphinx documentation builder.
 
 import os
+import re
+
+from docutils import nodes
 
 # -- Project information
 project = "rules_python"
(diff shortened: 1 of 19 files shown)


docs/conf.py
```diff
@@ -219,9 +222,19 @@
 ]
 
 
+def _pep_role(name, rawtext, text, lineno, inliner, options={}, content=[]):
+    match = re.search(r"\d+", text)
```
dougthor42, line 226: I'm not familiar with what Sphinx/MyST passes as `text` to this function - is it the full text of the markdown file? Or just what's inside the role backticks?

If the former, this pattern is overly broad, yeah?

dougthor42 (approved):

docs/conf.py
rickeylev: It's passed what's inside the backticks

PR #3988, opened 2026-08-01
refactor(tests): convert legacy py_extension in local_toolchains to official rule

rickeylev:
The local_toolchains integration test workspace previously relied on a
custom, partial implementation of py_extension that predated the
official rule.

Replace this legacy test implementation with the official
@rules_python//python/cc:py_extension.bzl rule, simplifying the target
definition to use sources directly.


rickeylev:
Oh windows...making this difficult again. Lets see how many commits it takes to figure it out this time

aignas (approved):

PR #3993, opened 2026-08-03
refactor(cc): simplify _current_py_cc_libs_impl

rickeylev:
Commit 0d6016ffe5307e2e7c30a90319f14f8efa801d2b accidentally added complex filtering logic for static libraries and DLLs to `_current_py_cc_libs_impl` in `python/private/current_py_cc_libs.bzl`. That extra filtering logic was actually a hack to support another hack in `py_extension`, and shouldn't have been added.

To simplify, update `_current_py_cc_libs_impl` to return `py_cc_toolchain.libs.providers_map.values()` directly.

Since Windows was relying on that hack, move it into a helper rule in the
py_extension implementation.

### DIFF
--- a/python/private/cc/py_extension_macro.bzl
+++ b/python/private/cc/py_extension_macro.bzl
@@ -151,6 +151,7 @@ def py_extension(
     # from CcInfo in py_cc_toolchain because system_provided=True in cc_import leaves DefaultInfo empty.
     py_extension_libs(
         name = win_libs_name,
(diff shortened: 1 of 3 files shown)


rickeylev:
Ok, this is ready now. I think there's a more proper fix that needs to go into the toolchain, but i couldn't figure out how windows .lib files are supposed to make their way into downstream consumers. So this works for now.

aignas (approved):

Issue #2945, opened 2025-05-29
Remove automatic __init__.py generation (set incompatible_default_to_explicit_init_py to true)


rickeylev:
The automatic generation of `__init__.py` files has been a source of odd, surprising, and incorrect behavior. e.g. it makes directories importable that shouldn't be importable. This is because it's logic is extremely simple: create an init.py file in every parent directory for a path that ends with .py or .so.

It's also expensive: it requires flattening the runfiles at analysis time.

Some of the venv site package work incidentally side-steps this because the way it creates symlinks "hides" some of the paths from the regular set of runfiles.

Addressing this is a bit more complicated because the incompatible_default_to_explicit_init_py lives in bazel, but the usage is in rules_python.

In any case, I think the todo list is:

1. [ ] Add a rules_python flag to replace incompatible_default_to_explicit_init_py
2. [ ] Disable the behavior by default
3. [ ] Remove the dead code paths

Users can use `--flag_alias` to map (1) to the bazel builtin flag.

This is the rules_python side of https://github.com/bazelbuild/bazel/issues/7386

brandonchinn178:
This also causes issues if you have a non-empty `__init__.py` file (e.g. at the root of your project for users to import everything from), as Bazel will just clear the file entirely

asymingt:
Removing the automatic `__init__.py` would potentially also allow us to leverage namespaced package (PEP420) support in Python 3.3+ to stitch together a single module from parts generated by an aspect run over a collection of targets. This would be very useful for generating python bindings for IDLs, like proto files and ROS messages.

limdor:
What is the status of this task? Is there any timeline? Right now this is the main blocker to be able to flip the default.
It would be great if by default users would not get automaticly created init py files by default, that is very counterintuitive in Python that you get init files where you did not ask for them.

andre-dialpad:
I have run into an issue where my __init__.py was emptied and made my tests to fail for seemingly no reason only to find out this diabolical behavior is in place. 

Are there any news on this ?

rickeylev:
> diabolical behavior

love this description  of it. Whenever we put a deprecation warning in for this, I'm totally putting this in the text of it.

nate-england:
What's the recommended usage when maintaining py_binary and py_library targets on a file basis instead of package? Since there's typically only one source per target, I'm not understanding what the value of me making init files to go along with each one of these. We use `# gazelle:python_generation_mode file` to get the target-per-file mapping.

[...]

PR #3997, opened 2026-08-03
feat(bzlmod): make __init__.py generation configurable module-wide

armandomontanez:
In #3841, a warning pushing users to migrate away from implicit `__init__.py` generation was added. While it's good to flag this bad behavior, silencing it requires users to either explicitly configure this option on every `py_binary` and `py_test` target, or configure the option globally in their `.bazelrc`.

To better facilitate a migration, this change introduces a mechanism for modules to configure this option module-wide. This has multiple benefits:

1. Everyone working in the module doesn't need to remember to explicitly set `legacy_create_init` on every target.
2. Everyone that depends on the module receives the correct behavior as configured by the module.
3. It becomes possible to tell BCR-wide which modules have adopted this migration.

Work towards #2945


rickeylev (approved):
Thanks for this! Overall this was pretty good. I cleaned it up a bit and renamed the api to explicit_init_py(default=...).

armandomontanez:
Thanks for ushering this in!

Issue #3977, opened 2026-07-30
sphinxdocs: relative markdown links between pages fail to resolve (myst.xref_missing)


ctcjab:
> [!IMPORTANT]
> **Corrected below** — see [this comment](https://github.com/bazel-contrib/rules_python/issues/3977#issuecomment-5137805864).
> Two cells of the matrix in this description are wrong: they were action-cache hits that never re-ran Sphinx.
> Stock sphinxdocs fails in **all** execution modes, not only sandboxed, and copying sources instead of
> symlinking them *does* fix the default and `local` cases. The description below is kept as filed.

### What happened

With `sphinx_docs` + `myst_parser`, a relative markdown link between two pages
(e.g. `[Add an entity](add-entity.md)`) resolves correctly when the Sphinx action
runs unsandboxed, but fails with `myst.xref_missing` when the action runs under
`linux-sandbox`. With `-W` (which we use as a docs-freshness gate) this is a hard
build failure.

```
_sources/how-to/register-dataset.md:8: WARNING: Unknown source document
  '/home/.../execroot/_main/bazel-out/k8-fastbuild/bin/.../_ontology_docs/_sources/how-to/add-entity'
  [myst.xref_missing]
```

Note the two paths involved. The document being processed is reported at the
**sandbox** execroot:

```
/home/.../sandbox/linux-sandbox/58/execroot/_main/bazel-out/.../_sources/how-to/register-dataset.md
```

but the resolved link target is at the **real** execroot:

```
/home/.../execroot/_main/bazel-out/.../_sources/how-to/add-entity
```

Inside the sandbox, `bazel-out` is a symlink into the real output tree. MyST
appears to resolve the relative target to a canonical (symlink-resolved) absolute
path, which drops the sandbox prefix, while Sphinx's docnames remain relative to
the `srcdir` it was given (the sandbox path). The two can then never match.

### Reproduction matrix

Same target, same `conf.py`, same sources, only the execution strategy varies:

| strategy | result |
| --- | --- |
| worker (default) | pass |
| `--strategy=SphinxBuildDocs=local` | pass |
| `--strategy=SphinxBuildDocs=sandboxed` | **fail** (`myst.xref_missing`) |

Because worker mode is the default, this is easy to miss — and it means a docs
build is only correct as long as it is not sandboxed.

### What does and doesn't break

- `:doc:` roles and `toctree` entries resolve fine in all three modes.
- Only **relative markdown links** between pages fail.

[...]

PR #4006, opened 2026-08-05
fix(sphinxdocs): fix path resolution for conf.py in separate directory

### LINKED ISSUE #3999: sphinx_docs cannot use a generated conf.py when its output path conflicts with a sibling executable target
`sphinx_docs` currently derives the Sphinx source directory from the path of its `config` label after materialising it into its private source tree.

This becomes problematic when the config is generated and the documentation source directory has the same name as another Bazel target.

Minimal example:

```starlark
_generated_conf(
    name = "generated_conf",
    output_path = "docs/conf.py",
)

py_binary(
    name = "docs",
    # ...
)

sphinx_docs(
    name = "needs_json",
    config = ":generated_conf",
    # ...
)
```

Bazel rejects this because the generated output `docs/conf.py` is below the executable output `docs`:

```text
One of the output paths '.../docs/conf.py' and '.../docs'
is a prefix of the other.
```

Using a generated config at another output path avoids the Bazel conflict, but then `sphinx_docs` materialises the config at that path in its private source tree and uses its parent as the Sphinx source directory. Consequently, the actual documentation sources are no longer under the selected source root.

Sphinx itself supports separating the source directory and configuration directory, but this separation is currently not expressible through `sphinx_docs`.

Expected behavior: a generated config should be usable without requiring its Bazel output path to be located below the documentation source directory.

rickeylev:
Sphinx documentation builds failed when conf.py was generated or located
in a separate directory because _relocate used the relative short_path of
config, setting sphinx_source_dir_path to the config subdirectory rather
than the source tree root.

Relocate ctx.file.config using paths.basename(ctx.file.config.path) so it
is always placed at the root of the generated _sources directory.

Work towards #3977.
Fixes #3999

### DIFF
--- a/sphinxdocs/sphinxdocs/private/sphinx.bzl
+++ b/sphinxdocs/sphinxdocs/private/sphinx.bzl
@@ -100,7 +100,7 @@ def sphinx_docs(
         sphinx,
         config,
         formats,
-        strip_prefix = "",
+        strip_prefix = None,
         extra_opts = [],
         tools = [],
         allow_persistent_workers = True,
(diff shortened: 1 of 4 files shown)


sphinxdocs/sphinxdocs/private/sphinx.bzl
```diff
@@ -378,7 +381,9 @@ def _sphinx_source_tree_impl(ctx):
     # Though Sphinx has a -c flag, we move the config file into the sources
     # directory to make the config more intuitive because some configuration
     # options are relative to the config location, not the sources directory.
-    source_conf_file = _relocate(ctx.file.config)
```
aignas, line 381: I am wondering if the relocation trick should be also done for the lock rule so that the workspaces can be more easily included.

aignas (approved):

sphinxdocs/sphinxdocs/private/sphinx.bzl
rickeylev: Yeah, I was trying to shepherd slopbot to do that, but it was being stubborn about it. One of my goals with that was so that both regular uv and bazel-invoked uv work with the same pyproject.toml file, i.e. no paths in pyproject.toml that only exist in the bazel runfiles/exec environment

aignas: Agree with the goal, maybe we could have something like:
```starlark
    workspaces = {
        "foo/pyproject.toml": "@foo//:pyproject.toml",
        "bar/pyproject.toml": "//bar:pyproject.toml",
    },
```

That way we can do the remapping of the files into the local environment.

---

Another thing that we may want to do is publish `bazel-sphinxdocs` as a wheel, so that the dependencies, etc, set correctly and users can easily install it.

PR #4008, opened 2026-08-06
refactor(toolchain): rename abi_tag to soabi to eliminate PEP ambiguity

rickeylev:
Rename `abi_tag` to `soabi` in `PyCcToolchainInfo` and associated
toolchain implementation and macro parameters.

I found the `abi_tag` term a bit ambiguous
with PEP 425 wheel ABI tags and PEP 3149 C extension filename SOABI tags.
So rename it to `soabi` to directly reflect the SOABI sysconfig var
that it is traditionally derived from.

This also uncovered a bug where the value was _just_ the Python ABI-ness
(`cpython-311`), while it should be the full ABI infix that includes
the platform (`cpython-311-x86_64-linux-gnu`). This also helped simplify
some of the Windows code paths.


rickeylev:
Ready, PTAL

aignas (approved):
Nice, this is easier to understand, thank you!

PR #4020, opened 2026-08-08
build(pyrefly): enable Pyrefly static type checking across Python targets

rickeylev:
Enables Pyrefly static type checking across Python source and test
targets in rules_python and sphinxdocs.

Previously, static type analysis was not uniformly applied across
all Python targets, leaving potential type inconsistencies and
unbound variable edge cases uncaught during builds and CI runs.

This change configures the global Pyrefly aspect to evaluate Python
targets by default in Bzlmod mode, adds type annotations and
narrowing assertions on runfiles resolution, and annotates dynamic
imports (such as generated protobuf stubs and compiled C extensions)
with explicit type ignore comments. Targets utilizing dynamically
generated bootstrap wrappers retain opt-out tags with documented
rationales.


rickeylev:
Apparently slopbot was lying about the tests it ran. I'll switch this back to non-draft once its sorted

.agents/rules/python.md
```diff
@@ -13,3 +13,13 @@
 * **External Objects**: When defining a `TypedDict` for an external object,
   link to its definition in the docstring.
 
+## Type Checking & Annotations
+* **Target skipping vs in-file disables**: Prefer disabling specific errors in
+  source files (e.g. `# type: ignore[...]` / `# pyrefly: ignore[...]`) over
```
dougthor42: FYI pyrefly treats `# type: ignore[...]` as a blanket `# type: ignore`. I recommend against using it.

.agents/rules/python.md
```diff
@@ -13,3 +13,13 @@
...
+  disabling type checking on targets (e.g. `tags = ["no-pyrefly"]`).
+* **Type assertions**: When adding assertions for type narrowing, add an
+  end-of-line comment: `assert foo is not None  # type assert`.
```
dougthor42, line 22: `assert` is removed when python is run with [`-O`](https://docs.python.org/3/using/cmdline.html#cmdoption-O). I doubt this ever applies, but it can potentially cause issues.

Perhaps recommend `if foo is not None: raise AssertionError(...)`?

Non-blocking.

(Such asserts in test files are obviously just fine.)

examples/wheel/main.py
```diff
@@ -12,9 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import examples.wheel.lib.module_with_data as module_with_data
-import examples.wheel.lib.module_with_type_annotations as module_with_type_annotations
-import examples.wheel.lib.simple_module as simple_module
+import examples.wheel.lib.module_with_data as module_with_data  # type: ignore[import-not-found]
```
dougthor42: Use [`# pyrefly: ignore[missing-import]`](https://pyrefly.org/en/docs/error-kinds/#missing-import) instead because `# type: ignore[...]` is treated as a blanket ignore.

Further similar comments suppressed.

rickeylev:
Ready, PTAL

python/private/pypi/whl_installer/arguments.py
```diff
@@ -57,7 +57,7 @@ def parser(**kwargs: Any) -> argparse.ArgumentParser:
     return parser
 
 
-def deserialize_structured_args(args: Dict[str, str]) -> Dict:
+def deserialize_structured_args(args: dict[str, Any]) -> dict[str, Any]:
```
dougthor42, line 60: Just confirming that we've dropped support for py3.8 right? `dict` as a type hint was [added in 3.9](https://docs.python.org/3/whatsnew/3.9.html#type-hinting-generics-in-standard-collections).

python/runfiles/runfiles.py
```diff
@@ -728,6 +728,20 @@ def Create(env: Optional[Dict[str, str]] = None) -> Optional["Runfiles"]:
 
         return None
 
+    # TODO: Update return type to Self when 3.11 is the min version
+    # https://peps.python.org/pep-0673/
+    @staticmethod
+    def CreateOrRaise(env: Optional[Dict[str, str]] = None) -> "Runfiles":
```
dougthor42: New feature; should be added to news.

sphinxdocs/sphinxdocs/private/proto_to_markdown.py
```diff
@@ -16,9 +16,9 @@
 import itertools
 import pathlib
 import sys
-from typing import Callable, TextIO, TypeVar
+from typing import Callable, Iterator, Optional, Sequence, TextIO, TypeVar
```
dougthor42: Many of these are [deprecated since 3.9](https://docs.python.org/3/library/typing.html#typing.Iterator); use `collections.abc` instead.

sphinxdocs/sphinxdocs/private/sphinx_build.py
```diff
@@ -10,13 +12,55 @@
...
 import traceback
-import typing
+import types
+from typing import TextIO, TypedDict
+
+import sphinx.application  # type: ignore[import-not-found]
+from sphinx.cmd.build import main  # type: ignore[import-not-found]
+
+
+class WorkRequestInput(TypedDict, total=False):
```
dougthor42, line 22: Q: Why a TypedDict instead of a dataclass?

Same for the other instances.

sphinxdocs/sphinxdocs/private/sphinx_build.py
```diff
@@ -57,17 +101,17 @@ def __init__(
         self,
         srcdir: pathlib.Path,
         destdir: pathlib.Path,
-        max_workers: typing.Optional[int] = None,
+        max_workers: int | None = None,
```
dougthor42, line 104: Just confirming that we only support 3.10 and above. `|` operator was [added in 3.10](https://docs.python.org/3/library/typing.html#typing.Iterator).

This obviously supersedes the question about 3.8/3.9 in a previous comment.

sphinxdocs/sphinxdocs/src/sphinx_bzl/bzl.py
```diff
@@ -1645,7 +1673,7 @@ def resolve_any_xref(
         return matches
 
     @override
-    def resolve_xref(
+    def resolve_xref(  # pyrefly: ignore[bad-override]
```
dougthor42, line 1692: What's pyrefly complaining about here?

Adding a lot of `ignore[bad-override]` is code smell.

sphinxdocs/sphinxdocs/src/sphinx_bzl/bzl.py
```diff
@@ -1744,6 +1772,8 @@ def add_object(self, entry: _ObjectEntry, alt_names=None) -> None:
 
         if alt_names is not None:
             alt_names = list(alt_names)
+        else:
+            alt_names = []
```
dougthor42: Maybe just

```python
alt_names = list(alt_names) if alt_names else []
```

?

sphinxdocs/sphinxdocs/private/proto_to_markdown.py
```diff
@@ -16,9 +16,9 @@
...
 
-from stardoc.proto import stardoc_output_pb2
+from stardoc.proto import stardoc_output_pb2  # type: ignore
```
dougthor42: Prefer to not use blanket ignores.

Duplicate comments suppressed.

tests/cc/py_extension/py_extension_pkg_test.py
```diff
@@ -9,7 +9,7 @@ def test_import_via_package(self):
 
     def test_direct_import(self):
         with self.assertRaises(ModuleNotFoundError):
-            import ext_pkg_test  # buildifier: disable=g-import-not-at-top # noqa: F401
+            import ext_pkg_test  # type: ignore # buildifier: disable=g-import-not-at-top # noqa: F401
```
dougthor42: Is the `buildifier: disable still needed?

tests/pytest_test/BUILD.bazel
```diff
@@ -10,6 +10,9 @@ pytest_test(
         "@rules_python//python/config_settings:bootstrap_impl": "script",
         "@rules_python//python/config_settings:venvs_site_packages": "yes",
     },
+    # pytest_test uses a generated bootstrap runner (_boot.py) from template
+    # expansion which cannot be statically checked from source by Pyrefly.
```
dougthor42: This doesn't necessarily mean the _entire target_ can't be type checked, right? Is it possible to use a more narrow-focused ignore?

.agents/rules/python.md
rickeylev: That's fine. These are "type narrowing" asserts, so they're only relevant to static analysis.

* self.xx set by `setUp()`
* `def foo(x = None, y = None)`, where x and y are mutually exclusive
* Functions that could return None, but wouldn't in practice because of an implicit pre-condition.

python/private/pypi/whl_installer/arguments.py
rickeylev: Yes, we dropped 3.8

I have another PR out to raise the minimum from 3.9 to 3.10

python/runfiles/runfiles.py
rickeylev: Good catch. I'll move this to a separate PR.

sphinxdocs/sphinxdocs/private/sphinx_build.py
rickeylev: It's dealing with JSON data, so it's dicts being passed around. Using a dataclass would require manually deserializing those objects into more dataclasses, which isn't really necessary.

Actually, now that we have the unified hub, I think this could be switched over to using protobuf messages instead of using JSON. I'll defer that to another PR though.

sphinxdocs/sphinxdocs/private/sphinx_build.py
rickeylev: For sphinxdocs, yes. There's not a strong support policy with sphinxdocs anyway.

sphinxdocs/sphinxdocs/private/proto_to_markdown.py
rickeylev: Done

tests/cc/py_extension/py_extension_pkg_test.py
rickeylev: no idea why the buildifier comment was there. It wouldn't do anything. Removed.

.agents/rules/python.md
rickeylev: TIL that `ignore[...]` was a special form. Fixed to tell it to never do that

sphinxdocs/sphinxdocs/private/proto_to_markdown.py
rickeylev: Removed all the blanket suppressions

sphinxdocs/sphinxdocs/src/sphinx_bzl/bzl.py
rickeylev: Done

examples/wheel/main.py
rickeylev: Done,removed all the blanket suppressions

tests/pytest_test/BUILD.bazel
rickeylev: Switched to error-specific disables throughout
(I told slop bot to do that, and it did until it didn't anymore)

sphinxdocs/sphinxdocs/private/proto_to_markdown.py
```diff
@@ -16,9 +16,12 @@
 import itertools
 import pathlib
 import sys
-from typing import Callable, TextIO, TypeVar
+from collections.abc import Callable, Iterator, Sequence
+from typing import Optional, TextIO, TypeVar
```
dougthor42: Optional Nit: Now that I know sphinxdocs uses 3.10+, you can use `| None` instead of `Optional`.

sphinxdocs/sphinxdocs/private/sphinx_build.py
```diff
@@ -281,28 +325,31 @@ def run(self) -> None:
...
-    def _send_response(self, response: "WorkResponse") -> None:
+    def _send_response(self, response: WorkResponse) -> None:
         self._outstream.write(json.dumps(response) + "\n")
         self._outstream.flush()
 
-    def _prepare_sphinx(self, request):
-        sphinx_args = request["arguments"]
+    def _prepare_sphinx(self, request: WorkRequest):
+        sphinx_args = request.get("arguments", [])
         srcdir = pathlib.Path(sphinx_args[0])
```
dougthor42, line 340: This'll raise `IndexError` if the `request.get("arguments", [])` returns the default value.

Intentional that the error bubbles up? Or do you want to handle it?

Same later on with `bazel_outdir = sphinx_args[1]`.

sphinxdocs/sphinxdocs/src/sphinx_bzl/bzl.py
```diff
@@ -21,19 +23,26 @@
 from collections.abc import Collection
 from typing import Callable, Iterable, TypeVar
```
dougthor42: Not sure how much "general cleanup" you want to do in this PR, but this file could use a consistency refresh. `import typing` vs `from typing import ...` for one, the the migration to `collections.abc` as well.

Non-blocking.

python/private/pypi/whl_installer/arguments.py
```diff
@@ -14,7 +14,7 @@
 
 import argparse
 import json
-from typing import Any, Dict, Set
+from typing import Any, Set
```
dougthor42: Seems a bit silly to only do a partial cleanup. `s/Set/set/g` too.

sphinxdocs/sphinxdocs/src/sphinx_bzl/BUILD.bazel
```diff
@@ -4,7 +4,6 @@ package(
     default_visibility = ["//sphinxdocs:__subpackages__"],
 )
 
-# NOTE: This provides the library on its own, not its dependencies.
```
dougthor42, line 7: I think this comment is still accurate, yes?

sphinxdocs/sphinxdocs/private/sphinx_build.py
rickeylev: Changed to removed the .get() call. Slopbot loves adding defaults like that.

It's expected that the first arg is the srcdir (sphinx requires this), so we'll leave it as it was.

sphinxdocs/sphinxdocs/src/sphinx_bzl/BUILD.bazel
rickeylev: Restored.

sphinxdocs/sphinxdocs/src/sphinx_bzl/bzl.py
rickeylev: Done. I think.

python/private/pypi/whl_installer/arguments.py
rickeylev: Done throughout.

sphinxdocs/sphinxdocs/private/proto_to_markdown.py
rickeylev: Done throughout.

sphinxdocs/sphinxdocs/src/sphinx_bzl/bzl.py
rickeylev: Ok, dug into this. A few of them could be fixed. Most of them are occuring because pyrefly isn't able to resolve the imports for some reason (the missing-import suppressions). Fixed what could be, added comments to the suppressions about why they still need it.

rickeylev:
PTAL

dougthor42 (approved):

PR #4023, opened 2026-08-09
chore: update default Python version to 3.14

rickeylev:
Update the default Python toolchain version from 3.11 to 3.14 to align
with newer Python releases and provide modern defaults out of the box.

This updates the default toolchain version configurations in the module
extension, documentation, and test suites.


rickeylev (commented):
@aignas Can you double check the logic in python/private/pypi/whl_extract.bzl?

It said that the RECORD file had invalid paths: the original paths in the wheel were present, but its actually supposed to reflect the install path. In this case, we have e.g. https://github.com/bazel-contrib/rules_python/blob/main/tests/repos/whl_with_data1

Up above, L40-50 is where the remap of schemes occurs.

From what I can tell, I think its right. Starting in Python 3.14 it seems that metadata.files() is stricter.

That said, I'm not entirely confident in the rewritten values it gives. e.g. `../data/` ? I guess it's up-referencing up out of the site-packages dir and into the venv? I suppose that's correct.

rickeylev:
I created https://github.com/bazel-contrib/rules_python/pull/4025 to show as a repro for this metadata.files bug. I think this is pre-existing? It might just be that 3.14 got more strict about things.

aignas:
> @aignas Can you double check the logic in python/private/pypi/whl_extract.bzl?

Hmm, if it is install path, then indeed the RECORD files will be incorrect and it seems that we need to rewrite them and we are not. We are just leaving whatever was in the wheel to begin with.

aignas:
Looks like this is very close - only requirements locking tests are remaining.

rickeylev:
PTAL. Finally ready; this is pretty much just 3.XX -> 3.YY now.

aignas (approved):

Issue #3024, opened 2025-06-24
importlib.metadata.files() doesn't work because RECORD is missing


rickeylev:
It appears that `importlib.metadata.files()` is broken with the pypi integration because the RECORD file gets ommited.

Per docs at https://docs.python.org/3/library/importlib.metadata.html#distribution-files

> In the case where the metadata file listing files (RECORD or SOURCES.txt) is missing, files() will return

This came to my attention via @mering trying to use the rtree package, which tries to use its distribution metadata to find some C libraries to load.

Excluding RECORD seems somewhat intentional, though? The code has this comment:

https://github.com/bazel-contrib/rules_python/blob/main/python/private/pypi/whl_library_targets.bzl#L315-L318

> # RECORD is known to contain sha256 checksums of files which might include the checksums
> # of generated files produced when wheels are installed. The file is ignored to avoid
> # Bazel caching issues.

What's the caching issue, exactly? For a downloaded whl, what generated files is it referring to? If a whl is built at repo time from an sdist, then files shouldn't be changing, and the hash should be stable, so how does a cache issue come up?

groodt:
I agree. I never liked removal of it.

However, I think it's to support sdist. Wheel building outside specialized tools like bazel can be non-reproducible.

My solution to that problem is to drop support for sdist in lockfiles, but that doesn't make me very fun at parties. 😜

aignas:
To make it reproducible one could run:
* `repairwheel`
* repack the wheel with our script that attempts to make the RECORD reproducible.

groodt:
I guess we could also create an invalid RECORD tbh? Just listing of the files without hashes.

I don't understand how repacking the wheel would work tbh (unless we are saying the same thing) because if we attempt to create a valid RECORD file, the source of the non-determinism is the file hashes. And guaranteeing determinism of arbitrary files from an sdist is the problem. 

A thought that I've had is that maybe instead of hiding the problem, we should rather document the behaviour. So if somebody notices they see cache misses, they should correct the source of non-determinism. Either through creating wheels upfront and locking to those, or some other environmental changes to squeeze out non-determinism in the sdist builds.

rickeylev:
I still don't understand why building from sdist (in repo phase) is a problem.

If the sdist-built-during-repo-phase files are non-deterministic, how does ignoring the RECORD file change that? Those files are still inputs into the build. The RECORD is basically just recording what bazel already will (content hash).

[...]

PR #4025, opened 2026-08-09
fix(pypi): rewrite RECORD file entries for extracted .data contents

rickeylev:
Extracting wheels with .data/ subdirectories moves files to their
target directories and deletes .data/, but left .dist-info/RECORD
referencing deleted .data/ paths. This causes tools like
importlib.metadata.files() to fail locating or reading files.

Per PEP 427 and PEP 376, rewrite .dist-info/RECORD entries during
wheel extraction to match installed locations relative to
site-packages.

* Also adds Starlark unit tests for RECORD rewriting and expands
  importlib_metadata_test.

Work towards #3024


aignas:
A few open questions:
* Should we actually do this path manipulation where we extract it differently from the wheel layout? We are doing this because this was the layout before when we used to extract with `installer`.
* Should we rename the RECORD file to something else, like `RECORD.orig`? We can generate a RECORD file from just the globs in a build action and that would mean it is always deterministic.

rickeylev:
AFAIK, we're extracting the wheel the same as what other installers do, sans the later "install" part where it shows up into a binary's venv.

e.g. given https://github.com/bazel-contrib/rules_python/blob/main/tests/repos/whl_with_data1/whl_with_data1-1.0.dist-info/RECORD


* platlib/ and purelib/ goes into site-packages
* scripts goes into scripts; venv later puts it into $venv/bin/
* headers/ goes into headers; venv later puts it into $venv/headers/
* data/ goes into data; venv later puts it into $venv/ directly (e.g. `data/XXX/foo.h` will show up as `$venv/XXX/foo.h`

rickeylev:
Ready, PTAL

aignas (approved):
Just to mention related code:

When patching the wheels we are rewriting a RECORD file. Since we are doing this here as well, we should stop doing it. Just patching the files is enough. That brings a question if we should rezip the patched wheel itself in a build action?

python/private/pypi/BUILD.bazel
```diff
@@ -41,6 +41,15 @@ alias(
     visibility = ["//visibility:public"],
 )
 
+alias(
+    name = "wheel_record_rewriter",
+    actual = select({
+        "@platforms//os:windows": "wheel_record_rewriter.ps1",
+        "//conditions:default": "wheel_record_rewriter.sh",
```
aignas, line 48: In the repack_whl.py we have a `RECORD` writer that provides compliant platform agnostic solution. I'd be +1 on using that. However it has a higher overhead of a python toolchain attached to the process.

aignas: Ah, I take it back, it is not handling the Windows vs non-Windows differences this has uncovered.

rickeylev:
> should rezip the patched wheel itself in a build action?

I would lean towards no, for two reasons: (1) keep our API smaller, and (2) does doing so benefit our code at all?

I don't entirely remember why re-zipping was required...did it stem from using piptools or something? Something that required a .whl to process?

TBH I'm not a fan of the `:whl` targets that let you reference the .whl file. They double the amount of disk used and I'm not sure they provide sufficient value.

aignas:
> TBH I'm not a fan of the :whl targets that let you reference the .whl file. They double the amount of disk used and I'm not sure they provide sufficient value.

The only internal usage is gazelle_python.yaml generation, but we should be able to use the `extracted_whl_srcs` for this.

For external usage, there is definitely something, but it's hard to know. I think patching + whl reziping is such a niche use-case, that we could just not support forward compatibility in there.

PR #4026, opened 2026-08-09
fix: preserve arguments with spaces

stanbot8:
The runtime environment launcher collapsed interpreter arguments into one shell word. It now forwards each argument without reparsing it, preserving the original argument boundaries.

This includes a regression test for interpreter arguments containing spaces.

Related: https://github.com/bazelbuild/bazel/issues/30644


katre (approved):

rickeylev (approved):

aignas:
The CI failures are legit.

stanbot8:
Thanks, fixed.

PR #4029, opened 2026-08-09
feat(uv): add --directory support for locking

aignas:
Currently, requirements locking targets are tightly coupled to the Bazel
package directory where the lock rule is declared. This prevents having
reusable invocations across different example packages (e.g. `examples/bzlmod`
and `examples`) that produce identical lock files, hindering migration from
`compile_pip_requirements` to the `lock` rule.

To resolve this, add a `directory` attribute to the `lock` macro and rules to
pass `--directory` to `uv lock` and `uv pip compile`. Reroot input sources,
constraints, and project root arguments relative to the specified working
directory, and update UNIX and Windows execution templates to handle directory
remapping seamlessly.


aignas:
@rickeylev, if you wish, you could point your agent at this. It would be great to finish the Windows parts. This should make it possible to have the `lock` rule used for example lock regeneration and it should be invariant where we are running the lock. We could also build on top of this for the uv workspace files.

rickeylev:
_slopbot, roll out!_ 
slopbot

dev/BUILD.bazel
```diff
@@ -12,6 +12,7 @@ lock(
         "--universal",
         "--upgrade",
     ],
+    directory = "dev",
```
rickeylev: maybe directory should default to package_directory()? It feels redundant to have to specify the directory the rule is in.

rickeylev (commented):
OK, agent carried windows support through.

Not sure if I entirely agree/understand some of its changes. It made a bunch of changes to the add_all helper struct which i dont' fully grok.

dev/BUILD.bazel
aignas: Hmm, this would force all users to update the lock-files, we can flip it later. I think we can merge it as is for now.

aignas: Although this brings more compatibility with the legacy `compile_pip_requirements` so let's do this.

aignas:
I have also added lock file targets to the `//private:requirements.update`. This should cover almost all of the cases.

python/uv/private/lock.bzl
```diff
@@ -636,6 +696,10 @@ def lock(
         lock_target_kwargs["build_constraints"] = build_constraints
     if constraints:
         lock_target_kwargs["constraints"] = constraints
+    if directory == None:
+        directory = native.package_name()
```
aignas, line 703: FYI, I noticed that this can break user workflows if build constraints are included from somewhere else than subpackages of the current package due to the `_reroot` limitations.

We have 3 choices:
* not set the default.
* leave as is.
* improve the reroot to also work for non-prefix cases.

Any preferences?

aignas: Used `//python/private:common.bzl#relative_path`.

rickeylev (approved):

PR #4041, opened 2026-08-13
feat(sphinxdocs): make SphinxDocsLibraryInfo provider public

JoHoenk:
Custom rules that produce doc files for sphinx_docs currently must depend on the private sphinx_docs_library rule implementation, since SphinxDocsLibraryInfo lives under sphinxdocs/private. Expose it via //sphinxdocs:sphinx_docs_library_info.bzl so custom rules can supply docs without that dependency.

Add a custom_docs_library test rule and output test verifying a non-sphinx_docs_library rule can supply docs to sphinx_docs through the public provider.

### DIFF
--- a/sphinxdocs/sphinxdocs/private/sphinx_docs_library_info.bzl
+++ b/sphinxdocs/sphinxdocs/private/sphinx_docs_library_info.bzl
@@ -1,4 +1,8 @@
 """Provider for collecting doc files as libraries."""
+
+# NOTE: A provider is used for memory efficiency because providers perform key
+# sharing.
+# buildifier: disable=name-conventions
 SphinxDocsFileset = provider(
     doc = "A set of doc files sharing the same path manipulation.",
     fields = {
@@ -10,12 +14,14 @@ The documentation files. A tuple because depset elements must be immutable.
         "prefix": """
 :type: str
 
-Prefix to prepend to file paths in `files`. Added after `strip_prefix` is removed.
+Prefix to prepend to file paths in `files`. Added after `strip_prefix` is
+removed.
 """,
         "strip_prefix": """
 :type: str
 
-Prefix to remove from file paths in `files`. Removed before `prefix` is prepended.
+Prefix to remove from file paths in `files`. Removed before `prefix` is
+prepended.
 """,
     },
 )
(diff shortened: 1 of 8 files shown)


news/sphinx_docs_library_info.added.md
rickeylev, line 1: We don't do news files for sphinxdocs, so go ahead and remove this.

sphinxdocs/sphinxdocs/private/sphinx_docs_library_info.bzl
```diff
@@ -24,7 +24,12 @@ is prepended.
 :type: depset[struct]
```
rickeylev: Since we're making this public, lets more formally define the type be:

* Define a SphinxDocsFileset provider with fields (a provider is used for memory efficiency; it does dict key sharing)
* Change the annotation here to `depset[SphinxDocsFileset]`
* Update the plain struct(...) calls to SphinxDocsFileset

sphinxdocs/sphinxdocs/sphinx_docs_library_info.bzl
```diff
@@ -0,0 +1,50 @@
+# Copyright 2026 The Bazel Authors. All rights reserved.
```
rickeylev: I'm assuming you aren't a member of the Bazel organization who is contributing the code on behalf of them and asserting their copyright, so please remove the copyright text here

sphinxdocs/sphinxdocs/sphinx_docs_library_info.bzl
```diff
@@ -0,0 +1,50 @@
...
+        prefix = "my_docs/",
+        strip_prefix = ctx.label.package + "/",
+    )
+    return [SphinxDocsLibraryInfo(
+        files = entry.files,
+        prefix = entry.prefix,
+        strip_prefix = entry.strip_prefix,
+        transitive = depset(
+            direct = [entry],
+            transitive = [d[SphinxDocsLibraryInfo].transitive for d in ctx.attr.deps],
```
rickeylev: Lets also create a merge() function to insulate callers from having to hard-code the fields that get copied over.

rickeylev (changes requested):
Thanks for this. This mostly LGTM.

May I ask what sort of customized logic you are needed to implement? I'm just curious to see if there's something to consider adding to the sphinx_docs_library rule.

JoHoenk:
> Thanks for this. This mostly LGTM.
> 
> May I ask what sort of customized logic you are needed to implement? I'm just curious to see if there's something to consider adding to the sphinx_docs_library rule.

Thanks a lot for the quick review and your interest!

So reason behind it is that currently we are developing [rules_score](https://eclipse-score.github.io/tooling/latest/) which is basically a bazel framework for generating software in a safety reated context. This means that we are structuring our software into "dependable elements" which include both code and safety documentation. And there we want to build a sphinx documentation per dependable element. A dependable element itself however can also depend on another dependable element (with its own sphinx documenation).

So that we now can wire it together in sphinx we already load from the private path the SphinxDocsLibraryInfo and we would like to make this now a (stable) public api.

Also for our case one monolithic sphinx_docs (with its one big source tree) is really an issue for scaleability, as the complete Sphinx Build - including the deps - will be rebuild due to one tiny change. That´s why we implemented smaller scale sphinx invocations (per dependable element) which we then merge afterwards in a [sphinx module](https://github.com/eclipse-score/tooling/blob/main/bazel/rules/rules_score/private/sphinx_module.bzl#L651).

rickeylev:
> merging different sphinx invocations

Oh that's pretty cool. Yeah, something I don't like about sphinx is its incremental support isn't great. The sphinxdocs worker mode installs a custom extension to use a hash-based incremental mode, but i'm not sure how incremental it actually makes things under the sphinx hood.

I've been wanting to do something similar with our docs (our docs are actually rules_python, sphinxdocs, and rules_python_gazelle all as one site, and I'd rather they be independent sites that can crossref easily)

rickeylev (approved):

PR #4043, opened 2026-08-14
fix(pypi): do not return rctx.metadata for pip_archive

aignas:
There was a small regression in #3948 that may affect
experimental repository cache users.

Related to #3791


aignas:
Might be a good idea to cut 2.3.1 with this.

rickeylev (approved):

PR #4044, opened 2026-08-14
refactor: split code into multiple files

aignas:
A followup to #4043 to split code into multiple files to make the interaction
and reuse more intentional and explicit.

Work towards #2948.


rickeylev (approved):

rickeylev:
Ah, it would appear i broke the build with some type errors in the release tools. I'll send a PR to fix that tonight/tomorrow

aignas:
Thank you very much!

PR #4054, opened 2026-08-15
fix(pypi): support source-less wheels with dependencies

### LINKED ISSUE #4053: `pip.parse` fails for source-less wheels with dependencies
(I diagnosed this with heavy help from Opus.)

# 🐞 bug report

### Affected Rule

`pip.parse`.

### Is this a regression?

Yes, I believe `2.2.0` works, while [`2.3.1`](https://github.com/bazel-contrib/rules_python/issues/4049) fails.

### Description

I expected `py-spy==0.4.1` to work as a `py_library` dependency, but observed that `whl_library_targets` points the wrapper's `srcs` at an empty source target. Bazel then rejects the wrapper during analysis.

## 🔬 Minimal Reproduction

See [`examples/py-spy`](https://github.com/hartikainen/rules_python/tree/5ea758d017ddcac58b47bbe0a1bbf900ba2d97e8/examples/py-spy).

From that directory, run:

```console
bazel test //:py_spy_test
```

## 🔥 Exception or Error

```text
... //:srcs does not produce any py_library srcs files (expected .py or .py3)
```

## 🌍  Your Environment

**Operating System:**

```text
macOS 26.5.2 arm64
```

```text
Build label: 9.2.0
```

**Rules_python version:**

```text
2.3.1
```

**Anything else relevant?**

The workaround is to stay on `2.2.0`. I have a small fix and unit tests on the separate [`py-spy-fix`](https://github.com/hartikainen/rules_python/tree/eb2978b0e63aede343374d16d23f44adccc9f2e9) branch.

hartikainen:
Source-less wheels with dependency metadata fail analysis because the generated wrapper `py_library` puts an empty `:srcs` target in its `srcs` attribute.

This forwards `:srcs` through the wrapper only when it produces Python sources, while retaining it in `deps` for `PyInfo` propagation. Before this change, [`py-spy==0.4.1`](https://pypi.org/project/py-spy/0.4.1/) dependency fails during analysis. Afterward, it can be used as a dependency.

Adds regression coverage for source-less, sourceful, and generated namespace-package source targets, along with a news entry.

Largely implemented with Opus.

Fixes #4053.

### DIFF
--- a/python/private/pypi/whl_library_targets.bzl
+++ b/python/private/pypi/whl_library_targets.bzl
@@ -101,7 +101,7 @@ def whl_library_targets(
         **kwargs: Extra args passed to the {obj}`whl_library_deps_targets` and {obj}`whl_library_srcs`.
     """
     create_extra_targets = bool(requires_dist or group_name) and dep_template
-    whl_library_srcs(
+    wrapper_srcs = whl_library_srcs(
         name = name,
         sdist_filename = sdist_filename,
         data_exclude = data_exclude,
@@ -133,6 +133,7 @@ def whl_library_targets(
             dep_template = dep_template,  # only needed if requires_dist is present
             repo = None,  # set aliases in the same repo
             aliases = {},
+            srcs = wrapper_srcs,
             **kwargs
         )
 
@@ -193,9 +194,14 @@ def whl_library_srcs(
         pkg_name: {type}`str` The label name to use for the py_library target.
         native: {type}`native` The native struct for overriding in tests.
         rules: {type}`struct` A struct with references to rules for creating targets.
+
+    Returns:
+        The source labels attached to the generated `py_library`, or `None` when
+        `rules` does not define `py_library`.
     """
     tags = sorted(tags)
     data = [] + data
+    wrapper_srcs = None
 
     bins_for_data_label = []
 
@@ -301,6 +307,7 @@ def whl_library_srcs(
             # pure-Python code, e.g. pymssql, which is written in Cython.
             allow_empty = True,
         )
+        wrapper_srcs = [pkg_name] if srcs else []
 
         # NOTE: pyi files should probably be excluded because they're carried
         # by the pyi_srcs attribute. However, historical behavior included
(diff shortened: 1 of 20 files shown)


aignas:
I think it would be good to see if we can use //python:none target in these cases. It provides pyinfo provider afterall.

python/private/pypi/whl_library_targets.bzl
```diff
@@ -133,6 +133,7 @@ def whl_library_targets(
             dep_template = dep_template,  # only needed if requires_dist is present
             repo = None,  # set aliases in the same repo
             aliases = {},
+            srcs = wrapper_srcs,
```
aignas: Hmmm. This particular approach means that defining the dependency graphs and the srcs in separate repositories is not going to be possible.

Thoughts:
* I am OK to merge this because this will fix a regression.
* I don't wanna merge it because it is one step back. I am curious how we can resolve this.

I'd prefer a fix in the `py_library` validation logic itself, TBH. Any thoughts?

aignas: Hmm, looking at the tests it does not seem that we are generating the message or at least I cannot find it. So I am thinking that we should attempt solving this with `//python:none` instead.

hartikainen: Thanks for the comments. I'll try poking around if I can get `//python:none` working.

hartikainen: From what I can tell, `//python:none` provides `SentinelInfo`, not `PyInfo`.

I moved the fix into `py_library` validation in 74616c517cac. It now allows only empty `PyInfo` targets while non-Python outputs still fail. I verified that the `py-spy` example linked in the issue. What do you think?

aignas: Wondering if the `to_list` calls here make this lower performance in practice. I am thinking that we should have something like:

```
src_targets:
    config_setting(
        name = "is_empty_srcs",
    )

    # rest is as is currently

deps_targets:
    py_library(
        srcs = select({
            "is_empty": [],
            "default": ["srcs_label"],
        }),
        deps = [
             ":src_label",
             # all other deps,
        ]
```

That way we:
* satisfy the `$(locations library)` to return all src files
* satisfy the empty .py error
* do not change the py_library validation

@rickeylev, any suggestions here?

aignas (approved):
Could not think of anything better than this, TBH. Tried using `FeatureFlagInfo` in the `whl_library_targets` macro, but this is way more elegant.

For future optimization - we are already filtering targets that only have `py` targets - should we do some cleanup here?.

hartikainen:
Thanks for such a quick turnaround!

rickeylev:
Sorry, just saw this PR.

Was py_library.srcs requiring itself to be non-empty? I don't recall it doing that...

In anycase, py_library should allow empty srcs. There are valid reasons to have a py_library with no sources and only deps.

Oh, was the problem file extension restriction? it was requiring .py extension? Which then broke passing a directory artifact as input?

Calling to_list for srcs in the rule impl is mostly fine. The `ctx.files.srcs` API does that for you implicitly.

PR #4064, opened 2026-08-16
test(pytest): support python_versions in pytest_test macro

rickeylev:
test(pytest): support python_versions in pytest_test macro

Testing pytest-based suites across multiple Python versions previously
required manually defining individual targets for each Python version.

Update the \`pytest_test\` macro in test support helpers to accept an
optional \`python_versions\` list. When specified, version-specific
\`pytest_test\` targets with formatted names are generated and grouped
under a root \`test_suite\`.

### DIFF
--- a/tests/pytest_test/BUILD.bazel
+++ b/tests/pytest_test/BUILD.bazel
@@ -34,8 +34,8 @@ pytest_multipy_test(
         "basic_test.py",
     ],
     python_versions = [
-        "3.11",
-        "3.12",
+        "3.14",
+        "3.13",
     ],
     target_compatible_with = SUPPORTS_BZLMOD,
 )
--- a/tests/pytest_test/BUILD.bazel
+++ b/tests/pytest_test/BUILD.bazel
@@ -1,5 +1,12 @@
 load("//tests/support:support.bzl", "SUPPORTS_BZLMOD")
-load("//tests/support/pytest_test:pytest_test.bzl", "pytest_test")
+load(
+    "//tests/support/pytest_test:pytest_test.bzl",
+    "pytest_multipy_test",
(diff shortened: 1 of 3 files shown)


tests/pytest_test/BUILD.bazel
```diff
@@ -20,3 +27,15 @@ pytest_test(
...
+
+pytest_multipy_test(
+    name = "pytest_multipy_default_test",
+    srcs = [
+        "basic_test.py",
+    ],
+    python_versions = [
+        "3.14",
+        "3.13",
+    ],
```
aignas, line 38: I really like this - it is quite important to have this sort of thing. Could we create `pytest_test` macro and accept `python_version` or `python_versions` attributes which would do the underlying wiring by themselves?

aignas (approved):

aignas:
This actually makes me think that this could be the general API for the ruleset:
* `py_library` - `python_versions` adds `target_compatible_with` flags. to mark compatibility.
* `py_test` - `python_versions` tests for all versions.
* `py_binary` - no need to pass `python_versions`?

tests/pytest_test/BUILD.bazel
rickeylev: I actually originally did that but wasn't sure if I liked it. Since it was also your first instinct, lets go with that.



I also remembered config_settings (because of bootstrap_impl tests) and thought: well, why not accept arbitrary settings?

And then the api looked like this:

```
pytest_test(
  config_setting_variants = {
    "py3.14_foo": {
      "@rules_python//python/config_settings:python_version": "3.14",
      "//some:flag": "foo",
    }
  }
)
```

Which, eh...

thoughts?

PR #4070, opened 2026-08-17
feat: add PEP 639 license metadata support

sanchiagarwal0:
## What does this PR do?

Adds support for PEP 639 license metadata in the `py_wheel` rule.

## Changes

- Adds the `license_expression` attribute for SPDX license expressions.
- Adds the `license_files` attribute for license files.
- Uses Metadata-Version 2.4 when PEP 639 license metadata is provided.
- Adds `License-Expression` and `License-File` metadata fields.
- Prevents `license` and `license_expression` from being used together.

## Testing

- `git diff --check` passes.
- Bazel tests were attempted, but dependency downloads from GitHub timed out in the local environment.

## Related issue

N/A

### DIFF
--- a/python/private/py_wheel.bzl
+++ b/python/private/py_wheel.bzl
@@ -251,6 +251,14 @@ be moved under that directory.
         doc = "A string specifying the license of the package.",
         default = "",
     ),
+    "license_expression": attr.string(
+        doc = "An SPDX license expression for the package.",
+        default = "",
+    ),
+    "license_files": attr.label_keyed_string_dict(
+        doc = "License files to include under the .dist-info/licenses/ directory.",
+        allow_files = True,
+    ),
     "project_urls": attr.string_dict(
         doc = ("A string dict specifying additional browsable URLs for the project and corresponding labels, " +
                "where label is the key and url is the value. " +
@@ -419,7 +427,17 @@ def _py_wheel_impl(ctx):
     # Note: Description file and version are not embedded into metadata.txt yet,
     # it will be done later by wheelmaker script.
     metadata_file = ctx.actions.declare_file(ctx.attr.name + ".metadata.txt")
-    metadata_contents = ["Metadata-Version: 2.1"]
+    if ctx.attr.license and ctx.attr.license_expression:
+        fail(
+            "`license` and `license_expression` are mutually exclusive. "
+            + "Please use only one of them."
+        )
+
+    metadata_version = "2.4" if (
+        ctx.attr.license_expression or ctx.attr.license_files
+    ) else "2.1"
+
+    metadata_contents = ["Metadata-Version: %s" % metadata_version]
     metadata_contents.append("Name: %s" % ctx.attr.distribution)
 
     if ctx.attr.author:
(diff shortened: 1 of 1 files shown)


python/private/py_wheel.bzl
```diff
@@ -251,6 +251,14 @@ be moved under that directory.
         doc = "A string specifying the license of the package.",
         default = "",
     ),
+    "license_expression": attr.string(
+        doc = "An SPDX license expression for the package.",
+        default = "",
+    ),
+    "license_files": attr.label_keyed_string_dict(
```
rickeylev, line 258: The license file paths are used, but the files themselves aren't being included in the output

python/private/py_wheel.bzl
```diff
@@ -251,6 +251,14 @@ be moved under that directory.
...
+        doc = "License files to include under the .dist-info/licenses/ directory.",
```
rickeylev, line 259: What is the value for this dict? the path under licenses to put the files under? The behavior needs to be documented.

How are files mapped under the licenses directory? Files could come from elsewhere in the repo

This attribute mostly redundant with the extra_distinfo_files attribute. The only differnce is where in dist-info files go (this puts them under licenses)

Given all this, I think we should just remove this field.

python/private/py_wheel.bzl
```diff
@@ -430,6 +448,12 @@ def _py_wheel_impl(ctx):
         metadata_contents.append("Home-page: %s" % ctx.attr.homepage)
     if ctx.attr.license:
         metadata_contents.append("License: %s" % ctx.attr.license)
+    if ctx.attr.license_expression:
+        metadata_contents.append(
+            "License-Expression: %s" % ctx.attr.license_expression
+        )
+    for _, license_file in sorted(ctx.attr.license_files.items()):
+        metadata_contents.append("License-File: %s" % license_file)
```
rickeylev, line 456: Is every license file supposed to be listed as a License-File header? I'm guessing a PEP specifies this? Which one?

Because extra_distinfo_files could add license files, I'm thinking the population of License-File should move to the execution phase.

rickeylev (changes requested):
Thanks for this!

* PR description needs to be rewritten to be an appropriate commit message.
* Needs tests.

rickeylev:
@sanchiagarwal0 I get the sense you want to see this PR through on your own? I just ask because normally I would throw my agent at a PR and have it address most things. However, if you'd rather work on it yourself, I can hold off on that

python/private/py_wheel.bzl
sanchiagarwal0: Yes. This is specified by PEP 639. In the `license-files` section, it says build tools MUST include all files matched by the configured patterns and MUST list each matched file path under a `License-File` field in the Core Metadata.

The `License-File` field itself is also defined in the Core Metadata section of PEP 639 as a multi-use field, where each instance lists one license-related file.

https://peps.python.org/pep-0639/

sanchiagarwal0:
Yes, I'd like to work on it myself. Thanks for offering to have your agent handle it — I appreciate it! I'll look into the `extra_distinfo_files` handling and the execution phase and make the necessary changes.

PR #4071, opened 2026-08-17
fix(windows): use extended paths in Python bootstraps

belitskiy:
Implicit long-path support is not universal across the Win32 API. The documented set of APIs covered by the long-path opt-in does not include DLL loading functions, e.g., LoadLibraryExW:

https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation#functions-without-max_path-restrictions

Always use extended-length paths in Windows bootstrap code and correctly convert UNC paths to the \\?\UNC\ form.

Experienced this breakage in JAX after the project being switched over to Bzlmod, which made some paths too long:
https://github.com/jax-ml/jax/actions/runs/31674295160/job/94365413380

```
    File "c:\botcode\w\bazel-out\x64_windows-opt\bin\jax\experimental\jax2tf\tests\multiprocess\jax2tf_multiprocess_test_cpu.exe.runfiles\rules_python++pip+jax_pypi_312_ml_dtypes_cp312_cp312_win_amd64_c1a95399\site-packages\ml_dtypes\_finfo.py", line 17, in <module>
      from ml_dtypes._ml_dtypes_ext import bfloat16
  ImportError: DLL load failed while importing _ml_dtypes_ext: The filename or extension is too long.
```

Already used as a patch in https://github.com/jax-ml/jax/pull/39961

### DIFF
--- a/news/4071.fixed.md
+++ b/news/4071.fixed.md
@@ -1,1 +1,1 @@
-Fixed loading Python extension modules from paths longer than `MAX_PATH` on Windows.
+(rules) Fixed loading Python extension modules from paths longer than `MAX_PATH` on Windows.
--- a/news/4071.fixed.md
+++ b/news/4071.fixed.md
@@ -1,0 +1,1 @@
+Fixed loading Python extension modules from paths longer than `MAX_PATH` on Windows.
--- a/python/private/site_init_template.py
+++ b/python/private/site_init_template.py
@@ -93,35 +93,62 @@ def _get_windows_path_with_unc_prefix(path):
     if not _is_windows() or sys.version_info[0] < 3:
         return path
 
-    # Starting in Windows 10, version 1607(OS build 14393), MAX_PATH limitations have been
-    # removed from common Win32 file and directory functions.
-    # Related doc: https://docs.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=cmd#enable-long-paths-in-windows-10-version-1607-and-later
-    import platform
-
-    win32_version = None
-    # Windows 2022 with Python 3.12.8 gives flakey errors, so try a couple times.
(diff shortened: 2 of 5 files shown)


rickeylev:
CI flagged a variety of edge cases: https://buildkite.com/bazel/rules-python-python/builds/17072/list

Would you be able to create a repro for this case? i.e. one that triggers LoadLibraryExW with a long path. There's now a py_extension rule that can create windows DLLs, so it should possible to create a small no-op python c module at a long path location to verify it can be loaded

I'm gonna throw my ai at this and see if it can hash out the failures, too.

belitskiy:
Sorry, I expected to get emails about replies here, but I didn't for some reason (or they were filtered out, etc.), so I missed this.
It seems like you have already done what you asked me to do - please let me know if something else is needed of me.

(fixed notifications, they were marked Spam for some reason...)

rickeylev:
If you could cross check the current state, that'd be helpful. I threw my agent at it, but haven't vetted what it's done.

belitskiy:
I threw an agent at it as well, and it came up with a more filtered solution (explanation in the commit). PTAL

Edit: nvm, buildkite doesn't need approval - will check results myself and then let you know

belitskiy:
Buildkite is also passing, except one unrelated, non-Windows, at HEAD failure.

news/4071.fixed.md
```diff
@@ -0,0 +1 @@
+Fixed loading Python extension modules from paths longer than `MAX_PATH` on Windows.
```
aignas:
```suggestion
(rules) Fixed loading Python extension modules from paths longer than `MAX_PATH` on Windows.
```

aignas (approved):
Thank you

Issue #2703, opened 2025-03-26
exec tools exec_interpreter broken on RBE


rickeylev:
As found in https://github.com/bazel-contrib/rules_python/pull/2657, the `exec_interpreter` attribute of the exec tools toolchain doesn't work with RBE.

What happens is the `current_interpreter_executable()` helper rule does `declare_file(); symlink()`, which puts the file at e.g. `python/private/python3`. Locally, this creates a symlink and when run, Python follows the symlink back to the actual location of the installed runtime ("python home"); this works. On RBE, however, it creates copy of the file. Python then can't find its way back to "python home" and fails to find e.g. shared libs, stdlib, etc.

The intent of the `exec_interpreter` field was to make it easy to run an interpreter in a build action by having it support `DefaultInfo.files_to_run`, which lets it be passed directly to `ctx.actions.run(executable=...)`. Without a FilesToRun provider, the interpreter and its files have to be manually passed in. Additionally, because actions can't directly accept runfiles, it means the usual way to express the runfiles necessary for an executable (DefaultInfo.default_runfiles) can't be used.

-----

While looking into solutions, a promising option looked something like:
* Have py_runtime() return DefaultInfo with executable set.
* In order to do this, re-declare the interpreter file as a file. This, apparently, works, even if it "clobbers" the original file. This is necessary because DefaultInfo.executable requires the file to be produced by the same rule.
* In py_runtime_pair, have it put the DefaultInfo that py_runtime produced on the ToolchainInfo it returned
  * Alternative: create a new PyRuntimeInfo with a replaced interpreter field

Consumers are then able to do e.g.

```
exe = ctx.toolchains["exec_tools"].interpreter # DefaultInfo.files_to_run object
```

Unfortunately, I didn't see a decent way to preserve the "exec_interpreter is a target that you can pass to ctx.actions.run" behavior. _Maybe_ this would be possible if the underlying py_runtime target (which would have DefaultInfo.executable) could be more directly plumbed though. I'm not sure that's possible, though.

PR #4079, opened 2026-08-18
fix(py_test): fix Windows crash in `py_test` main validation

rdesgroppes:
`py_test` main validation (`--validate_test_main=enabled`) crashes on Windows, because it invokes the interpreter via `exec_tools.exec_interpreter`, which resolves through `current_interpreter_executable()`'s relocated copy of the interpreter.

The crash may be reproduced on existing tests in the repo, for instance:
```
bazel build \
  --@rules_python//python/config_settings:validate_test_main=enabled  \
  //tests/validate_test_main:validate_test_main_test
...
ERROR: .../tests/validate_test_main/BUILD.bazel:3:8: Validating py_test
 main //tests/validate_test_main:validate_test_main_test failed:
 (Exit -1073741515): python.exe failed: error executing
 PyValidateTestMain command
```
```
cd tests/integration/validate_test_main
bazel build \
  --@rules_python//python/config_settings:validate_test_main=enabled \
  //:good_test
...
ERROR: .../tests/integration/validate_test_main/BUILD.bazel:11:8:
 Validating py_test main //:good_test failed: (Exit -1073741515):
 python.exe failed: error executing PyValidateTestMain command
```

`-1073741515` is `STATUS_DLL_NOT_FOUND`: the relocated copy can't find its DLLs beside itself.

This is also the Windows-local manifestation of #2703 (`exec_interpreter` broken on RBE): the same relocation severs the interpreter from files resolved relative to itself, just triggered differently: RBE's copy materialization there, a lack of symlink privilege here.

Colocating the DLLs alone (a first attempt) traded this for a second, still fatal error, `ModuleNotFoundError: No module named 'encodings'`, because the copy still can't find its stdlib.
Patching each missing file individually doesn't scale: DLLs today, stdlib tomorrow, whatever else a future toolchain needs beside itself after that.

`_maybe_add_test_main_validation` now uses `actions_run()` with `exec_runtime` instead of `exec_tools.exec_interpreter`'s relocated `DefaultInfo.files_to_run`, matching `PyExecToolsInfo.exec_interpreter`'s own documented recommendation and the pattern `common.bzl`'s `actions_run()` and `py_zipapp_rule.bzl` already use.
`exec_runtime.interpreter` is the real file, used directly, with its real files as plain action inputs, so nothing is relocated and nothing loses its siblings.

With the proposed fix[^1], above examples now build cleanly, with no relocated runfiles tree for the interpreter at all, and `tests/integration/validate_test_main`'s `inert_test` still fails with its intended "will not run any tests" message rather than a crash.

`tests/integration/validate_test_main_test` is the corresponding integration test, but it was not exercised on Windows, where it was failing on `OSError: [WinError 193] %1 is not a valid Win32 application` in `tests/integration/runner.py`'s Bazel-in-Bazel invocation, itself unable to run `bazel_from_env`'s `#!` shebang line the way POSIX's `exec` does.
The present change therefore fixes this, by resolving the shebang's interpreter itself, and enables the test on Windows.

[^1]: This does not fix `exec_interpreter`/#2703 itself: `precompile.bzl` still resolves the interpreter via the relocated path and would need the same migration.

### DIFF
--- a/news/4079.fixed.md
+++ b/news/4079.fixed.md
@@ -1,0 +1,4 @@
+(toolchain) Fixed a crash in `py_test` main validation
+(`validate_test_main`) on Windows: the interpreter used for the check
+couldn't find its DLLs or its stdlib once relocated
+([#4079](https://github.com/bazel-contrib/rules_python/pull/4079)).
--- a/news/4079.fixed.md
+++ b/news/4079.fixed.md
@@ -1,4 +1,4 @@
-(toolchain) Fixed a crash in `py_test` main validation
-(`validate_test_main`) on Windows: the interpreter used for the check
+(toolchain) Fixed a crash in {obj}`py_test` main validation
+({obj}`validate_test_main`) on Windows: the interpreter used for the check
 couldn't find its DLLs or its stdlib once relocated
-([#4079](https://github.com/bazel-contrib/rules_python/pull/4079)).
+([#4079](https://github.com/bazel-contrib/rules_python/issues/4079)).
--- a/python/private/py_executable.bzl
+++ b/python/private/py_executable.bzl
@@ -64,7 +64,6 @@ load(":py_cc_link_params_info.bzl", "PyCcLinkParamsInfo")
 load(":py_executable_info.bzl", "PyExecutableInfo")
 load(":py_info.bzl", "PyInfo", "VenvSymlinkKind")
 load(":py_internal.bzl", "py_internal")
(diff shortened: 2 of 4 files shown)


rickeylev (approved):
Thanks for the clean PR! Just needed a couple style fixes. Otherwise LGTM

rdesgroppes:
> Thanks for the clean PR! Just needed a couple style fixes. Otherwise LGTM

@rickeylev merci !

PR #4091, opened 2026-08-22
fix(toolchain): exclude libpython from runtime in recent releases

### LINKED ISSUE #3534: bin/python size exploded 2000 times larger
Is there a reason why the binary bin/python3.10 in latest python 3.10.19 has size 43963640 which is more than 2000 times larger than bin/python3.10 in python 3.10.15 (size 20944)?

```
python.toolchain(
    is_default = True,
    python_version = "3.10.15",
)
```

aignas:
Use the already existing auto-detection and exclude libpython and
friends as suggested in the upstream ticket.

Fixes #3534

### DIFF
--- a/news/4091.fixed.md
+++ b/news/4091.fixed.md
@@ -1,0 +1,5 @@
+(toolchain) users can now request libpython to be not implemented via
+{target}`//python/config_settings:py_runtime_include_libpython=no`.
+The default is to include it, but it may change in the future. Ensure
+that you include `libpython` via the `<toolchain_repo>//:libpython` if
+it is actually needed at runtime.
--- a/python/config_settings/BUILD.bazel
+++ b/python/config_settings/BUILD.bazel
@@ -170,6 +170,15 @@ string_flag(
 )
 
 config_setting(
+    name = "_is_py_runtime_include_libpython_auto",
+    flag_values = {
+        ":py_runtime_include_libpython": PyRuntimeIncludeLibPython.AUTO,
+    },
+    # NOTE: Only public because it is used in python_repository repos.
+    visibility = NOT_ACTUALLY_PUBLIC,
+)
+
+config_setting(
     name = "_is_py_runtime_include_libpython_yes",
     flag_values = {
         ":py_runtime_include_libpython": PyRuntimeIncludeLibPython.YES,
@@ -179,9 +188,9 @@ config_setting(
 )
 
 config_setting(
-    name = "_is_py_runtime_include_libpython_auto",
+    name = "_is_py_runtime_include_libpython_no",
     flag_values = {
-        ":py_runtime_include_libpython": PyRuntimeIncludeLibPython.AUTO,
+        ":py_runtime_include_libpython": PyRuntimeIncludeLibPython.NO,
     },
     # NOTE: Only public because it is used in python_repository repos.
     visibility = NOT_ACTUALLY_PUBLIC,
--- a/python/config_settings/BUILD.bazel
+++ b/python/config_settings/BUILD.bazel
@@ -10,6 +10,7 @@ load(
     "LibcFlag",
     "PrecompileFlag",
     "PrecompileSourceRetentionFlag",
+    "PyRuntimeIncludeLibPython",
     "ValidateTestMainFlag",
     "VenvsSitePackages",
     "VenvsUseDeclareSymlinkFlag",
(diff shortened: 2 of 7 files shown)


aignas:
Re-read the comment and realized that this should be config flag gated, will PoC the interface

rickeylev:
Re: interface: I had three thoughts 

One: a coarse/vague string flag like minimal/full/some-use-case-name

Two: a string flag of pieces to include/exclude, similar to how output group flag works. -libs,+pip,-headers,-tcl,etc

Three: a flag to pick the runtime flavor, ie install_only (which is a 26M install compared to the other tars) (this flag will be added regardless, eventually)

aignas:
I am done for the day so I'll quickly respond:

> One: a coarse/vague string flag like minimal/full/some-use-case-name

I chose to be very specific here with respect to what we are including. I chose to have a runtime inclusion behaviour because we already have the thing extracted, so we can support various configurations given the extracted contents and what is present.

> Two: a string flag of pieces to include/exclude, similar to how output group flag works. -libs,+pip,-headers,-tcl,etc

This is a nice idea, but could feel clunky? Maybe not... The nice thing here is that we could trampoline off of what user supplies and use FeatureFlagInfo to get derived flags for what we need. I like this because it keeps the API surface small and extendible.

> Three: a flag to pick the runtime flavor, ie install_only (which is a 26M install compared to the other tars) (this flag will be added regardless, eventually)

This can be already achieved via what is provided to the machinery as the URLs. I thought the `install_only` still has the libpython?

Feel free to hack on this and mold it to your liking.

news/4091.fixed.md
```diff
@@ -0,0 +1,5 @@
+(toolchain) users can now request libpython to be not implemented via
+{target}`//python/config_settings:py_runtime_include_libpython=no`.
+The default is to include it, but it may change in the future. Ensure
+that you include `libpython` via the `<toolchain_repo>//:libpython` if
```
rickeylev, line 4: We shouldn't tell people to reference the toolchain repos directly, as those are implementation details.

The py_cc_libs target is the one to depend on if build-time linking. I suppose that could be used to get libpython into runtime data deps with some extra steps. (depending on py_cc_libs might already do that? not sure).

What's the use case for a target to have libpython at runtime when the toolchain says to not included it?

python/config_settings/BUILD.bazel
```diff
@@ -161,6 +162,40 @@ string_flag(
...
 
+string_flag(
+    name = "py_runtime_include_libpython",
+    build_setting_default = PyRuntimeIncludeLibPython.YES,
+    values = PyRuntimeIncludeLibPython.flag_values(),
+    visibility = ["//visibility:public"],
+)
+
+config_setting(
+    name = "_is_py_runtime_include_libpython_auto",
```
rickeylev, line 173: Move the private config setting definitions into the bzl code; there's a helper in there that defines all the various private config settings

python/private/python_repository.bzl
```diff
@@ -218,26 +229,25 @@ def _python_repository_impl(rctx):
...
+        # building on.
+        #
+        # Link to the first affected release:
+        # https://github.com/astral-sh/python-build-standalone/releases/tag/20240224
+        rctx.delete("share/terminfo")
+
+    if release and release >= 20250517:
+        # Starting with 20250517 we have python3 linked statically
+        # https://github.com/astral-sh/python-build-standalone/issues/941
+        python3_statically_links_libpython = False
```
rickeylev: This logic looks backwards? The conditional doesn't match the comment

python/private/python_repository.bzl
```diff
@@ -200,6 +200,17 @@ def _python_repository_impl(rctx):
     elif rctx.attr.distutils_content:
         rctx.file(distutils_path, rctx.attr.distutils_content)
 
+    # Support not including libraries into runtime
```
rickeylev, line 203: nit: remove comment. The release is used for other things, too.

python/private/python_repository.bzl
```diff
@@ -218,26 +229,25 @@ def _python_repository_impl(rctx):
...
+        # on host platforms that are not Linux. Delete the files no
+        # matter the host platform so that the cross-built artifacts
+        # are the same irrespective of the host platform we are
+        # building on.
+        #
+        # Link to the first affected release:
+        # https://github.com/astral-sh/python-build-standalone/releases/tag/20240224
+        rctx.delete("share/terminfo")
+
+    if release and release >= 20250517:
```
rickeylev, line 245: Hrm. I don't like this way of deciding it, but not sure what options we have. I'm willing to live with this, but that said:

It would be best to check if the executable has dt_needed for libpython somehow. A build rule that looks at python3 and then conditionally emits the libpython file or...not sure what the "or" is. error or empty file? Maybe a params file of link opts?

If there was a prebuilt patchelf or some such available, we could use that during the repo phase. Or just call out to system patchelf, if available.

I've really been wanting to create some of our own repo-phase tools for stuff like this.

python/private/hermetic_runtime_repo_setup.bzl
```diff
@@ -80,15 +93,28 @@ def define_hermetic_runtime_toolchain_impl(
     ]
     files_exclude += extra_files_glob_exclude
 
-    native.filegroup(
+    native.alias(
```
rickeylev, line 96: I'd say remove the alias. The extra level of indirection for the slightly clearer "_all" name doesn't seem worth it.

rickeylev (commented):
Needs test. I think a simple py_runtime analysis test that verifies libpython isn't in the files is sufficient.

python/private/python_repository.bzl
aignas: Yes it is backwards, I renamed the param too many times.
```suggestion
        python3_statically_links_libpython = True
```

PR #4099, opened 2026-08-23
chore: migrate requirements.in files to pyproject.toml

rickeylev:
Standardize Python dependency declarations across tools, examples, and
tests to use standard pyproject.toml configuration instead of
requirements.in files.

Replace requirements.in inputs with pyproject.toml, update corresponding
BUILD.bazel and documentation references, and regenerate all associated
requirement lockfiles and Gazelle manifests.


aignas (approved):
I think we have one place where we have integration tests checking that we can consume `requirements.in`, so this is a great start to cleanup the codebase.

Thank you!

rickeylev:
Yes, feels good to clean this up. I'm going start another PR to switch them over to the uv-based locking next.

PR #4100, opened 2026-08-23
fix(runfiles): update Path method signatures for Python 3.14 compatibility and fix match()

rickeylev:
Align runfiles.Path method signatures with Python 3.14 typeshed stubs to
avoid type checker errors on newer Python releases. This also fixes a
bug where pattern matching failed on Python 3.12+.

This change splits the runfiles compatibility updates from PR #4023.


aignas (approved):
I really like the `python_versions` attr for things like this.

PR #4104, opened 2026-08-23
fix(bootstrap): remap stdlib sys.path entries to runfiles during site init

rickeylev:
When CPython resolves interpreter symlinks during startup, it leaks
underlying repository cache or external repository paths into sys.path
and sys prefixes instead of their runfiles locations.

This breaks runfiles isolation and prevents using a mixture of
generated and non-generated files as part of the python runtime, as
standard library modules and runtime site-packages are loaded from
locations outside the runfiles tree.

Fix this by passing interpreter_actual_path into site_init_template.py
during virtual environment creation. During site initialization,
collect any non-runfiles sys prefixes into a set and remap matching
sys.path entries, sys prefixes, and site.PREFIXES to the runfiles
runtime root.

Also adds a reproduction test target
(stdlib_symlink_syspath_bootstrap_script_test) to verify that sys.path
entries remain within the runfiles tree.

Work towards https://github.com/bazel-contrib/rules_python/issues/1653


rickeylev:
For context, this came out of trying to zip the stdlib.

What happens is the `python3` file is a symlink that points back to the extracted repo directory. Since Python reads through symlinks when computing its initial sys.path, it ends up putting the repo download directory on sys.path. Since all the Python runtime files are there, things end up working.

If a generated file is added to the runtime's files (e.g. adding a generated file to py_runtime), it gets put into the exec root. Since the exec root was never added to sys.path, those files aren't importable. It also means the runtime's files are split up into two directories (the repo directory and the execroot directory). Normally runfiles unifies things (as happens with regular libraries), but Python's resolving the symlinks foils that.

The only other option I see is to turn all the files in the runtime into generated files. i.e. make py_runtime loop over every file and symlink it (and copy python3 itself so it doesn't point back to the repo directory). This wouldn't get the runfiles directory onto sys.path, but would put all the runtime files in one location (the execroot). I'm not sure if that's better or not than futzing with sys.path at startup.

aignas (approved):
Nice find. please add `news`. People may look at it sometime in the future, or we may need to refer to this.

rickeylev:
Added news entry in `news/4104.fixed.md`.

rickeylev:
agent has been getting a bit wonky with this PR, so adding do-not-merge in case it gets too creative

gpshead:
I think you could set PYTHONHOME=, but that is gross and carries the caveats of depending on the environment that various things will break.

I guess the _real_ desirable fix here is your [draft PEP-796: Relative Virtual Environments](https://github.com/python/peps/pull/4476/files) aka https://github.com/python/cpython/issues/136051.

rickeylev:
Yeah, that's the real fix.

Using PYTHONHOME is tricky because running sys.executable directly needs to work, and PYTHONHOME has to be set before the interpreter is set. Which means sys.executable has to point to some sort of wrapper. Which then interferes with sys.prefix calculation. There's a stack of hacks in [runtime_env_interpreter.sh](https://github.com/bazel-contrib/rules_python/blob/main/python/private/runtime_env_toolchain_interpreter.sh) to make that type of thing work, but it relies on PYTHONEXECUTABLE (which I saw discussion on removing entirely)

PR #4105, opened 2026-08-23
feat(pypi): allow defining deps in through pkg_aliases macro

aignas:
Summary:
- If `requires_dist` are passed we create a `py_library` using the
  aliased `srcs` target.
- Currently this is only exercised by tests.

Split out of #4059

### DIFF
--- a/python/private/pypi/BUILD.bazel
+++ b/python/private/pypi/BUILD.bazel
@@ -342,6 +342,7 @@ bzl_library(
     srcs = ["pkg_aliases.bzl"],
     deps = [
         ":labels",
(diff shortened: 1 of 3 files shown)


rickeylev (changes requested):
Is this actually feat? Looks more like part of the internal refactoring you're doing?

aignas:
This will become a `feat` when we can use the `uv.lock` dep-graph. This change is not a refactor - defining the dep graph in the hub repo is something new.

I am OK to mark this as `internal(pypi): ...`.

rickeylev (approved):

PR #4106, opened 2026-08-23
fix: various fixes for the deps library

aignas:
Correctly wire args around and disable precompiling
for the deps nodes.

Work towards #2948

### DIFF
--- a/python/private/pypi/whl_library_deps_targets.bzl
+++ b/python/private/pypi/whl_library_deps_targets.bzl
@@ -168,7 +168,7 @@ def whl_library_deps_targets(
                 package_deps = package_deps,
                 tmpl = dep_template.format(name = "{}", target = PY_LIBRARY_PUBLIC_LABEL),
             ),
-            # do not precomple `pyc` because this does not have any sources. The sources
+            # do not precompile `pyc` because this does not have any sources. The sources
             # will be pre-compiled in the srcs folder once and for all.
             precompile = "disabled",
             tags = tags,
--- a/python/private/pypi/whl_library_deps_targets.bzl
+++ b/python/private/pypi/whl_library_deps_targets.bzl
@@ -23,8 +23,8 @@ def whl_library_deps_targets(
         repo,
         aliases = None,
         metadata_name,
-        requires_dist,
-        extras,
+        requires_dist = [],
+        extras = [],
         include = [],
         group_deps = [],
         group_name = None,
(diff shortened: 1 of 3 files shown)


python/private/pypi/whl_library_deps_targets.bzl
```diff
@@ -168,6 +168,9 @@ def whl_library_deps_targets(
                 package_deps = package_deps,
                 tmpl = dep_template.format(name = "{}", target = PY_LIBRARY_PUBLIC_LABEL),
             ),
+            # do not precomple `pyc` because this does not have any sources. The sources
+            # will be pre-compiled in the srcs folder once and for all.
+            precompile = "disabled",
```
rickeylev, line 180: This sounds like a bug in precompiling.

Precompile should be OK with empty sources.

Is it actually for directory sources?

python/private/pypi/whl_library_deps_targets.bzl
```diff
@@ -46,8 +46,8 @@ def whl_library_deps_targets(
             direct cycles. These dependencies will be provided at runtime by the
             group rules which wrap this library and its fellows together.
         requires_dist: {type}`list[str]` The list of `Requires-Dist` values from
-            the whl `METADATA`.
-        extras: {type}`list[str]` The list of requested extras. This essentially includes extra transitive dependencies in the final targets depending on the wheel `METADATA`.
+            the whl `METADATA`. Optional because some packages don't have them.
+        extras: {type}`list[str]` The list of requested extras. This essentially includes extra transitive dependencies in the final targets depending on the wheel `METADATA`. Optional because some packages don't request them.
```
rickeylev: nit: wrap at 80

python/private/pypi/whl_library_targets.bzl
```diff
@@ -86,16 +86,18 @@ def whl_library_targets(
...
-        group_deps = group_deps,  # only needed if group_name is present
-        group_name = group_name,  # must specify group_deps together
-        extras = extras,  # only needed if requires_dist is present
-        include = include,  # only needed if requires_dist is present
-        repo = None,  # set aliases in the same repo
-        aliases = {},
-        **kwargs
-    )
+    if deps_template:
+        # allow to not define the targets if we invoke `whl_archive` without deps_template
```
rickeylev: code self describes what this comment says. rephrase comment to be more informative or remove comment

rickeylev (approved):

rickeylev (commented):
this is titled fix, but there's no news entry. is it fixing a released user visible behavior? If so, please add news

python/private/pypi/whl_library_deps_targets.bzl
rickeylev: Ohh, is the idea here to pass a py_library as sources? And then precompile has to be disabled because, if the py_library is generating pyc, then those end up in srcs, and then the precompiler gives an error?

In any case, I have a WIP to make precompiling work with directory inputs and handle existing pyc in sources.

aignas:
> this is titled fix, but there's no news entry. is it fixing a released user visible behavior? If so, please add news

The APIs have never been said to be public and available to be used. Though, I can add a news item if needed.

rickeylev (approved):

rickeylev:
I'm fine with omitting a news entry. I'm fine with omitting news for non-public behavior

PR #4110, opened 2026-08-24
chore(examples): replace compile_pip_requirements with uv lock()

rickeylev:
Using `pip-compile` via `compile_pip_requirements` is slower and being
superseded by `uv` for requirement locking under Bzlmod. Demonstrating
`lock()` in the examples ensures users follow modern best practices
and benefit from faster dependency resolution.

Migrate Bzlmod-enabled example workspaces and documentation to use the
`lock()` macro in place of `compile_pip_requirements`, updating the
associated requirements targets and lockfiles. Additionally,
canonicalize platform labels within `lock()` so submodules resolve
targets reliably.

DO NOT MERGE: consider making uv auto-used before switching everything
over to it.


examples/bzlmod/other_module/MODULE.bazel
```diff
@@ -40,3 +40,12 @@ pip.parse(
     requirements_lock = ":requirements_lock_3_11.txt",
 )
 use_repo(pip, "other_module_pip")
+
+# DO NOT MERGE: consider making uv auto-used before switching
+# everything over to it.
```
aignas, line 45: What about `uv` not being present in WORKSPACE?

aignas:
Do we need #4029 to be able to fully switch?

examples/bzlmod/other_module/MODULE.bazel
rickeylev: This is a good question. CI didn't have any failures for workspace mode. Which is weird -- how is that possible? Shouldn't at least one of them failed?

rickeylev:
> Do we need #4029 to be able to fully switch?

I don't see why? That PR seems more like an ergonomic fix, e.g. being able to generate a lock for something in an arbitrary location. Each sub-module will still need its own lock() target, no?

aignas:
> Each sub-module will still need its own lock() target, no?

My thinking was that:
* each submodule owns the lock file target.
* with the `directory` we can add extra targets at the root `rules_python` workspace where we could run the locking for all submodules at once. We would need to set `--directory` and `lock_target` to make this work.

But maybe it is fine to not have that ergonomic feature.

rickeylev:
> my thinking was ... so we can run all locking at once

That actually sounds appealing. I was trying to process the dependabot backlog and it mostly boiled down to "update requirements" in all the various examples and integration tests. AI made this easier, but it was still a bit of a chore having to find and run all the lock files. Presumably we would just have a single e.g. `//requirements:update_all` target?

aignas:
Yeah, I started that in the `//private` package, but never had the bandwidth to finish.

PR #4113, opened 2026-08-25
fix(precompile): skip precompilation for directory and pyc inputs in srcs

rickeylev:
Precompiling `py_library`, `py_binary`, or `py_test` targets fails when
`srcs` contains directory artifacts or `.pyc` files. The precompiler
declares invalid actions on directories and drops non-`.py` sources from
default outputs.

This change:
* Expands direct sources to include Python files, `.pyc` files, and
  directories.
* Skips precompilation for directories and non-Python files while
  retaining them in default outputs.
* Updates `srcs` validation and documentation.


python/private/py_library.bzl
aignas: This can be removed.

aignas: Should this be also consolidated in the `common.bzl`?

aignas (approved):
Pending the comments, LGTM.

python/private/py_library.bzl
rickeylev: Done

rickeylev: Done

PR #4120, opened 2026-08-28
fix(pypi): include none and any tags in wheel selection and fix prerelease platform parsing

rickeylev:
Pure-Python wheels with the `none` ABI tag and `any` platform tag fail
to match target platform configurations when explicit platform tags
omit them. In addition, prerelease Python version strings (such as
3.13.0rc1) cause improper platform key formatting in
`requirements_files_by_platform`.

To resolve these issues:
* Add `none` ABI tags and ensure `any` platform tags are present in
  platform descriptors within hub builder and extension defaults.
* Use `version.parse` in `requirements_files_by_platform` to format
  prerelease Python version platform prefixes correctly.
* Support exact platform names without trailing wildcards in
  `requirements_files_by_platform`.


aignas (commented):
This is a confusing PR, the `select_whl` should do `none-any` wheel selection by default. There should be no need to change all of this. I think the best way would be to change only that part.

EDIT: https://github.com/bazel-contrib/rules_python/blob/main/python/private/pypi/platform.bzl#L31

This is the code that I had in mind writing the comment, it is not the `select_whl` but the `platform` construction code. That part will add `none` and `any` if they are not present. So I am curious what is happening here.

PR #4135, opened 2026-09-01
fix(gazelle): fix module map for old-style namespaces

### LINKED ISSUE #3528: Wrong Gazelle mappings after upgrading to 1.8
We are seeing a wrong change for a specific dependency in gazelle_python.yaml after upgrading the Gazelle plugin from 1.6 to 1.8.1.

Image

```
# requirements.in
bosdyn-client==5.0.1.2
bosdyn-orbit==5.0.1.2
```

Build error:
```
E   ModuleNotFoundError: No module named 'bosdyn.client'
```

I suspect it was introduced by https://github.com/bazel-contrib/rules_python/pull/3415/changes

Adding manual resolve directives solves the build errors, but that's just a workaround.
```
# gazelle:resolve py bosdyn.api @python_deps//bosdyn_api
# gazelle:resolve py bosdyn.bddf @python_deps//bosdyn_core
# gazelle:resolve py bosdyn.client @python_deps//bosdyn_client
# gazelle:resolve py bosdyn.deprecated @python_deps//bosdyn_core
# gazelle:resolve py bosdyn.geometry @python_deps//bosdyn_core
# gazelle:resolve py bosdyn.util @python_deps//bosdyn_core
```

gholms:
The performance tweak in #3415 which parallelized gazelle module map generation calls simplify() on wheels individually, which always results in entries which collapse to the highest level modules which contain __init__.py files.  But when two wheels do old-style namespace packages by both including that file under their top-level module, they simplify down to the top-level module, clobber each other when merged, and wind up with one wheel getting mapped for the entire namespace. This causes bug #3528.

We fix this by deferring our call to simplify() until after we've merged all the wheels' module info.

Fixes #3528

### DIFF
--- a/gazelle/modules_mapping/generator.py
+++ b/gazelle/modules_mapping/generator.py
@@ -54,23 +54,6 @@ class Generator:
                 else:
                     self.module_for_path(path, whl)
 
-    def simplify(self):
-        simplified = {}
-        for module, wheel_name in sorted(self.mapping.items(), key=lambda x: x[0]):
-            mod = module
-            while True:
-                if mod in simplified:
-                    if simplified[mod] != wheel_name:
-                        break
-                    wheel_name = ""
-                    break
-                if mod.count(".") == 0:
-                    break
-                mod = mod.rsplit(".", 1)[0]
-            if wheel_name:
-                simplified[module] = wheel_name
-        self.mapping = simplified
-
     def module_for_path(self, path, whl):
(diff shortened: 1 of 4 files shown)


aignas (approved):
I can see that this is just code move and it makes sense, please add `news/4135.fixed.md` file describing what is fixed. I am happy to backport it to the current dot release.

aignas (approved):

PR #4137, opened 2026-09-02
build: configure repository cache in .bazelrc

rickeylev:
Configure a shared Bazel repository cache and enable hardlinks for
local builds.

Developer workflows that use short-lived worktrees (such as agentic
development) repeatedly re-download external dependencies and consume
excessive disk space.


aignas:
Nit, if possible, we could also turn on remote repo contents cache to test for regressions in #3791.

aignas (approved):

PR #4141, opened 2026-09-03
feat: Add win32 versions

rsartor-cmd:
Adds runtimes for win32, versions 3.9.25, 3.10.20, 3.11.15, 3.12.13, 3.13.13, and 3.14.4. This is useful for some downstream consumers (e.g. protobuf) who need to support that. This doesn't affect any other versions or platforms.


rickeylev:
LGTM. 32bit windows is pretty niche, but I don't see a big issue with adding it. We've refactored the way toolchains and listed and registered to make it much easier to add such niche cases. Windows is also tier 2, so adding this without tests also SGTM.

rickeylev (approved):

PR #4151, opened 2026-09-06
refactor(zipapp): implement rust exe_zip_maker program

rickeylev:
Creating self-executable zip archives currently relies on Python script
execution during builds, which incurs interpreter startup overhead and
requires a Python runtime.

Provide a compiled Rust implementation of exe_zip_maker. The tool
computes the SHA-256 digest of the input zip archive, substitutes the
%ZIP_HASH% placeholder within the executable preamble script, and
concatenates the modified preamble with the zip payload.

This change only adds a Rust implementation. Subsequent changes will
wire it into the overall build process as a prebuilt tool.


tests/exe_zip_maker/exe_zip_maker_test.rs
```diff
@@ -0,0 +1,155 @@
+use std::env;
+use std::fs;
+
+use exe_zip_maker_lib::{
+    compute_file_sha256_hex, create_exe_zip, replace_bytes, PLACEHOLDER,
+};
+
+#[test]
```
aignas, line 8: My thinking is that we should write idiomatic Rust code and what I've seen in the past is:
* Rust code with unit tests is the same source file.
* Integration tests is a separate file.

What do you think about this convention?

crates/BUILD.bazel
```diff
@@ -0,0 +1,10 @@
+package(default_visibility = ["//:__subpackages__"])
```
aignas, line 1: Should we have a convention to keep the rust tools part of `//tools`?

aignas (commented):
This is exciting!

tests/exe_zip_maker/exe_zip_maker_test.rs
rickeylev: It originally generated code like that and I thought it was just being lazy. Another person confirmed it, too, though. I think that's a wacky convention. 

But, if that'd idiomatic, then, well, when in Rust-ome

aignas: I personally like that the tests become sort of executable docs.

But I agree, it's a little whacky :D

Though it allows to keep the exposed symbols to minimum and keep the unit tests exercising the implementation details in optimum way.

Kind of no longer need to do "buildifier ignore private" if you follow this convention.

rickeylev: Hm, I also saw the rust_test target was outside of //tests. I'm more resistant to that -- I really like being able to run `bazel test //tests/...` to capture almost all the tests. It also helps keep test-only bzl code from sneaking inside the non-test directories

crates/BUILD.bazel
rickeylev: maybe? idk.

IMHO, tools/ is for things that are run as build tools. That would be most of the rust stuff, but there's a couple cases that wouldn't be:

1. When we replace the shell bootstrap with a native executable bootstrap
2. Similar for entry points in the venv bin/ directory (right now its a polyglot shell thing)

There were two other reasons I ended up going with the `crates/<project>` layout
1. From what I could tell, `crates/<project>` seems the common idiom? This was just after some light searching, didn't look universal
2. We had previously talked about creating a src/ directory for python stuff, e.g. segmenting code by language

I don't have a strong opinion, though I haven't thought about this much, and my brain is mostly mush right now.

The most important thing on my mind is preventing unnecessary dependencies from sneaking into prod code (e.g. a dev/test-only load occurs and then a BULID file can't load)

aignas: OK, I am happy with `//crates` for the rust stuff. We could also strip that sub-tree from the distributable rules when we start shipping binaries of those rust tools.

tests/exe_zip_maker/exe_zip_maker_test.rs
aignas: The rust test code is stripped by the compiler, so technically speaking it won't sneak in.

I like what you are saying about `bazel test //tests/...` but I think not going against the grain with the tooling might be good, so solution-wise we could:
* Have the unit test code together with the main code as it is idiomatic.
* Define the test targets together in the tools. Each sub-package in `//crates` should be self-contained and we could have a convention `//crates/foo:test`.
* Add a `native.test_suite` to include the tests for the tools.

I'd be happy to merge the code as is and move along and learn how to work with rust in this repo rather than bikeshed it too much at this point.

aignas (approved):

PR #4152, opened 2026-09-06
refactor: introduce helper for defining distribution filegroups

rickeylev:
Previously, distribution filegroups required manually enumerating each
direct subpackage's distribution target in `srcs`. This created
repetitive boilerplate across BUILD files and risked omitting newly
added subpackages from release distributions.

Introduce the `distribution_filegroup` macro to automatically glob
package files and discover direct subpackages using
`native.subpackages()`, with support for an `exclude` list.

* Also updates BUILD files across the repository to use the new macro


aignas (approved):
Very elegant, thanks

PR #4154, opened 2026-09-07
feat(toolchain): wrap exec interpreter in launcher script for build actions

### LINKED ISSUE #2703: exec tools exec_interpreter broken on RBE

rickeylev:
Build actions that invoke the toolchain's execution-platform Python
interpreter directly fail under sandboxing or remote execution because
bare binaries and raw symlinks lack the necessary runfiles tree and
PYTHONHOME environment to locate standard library modules hermetically.

Wrap the interpreter in platform-specific launcher scripts for POSIX
and Windows that resolve the underlying interpreter through standard
Bazel runfiles lookup mechanisms and set PYTHONHOME to point to the
runfiles root. Wire these launchers into current_interpreter_executable
and interpreter_binary, and run them through actions_run and
py_console_script_gen.

Fixes #2703

### DIFF
--- a/python/private/BUILD.bazel
+++ b/python/private/BUILD.bazel
@@ -33,6 +33,8 @@ package(
 licenses(["notice"])
 
 exports_files([
+    "interpreter_tmpl.bat",
+    "interpreter_tmpl.sh",
     "runtime_env_toolchain_interpreter.sh",
     "runtimes_manifest.txt",
     "runtimes_manifest_workspace.bzl",
@@ -205,6 +207,14 @@ py_library(
     visibility = [
         "//tests/entry_points:__pkg__",
     ],
+)
+
+filegroup(
+    name = "exec_interpreter_template",
+    srcs = select({
+        "@platforms//os:windows": ["interpreter_tmpl.bat"],
+        "//conditions:default": ["interpreter_tmpl.sh"],
+    }),
 )
 
 # The current toolchain's interpreter as an excutable, usable with
@@ -537,6 +547,7 @@ bzl_library(
     name = "py_exec_tools_toolchain",
     srcs = ["py_exec_tools_toolchain.bzl"],
     deps = [
+        ":common",
         ":common_labels",
         ":py_exec_tools_info",
         ":sentinel_impl",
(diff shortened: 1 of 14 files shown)


aignas:
Curious if we could use https://github.com/hermeticbuild/hermetic-launcher here instead of creating the script templates. Nothing against the template approach though, just wanted to bring this up as an alternative.

PR #4156, opened 2026-09-08
feat: Add win32 python versions

rsartor-cmd:
Re-run of #4141 for the 2.3 release branch instead of main.

Adds runtimes for win32, versions 3.9.25, 3.10.20, 3.11.15, 3.12.13, 3.13.13, and 3.14.4. This is useful for some downstream consumers (e.g. protobuf) who need to support that. This doesn't affect any other versions or platforms.


rickeylev (approved):
while this is technically a feat, its low risk enough we think its fine to backport to the 2.3 branch

PR #4161, opened 2026-09-15
fix(zipapp): forward termination signals

### LINKED ISSUE #3809: Zipapp doesn't forward signals (SIGTERM and friends) to the process, resulting in orphaned processes and no cleanup
# 🐞 bug report

### Affected Rule

 The issue is caused by the rule: `py_binary`

### Is this a regression?

 Yes, the previous version in which this bug was not present was: ....

Unsure

### Description

When running a `py_binary` zipapp (E.g. `bazel build --build_python_zip <tgt>`) signals such as `SIGTERM` aren't forwarded to the binary itself, meaning the process doesn't get a chance to clean up.

I suspect https://github.com/bazel-contrib/rules_python/blob/25ad57c85f0094889520486cd9a5738278fd0541/python/private/zipapp/zip_main_template.py#L279-L281 is the source of the issue

## 🔬 Minimal Reproduction

Given

```
import signal

signal.signal(signal.SIGTERM, lambda _1, _2: print("SIGTERM"))

while True:
    pass
```

Then `bazel run <tgt>` and a `kill -s TERM <pid>` will print `SIGTERM`. But `bazel build --build_python_zip <tgt>` followed by a `python <tgt>.zip` and a `kill -s TERM <pid>` (of the outer process) results in the process being terminated and the underlying process being orphaned 💔 

## 🔥 Exception or Error
 
N/A

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
x64 Ubuntu
  </code>
</pre>

<pre>
  <code>
Build label: 8.6.0
Build time: Thu Feb 26 19:55:20 2026 (1772135720)
Build timestamp: 1772135720
Build timestamp as int: 1772135720
  </code>
</pre>

**Rules_python version:**

<pre>
  <code>
1.9.0
  </code>
</pre>

**Anything else relevant?**

jpneufeld:
Executable ZIP launchers can keep an intermediate process alive for cleanup, so signals sent to the launcher do not reach the application. The same issue occurs when a zipapp is run as `python app.pyz`.

This change:

* uses `exec` when the stage 2 bootstrap owns ZIP cleanup;
* forwards `SIGHUP`, `SIGINT`, `SIGQUIT`, and `SIGTERM` from POSIX Python launchers;
* keeps `Popen.wait()` as the only child reaper and preserves nonzero or signal-derived exit statuses; and
* leaves Windows subprocess behavior unchanged.

Regression tests cover executable zipapps, `python app.pyz`, and both legacy ZIP bootstrap implementations. They verify handled SIGTERM, default SIGTERM, exit status 17, and existing ZIP behavior.

Fixes #3809

Testing:

* `bazel test --config=fast-tests //tests/py_zipapp:system_python_zipapp_test //tests/bootstrap_impls:run_binary_zip_yes_test //tests/bootstrap_impls:run_binary_bootstrap_script_zip_yes_test --runs_per_test=10 --test_output=errors`
* changed-file pre-commit hooks
* shell syntax checks for both modified shell templates

— Generated by GPT-5.6-Sol


aignas:
Thank you for the PR, please:
* Fix `buildifier` errors
* Windows does not have `signal.SIGKILL`, so making the tests UNIX only could be a way forward.

news/3809.fixed.md
```diff
@@ -0,0 +1 @@
+(zipapp) Forwarded termination signals to the application and preserved its exit status.
```
aignas, line 1:
```suggestion
(zipapp) Forwarded termination signals to the application and preserved its exit status.
([#3809](https://github.com/bazel-contrib/rules_python/issues/3809))
```

jpneufeld:
Addressed in `f816e365`:

* Moved the signal helpers and cases into a class guarded by `@unittest.skipUnless(os.name == "posix", ...)`. The existing cross-platform smoke test remains separate.
* Applied the issue-link suggestion to the news entry.
* Ran both Buildifier hooks across the repository; `buildifier` and `buildifier-lint` pass locally. The previous CI job failed before scanning the repository because its helper tried to download nonexistent Buildifier v10.0.0, despite the v8.2.1 configuration. This push starts a new CI run.

— James' Codex

news/3809.fixed.md
jpneufeld: Applied in `f816e365`.

— James' Codex

jpneufeld:
Follow-up `780723a7` fixes the Windows Pyrefly failure. The class-level skip prevents runtime execution, but Pyrefly still analyzes `signal.SIGKILL` against the Windows stubs. Cleanup now sends `SIGTERM`; both fixtures terminate after that signal.

The Pyrefly-enabled Bazel build, Ruff, and all three focused signal targets pass locally. Fresh CI is running.

— James' Codex

PR #4165, opened 2026-09-17
fix(zipapp): reduce bundled interpreter size

### LINKED ISSUE #3534: bin/python size exploded 2000 times larger

### LINKED ISSUE #4163: zipapp exploded the python interpreter
We are using python 3.10 and the static python3 interpreter binary is about 45MB, but zipapp packs it three times so we end up with 150MB just for the interpreter alone, making the zipapp binary bloated.

```
$ ls -l runfiles/rules_python++python+python_3_10_x86_64-unknown-linux-gnu/bin/python*
-rwxrwxr-x 1 koi koi 44667000 Jan  1  1980 python
-rwxrwxr-x 1 koi koi 44667000 Jan  1  1980 python3
-rwxrwxr-x 1 koi koi     3149 Jan  1  1980 python3-config
-rwxrwxr-x 1 koi koi 44667000 Jan  1  1980 python3.10
-rwxrwxr-x 1 koi koi     3149 Jan  1  1980 python3.10-config
```

Also the static interpreter wouldn't need the shared libraries, but that is also being packed into zipapp, increase the size by another 45MB:
```
$ ls -l runfiles/rules_python++python+python_3_10_x86_64-unknown-linux-gnu/lib/
-rwxrwxr-x  1 koi koi 45131088 Jan  1  1980 libpython3.10.so.1.0
-rwxrwxr-x  1 koi koi    16800 Jan  1  1980 libpython3.so
-rwxrwxr-x  1 koi koi  2340016 Jan  1  1980 libtcl9.0.so
-rwxrwxr-x  1 koi koi  3082472 Jan  1  1980 libtcl9tk9.0.so
...
```

gfrankliu:
Bazel reports source symlinks as regular files, causing Python executable aliases to be copied into zipapps as full binaries. Hermetic runtimes also include shared libpython artifacts even when the interpreter links Python statically.

Preserve source symlinks through sandbox indirection and exclude libpython shared objects from runtime files while retaining them for explicit native dependencies. Add regression coverage for archive structure and native extension loading.

Fixes #3534
Fixes #4163


gfrankliu:
The two failures in buildkite seems to relate to using python 3.13.1. The transition where Astral switched python-build-standalone to statically link libpython directly into the bin/python executable occurred in May 2025. The custom `install-only-stripped` artifact predates Astral’s May 2025 static-link transition, and that version of python 3.13.1 still dynamically links libpython, which is removed in this PR, causing test failure.

A few options:
1) revert the removal of libpython.so in my PR and live with the extra 250MB in zipapp.
2) use an Astral python build version after May 2025, eg: https://github.com/astral-sh/python-build-standalone/releases/tag/20250604 gives 3.14.0b1, 3.13.4, 3.12.11, 3.11.13, 3.10.18 and 3.9.23. So any of those versions, or higher would have statically linked python executable. We can change https://github.com/bazel-contrib/rules_python/blob/main/MODULE.bazel#L195 https://github.com/bazel-contrib/rules_python/blob/main/tests/toolchains/BUILD.bazel#L29 etc from 3.13.1 to 3.13.x (x>=4).
3) add some auto detection of which Astral version is used and then decide if libpython is needed. (not sure if this is possible).

Thoughts?

gfrankliu:
Implemented the Policy:

- Official Astral builds dated 20250604+: exclude libpython*.so*.
- Older, custom, or unknown builds: retain versioned libraries.

python/private/pbs_manifest.bzl
```diff
@@ -112,6 +118,27 @@ def parse_filename(filename):
         "vendor": vendor,
     }
 
+# buildifier: disable=function-docstring-args
+# buildifier: disable=function-docstring-return
```
aignas, line 122: This is saving a little bit of time instead of readability for later. Consider adding a comment that includes type-hints for the args, since this would be a trivial addition.

python/private/python_repository.bzl
```diff
@@ -280,13 +282,15 @@ define_hermetic_runtime_toolchain_impl(
   name = "define_runtime",
   extra_files_glob_include = {extra_files_glob_include},
   extra_files_glob_exclude = {extra_files_glob_exclude},
+    interpreter_has_static_libpython = {interpreter_has_static_libpython},
```
aignas, line 290: nit: Please either fix the indentation to 4 spaces for all or keep 2 spaces.

python/private/pbs_manifest.bzl
```diff
@@ -112,6 +118,27 @@ def parse_filename(filename):
...
+    if (
+        not build_version.isdigit() or
+        int(build_version) < _ASTRAL_STATIC_LIBPYTHON_RELEASE
+    ):
+        return False
+
+    return any([
+        url.startswith(prefix)
+        for url in urls
+        for prefix in _ASTRAL_RELEASE_URL_PREFIXES
```
aignas, line 140: Is the thinking here that only astral releases are supported and if someone is re-hosting this from their local, they will not be able to benefit from this?

I am thinking that it probably makes sense to expose a flag in the `python` extension to set the static library detection to a particular value. The extension could have `static_libpython_build="auto"`, "yes", "no" parameters where we use this function if it is `"auto"`. Do you think it is complex to set it?

python/private/python_repository.bzl
```diff
@@ -162,6 +163,7 @@ def _python_repository_impl(rctx):
         *python_version_info
     )
     urls = rctx.attr.urls or [rctx.attr.url]
+    interpreter_has_static_libpython = is_astral_static_libpython_build(urls, release_filename)
```
aignas:
```suggestion
    if rctx.attr.static_libpython_build == "auto":
        interpreter_has_static_libpython = is_astral_static_libpython_build(urls, release_filename)
    elif rctx.attr.static_libpython_build == "yes":
        interpreter_has_static_libpython = True
    else:
        interpreter_has_static_libpython = False
```

aignas (approved):
LGTM, but it would be good to expose a python extension configuration to disable the auto-detection. Please let me know if you would like to do this as part of this PR.

I am OK to just do `python_repository` attribute with an `"auto"` default where the user can very easily do a patch for now to override this if needed.

python/private/python_repository.bzl
gfrankliu: Resolved in 8303a1f

python/private/pbs_manifest.bzl
gfrankliu: Resolved in 2ab0e89

python/private/pbs_manifest.bzl
gfrankliu: Resolved in 2ab0e89

python/private/python_repository.bzl
gfrankliu: Resolved in 2ab0e89

gfrankliu:
It seems one of the CI jobs hangs, how to re-trigger it?

PR #4166, opened 2026-09-18
fix(cc): include toolchain libraries in cc_shared_library

dzbarsky:
`current_py_cc_libs` forwards linker inputs with the Python toolchain library's owner labels. `cc_shared_library` does not discover those labels through the toolchain, so it silently omits the static libraries. Return `CcSharedLibraryHintInfo` with the forwarded owner labels.

Validation: added a `rules_testing` regression with a static library supplied through `py_cc_toolchain`. It fails before the fix; all three `current_py_cc_libs` analysis tests pass afterward on Bazel 9.1.1.


aignas (approved):
Thanks!

